PR #1617, opened 2017-11-09
reduce Heading line-height, adjust relative size of sections in Setti…

abramin:
screen shot 2017-11-13 at 13 47 47

### DIFF
--- a/src/components/heading/heading.scss
+++ b/src/components/heading/heading.scss
@@ -31,8 +31,8 @@
 .carbon-heading__title {
   display: inline-block;
   font-weight: 900;
-  line-height: 22px;
-  margin: 0 0 10px 0;
+  line-height: 32px;
+  margin: 0;
   vertical-align: middle;
 }
 
--- a/src/components/heading/heading.scss
+++ b/src/components/heading/heading.scss
@@ -31,8 +31,8 @@
 .carbon-heading__title {
   display: inline-block;
   font-weight: 900;
-  line-height: 32px;
-  margin: 0;
+  line-height: 22px;
+  margin: 0 0 10px 0;
   vertical-align: middle;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/settings-row/settings-row.scss
```diff
@@ -31,12 +31,12 @@
   box-sizing: border-box;
   clear: both;
   float: left;
-  width: 25%;
+  width: calc(30% - 10px);
```
andrewjtait: since you've changed the other value to `65`, should this not be `35`? Perhaps with a larger margin than `10px`?

src/components/heading/heading.scss
```diff
@@ -31,7 +31,7 @@
 .carbon-heading__title {
   display: inline-block;
   font-weight: 900;
-  line-height: 32px;
+  line-height: 22px;
```
andrewjtait: Looks like reducing this has caused the text to get too close to the separator. Maybe it needs a bottom margin applied? 

This is what the spacing used to be like:

screen shot 2017-11-10 at 10 07 13

andrewjtait:
Looks like this has had an undesired effect on `Heading` in other components such as `Dialog`.

screen shot 2017-11-13 at 13 21 05

Maybe the CSS you need to change would be better aimed at the `SettingsRow` - and leaving `Heading` as it was? (seeing as though `SettingRow` overrides the font-size etc, it could also override the line-height and margin?

andrewjtait:
UX tested with @harpalsingh

andrewjtait:
Will postpone this to 2.4.0 as we need to cut 2.3.0

ianrobsonsage:
QA Passed.

Pointed GAC at the Carbon branch that includes the changes. Checked the Settings Row, Heading and Dialog Carbon components, all looking as expected.

toni-leigh:
For future reference, using `em` based line-heights might be worth considering?

PR #1620, opened 2017-11-14
Add data-state to modal

philearley:
# Description

Add data state to modal to aid in test running.
This change ensures the data state is updated once all transitions have fininshed.
Selenium was attempting to select dialogs before they had finished rendering on the page, causing false failures.

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -5,6 +5,7 @@ import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
 import Events from './../../utils/helpers/events';
 import Browser from './../../utils/helpers/browser';
 
+const TIMEOUT = 500;
 /**
  * A Modal Component
  *
@@ -102,8 +103,8 @@ class Modal extends React.Component {
     modal: PropTypes.object
   }
 
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
 
     /**
      * Tracks if event listeners are on for modal
@@ -112,7 +113,29 @@ class Modal extends React.Component {
      * @type {Boolean}
      */
     this.listening = false;
-    this.state = { state: 'default' };
+
+    this.state = {
+      /**
+       * Sets the initial data-state of the modal
+       * @property state
+       * @type {String}
+       */
+      state: this.props.open ? 'open' : 'closed'
+    };
+  }
+
+  /**
+   * Updates the value used for the css data-state. This uses a timeout to match the length of any transition to ensure
+   * UI automation is able to target elements once they are visually ready.
+   * @method updateDataState
+   * @return {void}
+   *
+   */
+  updateDataState = () => {
+    clearTimeout(this.openTimeout);
+    this.openTimeout = setTimeout(() => {
+      this.setState({ state: this.props.open ? 'open' : 'closed' });
+    }, TIMEOUT);
   }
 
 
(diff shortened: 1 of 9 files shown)


src/components/modal/modal.js
```diff
@@ -155,10 +156,14 @@ class Modal extends React.Component {
       // it was added as the Portal library we use messes up the callstack
       // when assigning the ref and triggering componentDidUpdate
       setTimeout(() => {
+        clearTimeout(this.openTimeout);
+        this.openTimeout = setTimeout(() => { this.setState({ state: 'open' }); }, 500);
```
andrewjtait: would it be safer to set the state depending on `this.props.open`.

eg. 

```
this.setState({ state: this.props.open ? 'open' : 'closed' })
```

src/components/modal/modal.js
```diff
@@ -112,6 +112,7 @@ class Modal extends React.Component {
      * @type {Boolean}
      */
     this.listening = false;
+    this.state = { state: 'default' };
```
andrewjtait: should this be `open` or `closed`? not sure what benefit you get from `default`?

src/components/modal/modal.js
andrewjtait: (if you do this you can pull it into a reusable function)

src/components/modal/modal.js
```diff
@@ -155,10 +156,14 @@ class Modal extends React.Component {
...
       // when assigning the ref and triggering componentDidUpdate
       setTimeout(() => {
+        clearTimeout(this.openTimeout);
+        this.openTimeout = setTimeout(() => { this.setState({ state: 'open' }); }, 500);
         this.onOpening; // eslint-disable-line no-unused-expressions
       });
       _window.addEventListener('keyup', this.closeModal);
     } else if (!this.props.open && this.listening) {
+      clearTimeout(this.openTimeout);
+      this.openTimeout = setTimeout(() => { this.setState({ state: 'closed' }); }, 500);
```
andrewjtait: could `500` be pulled out into a constant in the file? and shared between this and the animations?

src/components/modal/modal.js
philearley: We have used idle and default in other locations, it felt the correct thing to do, to initialise with default. Happy to set this to default as a closed if it makes more sense.

src/components/modal/modal.js
```diff
@@ -112,6 +113,12 @@ class Modal extends React.Component {
      * @type {Boolean}
      */
     this.listening = false;
+    this.state = { state: 'closed' };
```
andrewjtait: iit could be open by default. I'd also add some comments to this to explain what it is used for

src/components/modal/modal.js
```diff
@@ -112,6 +113,12 @@ class Modal extends React.Component {
...
+  }
+
+  updateState = () => {
```
andrewjtait: maybe add a description to explain what this functions purpose is?

src/components/modal/modal.js
```diff
@@ -112,6 +113,12 @@ class Modal extends React.Component {
...
+    clearTimeout(this.openTimeout);
+    this.openTimeout = setTimeout(() => { this.setState({ state: this.props.open ? 'open' : 'closed' }); }, TIMEOUT);
```
andrewjtait: maybe break this onto a few lines so its easier to read
```js
this.openTimeout = setTimeout(() => {
  this.setState({ state: this.props.open ? 'open' : 'closed' });
}, TIMEOUT);
```

src/components/modal/modal.js
```diff
@@ -155,11 +162,13 @@ class Modal extends React.Component {
       // it was added as the Portal library we use messes up the callstack
       // when assigning the ref and triggering componentDidUpdate
       setTimeout(() => {
+        this.updateState();
```
andrewjtait: `updateState` is a weird name inside of a React component - maybe it should be something more clear?

src/components/modal/modal.js
philearley: open is a defaultProp and is set to false.
What would you suggest to be the best wait to mirror the component that loads this?

src/components/modal/modal.js
```diff
@@ -148,18 +172,20 @@ class Modal extends React.Component {
    */
   componentDidUpdate() {
     const _window = Browser.getWindow();
+    // this.setState((prevState, props) => {
+    //   return { counter: prevState.counter + props.step };
+    // });
```
andrewjtait: comment

src/components/modal/modal.js
```diff
@@ -148,18 +172,20 @@ class Modal extends React.Component {
...
 
     if (this.props.open && !this.listening) {
       this.listening = true;
-      // TODO: try to remove this with React 16 upgrade with native Portals
-      // it was added as the Portal library we use messes up the callstack
-      // when assigning the ref and triggering componentDidUpdate
```
andrewjtait: this is still a valid `TODO`

philearley:
Problematic spec continues to pass after latest changes.

PR #1624, opened 2017-11-16
UX improvement to address padding in dialogs

philearley:
# Description
A UX request to reduce the spacing between the dialog header and the line spacer.

# Screenshots / Gifs
_**Original**_
screen shot 2017-11-16 at 11 31 26

_**New**_
screen shot 2017-11-16 at 11 31 45

# Testing
At desk demonstration done with UX.

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -105,6 +105,7 @@ $dialog-sizes: (
 
   .carbon-heading__header {
     margin-bottom: 0;
+    padding-bottom: 15px;
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 2.4.0
+
+## UX Improvements
+
+* `Dialog` Ensure the heading used within a dialog has 15px padding instead of 20px.
+
 # 2.3.0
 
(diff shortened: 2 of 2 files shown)


src/components/dialog/dialog.scss
```diff
@@ -105,6 +105,7 @@ $dialog-sizes: (
 
   .carbon-heading__header {
     margin-bottom: 0;
+    padding-bottom: 15px;
```
andrewjtait: Why not adjust the padding on the `Heading` component rather than on the Dialog's version of it?

harpalsingh: This is because as a Header on a page, the spacing was required to offset the gap on the top, and in a dialog, it is smaller, so the padding needs to be reduced, but this is specific in a dialog.

andrewjtait: OK - would have preferred to avoid contextual minor changes like this if possible as it makes the components more brittle to change, but if it is the way it has to be ✅

toni-leigh: If you look at the way display flex and display grid work they both change the layout of the children based on context provided by the immediate parent, so I think this is a reasonable thing to do but crucially just for layout styles.

PR #1627, opened 2017-11-23
css updates for button, message and base-text

andrewjtait:
## CSS Changes

* Adds an app wide `line-height` default of `18px`.
* Fixes `Message` to remove excess padding right when not dismissable.
* Updates `Button` so inline siblings have a default `10px` gap.

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -23,6 +23,10 @@
   text-align: center;
   text-decoration: none;
 
+  + .carbon-button {
+    margin-left: 10px;
+  }
+
   &:focus {
     @include transition(box-shadow linear 0.1s);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## CSS Changes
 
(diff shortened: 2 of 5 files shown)


lukeemmerson:
@andrewjtait Reckon we could go any higher with that line-height app-wide? 16px looks a tad cramped whereas 18px would be the sweet spot.

andrewjtait:
@lukeemmerson if that is the UX consensus then sure 👍  it would need a change in SOP as well. Let me know if confirmed.

lukeemmerson:
@andrewjtait just chatted to the team, we're happy to go ahead with 18px.

andrewjtait:
@lukeemmerson done

src/components/button/button.scss
```diff
@@ -23,6 +23,10 @@
   text-align: center;
   text-decoration: none;
 
+  + .carbon-button {
+    margin-left: 10px;
```
andrewjtait: need to investigate how this works when buttons collapse onto multiple lines.

Issue #1631, opened 2017-11-28
Parse response needs to handle a 204


toni-leigh:
Currently the response parsing in the service util can't handle a 204 which is a response with an empty body. We need to edit this to return an empty JSON object in a string is the response body is not present.

PR: https://github.com/Sage/carbon/pull/1632

andrewjtait:
Fixed https://github.com/Sage/carbon/pull/1632

PR #1632, opened 2017-11-28
Ensure response data is an object #1631

toni-leigh:
# Description

* We take an empty string in response data and convert it to an object
* This ensures 204 responses with empty data will still be correctly parsed

# TODO
- [x] Release notes

https://github.com/Sage/carbon/issues/1631

# Testing Instructions

### DIFF
--- a/src/utils/service/service.js
+++ b/src/utils/service/service.js
@@ -71,11 +71,6 @@ class Service {
    * @return {Object} the response data
    */
   handleSuccess(response) {
-    if (response.data === '') {
-      response.data = {};
-      return response.data;
-    }
-
     if (!response.data.message) {
       return response.data;
     }
@@ -260,7 +255,7 @@ class Service {
    * @return {Void}
    */
   responseTransform(response) {
-    return JSON.parse(response);
+    return JSON.parse(!response ? '{}' : response);
   }
 
   /**
--- a/src/utils/service/service.js
+++ b/src/utils/service/service.js
@@ -71,6 +71,11 @@ class Service {
    * @return {Object} the response data
    */
   handleSuccess(response) {
+    if (response.data === '') {
+      response.data = {};
+      return response.data;
+    }
+
     if (!response.data.message) {
       return response.data;
     }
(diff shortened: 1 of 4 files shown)


src/utils/service/service.js
```diff
@@ -71,6 +71,11 @@ class Service {
    * @return {Object} the response data
    */
   handleSuccess(response) {
+    if (response.data === '') {
+      response.data = {};
+      return response.data;
```
adamgeorgeson: Could you just `return {}`?

src/utils/service/service.js
```diff
@@ -71,6 +71,11 @@ class Service {
    * @return {Object} the response data
    */
   handleSuccess(response) {
+    if (response.data === '') {
```
SeanArmstrong: Could this be if `(!response.data) {`? Would then catch nulls and undefineds.

adamgeorgeson: Ah yes, had it this way when we spiked it in `responseTransform`

src/utils/service/service.js
```diff
@@ -255,7 +255,7 @@ class Service {
    * @return {Void}
    */
   responseTransform(response) {
-    return JSON.parse(response);
+    return JSON.parse(!response ? '{}' : response);
```
ianoxley: I'm not sure returning an empty object here is the right approach, as if there is no response body it could make it seem as though there is e.g. with an empty response body returning an empty object from `responseTransform` would evaluate to `truthy`.

Would it be better to allow `responseTransform` to return `undefined` when there is no response body?

toni-leigh: I think we need to do this in order for the response to be correctly parsed?

@adamgeorgeson might be able to shed some light, I picked this up quickly before the dentist so wasn't fully clear on the underlying issue!

adamgeorgeson: @ianoxley That seems fair. Will pick it up.

adamgeorgeson: Parking this for now, as returning `undefined` in `responseTransform` ends up requiring us to capture `response.data` being `undefined` in `handleSuccess` which would need to be handled correctly in order to ensure we trigger callbacks. Will have a catchup with @ianoxley and @andrewjtait in the morning.

adamgeorgeson: This has been addressed in 6c77470 @andrewjtait @ianoxley

PR #1635, opened 2017-12-07
Set Component.displayName in all decorators

ianoxley:
Set the `Component.displayName` in all decorators to that of the wrapped component. This fixes an issue found using Jest snapshots where snapshots were not using the correct component name.

Update Node.js and npm version used in Travis CI builds:

* Node.js is now `8.9.2`
* npm is now `5.5.1`

Update `package-lock.json` (automatically updated when running `npm install`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Fixes
 
-* Update Jest snapshots for `Alert`, `Dialog`, and `Sidebar`.
+* Set the `Component.displayName` on all decorators.
 
 # 2.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## Fixes
 
-* Set the `Component.displayName` on all decorators.
+* Set the `Component.displayName` on all decorators. **NB** You may need to update your snapshots as a result of this to change the component name to something more accurate
 
 # 2.4.0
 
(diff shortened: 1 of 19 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 * Changed type of prop `tooltipMessage` of `tooltip-decorator` from string to node to allow children.
 
+## Fixes
+
+* Set the `Component.displayName` on all decorators.
```
andrewjtait: You might want to highlight the fact that this may effect snapshots in peoples repos as they upgrade, to change the component name to something more accurate.

src/utils/decorators/input-label/__spec__.js
```diff
@@ -295,4 +295,35 @@ describe('InputLabel', () => {
...
+  describe('displayName', () => {
+    class Foo extends React.Component { // eslint-disable-line react/no-multi-comp
+      bar = () => {
+        return 'bar';
+      }
+    }
+
+    const displayName = 'FooClass';
+
+    describe('when ComposedComponent.displayName is defined', () => {
```
SeanArmstrong: [minor] We could probably move these specs into something more generic and pass in the decorator. e.g. how the data attribute specs are done

SeanArmstrong:
At desk demo'ed on Carbon Demo site. 

Regression tested decorators:

Input, Input Labels, Input Validation, input Icons, tooltips

PR #1636, opened 2017-12-08
Allow Portrait to render an empty alt attribute

ianoxley:
Update the `Portrait` component so that can render empty `alt` attributes: when it receives the empty string for the `alt` prop it will render it as an empty `alt` attribute on the `` tag.

Set the `defaultValue` for the `alt` prop to `''`.

### DIFF
--- a/src/components/portrait/portrait.js
+++ b/src/components/portrait/portrait.js
@@ -280,7 +280,7 @@ class Portrait extends React.Component {
         data-element='initials'
         className='carbon-portrait__img carbon-portrait__initials'
         src={ this.generateInitials }
-        alt={ this.props.alt || '' }
+        alt={ this.props.alt }
       />
     );
   }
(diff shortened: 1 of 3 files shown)


src/components/portrait/portrait.js
```diff
@@ -279,7 +280,7 @@ class Portrait extends React.Component {
         data-element='initials'
         className='carbon-portrait__img carbon-portrait__initials'
         src={ this.generateInitials }
-        alt={ this.props.alt }
+        alt={ this.props.alt || '' }
```
SeanArmstrong: Do we need this if its defaulted

ianrobsonsage:
QA Passed. Loaded the branch in Carbon Demo site. Confirmed that the img tag now includes an empty alt attirbute.

screen shot 2017-12-11 at 16 39 04

PR #1637, opened 2017-12-11
Update Portrait and Profile CSS

ianoxley:
Update the text colour for `Profile`,  and the background colour used for `darkBackground={ true }` in the `Portrait` component, to `$onyx`.

Update the `Portrait` component used by `Profile` to use `darkBackground={ true }`.

## Screenshots
profile

portrait

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Improvements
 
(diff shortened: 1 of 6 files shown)


ianrobsonsage:
Loaded Demo site, confirmed correct colours and font now being used.

QA Passed

PR #1638, opened 2017-12-12
Fix documentation for SettingsRow

ianoxley:
Change `description` type to `Node`, and add the missing prop type and description for `className`.


src/components/settings-row/__definition__.js
```diff
@@ -13,21 +13,23 @@ let definition = new Definition('settings-row', SettingsRow, {
 * Laying out a page in columns? [Try Row](/components/row).
  `,
   propTypes: {
-    children: "Node",
-    description: "String",
-    divider: "Boolean",
-    title: "String"
+    children: 'Node',
+    className: 'String',
+    description: 'Node',
```
SeanArmstrong: Does this still work correctly in the demo site?

src/components/settings-row/settings-row.js
```diff
@@ -60,9 +60,12 @@ class SettingsRow extends React.Component {
      * Heading description
      *
      * @property  description
-     * @type      {Node}
+     * @type      {String|Object}
      */
-    description: PropTypes.node,
+    description: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.object
```
andrewjtait: Just curious about this change? Isn't `node` more flexible?

ianoxley: I suppose so. And I suppose there is a risk of this introducing a breaking change.

However, would the `description` be set to anything other than a string or an JSX object?
And we use `PropTypes.oneOfType([PropTypes.string, PropTypes.object])` elsewhere e.g. in `Heading`: https://github.com/Sage/carbon/blob/master/src/components/heading/heading.js#L36

src/components/settings-row/__definition__.js
SeanArmstrong: It does if you add a span to the value

PR #1640, opened 2017-12-12
[CARBON-1007] - Upgrade enzyme and fixing resulting broken specs

krillatron:
# Description
Updating Enzyme to 3.2, with adapters for react-15.x. 
Also updating related now-breaking specs due to major version bump (https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md)

```

https://github.com/Sage/carbon/pull/1578

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,6 +9,7 @@ var deploy = require('./script/deploy').default;
 var setConfig = require('./script/set-config').default;
 var express = require('express');
 var gutil = require('gulp-util');
+const jestConfig = require('./jest.conf.json');
 
 var argv = yargs.argv;
 
@@ -44,10 +45,12 @@ gulp.task('build', BuildTask({
   imageDest: './' + dir + '/assets/images'
 }));
 
+var specOpts = { jestConfig };
+
 gulp.task('run-deploy', deploy);
 
 gulp.task('default', ['prepare-demo', 'webserver', 'build']);
 gulp.task('deploy', ['prepare-demo', 'build', 'run-deploy']);
 
-gulp.task('test', SpecTask());
+gulp.task('test', SpecTask(specOpts));
 gulp.task('lint', LintTask());
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -45,12 +45,10 @@ gulp.task('build', BuildTask({
   imageDest: './' + dir + '/assets/images'
 }));
 
-var specOpts = { jestConfig };
-
 gulp.task('run-deploy', deploy);
 
 gulp.task('default', ['prepare-demo', 'webserver', 'build']);
 gulp.task('deploy', ['prepare-demo', 'build', 'run-deploy']);
 
-gulp.task('test', SpecTask(specOpts));
+gulp.task('test', SpecTask({ jestConfig }));
 gulp.task('lint', LintTask());
--- a/jest.conf.json
+++ b/jest.conf.json
@@ -1,0 +1,7 @@
+{
+  "preset": "./node_modules/carbon-factory/jest.conf.json",
+  "notify": false,
+  "setupFiles": [
+    "./src/__spec_helper__/index.js"
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -1,16 +1,17 @@
 {
   "name": "carbon-react",
-  "version": "2.4.0",
+  "version": "2.5.2",
   "description": "A library of reusable React components and an interface for easily building user interfaces based on Flux.",
   "engineStrict": true,
   "engines": {
     "npm": "^5.0.0"
   },
   "scripts": {
+    "set-config": "node -e \"require('./script/set-config').default()\"",
     "test": "jest --config=./jest.conf.json",
     "debug": "node --inspect ./node_modules/.bin/jest --watch --config=./jest.conf.json",
     "release-version": "node check-version.js && npm install && npm publish",
-    "prepublishOnly": "rm -rf ./lib && node -e \"require('ncp').ncp('./src', './lib', { filter: new RegExp('^((?!__spec__.js|.md).)*$') }, function() { })\" && NODE_ENV=production babel --stage 0 ./src --out-dir ./lib --ignore '**/*/__spec__.js' --quiet"
+    "prepublishOnly": "npm run set-config && rm -rf ./lib && node -e \"require('ncp').ncp('./src', './lib', { filter: new RegExp('^((?!__spec__.js|.md).)*$') }, function() { })\" && NODE_ENV=production babel --stage 0 ./src --out-dir ./lib --ignore '**/*/__spec__.js' --quiet"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "carbon-react",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "description": "A library of reusable React components and an interface for easily building user interfaces based on Flux.",
   "engineStrict": true,
   "engines": {
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "test": "jest",
-    "debug": "node --inspect ./node_modules/.bin/jest --watch",
+    "debug": "node --inspect ./node_modules/.bin/jest --watch --config=./jest.conf.json",
     "release-version": "node check-version.js && npm install && npm publish",
     "prepublishOnly": "rm -rf ./lib && node -e \"require('ncp').ncp('./src', './lib', { filter: new RegExp('^((?!__spec__.js|.md).)*$') }, function() { })\" && NODE_ENV=production babel --stage 0 ./src --out-dir ./lib --ignore '**/*/__spec__.js' --quiet"
   },
@@ -29,10 +29,9 @@
   },
   "devDependencies": {
     "babel-standalone": "~6.17.0",
-    "carbon-factory": "v2.2.0",
+    "carbon-factory": "v3.0.0-rc1",
     "enzyme": "^3.2.0",
     "enzyme-adapter-react-15": "^1.0.5",
-    "enzyme-to-json": "^3.2.2",
     "express": "~4.14.0",
     "flux": "^3.1.1",
     "gulp": "~3.9.0",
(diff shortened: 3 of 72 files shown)

### CODE REVIEW

File: jest.conf.json [hunk 1]
```diff
@@ -0,0 +1,7 @@
+{
+  "preset": "./node_modules/carbon-factory/jest.conf.json",
+  "notify": false,
+  "setupFiles": [
+    "./src/__spec_helper__/index.js"
+  ]
```
andrewjtait: 👍  but you will need to also update the `package.json` scripts to tell the test commands where the jest config is

https://github.com/Sage/carbon/blob/master/package.json#L10-L11

will both need ` --config=./jest.conf.json` adding to it

src/components/alert/__snapshots__/__spec__.js.snap
```diff
@@ -1,81 +1,185 @@
...
+            <div
+              className="carbon-modal__background"
+              key=".0"
+            />
+          </CSSTransitionGroupChild>
+        </div>
+      </TransitionGroup>
+    </CSSTransitionGroup>
+  </div>
+</Alert>
```
andrewjtait: should avoid snapshotting mounted components, as it will create HTML tree of every single child component, whereas really all we care about is the top level (via shallow).

If we snapshot mounted components the snapshot will be incredibly brittle to changes in other components.

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "test": "jest",
```
andrewjtait: also need the `--config` adding to this line

src/components/confirm/__spec__.js
```diff
@@ -10,7 +10,7 @@ describe('Confirm', () => {
     onCancel = jasmine.createSpy('cancel');
     onConfirm = jasmine.createSpy('confirm');
 
-    wrapper = shallow(
+    wrapper = mount(
```
andrewjtait: ditto on snapshotting mounted components

src/components/dialog/__spec__.js
```diff
@@ -313,7 +313,7 @@ describe('Dialog', () => {
   describe('dialog headers', () => {
     describe('when a props title or subtitle is passed', () => {
       it('sets a dialog headers', () => {
-        const wrapper = shallow(
+        const wrapper = mount(
```
andrewjtait: ditto on snapshotting mounted components

src/utils/decorators/input-validation/__spec__.js
```diff
@@ -1213,11 +1213,13 @@ describe('InputValidation', () => {
 
       it('returns an info icon', () => {
         expect(instance.validationHTML[0].props.type).toEqual('info');
+        // TODO - 1007 - find not returning element though it is printed in the html()
         expect(wrapper.find('.common-input__icon.common-input__icon--info').exists()).toBeTruthy();
```
andrewjtait: not sure why it's not working - but you could do the same check we perform for other kinds of errors instead https://github.com/Sage/carbon/blob/4467a85bc8a72dfdce58b0f8d58c35b739735ceb/src/utils/decorators/input-validation/__spec__.js#L1114

src/utils/decorators/input-validation/__spec__.js
```diff
@@ -1213,11 +1213,13 @@ describe('InputValidation', () => {
...
       });
 
       it('returns a div for the info message', () => {
-        expect(wrapper.find('.common-input__message-wrapper').exists()).toBeTruthy();
+        // TODO - 1007 - find not returning element though it is printed in the html()
+        expect(wrapper.find('.common-input__message-wrapper').exists()).toEqual(true);
```
andrewjtait: again, there is another method used in this spec to check the class name (checking the prop rather than using enzyme).

src/components/table/__spec__.js
```diff
@@ -1019,20 +1021,22 @@ describe('Table', () => {
...
-      const wrapper = shallow(
+      const wrapper = mount(
         <Table theme='secondary' />
       );
       expect(wrapper.find('.carbon-table--secondary').exists()).toBeTruthy();
     });
 
     it("renders a --primary if the theme is missing (default)", () => {
-      const wrapper = shallow(
+      const wrapper = mount(
```
andrewjtait: Rather than changing all of these to `mount`, we could add an early return to this function if `this._table` does not exist https://github.com/Sage/carbon/blob/master/src/components/table/table.js#L747

src/components/table-ajax/__spec__.js
```diff
@@ -415,7 +421,9 @@ describe('TableAjax', () => {
   });
 
   describe("tags on component", () => {
-    let wrapper = shallow(
+    // TODO - 1007 - shallow seems to fail when creating Table
+    // data-component doesnt exist as prop on mount
+    let wrapper = mount(
```
andrewjtait: maybe this will fix the shallow render https://github.com/Sage/carbon/pull/1640/files#r156735929

src/components/rainbow/__spec__.js
```diff
@@ -63,7 +62,7 @@ describe('Rainbow', () => {
 
     describe('when data changes', () => {
       it("calls setData", () => {
-        const newData = Immutable.fromJS({ foo: "bar" });
+        const newData = fromJS({ foo: "bar" });
```
andrewjtait: I don't think this had to change, tried it on my machine and it works fine with what it was originally.

src/components/rainbow/__spec__.js
```diff
@@ -97,7 +96,7 @@ describe('Rainbow', () => {
       <Rainbow
         data-element='bar'
         data-role='baz'
-        data={ {} }/>
+        data={ fromJS({}) }/>
```
andrewjtait: however this test fails without first ensure it is immutable 👍

src/components/form/__spec__.js
```diff
@@ -137,7 +138,7 @@ describe('Form', () => {
 
   describe('checkStickyFooter', () => {
     beforeEach(() => {
-      wrapper = shallow(<Form />);
+      wrapper = mount(<Form />);
```
andrewjtait: is this to do with the use of refs again? (`this._form`) - there may be a way we can get around not having to use `mount`.

src/components/rainbow/__spec__.js
JohnStuartBarrie: Yes you are right on this. Just changed it back.

src/components/rainbow/__spec__.js
JohnStuartBarrie: This must have been failing silently before. It has to be immutable in the rainbow.

function generateConfig(immutableData, title) {
  const data = immutableData.toJS();

src/components/alert/__spec__.js
```diff
@@ -12,7 +12,7 @@ describe('Alert', () => {
   beforeEach(() => {
     wrapper = shallow(
       <Alert
-        open
+        open={ false }
```
andrewjtait: why'd you change this? It looks like it has affected the snapshot

src/components/confirm/__spec__.js
```diff
@@ -12,7 +12,7 @@ describe('Confirm', () => {
 
     wrapper = shallow(
       <Confirm
-        open
+        open={ false }
```
andrewjtait: again why has this changed?

src/components/dialog/dialog.js
```diff
@@ -233,6 +234,7 @@ class Dialog extends Modal {
   }
 
   focusDialog() {
+     if (!this._dialog) { return; }
```
andrewjtait: one too many space in front if `if` - our linter should pick this up. Can you try running `gulp lint` on the code base?

src/utils/helpers/element-resize/element-resize.js
```diff
@@ -20,6 +20,7 @@ const ElementResize = {
    * @param {Function} fn - the callback
    */
   addListener: (element, callback) => {
+    if (! element) { return };
```
andrewjtait: one too many spaces between ! and element

src/components/alert/__spec__.js
krillatron: When open is left as the default (set to true) it ends up calling modalHTML in Dialog (which Alert inherits from (which is actually called from the Modal render() action)) which uses a 'ref', which shallow doesn't have access to. 
The two paths appeared to be either:  Use mount rather than shallow for the alert, but then we'd be snapshotting a mount, or set open to false so it doesn't attempt to render the Alert out. 
It looks like this is not causing the issue though of we're not actually testing the Alert in a meaningful way via. the snapshot any more.
Would you suggest we actually snapshot the Alert but with mount rather than shallow in this case? Or have any other suggestions of how to move forward with these ones?

src/components/confirm/__spec__.js
JohnStuartBarrie: Sorry We will change this back. We were fiddling around and it got left in place.

src/components/alert/__spec__.js
JohnStuartBarrie: Actually we've changed  these back to having the Alert open.

src/components/confirm/__spec__.js
JohnStuartBarrie: Actually we've changed  these back to having the Confirm open.

src/components/dialog/dialog.js
JohnStuartBarrie: My bad. I just changed editors and I am not visually picking up on the lint errors.

gulpfile.js
```diff
@@ -44,10 +45,12 @@ gulp.task('build', BuildTask({
   imageDest: './' + dir + '/assets/images'
 }));
 
+var specOpts = { jestConfig };
```
ianoxley: Looks like this could be a `const` instead of a `var`.

src/components/table/__spec__.js
```diff
@@ -14,7 +14,7 @@ describe('Table', () => {
     spy = jasmine.createSpy('onChange spy');
 
     row = (
-      <TableRow>
+      <TableRow >
```
ianoxley: [Minor] Remove space

andrewjtait:
tested components that have changed in demo site ✅

PR #1643, opened 2017-12-14
fixes transforms on service class

andrewjtait:
Adding a transform to an instance of a service class was applying the transform to all other instances. This change ensures the transform is only applied to the specific instance.


andrewjtait:
Tested by @melaniesmith29 and @paulregnart ✅

PR #1654, opened 2017-12-21
updated superagent and moment for security fixes

andrewjtait:
* Superagent has been updated to v3.8.2 for a security fix
* moment has been updated to v2.20.1 for a security fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "immutable": "~3.8.1",
     "lodash": "~4.15.0",
     "marked": "~0.3.6",
-    "moment": "~2.15.1",
+    "moment": "~2.20.1",
     "prop-types": "^15.5.8",
     "react-day-picker": "~6.1.0",
     "react-dnd": "~2.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "immutable": "~3.8.1",
     "lodash": "~4.15.0",
     "marked": "~0.3.6",
-    "moment": "~2.15.1",
+    "moment": "^2.20.1",
```
ianoxley: Do we want to change from `~` to `^`?

ianoxley: Do we want to change from `~` to `^`?

package.json
```diff
@@ -74,6 +74,6 @@
     "react-portal": "^3.1.0",
     "react-router": "^3.0.0",
     "react-transition-group": "~1.1.3",
-    "superagent": "~2.2.0"
+    "superagent": "^3.8.2"
```
ianoxley: As above: do we want to change from `~` to `^`?

PR #1655, opened 2017-12-21
max-width on setting-row

andrewjtait:
Sets a max-width on the left hand segment of the SettingRow component, to ensure the content has the majority of the width of larger screens.

### Before

screen shot 2017-12-21 at 14 10 34

### After
screen shot 2017-12-21 at 14 09 36


andrewjtait:
UX reviewed with @benwilsonsage ✅

andrewjtait:
Tested in App and Demo Site ✅

PR #1660, opened 2018-01-04
fixes cursor type on menu items

andrewjtait:
Menu Items were displaying the wrong cursor when hovering them.


andrewjtait:
At desk tested with @ianrobsonsage  ✅

PR #1667, opened 2018-01-10
Carbon 1025 react portal upgrade

JohnStuartBarrie:
# Description
Use React v16 portal functionality

-Ticket No: Carbon 1025 

# Screenshots / Gifs

# Testing Instructions
gulp test


src/components/portal/portal.js
```diff
@@ -1,8 +1,14 @@
 import React from 'react';
-import { Portal as ReactPortal }  from 'react-portal';
 import PropTypes from 'prop-types';
+import { createPortal } from 'react-dom';
 
+const isDomAvailable = !!(
```
andrewjtait: This would be a useful function to belong on our browser helper https://github.com/Sage/carbon/blob/master/src/utils/helpers/browser/browser.js

src/components/portal/portal.js
```diff
@@ -18,16 +24,40 @@ class Portal extends React.Component {
...
+     *
+     * @property node
+     * @type {Any}
+     */
+    node: PropTypes.any
+  }
+
+  componentWillUnmount() {
+    if (this.defaultNode) {
+      document.body.removeChild(this.defaultNode);
```
andrewjtait: Use `getDocument` https://github.com/Sage/carbon/blob/master/src/utils/helpers/browser/browser.js#L28

src/components/portal/portal.js
```diff
@@ -18,16 +24,40 @@ class Portal extends React.Component {
...
+    this.defaultNode = null;
   }
 
   render() {
+    if (!this.props.open && !isDomAvailable) {
+      return null;
+    }
+
+    if (!this.props.node && !this.defaultNode) {
+      this.defaultNode = document.createElement('div');
```
andrewjtait: Use `getDocument` https://github.com/Sage/carbon/blob/master/src/utils/helpers/browser/browser.js#L28

src/components/portal/portal.js
```diff
@@ -18,16 +24,40 @@ class Portal extends React.Component {
...
+    }
+
+    this.node = (this.props.node) ? this.props.node : this.defaultNode;
     return (
-      this.props.open && <ReactPortal>
-        <div ref={ (node) => { this._portal = node; } }>
-          { this.props.children }
-        </div>
-      </ReactPortal>
+      this.props.open && createPortal(
```
andrewjtait: Let's think about removing the `open` prop.

andrewjtait:
@JohnStuartBarrie for the changes in the package-lock I'd try this trick https://github.com/npm/npm/issues/16938#issuecomment-339863980

src/utils/helpers/browser/browser.js
```diff
@@ -11,6 +10,13 @@ import Form from './../../../components/form';
 *
 */
 const Browser = {
+  isDomAvailable: () => {
+    return (!!(
+      typeof window !== 'undefined' &&
```
andrewjtait: Should use `Browser.getWindow()` or `global.window` at least.

Will need a test for this as well.

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -339,7 +339,7 @@ const TooltipDecorator = (ComposedComponent) => {
     get tooltipHTML() {
       if (!this.props.tooltipMessage) { return null; }
       return (
-        <Portal open={ this.state.isVisible }>
+        this.state.isVisible && <Portal>
```
andrewjtait: maybe this could be combined with line 340, something like

```js
if (!this.props.tooltipMessage || !this.state.isVisible) return null;
```

src/components/portal/portal.js
```diff
@@ -1,35 +1,63 @@
...
+  componentWillUnmount() {
+    if (this.defaultNode) {
+      Browser.getDocument().body.removeChild(this.defaultNode);
+    }
+    this.defaultNode = null;
+  }
+
+  getDefaultDiv() {
+    
+    if (!this.props.node && !this.defaultNode && Browser.isDomAvailable()) {
```
andrewjtait: would `Browser.isDomAvailable()` ever be `false` here seeing as you check on the first line of the `render` function that it is `true`?

andrewjtait: Similarly would `this.props.node` ever be `true`, seeing as you're testing it in the `render` too?

src/components/portal/portal.js
```diff
@@ -1,35 +1,63 @@
...
+    }
+    return this.defaultNode;
   }
 
   render() {
+    if (!Browser.isDomAvailable()) {
+      return null;
+    }
+
+    this.node = (this.props.node) ? this.props.node : this.getDefaultDiv();
```
andrewjtait: is there any benefit from storing this on `this.node`?

src/components/portal/portal.js
andrewjtait: Maybe you could clean up this confusion by getting rid of the ternary operator in the `render` function and having this function just return whatever the correct node is (whether it creates one, uses the memoized one, or uses the one from `props.node`).

src/components/portal/portal.js
```diff
@@ -1,35 +1,63 @@
...
+    }
+    this.defaultNode = null;
+  }
+
+  getDefaultDiv() {
+    
+    if (!this.props.node && !this.defaultNode && Browser.isDomAvailable()) {
+      this.defaultNode = Browser.getDocument().createElement('div');
+      this.defaultNode.classList.add('portal');
+      document.body.appendChild(this.defaultNode);
```
andrewjtait: retrieve document from `Browser.getDocument` and use it here too.

src/components/portal/portal.js
```diff
@@ -1,35 +1,63 @@
...
-        <div ref={ (node) => { this._portal = node; } }>
-          { this.props.children }
-        </div>
-      </ReactPortal>
+      <ReactPortal node={ this.node }>{this.props.children}</ReactPortal>
     );
   }
 }
 
+function ReactPortal(props) {
```
andrewjtait: any particular reason you've done this as a separate component, rather than just using `createPortal` method in the `render` function?

src/components/portal/portal.js
JohnStuartBarrie: Yes I could feel there was something sloppy here. in the general confusion, I forgot to come back to it.

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,74 @@
...
+    beforeEach(() => {
+      wrapper = shallow(
+        <Portal>
+          <p className='child-element' />
+        </Portal>
+      );
+    });
+
+    it('div with name portal should be created ', () => {
+      expect(wrapper.instance().node.classList[0]).toEqual('portal');
```
andrewjtait: snapshot probably covers this test?

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,74 @@
...
+    });
+
+    it('to match snapshot ', () => {
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('will unmount correctly', () => {
+      expect(wrapper.unmount());
```
andrewjtait: should you test that it cleans up the DOM?

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,74 @@
...
+        </Portal>
+      );
+    });
+
+    it('to match snapshot ', () => {
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('the node prop has been added to the DOM', () => {
+      expect(wrapper.instance().node.classList[0]).toEqual('my-node');
```
andrewjtait: not sure what this is testing other than it is assigning your custom element to the instance?

Maybe it should test that the portal has mounted the child components within your custom node?

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,74 @@
...
+    });
+
+    it('the node prop has been added to the DOM', () => {
```
andrewjtait: duplicate

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,74 @@
...
+      expect(wrapper.instance().node.classList[0]).toEqual('my-node');
+    });
+
+    it('will unmount correctly', () => {
+      expect(wrapper.unmount());
```
andrewjtait: this should test that it does not clean up the DOM

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
JohnStuartBarrie: How about ?
```js
(this.props.tooltipMessage && this.state.isVisible)  && <Portal>
```

src/components/icon/__spec__.js
```diff
@@ -160,24 +162,21 @@ describe('Icon', () => {
...
           tooltipAlign='left'
           tooltipPosition='top'
           type='tick'
         />);
-
-      elementsTagTest(wrapper, [
-        'tooltip'
-      ]);
+      wrapper.setState({ isVisible: true });
+      expect(document.getElementsByTagName('body')[0].innerHTML).toEqual(expect.stringContaining('data-component="tooltip" data-element="tooltip"'));
```
andrewjtait: need to find a better way of testing this

src/components/portal/portal.js
JohnStuartBarrie: I think this is my ignorance speaking. I seem to have unstood that we wanted to be able to access the Portal html node for testing purposes. There may be another way of access this through Enzyme

src/components/portal/portal.js
JohnStuartBarrie: Opps. Done

src/components/icon/__spec__.js
andrewjtait:
```js
const tooltip = Icon.find(Tooltip);
elementsTagTest(tooltip, [
  'tooltip'
]);
```
??

src/components/portal/portal.js
andrewjtait: looks like `wrapper.find` should work now?

JohnStuartBarrie: I suppose that I should take it out and we will see the need when it arises.

src/utils/helpers/browser/browser.js
```diff
@@ -11,6 +10,13 @@ import Form from './../../../components/form';
 *
 */
 const Browser = {
+  isDomAvailable: () => {
+    return (!!(
+      typeof Browser.getWindow() !== 'undefined' &&
```
andrewjtait: could you store the result of the function:

```js
return !!(
  typeof _window !== 'undefined' &&
  _window.document &&
  _window.document.createElement
);
```

src/utils/helpers/browser/__spec__.js
```diff
@@ -29,6 +29,12 @@ describe('Browser', () => {
     });
   });
 
+  describe('isDomAvailable', () => {
+    it('returns the window object', () => {
+      expect(Browser.isDomAvailable()).toBe(true);
```
andrewjtait: This is only testing a single branch, whereas the code offers 3 conditions - should try and test each (is this not why your coverage is failing?)

src/components/icon/__spec__.js
JohnStuartBarrie: This was the test I was talking earlier. Yes not good. Your solution looks like the way to go but, inside the function `elementsTagTest` tries `wrapper.find({ 'data-element': element }).length` and return ,'2' for the length so it fails. Any idea of why  it would do that?

src/utils/helpers/browser/browser.js
JohnStuartBarrie: Actually the problem with storing it is that it is a permanent value that doesn't reflect the real state of the environment. Also for simple testing reason where we are 'interfering' with the env it to make this function send false, we need thsi value to reflect the new state.

andrewjtait: I just meant within the scope of the function, so it's not recalling the same function 3 times at once.

src/components/portal/__spec__.js
andrewjtait: We should test what we expect it to do though?

src/utils/helpers/browser/__spec__.js
JohnStuartBarrie: yes I agree but Browser.getWindow() is a static so I can't stub it their window and document.
So can't really test those branches. Am I missing something?

src/components/portal/__spec__.js
JohnStuartBarrie: Sorry of course. This was an mistake. I was obviously tired on Friday. Oops.

src/utils/helpers/browser/__spec__.js
andrewjtait:
```
spyOn(Browser, 'getWindow').and.returnValue(undefined);
```
should help

andrewjtait: something along the lines of

```
describe('when window does not exist', () => {
  it('returns false', () => {
    spyOn(Browser, 'getWindow').and.returnValue(undefined);
    expect(Browser.isDomAvailable()).toEqual(false);
  });
});

describe('when document does not exist', () => {
  it('returns false', () => {
    spyOn(Browser, 'getDocument').and.returnValue(undefined);
    expect(Browser.isDomAvailable()).toEqual(false);
  });
});

describe('when document.createElement does not exist', () => {
  it('returns false', () => {
    spyOn(Browser, 'getDocument').and.returnValue({ createElement: undefined });
    expect(Browser.isDomAvailable()).toEqual(false);
  });
});
```

JohnStuartBarrie: Nice. I didn't realise that you could do that with imported objects.

src/components/icon/__spec__.js
```diff
@@ -160,24 +162,24 @@ describe('Icon', () => {
...
           tooltipPosition='top'
           type='tick'
         />);
+      wrapper.setState({ isVisible: true });
 
-      elementsTagTest(wrapper, [
-        'tooltip'
-      ]);
+      const tooltip = wrapper.find(Tooltip);
+      expect(tooltip.find({ 'data-element': 'tooltip' }).length).toEqual(2);
```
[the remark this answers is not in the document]
JohnStuartBarrie: Makes a lot more sense. I was blindly following.

src/utils/helpers/browser/__spec__.js
JohnStuartBarrie: Didn't think they were truely static in Javascript. Live, Love and Learn !

src/components/portal/__spec__.js
```diff
@@ -1,27 +1,97 @@
...
-  beforeEach(() => {
-    wrapper = shallow(
-      <Portal open>
-        <p className='child-element'></p>
+    afterEach(() => {
+      wrapper.unmount();
+    });
+
+    it('will mount correctly on document', () => {
+      expect(document.body.innerHTML).toEqual('<div class="carbon-portal"><span class="carbon-icon icon-tick" type="tick" data-component="icon"></span></div>');
```
andrewjtait: not sure if it will, but would snapshot work for this?

src/utils/helpers/browser/browser.js
```diff
@@ -11,6 +10,15 @@ import Form from './../../../components/form';
 *
 */
 const Browser = {
+  isDomAvailable: () => {
+    const _window = Browser.getWindow();
+    const _document = Browser.getDocument();
+    return (!!(
+      typeof _window !== 'undefined' &&
+      _document &&
+      _document.createElement
```
SeanArmstrong: Could you use `_document.hasOwnProperty('createElement')` here instead of the `!!`? This is a question rather than a request to change. There isn't really any difference in them

andrewjtait: Think the `!!` is still required to ensure it is a boolean if window or document are undefined.

To be honest this code was taken from the third party lib we used to use https://github.com/tajo/react-portal/blob/master/src/utils.js

andrewjtait:
Will QA this as part of the React upgrade branch.

PR #1669, opened 2018-01-15
Fix decimal precision in react upgrade

abramin:
# Description

*Cap precision at 20 and log warning if set higher

* replace double quotes in spec

### DIFF
--- a/src/components/decimal/__spec__.js
+++ b/src/components/decimal/__spec__.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import I18n from "i18n-js";
+import I18n from 'i18n-js';
 import { shallow } from 'enzyme';
 import TestUtils from 'react-dom/test-utils';
 import Decimal from './decimal';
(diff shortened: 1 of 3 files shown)


src/components/decimal/__spec__.js
```diff
@@ -393,24 +386,24 @@ describe('Decimal', () => {
...
       });
-    });
 
-    describe("handleOnClick", () => {
+    describe('handleOnClick', () => {
       let visible, selectionSpy;
 
       beforeEach(() => {
-        selectionSpy = jasmine.createSpy();
+        debugger
```
ianoxley: Stray `debugger`

src/components/decimal/decimal.js
```diff
@@ -111,14 +107,23 @@ class Decimal extends React.Component {
     precision: 2
   };
 
+  // Caps precision at 20 to prevent browser issues
+  verifiedPrecision = (precision) => {
+    if (precision > 20) {
+      console.warn('Decimal precision must not be greater than 20.'); // eslint-disable-line no-console
```
andrewjtait: should use our logger for this, then you don't need the eslint disable and it won't get output in production mode

andrewjtait:
Will QA this as part of the React upgrade branch.

PR #1673, opened 2018-01-18
Date component now keeps an invalid value and throw a validation error on the input.

adrianpastur:
# Description

Previously this component would not retain an invalid date value, we now keep the value and throw a validation error on the input.

# Screenshots / Gifs

carbon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 3.0.0
+
+## Bug fixes
+
+* `Date`: Previously this component would not retain an invalid date value, we now keep the value and throw a validation error on the input.
+
 # 2.6.4
 
 * Upgrade marked package from v0.3.6 to 0.3.9 to address security vulnerabilities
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.0.0
```
SeanArmstrong: Is this intentional? Version 3.0.0 will come with a React upgrade and therefore may be better as a Minor version. @andrewjtait

andrewjtait: @adrianpastur was unsure how quickly they need the fix, so just targeting the next planned release unless they decide otherwise

adrianpastur: We have had no reports from customers about this issue so there is no need for a quick release.

andrewjtait:
Had to push up a commit to resolve merge conflicts, so created a new PR https://github.com/Sage/carbon/pull/1677

PR #1674, opened 2018-01-18
CheckBox inline bug

JohnStuartBarrie:
# Description
Checkbox should not cover the last two letters and the full stop of the Help field text (see screenshots).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes
 
-* Checkbox no longer overlays the end of the Help field text when the Reverse box is ticked
+* Checkbox no longer overlays the end of the Help field text when the reverse prop is set to true
 
 # 2.6.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 3.0.0
+
+## Bug Fixes
+
+* Checkbox no longer overlays the end of the Help field text when the Reverse box is ticked
+
 # 2.6.4
 
 * Upgrade marked package from v0.3.6 to 0.3.9 to address security vulnerabilities
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -98,6 +98,7 @@
     display: inline;
     margin-right: 5px;
     position: relative;
+    margin-left: 0;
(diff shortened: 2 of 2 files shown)


src/components/checkbox/checkbox.scss
```diff
@@ -98,7 +98,7 @@
     display: inline;
     left: 21px;
     margin-left: -21px;
```
andrewjtait: It looks like the `left` and `margin-left` properties may be the issue here, doesn't make much sense why they would be overriding each other - and the `left` prop is probably what is throwing the input off.

Would it work if you left `margin-right` as `5px`, changed `margin-left` to `0`, and deleted `left`?

JohnStuartBarrie: I tried this actually and discover the left seemed important for something. Can't remember what. I will have another look.

src/components/checkbox/checkbox.scss
```diff
@@ -96,10 +96,9 @@
 
   &.carbon-checkbox__help-text--inline {
     display: inline;
-    left: 21px;
-    margin-left: -21px;
-    margin-right: 5px;
+    left: 0;
```
andrewjtait: screen shot 2018-01-19 at 15 33 28

setting `left: 0` and removing `margin-left` breaks the non-reversed version. Looks like it works if you remove `left` and set `margin-left` to `0`.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.0.0
+
+## Bug Fixes
+
+* Checkbox no longer overlays the end of the Help field text when the Reverse box is ticked
```
ianoxley: Minor - replace 'when the Reverse box is ticked' with 'when the `reverse` prop is set to `true`.

src/components/checkbox/checkbox.scss
JohnStuartBarrie: Maybe I am blind (but I can't see what broke).I removed both. We should need either. right?

andrewjtait: The gap between the icon and the help text is too wide

andrewjtait: `margin-left: 0` ensures it is correct

PR #1677, opened 2018-01-22
Date component now keeps an invalid value and throw a validation error on the input.

andrewjtait:
# Description

Previously this component would not retain an invalid date value, we now keep the value and throw a validation error on the input.

# Screenshots / Gifs

carbon


andrewjtait:
✅

PR #1678, opened 2018-01-23
Carbon 968 portal input validation

JohnStuartBarrie:
# Description
Modify the input validation component to work with the new Portal


src/components/portal/portal.scss
```diff
@@ -0,0 +1,5 @@
+.carbon-portal {
+  z-index: 1002;
+  position: absolute;
+  top: 0;
```
andrewjtait: how come we need the CSS for this? we didn't need it for datepicker?

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -633,9 +625,22 @@ const InputValidation = (ComposedComponent) => {
         iconStyle = { [`${this.props.align}`]: `${100 - this.props.labelWidth}%` };
       }
 
-      if (this.state.messageLocked) { messageClasses += ' common-input__message--locked'; }
```
andrewjtait: do we not need this class any more?

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -633,9 +625,22 @@ const InputValidation = (ComposedComponent) => {
...
       if (this.flipped) { messageClasses += ' common-input__message--flipped'; }
 
+      this.errorMessage = (this.state.messageShown || this.state.messageLocked) &&
```
andrewjtait: this could just be `const`, rather than assigning it to the instance.

src/utils/decorators/input-validation/input-validation.js
JohnStuartBarrie: I provoked the question with the rename. I just renamed it to ```common-input__message--shown``` because I could see no reason why it shouldn't be in the same css state if it is shown or locked.

src/components/portal/portal.scss
JohnStuartBarrie: DatePicker does have css in ../component/date/date.scss 
Suppose we could do this input-validation css but shouldn't we generalise this for the Portal. Would we ever need the Portal anywhere else than position 0x0

andrewjtait: that's a good point, maybe we could generalise this a bit. Does that mean we could remove that CSS from datepicker too?

Might be good to use the `z-index` sass file https://github.com/Sage/carbon/blob/master/src/style-config/z-index.scss

src/components/button-toggle-group/__spec__.js
```diff
@@ -141,10 +153,34 @@ describe('ButtonToggleGroup', () => {
...
+          <ButtonToggleGroup
+            label={ 'Label' }
+            value={ 'foo' }
+          >
+            <ButtonToggle>Foo</ButtonToggle>
+          </ButtonToggleGroup>
+        );
+      });
+  
+      afterEach(() => {
```
abramin: Could the three lines around Timers be called once in a top level beforeEach?

src/utils/decorators/input-validation/__spec__.js
```diff
@@ -393,46 +419,106 @@ describe('InputValidation', () => {
...
+            expect(wrapper.find(Component).instance().validationMessage.className).toEqual(expect.stringContaining('common-input__message--flipped'));
+            expect(wrapper.find(Component).instance().validationMessage.style.left).toEqual('610px');
+            expect(wrapper.find(Component).instance().validationMessage.style.top).toEqual('50px');
           });
         });
 
         describe('when offscreen', () => {
+          let wrapper;
+          beforeEach(() => {
+            jest.useFakeTimers();
```
abramin: ditto - could the timers code be moved up the hierarchy?

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -254,45 +253,39 @@ const InputValidation = (ComposedComponent) => {
...
+            } else {
+              message.style.left = `${icon.getBoundingClientRect().left + (icon.getBoundingClientRect().width / 2)}px`;
+              message.style.top = `${(icon.getBoundingClientRect().top - message.getBoundingClientRect().height)
+                                      - (icon.getBoundingClientRect().height)}px`;
+              message.classList.remove('common-input__message--flipped');
+              this.flipped = false;
+            }
           }
-        }
+        }, 0);
```
abramin: unnecessary `0`

src/components/button-toggle-group/__spec__.js
JohnStuartBarrie: Yes I should have tidied this up. It's done now

src/utils/decorators/input-validation/__spec__.js
JohnStuartBarrie: ditto

src/utils/decorators/input-validation/input-validation.js
JohnStuartBarrie: done

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -633,9 +625,21 @@ const InputValidation = (ComposedComponent) => {
         iconStyle = { [`${this.props.align}`]: `${100 - this.props.labelWidth}%` };
       }
 
-      if (this.state.messageLocked) { messageClasses += ' common-input__message--locked'; }
       if (this.flipped) { messageClasses += ' common-input__message--flipped'; }
 
+      const errorMessage = (this.state.messageShown || this.state.messageLocked) &&
```
andrewjtait: I think this condition here is what is breaking the animation

iliaoni:
https://jira-sage.valiantyscloud.net/browse/PF-3708 Needs resolving. Will continue QA when fixed.

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -246,6 +248,12 @@ const InputValidation = (ComposedComponent) => {
       }
     }
 
+    windowChanged = () => {
+      this.positionMessage();
+      // render to to add any eventual css rules
+      this.forceUpdate();
```
andrewjtait: This seems like it would be expensive. What if we changed `this.flipped` to state, and called `setState` when it changed? If not then it may be less expensive to control the flipped classes as part of the `positionMessage` function like we used to.

JohnStuartBarrie: Good point

PR #1679, opened 2018-01-23
Carbon 996 portal modal

JohnStuartBarrie:
# Description
Modal portal


src/components/dialog-full-screen/dialog-full-screen.js
```diff
@@ -134,14 +134,13 @@ class DialogFullScreen extends Modal {
...
       <FullScreenHeading>
+        { title }
         <Icon
           className='carbon-dialog-full-screen__close'
           data-element='close'
           onClick={ this.props.onCancel }
           type='close'
         />
-
-        { title }
```
andrewjtait: what was the reason for this change? does it effect the UI at all?

src/components/dialog/dialog.scss
```diff
@@ -37,8 +37,11 @@ $dialog-sizes: (
   box-shadow: 0 10px 20px rgba($grey-shadow, 0.2);
   position: fixed;
   top: 50%;
-  z-index: $z-dialog;
-
+  min-width: 664px;
+  max-width: 800px;
+  width: calc(50%);
+  height: calc(80%);
```
andrewjtait: what was the reason for adding these?

JohnStuartBarrie: The dialog div wasn't narrowing. There was nothing with a width above it in the html hierarchy.

andrewjtait: This appears to break the width/height of dialogs.

andrewjtait: Try the `size` prop.

src/components/dialog-full-screen/dialog-full-screen.js
JohnStuartBarrie: The icon has to be above the title bar. I couldn't see why this worked before tbh.

src/components/dialog/dialog.scss
JohnStuartBarrie: I think I must have 'fiddled' with the size prop by accident and wondering why the width had change. I think we do however need the height. This might have already been an issue before when the dialog content height was bigger than the screen height

src/components/dialog/dialog.scss
```diff
@@ -37,8 +37,8 @@ $dialog-sizes: (
   box-shadow: 0 10px 20px rgba($grey-shadow, 0.2);
   position: fixed;
   top: 50%;
-  z-index: $z-dialog;
-
+  height: calc(95%);
```
andrewjtait: this breaks the height of the dialogs

screen shot 2018-02-01 at 11 41 33

andrewjtait:
✅  Tested in demo site and GAC app.

PR #1685, opened 2018-01-29
Add deferredOnChange callback to inputs

### LINKED ISSUE #1199: Deferred Action
Could we add a deferred action to inputs, which will execute an action after a short delay and will defer further if the component detects another user action.

```js
<Textbox
  value={ this.data.get('value') }
  onChange={ Actions.handleChange }
  onChangeDeferred={ Actions.deferredCallback }
  deferTimeout="200" // defer length (defaults to 200 or something)
/>
```

whomwah:
Fixes #1199 by adding 2 new optional input props:

`deferTimeout` - timeout value for `setTimeout` triggered on `onChange` event
`onChangeDeferred` - callback triggered when above `setTimeout` fires

If the `onChange` event is fired before the `deferTimeout` time, the `onChangeDeferred` is deferred again.

# TODO
- [x] Release notes

#1199

# Screenshots / Gifs
carbon-870


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.1.0
+
+## Package Updates
```
andrewjtait: `Package Updates` would normally refer to version changes in the package. I'd use `Improvements` as a heading instead.

Issue #1684, opened 2018-01-28
Props passed to base connected component should be resolved such that mapToProps has higher precedence over props passed to the HOC


amaalali:
For a HOC composed using [connect](https://github.com/Sage/carbon/blob/master/src/utils/flux/connect/connect.js), the props passed to the base component should be of higher precedence for [`mapToProps `](https://github.com/Sage/carbon/blob/master/src/utils/flux/connect/connect.js#L48) over props passed to the HOC.

It is likely that based on the state from stores passed to `mapToProps`, the user will want to modify the props passed.

# Example
```
import React from 'react';
import ReactDom from 'react';

const SimpleDiv = ({ text }) => <div>{text}</div>;

mapToProps = (props) => ({
  text: `MODIFIED ${props.text}` // the props can be modified here
})

const HigherOrderDiv = connect(mapToProps)(SimpleDiv);

ReactDOM.render(
  <HigherOrderDiv text='text' />,
  document.getElementById('app')
);
```

**Expected Output**
```
<Connect(SimpleDiv)>
  <SimpleDiv>
    <div>MODIFIED text</div>
  <SimpleDiv>
<Connect(SimpleDiv)>
```
**Current Output**
```
<Connect(SimpleDiv)>
  <SimpleDiv>
    <div>text</div>
  <SimpleDiv>
<Connect(SimpleDiv)>
```

amaalali:
I have a solution for this that I can open a PR for this week.

PR #1686, opened 2018-01-29
mapToProps takes precedence over props passed to HOC

amaalali:
# Description
For a HOC composed using [connect](https://github.com/Sage/carbon/blob/master/src/utils/flux/connect/connect.js), the props passed to the base component should be resolved such that [`mapToProps `](https://github.com/Sage/carbon/blob/master/src/utils/flux/connect/connect.js#L48) has higher precedence over props passed to the HOC.

# TODO
- [x] Release notes

https://github.com/Sage/carbon/issues/1684

# Testing Instructions

Advice needed


src/utils/flux/connect/__spec__.js
```diff
@@ -86,6 +86,15 @@ describe('connect', () => {
     expect(mapStateToProps).toHaveBeenCalledWith(testStore.getState(), wrapper.props());
   });
 
+  it('props from mapStateToProps take precedence', () => {
+    const Presenter = () => null;
+    const mapStateToProps = props => ({ text: `modified ${props.text}` });
+    const ConnectedComponent = connect(mapStateToProps)(Presenter);
+    const wrapper = mount(<ConnectedComponent text={ 'sample text' } />).find(Presenter);
```
toni-leigh: Have you see [`dive()`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/dive.html) which I used to get round having to use `mount` when using HOCs (not sure if it would actually work here, but worth knowing about)

toni-leigh:
Also nice that you're keeping this properly open source by using issues in Github rather than Jira tickets!

You can tag the commits too, but not with a `#` at the front:

`some commit message #1234`

src/utils/flux/connect/__spec__.js
amaalali: I would have prefered to have done this but after shallow rendering then looking at the sub-component using `dive` didn't return the `Presenter ` component ¯\_(ツ)_/¯

amaalali: I've changed the test to use `shallow`, but it turns out we don't need to use `dive` here as shallow returns the return of the HOC, which will be the base component that is being wrapped.

Issue #1048, opened 2017-02-01
Automatic Disabling of Form


andrewjtait:
When a form is submitted we should automatically disable the submit button.

The tricky part is when to re-enable the form (server error etc). It could be that we pass a callback function as part of the `onSubmit` and `onSave` props so the developer can call these functions when the server responds with an error for example.

ianoxley:
Resolved by #1688

PR #1688, opened 2018-01-31
Carbon 638 auto disabling form save

JohnStuartBarrie:
# Description
Autodisable prop added to submit


src/components/form/form.js
```diff
@@ -522,10 +540,14 @@ class Form extends React.Component {
     }
 
     if (valid && this.props.onSubmit) {
-      this.props.onSubmit(ev);
+      this.props.onSubmit(ev, this.enableForm);
```
andrewjtait: We should also pass this to `afterFormValidation` above ^

src/components/form/form.js
```diff
@@ -509,6 +523,10 @@ class Form extends React.Component {
    * @return {void}
    */
   handleOnSubmit = (ev) => {
+    if (this.props.autoDisable) {
+      this.setState({ submitted: true });
+    }
```
andrewjtait: if client side validation fails we could re-enable automatically, after

```
const valid = this.validate();
```

inside the `if (!valid)` check that is already there probably

src/components/form/form.js
JohnStuartBarrie: It's just an idea but shouldn't we just depricate afterFormValidation?

SeanArmstrong: We may want to do something/show something to the user extra when the client side validation fails.
`if valid` check might be useful although maybe that is a rare case?

src/components/form/form.js
```diff
@@ -109,7 +109,14 @@ class Form extends React.Component {
...
     validateOnMount: PropTypes.bool,
-
+    /**
+     * If true, will disable the savebutton when clicked
+     *
+     * @property autoDisable
+     * @type {Boolean}
+     * @default false
+     */
+    autoDisable: PropTypes.bool,
```
SeanArmstrong: Would be tempted to make this true by default in the future although that would be a breaking change

src/components/form/form.js
andrewjtait: Yeah its just a hook that gives them more control if they need it.

src/components/form/form.js
```diff
@@ -518,14 +532,21 @@ class Form extends React.Component {
...
     if (this.props.afterFormValidation) {
-      this.props.afterFormValidation(ev, valid);
+      this.props.afterFormValidation(ev, valid, this.enableForm);
     }
 
     if (valid && this.props.onSubmit) {
-      this.props.onSubmit(ev);
+      this.props.onSubmit(ev, this.enableForm);
+      if (this.props.autoDisable) {
+        this.setState({ submitted: true });
```
andrewjtait: prefer this where you had it as `this.validate` might be slow if it is a big form - the sooner we disable it the better.

andrewjtait:
@SeanArmstrong yeah the intention here was to simplify the common functionality that a developer would need to do. It is not very React, but is a stepping stone to possibly turning this on by default once we figure out a decent way to automatically re-enable the form.

With this though devs can start using it now, as long as they use the callback.

andrewjtait:
Its probably more understandable to say that the problem isn't that devs couldn't do this the React way and manage the state themselves, it's that they don't. This is to make it easier and to eventually ensure it does happen, without relying on the dev to remember or care to do it.

src/components/form/__definition__.js
```diff
@@ -45,6 +46,7 @@ let definition = new Definition('form', Form, {
     summary: "Boolean",
   },
   propDescriptions: {
+    autoDisable: "Set to true will disable the save button on submit",
```
ianoxley: Please change this to: "Set to true to disable the save button once the form has been submitted".

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.0.2
+
+### AutoDisabling form
```
andrewjtait: one to many `#`

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 3.0.2
```
andrewjtait: `3.1.0`

src/components/form/form.js
```diff
@@ -109,7 +109,14 @@ class Form extends React.Component {
      * @default false
      */
     validateOnMount: PropTypes.bool,
-
```
andrewjtait: put line break back in

src/components/form/form.js
andrewjtait: missing line break below

src/components/form/form.js
```diff
@@ -266,7 +273,14 @@ class Form extends React.Component {
      * @property warningCount
      * @type {Number}
      */
-    warningCount: 0
+    warningCount: 0,
```
andrewjtait: missing line break below

src/components/form/form.js
```diff
@@ -518,14 +532,21 @@ class Form extends React.Component {
...
     if (valid && this.props.onSubmit) {
-      this.props.onSubmit(ev);
+      this.props.onSubmit(ev, this.enableForm);
+      if (this.props.autoDisable) {
+        this.setState({ submitted: true });
+      }
     }
   }
 
+  enableForm = () => {
```
andrewjtait: need some comments above this function to describe what it is used for

src/components/form/form.js
JohnStuartBarrie: I don't understand what you mean here. You mean put it further up ? If so, surely we have to make sure that the form has validated before disabling the save button. If not the use would try and send with invalid info, then when he corrects the info and tries to resend the button will be disabled. Or have I not understood your demand

src/components/form/form.js
JohnStuartBarrie: Done

src/components/form/form.js
ianoxley: Will passing `this.enableForm` as a callback function work? Or will the `this` inside `enableForm` be pointing elsewhere when it gets called?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # 3.1.0
 
+
+## AutoDisabling form
+
+Form autoDisables after submit when the prop `autoDisable` is set to true. The props `afterFormValidation` and `onSubmit` are passed a `enableForm` callback function which can be used to reactivate the form.
+
```
ianoxley: Some example code to illustrate this would be nice.

src/components/form/form.js
andrewjtait: should be ok as `enableForm` is a class property, so should be automatically bound to the instance of the component

src/components/form/form.js
JohnStuartBarrie: https://github.com/Sage/carbon/blob/0268cd392a50784427e553493af965064b58a8c8/src/components/form/form.js#L549 should it go here ?

src/components/form/form.js
JohnStuartBarrie: I just booted up the code I wrote to test it in GAC. It works fine.

src/components/form/form.js
andrewjtait: I thought you had put it here? https://github.com/Sage/carbon/pull/1688/files#diff-ce2b1986d94d1af0464dc730fd2bbe29R539

(my original comment was from 7 days ago)

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+### Example Code
+  ``` 
+  <Form
+    autoDisable
+    onSubmit={ this.saveContact }
+    cancel={ false }>
```
andrewjtait: do you need cancel false in your example? also the `>` would normally fall onto the next line

```
<Form
  onSubmit={ this.saveContact }
  autoDisable
>
  { children }
</Form>
```

src/components/form/form.js
JohnStuartBarrie: done

ianoxley:
Fixes #1048

PR #1689, opened 2018-02-01
Carbon 971 portal on toast

JohnStuartBarrie:
# Description

Portal on Toast

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,4 @@
 # 3.1.0
-
 ## Modal Improvement
 
 The Modal now uses the Portal Component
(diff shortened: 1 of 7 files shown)


src/components/toast/__definition__.js
```diff
@@ -29,7 +29,7 @@ let definition = new Definition('toast', Toast, {
   },
   propValues: {
     children: "Talkie's the name, toasting's the game. Anyone like any toast?",
-    open: true
+    open: false
```
andrewjtait: revert

CHANGELOG.md
```diff
@@ -1,5 +1,4 @@
 # 3.1.0
-
```
andrewjtait: revert

src/components/toast/toast.scss
```diff
@@ -15,7 +15,6 @@ $toast-width: 300px !default;
   right: $toast-anim-end;
   top: $toast-top-position;
   width: $toast-width;
-  z-index: $z-toast;
```
andrewjtait: The Toast needs to be higher in `z-index` than dialogs. It should probably sit above dialogs regardless of where it is rendered.

Issue #1291, opened 2017-06-01
Automation Support for Dropdown


andrewjtait:
Investigate support for automation and Dropdowns. Currently we render the options for the dropdown, even when they are hidden (for automation to work correctly). We should spend some time to investigate alternatives.

SeanArmstrong:
We also need to investigate the accessibility of the dropdown that we have. Can the options even be accessed by page readers. Will it help to also render the options or will given focus to the dropdown allow for it to read all the options

SeanArmstrong:
The reason for originally hiding this information is that we didn't need the markup if there was no focus. Performance reasons. If the dropdown items had lots of input listeners.

ianforster-sage:
Fixed in PR #1690

PR #1690, opened 2018-02-02
Carbon 969 dropdown portal

JohnStuartBarrie:
# Description
Dropdown using Portal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # 3.1.0
 
+## Dropdown Improvement
+
+The DropDown Component now uses the Portal Component
+
 ## Modal Improvement
 
 The Modal now uses the Portal Component
--- a/src/components/dropdown/__spec__.js
+++ b/src/components/dropdown/__spec__.js
@@ -1,11 +1,12 @@
 import React from 'react';
 import TestUtils from 'react-dom/test-utils';
-import Dropdown from './dropdown';
+import { shallow, mount } from 'enzyme';
 import Immutable from 'immutable';
+import Dropdown from './dropdown';
 import Events from './../../utils/helpers/events';
-import { shallow } from 'enzyme';
 import { elementsTagTest, rootTagTest } from '../../utils/helpers/tags/tags-specs';
 import ImmutableHelper from './../../utils/helpers/immutable';
+import Portal from './../portal';
 
 /* global jest */
 describe('Dropdown', () => {
@@ -13,7 +14,7 @@ describe('Dropdown', () => {
 
   beforeEach(() => {
     instance = TestUtils.renderIntoDocument(
-      <Dropdown name="foo" options={ Immutable.fromJS([]) } value="1" />
+      <Dropdown name='foo' options={ Immutable.fromJS([]) } value='1' />
     );
   });
 
@@ -32,7 +33,7 @@ describe('Dropdown', () => {
   describe('componentWillReceiveProps', () => {
     describe('when cacheVisibleValue is false', () => {
       it('resets visibleValue', () => {
-        instance.visibleValue = "foobar";
+        instance.visibleValue = "exitfoobar";
         instance.componentWillReceiveProps({
           value: "2"
         });
@@ -921,9 +922,18 @@ describe('Dropdown', () => {
       });
     });
 
-    it('returns the list', () => {
-      let classes = 'carbon-dropdown__list-block carbon-dropdown__list-hidden';
-      expect(instance.additionalInputContent[1].props.className).toEqual(classes);
+    it('creates the list in a Portal', () => {
+      const wrapper = mount(
+        <Dropdown
+          name='foo'
+          options={ Immutable.fromJS([{ id: 1, name: 'foo' }, { id: 2, name: 'bar' }]) } value='1'
+        />);
+      const portal = wrapper.find(Portal);
+      expect(portal.props().children.key).toEqual('listBlock');
+
+      const portalDropdownList = wrapper.find(Portal).props().children.props.children.props;
+      expect(portalDropdownList.className).toEqual('carbon-dropdown__list');
+      expect(portalDropdownList.children.length).toBe(2);
     });
   });
 
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -7,6 +7,7 @@ import InputValidation from './../../utils/decorators/input-validation';
 import InputIcon from './../../utils/decorators/input-icon';
 import Events from './../../utils/helpers/events';
 import { validProps } from '../../utils/ether';
+import Portal from './../portal';
 
 /**
  * A dropdown widget.
@@ -649,6 +650,10 @@ class Dropdown extends React.Component {
     return results;
   }
 
+  positionOptions() {
+    console.log(!!this.listBlock);
+  }
+
   /**
    * Extends the input content to include the input icon.
    *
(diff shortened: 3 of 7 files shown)


src/components/dropdown/__spec__.js
```diff
@@ -921,9 +922,18 @@ describe('Dropdown', () => {
...
+    it('creates the list in a Portal', () => {
+      const wrapper = mount(
+        <Dropdown
+          name='foo'
+          options={ Immutable.fromJS([{ id: 1, name: 'foo' }, { id: 2, name: 'bar' }]) } value='1'
+        />);
+      const portal = wrapper.find(Portal);
+      expect(portal.props().children.key).toEqual('listBlock');
+
+      const portalDropdownList = wrapper.find(Portal).props().children.props.children.props;
```
abramin: Just curious if there is a cleaner way to drill down - maybe something in enzyme to fetch the list?

andrewjtait: Yeah you should just be able to use `find`.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # 3.1.0
 
+## Dropdown Improvement
+
+The DropDown Component now uses the Portal Component
```
andrewjtait: `Dropdown` (lowercase d)

src/components/dropdown/dropdown.js
```diff
@@ -649,6 +652,15 @@ class Dropdown extends React.Component {
     return results;
   }
 
+  positionOptions() {
```
andrewjtait: `positionOptions` sounds like it'd return some options - maybe `calculatePosition` or `handlePosition` or something would be more descriptive?

src/components/dropdown/dropdown.js
```diff
@@ -649,6 +652,15 @@ class Dropdown extends React.Component {
...
+    if (this.dropDown) {
+      const input = this.dropDown.getElementsByClassName('carbon-dropdown__input common-input__input')[0];
```
andrewjtait: I think all input components have a ref available of `input` or `_input` or something - you should just be able to use this.

src/components/dropdown/dropdown.js
```diff
@@ -716,6 +731,9 @@ class Dropdown extends React.Component {
         className={ this.mainClasses }
         { ...this.componentTags(this.props) }
         data-state={ this.requestingState() }
+        ref={ (dropDown) => {
+          this.dropDown = dropDown;
```
andrewjtait: don't think you will need this.

src/components/dropdown/dropdown.js
```diff
@@ -663,11 +675,14 @@ class Dropdown extends React.Component {
...
-        { this.listHTML }
-      </div>
+      <Portal onReposition={ () => { this.positionOptions(); } }>
+        <div { ...this.listBlockProps } ref={ (node) => { this.listBlock = node; } }>
+            { this.listHTML }
+        </div>
+      </Portal>
     );
 
+    this.positionOptions();
```
andrewjtait: should we trigger this when the list opens rather than during a method that gets called on render? (means it would get called less)

src/components/dropdown/dropdown.js
andrewjtait: looks like `_input` - https://github.com/Sage/carbon/blob/master/src/utils/decorators/input/input.js#L236

andrewjtait:
Can you also add comments to the functions, similar to the rest of the file?

src/components/dropdown/dropdown.js
andrewjtait: maybe `onFocus`?

src/components/dropdown/__spec__.js
JohnStuartBarrie: Yip you are both right. I added a snap shot.

CHANGELOG.md
JohnStuartBarrie: done

src/components/dropdown/dropdown.js
JohnStuartBarrie: done

src/components/dropdown/dropdown.js
JohnStuartBarrie: done

src/components/dropdown/dropdown.js
JohnStuartBarrie: done

src/components/dropdown/dropdown.js
JohnStuartBarrie: done

src/components/dropdown/dropdown.js
```diff
@@ -322,6 +325,7 @@ class Dropdown extends React.Component {
    * @return {Void}
    */
   handleFocus = () => {
+    this.calculatePosition();
```
andrewjtait: maybe as part of the `else` block? wouldn't need to calculate it if focus is blocked?

src/components/dropdown/__spec__.js
```diff
@@ -921,9 +922,15 @@ describe('Dropdown', () => {
...
-      let classes = 'carbon-dropdown__list-block carbon-dropdown__list-hidden';
-      expect(instance.additionalInputContent[1].props.className).toEqual(classes);
+    it('creates the list in a Portal', () => {
+      const wrapper = mount(
+        <Dropdown
+          name='foo'
+          options={ Immutable.fromJS([{ id: 1, name: 'foo' }, { id: 2, name: 'bar' }]) } value='1'
+        />);
+
+      const portalDropdownList = wrapper.find('.carbon-dropdown__list');
```
andrewjtait: this is snapshoting the list, but doesn't prove it is in a Portal?

src/components/dropdown/dropdown.js
```diff
@@ -649,6 +653,14 @@ class Dropdown extends React.Component {
     return results;
   }
 
+  calculatePosition() {
```
andrewjtait: need to add a descriptive comment to this function

src/components/dropdown/dropdown.js
```diff
@@ -663,9 +680,11 @@ class Dropdown extends React.Component {
     }
 
     content.push(
-      <div { ...this.listBlockProps } ref={ (node) => { this.listBlock = node; } }>
-        { this.listHTML }
-      </div>
+      <Portal onReposition={ () => { this.calculatePosition(); } }>
```
andrewjtait: just noticed you don't need to create a function here...

```
<Portal onReposition={ this.calculatePosition }>
```
rather than

```
<Portal onReposition={ () => { this.calculatePosition(); } }>
```

src/components/dropdown/dropdown.js
```diff
@@ -648,6 +652,19 @@ class Dropdown extends React.Component {
 
     return results;
   }
+  /**
+   * positions the portal listBlock in relation to the input.
+   *
+   * @method calculatePosition
+   * @return {void}
+   */
+  calculatePosition() {
```
andrewjtait: change this to

```
calculatePosition = () => {
```

this uses class properties, and automatically binds the instance to the method.

src/components/dropdown/dropdown.js
```diff
@@ -648,6 +652,19 @@ class Dropdown extends React.Component {
...
+    this.listBlock.style.top = `${this._input.getBoundingClientRect().y
+                                    + (this._input.getBoundingClientRect().height)
+                                    + window.scrollY}px`;
+    this.listBlock.style.width = `${this._input.getBoundingClientRect().width}px`;
+    this.listBlock.style.left = `${this._input.getBoundingClientRect().x}px`;
+  }
```
ianoxley: We've got 4 calls to `getBoundingClientRect()` in 3 lines here, and 3 calls to `this.listBlock.style`. It'd be better to call each of these just the once:

``` js
const boundingRect = this._input.getBoundingClientRect();
const top = `${boundingRect.y + boundingRect.height + window.scrollY}px`;
const width = `${boundingRect.width}px`;
const left = `${boundingRect.x}px`;
this.listBlock.setAttribute('style', `left: ${left}; top: ${top}; width: ${width};`);
```

JohnStuartBarrie: Of course

andrewjtait:
Tested in demo site ✅ 

Might be good to take a look at this ticket soon for potential performance issues https://github.com/Sage/carbon/issues/1291

PR #1691, opened 2018-02-05
ESLint Fixes

andrewjtait:
Fixes new rules from eslint upgrades.

### DIFF
--- a/src/components/dialog-full-screen/dialog-full-screen.js
+++ b/src/components/dialog-full-screen/dialog-full-screen.js
@@ -102,14 +102,14 @@ class DialogFullScreen extends Modal {
    * Overrides the original function to disable the document's scroll.
    */
   get onOpening() {
-    this.document.documentElement.classList.add(DIALOG_OPEN_HTML_CLASS);
+    return this.document.documentElement.classList.add(DIALOG_OPEN_HTML_CLASS);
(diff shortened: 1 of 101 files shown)


src/components/dialog-full-screen/dialog-full-screen.js
```diff
@@ -102,14 +102,14 @@ class DialogFullScreen extends Modal {
...
-    this.document.documentElement.classList.add(DIALOG_OPEN_HTML_CLASS);
+    return this.document.documentElement.classList.add(DIALOG_OPEN_HTML_CLASS);
   }
 
   /**
    * Overrides the original function to enable the document's scroll.
    */
   get onClosing() {
-    this.document.documentElement.classList.remove(DIALOG_OPEN_HTML_CLASS);
+    return this.document.documentElement.classList.remove(DIALOG_OPEN_HTML_CLASS);
```
SeanArmstrong: Is this a problem with it being a get method rather than its not returning

andrewjtait: correct - would need a larger refactor to resolve properly

andrewjtait:
Agree on the getters - however that would be a breaking change due to the inheritance we currently rely on.

PR #1692, opened 2018-02-07
[CARBON-159] - Add Dirty Forms

### LINKED ISSUE #298: Dirty forms mechanism
Carbon requires its own dirty forms mechanism to ensure users don't accidentally lose data with backspace or links.

krillatron:
# Description
Adding Dirty Forms to Carbon as per SOP, as a prop for the form component. 
If added, will generate a popup informing you of unsaved changes if present and asking if you're sure you want to navigate away from the page. Triggers on form unload (link, url change) but not on react route redirects, as per SOP.

Does NOT work on react router page transitions
Does NOT consistently work for browser back/forwards buttons as these use back/forwards cache and do not always trigger the beforePageUnload. 

Resolves #298

# Testing Instructions
- If savePrompt prop is true and form has been changed, confirmation popup should show when navigating away
- If savePrompt prop is true and form hasn't been changed, no popup should show when navigating away
- If savePrompt prop is false, no popup should show when navigating away

# Example
Example of prop + usage: 
[Dirty Forms - Carbon.zip](https://github.com/Sage/carbon/files/1711212/Dirty.Forms.-.Carbon.zip)


src/components/form/form.js
```diff
@@ -305,6 +326,10 @@ class Form extends React.Component {
     if (this.props.validateOnMount) {
       this.validate();
     }
+
+    if (this.props.savePrompt) {
+      this.addSavePromptListener();
```
andrewjtait: need to remove listener on `componentWillUnmount`

src/components/form/form.js
```diff
@@ -348,6 +379,28 @@ class Form extends React.Component {
...
   }
 
+  /**
+   * Sets the form to Dirty
+   *
+   * @method setIsDirty
+   * @return {void}
+   */
+  setIsDirty = () => {
+    this.isDirty = true;
```
andrewjtait: why do you need to track `this.isDirty` if you already have `this.state.isDirty`?

src/components/form/form.js
```diff
@@ -348,6 +379,28 @@ class Form extends React.Component {
...
+    this.setState({ isDirty: this.isDirty });
```
andrewjtait: could wrap this in a `if (!this.state.isDirty) {` to block unrequired calls to `setState`?

src/components/form/form.js
```diff
@@ -348,6 +379,28 @@ class Form extends React.Component {
...
+
+  /**
+   * Sets the form to Clean
+   *
+   * @method resetIsDirty
+   * @return {void}
+   */
+  resetIsDirty = () => {
+    this.isDirty = false;
+    this.setState({ isDirty: this.isDirty });
```
andrewjtait: could wrap this in an `if`

src/components/form/form.js
```diff
@@ -381,6 +434,25 @@ class Form extends React.Component {
...
+  removeSavePromptListener = () => {
+    this._window.removeEventListener('beforeunload', this.checkIsFormDirty);
+  }
+
+  checkIsFormDirty = () => {
+    let confirmationMessage = '';
+    if (this.isDirty) {
+      // Confirmation message is usually overridden by browsers with a similar message
+      confirmationMessage = 'Do you want to reload this site?'
+                              + 'Changes that you made may not be saved.';
```
andrewjtait: Use I18n for the message.

src/components/form/form.js
```diff
@@ -514,7 +586,11 @@ class Form extends React.Component {
 
     const valid = this.validate();
 
-    if (!valid) { ev.preventDefault(); }
+    if (valid) {
+      this.resetIsDirty();
```
andrewjtait: not sure about this - what if it is sent through a POST and the server responds with an error, meaning the data has not been submitted? Not saying this is wrong, just wondering if you have any thoughts on it.

src/components/form/form.js
```diff
@@ -381,6 +434,25 @@ class Form extends React.Component {
     }
   }
 
+  addSavePromptListener = () => {
+    this._window.addEventListener('beforeunload', this.checkIsFormDirty);
```
andrewjtait: what happens if there are multiple forms on the page?

SeanArmstrong:
Would be good to get some screenshots or gifs of the work for UX review

src/components/form/form.js
krillatron: I just attempted this locally with 2 forms being present on a page, one with savePrompt enabled and one with it disabled. The pop-up only displayed when the first form was changed, and didn't when the latter was changed, which is what I'd expect.
In the case of 2 forms both with savePrompt enabled, editing both forms then attempting to navigate away from the page brought up the prompt once, wether choosing to stay on the page or to navigate away from the page

src/components/form/form.js
```diff
@@ -237,6 +248,7 @@ class Form extends React.Component {
   static defaultProps = {
     buttonAlign: 'right',
     cancel: true,
+    savePrompt: true,
```
SeanArmstrong: Im not sure this is the best name as its not prompting you when you save. Trying to think of something better now

`dirtyFormWarning`
`dirtyFormCheck`
`warnOnDirtyForm`

andrewjtait: `unsavedWarning`?

src/components/form/form.js
```diff
@@ -332,12 +359,20 @@ class Form extends React.Component {
     if (!nextProps.stickyFooter && this.props.stickyFooter) {
       this.removeStickyFooterListeners();
     }
+
+    if (nextProps.unsavedWarning) {
```
andrewjtait: you should check here if `this.props.unsavedWarning` is different, otherwise it will add listener after listener after listener

src/components/form/form.js
```diff
@@ -332,12 +359,20 @@ class Form extends React.Component {
...
+      this.addUnsavedWarningListener();
+    } else {
```
andrewjtait: should check if it actually needs to remove the listener?

harpalsingh:
Not tested this myself, but for QA, as long its the same as SOP, passes UX.

toni-leigh:
As this is an open source repo, ticket numbers don't make a great deal of sense.

We can link to the issue using `#num` in commits and PRs though. If the issue is thin on info I'd suggest expanding on this with some anonymised info from the ticket - this makes the repo more rich for anyone outside of Sage who finds it

Issue #905, opened 2016-11-18
Inconsistent CHANGELOG


SeanArmstrong:
There is no standard for how our CHANGELOG is structured and therefore we are getting similar headings e.g. Improvements, Minor Improvements, Component Changes, Component Improvements. Sometimes each change is put into a heading other times a improvement is separated to its own main heading

Should we look to standardise this for each set of release notes. This would also make the task of change log generator much easier.

Possible Headings

- Breaking Changes
- Component Changes/Improvements
- New Components
- Non Component/Utils Changes/Improvements
- CSS Changes
- Dependency Change ( update version / new dependency )
- Bug Fixes ( Maybe only for patch versions otherwise it would be a change )

Maybe we shouldn't restrict the possible headings in CHANGELOG but i don't think we should switch between improvements/minor improvements/bug fixes etc...

PR #1695, opened 2018-02-12
Add renogen

SeanArmstrong:
# Example Output:

# v3.1.0 (2018-02-12)
### Bug Fixes
* `mapToProps` takes precedence over props passed to HOC in `connect` function.

### Other Improvements
* Resolved new ESLint errors from carbon-factory upgrade
* Input Validation: The InputValidation error tooltips now use the Portal Component

### Component Changes
* Modal: The Modal now uses the Portal Component
* Toast: Toast component now uses the Portal Component
* Input: Input has 2 new props. `onChangeDeferred` allows a deferred callback after an onChange event. `deferTimeout` allows you to customise the default: `750`.
* Form

  Form autoDisables after submit when the prop `autoDisable` is set to true. The props `afterFormValidation` and `onSubmit` are passed a `enableForm` callback function which can be used to reactivate the form.

  Example Code
  ```
    <Form
      onSubmit={ this.saveContact }
      autoDisable
    >
      {children}
    </Form>
  ```

  ```
    saveContact = (ev, valid, enableForm) => {
      ...
      Actions.submitForm(...);
      enableForm();
    };
  ```

* Dropdown: Component now uses the Portal Component

https://github.com/Sage/carbon/issues/905


script/generate-release-notes.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+if [ $# -eq 0 ]; then
```
andrewjtait: this version is better

```
#!/bin/bash

PACKAGE_VERSION=$(cat package.json \
  | grep version \
  | head -1 \
  | awk -F: '{ print $2 }' \
  | sed 's/[",]//g')

dot_version=`echo $PACKAGE_VERSION | sed -E "s/-rc.*$//"`
dash_version=`echo ${dot_version} | tr '.' '_'`

RELEASE_NOTES_FILE="release_notes/v${dash_version}.html.md"

renogen --format markdown v${dot_version} >> ${RELEASE_NOTES_FILE}

mkdir change_log/v${dash_version}
mv change_log/next/* change_log/v${dash_version}/

cat ${RELEASE_NOTES_FILE}
```

chrisbarber86:
we should look at adding/running `renogen test` as part of the CI - this would prove that release notes submitted are valid at the point of pull request - rather than trying to fix problems at the end when generating them

README.md
```diff
@@ -58,14 +58,24 @@ To run the example, do the following steps:
...
+
 ## Submitting a Release
 
 * N.B. Releases are created by Sage Carbon Team. We use [semantic versioning](http://semver.org/)
 
 * If releasing a minor version, create a branch from `master`.
 * If releasing a patch version, create a branch from the tag you want to patch. This should be the latest tag apart from exceptional circumstances.
 * Bump the version in `package.json`.
-* Ensure the `CHANGELOG.md` is up to date.
+* Generate Release Notes using the provided script `./script/generate_release_notes.sh 1.2.3`
```
andrewjtait: dont need number on the end of this any more

script/generate-release-notes.sh
```diff
@@ -0,0 +1,19 @@
...
+  | head -1 \
+  | awk -F: '{ print $2 }' \
+  | sed 's/[",]//g')
+
+dot_version=`echo $PACKAGE_VERSION | sed -E "s/-rc.*$//"`
+dash_version=`echo ${dot_version} | tr '.' '_'`
+
+RELEASE_NOTES_FILE="release_notes/v${dash_version}.html.md"
+
+renogen --format markdown v${dot_version} >> ${RELEASE_NOTES_FILE}
```
chrisbarber86: We'll need a Gemfile/Gemfile.lock to ensure `renogen` is present

SeanArmstrong: Do we as its installed globally on a machine. Or can be. Didn't want to cruft up the repo

SeanArmstrong: Although if we want the check as part of build we will probably need it

chrisbarber86: Global install is fine, we need instructions for that somewhere though right?

PR #1696, opened 2018-02-20
bug fix and tests

JohnStuartBarrie:
# Description

When loading the GAC Gateway pointing at local Master branch of Carbon.

Click on Contacts and Create and New Customer.

On the New Customer Screen open the dropdown for either Account Default or Country

Expected

The dropdown to be directly underneath the button

Observed

It appears off to the side of the page. See attachment.

# Testing Instructions

### DIFF
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -327,7 +327,6 @@ const Dropdown = Input(InputIcon(InputLabel(InputValidation(class Dropdown exten
     if (this.blockFocus) {
       this.blockFocus = false;
     } else {
-      this.calculatePosition();
       this.setState({ open: true });
     }
   }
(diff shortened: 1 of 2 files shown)


iliaoni:
Tested on FF/Chrome/Safari - dropdowns open in correct position. QA passed.

src/components/dropdown/dropdown.js
```diff
@@ -678,13 +677,15 @@ const Dropdown = Input(InputIcon(InputLabel(InputValidation(class Dropdown exten
...
     }
 
-    content.push(
-      <Portal onReposition={ this.calculatePosition }>
-        <div { ...this.listBlockProps } ref={ (node) => { this.listBlock = node; } }>
-          { this.listHTML }
-        </div>
-      </Portal>
-    );
+    if (this.state.open) {
```
andrewjtait: unfortunately we cannot do this due to https://jira-sage.valiantyscloud.net/browse/CARBON-1044

JohnStuartBarrie: This branch was approved tested and merged. Should we revert. This being said, when I read the other ticket, it seems that CARBON-1044 is resolved by this fix right?

andrewjtait: The issue is that automation expects the dropdown list to be present in the DOM, even when closed. Now that you have removed the list from the DOM, automation will fail.

The ticket I linked to needs a solution that will resolve the issue with automation as well.

Issue #1700, opened 2018-02-27
Raise max event emitters


toni-leigh:
We have periodically seen the following error in our code:

```
(node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.
```

As the HOC and Connect pattern is favoured, along with multiple stores being used in some cases as well as using different layouts to organise our components, meaning they need to be broken down into quite small chunks, all attached to stores, we need to raise the number of allowed event listeners in the Store object to avoid a JS warning.

IMO using a single store / Redux which seems to be the most popular option in wild is the best solution but we are currently not architected like that and so need to make this change

toni-leigh:
This is has been merged and released in version 3.1.0

PR #1701, opened 2018-02-27
Increase maxListeners #1700

toni-leigh:
## Description

* I chose 50, which would allow for a complex layout with several store on some of it's children, this should be enough room to manoeuvre for a varied array of applications without allowing anything really silly to happen with exponentials

## TODO
- [x] Release notes

## Related Issues / Pull Requests
https://github.com/Sage/carbon/issues/1700

## Testing Instructions
* Has been dev checked in the context where it was happening

### DIFF
--- a/src/utils/flux/store/__spec__.js
+++ b/src/utils/flux/store/__spec__.js
@@ -2,6 +2,7 @@ import Store from './store';
 import Flux from 'flux';
 import Logger from './../../logger';
 import { Dispatcher } from './../../flux';
+import { EventEmitter } from 'events';
 
 describe('Store', () => {
   let instance;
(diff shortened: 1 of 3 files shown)


src/utils/flux/store/__spec__.js
```diff
@@ -54,6 +55,14 @@ describe('Store', () => {
         expect(instance.trackHistory).toBeTruthy();
       });
     });
+
+    describe('max event listeners', () => {
+      it('set to 50', () => {
+        spyOn(EventEmitter.prototype, 'setMaxListeners');
+        instance = new Store('foo', 'some data');
+        expect(EventEmitter.prototype.setMaxListeners).toHaveBeenCalledWith(50);
```
amaalali: I this this should instead be: 
```
expect(instance.getMaxListeners()).toEqual(50)
```

PR #1708, opened 2018-03-07
Fix invalid HTML when using CSSTransitionGroup

ianoxley:
Pass the `component='div'` prop to all instances of `CSSTransitionGroup` that contain block-level content.

By default `CSSTransitionGroup` uses a `<span>` tag, which can lead to invalid HTML such as:

```html
<span>
  <div>
    My component
  </div>
</span>
```

By setting the `component` prop to the string `div` we ensure that the component HTML is valid HTML and doesn't include incorrectly nested tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 3.1.3
+
+## Bug Fixes
+
+* `AnimatedMenuButton`, `Carousel`, `Flash`, `ShowEditPod`, `Table`, and `Toast` all pass the `component='div'` prop to their respective `CSSTransitionGroup` components. This fixes incorrectly nested HTML e.g. `<div>` tags nested within `<span>` tags.
+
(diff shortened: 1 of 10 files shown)


SeanArmstrong:
Do we think this could be a breaking change due to the CSS differences? QA would probably pick up on this if it is the case

ianoxley:
@SeanArmstrong It _shouldn't_ be, as all that has changed is the tag used in the selector.

andrewjtait:
devs shouldn't be hanging off element types really so wouldn't worry about it being a breaking change

ianrobsonsage:
loaded the branch 'fix-csstransition-html' of carbon, compared the following components with carbon.sage.com. No differences found. QA Passed.

- AnimatedMenuButton
- Carousel
- Flash
- ShowEditPod
- Table
- Toast

PR #1712, opened 2018-03-13
CARBON-970 Use Portals for the Flash component

JohnStuartBarrie:
# Description

Flash messages should use the Portal


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 Fixes auto-deployment of tags using Travis CI.
 
+Flash component now uses Portal
```
andrewjtait: release notes are in the wrong place (should target 3.2.0)

src/components/flash/flash.js
```diff
@@ -133,9 +135,16 @@ class Flash extends React.Component {
...
    * @return(Void)
    */
-  componentWillReceiveProps(prevProps) {
-    if (prevProps.open !== this.props.open) {
-      this.setState({ dialogs: {} });
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open === this.props.open) { return; }
+
+    if (nextProps.open) {
+      this.setState({ dialogs: {}, open: nextProps.open });
```
andrewjtait: you don't need to set all state params on `setState`, just the ones you want to update

src/components/flash/flash.js
```diff
@@ -133,9 +135,16 @@ class Flash extends React.Component {
...
+    } else {
+      clearTimeout(this.fadeLeaveTimeout);
+      this.fadeLeaveTimeout = setTimeout(() => {
+        this.setState({ dialogs: {}, open: nextProps.open });
```
andrewjtait: same again, don't think you need to set `dialogs`?

src/components/flash/flash.js
```diff
@@ -456,34 +465,38 @@ class Flash extends React.Component {
...
-        </div>
-
-        { this.dialogs }
-      </div>
+      this.state.open && (
+        <Portal>
+          <div { ...tagComponent('flash', this.props) }>
+            <div className={ this.classes }>
+              <CSSTransitionGroup
+                transitionAppear
```
ianoxley: You'll need a `component='div'` prop here, otherwise the HTML will be invalid.

src/components/flash/flash.js
```diff
@@ -456,34 +465,38 @@ class Flash extends React.Component {
...
+                transitionAppearTimeout={ 500 }
+                transitionName='carbon-flash__slider'
+                transitionEnterTimeout={ 600 }
+                transitionLeave
+                transitionLeaveTimeout={ 2600 }
+              >
+                { sliderHTML }
+                <CSSTransitionGroup
+                  transitionAppear
```
ianoxley: And again: you'll need a `component='div'` prop here, otherwise the HTML will be invalid.

See #1708.

src/components/flash/flash.js
ianoxley: I'm curious why we need `this.state.open`, when all we seem to do is set it to the value of `this.props.open`?

Can we not just use `this.props.open`?

JohnStuartBarrie: If we did this ```this.prop.open && (
        <Portal>``` Portal and content are automatically removed. ```clearTimeout(this.fadeLeaveTimeout);
      this.fadeLeaveTimeout = setTimeout(() => {
        this.setState({ open: nextProps.open });
      }, 2000);``` gives us latency for the fadeout to operate before the Portal and children are torn down. If you can see a better way, sincerely I am all ears.

src/components/flash/flash.js
JohnStuartBarrie: Here we don't need it here for sure.

src/components/flash/flash.js
JohnStuartBarrie: 👍

src/components/flash/__spec__.js
```diff
@@ -348,40 +429,22 @@ describe('Flash', () => {
...
-        'info-dialog',
-        'message',
-        'more-info'
-      ]);
-    });
+  describe("snapshot", () => {
+    guid.mockImplementation(() => 'guid-12345');
+    
+    jest.useFakeTimers();
+    let flashInfo = mount(
```
andrewjtait: Do not snapshot mounted components, only snapshot shallow

src/components/flash/flash.js
```diff
@@ -133,9 +135,14 @@ class Flash extends React.Component {
    * @method componentWillReceiveProps
    * @return(Void)
    */
-  componentWillReceiveProps(prevProps) {
-    if (prevProps.open !== this.props.open) {
-      this.setState({ dialogs: {} });
+  componentWillReceiveProps(nextProps) {
+    clearTimeout(this.removePortalTimeout);
+    if (nextProps.open) {
```
andrewjtait: this should have some additional guard on it so it doesn't call it unnecessarily.

src/components/flash/flash.js
```diff
@@ -133,9 +135,14 @@ class Flash extends React.Component {
...
+      this.setState({ dialogs: {}, open: nextProps.open });
+    } else {
```
andrewjtait: this should have some additional guard on it so it doesn't call it unnecessarily.

src/components/flash/flash.js
andrewjtait: maybe you could just return early if `nextProps.open === this.props.open` to catch both guards in one

src/components/flash/flash.js
```diff
@@ -124,7 +125,8 @@ class Flash extends React.Component {
      * @property dialogs
      * @type {Object}
      */
-    dialogs: {}
+    dialogs: {},
+    open: false
```
ianoxley: [Minor] We should probably add some docs to `open` here, as all the other state and props have some.

iliaoni:
All good on Safari/Chrome/FF. QA passed, nothing broken.

PR #1715, opened 2018-03-28
Carbon 1003 New connect guide

JohnStuartBarrie:
# Description

New guides have been included to help explain the  flux connect function.

### DIFF
--- a/docs/guides/flux.md
+++ b/docs/guides/flux.md
@@ -126,18 +126,14 @@ class AccountInvitation extends React.Component {
   render() {
     return (
       <div>
-        <Textbox name="invitationid"
+        <Textbox name="invitationidTextBox"
           value={ props.invitationid }
         />
-        <Textbox name="firstName"
+        <Textbox name="firstNameTextBox"
           value={ props.firstname }
           onChange={ UserActions.userFirstNameValueUpdated } />
         />
-        <Textbox name="surname"
-          value={ props.surname }
-          onChange={ UserActions.useSurnameValueUpdated } />
-        />
-        <Textbox name="gift"
+        <Textbox name="giftTextBox"
           value={ props.gift }
         />
       </div>
@@ -148,7 +144,6 @@ const mapStateToProps = (userState, giftsState, props) => {
   return {
     invitationid: props.invitationid,
     firstname: props.firstname || userState.get('firstname'),
-    surname: props.surname || userState.get('surname'),
     gift: giftsState.get('gift')
   }
 }
@@ -161,7 +156,7 @@ At the core of it, this is just a React component. Our component renders a Carbo
 
 However, on the last line it calls a connect function (provided by Carbon) to connect our component with our store. This function sets up event listeners for when the store is updated - when it detects a change in the store it will call `setState` on itself with the new data. It also makes the stores data available through the components state using the name we defined for the store (in this case, it is available as `this.state.userStore`).
 
-Through this connection, we can setup several Textboxes, the first Textbox has a static value which is taken from a prop. The second Textbox takes its value from the prop 'firstname'. The 'surname' takes its value from the Userstore. We can also set the `onChange` event to trigger the action we defined earlier - completing the Flux loop!
+Through this connection, we can setup several Textboxes, the `invitationidTextBox` Textbox has a static value which is taken from a prop. The `firstNameTextBox` Textbox takes its value from the prop `firstname` depending on whether the prop has been set, and if not the state from the `UserStore` will be used. The `giftTextBox` will take it's value from the `GiftsStore`. We can also set the `onChange` event to trigger the action we defined earlier - completing the Flux loop!
 
 So this User component could be set up as follows.
 
@@ -173,7 +168,7 @@ class Main extends React.Component {
 
     render() {
       return (
-          <User invitationid={ '912ec803b2ce49e4a541068d495ab570' firstname={ 'John' }/>
+        <User invitationid={ '912ec803b2ce49e4a541068d495ab570' firstname={ 'John' }/>
       )
     }
 }
--- a/docs/guides/flux.md
+++ b/docs/guides/flux.md
@@ -110,6 +110,7 @@ So we now have an action called `userValueUpdated`, which when called dispatches
 ### Connecting a React Component to our Flux Store
 
 The store is now updating its data - but we have no React components connected to the store! Lets set one up:
+Here is an example of how to set up a connected Componen. It shows several that a component can connect to Stores or simply override with props.
 
 ```js
 // ./src/views/user/index.js
@@ -146,8 +147,8 @@ class AccountInvitation extends React.Component {
 const mapStateToProps = (userState, giftsState, props) => {
   return {
     invitationid: props.invitationid,
-    firstname: userState.get('firstname'),
-    surname: userState.get('surname'),
+    firstname: props.firstname || userState.get('firstname'),
+    surname: props.surname ||userState.get('surname'),
     gift: giftsState.get('gift')
   }
 }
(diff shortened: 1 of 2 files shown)


docs/guides/flux.md
```diff
@@ -129,9 +129,14 @@ class UserView extends React.Component {
...
+const mapStateToProps = (userState, props) => {
+  return {
+    value: userState.get('foobar')
+  }
+}
 
 // connect the view component to our store
-export default connect(UserView, UserStore);
+export default connect(UserStore, mapStateToProps)(UserView);
 ```
```
whomwah: Would it be better to use the example from the `connect` source ?

```
function mapStateToProps(stateA, stateB, props) {
  return {
    propFromA: stateA.get('someProp') + props.someProp,
    propFromB: stateB.get('someProp')
  }
}

const ConnectedComponent = connect(storeA, storeB, mapStateToProps)(TargetComponent)
```

docs/guides/flux.md
```diff
@@ -110,35 +110,74 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+      </div>
     );
   }
 }
+const mapStateToProps = (userState, giftsState, props) => {
+  return {
+    invitationid: props.invitationid,
+    firstname: props.firstname || userState.get('firstname'),
+    surname: props.surname ||userState.get('surname'),
+    gift: giftsState.get('gift')
```
whomwah: You've got a typo in there `props.surname ||userState.get('surname')`, you're missing a space. I actually think it was OK when you just had the `props.invitationid` and no other mention of props in there.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+          value={ props.invitationid }
+        />
+        <Textbox name="firstNameTextBox"
+          value={ props.firstname }
+          onChange={ UserActions.userFirstNameValueUpdated } />
+        />
+        <Textbox name="giftTextBox"
+          value={ props.gift }
+        />
+      </div>
```
andrewjtait: you have some syntax issues in this code example (an extra `/>` for example) I would put each prop on a new line too.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
 
 The store is now updating its data - but we have no React components connected to the store! Lets set one up:
+Here is an example of how to set up a connected Component. It shows several ways that a component can connect to Stores or simply be overridden with props.
 
 ```js
 // ./src/views/user/index.js
 
 import React from 'react';
 import { connect } from 'utils/flux';
-import Textbox from 'carbon-react/lib/components/textbox';
```
andrewjtait: we ned the Textbox component

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+import AccountInvitation from 'test/lib/components/accountinvitation';
```
andrewjtait: I don't know why you're importing this, you're defining it in this file

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
 import UserStore from 'stores/user';
+import GiftsStore from 'stores/gifts';
```
andrewjtait: I appreciate you're showing how to connect to multiple stores, but we haven't set up a GiftStore in the example. I would keep it simple with the one store.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
 import UserActions from 'actions/user';
 
-class UserView extends React.Component {
+class AccountInvitation extends React.Component {
```
andrewjtait: Lets keep with the UserView

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
   render() {
     return (
-      <Textbox name="foobar"
-        value={ this.state.userStore.get('foobar') }
-        onChange={ UserActions.userValueUpdated } />
+      <div>
+        <Textbox name="invitationidTextBox"
+          value={ props.invitationid }
+        />
```
andrewjtait: Lets keep it simple with one textbox - the purpose of this demo is to demonstrate a really simple and clear example.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+}
 
 // connect the view component to our store
-export default connect(UserView, UserStore);
+export default connect(UserStore, GiftsStore, mapStateToProps)(UserView);
 ```
 
 At the core of it, this is just a React component. Our component renders a Carbon Textbox and gives it a name.
 
 However, on the last line it calls a connect function (provided by Carbon) to connect our component with our store. This function sets up event listeners for when the store is updated - when it detects a change in the store it will call `setState` on itself with the new data. It also makes the stores data available through the components state using the name we defined for the store (in this case, it is available as `this.state.userStore`).
```
andrewjtait: The last part of this paragraph is no longer true (about the state being available).

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+
+```js
+//./src/view/main.js
+import User from './src/views/user/';
+
+class Main extends React.Component {
+
+    render() {
+      return (
+        <User invitationid={ '912ec803b2ce49e4a541068d495ab570' firstname={ 'John' }/>
```
andrewjtait: You don't need this.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
 
 However, on the last line it calls a connect function (provided by Carbon) to connect our component with our store. This function sets up event listeners for when the store is updated - when it detects a change in the store it will call `setState` on itself with the new data. It also makes the stores data available through the components state using the name we defined for the store (in this case, it is available as `this.state.userStore`).
 
-Through this connection, we can set the Textbox's value to use the value from the store. We can also set the `onChange` event to trigger the action we defined earlier - completing the Flux loop!
+Through this connection, we can setup several Textboxes, the `invitationidTextBox` Textbox has a static value which is taken from a prop. The `firstNameTextBox` Textbox takes its value from the prop `firstname` depending on whether the prop has been set, and if not the state from the `UserStore` will be used. The `giftTextBox` will take it's value from the `GiftsStore`. We can also set the `onChange` event to trigger the action we defined earlier - completing the Flux loop!
```
andrewjtait: Let's simplify this explanation, and keep the example simple.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
 ### Connecting a React Component to our Flux Store
 
 The store is now updating its data - but we have no React components connected to the store! Lets set one up:
+Here is an example of how to set up a connected Component. It shows several ways that a component can connect to Stores or simply be overridden with props.
```
amaalali: A minor error here or vaguely stated. The incoming key-value pair from the `props` object passed to the HOC are overridden if the `mapStateToProps` function in your example have a key-value pair of the same key. So in that way, the props have the lowest precedence.

docs/guides/flux.md
```diff
@@ -110,35 +110,69 @@ So we now have an action called `userValueUpdated`, which when called dispatches
...
+          value={ props.gift }
+        />
+      </div>
     );
   }
 }
+const mapStateToProps = (userState, giftsState, props) => {
+  return {
+    invitationid: props.invitationid,
+    firstname: props.firstname || userState.get('firstname'),
```
amaalali: I'm not sure you want this in an example. I think it might be a bit of a bad pattern as this would be pushing the responsibility of setting a sensible default to the component developer and not the consumer. If it did need a value it would probably be marked as a required prop.

docs/guides/flux.md
amaalali: Also, this can be a functional component, which really highlights how `connect` can be leveraged: lift state up into a function like `mapStateToProps` or higher in the component tree

PR #1727, opened 2018-04-20
Portal ie fix

toni-leigh:
* The code was using `x` and `y` instead of `left` and `top` and those two values are not reported by the IE's none standard object returned by `getBoundingClientRect` - switching to `left` and `top` fixes the problem.

* https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect


andrewjtait:
rebased and re-targeted here https://github.com/Sage/carbon/pull/1728

Issue #1718, opened 2018-04-05
Inline labels position themselves at the bottom of a Textarea


patrickmelvin:
Enforcing inline on a label for a textarea results in it floating near the bottom of the component (level with the bottom of the textarea field, above the field help)

```
<Textarea
  label='Test title'
  labelInline
  fieldHelp='Field Help'
  rows={ 4 }
/>
```

![screen shot 2018-04-05 at 11 52 31](https://user-images.githubusercontent.com/8832592/38359217-f9da0ce4-38c7-11e8-9a1b-b03fe5e4f50b.png)

PR #1732, opened 2018-04-23
Position textarea label at the top

patrickmelvin:
#1718

Moves the label for the textarea to the top of the component.

screen shot 2018-04-23 at 17 14 03

# TODO
- [x] Release notes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# next
+
+## Bug fixes
+
+* Position textarea label at the top of the component
+
 # 3.2.1
 
 Merged in v3.1.3 and v3.1.4
--- a/src/components/textarea/textarea.scss
+++ b/src/components/textarea/textarea.scss
@@ -1,8 +1,8 @@
 @import 'colors';
 
 .carbon-textarea .common-input__label {
-  vertical-align: top;
   margin-top: 9px;
+  vertical-align: top;
 }
 
 .carbon-textarea__input {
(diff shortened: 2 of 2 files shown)


src/components/textarea/textarea.scss
```diff
@@ -1,5 +1,10 @@
 @import 'colors';
 
+.carbon-textarea .common-input__label {
```
ianoxley: This is OK, but I'm wondering whether there's a better way to do this using BEM? e.g. perhaps using `.common-input__label--textarea` or `.carbon-textarea__label` instead of the descendant selector you are using.

Not sure how that'd work - or if it's possible - with the `InputLabel` decorator though. @andrewjtait might know?

src/components/textarea/textarea.scss
```diff
@@ -1,5 +1,10 @@
...
+  vertical-align: top;
+  margin-top: 9px;
```
ianoxley: Can you sort these alphabetically please?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# next
```
ianoxley: I think you can target `3.2.2`.

src/components/textarea/textarea.scss
toni-leigh: I think this is OK as it's using the parent to control the layout of the children, which is how `flex` and `grid` both work - I'm come to the conclusion that using inheritance is OK in CSS in this context

ianoxley:
QA: tested in Chrome, Firefox, Safari, Opera, Edge, IE11 ✅

PR #1742, opened 2018-04-27
Fix Flash component not opening on initial render

### LINKED ISSUE #1741: Flash component doesn't open on initial render
If you set the `Flash` component's `open` prop to `true` on the initial render it doesn't actually open.

``` js
<Flash
  open
  message='This should open'
  onDismiss={ ... }
/>
```

It looks like an issue to do with only updating `this.state.open` on `componentWillReceiveProps`.

ianoxley:
Resolves #1741.

Set `this.state.open` to `this.props.open` in the constructor, so that if `this.props.open` is set to `true` the `Flash` component will open on initial render.

### DIFF
--- a/src/components/flash/flash.js
+++ b/src/components/flash/flash.js
@@ -136,6 +136,14 @@ class Flash extends React.Component {
     open: false
   }
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      dialogs: {},
+      open: this.props.open
+    };
+  }
+
   /**
    * Resets the dialog open states if flash is opened/closed.
    *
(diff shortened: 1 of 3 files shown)


src/components/flash/flash.js
```diff
@@ -136,6 +136,14 @@ class Flash extends React.Component {
     open: false
   }
 
+  constructor(props) {
```
abramin: Do we still need the state object above this code if we're initializing it here now.

ianrobsonsage:
QA Passed, loaded the carbon demo site, edited code to make flash message appear of page laod, it worked as expected.

PR #1757, opened 2018-06-01
Ensure specificity of margin over-ride

toni-leigh:
## Description

* Due to unpredictable load order we need to ensure this margin applies

## TODO
- [ ] Release notes

### Before
image

### After
image

## Testing notes
* This is not apparent on the demo site where the load order obfuscates it, so check the inspector


andrewjtait:
This is being refactored into styled-components so will close.

PR #1759, opened 2018-06-06
Make confirm button contents more flexible

toni-leigh:
* Quick fix to make the Confirm button more flexible


ianoxley:
I'm not sure what the use case is here. Can you give an example of where this would be useful please?

toni-leigh:
@ianoxley I needed to put an icon into a confirm button

andrewjtait:
Release notes need updating to match new format.

toni-leigh:
@andrewjtait addressed

andrewjtait:
Currently being refactored so will close.

PR #1762, opened 2018-06-15
Stop rendering a type attribute for the Icon's type prop

ianoxley:
Do not render the `type` prop as a `type` attribute, as this is invalid HTML.

### DIFF
--- a/src/components/icon/icon.js
+++ b/src/components/icon/icon.js
@@ -101,7 +101,6 @@ const Icon = TooltipDecorator(class Icon extends React.Component {
     delete props.bgSize;
     delete props.bgShape;
     delete props.bgTheme;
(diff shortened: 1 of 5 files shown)


src/components/icon/icon.js
```diff
@@ -101,7 +101,6 @@ const Icon = TooltipDecorator(class Icon extends React.Component {
     delete props.bgSize;
     delete props.bgShape;
     delete props.bgTheme;
-    props.type = this.type;
```
SeanArmstrong: Would there be any use in having this as a data attribute?

ianoxley: Maybe. Although it's already included in the `className` e.g. `icon-tick`.

Issue #1763, opened 2018-06-19
TextArea default height is not the same as other inputs


adrianpastur:
![screen shot 2018-06-19 at 11 13 15](https://user-images.githubusercontent.com/10499070/41591840-c2f9b1ae-73b2-11e8-914f-2949bb05e909.png)

```
.carbon-textarea__input {
    min-height: 31px;
}
```

Will fix this

ianoxley:
Resolved by #1764

PR #1764, opened 2018-06-19
#1763 Textarea not same height as other inputs.

adrianpastur:
# Description

# TODO
- [x] Release notes

https://github.com/Sage/carbon/issues/1763

# Screenshots / Gifs
Before change:
screen shot 2018-06-19 at 11 13 15

# Testing Instructions


ianoxley:
I'm not sure we'd want `<textarea>` to be the same height as other inputs, as typically they're used for multiline input. If they're the same height as other inputs this isn't immediately clear.

It might be better to default it to `2 * $input-common-height` instead?

harpalsingh:
@ianoxley PR removes the auto and makes a "min-height" to match inputs, so it always is at least the same height, which I think is fine?

ianoxley:
@harpalsingh I'm not sure. A `<textarea>` rendered by Carbon isn't resizable. If its height defaults to the same height as other inputs do you think that might make it harder to enter any sizeable amount of text?

harpalsingh:
Min-height though, which means it can get bigger? A textarea doesn't resize unless you give it parameters and/or add content.

adrianpastur:
I've got some more information that might help decide what's best for this. On one of the pages i am working on, when making all inputs the same height, the TextArea will expand automatically when the input goes over to the next line,  but when applying the maximum characters limit, it seems to stop it from expanding. So after discussing with Patrick Port, we believe that (for that page anyway) it will look better with the minimum height of TextAreas to be set to twice the size of the height of other inputs, just in case the maximum character validation is added on the component.

src/components/textarea/textarea.scss
```diff
@@ -1,12 +1,14 @@
...
+@import 'input-field.scss';
+
 
 .carbon-textarea .common-input__label {
   margin-top: 9px;
   vertical-align: top;
 }
 
 .carbon-textarea__input {
-  height: auto;
```
andrewjtait: we should not remove the `height: auto` - this is important to ensure Textarea is a proper height. It should only be the same height as other inputs if `expandable` is set I think.

adrianpastur: So should I set the `height: auto` next to `min-height`? Or cancel the PR and make my changes for the page I need?

andrewjtait: I think having `height: auto` and `min-height: $input-common-height` should work nicely together 👍

adrianpastur: @ianoxley @harpalsingh 

So do you guys think the `min-height` should be ` $input-common-height` or `2 *  $input-common-height`?

andrewjtait: I would think `min-height` should be `$input-common-height` as it needs to sit inline with other inputs when `expandable` prop is set to `true`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * `Browser` now has a `redirectAfter` method that redirects to the given URL after the given number of seconds.
 * `Menu` now outputs semantic HTML. Links are rendered in HTML lists, with submenus rendered with nested lists.
+* `TextArea` now has the same height as the rest of the input fields.
```
ianoxley: I think this could do with a slight change - can you change it to say something like:

> `TextArea` now sets its `min-height` to match the height of other input fields.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * `Browser` now has a `redirectAfter` method that redirects to the given URL after the given number of seconds.
 * `Menu` now outputs semantic HTML. Links are rendered in HTML lists, with submenus rendered with nested lists.
+* `TextArea` TextArea now sets its `min-height` to match the height of other input fields.
```
andrewjtait: just realised you need to move this to new /change_log/next !

Issue #1766, opened 2018-06-22
Warning icon overlapping text inside input field.


adrianpastur:
**Issue:**
![screen shot 2018-06-20 at 15 04 42](https://user-images.githubusercontent.com/32830770/41782551-7ccc7900-7632-11e8-954b-229da1f80d4b.png)

**Work around:**
```
.carbon-dropdown__input.common-input__input--warning.common-input__input {
    text-overflow: ellipsis;
    padding-right: 60px;
}
```
PR Raised:

PR #1767, opened 2018-06-22
1766: Fixed warning icon overlapping text in input field.

adrianpastur:
# Description

Added padding and handled text overflow to avoid warning icon overlapping the text inside input fields.

# TODO
- [x] Release notes

https://github.com/Sage/carbon/issues/1766

# Screenshots / Gifs

In-app QA:
screen shot 2018-06-22 at 15 47 21

Carbon website QA:
screen shot 2018-07-17 at 10 28 23
screen shot 2018-07-17 at 10 28 39
screen shot 2018-07-17 at 10 29 21

# Testing Instructions

Verify input fields that will use icons such as the warning icon work as expected.


andrewjtait:
ℹ️ Need to QA that the input width doesn't jump around as a result of the padding changing.

src/utils/decorators/input/input.scss
```diff
@@ -255,6 +255,8 @@
   .common-input__input--#{$type},
   .common-input__input--#{$type}:hover {
     border-color: $color !important; // this is the correct use of !important
+    text-overflow: ellipsis;
+    padding-right: 60px;
```
ianoxley: Can you sort the CSS properties please?

adrianpastur:
@andrewjtait @SeanArmstrong would an IE QA of the carbon website suffice?

src/utils/decorators/input/input.scss
```diff
@@ -255,6 +255,8 @@
   .common-input__input--#{$type},
   .common-input__input--#{$type}:hover {
     border-color: $color !important; // this is the correct use of !important
+    padding-right: 60px;
```
andrewjtait: Is this not going to affect ALL inputs? Whereas you want to target just dropdowns (or inputs with a button to the right?) Date for example may need updating in a different way.

adrianpastur: Our scope was just for dropdowns. We can test how it affects other inputs and see if there are any issues, or I can target the change just for dropdowns

andrewjtait: You need to regression test other inputs at least, as this is a shared module you have changed. I suspect they will be negatively affected in this case.

src/components/dropdown-filter/dropdown-filter.scss
```diff
@@ -41,3 +41,25 @@
     }
   }
 }
+
+@mixin common-input-text-wrapping($type, $color) {
+  .common-input__input--#{$type},
+  .common-input__input--#{$type}:hover {
+    border-color: $color !important; // this is the correct use of !important
```
andrewjtait: did you not already have border-color?

change_log/next/input_icon_overlap.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "`Input` now warning icon will not overlap over text inside input field"
```
andrewjtait: You should update this to `Dropdown` rather than `Input`

PR #1768, opened 2018-06-25
upgraded carbon-factory to v4.2.2

andrewjtait:
This includes a change to support compilation on Windows environments, and required updating paths in the Webpack config.


abramin:
Windows? Does Carbon also support Ubuntu Linux?

andrewjtait:
@abramin it works in Peppermint ;)

PR #1770, opened 2018-06-27
[1767] CustomDragLayer can now take className prop

JanTaras29:
# Description
For draggable lines work I require to extend to apply className to customDragLayer component - this would be a neat feature to have in Carbon to make customLayer even more custom.

# TODO
- [x] Release notes

# Screenshots / Gifs

# Testing Instructions

### DIFF
--- a/src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
+++ b/src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import { DragLayer } from 'react-dnd';
 
 const collect = (monitor) => {
@@ -62,6 +63,10 @@ class CustomDragLayer extends React.Component {
     };
   }
 
+  getClassName = (props) => {
+    return classNames('custom-drag-layer', props.className);
+  }
+
   setClonedChildWidth(props) {
     this.width = props.draggableNode.getBoundingClientRect().width;
   }
@@ -82,7 +87,7 @@ class CustomDragLayer extends React.Component {
 
   render() {
     return (
-      <div className='custom-drag-layer'>
+      <div className={ this.getClassName(this.props) }>
         <div
           className='custom-drag-layer__container'
           ref={ (node) => { this._container = node; } }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Improvements
 
 * `Menu` now outputs semantic HTML. Links are rendered in HTML lists, with submenus rendered with nested lists.
+* `CustomDragLayer` component can now take className prop. 
 
 # 4.1.0
 
--- a/src/components/drag-and-drop/custom-drag-layer/__spec__.js
+++ b/src/components/drag-and-drop/custom-drag-layer/__spec__.js
@@ -230,4 +230,22 @@ describe('CustomDragLayer', () => {
       })
     })
   });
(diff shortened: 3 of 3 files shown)


src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
```diff
@@ -62,6 +62,10 @@ class CustomDragLayer extends React.Component {
     };
   }
 
+  getKlassName = (props) => {
+    return `custom-drag-layer ${props.className}`;
```
andrewjtait: if `props.className` is not set, then this will add a class of `undefined` into the HTML.

Instead please use the [classNames](https://github.com/JedWatson/classnames) package we use in Carbon:

```
classNames('custom-drag-layer', props.className);
```

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
```diff
@@ -82,7 +86,7 @@ class CustomDragLayer extends React.Component {
 
   render() {
     return (
-      <div className='custom-drag-layer'>
+      <div className={ this.getKlassName(this.props) }>
```
andrewjtait: as you have set this up to pass an argument to the function, it might as well send the relevant arg: `this.getKlassName(this.props.className)`

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
```diff
@@ -62,6 +62,10 @@ class CustomDragLayer extends React.Component {
     };
   }
 
+  getKlassName = (props) => {
```
andrewjtait: I don't think we use `klass` with a `k` anywhere else, might as well stick with `getClassName`?

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
andrewjtait: alternatively your function has access to `this` - but I guess that is not as functional!

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
```diff
@@ -62,6 +62,10 @@ class CustomDragLayer extends React.Component {
...
+    return `custom-drag-layer ${props.className || ''}`;
```
andrewjtait: We use the [classNames](https://github.com/JedWatson/classnames) package in Carbon to handle this:

```
classNames('custom-drag-layer', props.className);
```

toni-leigh:
Linking to the ticket has no meaning in an open source repo, anyone outside of Sage won't be able to view the ticket details and the codes aren't useful either

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
toni-leigh: We solve the same problem with the key `class` by using `className` instead (though `klass` is also an alternative in similar situations, `className` is preferred here as it's just a name being defined rather than an object)

JanTaras29: Correct - will refactor that 👍.

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.js
JanTaras29: I styled access to props to be similar to `this.getItemStyles(this.props)` in this file.

JanTaras29:
Local QA passed

PR #1771, opened 2018-06-28
Add redirectAfter method to Browser

ianoxley:
Add `redirectAfter` to the `Browser` class. This method takes 2 arguments: url, and seconds.

It redirects to the given URL after the given number of seconds have elapsed e.g. if you need to redirect after displaying a `Flash` message.

### DIFF
--- a/src/utils/helpers/browser/browser.js
+++ b/src/utils/helpers/browser/browser.js
@@ -58,6 +58,17 @@ const Browser = {
   },
 
   /**
+   * Redirects to URL after the given number of seconds have elapsed
+   *
+   * @method redirectAfter
+   * @param url => URL string format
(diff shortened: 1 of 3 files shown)


src/utils/helpers/browser/browser.js
```diff
@@ -57,6 +57,17 @@ const Browser = {
...
 
+  /**
+   * Redirects to URL after the given number of seconds have elapsed
+   *
+   * @method redirectAfter
+   * @param url => URL string format
+   * @param seconds => the number of seconds to wait before redirecting
+   */
+  redirectAfter: (url, seconds) => {
+    return setTimeout(() => Browser.redirectTo(url), seconds);
```
chrisbarber86: Isn't this milliseconds rather than seconds?

PR #1777, opened 2018-07-03
Adding transitionAppear and transitionAppearTimeout to modal component

halezmo:
# Description
Adding transitionAppear and transitionAppearTimeout to modal component, so we have animation on initial mount

# TODO
- [x] Release notes

# Screenshots / Gifs

# Testing Instructions


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 4.1.1
```
SeanArmstrong: I think we should reserve patch releases for quick little fixes on major/minor versions.

This can just go into 4.2.0 unless there is a specific reason you need the change urgently

halezmo: Yea I can put it for 4.2.0

halezmo:
@SeanArmstrong in this case I had to create branch from 4.1.0 as some change is master are causing some tests in s1_gac_ui/artefacts to fail, as I remember failure was caused by some icon/icon type

SeanArmstrong:
Yeah that will cause some snapshots to fail, maybe that should of been breaking? But it should be simple just to update them.

If you want to do a patch you will probably have to do a PR into release-candidate or release depending on the state of those branches. I don't think we want to do that though @andrewjtait

halezmo:
@SeanArmstrong Can we have approval on this one, release notes updated

halezmo:
@SeanArmstrong I am still not sure which branch I should target with this PR

PR #1790, opened 2018-07-13
1790 - the WithDrag allowTextSelection method fix for Internet Explorer

JanTaras29:
# Description
When withDrag element was present on the page the overly specific binding on `startselect` event caused the inputs on the page to be unselectable for Microsoft browsers. The preventing of event propagation was scoped only to icon by which the user will drag the icon.

# TODO
- [x] Release notes

# Screenshots / Gifs

# Testing Instructions

The app was tested on Safari and IE and both tests were successful.


src/components/drag-and-drop/with-drag/with-drag.js
```diff
@@ -53,7 +53,7 @@ class WithDrag extends React.Component {
   // In Safari it changes the mouse cursor when dragging because it thinks text is being selected
   // We test if the target is an html element (not text) or if we already know the user is dragging
   allowTextSelection = (event) => {
-    if ((event.target instanceof HTMLElement) || this.dragging) {
+    if (event.target.tagName === 'SPAN' || this.dragging) {
```
andrewjtait: this is true for `Icon`, but this component is not limited to just dragging `Icon` components

JanTaras29: @andrewjtait What other elements should we protect against?

andrewjtait: what about checking if the tag is a HTMLElement but not an Input or something?

JanTaras29: @andrewjtait I think it would work, we will see how to refactor that. 👍

halezmo:
@andrewjtait Any suggestion on what would be the best approach on testing this carbon change? 🤔

PR #1792, opened 2018-07-16
Adds onMaxRetries call to Poller

toni-leigh:
## Description

* We need to be able to handle a user defined action when the Poller reaches it's try limit rather then outputting a `console.warn`.
* The console output will still remain the default.

## TODO
- [x] Release notes


src/utils/helpers/poller/__spec__.js
```diff
@@ -135,6 +134,22 @@ describe('poller', () => {
       });
     });
 
+    describe('when polling termkinates and theres an onMaxRetries function', () => {
```
samcartwrigh: typo on terminates

src/utils/helpers/poller/__spec__.js
```diff
@@ -135,6 +134,22 @@ describe('poller', () => {
...
 
+    describe('when polling terminates and theres an onMaxRetries function', () => {
+      it('calls the onMaxRetries function', () => {
+        const onMaxRetriesSpy = jest.fn();
+        functions.onMaxRetries = onMaxRetriesSpy;
+
+        Poller({ url }, functions, { interval: 1000, retries: 2 });
+        jest.runTimersToTime(1000);
+
+        for (let i = 0; i < 2; i++) {
```
andrewjtait: do you want to asset above this that the function has NOT been called? having your assertion that it has been called at the bottom does not assert _when_ it was called.

src/utils/helpers/poller/poller.js
```diff
@@ -126,7 +127,10 @@ function getFunctions(functions) {
     handleError: functions.handleError || null,
     conditionMet: functions.conditionMet || (() => { return false; }),
     conditionNotMetCallback: functions.conditionNotMetCallback || (() => { return false; }),
-    terminate: functions.terminate || (() => { return false; })
+    terminate: functions.terminate || (() => { return false; }),
+    onMaxRetries: functions.onMaxRetries || (() => {
+      console.warn('The poller has made too many requests - terminating poll'); // eslint-disable-line no-console
```
SeanArmstrong: This should probably use the logger rather than warning itself.

This will stop anything showing in production and also concentrates the no-console ESLint ignore to one place

src/utils/helpers/poller/__spec__.js
samcartwrigh: @andrewjtait - just tidying up this PR. Is your comment answered by line 146?

src/utils/helpers/poller/__spec__.js
```diff
@@ -129,11 +129,27 @@ describe('poller', () => {
...
+    describe('when polling terminates and theres an onMaxRetries function', () => {
+      it('calls the onMaxRetries function', () => {
+        const onMaxRetriesSpy = jest.fn();
+        functions.onMaxRetries = onMaxRetriesSpy;
+
+        Poller({ url }, functions, { interval: 1000, retries: 2 });
+
+        for (let i = 0; i < 2; i++) {
+          expect(onMaxRetriesSpy).not.toHaveBeenCalled();
+          jest.runTimersToTime(1000);
```
patrickmelvin: Does this loop tie up the test for 2 seconds?

PR #1794, opened 2018-07-17
carbon-factory v5 upgrade

andrewjtait:


src/components/carousel/carousel.js
```diff
@@ -164,7 +164,7 @@ class Carousel extends React.Component {
 
     if (newIndex === currentIndex) {
       return;
-    } else if (newIndex > currentIndex) {
+    } if (newIndex > currentIndex) {
```
andrewjtait: err - not sure about this?

SeanArmstrong: huh

SeanArmstrong: Maybe if it was
```
if (newIndex === currentIndex) { return; }
if (newIndex > currentIndex) {
  this.transitionDirection = NEXT;
} else {
  this.transitionDirection = PREVIOUS;
```

andrewjtait: yeah maybe it is ok - it highlights that the else is unecessary

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -158,9 +158,11 @@ class DropdownFilterAjax extends DropdownFilter {
    */
   handleBlur = () => {
     if (!this.blockBlur) {
-      const filter = this.props.create ? this.state.filter : null;
       // close list and reset filter
-      this.setState({ open: false, filter });
+      this.setState(prevState => ({
+        open: false,
+        filter: this.props.create ? prevState.filter : null
```
SeanArmstrong: Is this just because it was a ternary operator?

src/components/grouped-character/grouped-character.js
```diff
@@ -139,12 +139,12 @@ const GroupedCharacter = Input(InputLabel(InputValidation(
 
   isValidKeypress = (ev) => {
     return (
-      !Events.isNumberKey(ev) &&
-      !Events.isAlphabetKey(ev) &&
-      !Events.isTabKey(ev) &&
-      !Events.isDeleteKey(ev) &&
-      !Events.isBackspaceKey(ev) &&
-      !Events.isNavigationKey(ev)
+      !Events.isNumberKey(ev)
```
SeanArmstrong: I prefer the previous

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
andrewjtait: It is bad to `setState` with existing state as it is not always a synchronous event, so we needed to use the function version.

src/components/grouped-character/grouped-character.js
andrewjtait: I'm indifferent - will go with whatever people prefer. See https://eslint.org/docs/rules/operator-linebreak

andrewjtait: One advantage to having them down the left means you can look straight down the left edge to see the operators.

SeanArmstrong:
Change is scary but 👍

src/components/grouped-character/grouped-character.js
andrewjtait: I tried switching it and comparing the diff, it is actually easier to understand whats happening with the operator on the left - otherwise it gets lost at the end of the line.

andrewjtait:
Tested components with significant changes within the demo site ✅

PR #1796, opened 2018-07-20
Add formatResponse prop to DropdownFilterAjax

gedskikas:
# Description
Adding `formatResponse` callback function to DropDownTableAjax props, and call `formatResponse` in `ajaxUpdateList` if it is defined. This allows us to customise the response we receive.

Adding `formatRequest` callback function to DropDownTableAjax props, and call `formatRequest` in `getParams ` if it is defined. This allows us to give complete freedom over param creation to the user. The user will now have the ability to customise what, if any, the parameter names are. This is especially useful when in use with APIs, as the user can accommodate the parameter names used by a given API.

# Sample use case 
Assuming we're using our custom API endpoint:

`/my_custom_api?p={num pages}&r={num row}&s={search term}`

which responds with following JSON:
```json
{ 
  "p": 1,
  "c": 100,
  "r": [ 
    { "id": 1, "name": "foo" },
    { "id": 2, "name": "bar" }
  ]
}
```
We define React component as (pseudocode):
```javascript
class MyDropdownAjaxFilter {

  render() {
    return (
        <DropdownAjaxFilter 
          path="/my_custom_api"
          formatResponse={this.formatResponse}
          formatRequest={this.formatRequest}
        />
    );
  }

  formatResponse = (responseBody) => {
    return {
      records: responseBody.c,
      items: responseBody.r,
      page: responseBody.p
    }
  }

  formatRequest = (params) => {
    return {
      p: params.page,
      r: params.rows,
      s: params.query
    }
  }
}
```
The above callbacks will allow us to customise both, request URI to follow our endpoint conventions and also re-structure the response to provide what the component expects.

# TODO
- [x] Release notes


src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -221,12 +230,13 @@ class DropdownFilterAjax extends DropdownFilter {
...
       .end(this.ajaxUpdateList);
   }
-
   /**
    * Applies some data from AJAX to the list
    */
   ajaxUpdateList = (err, response) => {
-    this.updateList(response.body.data[0]);
+    this.updateList(
+      this.props.formatResponse ? this.props.formatResponse(response.body.data[0]) : response.body.data[0]
```
SeanArmstrong: Should we just pass response or response.body here as there may be custom logic depending on if a error occurred and we don't know for sure that data will be an array?

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -207,26 +225,39 @@ class DropdownFilterAjax extends DropdownFilter {
    *
    * @method getData
    * @param {String} query The search term
-   * @param {Object} page The page number to get
    */
-  getData = (query = '', page = 1) => {
+  getData = (query = '') => {
```
noizex: After checking a bit more in depth on it - we need to keep `page` param on `getData` because it's being called all over the place and that's the real value of page, not what's in the state it would seem. Check here: 
https://github.com/Sage/carbon/blob/595298425833d81f6f0d47a92e86ab6271f26471/src/components/dropdown-filter-ajax/dropdown-filter-ajax.js#L169
https://github.com/Sage/carbon/blob/595298425833d81f6f0d47a92e86ab6271f26471/src/components/dropdown-filter-ajax/dropdown-filter-ajax.js#L217

So rather than pick the page value later on in `getParams`, we need to pass it down form `getData` call.

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -207,26 +225,39 @@ class DropdownFilterAjax extends DropdownFilter {
...
+    return params;
+  }
+
   /**
    * Applies some data from AJAX to the list
    */
   ajaxUpdateList = (err, response) => {
-    this.updateList(response.body.data[0]);
+    this.updateList(
+      this.props.formatResponse ? this.props.formatResponse(response.body.data) : response.body.data[0]
```
noizex: It has to be `response.body` as `data` already goes into the JSON response structure and makes assumption about what the response looks like. So while we keep it `body.data[0]` for old path, we need just `body` when we use formatting callback.

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -126,6 +126,24 @@ class DropdownFilterAjax extends DropdownFilter {
      */
     create: PropTypes.func,
 
+    /**
+     * A callback function used to format the Ajax
+     * response into the format required by the table
+     *
+     * @property formatResponse
+     * @type {Function}
```
noizex: Here it's probably quire important to annotate what the return value should be for this method, as the component expects quite strict value here - it has to be object with just following fields: 
{ 
`records` - number of items returned 
`items` - array of items in a format `{ id: ..., name: ... }`
`page` - current page number
}
Otherwise there will be error, so we should maybe even consider asserting on the return value and checking it contains all required fields or issuing warning in console? What's our approach here @andrewjtait ?

PR #1798, opened 2018-07-23
Added a new theme for buttons

sendorf:
# Description
Created a new theme for buttons `magenta-dull`
# TODO
- [x] Release notes

# Screenshots / Gifs
magenta-dull-primary
magenta-dull-secondary

# Testing Instructions


src/style-config/buttons.scss
```diff
@@ -28,6 +28,7 @@ values are [ background/border color, hover, active, text ]
 $buttonColorSets: (
  ("blue", $blue, $azul, $blue-bright, $white),
  ("magenta", $magenta, $magenta-dull, $magenta-light, $white),
+ ("magenta-dull", $magenta-dull, $magenta-dull, $magenta-light, $white),
```
toni-leigh: There'll be no hover style here, I'd ask UX about this @harpalsingh

toni-leigh: Maybe we could use `darken`?

harpalsingh: The new DLS work actually using a 20% black color on top of the base color to imply "darker", so not the darken() sass mixin but rather mix(black, $magenta-dull, 20), since darken does not directly only add black.

toni-leigh: Should we use that in this case?

harpalsingh: Yea, since we have no other set color, we should do that, as they will all be updated in time to use that formula

laurenvthompson:
As per @toni-leigh and @harpalsingh's comments, we'll pass the UX review when a colour style is implemented.

src/style-config/buttons.scss
```diff
@@ -28,6 +28,7 @@ values are [ background/border color, hover, active, text ]
 $buttonColorSets: (
  ("blue", $blue, $azul, $blue-bright, $white),
  ("magenta", $magenta, $magenta-dull, $magenta-light, $white),
+ ("magenta-dull", $magenta-dull, mix(black, $magenta-dull, 20), mix(white, $magenta-dull, 20), $white),
```
toni-leigh: I'm not blocking on this, but should we repeat this pattern for all the colours here? I feel like this could be an option, but also that the black / white mix might need to be different depending on the shade @harpalsingh curious

PR #1807, opened 2018-08-07
Extending DropdownFilterAjax component

gedskikas:
# Description
Extending `DropdownFilterAjax` component to allow us to track the Request. Also introducing new handling for the Requests, which reduces the amount of unnecessary calls. 

Removing `"name": "events"` from `package.json`, as in some cases, this was breaking the test suite.

# TODO
- [x] Release notes


src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -246,12 +246,13 @@ class DropdownFilterAjax extends DropdownFilter {
...
   getData = (query = '', page = 1) => {
     this.setState({ requesting: true });
-    Request
+    const request = Request
       .get(this.props.path)
       .query(this.getParams(query, page))
       .query(this.props.additionalRequestParams)
       .set('Accept', this.props.acceptHeader)
       .end(this.ajaxUpdateList);
+    this.setState({ request });
```
andrewjtait: I don't think this should be stored in state. Can discuss with you if you want to go over what you're trying to achieve.

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -182,7 +198,14 @@ class DropdownFilterAjax extends DropdownFilter {
...
   handleVisibleChange(ev) {
     super.handleVisibleChange(ev);
-    this.getData(ev.target.value, 1);
+    if (this.getDataTimeout) {
+      clearTimeout(this.getDataTimeout);
+    }
+    const query = ev.target.value;
+    this.getDataTimeout = setTimeout(
+      () => this.getData(query, 1),
+      (this.props.dataRequestTimeout || 500)
```
andrewjtait: you could add this as a `defaultProp`

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -198,6 +221,10 @@ class DropdownFilterAjax extends DropdownFilter {
         filter: this.props.create ? prevState.filter : null
       }));
 
+      if (this.pendingRequest !== null) {
```
andrewjtait: should this also clear the timeout if there is one?

change_log/next/dropdown_filter_ajax_request.yml
```diff
@@ -0,0 +1 @@
+Improvement: "Improved DropdownFilterAjax component Request handling"
```
andrewjtait: Can you add a bit more detail here so people know what to expect with the upgrade.

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
halezmo: Yea makes sense to add it also here

change_log/next/dropdown_filter_ajax_request.yml
```diff
@@ -0,0 +1 @@
+Improvement: "Improved DropdownFilterAjax component Request handling, timeout added so request creation will be delayed (will wailt until typing is completed)"
```
andrewjtait: typo in `wait`

src/components/dropdown-filter-ajax/dropdown-filter-ajax.js
```diff
@@ -198,6 +222,14 @@ class DropdownFilterAjax extends DropdownFilter {
         filter: this.props.create ? prevState.filter : null
       }));
 
+      if (this.getDataTimeout) {
```
noizex: This sounds a bit confusing, getDataTimeout sounds like a getter method rather than class member - `dataFetchTimeoutId` would sound quite a bit more like a variable that holds timeout id not an action? or something shorter but not starting with a verb.

halezmo:
Passed testing as part of test session perfomed on https://github.com/Sage/s1_gac_ui/pull/245

gedskikas:
Passed QA as mentioned by most recent comment made by Marko, sanity tested on Carbon site to ensure it still works as well.

Issue #1808, opened 2018-08-07
Empty table should allow configurable content


toni-sharpe:
Currently the empty table just displays 'no results to show'.

We need an enhancement that will allow the empty table to display some user defined content.

A prop called `contentForEmptyTable` is to be added to the `Table` component which will be used to render in the empty table cell if the table is empty.

philearley:
This was implemented by: https://github.com/Sage/carbon/pull/1938

PR #1810, opened 2018-08-07
Empty Table #1808

toni-leigh:
# Description

* Created new empty table handling code.
* [scouting] extracted out current empty table logic into a component to reduce the complexity of the table component rather than adding to it with this new functionality.

# TODO
- [x] Release notes

https://github.com/Sage/carbon/issues/1808


toni-leigh:
Invalid, it's better to handle this in the component then the dev can do it case by case for their particular context

Issue #1818, opened 2018-08-13
Convert components which extend to compose


andrewjtait:
We currently have many components which extend others and override methods. This causes too much complexity and ties the components too closely together. We should instead look to update these to compose other components.

For example we have many Dialog components, such as Alert which extend Dialog. Instead they should render a Dialog and pass props in or change CSS based on context.

ianforster-sage:
Updated as `pinned` as this is still a current issue

PR #1819, opened 2018-08-13
updated Alert to render a Dialog instead of extend it

andrewjtait:
Alert now renders a Dialog, instead of extending it - see https://github.com/Sage/carbon/issues/1818

### DIFF
--- a/change_log/next/CARBON-1818.yml
+++ b/change_log/next/CARBON-1818.yml
@@ -1,0 +1,1 @@
+Other: "[CARBON-1818](https://github.com/Sage/carbon/issues/1818) - Updated Alert to not extend Dialog, but render a Dialog instead."
--- a/src/components/alert/__definition__.js
+++ b/src/components/alert/__definition__.js
@@ -21,7 +21,6 @@ let definition = new Definition('alert', Alert, {
     size: OptionsHelper.sizesFull
   },
   propValues: {
-    ariaRole: 'alertdialog',
     title: 'Attention!',
     children: 'This is an example of a alert.'
   },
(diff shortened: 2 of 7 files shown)


src/components/alert/alert.js
```diff
@@ -1,82 +1,21 @@
...
-      // Firefox loses focus unless we wrap the call to
-      // this.focusDialog in setTimeout
-      setTimeout(this.focusDialog);
-    }
-  }
-}
+Alert.defaultProps = {
+  'data-role': 'alertdialog',
+  size: 'extra-small'
+};
```
ianoxley: We should set `ariaRole: 'alertdialog'` here as well - https://www.w3.org/TR/wai-aria/#alertdialog

src/components/alert/__definition__.js
```diff
@@ -21,7 +21,6 @@ let definition = new Definition('alert', Alert, {
     size: OptionsHelper.sizesFull
   },
   propValues: {
-    ariaRole: 'alertdialog',
```
ianoxley: We should keep this so the demo site renders the `Alert` with the correct ARIA role.

change_log/next/CARBON-1818.yml
```diff
@@ -0,0 +1 @@
+Other: "[CARBON-1818](https://github.com/Sage/carbon/issues/1818) - Updated Alert to not extend Dialog, but render a Dialog instead."
```
toni-leigh: Jira link will be broken in the release notes. This will make things difficult for all the developers outside of Sage who use Carbon

PR #1820, opened 2018-08-15
Replace ncp with cpx in generate-docs.js script

ianoxley:
Add `ncp` back to `devDependencies` as it is used in the `generate-docs` script.


ianoxley:
At desk QA done with @victormartins ✅

PR #1821, opened 2018-08-16
Modal js refactor

JanTaras29:
# Description

The modal element was refactored to simplify its event binding mechnanisms. Previously it was handled in `componentDidUpdate` method which proved to be stable for most, but not all cases.
Now the binding of 'keyup' event was split into `componentDidMount` & `componentWillUnmount` to make sure that the binding is there regardless of the props. The check that assured that only opened component is closed was moved to `closeModal` method.

# TODO
- [ ] Release notes

# Screenshots / Gifs

# Testing Instructions

### DIFF
--- a/src/components/dialog/__spec__.js
+++ b/src/components/dialog/__spec__.js
@@ -120,6 +120,35 @@ describe('Dialog', () => {
           });
         });
       });
+
+      describe('when Dialog is closed and listening', () => {
+        beforeEach(() => {
+          jest.useFakeTimers();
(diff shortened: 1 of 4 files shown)


ianoxley:
This will need QA'ing in the app against all components that are derived from Modal.

JanTaras29:
@ianoxley I agree, this changes the behaviour of a bunch of components. We intend to test all of them in the Carbon demo page and in the app itself.

PR #1823, opened 2018-08-17
Input validation responsiveness improvement

gedskikas:
# Description
Improving responsiveness of input validation by allowing us to customise the amount of time it takes for a validation message to disappear (unless the `timeToDisappear` prop is passed, by default it is now instant) and also removing the fade in/out effect.

# TODO
- [x] Release notes

# Screenshots / Gifs
Old behaviour with the fade
validation-fade

New behaviour, no fade
responsive-contact

New behaviour, but with the prop set to make the disappearance a bit delayed
delayed-error-validation

# Testing Instructions
Ensure input field validation still works with warning/error/info messages


src/utils/decorators/input-validation/input-validation.js
```diff
@@ -630,7 +652,7 @@ const InputValidation = (ComposedComponent) => {
         iconStyle = { [`${this.props.align}`]: `${100 - this.props.labelWidth}%` };
       }
 
-      const errorMessage = (!this.state.immediatelyHideMessage || this.state.messageLocked) && (
```
noizex: As we removed `immediatelyHideMessage` flag from checking here it has no effect anymore. It would make sense to remove its occurance from anywhere in the code, but first I suggest searching if the `immediatelyHideMessage` call is not used somewhere to hide the message, as it will be no longer working.

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -173,7 +177,16 @@ const InputValidation = (ComposedComponent) => {
        * @property
        * @type {Array}
        */
-      info: PropTypes.array
+      info: PropTypes.array,
+
+      /**
+       * Number which sets timing of when the message will disappear
```
noizex: 💄 Cosmetic thing but it would help to state here that the time is expected to be in miliseconds

src/components/textbox/__spec__.js
```diff
@@ -37,7 +37,7 @@ describe('Textbox', () => {
     });
 
     it('is decorated with a validation if a error is present', () => {
-      instance.setState({errorMessage: 'Error', valid: false});
+      instance.setState({errorMessage: 'Error', valid: false, messageShown: true});
```
noizex: ❓ Why do we need to set it explicitly now as state, compared to how it worked before?

adrianpastur:
Tested on the demo site and in-app. ✅

PR #1833, opened 2018-09-04
Remove Parcelify Loader

### LINKED ISSUE #1832: Webpack not recompiling Sass files when changed
Steps to reproduce:

1. Run `npm start`
2. Edit any `.scss` file e.g. `menu-item.scss`

Expected results:
The change to the Sass file causes webpack to recompile

Actual results:
Nothing is recompiled.

It looks like the fix is to require / import the relevant Sass file(s) in the components that use them:

``` js
// For example
import './menu-item.scss';
```

ianoxley:
Remove Parcelify Loader and import the Sass files into the relevant files e.g. `menu-item.js` now imports `menu-item.scss`. ~Without this webpack doesn't know about the Sass files, hence why it wasn't recompiling them.~

Resolves #1832


andrewjtait:
Might also be worth updating the docs to remove references to parcelify loader https://github.com/Sage/carbon/search?q=parcelify&unscoped_q=parcelify ?

demo/views/chrome/footer/footer.js
```diff
@@ -1,26 +1,46 @@
...
+                <Icon type='dribbble' />
+              </a>
+              <a
+                href='https://github.com/Sage/carbon'
+                target='_blank'
+                rel='noopener noreferrer'
+                className='demo-footer__link'
+              >
+                <Icon type='github' />
+              </a>
```
andrewjtait: why was this changed? for the `rel` attr? If so you do you not just need to whitelist this prop as safe? https://github.com/Sage/carbon/blob/master/src/components/link/link.js#L153

(but also is not unrelated to this PR?)

ianoxley: It was to fix a CSS specificity issue when running the demo site locally, meaning the footer links were the wrong colour.

ESLint complained about the lack of a `rel` attribute on external links, so I added that as well for good measure.

PR #1836, opened 2018-09-10
Upgrade Carbon Factory from v5.0.0 to v5.2.1

ianoxley:
Upgrade `carbon-factory` from `v5.0.0` to `v5.2.1`, which enables source maps by default in the webpack config.

Fixed a test in the router specs, as [Jest now sets `testURL` to localhost](https://github.com/facebook/jest/pull/6792) instead of `about:blank`.
which enables source maps by default in the webpack config.


sianford:
Checked layouts, components, styles, and other features had not broken in Safari and Chrome browsers. Also checked source maps. Found some issues (CARBON-1103 - CARBON-1108) but not related to these changes. Tests passed.

PR #1837, opened 2018-09-10
DropdownFilter/Ajax Improvements: Add additional prop to filter ajax and allow clearing of selection

philearley:
# Description
### Additional prop
When making network requests using the dropdown-filter-ajax component, we require the ability to pass cookies in the network request.
To allow developers to control this during implementation, we have added a Boolean prop of withCredentials.

### Ability to clear selected value
When a user clears the selected value in the dropdown filter or dropdown filter ajax it should trigger the onchange event to clear the selection.

# Testing Instructions
Testing to be done during an integration with another application.
Due to the way the mock is setup for drop down filter ajax, it is not doing successful network requests on selection.

### DIFF
--- a/change_log/next/add-ability-to-clear-dropdown-filter.yml
+++ b/change_log/next/add-ability-to-clear-dropdown-filter.yml
@@ -1,0 +1,1 @@
+New Features: "DropdownFilter/Ajax improvement: When clearing the dropdown in full, it now clears the selection."
--- a/change_log/next/add-ability-to-clear-dropdown-filter.yml
+++ b/change_log/next/add-ability-to-clear-dropdown-filter.yml
@@ -1,1 +1,1 @@
-New Features: "DropdownFilter/Ajax improvement: When clearing the dropdown in full, it now clears the selection."
+New Features: "DropdownFilter and DropdownFilterAjax improvement: When clearing the dropdown in full, it now clears the selection."
--- a/src/components/dropdown-filter-ajax/__spec__.js
+++ b/src/components/dropdown-filter-ajax/__spec__.js
@@ -291,6 +291,30 @@ describe('DropdownFilterAjax', () => {
         });
       });
     });
+
+    describe('when the param withCredentials is passed', () => {
+      beforeEach(() => {
+        instance = TestUtils.renderIntoDocument(
+          <DropdownFilterAjax
+            name="foo"
+            value="1"
+            path="/foobar"
+            create={ function() {} }
+            withCredentials
+            additionalRequestParams={ {foo: 'bar'} }
+          />
+        );
+      });
+
+      it('calls with credentials', () => {
+        Request.query = jest.fn().mockReturnThis();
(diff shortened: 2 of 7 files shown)


change_log/next/add-ability-to-clear-dropdown-filter.yml
```diff
@@ -0,0 +1 @@
+New Features: "DropdwonFilter/Ajax improvement: When clearing the dropdown in full, it now clears the selection."
```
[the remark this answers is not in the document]
philearley: Corrected and force pushed this change.

change_log/next/add-ability-to-clear-dropdown-filter.yml
```diff
@@ -0,0 +1 @@
+New Features: "DropdownFilter/Ajax improvement: When clearing the dropdown in full, it now clears the selection."
```
toni-leigh: forward slash in the name is out of place

src/components/dropdown-filter-ajax/__spec__.js
```diff
@@ -291,6 +291,30 @@ describe('DropdownFilterAjax', () => {
...
+    describe('when the param withCredentials is passed', () => {
+      beforeEach(() => {
+        instance = TestUtils.renderIntoDocument(
+          <DropdownFilterAjax
+            name="foo"
+            value="1"
+            path="/foobar"
+            create={ function() {} }
+            withCredentials
+            additionalRequestParams={ {foo: 'bar'} }
```
toni-leigh: Spaces inside an object (linting didn't get this?)

`{ foo: 'bar' }`

philearley: Specs appear to be added to the eslint ignore.

change_log/next/add-ability-to-clear-dropdown-filter.yml
philearley: This was done as this impacts `DropdownFilter` and `DropdownFilterAjax`

toni-leigh: I'd write both names out in order to help someone searching for changes to a particular component

src/components/dropdown-filter-ajax/__spec__.js
toni-leigh: Carbon hasn't caught up yet then :-P

I'd still change it for consistency, but it's not a blocker!

src/components/dropdown-filter/__spec__.js
```diff
@@ -185,9 +185,35 @@ describe('DropdownFilter', () => {
         );
         spyOn(instance, 'emitOnChangeCallback');
         TestUtils.Simulate.change(instance._input, {
-          target: { value: 'foo' }
+          target: { value: '' }
```
ianoxley: Just wondering why the value was removed from this spec?

PR #1838, opened 2018-09-11
Auto scroll when using drag and drop

adrianpastur:
# Description

Implementation of automatic scrolling when using drag and drop.

# TODO
- [x] Release notes

# Screenshots / Gifs

sl-438-carbon-demo

# Testing Instructions


src/components/drag-and-drop/draggable-context/draggable-context.js
```diff
@@ -78,13 +86,27 @@ class DraggableContext extends React.Component {
...
+    super(props);
+
+    // Default speed of auto scrolling
+    this.speed = 0;
+
+    // Frame callback ref
+    this.frame = null;
+
+    // Frame callback ref
+    this.dragging = false;
```
andrewjtait: don't think this is used?

src/components/drag-and-drop/draggable-context/draggable-context.js
```diff
@@ -132,11 +207,17 @@ class DraggableContext extends React.Component {
...
+   * Sets state of dragging to true which
+   * will trigger the start of auto scroll
+   * if it is enabled.
+   *
    * @method handleBeginDrag
    * @param {Object} props
    * @return {Void}
    */
   handleBeginDrag = (props) => {
+    this.setState({ dragging: true });
```
andrewjtait: instead of `state.dragging`, does `state.activeIndex` also not determine if drag is in progress or not?

src/components/drag-and-drop/draggable-context/draggable-context.js
```diff
@@ -109,6 +131,59 @@ class DraggableContext extends React.Component {
...
+    if (!this.state.dragging) {
+      return;
+    }
+
+    // constant for the position of the mouse pointer
+    // on the y axis
+    const { clientY } = event;
+
+    // constant for the threshold where the auto scroll begins
+    const threshold = Math.max(140, window.innerHeight / 4);
```
andrewjtait: You should get the window object from the `Browser` helper.

src/components/drag-and-drop/draggable-context/draggable-context.js
```diff
@@ -109,6 +131,59 @@ class DraggableContext extends React.Component {
...
+  }
+
+  tick = () => {
+    if (!this.speed) {
+      this.frame = null;
+      return;
+    }
+
+    window.scrollTo(0, window.scrollY + (this.speed * 10));
+    this.frame = window.requestAnimationFrame(this.tick);
```
andrewjtait: should this be `this.startScrolling()`?

src/components/drag-and-drop/draggable-context/draggable-context.js
andrewjtait: be careful though that index could be `0` which is falsey in JavaScript.

src/components/drag-and-drop/draggable-context/draggable-context.js
adrianpastur: That won't work. I believe that `startScrolling` is only used when the scrolling needs to be initiated when there is no `frame` object. When swapping that line with a call to the startScrolling method, it breaks the whole auto scroll,

src/utils/helpers/scrollable-parent/_spec_.js
```diff
@@ -0,0 +1,23 @@
+// import Browser from '../browser';
```
JanTaras29: @adrianpastur What is to be done about this commented test? I covered this module in draggablecontext, but may also write module specific tests with your permission.

PR #1839, opened 2018-09-11
DropdownFilterAjax: Add the option to pass cookies with ajax call

philearley:
# Description
Add withCredentials prop to DropdownFilterAjax to allow for the passing of cookies in the request.

# Testing Instructions
Testing to be done as part of the integration within an application as the demo site mocks the api call.


philearley:
We have tested that when passing withCredentials as a prop, you see the cookie passed in the request. When withCredentials is not passed or is declared as false, it does not pass the cookie in the request.

PR #1845, opened 2018-09-27
Adding Support for Redux

andrewjtait:
Updates the Flux implementation to pipe the data through Redux. Sets up a ReducerRegistry to dynamically load reducers in to the Redux store.

### DIFF
--- a/src/utils/flux/flux.js
+++ b/src/utils/flux/flux.js
@@ -51,6 +51,7 @@ export function connect(ComposedView, stores) {
 
   const connectedView = reduxConnect(mapStateToProps)(View);
   connectedView.displayName = ComposedView.displayName || ComposedView.name;
+  connectedView._requiresReduxStore = true;
   connectedView._legacyConnect = true;
   return connectedView;
 }
--- a/src/utils/flux/flux.js
+++ b/src/utils/flux/flux.js
@@ -49,5 +49,8 @@ export function connect(ComposedView, stores) {
     return buildPropsFromStores(stores, state);
   }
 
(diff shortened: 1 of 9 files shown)


src/utils/flux/flux.js
```diff
@@ -1,144 +1,57 @@
...
-    _getStoreStates = () => {
-      const states = {};
-
-      for (const key in _stores) {
-        states[key] = _stores[key].getState();
-      }
-
-      return states;
+      const props = buildPropsFromStores(stores, nextProps);
+      this.setState(props);
```
andrewjtait: Could maybe use `getDerivedStateFromProps` instead https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

src/utils/flux/reducer-registry.js
```diff
@@ -0,0 +1,21 @@
+import { createStore, combineReducers } from 'redux'
+
+const reducers = {};
+
+const ReducerRegistry = function() {
+  this.register = (name, reducer) => {
+    reducers[name] = reducer;
```
andrewjtait: need to error handle reducers being registered under the same key

andrewjtait:
This PR will redirect all Flux stores through Redux, however there are some issues.

It breaks all unit tests in an application, without an easy way to manage this it is incredibly difficult to make the switch.

We could let developers switch on a store per store basis, but this will mix up the Flux/Redux implementation too much - particularly if some stores share actions.

It may be easier to not go any further with this change, and instead have developers refactor Flux stores into Redux stores as they go.

PR #1846, opened 2018-09-28
Add in a compile command for docker

toni-leigh:
A quirk of docker containers seems to be that the volume mounting doesn't handle fully deleted directories very well so we need to remove this in order for implanting to work


andrewjtait:
the package scripts have been updated to work better cross environments so hopefully this has been resolved

toni-leigh:
Hopefully, though I have no idea what this problem actually was

PR #1847, opened 2018-10-04
Using composition of components

amaalali:
@andrewjtait This approach that you've taken in `CARBON-1818-alert` can lead to massive simplification of the library. A library like `recompose` might provide additional tooling to make this easier like shown here. 

Alternatively, if you're not looking to add an additional dependency, HOC is just merging the props so this can be a utility that Carbon decides to build. Either way, it takes a more functional approach by using composition.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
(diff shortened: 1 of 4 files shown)


andrewjtait:
👍 thanks @amaalali - this looks good, not sure if we should commit to the additional dependency though. Looking to reduce those at the moment 😄

PR #1849, opened 2018-10-08
remove core-js and promise dep

andrewjtait:
Remove core-js dependency and promise polyfill. If you were using the promise polyfill from Carbon previously we recommend using the polyfill from babel or from core-js manually. carbon-factory v6.1.0 provides a preconfigured `.babelrc` file for this.


andrewjtait:
Yeah I will cut them both together

PR #1857, opened 2018-10-29
Fixed missing SCSS import for Page component

kamilniechajewicz:
# Description
Carbon version 5.0.0 introduced a small regression by removing SCSS import from package.json but not adding it to `Page` component itself, making it miss few CSS rules.

# TODO
- [ ] Release notes

# Screenshots / Gifs

# Testing Instructions
Confirm that the padding required in Journals View is back. This should be enough to ensure that the Page SCSS is now present.


andrewjtait:
screen shot 2018-10-29 at 15 38 20

Confirmed fix in demo site ^ ✅

Issue #1865, opened 2018-11-20
Step Indicator Component


andrewjtait:
There is a requirement to include a new component to visualise a multi step flow.

The following designs are early mockups of what this may look like, but not definitive of what we should end up with:

![screen shot 2018-11-20 at 14 06 43](https://user-images.githubusercontent.com/1668080/48778580-89d9c000-eccd-11e8-9635-c760c58e5cc7.png)
![screen shot 2018-11-20 at 14 05 51](https://user-images.githubusercontent.com/1668080/48778581-89d9c000-eccd-11e8-8dbb-8ade3443ce03.png)

I think for this to work effectively we only need to supply the UI for the indication of what step you are on (rather than an entire component suite that includes `next`/`previous` buttons and a placeholder for the actual step's content). This way it can be used in various contexts if needed, and the steps themselves do not need to follow any particular convention.

**For example:**

The following is an example of state that could be served via Redux, it does not need to follow any particular convention:
```js
steps: [{
  id: 'one',
  state: 'complete'
}], [{
  id: 'two',
  state: 'active'
}, {
  id: 'three',
  state: null
}];
```

The following is an example of how a developer may use the Steps/Step components with the demo state provided above:
```js
<Steps>
  {
    steps.map(step => (
      <Step
        label={ I18n.t(`steps.${step.id}`) }
        status={ step.state }
      />
    )
  }
</Steps>
```

They can reuse the same state to render the correct content, wherever that may live:

```js
const activeStep = findActiveStep(steps);

switch (activeStep.id) {
  case 'one':
    return <StepOne />;
  case 'two':
    return <StepTwo />;
  case 'three':
    return <StepThree />;
}
```

This uses no internal state, and is managed completely by the external state - this allows developers to fully control how and when steps change state etc.

PR #1866, opened 2018-11-20
Initial look at step sequence components

toni-leigh:
Step sequence component wrapped in a wizard, with a demo site page 'play-area' into which it's been added.

There's also some basic state management so it should react to clicks of the buttons

Intending to help with the [issue about step sequences](https://github.com/Sage/carbon/issues/1865)

steps

Load carbon and visit: http://0.0.0.0:8095/play-area to see

### DIFF
--- a/demo/views/pages/play-area/wizard-wrapper-stub/wizard-wrapper-stub.js
+++ b/demo/views/pages/play-area/wizard-wrapper-stub/wizard-wrapper-stub.js
@@ -1,0 +1,65 @@
+import Button from 'components/button';
+import React from 'react';
+import Wizard from 'components/wizard';
+
+class WizardWrapperStub extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      steps: [
+        { label: 'foo', state: 'complete' },
+        { label: 'bar', state: 'current' },
+        { label: 'baz', state: 'incomplete' }
+      ],
+      current: 2
+    };
+  }
+
+  setStepState = current => (step, i) => {
+    let state = 'incomplete';
+    if (current === i) { state = 'current'; }
+    if (current > i) { state = 'complete'; }
+    console.log({ state, label: step.label });
+    return { state, label: step.label };
+  };
+
+  previous = () => {
+    let { current, steps } = this.state;
+
+    current = current === 0 ? 0 : current - 1;
+
+    steps = steps.map(this.setStepState(current));
+
+    this.setState({ current, steps });
+  }
+
+  next = () => {
+    let { current } = this.state;
+    let { steps } = this.state;
+
+    current = current === steps.length - 1 ? current : current + 1;
+
+    steps = steps.map(this.setStepState(current));
+
+    this.setState({ current, steps });
+  }
+
+  render() {
+    return (
+      <div className='demo-wizard-wrapper-stub'>
+        <Wizard { ...this.state }>
+          <div>Step 1</div>
+          <div>Step 2</div>
+          <div>Step 3</div>
+        </Wizard>
+
+        <div className='demo-wizard-wrapper-stub__buttons'>
+          <Button onClick={ this.previous }>Previous</Button>
+          <Button onClick={ this.next }>Next</Button>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default WizardWrapperStub;
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
   "globals": {
     "jest": false,
     "global": false,
-    "process": false
+    "process": false,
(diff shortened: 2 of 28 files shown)


demo/views/pages/play-area/wizard-wrapper-stub/wizard-wrapper-stub.js
```diff
@@ -0,0 +1,65 @@
...
+    this.setState({ current, steps });
+  }
+
+  render() {
+    return (
+      <div className='demo-wizard-wrapper-stub'>
+        <Wizard { ...this.state }>
+          <div>Step 1</div>
+          <div>Step 2</div>
+          <div>Step 3</div>
```
andrewjtait: I don't think this is particularly intuitive, unless you know how `Wizard` behaves it looks like it is rendering all 3 items at once.

I think it is also particularly restricting as the `Wizard` is responsible for deciding what to render and where it is rendered. It is also heavily dictated by complex props, which need to adhere to a certain format shared down with it's children. These children also have to adhere to the correct index, which may cause issues if you ever want to easily add conditional step etc.

If we instead have developers write this implementation detail in their app it makes it more understandable and flexible to different situations, rather than having to maintain a heavy API that will be prone to change and snowball into even more complexity over time. We have definitely been guilty of this in the past, and it is a pattern we need to break.

```js
render() {
  return (
    <div className='demo-wizard-wrapper-stub'>
      <StepIndicators>
        {
          this.state.steps.map(step => (
            <StepIndicator status={ step.state }>{ step.label }</StepIndicator>
          ));
        }
      </StepIndicators>

      { this.state.current === 1 && <div>Step 1</div> }
      { this.state.current === 2 && <div>Step 2</div> }
      { this.state.current === 3 && <div>Step 3</div> }

      <div className='demo-wizard-wrapper-stub__buttons'>
        <Button onClick={ this.previous }>Previous</Button>
        <Button onClick={ this.next }>Next</Button>
      </div>
    </div>
  );
}
```

While this means there is a bit more code on the implementation side, it is easier to understand exactly what is going on, and can be easily manipulated in various ways if needed with no dictation on the data structure. In reality, this would probably be broken down into a few different components - one handling the indicators and another handling the step rendering.

toni-leigh: I put the wizard together quickly to demonstrate the StepSequence (same as StepIndicators) which is basically built like you have above

This is it here: https://github.com/Sage/carbon/pull/1866/files?utf8=%E2%9C%93&diff=unified#diff-f76d855afc9eb13f8df6f2d66db69cf8

toni-leigh: You're suggesting that the Wizard contents are built manually by the dev:

```
<Wizard>
  <StepSequence steps={ steps } />
  { this.state.current === 1 && <div>Step 1</div> }
  { this.state.current === 2 && <div>Step 2</div> }
  { this.state.current === 3 && <div>Step 3</div> }
  <Nav />
</Wizard>
```

rather than

```
<Wizard configuredJSON>
  <div>Step 1</div>
  <div>Step 2</div>
  <div>Step 3</div>
</Wizard>
```

toni-leigh: I wonder about doing both though,

In the first example, the `Wizard` wrapper seems to have no purpose, the user can put together a StepSequence, ButtonPair and their own content and state implementation and I think all I'd need to deliver would be a step sequence

Why not also provide a Wizard that does it for you for the basic cases, then pull it apart if you need to? It could even provide the state for a quick option for simple wizards

andrewjtait: From what I have experienced there is never a basic case with something as complex as this. There is your current implementation, but the next squads implementation will be different in some way so they cannot reuse your `Wizard` component - or they need to add their little adjustments to it and it soon escalates into a mess of configurations. Maybe the confusion here is that the `Wizard` you're putting together might better sit in your app as the view layer rather than in Carbon as a reusable component, but even then I think being a little bit more verbose with the code opens up so much more flexibility and readability for future development. You're hiding the readability of React within a JSON config object which is doing some magic behind the scenes.

There is some merit in delivering reusable compositions of components, and it is something we're discussing with UX, but we haven't worked out what they are exactly yet or whether they are a valid thing for Carbon to own. 

Also in your specific case, are you sure the wizard nav even sits outside of the steps? Are they not buttons inside each step which act as an individual form-per-step?

toni-leigh:
Closing this, the components provided will handle step sequence and step

PR #1867, opened 2018-11-22
Add step sequence component #1865

toni-leigh:
## Description

* Horizontal and vertical versions are provided.
* Flex is used to make this completely flexible. Any size label, parent width or number of items will render correctly. Step indicators are also fully flexible.
* Style config is used to allow both components to have access to the sdame colours without depending on each other or duplicating.
* Inheritence is used to make sure the sequence styles (mostly layout) are handled by the sequence and not by the item, again for fully flexible re-use.

## TODO
- [ ] Release notes
- [ ] Demo site entry

### Stepping Through
animated

### Small
screen shot 2018-11-22 at 10 51 37

### Large
screen shot 2018-11-22 at 10 51 47

### Lots
screen shot 2018-11-22 at 10 53 29

### Unusually large label
screen shot 2018-11-22 at 10 54 01

### Large indicators
screen shot 2018-11-22 at 10 56 19

### Animated with large indicators
animated-ind

## Testing Instructions
* Use demo site


src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,35 @@
...
+  <li className={ classes(state) }>
+    <span className='carbon-step-sequence-item__label'>
+      <span className='carbon-step-sequence-item__indicator'>{ stepMarker(state, indicator) }</span>
+      { label }
+    </span>
+  </li>
+);
+
+StepSequenceItem.propTypes = {
+  label: PropTypes.string,
```
andrewjtait: Instead of `label` can you just used `children`? Just in case they ever want to use something more complicated than a string it makes it a bit more intuitive.

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,35 @@
...
+const baseClass = 'carbon-step-sequence-item';
+const classes = state => `${baseClass} ${baseClass}--${state}`;
+
+const stepMarker = (state, indicator) => {
+  return state === 'complete' ? <Icon type='tick' /> : indicator;
+};
+
+const StepSequenceItem = ({ label, state, indicator }) => (
+  <li className={ classes(state) }>
+    <span className='carbon-step-sequence-item__label'>
```
andrewjtait: is this `span` actually required? looks like it may be superfluous

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,35 @@
...
+StepSequenceItem.propTypes = {
+  label: PropTypes.string,
+  state: PropTypes.string,
+  indicator: PropTypes.string
+};
+
+StepSequenceItem.defaultProps = {
+  label: '',
+  state: 'incomplete',
+  indicator: ''
```
andrewjtait: defaulting to an empty string? not sure I see the point in this?

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,35 @@
...
+    <span className='carbon-step-sequence-item__label'>
+      <span className='carbon-step-sequence-item__indicator'>{ stepMarker(state, indicator) }</span>
+      { label }
+    </span>
+  </li>
+);
+
+StepSequenceItem.propTypes = {
+  label: PropTypes.string,
+  state: PropTypes.string,
```
andrewjtait: you could use the prop type here which describes the 3 valid values? (can't remember the name),

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,35 @@
+import Icon from 'components/icon';
```
andrewjtait: this needs to be a relative path, otherwise it won't work when it is packaged up in a tag

src/components/step-sequence/step-sequence.js
```diff
@@ -0,0 +1,21 @@
...
+
+const StepSequence = ({ children, orientation = 'horizontal' }) => (
+  <ol className={ classes(orientation) }>
+    { children }
+  </ol>
+);
+
+StepSequence.propTypes = {
+  children: PropTypes.node,
+  orientation: PropTypes.string
```
andrewjtait: this could also use the prop type to describe the 2 valid values

src/components/step-sequence/step-sequence.js
```diff
@@ -0,0 +1,21 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import './step-sequence.scss';
+import './step-sequence-orientation.scss';
+
+const baseClass = 'carbon-step-sequence';
+const classes = orientation => `${baseClass} ${baseClass}--${orientation}`;
+
+const StepSequence = ({ children, orientation = 'horizontal' }) => (
```
andrewjtait: for consistency you should probably use `defaultProps`? at first I was confused where the default was coming from as I couldn't find `defaultProps` defined.

src/components/step-sequence/step-sequence-item/step-sequence-item.js
toni-leigh: This is required to bind the two elements together in order to keep them in line when the sequence switches to vertical,

Without it they were vertically oriented. I tried `flex` and `grid` to fix this but it became complex before working so this seemed a better solution

src/components/step-sequence/step-sequence.js
```diff
@@ -0,0 +1,28 @@
...
+import React from 'react';
+
+import './step-sequence.scss';
+import './step-sequence-orientation.scss';
+
+const baseClass = 'carbon-step-sequence';
+const classes = orientation => `${baseClass} ${baseClass}--${orientation}`;
+
+const StepSequence = ({ children, orientation }) => (
+  <ol className={ classes(orientation) }>
```
ianoxley: It might be worth adding an `aria-label` attribute on here, saying something like "step 3 of 5: delivery details".

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,37 @@
...
+
+const baseClass = 'carbon-step-sequence-item';
+const classes = state => `${baseClass} ${baseClass}--${state}`;
+
+const stepMarker = (state, indicator) => {
+  return state === 'complete' ? <Icon type='tick' /> : indicator;
+};
+
+const StepSequenceItem = ({ children, state, indicator }) => (
+  <li className={ classes(state) }>
```
ianoxley: Add `aria-current="step"` to the current step (see https://www.w3.org/TR/wai-aria-1.1/#aria-current)

src/components/step-sequence/step-sequence.js
toni-leigh: How would this work with flexible step labels,

For example, steps might be labeled `a` to `e`.

Would I just use an index and step count for this, regardless of the indicator used by the developer?

ianoxley: Yes, I think that should work 👍

toni-leigh: The label will be hard now that it's not just a string, should I drop this if it's not a string?

ianoxley: Maybe just truncate it to 'step 3 of 5'?

toni-leigh: Decided to put the label in as it's own prop rather than deriving it from other props

ianoxley: Hmm, I'd initially meant for this to go on the `<ol>` element, but adding it to each step might be better for when the step contains complicated, non-string children.

It the steps do contain just strings though you probably won't need the `aria-label`, as you'd be duplicating the content. So these could probably be optional props instead?

Also, have a look at https://www.w3.org/WAI/tutorials/forms/multi-page/#using-step-by-step-indicator - it'd be good if we can add visually hidden 'completed' and 'current' indicators too.

src/components/step-sequence/step-sequence.js
```diff
@@ -0,0 +1,28 @@
...
+    'horizontal',
+    'vertical'
+  ])
+};
+
+StepSequence.defaultProps = {
+  orientation: 'horizontal'
+};
+
+export default StepSequence;
```
andrewjtait: Just thought, it'd be very useful if this exported both the `StepSequence` and the `StepSequenceIndicator` components so you could do this instead:

```
import { StepSequence, StepSequenceIndicator } from 'carbon-react/lib/components/step-sequence';
```

toni-leigh: Yep I just noticed it while using it!

That and the aria changes are the last things to do here

src/components/step-sequence/step-sequence-item/step-sequence-item.js
```diff
@@ -0,0 +1,77 @@
...
+  ariaLabel,
+  children,
+  indicator,
+  status,
+  hiddenCompleteLabel,
+  hiddenCurrentLabel
+}) => (
+  <li
+    className={ classes(status) }
+    { ...ariaLabelProp(ariaLabel) }
```
andrewjtait: Maybe we should just let developers add any additional props they want rather than adding specific support, otherwise we need to support/maintain these props. Eg.

```
const StepSequenceItem = ({
  children,
  indicator,
  status,
  hiddenCompleteLabel,
  hiddenCurrentLabel,
  ...props
}) => (
  <li
    { ...props }
    className={ classes(status) }
    { ...ariaRoleProp(status) }
  >
```

PR #1873, opened 2018-12-03
Extending Dropdown component to allow auto completion with tab key press

gedskikas:
# Description
Extending Dropdown component to allow users auto completion of selected items within the dropdown with the press of a tab button

# TODO
- [x] Release notes

# Testing Instructions
Test Dropdown/Dropdown Filter/ Dropdown Filter Ajax components to see if the tab key autocompletes selected items.


toni-leigh:
Proprietary app screen shots shouldn't be in the open source repo

gedskikas:
@toni-leigh removed the screenshot

toni-leigh:
You should be able to demonstrate the change in the carbon demo site, which you can access by running `npm start` in the carbon root then visiting http:/0.0.0.0:8095

In fact, the component should be QA'd in this context to avoid the situation where the change works in your context and not in the general one. Demo site screenshots are best for PRs.

src/components/dropdown/dropdown.js
```diff
@@ -382,7 +382,7 @@ const Dropdown = Input(InputIcon(InputLabel(InputValidation(class Dropdown exten
     let nextVal;
 
     switch (ev.which) {
-      case 13: // return
+      case 9: case 13: // tab key or return key
```
kamilniechajewicz: 💄 `case`'s should be on separate lines always for readability
but what concerns me more is that we're preventing default behaviour of "tab" which is going to the next element - we're basically forcing people to press tab twice per dropdown, which makes it less smooth experience than just tabbing straight away to the next - IMO in case of tab default action should not be prevented, I quickly checked it and not doing it gives the experience I'd expect from tabbing (needs investigation of possible side effects but not sure there will be any)

shauncox1:
Passed manual Q/A ✅see: https://github.com/Sage/s1_gac_ui/pull/704 for links to test sessions as part of this work

PR #1898, opened 2019-01-24
Fe 1197 adapt scrollable list component to accept different types of children

samhinton88:
# Description
Enhances a previous version of the SelectableList component to support non-selectable items.

The list may be used with the `SelectableItemComponent` for out of the box styling and select functionality.

Devs may pass their custom components to the list. These custom components will work with the keyboard / mouse select functionality if passed the `isSelectable` prop, else they will be rendered but will not respond 

Adds a local test-utils folder which a number of functions consumed by the spec in an effort to make unit tests more readable. The utils are implemented with heavy use of functional composition.

# TODO
- [] Release notes

# Screenshots / Gifs
scrollable-list-mkii


src/components/scrollable-list/scrollable-list.component.js
```diff
@@ -0,0 +1,175 @@
...
+  buildHeight = (acc, { offsetHeight }) => acc + offsetHeight
+
+  handleScroll = ({ target: { scrollTop, scrollHeight } }) => {
+    if (!this.props.onLazyLoad) return;
+
+    if ((scrollHeight - scrollTop) < 200) this.props.onLazyLoad();
+  }
+
+  handleKeyDown = (e) => {
+    e.preventDefault();
```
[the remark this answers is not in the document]
andrewjtait: tried this in the browser and it does block all key events!

src/components/scrollable-list/test-utils.js
```diff
@@ -0,0 +1,90 @@
...
+  selectedItemOf,
+  childrenFrom,
+  renderListItems,
+  keyboard,
+  assertKeyboardTraversal,
+  assertHoverTraversal,
+  traverseListObject,
+  listFrom,
+  click
+};
```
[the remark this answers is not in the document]
samhinton88: Agreed that helpers like this should be shared across the library. The intention with this was to offer some suggestions as to how we can clean up our approach to writing these unit tests, rather than a finished solution. 

I argue that we should try to improve our approach in making unit tests more thorough and more readable. Declaring more at the top of test blocks and keeping individual `it` blocks as thin as possible, avoiding wherever possible testing instance methods on the components directly, etc. 

Also, I'm finding a lot of instances where we're having to alter the code in our components to work around the limitations of the Jest API and JSDom.

package.json
```diff
@@ -46,6 +46,7 @@
     "enzyme-adapter-react-16": "1.7.1",
     "flux": "^3.1.1",
     "highlight.js": "~9.6.0",
+    "jest-styled-components": "^6.3.1",
```
andrewjtait: package-lock looks like it has upgraded a lot of stuff - this may need re-generating to be safe

src/components/scrollable-list/scrollable-list-item.component.js
```diff
@@ -0,0 +1,20 @@
+import styled, { css } from 'styled-components';
+import propTypes from 'prop-types';
+
+const StyledListItem = styled.li`
```
andrewjtait: As there is no distinction in this case between the component and the style, do you think the name of this should reflect the component name so it makes sense in dev tools etc? eg. `ScrollableListItem`.

You could achieve this be either renaming this `const` or adding `StyledListItem.displayName = 'ScrollableListItem';`

src/components/scrollable-list/scrollable-list-item.component.js
```diff
@@ -0,0 +1,20 @@
...
+  width: 100%;
+  padding: 5px 6px;
+  ${({ isSelected }) => isSelected && css`background-color: #E6E9E7;`}
```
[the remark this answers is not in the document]
andrewjtait: still curious about this one

src/components/scrollable-list/scrollable-list-item.component.js
```diff
@@ -0,0 +1,20 @@
...
+  height: 15px;
+`;
+
+StyledListItem.propTypes = {
+  id: propTypes.number,
+  isSelected: propTypes.bool
+};
+
+StyledListItem.defaultProps = {
+  isSelectable: true
```
andrewjtait: I would perhaps add a comment next to this one to describe why it is required, as it wasn't clear to me before I figured out the API you're using. eg. `// defaulted to true so it integrates with ScrollableList by default`

src/components/scrollable-list/scrollable-list.component.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  scrollBox = React.createRef();
+
+  componentDidMount() {
+    if (this.props.keyNavigation) {
+      document.addEventListener('keydown', this.handleKeyDown);
+    }
+    const initialSelectedItem = this.nextSelectable('down', -1);
```
andrewjtait: does this mean the 1st item is automatically highlighted? it might be better to make nothing highlighted by default (eg. `-1`?) - if we do this then you can remove this and just add it to the initial state object on line 20.

src/components/scrollable-list/scrollable-list.component.js
```diff
@@ -0,0 +1,175 @@
...
+      // set the bottom of the scroll box to the bottom of the selected item
+      list.scrollTop = this.setScrollTop(item);
+    } else {
+      list.scrollTop = 0;
+    }
+  }
+
+  setScrollTop = (item) => {
+    const { current: list } = this.scrollBox,
+        { offsetHeight: listHeight, children } = list,
```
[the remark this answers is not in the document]
andrewjtait: still applicable?

src/components/scrollable-list/scrollable-list.component.js
```diff
@@ -0,0 +1,175 @@
...
+        { offsetHeight: itemHeight } = children[item];
+
+    // total height of list up to selected item
+    const scrollPos = [...children].slice(0, item).reduce(this.buildHeight, 0);
+
+    return scrollPos - listHeight + itemHeight;
+  }
+
+  buildHeight = (acc, { offsetHeight }) => acc + offsetHeight
```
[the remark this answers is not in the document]
andrewjtait: still applicable

src/components/scrollable-list/scrollable-list.component.js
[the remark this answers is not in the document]
andrewjtait: still applicable

src/components/scrollable-list/scrollable-item-wrapper.component.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import ScrollableListContext from './scrollable-list.context';
+
+const ScrollableItemWrapper = (CaptiveComponent, index, isSelected) => (
```
andrewjtait: As this isn't actually a component it might make sense to rename the file and const to something more applicable and less confusing

src/components/scrollable-list/scrollable-list-item.component.js
samhinton88: The `css` object is used when you need to add  a whole line of css, ie. the attr and value. In this instance I do not give a `background-color` at all unless there is a prop, so I use the css object add the line. If you prefer I can just set the background to  `transparent` by default?

src/components/scrollable-list/scrollable-list.component.js
samhinton88: I presumed that if, for example, the user types an `ab`, it might be more expected that the closest or top match (i.e. `abacus`) would be immediately selected for use?

src/components/scrollable-list/scrollable-list-item.component.js
andrewjtait: I think it'll work without using it, but I found this thread explaining why you should use it regardless https://github.com/styled-components/styled-components/issues/1178#issuecomment-332211886

andrewjtait: I think thats true as they type, but not on open - see https://carbon.sage.com/components/dropdown-filter

It may be that the select/option components will dictate this. Leave it for now and we will see how it works when we plug it all together 👍

src/components/scrollable-list/scrollable-list.component.js
andrewjtait: I think thats true as they type, but not on open - see https://carbon.sage.com/components/dropdown-filter

It may be that the select/option components will dictate this. Leave it for now and we will see how it works when we plug it all together 👍

src/components/scrollable-list/scrollable-item-wrapper.component.js
samhinton88: What do you think might be more applicable and less confusing? Something more imperative like `makeSelectable`?

PR #1905, opened 2019-01-30
select and option components

andrewjtait:
We have added a new Select and Option component. These are still classed as experimental while full Dropdown support is provided.


src/__experimental__/components/select/select-list.component.js
```diff
@@ -0,0 +1,46 @@
...
+    const left = `${inputBoundingRect.left}px`;
+    this._list.current.setAttribute('style', `left: ${left}; top: ${top}; width: ${width}; position: absolute;`);
+  }
+
+  render() {
+    if (!this.props.open) return null;
+
+    return (
+      <Portal onReposition={ this.positionList }>
+        <div ref={ this._list }>
```
ianoxley: We should probably use a `<ul>` or `<ol>`, and add `role="listbox"`.

src/__experimental__/components/select/option.component.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Option = ({ text, children, ...props }) => (
+  <div { ...props }>
```
ianoxley: If `<Select>` uses a list tag then this should be an `<li>`. It should also have `role="option"` too to help with accessibility.

src/utils/filter-children/filter-children.util.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+const renderChild = (child, callback) => {
+  if (callback) return callback(child);
+  return child;
+};
+
+const defaultFilter = (processedText, processedValue, filterType) => {
```
samhinton88: It might be better to change the signature of this so the `defaultFilter` is just a plain function. You then don't need to pass in both `filterType` and `filter` into `filterChildren`.

src/utils/filter-children/filter-children.util.js
```diff
@@ -0,0 +1,28 @@
...
+  return processedText[filterType](processedValue);
+};
+
+const filterChildren = ({
```
samhinton88: Some documentation about how to use this would be good.

andrewjtait: done 👍

src/utils/filter-children/filter-children.util.js
andrewjtait: done 👍

sianford:
Tested in the Carbon Demo site sandbox. Component behaves as expected in Chrome, Safari and Firefox browsers. Minor issue remaining in IE11 and Edge whereby the Shift + Tab focus order is not correct but this will be resolved in a separate ticket. Testing passed.

src/__experimental__/components/select/select.spec.js
```diff
@@ -0,0 +1,260 @@
...
+  }, {
+    value: '2',
+    text: 'Blue'
+  }, {
+    value: '3',
+    text: 'Red'
+  }];
+
+  // utility functions to fetch various elements from the wrapper
+  const fetchList = wrapper => wrapper.find('SelectList');
```
samhinton88: Consider using the naming convention `[propName]Of = componentName => {}` so that in the tests you get that nice `expect(pillsOf(component)).toBe(whatever)`.

andrewjtait: done 👍

src/__experimental__/components/select/select.spec.js
```diff
@@ -0,0 +1,260 @@
...
+
+  describe('filter', () => {
+    it('updates the filter value when the textbox value is updated', () => {
+      const wrapper = openList(renderWrapper());
+      expect(listOf(wrapper).props().filterValue).toEqual(undefined);
+      textboxOf(wrapper).find('input').simulate('change', { target: { value: 'x' } });
+      expect(listOf(wrapper).props().filterValue).toEqual('x');
+    });
+
+    it('triggers custom onFilter event if one if passed', () => {
```
edleeks87: typo

PR #1906, opened 2019-01-31
Fe 1183 setup js framework for js color themes

samhinton88:
# Description

Adds a set of utility functions which create a palette of colours, with which a developer can create themes.
```
import { generatePalette, addOpacity } from 'carbon-react/lib/style';

const palette = generatePalette(baseColors);

const myTheme = {
  primary: palette.productBlueTint(20),
  primaryShadow: addOpacity(palette.productBlueTint(20)), 0.2),
  disabledText: palette.slateTint(50),
  modalFade: addOpacity(palette.black, 0.6)
};
```

The palette exposes functions which add tint or shade to a given base colour. So, given colour config of the value: 
```
const palette = generatePalette({ brilliantGreen: '00DC00' })
```
The developer will be able to call: 
```
palette.brilliantGreenTint(n);
palette.brilliantGreenShade(n);
```
Where n is the degree of white (in case of `tint`) or black (in case of `shade`) they wish to mix into the base colour.

The mix function replicates the Sass `mix` mixin.

Base colours are also returned with the palette, allowing `palette.brilliantGreen`.

`addOpacity` is a utility which converts hex colour values to rgb and then adds a given level of opacity.
```
addOpacity('006e3a', 0.2) // => rgba(0,110,58,0.2)
```


src/style/utils/mix.js
```diff
@@ -0,0 +1,19 @@
+const decToHex = decimal => decimal.toString(16);
+const hexToDec = hex => parseInt(hex, 16);
+const hasHashSym = color => color[0] === '#';
+
+export default (inputColorA, inputColorB, weight = 50) => {
+  const colorA = hasHashSym(inputColorA) ? inputColorA.slice(1) : inputColorA,
+      colorB = hasHashSym(inputColorB) ? inputColorB.slice(1) : inputColorB;
```
grabkowski: why not just remove that character?
colorA = inputColorA.replace(/#/,'');

src/style/__spec__.js
```diff
@@ -0,0 +1,111 @@
...
+
+    it('returns base colors', () => {
+      const baseNames = Object.keys(config);
+      const paletteNames = Object.keys(palette);
+      baseNames.forEach(name => expect(paletteNames.includes(name)).toBeTruthy());
+    });
+
+    it('caches function calls', () => {
+      palette.productBlueShade(20);
+      palette.productBlueShade(20);
```
andrewjtait: you're not asserting anything here? is it just to pass coverage?

src/style/palette/index.js
```diff
@@ -0,0 +1,28 @@
...
+import shade from '../utils/shade';
+
+const cachedFunc = cb => (cache = {}) => (weight) => {
+  if (cache[weight]) { return cache[weight]; }
+  const color = cb(weight);
+  cache[weight] = color;
+  return color;
+};
+
+const palette = (config) => {
```
andrewjtait: some documentation on this would be useful - just so other developers can easily see what it is for and how to use it

src/style/utils/at_opacity.js
```diff
@@ -0,0 +1,9 @@
+export default (base) => {
+  const parse = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(base);
+
+  const r = parseInt(parse[1], 16),
+      g = parseInt(parse[2], 16),
+      b = parseInt(parse[3], 16);
+
+  return opactiy => `rgba(${r},${g},${b},${opactiy})`;
```
andrewjtait: typo on `opactiy`

andrewjtait: documentation on this would also be good please

change_log/next/FE-1183.yml
```diff
@@ -0,0 +1 @@
+New Features: "Added functions to create a palette of colors."
```
andrewjtait: Can you add a bit more detail here - eg. what functions and how to use them. Would like to get us better at producing enough detail in release notes.

andrewjtait: You can use `|` to add multi line.

andrewjtait:
moved my review to https://github.com/Sage/carbon/pull/1911

PR #1907, opened 2019-01-31
Add height styling to carousel to fix no scroll issue

alex-buchan:


src/components/carousel/carousel.js
```diff
@@ -506,6 +506,7 @@ class Carousel extends React.Component {
 
           <CSSTransitionGroup
             component='div'
+            style={ { height: '100%' } }
```
txusko: Why don't you use a CSS class instead of an inline style property?

alex-buchan: The only reason was that it's a small correction to the styling, I didn't think it needed a css class, unless you think it's necessary.

txusko: If I'm not wrong, the general case is to use a CSS class instead of an inline style (Inline styles should be used only to add dynamic content ¿?).
I really don't care if you use one or the other, I just wonder in which cases we can use the inline style.

mob1970: IMO it should go into `/src/components/carousel/carousel.scss`

change_log/next/reports_no_scroll_issue.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "Fixes AC-3061, where the Balance Sheet and Profit and Loss pages would not scroll when opening the full screen dialog"
```
mob1970: I think it shouldn't include any reference to any ticket (don't forget it's an open source project).

src/components/carousel/carousel.scss
```diff
@@ -202,3 +202,7 @@ $cubic-bezier: .34,1.10,.69,1.01;
   opacity: 0;
   transition: all $slide-animation-length cubic-bezier($cubic-bezier);
 }
+
+.css-transition-group {
```
patrickmelvin: `.carousel-transition-group-full-height` to avoid a collision.

alex-buchan: Oks

andrewjtait: I would go with a class name `carbon-carousel__transition` please, so it is consistent with the rest of the component

change_log/next/reports_no_scroll_issue.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "Fixes the Balance Sheet and Profit and Loss pages would not scroll when opening the full screen dialog"
```
andrewjtait: This should be more generic to the bug fix in the component, rather than what you're fixing in the app. We shouldn't really mention in app bugs in this open source project :)

change_log/next/reports_no_scroll_issue.yml
andrewjtait: @mob1970 @alex-buchan it is ok to reference the Jira ticket number if it helps, that'll not mean anything to anyone without access to Jira. But you shouldn't provide a link to our Jira app, and you shouldn't mentioned anything product related, for example what is broken in the app etc, and you shouldn't include screenshots of the bug in the app. Keep it focussed on the component, and include screenshots from the demo site etc.

andrewjtait: But its also how your phrase it... saying "This fixes AC-3061" will be confusing to anyone external. But "AC-3061 - This fixes an issue with..." it is just adding a reference to the release note which is fine and won't be confusing.

PR #1908, opened 2019-01-31
Fe 1182 setup carbon for typescript

edleeks87:
# Description
Begins the process of integrating TypeScript support into Carbon.
Converts the Button and Link components to .tsx files and removes propTypes in favour of interfaces.
As a result a new function is required to filter valid props in ether.js, this affects the demo site so a new solution needs to be explored for that, hence why this branch is marked "DO NOT MERGE".
Uses an upgraded branch of Carbon-Factory (see related PRs below) to handle all the new Babel and TypeScript plugins, this will be updated once that branch has passed review. 

# Related Pull Requests
https://github.com/Sage/carbon-factory/pull/222/files#diff-0a7c10f331bf0ba9ca651dc99e6be833


src/components/button/button.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+export interface ButtonProps {
+  as?: 'primary' | 'secondary';
+  disabled?: boolean;
+  theme?: string;
+  size?: string;
+  subtext?: any;
+}
+declare const Button: React.Component<ButtonProps, {}>;
+export default Button;
```
grabkowski: shouldn't this file be included in the ```lib``` folder instead of the ```component```?

src/components/button/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './button';
+export * from './button';
```
andrewjtait: Is there any difference between these lines? do we need them both?

src/components/button/index.js
```diff
@@ -0,0 +1,3 @@
+import Button from './button';
+
+export default Button;
```
andrewjtait: Quite like the single line syntax:

```
export { default } from './button';
```

package.json
```diff
@@ -15,12 +15,14 @@
     "test": "jest --config=./jest.conf.json",
     "debug": "node --inspect ./node_modules/.bin/jest --watch --config=./jest.conf.json",
     "lint": "./node_modules/.bin/eslint",
-    "precompile": "npm run clean-lib && npm run copy-files && npm run babel",
+    "lint-ts": "./node_modules/.bin/tslint -c tslint.json 'src/components/**/*.ts'",
+    "precompile": "npm run clean-lib && npm run copy-files && npm run typescript && npm run babel",
```
andrewjtait: Do we still need `npm run typescript` as part of the precompilation process? Should it just be part of the CI build?

package.json
```diff
@@ -15,12 +15,14 @@
     "test": "jest --config=./jest.conf.json",
     "debug": "node --inspect ./node_modules/.bin/jest --watch --config=./jest.conf.json",
     "lint": "./node_modules/.bin/eslint",
-    "precompile": "npm run clean-lib && npm run copy-files && npm run babel",
+    "lint-ts": "./node_modules/.bin/tslint -c tslint.json 'src/components/**/*.ts'",
```
andrewjtait: doesn't need to be scoped to `/src/components` - we should lint the entire `/src` directory?

src/components/button/button.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+export interface ButtonProps {
+  as?: 'primary' | 'secondary';
+  disabled?: boolean;
+  theme?: string;
```
andrewjtait: theme has set values that it can be too, can we add these here?

src/components/button/button.d.ts
```diff
@@ -0,0 +1,10 @@
...
+  size?: string;
```
andrewjtait: I think size also has supported set values.

src/components/link/__spec__.js
```diff
@@ -5,7 +5,7 @@ import TestUtils from 'react-dom/test-utils';
 import { elementsTagTest, rootTagTest } from '../../utils/helpers/tags/tags-specs';
 
 import Icon from './../icon';
-import Link from './link';
+import Link from './index';
```
andrewjtait: required change or can be reverted?

src/components/link/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './link';
+export * from './link';
```
andrewjtait: Do you need both here?

src/components/link/link.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+export interface LinkProps {
+  className?: string;
+  disabled?: boolean;
+  href?: string;
+  icon?: string;
+  iconAlign?: string;
```
andrewjtait: are these set values?

src/components/link/link.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  onClick?: (...args: any[]) => any;
+  onKeyDown?: (...args: any[]) => any;
+  tabbable?: boolean;
+  to?: string;
+  tooltipMessage?: string;
+  tooltipPosition?: string;
+  tooltipAlign?: string;
```
andrewjtait: are position and align set values?

tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./node_modules/carbon-factory/tsconfig.json",
+  "include": [
+    "src"
```
andrewjtait: does having this config mean you need to define the path at all in the script command?

blaky:
Hi, could you let us know when we can expect this PR merged to master? (and hopefully a new NPM release with it)

edleeks87:
@blaky we should be cutting a new version by the beginning of next week

src/components/button/button.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+export interface ButtonProps {
+  as?: 'primary' | 'secondary';
+  disabled?: boolean;
+  theme?: 'blue' | 'grey' | 'magenta' | 'magenta-dull' | 'red' | 'white';
+  size?: 'small' | 'medium' | 'large';
+  subtext?: any;
```
customcodepl: I wonder if `any` type is a good practice.

src/components/button/button.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+export interface ButtonProps {
```
customcodepl: Did you consider use types instead of interfaces?

src/components/button/button.d.ts
edleeks87: Yeah I've missed that one, I'm going to set it to string as that's what it is in the propTypes

src/components/button/button.d.ts
edleeks87: I have, my understanding is that either works fine as long as we're consistent throughout Carbon but that you can't merge declarations when using `type` and I can see scenarios where a developer may want to extend the definitions.  I'll talk to Andrew about it when he gets in as I chose to go with `interface` to allow for this, but will change it to `type` if it's not needed.

edleeks87: I've spoke to him, we're going to leave it as interface for now so that anyone using Carbon can extend the definition if they need to

customcodepl: I understand.
I've found it https://stackoverflow.com/questions/41385059/possible-to-extend-types-in-typescript but it's OK for me. As you said `as long as we're consistent`. :)

PR #1911, opened 2019-02-08
Fe-1219 set up style themes

samhinton88:
# Description
Adds first implementation of themes.

Each of `small`, `medium`, `large`,  and `classic`, consume a palette object and build on a base theme which contains attributes shared across all.

Currently the base follows a schema with a top level key of `colors`.

```
  const { atOpacity } = palette,
      blackWithOpacity = atOpacity('#000000');

{
      colors: {
        white: '#FFFFFF',
        error: palette.errorRed,
        warning: palette.gold,
        sageLogo: palette.brilliantGreen,
        success: palette.brilliantGreenShade(20),
        info: palette.productBlueShade(3),
        text: {
          body: blackWithOpacity(0.9),
          disabled: blackWithOpacity(0.55),
          placeholder: blackWithOpacity(0.3)
        }
      }
    }
```


src/style/__spec__.js
```diff
@@ -0,0 +1,111 @@
+import config from './color-config';
+import mix from './utils/mix';
+import { addOpacity, generatePalette } from './'
+
+const assertCorrectColorMix = (config, paletteObject) => {
+  Object.keys(config).forEach(col => {
+    const match = /([A-Za-z]+)([\d]?[\d]?)/.exec(col);
```
grabkowski: that expression could be simplified to:
(/([a-z]+)([\d]{0,2})/i)

src/style/palette/index.js
```diff
@@ -0,0 +1,28 @@
...
+    const tintBy = tint(config[baseName]),
+        shadeBy = shade(config[baseName]);
+
+    acc[`${baseName}Tint`] = cachedFunc(tintBy)();
+    acc[`${baseName}Shade`] = cachedFunc(shadeBy)();
+
+    return acc;
+  }, {});
+
+  return { ...funcs, ...config };
```
grabkowski: are that config entries needed in a palette object? Config could be imported directly from ```color-config```.

src/style/palette/index.js
```diff
@@ -0,0 +1,28 @@
...
+import shade from '../utils/shade';
+
+const cachedFunc = cb => (cache = {}) => (weight) => {
+  if (cache[weight]) { return cache[weight]; }
+  const color = cb(weight);
+  cache[weight] = color;
+  return color;
+};
+
+const palette = (config) => {
```
grabkowski: that name is inconsistent with what is imported in other files, shouldn't it be ```generatePalette```?

src/style/utils/at_opacity.js
```diff
@@ -0,0 +1,9 @@
+export default (base) => {
+  const parse = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(base);
```
grabkowski: this expression could be shortened to:
```(/[^#]{2}/g)```
```match``` could be used to split that color into an array with three hex digit pairs

File: src/style/utils/at_opacity.js [hunk 5]
```diff
@@ -0,0 +1,9 @@
...
+
+  const r = parseInt(parse[1], 16),
+      g = parseInt(parse[2], 16),
+      b = parseInt(parse[3], 16);
```
grabkowski: we could use ```map``` here:
```const rgb = hexPairs.map((hexPair)=> parseInt(hexPair, 16))```

File: src/style/utils/at_opacity.js [hunk 6]
```diff
@@ -0,0 +1,9 @@
...
+
+  return opactiy => `rgba(${r},${g},${b},${opactiy})`;
```
grabkowski: if we use map, then we can just write ```rgba(${rgb.join(',')},${opactiy})```

src/style/utils/merge_deep.spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+  it('merges the objects and creates a new key', () => {
+    const newKeyObj = { a: 1, b: { nestedB: 1 }, c: 'test' };
+    expect(mergeDeep(base, newKeyObj)).toEqual({ a: 1, b: { nestedB: 1 }, c: 'test' });
+  });
+
+  it('merges the objects but does not create a new key when not passed an object', () => {
+    expect(mergeDeep(base, undefined)).toEqual(base);
+  });
+});
```
grabkowski: I think that we need a test to check if arrays are merged properly

change_log/next/FE-1183.yml
```diff
@@ -0,0 +1 @@
+New Features: "Added functions to create a palette of colors."
```
andrewjtait: Can you add a bit more detail here - eg. what functions and how to use them. Would like to get us better at producing enough detail in release notes.

You can use `|` to add multi line.

change_log/next/FE-1219.yml
```diff
@@ -0,0 +1 @@
+New Features: "Added basic implementation of themes"
```
andrewjtait: Can you add a bit more detail here - eg. what functions and how to use them. Would like to get us better at producing enough detail in release notes.

src/style/__spec__.js
```diff
@@ -0,0 +1,111 @@
...
+
+    it('returns base colors', () => {
+      const baseNames = Object.keys(config);
+      const paletteNames = Object.keys(palette);
+      baseNames.forEach(name => expect(paletteNames.includes(name)).toBeTruthy());
+    });
+
+    it('caches function calls', () => {
+      palette.productBlueShade(20);
+      palette.productBlueShade(20);
```
andrewjtait: you're not asserting anything here? is it just to pass coverage?

src/style/palette/index.js
andrewjtait: also some documentation on this would be useful - just so other developers can easily see what it is for and how to use it

src/style/utils/at_opacity.js
andrewjtait: typo on `opactiy`

andrewjtait: documentation on this would also be good please

src/style/themes/base/base_theme.config.js
```diff
@@ -0,0 +1,22 @@
...
+        info: palette.productBlueShade(3),
+        text: {
+          body: blackWithOpacity(0.9),
+          disabled: blackWithOpacity(0.55),
+          placeholder: blackWithOpacity(0.3)
+        }
+      }
+    }
+  );
+};
```
andrewjtait: can you keep file names consistent with dashes rather than the odd underscore please?

src/style/themes/base/index.js
```diff
@@ -0,0 +1,17 @@
...
+import generatePalette from '../../palette';
+
+const palette = {
+  ...generatePalette(baseColors),
+  atOpacity
+};
+
+export default (configureTheme) => {
+  const baseTheme = configureBase(palette);
+  const themeToMergWithBase = configureTheme(palette);
```
andrewjtait: typo `themeToMergWithBase`

src/style/themes/base/index.js
```diff
@@ -0,0 +1,17 @@
...
+import atOpacity from '../../utils/at_opacity';
+import { mergeDeep } from '../../utils/merge_deep';
+import generatePalette from '../../palette';
+
+const palette = {
+  ...generatePalette(baseColors),
+  atOpacity
+};
+
+export default (configureTheme) => {
```
andrewjtait: can you separate the merge function from the theme? so you export the base theme here, and you export a merge function separately? Just thinking someone may want to use the base theme, or merge different themes together etc

src/style/palette/index.js
samhinton88: The palette needs to pass the base colours along with the functions, it seemed tidier to draw them into the palette here, otherwise they will have to be included into each individual theme file.

src/style/themes/themes.spec.js
```diff
@@ -0,0 +1,158 @@
...
+          primary: palette.productGreenShade(21),
+          secondary: palette.productGreenShade(41),
+          tertiary: palette.productGreenShade(61)
+        }
+      };
+
+      assertIsSubset(classicThemeConfig, classicTheme);
+    });
+  });
+});
```
andrewjtait: I would break your specs up, you can unit test each module individually and keeps your tests more concise.

src/style/themes/themes.spec.js
```diff
@@ -0,0 +1,158 @@
...
+  const comparisonKeys = Object.keys(comparison);
+
+  objKeys.forEach((key) => {
+    // assert that keys are present
+    expect(comparisonKeys.includes(key)).toBeTruthy();
+
+    // repeat for nested objects
+    assertIsSubset(obj[key], comparison[key]);
+  });
+};
```
andrewjtait: These tests don't test any scenario where keys are overridden - only that new keys are added.

src/style/palette/index.js
grabkowski: having both methods and entries in one object could be confusing, when their names are nouns.
I think that renaming these methods to verbs would be more meaningful.

src/style/themes/themes.spec.js
samhinton88: There is a spec on the merge-deep util that does this. Should I repeat this here?

andrewjtait: I didn't see that, thats fine - except the merge-deep spec doesn't test values are overridden either

samhinton88: I've since made changes so that it does.

src/style/palette/index.js
samhinton88: `tint` and `shade` are verbs, so to me the naming works. Agreed that it'd probably better to separate out the base colours, so I'll do that.

src/style/__spec__.js
samhinton88: Not entirely sure how to mock the caching functionality using Jest.

src/style/themes/base/index.js
samhinton88: Who needs `e`s anyway. Dull letter.

src/style/utils/merge-deep.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    expect(mergeDeep(base, newKeyObj)).toEqual({ a: 1, b: { nestedB: 1 }, c: 'test' });
+  });
+
+  it('merges objects and overwrites keys if they exist', () => {
+    expect(mergeDeep(base, objectToMergeWithSameKey)).toEqual({ a: 2, b: { nestedB: 1 } });
+  });
+
+  it('handles the merging of objects which contain arrays', () => {
+    expect(mergeDeep(base, objectWithArray)).toEqual({ a: 1, b: { nestedB: 1 }, c: ['inArray'] });
+  });
```
grabkowski: I think that we need a test that checks if arrays are merged properly - e.g.: mergeDeep({ a: [1,2] }, { a: [3] }) should result in { a: [1,2,3] }

src/style/utils/mix.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A Javascript implementation of the Sass `mix` mixin.
+ */
+export default (inputColorA, inputColorB, weight = 50) => {
+  let colorA = inputColorA.replace(/#/, ''),
+      colorB = inputColorB.replace(/#/, '');
+
+  colorA = isShortHex(colorA) ? colorA.repeat(2) : colorA;
+  colorB = isShortHex(colorB) ? colorB.repeat(2) : colorB;
```
grabkowski: that should be redundant if you use ```/[^#]{2}/g``` regExp

samhinton88: That regex will fail on hex codes with only three digits, however. I need to check the length of the string first (having before that stripped it of any #), repeat the string if necessary, and then find the pairs.

src/style/utils/merge-deep.spec.js
samhinton88: Do we want to allow this behaviour? All the other values are overwritten when they have the same key. i.e. if `mergeDeep({ a: 1 }, { a: 2 }) => { a: 2 }`, should it not follow that `mergeDeep({ a: [1] }, { a: [2] }) => { a: [2] }`.

grabkowski: You are right, I'll leave that as resolved.
I was using https://github.com/TehShrike/deepmerge, which does that by default and that behaviour is configurable , but on the other side lodash's ```_.merge({ a: [1,2] }, { a: [3] })``` will result in ```{ a: [3, 2] }```.

src/style/utils/mix.js
grabkowski: my bad, I haven't looked closely

on the other hand it could be extracted to another method to not repeat code and increase readability

PR #1913, opened 2019-02-12
FE-1241 new validator function

edleeks87:
# Description
Adds a new Promise based validation function, it allows for either an array or single validator to be passed to it and run on a value. The function also supports the legacy validators already existing in Carbon.


src/utils/validations/validator/validator.js
```diff
@@ -0,0 +1,36 @@
...
+        if (isLegacy(func)) {
+          const valid = func.validate(value, props);
+          if (valid) {
+            return Promise.resolve();
+          }
+          return Promise.reject(func.message());
+        }
+        return func(value, props);
+      });
+    } else if (isLegacy(validationFunctions)) {
```
samhinton88: You could transform `validationFunctions` into an array at the top of the function if needed, rather than need to introduce this second branch.

src/utils/validations/validator/validator.js
```diff
@@ -0,0 +1,36 @@
...
+      results = validationFunctions.map((func) => {
+        if (isLegacy(func)) {
+          const valid = func.validate(value, props);
+          if (valid) {
+            return Promise.resolve();
+          }
+          return Promise.reject(func.message());
+        }
+        return func(value, props);
+      });
```
customcodepl: I wonder if we can do it more readable and separate this function.

src/utils/validations/validator/validator.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+  return valid;
+};
+
+const isLong = value => new Promise((resolve, reject) => {
+  if (value.length > 5) {
+    resolve(value);
+  } else {
+    reject(new Error('value is too short!'));
```
samhinton88: You should use the variables you've assigned above for these messages.

src/utils/validations/validator/validator.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    done();
+  });
+
+  it('receives a single input and returns the error message when the promise rejects', async (done) => {
+    const validate = await runValidation(presence, '');
+    expect(validate).toEqual(presErr);
+    done();
+  });
+
+  it('handles async validations as part of an array and returns true when all th epromises resolve', async (done) => {
```
samhinton88: typo

src/utils/validations/validator/validator.js
edleeks87: Yeah agreed, I've stripped out half of the code and I'll separate some of it into a new function for the legacy, I'll push up the changes soon

src/utils/validations/validator/validator.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    if (value === 'foo') {
+      resolve();
+    } else {
+      reject(new Error('not foo!'));
+    }
+  }, 500);
+});
+
+describe('validator', () => {
+  it('receives an array of functions and returns true when all promises resolve', async (done) => {
```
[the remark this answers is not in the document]
edleeks87: Cheers removed the done callback

package-lock.json
```diff
@@ -12465,7 +12465,7 @@
     "redux": {
       "version": "3.7.2",
       "resolved": "https://registry.npmjs.org/redux/-/redux-3.7.2.tgz",
-      "integrity": "sha512-pNqnf9q1hI5HHZRBkj3bAngGZW/JMCmexDlOxw4XagXY2o1327nHH54LoTjiPJ0gizoqPDRqWyX/00g0hD6w+A==",
+      "integrity": "sha1-BrcxIyFZAdJdBlvjQusCa8HIU3s=",
       "requires": {
         "lodash": "^4.2.1",
         "lodash-es": "^4.2.1",
```
andrewjtait: don't think this should've changed?

package-lock.json
```diff
@@ -71,12 +71,12 @@
       }
     },
     "@babel/generator": {
```
andrewjtait: looks like even more changes than last time ;)

try `git commit master -- ./package-lock.json` and commit the updated file.

src/utils/validations/validator/validator.spec.js
```diff
@@ -0,0 +1,127 @@
...
+  await validate(value)
+    .then(() => {
+      valid = true;
+      index += 1;
+    })
+    .catch((err) => {
+      if (err.message) {
+        valid = err.message;
+      } else {
+        valid = Array.isArray(validations) ? validations[index].message() : validations.message();
```
andrewjtait: why do you need this if you have `handleLegacyValidation`?

edleeks87: Long story short, we don't, I should of refactored this test when I changed the validator code 🤦‍♂️

src/utils/validations/validator/validator.js
```diff
@@ -0,0 +1,28 @@
...
+
+const handleLegacyValidation = (func, value, props) => {
+  if (func.validate(value, props)) {
+    return Promise.resolve();
+  }
+  return Promise.reject(func.message());
+};
+
+const validator = validationFunctions => (value, props) => (
+  new Promise(async (resolve, reject) => {
```
grabkowski: we shouldn't use async function as a promise executor - it's an antipattern

src/utils/validations/validator/validator.js
```diff
@@ -0,0 +1,25 @@
...
+        return handleLegacyValidation(func, value, props);
+      }
+      return func(value, props);
+    });
+
+    Promise.all(results)
+      .then(() => resolve())
+      .catch(err => reject(err));
+  })
+);
```
grabkowski: I think that it could be simplified to:
```
const validator = validationFunctions => (value, props) =>{
  .
  .
  .
  return Promise.all(results);
};
```

src/utils/validations/validator/validator.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+const validator = validationFunctions => (value, props) => {
+  const validations = Array.isArray(validationFunctions) ? validationFunctions : [validationFunctions];
+  const results = validations.map((func) => {
+    if (isLegacy(func)) {
+      return handleLegacyValidation(func, value, props);
+    }
+    return func(value, props);
+  });
```
customcodepl: What do you think about something like:
```
const handleLegacyValidation = (func, value, props) => func.validate(value, props) ? Promise.resolve() : Promise.reject(func.message());
const handleValidation = (func) => isLegacy(func) ? handleLegacyValidation(func, value, props) : func(value, props);
...
const results = validations.map(handleValidation);
```

edleeks87: I like the suggestion for `handleValidation` function so I've added it, I've not changed the `handleLegacyValidation` to ternary one as I feel it's too complex and makes it less readable

andrewjtait:
I'll run some dev QA on this today

PR #1916, opened 2019-02-14
Fe 1219 set up style themes

samhinton88:
# Description
Changes to make the deepMerge function non-mutative.


src/style/utils/merge-deep.spec.js
```diff
@@ -41,4 +41,10 @@ describe('mergeDeep', () => {
   it('merges the objects but does not create a new key when not passed an object', () => {
     expect(mergeDeep(base, undefined)).toEqual(base);
   });
+
+  it('is not mutative', () => {
+    mergeDeep(base, toMerge);
+    const secondObject = mergeDeep(base, objectWithArray);
+    expect(secondObject).toEqual({ a: 1, b: { nestedB: 1 }, c: ['inArray'] });
```
andrewjtait: should the actual test be that `base` is still the same? as that was the thing that was getting mutated.

samhinton88: You might just be right.

src/style/utils/merge-deep.spec.js
```diff
@@ -41,4 +41,9 @@ describe('mergeDeep', () => {
   it('merges the objects but does not create a new key when not passed an object', () => {
     expect(mergeDeep(base, undefined)).toEqual(base);
   });
+
+  it('is not mutative', () => {
+    mergeDeep(base, toMerge);
+    expect(base).toEqual(base);
```
andrewjtait: This is always going to pass, its the same as `expect(true).toEqual(true);` - you need to assert against some actual properties of the original object to conclude they haven't changed after the merge.

PR #1923, opened 2019-02-18
TS declaration for Flash component

blaky:
# Description
TypeScript type declaration for the Flash component. We would like to use the Flash component in a TypeScript based project and we prefer to have a type declaration file instead of using a generic module declaration.

# TODO
- [x] Release notes


blaky:
Hi @edleeks87 , the missing release note has been added.

blaky:
@andrewjtait just a question: Are you happy if we keep adding type declarations as we go? I think the next one we will need is textboxes and numeric inputs.

andrewjtait:
@blaky Yes as many as you want to add - thanks!

PR #1926, opened 2019-02-19
Type declarations for texbox, numeric and decimal components

blaky:
# Description

New TypeScript type declarations for the following components:
- Textbox
- Numeric
- Decimal

Generic interface added for Validators

InputProps interface added aligned with the content of `demo/utils/definition/input-definition/input-definition.js`

# TODO
- [x] Release notes


blaky:
We tried to use the types yesterday for the first time. They didn't work as expected, but we managed to fix them:

Instead of 
`declare const Link: React.Component<LinkProps,{}>;`,
 it should be 
`declare const Link: React.ComponentType<LinkProps>;`

edleeks87:
Thanks for letting us know @blaky I'll fix that and release it as a patch version

PR #1930, opened 2019-02-20
Fe 1115 rewrite input icon decorator

grabkowski:
# Description
Added two components - ```input-icon-toggle``` and ```validated-input-icon-toggle```

```input-icon-toggle``` when placed under the same tag as an ```input``` commponent adds a toggle icon  to it (e.g. calendar or dropdown arrow),
```validated-input-icon-toggle``` is a temporary solution to display validation icons and messages in the ```input-icon-toggle``` component.

```input-icon``` decorator was not removed to ensure backward compatibility.

### CODE REVIEW

File: src/components/input-icon-toggle/input-icon-toggle.scss [hunk 1]
```diff
@@ -0,0 +1,62 @@
+@import "./../../style-config/colors";
```
samhinton88: We should be building new components using Styled Components and not scss.

src/components/input-icon-toggle/input-icon-toggle.component.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <label htmlFor={ inputId } key='label-icon'>
+      { content || <Icon type={ iconType } className='input-icon-toggle' /> }
+    </label>
+  );
+};
+
+InputIconToggle.propTypes = {
+  /**
```
samhinton88: I think we're trying to minimise this documentation, so that it only contains the long description. The type, prop name and other attributes are readable from the `PropTypes` config.

grabkowski: documentation trimmed.

src/components/input-icon-toggle/input-icon-toggle.scss
grabkowski: There is no possibility to use variables and methods from scss in styled components without exporting them, we also have to keep the backward compatibility with previous implementation.

samhinton88: For sure, but we'll need the styled components implementation also.

andrewjtait: Wondering whether its safer and more valuable if we leave Date and Dropdown using the old implementation, and rather set the new components up within the new input components under /__experimental - then you have no ties to the Sass implementation and can use the flexbox layout etc which would be more beneficial?

andrewjtait: There may be some cross over with the work @samhinton88 is doing to update the new input to use styled-components - maybe it would be easier if we consolidate what we have so far and the styling can be set up as part of his PR? Can discuss this morning after standup

grabkowski: we will also need some documentation regarding usage of the style themes

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,16 @@
...
+const { colors } = Colors;
+const borderColor = '#668592';
+
+const Label = styled.label`
+  background-color: ${props => (props.isHovered ? colors.tertiary : colors.primary)};
+  color: ${props => (props.isHovered ? colors.white : colors.text.body)};
+  border-left: 1px solid ${borderColor};
+  cursor: pointer;
+  padding: 6px;
+  margin: -2px -6px; /*correction for parent element padding*/
```
dquilter: This seems quite situational; is it always going to be necessary?

I'm probably just missing some context on the way this component is used...

grabkowski: It's a leftover from the previous implementation. Unfortunately it have to stay until parent component is refactored.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.spec.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
samhinton88: It might be a good idea to add some specs that are not simply snapshot testing, but rather tests the functionality in a more conventional, unit test manner.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+import Colors from 'style/themes/base';
```
samhinton88: This is not the correct way to extract themes from styled-components. The style rules are provided on the props of the component itself, and should be accessed on the callbacks passed into the template literals.

```
styled.div`
  background: ${(props) => props.theme.color};
`
```

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,16 @@
...
+const Label = styled.label`
+  background-color: ${props => (props.isHovered ? colors.tertiary : colors.primary)};
+  color: ${props => (props.isHovered ? colors.white : colors.text.body)};
+  border-left: 1px solid ${borderColor};
+  cursor: pointer;
+  padding: 6px;
+  margin: -2px -6px; /*correction for parent element padding*/
+`;
+
+export default Label;
```
samhinton88: After you declare the styles, you should add default props, and set defaultProps.theme to the baseTheme.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.spec.js
grabkowski: added a test to check if an `Icon` component is rendered without the `content` prop. Rest of it just tests the style changes based on props, so I think that snapshots are more proper for that.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
grabkowski: fixed

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
grabkowski: good point, fixed

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,45 @@
...
+
+function getColor({ theme, isHovered, disabled }) {
+  const { colors } = theme;
+
+  if (disabled) {
+    return colors.text.disabled;
+  }
+
+  if (isHovered) {
+    return colors.text.white;
```
samhinton88: I don't think that this is a valid value on the theme object.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return colors.text.body;
+}
+
+function getBackground({ theme, isHovered, disabled }) {
+  const { colors } = theme;
+
+  if (disabled) {
+    return colors.secondary;
```
samhinton88: Are these the correct colours?

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
grabkowski: it matches the base theme config:
https://github.com/Sage/carbon/blob/master/src/style/themes/base/base-theme.config.js

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
grabkowski: these are the only colors available through current base theme config

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
samhinton88: `colors.text.white` will return `undefined`

grabkowski: you're right, fixed :)

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
samhinton88: image

grabkowski: fixed

Issue #1921, opened 2019-02-18
Installing from Github release causes problem


blaky:
Hi there,

We tried to integrate carbon to one of our projects, but when we install it we got the following issue caused by a carbon dependency:

```
npm http fetch GET 500 https://github.com/fnando/i18n-js/archive/v3.0.0.rc12.tar.gz 72328ms attempt #3
npm timing stage:rollbackFailedOptional Completed in 1ms
npm timing stage:runTopLevelLifecycles Completed in 89230ms
npm verb stack Error: 500 Internal Server Error - GET https://github.com/fnando/i18n-js/archive/v3.0.0.rc12.tar.gz
npm verb stack     at res.buffer.catch.then.body (/Users/bencelaky/.nvm/versions/node/v8.9.4/lib/node_modules/npm/node_modules/npm-registry-fetch/check-response.js:104:15)
npm verb stack     at <anonymous>
npm verb stack     at process._tickCallback (internal/process/next_tick.js:188:7)
npm verb statusCode 500
npm verb pkgid i18n-js@http://github.com/fnando/i18n-js/archive/v3.0.0.rc12.tar.gz
npm verb cwd /Users/bencelaky/Dev/x3/etna
npm verb Darwin 18.0.0
```

I tried to get around it many ways, but I couldn't get rid of the error. It seemed to be caused by the following dependency:

https://github.com/Sage/carbon/blob/e75efca87d112131b6ed018727d27d2ff5d6d3e4/package.json#L81

Is there any chance you could upgrade it to the final release from NPM instead of the tar.gz file?

PR #1931, opened 2019-02-20
i18n-js upgrade

andrewjtait:
# Description

Upgrades i18n-js to the latest version. There is a breaking change with this version, only if you're relying on defining/providing translations outside of the common js system. We have noted this in the release notes to help developers upgrade.

#1921


andrewjtait:
Potentially we release this as a breaking change.

andrewjtait:
We have incompatibilities with this version in our main application, therefore I have moved this dep into a peer dep temporarily until we resolve these issues.

andrewjtait:
More detail on the incompatibilities, we have a custom extension to i18n-js which overrides several methods in the library to change its implementation. See the I18nScoped extension in GAC.

PR #1932, opened 2019-02-21
fixes definition files and exports type form helper file

edleeks87:
# Description
Fixes the incorrect definition files to allow Typescript to work as expected, currently the typing does not work so this will allow developers to use Intellisense features when using the components


release_notes/v6_1_1.html.md
```diff
@@ -0,0 +1,5 @@
+# v6.1.1 (2019-02-21)
+### Bug Fixes
+* Fixes the error in the type definition files
```
samhinton88: Should this not be a little more descriptive of the specific bug / error it's fixing?

edleeks87: Added some more to it

PR #1936, opened 2019-02-23
SelectAsync

andrewjtait:
SelectAsync renders a regular Select, but wraps it in additional functionality to allow it to fetch its options from an API endpoint.
 
### The Request

By default the component will execute a request with the following options:

```
{
  params: {
    page: 1
    items_per_page: 20,
    search: 'any filter text'
  }
}
```

This can be customised using the formatRequest prop, for example to enable withCredentials:

```
<SelectAsync
  endpoint='/endpoint'
  formatRequest={ opts => ({ ...opts, withCredentials: true }) }
/>
```

The opts object is used with axios, please see this URL for more options: https://github.com/axios/axios#request-config

### The Response
By default the response is expected in the following format:

```
{
  $items: [Array] - required, the items to render
  $page: [Integer] - optional, for paginated responses
  $total: [Integer] - optional, for paginated responses
}
```

If your response does not match this format you can modify it using the formatResponse prop:

```
<SelectAsync
  endpoint='/endpoint'
  formatResponse={ response => ({ ...response, data: { $items: response.myItems } }) 
/>
```

This follows the axios response schema: https://github.com/axios/axios#response-schema

### Rendering Options

By default options will be rendered under the assumption they have the following format:
 
```
{
  id [Integer] - the unique identifier
  displayed_as [String] - the human readable value
}
```

If your items do not match this format, or you want to have custom control over how they are rendered, you can provide a function as a child:
 
```
<SelectAsync endpoint='/endpoint'>
  {
    items => items.map(item => <Option>{ item.text }</Option>)
  }
</SelectAsync>
```


src/__experimental__/components/select-async/select-async.component.js
```diff
@@ -0,0 +1,185 @@
...
+      this.props.endpoint,
+      this.fetchOptions(page, search)
+    );
+    this.handleResponse(data);
+  }
+
+  fetchNextPage = () => {
+    if (this.state.fetching) return;
+    if (this.state.items.length === this.state.total) return;
+    this.fetchData({ page: this.state.page + 1 });
```
grabkowski: object destructuring could be used here, to refer ```this.state``` only once

src/__experimental__/components/select-async/select-async.component.js
```diff
@@ -0,0 +1,185 @@
...
+  buildOptions(page, items) {
+    return (page > 1) ? this.state.items.concat(items) : items;
+  }
+
+  renderOptions() {
+    if (this.props.children) return this.props.children(this.state.items);
+
+    if (!this.state.items.length) return null;
+
+    return this.state.items.map(item => (
```
grabkowski: ```this.state.items``` could be a const

src/components/scrollable-list/__spec__.js
```diff
@@ -320,13 +320,13 @@ describe('ScrollableList', () => {
...
+      scrollableList.simulate('scroll', { target: { scrollHeight: 400, scrollTop: 101 } });
       expect(onLazyLoad).toBeCalled();
     });
 
     it('does not trigger onLazyLoad when the scrollTop is not beyond the scrollHeight', () => {
       onLazyLoad.mockReset();
-      scrollableList.simulate('scroll', { target: { scrollHeight: 300, scrollTop: 99 } });
+      scrollableList.simulate('scroll', { target: { scrollHeight: 400, scrollTop: 99 } });
       expect(onLazyLoad).not.toBeCalled();
     });
```
grabkowski: shouldn't that file be renamed to: ```scrollable-list.spec.js```?

src/__experimental__/components/select-async/select-async.component.js
andrewjtait: ✅

src/__experimental__/components/select-async/select-async.component.js
andrewjtait: ✅

src/components/scrollable-list/__spec__.js
andrewjtait: ✅

src/__experimental__/components/select-async/select-async.spec.js
```diff
@@ -0,0 +1,170 @@
...
+      const wrapper = await openResults(render());
+      await getNextPage(wrapper);
+      expect(findResults(wrapper)).toMatchSnapshot();
+    });
+
+    it('does not refetch if a fetch is already in progress', async () => {
+      mockResponse();
+      const wrapper = await openResults(render());
+      expect(axios.get.mock.calls.length).toEqual(1);
+      getNextPage(wrapper);
```
samhinton88: Missing an `await` here I think.

src/__experimental__/components/select-async/select-async.spec.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+  });
+
+  describe('custom response', () => {
+    it('throws an error if the response is not supported', async () => {
+      mockResponse({ data: {} });
+      try {
+        await render().instance().fetchData({ page: 1 });
+      } catch ({ message }) {
+        expect(message).toEqual(`The server response does not contain an $items attribute.
```
samhinton88: Perhaps pop this in a different file and export it to both the test and component, so a change to error message doesn't break the test.

src/__experimental__/components/select-async/select-async.spec.js
andrewjtait: nah this is the subtle difference between this test and the next, this one is testing if a fetch is already in progress (so i do not want to wait for it to finish) maybe i should write a comment next to this line to that effect

src/__experimental__/components/select-async/select-async.spec.js
andrewjtait: alrighty

src/__experimental__/components/select-async/select-async.component.js
```diff
@@ -0,0 +1,185 @@
...
+      }
+    };
+
+    if (this.props.formatRequest) opts = this.props.formatRequest(opts);
+
+    return opts;
+  }
+
+  handleResponse(responseData) {
+    let data = responseData;
```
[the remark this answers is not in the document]
andrewjtait: The problem was a I cannot re-assign `responseData` - do you know if I can write a different way around that?

src/__experimental__/components/select-async/select-async.component.js
```diff
@@ -0,0 +1,185 @@
...
+
+  filterValue() {
+    return this.select.current.state.filter;
+  }
+
+  fetchOptions(page, search) {
+    let opts = {
+      params: {
+        page,
+        items_per_page: 20,
```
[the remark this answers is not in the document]
andrewjtait: I can pass this in as a prop 👍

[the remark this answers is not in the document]
andrewjtait: ✅

src/__experimental__/components/select-async/select-async.component.js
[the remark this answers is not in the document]
andrewjtait: ✅ changed it a little bit

src/__experimental__/components/select-async/select-async.spec.js
andrewjtait: ✅

src/__experimental__/components/select-async/select-async.spec.js
andrewjtait: ✅

PR #1937, opened 2019-02-25
i18n peer dep

andrewjtait:
We have changed i18n-js into a peer dependency. This means developers will need to install and manage which version of i18n-js they want to use in their own applications. We have done this as some applications using Carbon require an older version of i18n-js, while some would like to use newer versions.


blaky:
Hi there, when are you guys planning to merge and release this change?

edleeks87:
Hi @blaky I'll cut a new release as soon as this gets merged, hopefully by the end of the day

blaky:
Thanks @edleeks87 !

PR #1938, opened 2019-02-25
Extend Table component for custom empty row

philearley:
# Description
* At the moment, when using the TableAjax provided by Carbon if you want to change the experience of when the Table is empty, you lose the loading functionality.
* To prevent reimplementing all the logic for if the Table is in a loading state, this is allowing a backward compatibility of easily extending table.

Agreement with @andrewjtait to extend Carbon for this functionality.

# Testing Instructions
1. Create a TableAjax in the demo wrapper with no data
1. - [x] Observe that it is showing the default no records found 
1. Create a TableAjax in the demo wrapper with no data and a customEmptyRow
1. - [x] Observe that it is showing the custom empty row.

### DIFF
--- a/src/components/table/__spec__.js
+++ b/src/components/table/__spec__.js
@@ -885,7 +885,7 @@ describe('Table', () => {
     describe('if there are children that are immutable', () => {
       it('returns the children if there are children', () => {
         let data = Immutable.fromJS([{ foo: 1 }, { foo: 2 }]),
-            children = data.map((child, index) => { return <tr key={ index }></tr>; });
+            children = data.map((_, index) => <tr key={ index }></tr>);
         instance = TestUtils.renderIntoDocument(<Table>{ children }</Table>);
 
         expect(instance.tableContent).toEqual(children);
@@ -893,7 +893,7 @@ describe('Table', () => {
 
       it('returns the loadingRow if no children and not yet received data', () => {
         let data = Immutable.fromJS([]),
-            children = data.map((child, index) => { return <tr key={ index }></tr>; });
+            children = data.map((_, index) => <tr key={ index }></tr>);
         instance = TestUtils.renderIntoDocument(<Table>{ children }</Table>);
         instance._hasRetreivedData = false;
 
@@ -902,7 +902,7 @@ describe('Table', () => {
 
       it('returns the emptyRow if no children and has received data', () => {
         let data = Immutable.fromJS([]),
-            children = data.map((child, index) => { return <tr key={ index }></tr>; });
+            children = data.map((_, index) => <tr key={ index }></tr>);
         instance = TestUtils.renderIntoDocument(<Table>{ children }</Table>);
         instance._hasRetreivedData = true;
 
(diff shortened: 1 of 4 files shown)


src/components/table/__spec__.js
```diff
@@ -928,6 +928,18 @@ describe('Table', () => {
         expect(instance.tableContent.get(0)).toEqual(instance.props.children.get(0));
         expect(instance.tableContent.get(1)).toEqual(instance.emptyRow);
       });
+
+      describe('when a customEmptyRow has been provided', () => {
+        test('the custom empty row is rendered instead of the default', () => {
+          let data = Immutable.fromJS([]),
+              children = data.map((child, index) => { return <tr key={ index }></tr>; });
```
toni-leigh: If you aren't using a value in a function like this, but are forced to name it as you need the second, then using an `_` I think is best practice (in some languages at least):

`children = data.map((_, index) => { return <tr key={ index }></tr>; });`

Also, you can get rid of some of the clutter in this line with implied returns etc.

`children = data.map((_, index) => <tr key={ index }></tr>);`

philearley:
Tested with @Denise-Moorhouse 
Using my local Carbon with: `http://0.0.0.0:8095/sandbox`

We created a TableAjax component with no data, observing the table showing the default no results to display row.

We created a TableAjax component with no data and provided a customEmptyRow, observing the table showing the custom empty row component.

src/components/table/__spec__.js
```diff
@@ -885,15 +885,15 @@ describe('Table', () => {
     describe('if there are children that are immutable', () => {
       it('returns the children if there are children', () => {
         let data = Immutable.fromJS([{ foo: 1 }, { foo: 2 }]),
```
samhinton88: If a variable is not intended to be reassigned it should be declared with `const`

src/components/table/__spec__.js
```diff
@@ -885,15 +885,15 @@ describe('Table', () => {
...
-            children = data.map((child, index) => { return <tr key={ index }></tr>; });
+            children = data.map((_, index) => <tr key={ index }></tr>);
         instance = TestUtils.renderIntoDocument(<Table>{ children }</Table>);
 
         expect(instance.tableContent).toEqual(children);
       });
 
       it('returns the loadingRow if no children and not yet received data', () => {
         let data = Immutable.fromJS([]),
```
samhinton88: If a variable is not intended to be reassigned it should be declared with `const`

src/components/table/__spec__.js
```diff
@@ -902,7 +902,7 @@ describe('Table', () => {
 
       it('returns the emptyRow if no children and has received data', () => {
         let data = Immutable.fromJS([]),
```
samhinton88: If a variable is not intended to be reassigned it should be declared with `const`

src/components/table/__spec__.js
```diff
@@ -928,6 +928,18 @@ describe('Table', () => {
         expect(instance.tableContent.get(0)).toEqual(instance.props.children.get(0));
         expect(instance.tableContent.get(1)).toEqual(instance.emptyRow);
       });
+
+      describe('when a customEmptyRow has been provided', () => {
+        test('the custom empty row is rendered instead of the default', () => {
+          let data = Immutable.fromJS([]),
```
samhinton88: If a variable is not intended to be reassigned it should be declared with `const`

PR #1942, opened 2019-02-28
Fe 1218 implement styled components logic on input

samhinton88:
# Description
Introduces a function which queries global theme objects to return correct style rules for styled components.


src/__experimental__/components/textbox/textbox.stories.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean, text, select } from '@storybook/addon-knobs';
+import Textbox from '.';
+
+storiesOf('Textbox', module)
```
customcodepl: Would you like to change the label from `Textbox` to `Experimental/Textbox'?

src/style/utils/query-theme.js
```diff
@@ -0,0 +1,28 @@
+
+const isUpper = char => char.toUpperCase() === char;
+const camelToDash = (acc, char) => `${acc}${isUpper(char) ? `-${char.toLowerCase()}` : char}`;
+const toCSSCase = str => str.split('').reduce(camelToDash, '');
```
grabkowski: I think it's called kebab case :)

grabkowski: it could be done more straightforward: `str.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();`

src/__experimental__/components/input/input-presentation.component.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import './input-presentation.style.scss';
+import InputPresentationStyle from './input-presentation.style';
+// import './input-presentation.style.scss';
```
dquilter: Unneeded commented-out import.

src/__experimental__/components/input/input-presentation.spec.js
```diff
@@ -1,9 +1,37 @@
...
+const deletekeys = (obj, keys) => keys.forEach(k => delete obj[k]);
+const stripKeys = (obj, keys) => {
+  return Object.keys(obj).reduce((acc, key) => {
+    const strippedObj = deletekeys(obj[key], keys);
+    return { ...acc, [key]: { ...strippedObj } };
+  }, {});
+};
+
+const isUpperCase = char => char.toUpperCase() === char;
+const toCSSCase = (str) => {
```
dquilter: Seems like these methods (`deletekeys`, `stripKeys`, `isUpperCase`, `toCSSCase`) might be useful in other specs? Possibly move to a helper file?

src/__experimental__/components/input/input-presentation.spec.js
```diff
@@ -1,9 +1,37 @@
 import React from 'react';
-import { shallow } from 'enzyme';
-import { InputPresentation } from '.';
+import { shallow, mount } from 'enzyme';
+import { InputPresentation, Input } from '.';
+import baseTheme from '../../../style/themes/base';
+import 'jest-styled-components';
+
+const deletekeys = (obj, keys) => keys.forEach(k => delete obj[k]);
```
dquilter: Camelcase? `deleteKeys`

src/__experimental__/components/input/input-presentation.spec.js
dquilter: And `assertStyleMatch `

src/__experimental__/components/input/input-presentation.spec.js
```diff
@@ -16,6 +25,100 @@ describe('InputPresentation', () => {
...
+              background: inputStyleRules.base.background,
+              border: `2px solid ${warning}`
+            },
+            mountRender({ warning: true })
+          );
+        });
+      });
+    });
+  });
+
```
grabkowski: do we really need to test all those styles? I think that snapshots should do the same thing.

src/style/themes/base/components/input-presentation.js
```diff
@@ -0,0 +1,55 @@
+export default ({ palette, blackWithOpacity }) => ({
```
grabkowski: I think that component specific data should be placed in the folder of that component. It could be difficult to maintain later, when we have it spread in multiple places.

src/style/utils/query-theme.js
```diff
@@ -0,0 +1,63 @@
...
+      if (!rest[prop]) return subAcc;
+      return { ...subAcc, ...profile[key][prop] };
+    }, {});
+
+    return { ...acc, ...byState };
+  }, {});
+
+  const cssString = objToCssString({ ...profile.base, ...styleGivenContext });
+  return cssString;
+};
```
grabkowski: we need unit tests for those utils

src/__experimental__/components/select/__snapshots__/select.spec.js.snap
```diff
@@ -72,12 +72,17 @@ Object {
...
+    data-element="input"
+    id="guid-12345"
     role="presentation"
+    type="text"
+    value=""
   >
     <input
       autocomplete="off"
-      class="carbon-input"
+      class="sc-bxivhb jITibh"
```
grabkowski: `import 'jest-styled-components';` missing in `select.spec.js`?

src/__spec_helper__/test-utils.js
```diff
@@ -0,0 +1,21 @@
...
+  }, {});
+};
+const isUpper = char => char.toUpperCase() === char;
+const humpToDash = (acc, char) => `${acc}${isUpper(char) ? `-${char.toLowerCase()}` : char}`;
+
+
+export const isUpperCase = char => char.toUpperCase() === char;
+export const toCSSCase = (str) => {
+  return str.split('').reduce(humpToDash, '');
+};
```
grabkowski: isn't that the same as `toKebabCase` in `query-utils.js`?

PR #1948, opened 2019-03-04
FE-1242 validations component and context

edleeks87:
# Description
Creates some higher order components to handle the validation of forms and their child inputs using the new promise based validator function.
The parent HOC provides context and state to handle the management of the validations of its children components.
The child component registers itself and its validations with the parent when it mounts and the parent iterates over the registered children running each one's validations, blocking unless all pass.


src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+                }
+                break;
+              case 'warnings':
+                if (this.state.hasWarning !== false && this.context.adjustCount) {
+                  this.setState({ hasWarning: false });
+                  this.context.adjustCount('warning', -1);
+                }
+                break;
+              case 'info':
+                if (this.state.hasInfo !== false && this.context.adjustCount) {
```
grabkowski: is there a reason why it's not just `if (this.state.hasInfo && . . .)`?

edleeks87: 🤦‍♂️ been a long week haha, will change them

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+                  this.setState({ hasInfo: false });
+                  this.context.adjustCount('info', -1);
+                }
+                break;
+              default:
```
grabkowski: that could be removed as it does nothing

edleeks87: The linter complains without a default branch, I've added an action so it does something on default now

grabkowski: you could try to just add a comment:
`    // no default`
and linter should be happy :)

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+      if (this.state.hasInfo) type = 'info';
+      if (this.state.hasWarning) type = 'warning';
+      if (this.state.hasError) type = 'error';
+      if (type) {
+        if (Array.isArray(this.props.children)) {
+          return [...this.props.children, <Icon key={ `${type}-icon` } type={ type } />];
+        }
+        return [this.props.children, <Icon key={ `${type}-icon` } type={ type } />];
+      }
+      return this.props.children;
```
grabkowski: `this.props.children` could be a const

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+            resolve(false);
+          });
+      });
+    }
+
+    renderValidationMarkup = () => {
+      let type;
+      if (this.state.hasInfo) type = 'info';
+      if (this.state.hasWarning) type = 'warning';
+      if (this.state.hasError) type = 'error';
```
grabkowski: are this conditionals exclusive? If yes, then it should be `if/else if` statement

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: I've added the ignore comments for now open to changing it if anyone else doesn't like it

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: No they are inclusive I need it to evaluate them all

grabkowski: so there could be `hasInfo`, `hasWarning` and `hasError` in one case and hasError overrides them all?

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+      if (type) {
+        if (Array.isArray(this.props.children)) {
+          return [...this.props.children, <Icon key={ `${type}-icon` } type={ type } />];
+        }
+        return [this.props.children, <Icon key={ `${type}-icon` } type={ type } />];
```
grabkowski: `<Icon key={ `${type}-icon` } type={ type } />` could be a const

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+    wrapper2 = shallow(<WithNoName />);
+    expect(wrapper2).toMatchSnapshot();
+  });
+
+  describe('renderValidationMarkup', () => {
+    it('returns an Icon with info type when the state has info and no warning or validations', () => {
+      wrapper.setState({ hasInfo: true });
+      checkState([stateId[2]], wrapper.state(), true);
+      const markUp = wrapper.instance().renderValidationMarkup();
+      const size = markUp.length;
```
grabkowski: `markup`?

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+
+const checkState = (ids, state, res) => {
+  Object.keys(state).forEach((key) => {
+    if (ids.includes(key)) {
+      ids.forEach(id => expect(state[id]).toEqual(res));
+    } else {
+      expect(state[key]).toEqual(!res);
+    }
+  });
+};
```
grabkowski: I think that we shouldn't separate `expect` methods from the tests. It decreases code readability.

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+
+  it('matches the snapshot when the hoc is passed an anonymous function with no name or display name', () => {
+    wrapper2 = shallow(<WithNoName />);
+    expect(wrapper2).toMatchSnapshot();
+  });
+
+  describe('renderValidationMarkup', () => {
+    it('returns an Icon with info type when the state has info and no warning or validations', () => {
+      wrapper.setState({ hasInfo: true });
+      checkState([stateId[2]], wrapper.state(), true);
```
grabkowski: If I understand correctly: `checkState` method is testing if `hasWarning` state is set to `false`, by setting `true` as it's third argument?  It's a bit counterintuitive IMHO.
I think that `expect(wrapper.state().hasWarning).toEqual(false)` would be more readable.

grabkowski: do setting the state of `hasInfo` has impact on the `hasWarning` state?

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+      const markUp = wrapper.instance().renderValidationMarkup();
```
grabkowski: I think that should be `markup`?

src/utils/validations/component/validations/with-validation.spec.js
grabkowski: could be `const lastMarkup = markUp[markUp.length - 1]`

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+      if (this.props.onBlur) this.props.onBlur(ev);
+    }
+
+    render() {
+      return (
+        <WrappedComponent
+          onBlur={ this.handleBlur }
+          { ...this.props }
+        >
+          { this.renderValidationMarkup() }
```
grabkowski: I think that we could extract the children prop from the `renderValidationMarkup` method:
```
{ this.props.children }
{ this.renderValidationMarkup() }
```

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+      const size = markUp.length;
+      expect(markUp[size - 1].type.displayName).toEqual('Icon');
+      expect(markUp[size - 1].props.type).toEqual('info');
```
grabkowski: it could be just:
`expect(wrapper.contains(<Icon type='info' />)).toBeTruthy();`

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,348 @@
...
+      expect(wrapper).toMatchSnapshot();
```
grabkowski: I think that we don't need that snapshot

src/utils/validations/component/validations/with-unique-name.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    expect(wrapper.props().name).not.toEqual(undefined);
+  });
+
+  it('either allows the name prop to be overriden or randomnly generates one', () => {
+    const InputComponent = withUniqueName(MockComponent);
+    wrapper = shallow(<InputComponent name='foo' />);
+    expect(wrapper.props().name).toEqual('foo');
+    wrapper = shallow(<InputComponent />);
+    const { name } = wrapper.props();
+    expect(wrapper.props().name).toEqual(name);
```
grabkowski: that test is redundant

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,159 @@
...
+              case 'info':
+                this.setState({ hasInfo: e });
+                if (this.context.adjustCount) this.context.adjustCount('info', 1);
+                break;
+              default:
+            }
+            resolve(false);
+          });
+      });
+    }
```
grabkowski: That method is huge and contains multiple nested blocks. Could you flatten it by splitting it into separate methods?

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Yeah hasError takes precedence over the other states

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Agreed, I've made the changes as it does seem unnecessary to have it handled in `renderValidationMarkup`

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: I've split it up into some smaller functions, I noticed an edge case whilst doing it so had to add some further implementation to prevent the same error being counted multiple times etc

src/utils/validations/component/validations/with-validation.spec.js
edleeks87: Yeah, I removed it, what I was using the `expect` for was pointless really anyway

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,188 @@
...
+      adjustCount: PropTypes.func
+    }
+
+    state = {
+      hasError: false,
+      hasWarning: false,
+      hasInfo: false
+    };
+
+    currentFailedValidations = { error: [], warning: [], info: [] };
```
samhinton88: This could be passed back up to the context provider rather than stored here.

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,188 @@
...
+      if (result && types.includes('warnings')) result = await this.runValidation('warnings');
+      if (result && types.includes('info')) result = await this.runValidation('info');
+      return result;
+    }
+
+    validationsResult = (type, errorStatus) => {
+      switch (type) {
+        case 'validations':
+          if (errorStatus) {
+            this.addFailedValidation('error', errorStatus.message);
```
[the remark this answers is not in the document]
edleeks87: Good point, I don't think there is a way to avoid race conditions as the way we have validations set up now the Promise rejects on the first validation that fails, and this could be a different one each time if they are all asynchronous.

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,188 @@
...
+
+    addFailedValidation = (type, message) => {
+      const { adjustCount } = this.context;
+      this.currentFailedValidations[type].push(message);
+      adjustCount(type, 1);
+    }
+
+    removeFailedValidations = (type) => {
+      const array = this.currentFailedValidations[type];
+      const { adjustCount } = this.context;
```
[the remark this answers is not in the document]
edleeks87: Maybe a better approach is to just to treat the counts on the input rather than to try and count the number of failures per input? Asynchronous validations won't really allow and accurate aggregation to be calculated as the Promise.all will just reject on the first validation that fails.

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    removeInput = (name) => {
+      delete this.inputs[name];
+    }
+
+    adjustCount = (type, adjustment) => {
+      switch (type) {
+        case 'error':
+          this.setState((prev) => { return { errorCount: prev.errorCount + adjustment }; });
```
dquilter: Just wondering about the reason for the function. I would probably write this as:

`this.setState({errorCount: this.state.errorCount + adjustment };`

src/utils/validations/component/validations/validations.context.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export default React.createContext();
```
dquilter: Seems like this could be a one-liner in whatever files you're importing this into?

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,74 @@
...
+    addInput = (name, validate) => {
+      this.inputs[name] = validate;
+    }
+
+    removeInput = (name) => {
+      delete this.inputs[name];
+    }
+
+    adjustCount = (type, adjustment) => {
+      switch (type) {
```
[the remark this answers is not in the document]
edleeks87: Yeah I agree was thinking they could be more dynamic

src/utils/validations/component/validations/form-with-validations.hoc.js
edleeks87: The linter complains when you don't use a callback when referencing previous state, which will make the build fail

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,74 @@
...
+        removeInput: this.removeInput,
+        adjustCount: this.adjustCount
+      };
+    }
+
+    validate = () => {
+      let results;
+      return new Promise((resolve, reject) => {
+        results = Object.keys(this.inputs).map((name) => {
+          const validate = this.inputs[name];
```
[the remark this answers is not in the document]
edleeks87: Yeah good point, I'll rename it to `validateRegisteredInputs`

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+            errorCount={ this.state.errorCount }
+            warningCount={ this.state.warningCount }
+            infoCount={ this.state.infoCount }
+            validationTypes={ this.validationTypes }
+            { ...this.props }
+          />
+        </ValidationsContext.Provider>
+      );
+    }
+  };
```
andrewjtait: Its good practice to give you HOC a custom displayName so it is visible in dev tools etc.

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+
+    adjustCount = (type, validationResult) => {
+      if (this.validationTypes.includes(type)) {
+        const stateProp = `${type}Count`;
+        let adjustment = -1;
+
+        if (validationResult) adjustment = 1;
+        else if (this.state[stateProp] === 0) adjustment = 0;
+
+        this.setState((prev) => { return { [stateProp]: prev[stateProp] + adjustment }; });
```
andrewjtait: slightly tidier (less brackets):
```js
this.setState(prev => ({ [stateProp]: prev[stateProp] + adjustment }));
```

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+
+    getContext = () => {
```
andrewjtait: I'd avoid the use of class properties if possible, for example this method does not need to be one - as do a couple of others.

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+        addInput: this.addInput,
+        removeInput: this.removeInput,
+        adjustCount: this.adjustCount
+      };
+    }
+
+    validateRegisteredInputs = () => {
+      let results;
+      return new Promise((resolve, reject) => {
+        results = Object.keys(this.inputs).map((name) => {
```
andrewjtait: `map` isn't really the correct method to do here.

Consider using `for..in` instead.

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+          const validate = this.inputs[name];
+          return validate().catch(e => reject(e));
+        });
+        return Promise.all(results).then(() => resolve(true));
```
andrewjtait: is `results` here not just an array of booleans, all of which would be truthy?

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+        adjustCount: this.adjustCount
+      };
+    }
+
+    validateRegisteredInputs = () => {
+      let results;
+      return new Promise((resolve, reject) => {
+        results = Object.keys(this.inputs).map((name) => {
+          const validate = this.inputs[name];
+          return validate().catch(e => reject(e));
```
andrewjtait: does this iterate through all inputs? or will it stop at the first error?

src/utils/validations/component/validations/with-unique-name.hoc.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import guid from '../../../helpers/guid/guid';
+
+const withUniqueName = WrappedComponent => props => (
+  <WrappedComponent name={ guid() } { ...props } />
```
andrewjtait: is this not going to continuously generate a new guid each render?

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+
+const withValidation = (WrappedComponent) => {
+  class WithValidation extends React.Component {
+    static contextType = ValidationsContext;
+
+    static propTypes = {
+      children: PropTypes.node, // Children elements
+      name: PropTypes.string, // Name to uniquely identify the component
+      value: PropTypes.string, // The current value of the component
+      error: PropTypes.oneOfType([ // The error validations that should be run against the value
```
andrewjtait: this should be `validations` for backwards compat

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+        PropTypes.func,
+        PropTypes.arrayOf(PropTypes.func)
+      ]),
+      validationTypes: PropTypes.arrayOf(PropTypes.string), // the types of validations to be run on the component
+      warning: PropTypes.oneOfType([ // The warnings validations that should be run against the value
```
andrewjtait: this should be `warnings` for backwards compat

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+          adjustCount(type);
+        }
+        this.setState(newState);
+      }
+      return !errorStatus;
+    }
+
+    runValidation = (type) => {
+      return new Promise((resolve) => {
+        if (!this.context.adjustCount) return resolve(false);
```
andrewjtait: does this mean it cannot run the validation without the form?

src/utils/validations/component/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,73 @@
...
+
+    addInput = (name, validate) => {
+      this.inputs[name] = validate;
+    }
+
+    removeInput = (name) => {
+      delete this.inputs[name];
+    }
+
+    adjustCount = (type, validationResult) => {
```
grabkowski: I think that it could be a boolean value (maybe `hasError`?), as `validationResult` is an ErrorStatus passed from the validation.

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+    static contextTypes = {
+      addInput: PropTypes.func, // Function to allow child components to register with parent
+      removeInput: PropTypes.func, // Function to allow child components to unregister with parent
+      adjustCount: PropTypes.func // Function to allow child components to increment the count when a validation fails
+    }
+
+    state = {
+      hasError: false,
+      hasWarning: false,
+      hasInfo: false
```
grabkowski: shouldn't these values be strings?

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+
+        if (errorStatus) {
+          adjustCount(type, errorStatus);
+          newState = { [stateProp]: errorStatus };
+        } else if (this.state[stateProp]) {
+          adjustCount(type);
+        }
+        this.setState(newState);
+      }
+      return !errorStatus;
```
grabkowski: do we need it to return something?

src/utils/validations/component/validations/with-unique-name.hoc.js
edleeks87: The tests seem to indicate that the component's guid is persisted when I call update on the wrapper

src/utils/validations/component/validations/form-with-validations.hoc.js
edleeks87: Good point seems redundant to pass the actual error up here, I'll just pass in a boolean. I'm going to rename it `hasFailed` as it covers all of the validation subsets that may have failed

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: The return value is the boolean I pass into the resolve of the Promise, I could remove it and just pass it  a boolean instead I guess but not sure it makes much difference 🤔

grabkowski: `errorStatus` is passed as an argument in the `updateValidationStatus` method, so we are returning that argument converted to boolean and negated. This method could just update the state without returning anything.

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+        if (!this.context.adjustCount) return resolve(false);
+        if (this.props[type].length === 0) return resolve(true);
+
+        const validate = validator(this.props[type]);
+        return validate(this.props.value)
+          .then(() => {
+            return resolve(this.updateValidationStatus(type));
+          })
+          .catch((error) => {
+            return resolve(this.updateValidationStatus(type, error));
```
grabkowski: referring to my comment in the `updateValidationStatus` method:
```
.then(() => {
  this.updateValidationStatus(type);
  resolve(true);
})
catch((error) => {
  this.updateValidationStatus(type, error);
  resolve(false);
})
```

src/utils/validations/component/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,135 @@
...
+      }
+      return !errorStatus;
+    }
+
+    runValidation = (type) => {
+      return new Promise((resolve) => {
+        if (!this.context.adjustCount) return resolve(false);
+        if (this.props[type].length === 0) return resolve(true);
+
+        const validate = validator(this.props[type]);
```
grabkowski: it's a bit confusing when we have multiple methods with the same names, could you rename it?

grabkowski: or it could be just `validator(this.props[type])(this.props.value)`

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Yeah on second thought it is more readable to go with your approach, I will change it and push it up later

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: I'm going to leave it as false or error

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Removed the line

src/utils/validations/component/validations/form-with-validations.hoc.js
edleeks87: Changed the implementation to fix this

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Ok will change it back

src/utils/validations/component/validations/with-validation.hoc.js
edleeks87: Ok will change it back

edleeks87:
@stevebutler2210 yeah I was waiting to talk to him to clear some stuff up, I'll resolve them now an push up the new changes asap

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,383 @@
...
+
+  it('matches the snapshot when the HOC is passed an anonymous component with no name or display name', () => {
+    wrapper2 = shallow(<WithNoDisplayName name='foo' />);
+    expect(wrapper2).toMatchSnapshot();
+  });
+
+  it('calls the onBlur function to handle the blur event and start the validations of the component', () => {
+    const spy = jasmine.createSpy('blurred');
+    wrapper.instance().handleBlur();
+    expect(spy).not.toHaveBeenCalled();
```
dquilter: Seems unnecessary to test whether the `spy` is called _before_ you attach the `spy`?

src/utils/validations/component/validations/with-validation.spec.js
```diff
@@ -0,0 +1,383 @@
...
+
+    it('takes no action when an incorrect type with passing validations is passed to the function', async () => {
+      wrapper.setProps({ fooType: validations, value: 'foo' });
+      const spy = spyOn(wrapper.instance(), 'setState');
+      const validate = await wrapper.instance().runValidation('fooType');
+      expect(spy).not.toHaveBeenCalled();
+      expect(validate).toEqual(true);
+    });
+
+    it('it removes errors when the inputs validation passes', async () => {
```
dquilter: Double "it".

src/utils/validations/component/validations/with-validation.spec.js
edleeks87: Yeah no idea what my intention was there haha, removed

src/utils/validations/component/validations/form-with-validations.spec.js
```diff
@@ -0,0 +1,272 @@
...
+
+const MockComponent = props => <form { ...props } />;
+
+const WithNoName = withValidations(() => <form />);
+
+describe('when the formWithValidations HOC is wraps a component', () => {
+  let wrapper, context;
+
+  const Child = MockComponent;
+  const ValididationComp = withValidations(MockComponent);
```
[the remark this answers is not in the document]
edleeks87: Haha @dquilter any objections to renaming them to `ComponentWithValidations` and `ChildComponentWithValidation`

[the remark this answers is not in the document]
dquilter: Nope, sounds lovely. 👍

src/components/validations/form-with-validations.hoc.js
```diff
@@ -0,0 +1,86 @@
...
+        <ValidationsContext.Provider value={ this.getContext() }>
+          <WrappedComponent
+            validate={ this.validateRegisteredInputs }
+            validationsCount={ this.state.validationsCount }
+            warningsCount={ this.state.warningsCount }
+            infoCount={ this.state.infoCount }
+            isValidating={ this.state.formIsValidating }
+            { ...this.props }
+          >
+            errors: { this.state.validationsCount }
```
andrewjtait: dont think you meant to leave this in

src/components/validations/with-validation.hoc.js
```diff
@@ -0,0 +1,176 @@
...
+        this.context.addInput(this.props.name, this.validate);
+      }
+    }
+
+    componentWillUnmount() {
+      if (this.checkValidations(['info', 'warnings', 'validations'])) this.context.removeInput(this.props.name);
+    }
+
+    componentDidUpdate(prevProps) {
+      if (this.isUpdatedValidationProps(prevProps) && this.checkValidations(['info', 'warnings', 'validations'])) {
```
andrewjtait: Need to address this function, I will catch up with you.

PR #1949, opened 2019-03-06
Storybook Component Migration - MountInApp to Pod

stevebutler2210:
Adds stories and notes, and removes old documentation comments, for the following components:
- MountInApp
- MultiActionButton
- NavigationBar
- Number
- Pager
- Pages
- Pill
- Pod


src/components/mount-in-app/mount-in-app.stories.js
```diff
@@ -0,0 +1,51 @@
...
+          targetId='carbon-demo'
+        >
+          <div>{ children }</div>
+        </MountInApp>
+      </div>
+    );
+  }, {
+    info: {
+      text: `
+        # Mount In App component
```
samhinton88: This markdown is a little bunched up and uneasy on the eye when rendered in the story. I'm not sure we want to have a header like that below the component itself, either. This applies to all the components, but it looks worst in this one. 

image

src/components/multi-action-button/multi-action-button.stories.js
```diff
@@ -0,0 +1,58 @@
...
+      propTablesExclude: [Button]
+    }
+  })
+  .add('default', () => {
+    const disabled = boolean('disabled', false);
+    const label = text('text', 'Example Multi Action Button');
+    const as = label ? select('as', ['primary', 'secondary', 'transparent'], 'secondary') : undefined;
+    const align = select('align', OptionsHelper.alignBinary, OptionsHelper.alignBinary[0]);
+
+    return (
```
samhinton88: This shouldn't cover the component description when open.

image

src/components/navigation-bar/navigation-bar.stories.js
```diff
@@ -0,0 +1,43 @@
...
+import NavigationBar from './navigation-bar';
+import Link from '../link/link';
+import notes from './notes.md';
+
+const handleClick = (ev) => {
+  ev.target.preventDefault();
+  action('click');
+};
+
+storiesOf('Navigation Bar', module)
```
samhinton88: In the current demo site:

image

In this implementation:

image

src/components/number/number.stories.js
```diff
@@ -0,0 +1,65 @@
...
+    const labelInline = label ? boolean('labelInline', false) : undefined;
+    const labelWidth = labelInline ? text('labelWidth', '') : undefined;
+    const labelAlign = labelInline ? select(
+      'labelAlign',
+      OptionsHelper.alignBinary,
+      OptionsHelper.alignBinary[0]
+    ) : undefined;
+    const labelHelp = text('LabelHelp', 'This text provides more information for the label.');
+
+    return (
```
samhinton88: This message appears bunched up agains the top of the window.

image

src/components/pager/pager.stories.js
```diff
@@ -0,0 +1,59 @@
...
+    info: {
+      propTablesExclude: [State]
+    }
+  })
+  .add('default', () => {
+    const totalRecords = text('totalRecords', '100');
+    const pageSize = select('pageSize', OptionsHelper.pageSizes, Pager.defaultProps.pageSize);
+    const showPageSizeSelection = boolean('showPageSizeSelection', Pager.defaultProps.showPageSizeSelection);
+
+    return (
```
samhinton88: In the current demo site:

image

This implementation:

image

Whatever we put here it shouldn't be a string showing a call to `fromJS`.

src/components/pages/pages.stories.js
```diff
@@ -0,0 +1,89 @@
...
+
+
+storiesOf('Pages', module)
+  .addParameters({
+    info: {
+      propTablesExclude: [Button, DialogFullScreen, State]
+    }
+  })
+  .add('default', () => {
+    return (
```
samhinton88: I don't think we should see this.

image

src/components/pill/pill.stories.js
```diff
@@ -0,0 +1,47 @@
...
+import OptionsHelper from '../../utils/helpers/options-helper';
+
+storiesOf('Pill', module)
+  .add('default', () => {
+    const children = text('children', 'Pill');
+    const as = children ? select('as', OptionsHelper.colors, Pill.defaultProps.as) : undefined;
+    const fill = children ? boolean('fill', Pill.defaultProps.fill) : undefined;
+    const onDelete = boolean('onDelete', false);
+
+    return (
```
samhinton88: Again the component is pretty much overwhelmed by the title underneath it.

image

src/components/mount-in-app/mount-in-app.stories.js
stevebutler2210: Agree the markdown needs styling, had left as wasn't sure if we were going to be sorting out some universal styling for these further down the line, but happy to add now 👍 

I added all these description blocks off the back of the conversations in Slack yesterday, but looking now at what's actually on the existing demo site, I wonder if it might be better to just get rid of most of these bits, leaving just the component name / one-liner descriptions?

src/components/navigation-bar/navigation-bar.stories.js
stevebutler2210: If I leave the existing demo site setup in place, the component will be invisible initially (white Nav on a white background with no children). And having a text knob for `children` felt like an odd way to demo it, as not really representative of what would go in a nav bar, so I tried to go for something a little bit more conventional.

Happy to revert to what's on the current site, but it feels like re-styling what I've got here, or replacing the links with something else (a Menu maybe?) could be a better way to go?

src/components/mount-in-app/mount-in-app.stories.js
samhinton88: I like the descriptive elements in the markup, and think they're an improvement on the current demo site. It's mainly the Header and the formatting.

src/components/navigation-bar/navigation-bar.stories.js
samhinton88: I think you may be able to wrap the component in a container, so that it is visible when first loaded. 

In general, though we want to replicate the same info that's on the demo site, first and foremost we want to display the components and their functionality. Feel free to get creative, even if it means a departure from the exact implementation on the Demo site.

src/components/pages/pages.stories.js
```diff
@@ -0,0 +1,89 @@
...
+import Heading from '../heading/heading';
+import Button from '../button/button';
+
+const store = new Store({
+  open: false,
+  slideIndex: 0
+});
+
+const handleSlide = (ev) => {
+  action('slide')();
```
dquilter: If you pass the `event` as an argument, you'll be able to see it / drill down into it in the actions panel.

src/components/mount-in-app/mount-in-app.stories.js
stevebutler2210: I've updated all of these to use JSX now, and have added some styles which give the whole thing a look consistent with the StoryBook defaults (the default `storybook-info` styles differed from these)

src/components/multi-action-button/multi-action-button.stories.js
stevebutler2210: Fixed 👍

src/components/navigation-bar/navigation-bar.stories.js
stevebutler2210: Fixed, and have added a minor box-shadow so component is visible at default settings

src/components/number/number.stories.js
stevebutler2210: Fixed 👍

src/components/pager/pager.stories.js
stevebutler2210: Fixed. There was no obvious / easy way to do this, so I had to resort to a custom TableComponent implementation, as per the example here https://github.com/storybooks/storybook/tree/master/addons/info#options-and-defaults

src/components/pages/pages.stories.js
stevebutler2210: Removed 👍

src/components/pill/pill.stories.js
stevebutler2210: Fixed as per my other comment 👍

src/components/pages/pages.stories.js
stevebutler2210: Fixed 👍

.storybook/style/storybook-info-details.scss
```diff
@@ -0,0 +1,19 @@
+.storybook-info {
```
samhinton88: We're trying to remove our dependency on `scss` in favour of `styled components`, so we don't want to add more scss to the repo at this stage.

stevebutler2210: Updated now to use styled components.

I have, however, had to leave some `scss` in, where I needed to apply styles to the `#story-root` element, which we don't have direct access to any other way

PR #1950, opened 2019-03-07
Fe 1299 combine work on input decorators

grabkowski:
# Description

Combines the work from decorator rewriting tasks:

FE-1114 Rewrite Input Label Decorator
FE-1115 Rewrite Input Icon Decorator
FE-1218 Draft > Simple Character Inputs (textbox) by @samhinton88 (Rewrite Input decorator + styled-components theming solution)


src/__experimental__/components/input/input-presentation.component.js
```diff
@@ -64,7 +66,7 @@ class InputPresentation extends React.Component {
         className={ this.classNames() }
         ref={ this.container }
         onMouseDown={ this.handleMouseDown }
-        { ...props }
+        { ...filteredProps }
```
andrewjtait: Why would you want to remove these events? Shouldn't the developer be able to add them?

src/__experimental__/components/field-help/field-help.component.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * To import FieldHelp:
+ *
+ *   import FieldHelp from 'carbon-react/lib/__experimental__/components/field-help';
+ *
+ * To render FieldHelp:
+ *
+ *   <FieldHelp>This provides more information for the field.</FieldHelp>
+ */
```
dquilter: We're tending towards deleting these documentation comments from components as we're adding them in on Storybook. Don't know if we're taking that approach for the `experimental` components?

src/__experimental__/components/form-field/form-field.component.js
```diff
@@ -0,0 +1,66 @@
...
+
+FormField.defaultProps = {
+  size: 'medium'
+};
+
+FormField.propTypes = {
+  children: PropTypes.node,
+  fieldHelp: PropTypes.node,
+  label: PropTypes.node,
+  labelAlign: PropTypes.oneOf(['left', 'right']),
```
dquilter: Could use the `OptionHelper` here?

src/__experimental__/components/form-field/form-field.component.js
```diff
@@ -0,0 +1,66 @@
...
+  labelHelp: PropTypes.node,
+  labelInline: PropTypes.bool,
+  labelWidth: PropTypes.number,
+  size: PropTypes.oneOf(['small', 'medium', 'large'])
```
dquilter: Could use the OptionHelper here?

src/__experimental__/components/input-decorator-bridge/input-decorator-bridge.component.js
```diff
@@ -33,24 +31,22 @@ const InputDecoratorBridge = InputDecorator(InputLabel(InputValidation(InputIcon
...
           { this.validationHTML }
-          { this.fieldHelpHTML }
         </div>
       );
     }
   }
-))));
+));
 
 export default InputDecoratorBridge;
```
dquilter: Just a general comment, I thought we were trying to move away from the `Decorator` approach?

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,41 @@
...
+
+InputIconToggleStyle.defaultProps = {
+  size: 'medium',
+  theme: BaseTheme
+};
+
+InputIconToggleStyle.propTypes = {
+  error: PropTypes.string,
+  info: PropTypes.string,
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
```
dquilter: Could use the `OptionHelper` here?

src/__experimental__/components/input/input-presentation.spec.js
```diff
@@ -1,19 +1,80 @@
...
   });
 
-  it('renders the focus class when component has focus', () => {
-    const wrapper = shallowRender().setState({ hasFocus: true });
-    expect(wrapper
-      .find('.carbon-input-presentation')
-      .hasClass('carbon-input-presentation--has-focus')).toBeTruthy();
+  describe('style', () => {
+    describe('sizes', () => {
+      ['small', 'medium', 'large'].forEach((size) => {
```
dquilter: `OptionHelper`?

src/__experimental__/components/input/input-presentation.style.js
```diff
@@ -0,0 +1,85 @@
...
+  theme: baseTheme
+};
+
+InputPresentationStyle.propTypes = {
+  disabled: PropTypes.bool,
+  error: PropTypes.string,
+  hasFocus: PropTypes.bool,
+  info: PropTypes.string,
+  readOnly: PropTypes.bool,
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
```
dquilter: `OptionHelper`?

src/__experimental__/components/label/label.style.js
```diff
@@ -0,0 +1,46 @@
...
+  align: 'left',
+  inputSize: 'medium',
+  theme: BaseTheme,
+  width: 30
+};
+
+LabelStyle.propTypes = {
+  align: PropTypes.oneOf(['left', 'right']),
+  inline: PropTypes.bool,
+  inputSize: PropTypes.oneOf(['small', 'medium', 'large']),
```
dquilter: More `OptionHelpers`?

src/__experimental__/components/select/select.stories.js
```diff
@@ -0,0 +1,78 @@
...
+          disabled={ boolean('disabled') }
+          error={ text('error') }
+          info={ text('info') }
+          label={ text('label') }
+          labelAlign={ select('labelAlign', ['left', 'right']) }
+          labelInline={ boolean('labelInline') }
+          onChange={ (ev) => { singleSelectStore.set({ value: ev.target.value }); } }
+          placeholder={ text('placeholder') }
+          readOnly={ boolean('readOnly') }
+          size={ select('size', ['small', 'medium', 'large']) }
```
dquilter: `OptionHelpers` for `size` and `labelAlign`?

src/__experimental__/components/select/select.stories.js
```diff
@@ -0,0 +1,78 @@
...
+          disabled={ boolean('disabled') }
+          error={ text('error') }
+          info={ text('info') }
+          label={ text('label') }
+          labelAlign={ select('labelAlign', ['left', 'right']) }
+          labelInline={ boolean('labelInline') }
+          onChange={ (ev) => { multiSelectStore.set({ value: ev.target.value }); } }
+          placeholder={ text('placeholder') }
+          readOnly={ boolean('readOnly') }
+          size={ select('size', ['small', 'medium', 'large']) }
```
dquilter: `OptionHelpers` again?

src/__experimental__/components/textbox/textbox.stories.js
```diff
@@ -0,0 +1,58 @@
...
+        labelHelp={ text('labelHelp') }
+        labelInline={ boolean('labelInline') }
+        placeholder={ text('placeholder') }
+        readOnly={ boolean('readOnly') }
+        size={ select('size', ['small', 'medium', 'large']) }
+        warning={ text('warning') }
+        key='1'
+      />
+    ]);
+  });
```
dquilter: Still banging the drum for various `OptionHelpers`.

src/__experimental__/components/field-help/field-help.component.js
andrewjtait: Are we? Where are they being stored instead? There is some benefit in keeping them for IDE purposes, for example VSCode will show these comments for a given class

Screenshot 2019-03-14 at 09 15 48

src/__experimental__/components/form-field/form-field.component.js
andrewjtait: Yeah I will look into those. TBH I think we can do better with the PropTypes, in terms of sharing them across components that need them. For example Textbox does not have many props, but it is composed of many components that have other props, maybe those should be shared and added to the Textbox PropTypes.

src/__experimental__/components/input-decorator-bridge/input-decorator-bridge.component.js
andrewjtait: We are - this is a temporary bridge between the new and the old. https://github.com/Sage/carbon/blob/0db0e55c03dff9c8c865dd6ec87ae52ecfcb529a/src/__experimental__/components/input-decorator-bridge/input-decorator-bridge.component.js#L8-L9

src/__experimental__/components/field-help/field-help.component.js
andrewjtait: Though to be honest the import info is probably less relevant considering you've already imported it!

dquilter: I think that's what we were thinking with a lot of those docs when we were looking at them last week. It's rare that there's any really useful information in there.

Where there has been, we've been adding them to Storybook using the Info AddOn. @stevebutler2210 has done a load of work styling them too...

Screenshot 2019-03-14 at 09 27 59

src/style/themes/base/input.config.js
```diff
@@ -0,0 +1,38 @@
...
+    borderWidth: '2px'
+  },
+  fieldHelp: {
+    color: 'inherit',
+    marginSide: '0px',
+    marginTop: '8px'
+  },
+  label: {
+    marginSide: '0px'
+  }
```
grabkowski: I'm afraid we could end up with clogging the theme configs with component specific properties. Maybe we could think of an approach with placing that styles in appropriate components.

src/__experimental__/components/input/input-presentation-classic.style.js
```diff
@@ -0,0 +1,25 @@
+import { css } from 'styled-components';
```
samhinton88: Could probably do with some commentary on why we're extracting out `classic` properties in this way.

src/__experimental__/components/input/input-presentation.style.js
```diff
@@ -0,0 +1,68 @@
...
+    border-color: transparent !important;
+  `}
+  ${({ hasFocus, theme }) => hasFocus && css`
+    && { outline: 3px solid ${theme.colors.focus}; }
+  `}
+  ${stylingForValidation('info')}
+  ${stylingForValidation('warning')}
+  ${stylingForValidation('error')}
+
+  ${inputClassicStyling}
```
samhinton88: Will this end up printing a whole bunch of component-specific css to the browser (namely the rules generated above this line), that is then overridden? Do we care about that?

src/__experimental__/components/input/input-sizes.style.js
```diff
@@ -0,0 +1,14 @@
...
+  },
+  medium: {
+    height: '40px',
+    padding: '11px'
+  },
+  large: {
+    height: '48px',
+    padding: '13px'
+  }
+};
```
grabkowski: we could specify size styles as strings:
```javascript
{
  small: `
    min-height: 32px;
    padding-left: 8px;
    padding-right: 8px;
  `,
  medium: `
    min-height: 40px;
    padding-left: 11px;
    padding-right: 11px;
  `,
  large: `
    min-height: 48px;
    padding-left: 13px;
    padding-right: 13px;
  `
};
```
and then just insert whole blocks with ${({ size }) => sizes[size]}

src/__experimental__/components/input/input-presentation.style.js
grabkowski: that will happen only when a theme is set to `classic`

andrewjtait: With classic theme enabled, it will print both, with the latter overriding. It's a minor trade-off I suppose for it being a temporary theme.

src/__experimental__/components/input/input-sizes.style.js
andrewjtait: In this case they need to remain as JSON format, as I am using these config values in other components (eg. label) which reflects the same padding values.

https://github.com/Sage/carbon/blob/f986289744f8ea2678821746e296529f29814c83/src/__experimental__/components/label/label.style.js#L19

src/__experimental__/components/input/input-presentation-classic.style.js
andrewjtait: I think it's best we add this detail to the wiki on the repo, a segment on theming and how to handle classic theming. 

I plan on adding some detail to that over the next few days.

src/__experimental__/components/input/input-presentation.style.js
samhinton88: Fully aware that it'll only happen when theme is set to classic. Given that most of our components in the near future will be using classic theme, does it seem as minor when you consider that there is likely to be a majority of components rendering duplicate `css`, rather than just all of the inputs rendered on a page? 

Something like:
```
${(props) => {
  if(props.theme.name === 'classic') return inputClassicStyling(props):
  return css`/* non classic css */`;
}
```

src/__experimental__/components/input/input-sizes.style.js
grabkowski: we could import Label style into InputPresentation style and add that attributes here with `& ${Label}`. That will allow us to use the label separately without the Input component.

src/__experimental__/components/input/input-presentation.style.js
andrewjtait: The problem here is that the overrides are only over _some_ of the attributes, as few changes as possible, while still relying on the original styling. In most cases this will be the same (some padding, colours, or sizing slightly different). The DLS is not that far different from what classic is today. In some of the components I have done a "if classic else this" (such as here https://github.com/Sage/carbon/blob/f986289744f8ea2678821746e296529f29814c83/src/__experimental__/components/form-field/form-field.style.js#L9 ), but for InputPresentation in particular it seemed like an easier trade off to drop it in a single block. 

It may be a case of using both approaches, whichever makes more sense given the context?

src/__experimental__/components/form-field/index.js
```diff
@@ -1,3 +1 @@
-import FormField from './form-field.component';
-
-export default FormField;
+export { default } from './form-field.component';
```
dquilter: This is blowing my mind a little bit, didn't think to write a default `export` that way... But that's by the by.

Are these single line files necessary? Would it not make more sense to export directly from the component file?

andrewjtait: It is to aid imports being less verbose (`import 'form-field'` vs `form-field/form-field`) and in aid of IDEs searching for files and displaying filenames in tabs etc, rather than everything being in an index file.

dquilter: 👍 No problem. I can see it's use for when you have lots of exports. It was a super-opinionated comment.

src/__experimental__/components/select/select.component.js
```diff
@@ -51,6 +51,8 @@ class Select extends React.Component {
     ])
   }
 
+  static defaultProps = {}
```
dquilter: Possibly could do with a comment to indicate why empty `defaultProps` is included. (I think this was for Storybook?)

src/__experimental__/components/textbox/textbox.component.js
```diff
@@ -1,24 +1,37 @@
...
+  );
+};
 
 Textbox.propTypes = {
   children: PropTypes.node,
+  inputIcon: PropTypes.string,
   leftChildren: PropTypes.node
 };
 
+Textbox.defaultProps = {};
```
dquilter: Another comment for empty `defaultProps`?

src/__experimental__/components/input/input-presentation.component.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
```
dawidzarzycki: What do you think about to move every file related with `input-presentation` to external folder? It's going to make it more readable

andrewjtait: We are trying a flatter structure for directories, see the wiki https://github.com/Sage/carbon/wiki/Codebase-Conventions#multiple-components

We will keep it as it is for now, but if we find it gets unmanageable we can update later.

src/__experimental__/components/input/input-presentation.style.js
```diff
@@ -0,0 +1,68 @@
+import styled, { css } from 'styled-components';
+import PropTypes from 'prop-types';
+import baseTheme from '../../../style/themes/base';
+import OptionsHelper from '../../../utils/helpers/options-helper';
+import sizes from './input-sizes.style';
+import inputClassicStyling from './input-presentation-classic.style';
+
+const InputPresentationStyle = styled.div`
```
dawidzarzycki: Just thought... What do you think about split the code on two parts 'hardcode' and 'variable based code'. For now we have something like mixing variables with hardcoded values, what do you think about this solution?
```
const InputPresentationStyle = styled.div`
  align-items: center;
  background: transparent;
  box-sizing: border-box;
  cursor: text;
  display: flex;
  flex-wrap: wrap;
  flex: 1;
  margin: 0px;
  width: 100%;

  ${({
    theme, size, disabled, readOnly, hasFocus
  }) => css`
    border: 1px solid ${theme.colors.border};
    min-height: ${sizes[size].height};
    padding-left: ${sizes[size].padding};
    padding-right: ${sizes[size].padding};

    ${disabled
      && css`
        background: ${theme.disabled.input};
        border-color: ${theme.disabled.border};
        cursor: not-allowed;
      `}

    ${readOnly
      && css`
        background: transparent !important;
        border-color: transparent !important;
      `}

    ${hasFocus
      && css`
        && {
          outline: 3px solid ${theme.colors.focus};
        }
      `}
  `}

  ${stylingForValidation('info')}
  ${stylingForValidation('warning')}
  ${stylingForValidation('error')}

  ${inputClassicStyling}
`;
```

andrewjtait: I like this, but it might get more complicated when we start trying to follow it for components that have nested selectors or pseudo selectors? It may be a convention that is too restrictive or difficult to follow in those cases?

dawidzarzycki: Yup, you're right. It's can be a problem

.storybook/__helpers__/components/app-layout/app-layout.js
```diff
@@ -34,17 +38,12 @@ class AppLayout extends React.Component {
   }
 
   state = {
-    activeTheme: 'theme1',
+    activeTheme: 'classic',
```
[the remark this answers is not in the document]
andrewjtait: done 👍

.storybook/__helpers__/components/app-layout/app-layout.js
```diff
@@ -59,8 +58,10 @@ class AppLayout extends React.Component {
         <div>
           <SelectionContainer>
             <StyledSelect value={ activeTheme } onChange={ this.handleSelectChange }>
-              <option value='theme1'>Theme 1</option>
-              <option value='theme2'>Theme 2</option>
+              <option value='classic'>Classic</option>
+              <option value='small'>Small</option>
+              <option value='medium'>Medium</option>
+              <option value='large'>Large</option>
```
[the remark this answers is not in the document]
andrewjtait: done 👍

src/__experimental__/components/form-field/form-field.style.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components';
+import BaseTheme from '../../../style/themes/base';
+
+const FormFieldStyle = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+
+  & + & {
+    margin-top: ${({ theme }) => (theme.name === 'classic' ? '10px' : '16px')};
```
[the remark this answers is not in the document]
andrewjtait: done 👍

src/__experimental__/components/input/input-presentation-classic.style.js
```diff
@@ -0,0 +1,25 @@
...
+  ${hasFocus && css`
+    && {
+      border: 1px solid #255bc7;
+      outline: none;
+    }
+  `}
+
+  ${disabled && css`
+    background: #d9e0e4;
+    border-color: #d9e0e4 !important;
```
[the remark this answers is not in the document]
andrewjtait: We need this otherwise error border colors leak through, I think this is a valid use case for `!important`.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,43 @@
...
+  `}
+  ${({ size }) => {
+    if (size === 'small') return css`width: 32px;`;
+    if (size === 'large') return css`width: 48px;`;
+    return css`width: 40px;`;
+  }}
+`;
+
+InputIconToggleStyle.defaultProps = {
+  size: 'medium',
```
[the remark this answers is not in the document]
andrewjtait: Not sure how we could work out the default value from the options helper.

src/__experimental__/components/input-icon-toggle/input-icon-toggle.style.js
```diff
@@ -0,0 +1,43 @@
...
+  `}
+  ${({ warning, theme }) => warning && css`
+    color: ${theme.colors.warning};
+  `}
+  ${({ info, theme }) => info && css`
+    color: ${theme.colors.info};
+  `}
+  ${({ size }) => {
+    if (size === 'small') return css`width: 32px;`;
+    if (size === 'large') return css`width: 48px;`;
```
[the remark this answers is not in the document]
andrewjtait: We need to relate specific options to specific values, not sure how we can use the options helper to do that.

src/__experimental__/components/form-field/form-field.component.js
```diff
@@ -0,0 +1,56 @@
...
+    { fieldHelp && (
+      <FieldHelp labelInline={ labelInline } labelWidth={ labelWidth }>
+        { fieldHelp }
+      </FieldHelp>
+    ) }
+  </FormFieldStyle>
+);
+
+FormField.defaultProps = {
+  size: 'medium'
```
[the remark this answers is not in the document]
andrewjtait: Not sure how we could work out the default value from the options helper.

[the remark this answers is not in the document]
customcodepl: Ok, I forgot there is an array. We can use `OptionsHelper. sizesRestricted[1]`, but it will be not readable. It will be nice to have something like enums but it is not necessary not now.

src/__experimental__/components/input/input-presentation-classic.style.js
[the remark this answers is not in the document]
customcodepl: Ok there is an inline style so you are right.

PR #1951, opened 2019-03-07
FE-1303 - update storybook to 5.0.3

customcodepl:
# Description
Update storybook to v5.0.3. Adds a11y.

I reported an issue related to a11y addon.
https://github.com/storybooks/storybook/issues/5945

 

I investigated a11y addon and there is a method getElement https://github.com/storybooks/storybook/blob/master/addons/a11y/src/index.js#L14:
```
const getElement = () => {
  const storyRoot = document.getElementById('story-root');

  if (storyRoot) {
    return storyRoot.children;
  }
  return document.getElementById('root');
};
```
So for now we can't check elements rendered in portals like modals.

There are a few options for a11y - https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter.

It's possible to change default prop table. We can define custom PropTable based on https://github.com/storybooks/storybook/blob/master/addons/info/src/components/PropTable.js and using TableComponent parameter for addon-info.

# TODO
`package-lock.json` is for verification.

# Screenshots / Gifs
zrzut ekranu 2019-03-7 o 16 37 31

# Testing Instructions
`npm install`
`npm run storybook`


.storybook/config.js
```diff
@@ -1,11 +1,10 @@
 import React from 'react';
-import { configure, addDecorator } from '@storybook/react';
+import { configure, addDecorator, addParameters } from '@storybook/react';
 import { withKnobs } from '@storybook/addon-knobs';
-import { withNotes } from '@storybook/addon-notes';
```
dquilter: I think we're using this addon to display the Design Notes, does another plugin replace it? Or are you planning on replacing it with anything?

customcodepl: Hi Dave, I'm not removing this addon. It doesn't need to be imported here anymore.

andrewjtait:
We will merge this and run QA from master, so everyone is working from the same base.

PR #1959, opened 2019-03-13
BUG- FE-1302 - pager total records errors

stevebutler2210:
Fix for the following issues:

1) I can put every character to totalRecords input field
2) When I put negative number to totalRecords input field, then component show negative pages number

# Description:
- changes totalRecords propType to number, to prevent alpha input
- adds totalRecordsCount property, which defaults to zero when totalRecords prop is negative
- uses this new property for the maxPage calc, and in place of the existing totalRecords display
- adds specs to cover scenarios where totalRecords is negative or zero


src/components/pager/pager.stories.js
```diff
@@ -68,7 +68,7 @@ storiesOf('Pager', module)
     }
   })
   .add('default', () => {
-    const totalRecords = text('totalRecords', '100');
+    const totalRecords = number('totalRecords', '100');
```
dquilter: Presumably this works, but should 100 not be a number?

stevebutler2210: Agreed, fixed 👍

PR #1960, opened 2019-03-14
Cypress cucumber storybook

marcinstachowicz:
### Automation tests for Storybook components

**Tools and technologies:**
Cucumber with Gherkin syntax for easy to read test scenarios
Cypress for automation

**Scope:**
- Alert component
- Animated Menu Button component
- App Wrapper component
- Button component
- Button Toggle Group component
...and more for next sprints


samhinton88:
I'd want to see a reasonable description on any PR, especially on work which introduces something conceptually new.

cypress/integration/Alert.feature
```diff
@@ -0,0 +1,131 @@
+Feature: Alert component
```
samhinton88: The length of this file might be a readability issue, is there a way to split it into separate files?

cypress/support/step_definitions/buttonToggleGroupSteps.js
```diff
@@ -0,0 +1,85 @@
...
+When('I check label inline checkbox', () => {
+  lableInlineCheckbox().check()
+})
+
+When('I uncheck label inline checkbox', () => {
+  lableInlineCheckbox().uncheck()
+})
+
+Then('Button Toggle Group component has label-inline property', () => {
+  buttonToggleGroupPreview().should('have.class', 'common-input--label-inline')
```
samhinton88: We are working to remove classNames from our markup, should we be writing tests for our new components rather than the old ones?

cypress/integration/Alert.feature
sianford: I don't think it makes sense to split the file when all the scenarios relate to the same component. Is 131 lines excessive? We've ran this file several times without issue.

samhinton88: Running the file wont be an issue, but if we expect this file to grow at all it'll become more difficult to work with. Our specs are in general too long, so I thought I'd ask the question.

sianford: Ok, it might be wise to do this at this point then with DLS-related changes potentially causing files to increase.

marcinstachowicz: I can always split file, but in this case it provides inconsistency for cucumber scenarios in general. If Alert feature will grow then we can think about splitting scenarios. In my opinion 130 lines feature file is ok. I believe we have coding standards for java script so for cucumber feature files we can agree coding standards too

cypress/support/step_definitions/buttonToggleGroupSteps.js
marcinstachowicz: This test concerns Toogle Group Storybook Component that has been already developed and merged to master. If in future classNames will be removed then test will be updated with those changes

marcinstachowicz:
> I'd want to see a reasonable description on any PR, especially on work which introduces something conceptually new.

Thank you Sam, I add description, hope it is reasonable

cypress/support/step_definitions/commonSteps.js
```diff
@@ -0,0 +1,42 @@
...
+})
+
+When('I set label to {string}', (label) => {
+  labelInput().clear().type(label)
+})
+
+
+
+
+
```
dquilter: Lots of unneeded newlines?

marcinstachowicz: fixed

package.json
```diff
@@ -61,8 +63,14 @@
     "core-js": "^2.6.3",
     "cpx": "^1.5.0",
     "cross-env": "^5.2.0",
+    "cypress": "^3.2.0",
+    "cypress-cucumber-preprocessor": "^1.11.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "1.7.1",
+    "eslint": "^5.3.0",
+    "eslint-config-airbnb-base": "^13.1.0",
```
andrewjtait: you shouldn't need eslint, eslint-plugin-import or eslint-config-airbnb-base in this package.json as they are already installed via carbon-factory

cypress/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
andrewjtait: are we using typescript here? is this file actually use?

cypress/support/step_definitions/commonSteps.js
```diff
@@ -0,0 +1,37 @@
+import { asSelect, themeSelect, sizeSelect, subtextInput, heightInput, childrenTextArea, labelInput, labelHelpInput } from "../../locators/commonLocators";
+
+When('I set children to {string}', (text) => {
```
andrewjtait: I'm unclear what these steps are used for?

package.json
marcinstachowicz: fixed

package.json
```diff
@@ -61,8 +63,11 @@
     "core-js": "^2.6.3",
     "cpx": "^1.5.0",
     "cross-env": "^5.2.0",
+    "cypress": "^3.2.0",
+    "cypress-cucumber-preprocessor": "^1.11.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "1.7.1",
+    "eslint-plugin-cypress": "^2.0.1",
```
andrewjtait: the package-lock should probably have updated after you ran `npm install` -  can you commit that as well please? Make sure you're on at least version `6.8.0` of `npm` first otherwise you will get some discrepancies.

cypress/support/step_definitions/commonSteps.js
marcinstachowicz: This step set component field called 'children'
image

package.json
marcinstachowicz: package-lock.json commited after npm install (6.8.0)

cypress/support/step_definitions/commonSteps.js
andrewjtait: I meant I was unclear how these are used, but I've looked into it a bit more and think I understand now. It matches the sentence against the feature spec.

cypress/tsconfig.json
marcinstachowicz: this is for suggestion during coding
https://docs.cypress.io/guides/tooling/intelligent-code-completion.html#Set-up-in-your-Dev-Environment

cypress/support/step_definitions/buttonToggleGroupSteps.js
marcinstachowicz: I isolate Given steps to Background to reduce codelines

cypress/integration/Alert.feature
marcinstachowicz: I isolate Given steps to Background to reduce lines of code

PR #1964, opened 2019-03-19
fixes bug with event listeners on dialog

andrewjtait:
Event listeners were not being set up or torn down correctly if the dialog component was mounted in an open state.

Also includes the fixes for these:
https://github.com/Sage/carbon/pull/1965
https://github.com/Sage/carbon/pull/1966/


src/components/sidebar/__spec__.js
```diff
@@ -28,66 +28,6 @@ describe('Sidebar', () => {
...
-      });
-
-      it('removes event listeners for resize and closing', () => {
-        spyOn(mockWindow, 'removeEventListener');
-        wrapper.setProps({ title: 'Remove event handlers' });
-        expect(mockWindow.removeEventListener.calls.count()).toEqual(1);
-        expect(mockWindow.removeEventListener).toHaveBeenCalledWith('keyup', instance.closeModal);
-      });
-    });
-  });
```
andrewjtait: removed this test as it was a duplicate of what was already tested in dialog

PR #1970, opened 2019-03-20
Fe 1280 decimal component

dquilter:
Adds a new experimental `Decimal` component based on the existing implementation (includes specs and Storybook). As per new Decimal props, adds additional styling to `input` and `field-help` experimental components.

Screenshot 2019-03-28 at 08 31 07

Includes new script in package.json to update jest snapshots.


src/__experimental__/components/field-help/field-help.style.js
```diff
@@ -9,20 +9,20 @@ const FieldHelpStyle = styled.span`
   white-space: pre-wrap;
   width: 100%;
 
-  ${({ labelInline, inputWidth }) => labelInline && css`
+  ${({ labelInline, labelWidth }) => labelInline && css`
```
grabkowski: field help is a component directly below the input and it should have the same width as that input

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -0,0 +1,43 @@
...
+import OptionsHelper from '../../../utils/helpers/options-helper';
+
+storiesOf('Experimental/Decimal', module)
+  .add('default', () => {
+    const align = select(
+      'align',
+      OptionsHelper.alignBinary,
+      Decimal.defaultProps.align
+    );
+    const precision = number('precision', Decimal.defaultProps.precision);
```
izakrzewska: is the precision prop working correctly? it looks like changing the knob in the storybook makes no difference

src/__experimental__/components/decimal/decimal.style.js
```diff
@@ -0,0 +1,18 @@
+import styled, { css } from 'styled-components';
```
izakrzewska: is css import necessary?

src/__experimental__/components/decimal/decimal.stories.js
dquilter: It definitely wasn't, I've only finished writing it today. 😆

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,107 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import TextBox from '../textbox';
```
andrewjtait: `Textbox` rather than `TextBox`

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+    setDecimalValue(array.join(''));
+  }
+
+  return (
+    <TextBox
+      {...props}
+      onChange={ handleChange }
+      onBlur={ handleBlur }
+      value={ decimalValue }
```
andrewjtait: this looks like the value is driven by state rather than by props (eg. if i change props.value will it update?) I know the original used state too, but it listened for changes to prop.value and updated accordingly. Maybe there is a way to do it without state at all? eg. derived data

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,107 @@
...
+    const integer = parts[0];
+    
+    let newInteger = integer.replace(/,/g, '');
+    newInteger = newInteger.replace(addCommasRegex, '$1,');
+    let array = [newInteger];
+
+    if (parts[1]) {
+      array.push('.', parts[1]);
+    }
+    setDecimalValue(array.join(''));
```
andrewjtait: should you not just use the `formatDecimal` helper here?

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,107 @@
...
+      diff = precision - decimalLength;
+      for (let i = 0; i < diff; i++) {
+        newDecimal = newDecimal + '0';
+      }
+    } else if (precision < decimalLength) { // Precision decreased
+      diff = decimalLength - precision;
+      newDecimal = Math.round(newDecimal / (10 * diff));
+    }
+    
+    setDecimalValue([parts[0], '.', newDecimal].join(''));
```
andrewjtait: should you not just use the formatDecimal helper here?

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+  inputWidth: PropTypes.number,
+  /**
+   * The value of the input
+   */
+  value: PropTypes.string
+}
+
+Decimal.defaultProps = {
+  align: 'left',
```
andrewjtait: this should default to right

src/__experimental__/components/input/input.style.js
```diff
@@ -9,7 +9,8 @@ const StyledInput = styled.input`
   flex-grow: 1;
   font-size: ${({ theme }) => theme.text.size};
   outline: none;
-  width: 30px;
+  text-align: ${({ align }) => align}
```
andrewjtait: align prop needs a default value otherwise this is breaking the css for other components

dquilter: I've changed this slightly this morning, so `text-align` is only used if the `align` prop is used.

src/__experimental__/components/decimal/decimal.component.js
dquilter: Lovely, thanks, I hadn't spotted that.

src/__experimental__/components/decimal/decimal.component.js
dquilter: I've ended up sticking with storing the value in `state`. I couldn't figure out a way to do this with derived data, as there were too many instances where I needed to change the value in different ways (`value` in `onChange`, delimiters in `onBlur` and the length of the decimal when `precision` changes).

Using `state` also means that I always have the previous value easily available if a new value fails validation.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,132 @@
...
+      });
+    }
+  }
+
+  handleBlur = () => {
+    const { precision } = this.props;
+    const strippedDelimiters = this.stripDelimiters();
+
+    this.setState({
+      decimalValue: I18nHelper.formatDecimal(parseFloat(strippedDelimiters), precision)
```
andrewjtait: maybe abstract this into its own function as it used in exactly the same way in `handleBlur` and in `componentDidUpdate`.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,132 @@
...
+import Textbox from '../textbox';
+import I18nHelper from '../../../utils/helpers/i18n';
+import Logger from '../../../utils/logger';
+
+class Decimal extends React.Component {
+  constructor(props) {
+    super(props);
+    const { value } = this.props;
+    const isValidInitVal = this.testValue(value);
+    const initVal = isValidInitVal ? value : Decimal.defaultProps.value;
```
andrewjtait: should we fallback to the default if they're providing a dodgy value? or throw an error? if so this could be prop type validation

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,132 @@
...
+
+    if (!isValidInitVal) {
+      Logger.warn('Initial value is not a valid decimal');
+    }
+
+    this.state = {
+      decimalValue: initVal
```
andrewjtait: is `initVal` formatted?

samhinton88:
Pressing a non-numerical seems to skip the cursor to the end of the input.

decimal-jump-to-end

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,123 @@
...
+
+    if (precision > 15) {
+      Logger.warn('Precision cannot be greater than 15');
+      return 15;
+    }
+
+    return precision;
+  }
+
+  testValue = (value) => {
```
grabkowski: I think that the name is too general, could you rename this method to be more specific? Maybe `isValidDecimal`?

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,123 @@
...
+    const { precision } = this.props;
+    const format = I18nHelper.format();
+    const delimiter = `\\${format.delimiter}`;
+    const seperator = `\\${format.separator}`;
+    const isGoodDecimal = new RegExp(`^[\\d${delimiter}]*[${seperator}{1}]?\\d{0,${precision}}?$`);
```
grabkowski: that const name suggests it's a boolean

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,123 @@
...
+    // Return unformatted value if input is being used
+    if (input && this._document.activeElement === input.current) {
+      return value;
+    }
+
+    // Only format value if input is not active
+    // Strip delimiters otherwise formatDecimal Helper goes nuts
+    const format = I18nHelper.format();
+    const delimiter = `\\${format.delimiter}`;
+    const removeDelimiter = new RegExp(`[${delimiter}]*`, 'g');
```
grabkowski: that name suggests it's a method

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    <Decimal
+      onChange={ onChange }
+      value={ props.value }
+      precision={ props.precision }
+    />
+  );
+}
+
+function renderIntoDocument(props) {
+  return TestUtils.renderIntoDocument(
```
grabkowski: is there a specific reason to use this method?

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    <Decimal value={ props.value } precision={ props.precision } />
+  );
+}
+
+describe('Decimal', () => {
+  describe('Input validation', () => {
+    it('renders the correct value', () => {
+      const input = render({ value: '9.87' });
```
grabkowski: I think that should be `wrapper` as `input` suggests it's an input element

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    it('renders a value from props exactly even when invalid', () => {
+      const input = render({ value: '12abc.85' });
+      expect(input.find(Textbox).prop('value')).toEqual('12abc.85');
+    });
+
+    it('updates an external state onChange', () => {
+      const onChange = jest.fn();
+      const input = render({ onChange });
+      input.instance().onChange({ target: { value: '14.79' } });
+      expect(onChange).toHaveBeenCalledWith({ target: { value: '14.79' } });
```
grabkowski: you should trigger that event on an input instance instead of a decimal wrapper. Right now it will always be true.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      const instance = renderIntoDocument({ value: '99.99', precision: 4 });
+      expect(instance.formatValue()).toEqual('99.9900');
+    });
+
+    it('updates the value after decreasing the precison', () => {
+      const instance = renderIntoDocument({ value: '234.1234567', precision: 4 });
+      expect(instance.formatValue()).toEqual('234.1235');
+    });
+
+    it('does not allow the precison to be greater than 15', () => {
```
grabkowski: than the default value?

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      const instance = renderIntoDocument({ value: '4.1234', precision: 20 });
```
grabkowski: precision could be taken from the default value

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    it('forces change event if user blurs', () => {
+      const onChange = jest.fn();
+      const input = render({ onChange });
+      input.instance().onBlur({ target: { value: '1' } });
+      expect(onChange).toHaveBeenCalled();
+    });
+
+    it('formats with delimiters with input is not active', () => {
+      const instance = renderIntoDocument({ value: '1234567.00' });
+      expect(instance.formatValue()).toEqual('1,234,567.00');
```
samhinton88: You repeat `expect(instance.formatValue()).toEqual(/* */)` a lot here. You could put it in a helper function. 

Also, it may not be completely clear what the instance method `formatValue` actually means in the context of the component.

grabkowski: you could check that directly on the input

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -0,0 +1,65 @@
...
+        <Decimal
+          align={ align }
+          precision={ precision }
+          inputWidth={ inputWidth }
+          fieldHelp={ fieldHelp }
+          label={ label }
+          labelInline={ labelInline }
+          labelWidth={ labelWidth }
+          labelAlign={ labelAlign }
+          labelHelp={ labelHelp }
```
grabkowski: you could just import `Textbox` props from the `Textbox` story:
`{ ...getTextboxStoryProps() }`

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -0,0 +1,65 @@
...
+
+const setValue = (evt) => {
+  action('onChange')(evt);
+  store.set({ value: evt.target.value });
+};
+
+storiesOf('Experimental/Decimal', module)
+  .addParameters({
+    info: {
+      propTablesExclude: [State]
```
grabkowski: you could also add the `Textbox` props from the `Textbox` component

src/__experimental__/components/label/__snapshots__/label.spec.js.snap
```diff
@@ -12,7 +12,6 @@ exports[`Label renders the label 1`] = `
 <label
   className="c0"
   data-element="label"
-  width={30}
```
andrewjtait: we need this to act as a min width

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,123 @@
...
+  onChange = (evt) => {
+    const { target } = evt;
+    const { value, selectionEnd } = evt.target;
+    const isValid = this.testValue(value);
+
+    if (isValid) {
+      this.props.onChange(evt);
+
+      setTimeout(() => {
+        target.setSelectionRange(selectionEnd, selectionEnd);
```
andrewjtait: I'm not sure this is the correct place for this to happen. Surely the cursor is in the correct position while active and valid.

Looking at the old implementation it only sets the cursor if the value is invalid, in which it replaces the value with the previously valid value https://github.com/Sage/carbon/blob/master/src/components/decimal/decimal.js#L153-L161

Currently in your implementation if you try it in the browser and type really fast it breaks the cursor position.

andrewjtait: Think it should be
```
    if (isValid) {
      this.props.onChange(evt);
    } else {
      target.setSelectionRange(selectionEnd, selectionEnd);
    }
```

src/__experimental__/components/decimal/decimal.component.js
dquilter: I think this name is good, as it gives an indication as to how the regex is going to be used.

src/__experimental__/components/decimal/decimal.component.js
dquilter: Again, I'm happy with this name, as it gives an indication as to how the regex is going to be used.

src/__experimental__/components/decimal/decimal.spec.js
dquilter: The `formatValue` method checks for the `activeElement` in the `document` property. I was having issues with this test using `shallow` and `mount`, but they were resolved when I started using `renderIntoDocument`, which I took from the specs for the existing `Decimal` component.

src/__experimental__/components/decimal/decimal.component.js
samhinton88: I'm inclined to agree with Kuba, something like `delimiterMatcher` is much more intuitive. 

There's nothing inherent in that regular expression that connects it to deletion, it's just a matcher.

src/__experimental__/components/decimal/decimal.component.js
samhinton88: `validDecimalMatcher` or even `validDecimalInput` is much more intuitive here. We should try and stick to general naming conventions as much as possible.

src/__experimental__/components/decimal/decimal.spec.js
dquilter: Not sure what you mean here?

src/__experimental__/components/decimal/decimal.spec.js
dquilter: I don't think it can, there's possibly a misunderstanding here?

The default value for precision is `2`. The maximum precision allowed is `15`.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+describe('Decimal', () => {
+  describe('Input validation', () => {
+    it('renders the correct value', () => {
+      const input = render({ value: '9.87' });
+      expect(input.find(Textbox).prop('value')).toEqual('9.87');
+    });
+
+    it('renders a value from props exactly even when invalid', () => {
```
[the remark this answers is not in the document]
dquilter: There's a whole bunch of formatting that happens when a value is entered (removing unwanted characters, putting the delimiters in the right place, ensuring that the precision is correct). This only happens on `blur` though (as per existing component - constantly formatting the number is quite expensive).

There's also the possibility of the component receiving an invalid value via props. As per @andrewjtait , I've written the component so that invalid values are shown in all their glory, rather than giving an inaccurate indication of the value that'd been provided.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    it('updates an external state onChange', () => {
+      const onChange = jest.fn();
+      const input = render({ onChange });
+      input.instance().onChange({ target: { value: '14.79' } });
+      expect(onChange).toHaveBeenCalledWith({ target: { value: '14.79' } });
+    });
+
+    it('does not allow the user to enter letters or special characters', () => {
+      const input = render({ value: '12.34' });
+      input.instance().onChange({ target: { value: '1hello$1.27' } });
```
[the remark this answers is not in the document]
dquilter: Do we have any existing helpers or documentation for what characters we should be testing?

src/__experimental__/components/decimal/decimal.spec.js
dquilter: Added in a couple of helper functions to remove duplicated `expect`s.

I think it's fairly clear what `formatValue` is doing, happy to take suggestions though?

src/__experimental__/components/decimal/decimal.spec.js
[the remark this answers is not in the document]
samhinton88: I don't believe so, feel free to create some. 

Otherwise `!@£$%^&*()#_+=}{|":;'\<>?`
`qwertyuiopasdfghjklzxcvbnm` (with toUpperCase) should be fine. 

It's just that, if we're going to test for special chars, there's no point only checking for one.

src/__experimental__/components/decimal/decimal.spec.js
dquilter: Is there any reason to? The formatting happens in the Decimal component; I don't want to test that the value is being passed down, just that it's being formatted correctly.

src/__experimental__/components/decimal/decimal.stories.js
dquilter: I'm not sure I want to do I? Those variables are used to enable the Storybook knobs...

Maybe I'm misunderstanding?

src/__experimental__/components/decimal/decimal.stories.js
dquilter: I don't think I need to? The `Textbox` props aren't being shown by the `Info` addon?

src/__experimental__/components/decimal/decimal.spec.js
samhinton88: It's more that this way of testing exposes implementation detail. Our assertions shouldn't rely on the specific names of methods, but rather their results.

src/__experimental__/components/label/__snapshots__/label.spec.js.snap
dquilter: I don't think it does?

Screenshot 2019-04-01 at 14 42 18

The label goes full-width if I change `width` to `min-width`?

andrewjtait: oh yeah, my mistake 👍

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,124 @@
...
+    return I18nHelper.formatDecimal(
+      noDelimiters,
+      this.validatePrecision()
+    );
+  }
+
+  validatePrecision = () => {
+    const { precision } = this.props;
+
+    if (precision > 15) {
```
grabkowski: we shouldn't have magic numbers in our code. That should be defined as a const.

src/__experimental__/components/decimal/decimal.spec.js
grabkowski: my bad, I've mixed up the default precision with max precision

src/__experimental__/components/decimal/decimal.spec.js
grabkowski: my bad, I've mixed up that with the default precision. On the other hand, we shouldn't use magic numbers and I think that number should be specified in the tested component and then used in the test.

src/__experimental__/components/decimal/decimal.stories.js
grabkowski: yes, but they're specified in the `Decimal` component and then passed to the `Textbox` component, so I think they should be displayed for the documentation purpose.

src/__experimental__/components/decimal/decimal.stories.js
grabkowski: inputWidth, fieldHelp and label knobs are already specified in the Textbox story, so instead of repeating it we could reuse them.

src/__experimental__/components/decimal/decimal.spec.js
grabkowski: I can't see a reason it shouldn't work on the enzyme `shallow` render

src/__experimental__/components/decimal/decimal.stories.js
grabkowski: an example:
https://github.com/Sage/carbon/blob/master/src/__experimental__/components/number/number.stories.js#L35

src/__experimental__/components/decimal/decimal.spec.js
dquilter: I had an issue with changing `document` when using `shallow`:

```Cannot set property activeElement of [object Document] ```

Which kind of made sense to me. I'm sure there's some hack I could use to get this working, but using `renderIntoDocument` made it trivial, and based on the name, I assumed that had been created with something similar to this problem in mind.

Comment (dquilter) [reply src/__experimental__/components/decimal/decimal.spec.js]: Using `mount` actually helped me resolve another issue, so have switched over to that.

File: src/__experimental__/components/decimal/decimal.stories.js (hunk 22 above)
Comment (dquilter) [reply src/__experimental__/components/decimal/decimal.stories.js]: I'm not sure this is going to work for `Decimal` without re-writing your `getTextboxStoryProps` method. I include some defaults for these properties in `Decimal`, and my `inputWidth` is not reliant on he `labelInline` being defined.

In fact, looking at what you've done in `getTextboxStoryProps`, I think that the definitions for `labelWidth` and `inputWidth` are at odds with some of the styling work I've done in this PR. I've updated so that the `inputWidth` gets defined as a percentage of the remaining width (total width - `labelWidth`). Check out: https://github.com/Sage/carbon/pull/1970/files#diff-1ae771b61db1ee50897f7001bdc23d1dR22

File: src/__experimental__/components/decimal/decimal.spec.js (hunk 21 above)
Comment (dquilter) [reply src/__experimental__/components/decimal/decimal.spec.js]: Updated as per other change request.

File: src/__experimental__/components/decimal/decimal.stories.js (hunk 22 above)
Comment (grabkowski) [reply src/__experimental__/components/decimal/decimal.stories.js]: `inputWidth` prop should cause the same effects in all `Textbox` based components, if it should not be reliant on the `labelInline` then it should be a standalone prop in the `Textbox` also. 

I still think that you should reuse these `Textbox` props.

File: src/__experimental__/components/decimal/decimal.component.js [hunk 29]
```diff
@@ -0,0 +1,132 @@
...
+      setTimeout(() => {
+        target.setSelectionRange(newPosition, newPosition);
+      });
+    }
+  }
+
+  onBlur = (evt) => {
+    // Call onChange to force decimal formattting
+    // Alternative is to store input value in state.
+    this.onChange(evt);
```
andrewjtait: How using `forceUpdate` instead, then it won't trigger `onChange` unnecessarily https://reactjs.org/docs/react-component.html#forceupdate

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,132 @@
...
+   * Maximum value for precision
+   */
+  maxPrecision: PropTypes.number
+};
+
+Decimal.defaultProps = {
+  align: 'right',
+  precision: 2,
+  maxPrecision: 15,
+  inputWidth: 100,
```
andrewjtait: does it need this? As it is defaulted in `InputPresentationStyle`

src/__experimental__/components/decimal/decimal.stories.js
andrewjtait: `labelWidth` and `inputWith` should behave the same for all input components, and we should really try and replicate the current behaviour to mitigate any issues during upgrade.

src/__experimental__/components/decimal/decimal.spec.js
grabkowski: now that we have a default `maxPrecision`, could you use that value here?

src/__experimental__/components/decimal/decimal.component.js
dquilter: I'd always thought `forceUpdate` was a last resort type thing. Happy to change though. 👍

andrewjtait: It is, but I think it makes sense in this case.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,129 @@
...
+    const { precision } = this.props;
+    const format = I18nHelper.format();
+    const delimiter = `\\${format.delimiter}`;
+    const seperator = `\\${format.separator}`;
+    const validDecimalMatcher = new RegExp(`^[\\d${delimiter}]*[${seperator}{1}]?\\d{0,${precision}}?$`);
+
+    return validDecimalMatcher.test(value);
+  }
+
+  onChange = (evt) => {
```
dawidzarzycki: We decided to use `ev` as an action argument
https://github.com/Sage/carbon/wiki/Codebase-Conventions#action-argument-names

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -0,0 +1,51 @@
...
+import Textbox from '../textbox';
+import getTextboxStoryProps from '../textbox/textbox.stories';
+import OptionsHelper from '../../../utils/helpers/options-helper';
+import notes from './documentation';
+
+const store = new Store({
+  value: Decimal.defaultProps.value
+});
+
+const setValue = (evt) => {
```
dawidzarzycki: `ev` instead of `evt`
https://github.com/Sage/carbon/wiki/Codebase-Conventions#action-argument-names

src/__experimental__/components/decimal/index.js
```diff
@@ -0,0 +1,3 @@
+import Decimal from './decimal.component';
```
dawidzarzycki: Is there a reason that we can't use? 
```js
export { Decimal } from './decimal.component';
```

dquilter: That'll be because the Decimal component is the `default` export from that file.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -0,0 +1,133 @@
...
+      precision,
+      maxPrecision
+    } = this.props;
+
+    if (precision > maxPrecision) {
+      Logger.warn(`Precision cannot be greater than ${maxPrecision}`);
+      return maxPrecision;
+    }
+
+    if (!precision || precision < 0) {
```
andrewjtait: won't `!precision` be truthy if precision is `0`? should it not be `precision === false`? To be honest I'm not entirely clear what scenario this is trying to resolve?

dquilter: `0` is a falsey value in Javascript. With `!precision`, I was trying to account for that `0` but also, the Storybook knob passing an empty `string`.

andrewjtait: What I mean is, shouldn't we respect `0` as a value? This looks like it will apply the default value if 0 is passed in? Maybe it should be `precision === '' || precision < 0`?

There should probably be a limit though to what values we cater for vs quirks that storybook are throwing at us? I mean PropTypes are there to guide the developer to the correct value type, _should_ we also be coding around invalid types?

PR #1973, opened 2019-03-20
FE-1285 grouped character

samhinton88:
# Description
New Grouped Character component.

Replaces the older, class-based component with a functional version making use of the [Hooks api](https://reactjs.org/docs/hooks-intro.html).

# Screenshots / Gifs
 
grouped-char


src/__experimental__/components/grouped-character/grouped-character.spec.js
```diff
@@ -0,0 +1,48 @@
...
+      input = instance.find('input');
+    });
+
+    it('takes configuration for how text should be grouped', () => {
+      expect(input.props().value).toEqual('12-34-5678');
+    });
+    it('emits an unformatted string', () => {
+      input.simulate('change', { target: { value: '123456' } });
+      expect(onChange).toHaveBeenCalledWith({ target: { value: '123456' } });
+    });
```
grabkowski: this test is redundant

src/__experimental__/components/grouped-character/grouped-character.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    it('does not allow values of length greater than that allowed by the group config', () => {
+      instance = mountComponent({
+        separator, groups: basicGroupConfig, value: '1234567890', onChange
+      });
+      input = instance.find('input');
+      expect(input.props().value).toEqual('12-34-5678');
+    });
+    fit('formats the input value correctly when deleting characters next to separator', () => {
+      const deletePress = new KeyboardEvent('keydown', { which: '46' });
```
grabkowski: could you use `keyboard` test util here?

src/__experimental__/components/grouped-character/grouped-character.spec.js
```diff
@@ -0,0 +1,48 @@
...
+      instance = mountComponent({
+        separator, groups: basicGroupConfig, value: '12345', onChange
+      });
+      input = instance.find('input');
+      input.simulate('click');
+      document.dispatchEvent(deletePress);
+      expect(input.props().value).toEqual('12-34');
+    });
```
grabkowski: there is also a `backspace` key

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,24 @@
...
+import GroupedCharacter from '.';
+
+const groupedCharacterStore = new Store({
+  value: ''
+});
+
+storiesOf('Experimental/GroupedCharacter', module)
+  .add('Basic', () => {
+    return (
+      <State store={ groupedCharacterStore }>
```
grabkowski: you could use the `StateDecorator` instead of the `State` component

src/__experimental__/components/grouped-character/index.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState, useEffect } from 'react';
```
grabkowski: shouldn't that file be named `grouped-character.component.js`?

src/__experimental__/components/grouped-character/grouped-character.stories.js
samhinton88: What is the advantage in doing that?

src/__experimental__/components/grouped-character/grouped-character.spec.js
samhinton88: `onChange` is the callback supplied by the dev, so this test asserts that the value supplied to that callback is the raw string, rather than a string containing separators.

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,46 @@
...
+
+    if (stepIndices.includes(selectionEnd - 1)) modifier = 1;
+    if (stepIndices.includes(selectionEnd - 1) && Events.isBackspaceKey({ which: pressedKey })) modifier = -1;
+
+    props.onChange({ target: { value: e.target.value.split(separator).join('') } });
+    setTimeout(() => i.setSelectionRange(selectionEnd + modifier, selectionEnd + modifier));
+  };
+
+  return (
+    <Textbox
```
dquilter: Should we be using the InputDecoratorBridge here?

src/__experimental__/components/grouped-character/grouped-character.stories.js
grabkowski: in my opinion - with that decorator jsx looks cleaner

src/__experimental__/components/grouped-character/grouped-character.component.js
samhinton88: My spec was to use the new TextBox, why would you suggest using the bridge?

src/__experimental__/components/grouped-character/index.js
```diff
@@ -0,0 +1,52 @@
...
+    />
+  );
+};
+
+GroupedCharacter.propTypes = {
+  separator: PropTypes.string,
+  groups: PropTypes.array,
+  value: PropTypes.string,
+  onChange: PropTypes.func
+};
```
grabkowski: missing PropTypes descriptions

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,24 @@
...
+        <GroupedCharacter
+          groups={[2,2,4]}
+          inputWidth={'10px'}
+          separator={text('separator', '-')}
+          value={groupedCharacterStore.get('value')}
+          onChange={(ev) => { groupedCharacterStore.set({ value: ev.target.value }); }}
+        />
+      </State>
+    );
```
grabkowski: State PropTypes should be excluded from the info:
```
propTablesExclude: [State]
```

samhinton88: Thanks, I'll add that.

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,46 @@
...
+import Events from '../../../utils/helpers/events/events';
+
+const GroupedCharacter = (props) => {
+  const [pressedKey, updatePressedKey] = useState(null);
+
+  const { separator, groups, value } = props;
+  const stepIndices = groups.reduce(toIndexSteps, []);
+
+  const handleChange = (e) => {
+    const i = e.target;
```
edleeks87: Couldn't this be? Plus `i` seems a bit vague for a name imo
```
 const i = ev.target;
 const { selectionEnd } = i;
```

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,46 @@
...
+    const { selectionEnd } = e.target;
+    let modifier = 0;
+
+    if (stepIndices.includes(selectionEnd - 1)) modifier = 1;
```
edleeks87: If you switched these you could use an `if/ else if` and only adjust the `modifier` value once?

```
 if (stepIndices.includes(selectionEnd - 1) && Events.isBackspaceKey({ which: pressedKey })) modifier = -1;
 else if (stepIndices.includes(selectionEnd - 1)) modifier = 1;
```

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,27 @@
...
+storiesOf('Experimental/GroupedCharacter', module)
+  .addParameters({
+    info: { propTablesExclude: [State] }
+  })
+  .add('Basic',
+    () => {
+      return (
+        <State store={ groupedCharacterStore }>
+          <GroupedCharacter
+            groups={ object('groups', [2, 2, 4]) }
```
dawidzarzycki: Can we have knobs assigned to `const` before we use it in the component? 
```js
const groups = object('groups', [2, 2, 4]);

<GroupedCharacter 
  groups={groups}
  ...
```

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,27 @@
...
+            separator={ text('separator', '-') }
+            value={ groupedCharacterStore.get('value') }
+            onChange={ (ev) => { groupedCharacterStore.set({ value: ev.target.value }); } }
```
dawidzarzycki: Can we move it to `handleChange` function?

```js
const handleChange = (ev) => {
  groupedCharacterStore.set({ value: ev.target.value })
}
```

src/__experimental__/components/grouped-character/grouped-character.component.js
samhinton88: Agreed that `i` is vague, however it shouldn't matter which version I destructure the `selectionEnd` from.

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,50 @@
...
+  const { separator, groups, value } = props;
+  const stepIndices = groups.reduce(toIndexSteps, []);
+
+  const handleChange = (ev) => {
+    const i = ev.target;
+    const { selectionEnd } = ev.target;
+    let modifier = 0;
+
+    if (stepIndices.includes(selectionEnd - 1)) modifier = 1;
+    if (stepIndices.includes(selectionEnd - 1) && Events.isBackspaceKey({ which: pressedKey })) modifier = -1;
```
grabkowski: `stepIndices.includes(selectionEnd - 1)` could be a const

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,50 @@
...
+import Events from '../../../utils/helpers/events/events';
+
+const GroupedCharacter = (props) => {
+  const [pressedKey, updatePressedKey] = useState(null);
+
+  const { separator, groups, value } = props;
+  const stepIndices = groups.reduce(toIndexSteps, []);
+
+  const handleChange = (ev) => {
+    const i = ev.target;
```
grabkowski: could you provide a more meaningful name?

src/__experimental__/components/grouped-character/grouped-character.component.js
grabkowski: don't you think that would be a bit difficult to read

grabkowski: I would extract conditionals to consts and then:
```
if(hasSelectionEnd) {
 isBackspacePressed ? selectionEnd-- : selectionEnd++;
}
```

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,32 @@
...
+import { storiesOf } from '@storybook/react';
+import { Store, State } from '@sambego/storybook-state';
+import { text, object } from '@storybook/addon-knobs';
+import GroupedCharacter from './grouped-character.component';
+
+const groupedCharacterStore = new Store({
+  value: ''
+});
+
+const onChange = (ev) => { groupedCharacterStore.set({ value: ev.target.value }); };
```
dquilter: You could add a Storybook action here too: `action(change)(ev);`

src/__experimental__/components/grouped-character/grouped-character.spec.js
```diff
@@ -0,0 +1,94 @@
...
+import { mount } from 'enzyme';
+import GroupedCharacter from './grouped-character.component';
+
+const mountComponent = props => mount(<GroupedCharacter { ...props } />);
+const setCursorOn = (node, setSelectionRange) => (
+  selectionEnd => node.simulate('change', { target: { selectionEnd, value: '123', setSelectionRange } })
+);
+
+const assertSelectionRangeCalled = selectionFn => (
+  position => expect(selectionFn).toHaveBeenCalledWith(position, position)
```
izakrzewska: why is 'position' double here?

src/__experimental__/components/grouped-character/grouped-character.stories.js
```diff
@@ -0,0 +1,32 @@
...
+
+storiesOf('Experimental/GroupedCharacter', module)
+  .addParameters({
+    info: { propTablesExclude: [State] }
+  })
+  .add('Basic',
```
izakrzewska: if we want to stick to the naming convention we already have in storybook and keep some consistency, it may be better to change the name from 'basic' to 'default'

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,61 @@
...
+      onKeyPress={ (ev) => {
+        if (groups.reduce(toSum) === value.length) {
+          ev.preventDefault();
+        }
+      } }
+    />
+  );
+};
+
+
```
izakrzewska: in a previous implementation '-' was a default separator prop, is there any reason for removing it?

src/__experimental__/components/grouped-character/grouped-character.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    it('pressing a character at the point where a separator should appear adjusts the cursor positon forwards', () => {
+      setInputCursorTo(3);
+      jest.runAllTimers();
+      assertInputCursorAt(4);
+    });
+
+    it('pressing backspace after a separating character adjusts the cursor position backwards', () => {
+      // backspace
```
grabkowski: I think that we could just make a const:
```
const backspaceKeyCode = 8;
```

package-lock.json
```diff
@@ -22960,4 +22960,4 @@
       }
     }
   }
-}
+}
```
grabkowski: is that change needed?

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,67 @@
...
+    const eventRef = ev.target;
+    const { selectionEnd } = ev.target;
+    let modifier = 0;
+    const isAtOneBeyondSeparator = stepIndices.includes(selectionEnd - 1),
+        backspacePressed = Events.isBackspaceKey({ which: pressedKey });
+
+    if (isAtOneBeyondSeparator && backspacePressed) {
+      modifier = -1;
+    } else if (isAtOneBeyondSeparator) {
+      modifier = 1;
```
grabkowski: do we really need this modifier? I think that we could just do:
```
const newCursorPos = selectionEnd;
.
.
.
if (isAtOneBeyondSeparator && backspacePressed) {
      newCursorPos++;
    } else if (isAtOneBeyondSeparator) {
      newCursorPos--;
}
```

grabkowski:
removed the UX QA label as it's the reimplementation of the old component, no new DLS functionalities applied

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -0,0 +1,66 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import Textbox from '../textbox';
+import { generateGroups, toSum, toIndexSteps } from './grouped-character.utils';
+import Events from '../../../utils/helpers/events/events';
```
dawidzarzycki: There is a `package.json` so we don't need to use specific file
we can just use `'../../../utils/helpers/events`

src/__experimental__/components/grouped-character/index.js
```diff
@@ -0,0 +1,3 @@
+import GroupedCharacter from './grouped-character.component';
```
dawidzarzycki: Would you like to export component without import syntax?

src/__experimental__/components/grouped-character/grouped-character.component.js
grabkowski: changed

src/__experimental__/components/grouped-character/index.js
grabkowski: changed

PR #1978, opened 2019-03-22
FE-1224 DLS button

edleeks87:
# Description
Implements styled components to allow the theme based styling to be added to the Button component and bring it in line with what is described in the DLS.


src/components/button/button.component.js
```diff
@@ -0,0 +1,105 @@
...
+import Link from '../link';
+import OptionsHelper from '../../utils/helpers/options-helper';
+
+const Button = (props) => {
+  const {
+    as,
+    disabled,
+    iconPosition,
+    iconType,
+    theme,
```
grabkowski: you could rename that prop when destructuring:
`theme: variant`

src/components/button/button.component.js
```diff
@@ -0,0 +1,105 @@
...
+  if (props.href || props.to) {
+    return (
+      <Link
+        role='button'
+        variant={ theme }
+        { ...tagComponent('button', props) }
+        { ...rest }
+      >
+        { renderChildren(props) }
+      </Link>
```
grabkowski: it could be extracted to a separate method, e.g.: `renderLink` to increase readability

edleeks87: Yeah I agree it's a bit convoluted, I'll make the suggested change

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+
+describe('Button', () => {
+  describe('when no props other than children are passed into the component', () => {
+    it('renders the default props and children to match the snapshot', () => {
+      const wrapper = render({ children: 'foo' });
+      expect(wrapper.props().renderAs).toEqual('secondary');
+      expect(wrapper.props().size).toEqual('medium');
+      expect(wrapper.props().disabled).toEqual(false);
+      expect(wrapper.props().iconPosition).toEqual('');
+      expect(!wrapper.contains(<Icon type='filter' />)).toBeTruthy();
```
grabkowski: why is that exclamation mark for?

edleeks87: `toBeEdgy` 😆
Not sure what I was doing there, I've removed it and changed the expectation to `toBeFalsy()`

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+      const wrapper = render({ children: 'foo', iconType: 'filter', iconPosition: 'before' }, TestRenderer.create);
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('renders the default props and children to match the snapshot with the Icon after children', () => {
+      const wrapper = render({ children: 'foo', iconType: 'filter', iconPosition: 'after' }, TestRenderer.create);
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    ['before', 'after'].forEach((position) => {
```
grabkowski: you could use `describe.each` for that

src/components/button/__snapshots__/button.spec.js.snap
```diff
@@ -0,0 +1,1109 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
grabkowski: Theme object is repeated multiple times in these snapshots. It could be mapped to a string, when using `toJson` method of the `enzyme-to-json` library.

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
+import React from 'react';
```
grabkowski: I don't think that we need to check if the default props are set correctly in every test?

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+            paddingLeft: `${sizes[size][1].toString()}px`,
+            paddingRight: `${sizes[size][1].toString()}px`
+          }, wrapper);
+        });
+      });
+    });
+  });
+
+  describe('when the classic theme is applied', () => {
+    ['blue', 'grey', 'magenta', 'magenta-dull', 'red', 'white'].forEach((variant) => {
```
grabkowski: with `describe.each` you could even insert each string into a description using `%s`

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+    anchor = render({
+      href: '/foo',
+      children: 'Anchor'
+    }, mount);
+
+    to = render({
+      to: '/foo',
+      children: 'To'
+    }, mount);
+  });
```
grabkowski: I think it would be better to have each test with a separate data

edleeks87: These are the tests remaining from the legacy button so I just left them as is wherever possible as it demonstrates that the backwards compatibility is maintained

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+
+  describe('Passing a custom onClick', () => {
+    it('triggers when the button is clicked', () => {
+      const wrapper = render({
+        children: 'foo',
+        onClick: jest.fn()
+      });
+
+      wrapper.simulate('click');
+      expect(wrapper.props().onClick).toHaveBeenCalled();
```
grabkowski: this test is redundant - when triggering a click on the wrapper element React will always call the `onClick` method, you should trigger that event on a descendant of this component.

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+    describe('default', () => {
+      it('renders a button', () => {
+        const wrapper = render({ children: 'foo ' }, mount);
+        expect(wrapper.type()).toEqual(Button);
+      });
+    });
+
+    describe('with href', () => {
+      it('renders an anchor', () => {
+        expect(anchor.exists('_Link')).toEqual(true);
```
grabkowski: I think that `anchor.find(_Link).exists()` would be more readable. It's also better to import and check for a component instead of a string selector.

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+        expect(anchor.find('_Link').length).toEqual(1);
```
grabkowski: isn't the result the same as with the preceding `exists`?

src/components/button/button.spec.js
```diff
@@ -0,0 +1,364 @@
...
+      wrapper.simulate('click');
+      expect(wrapper.props().onClick).toHaveBeenCalled();
+    });
+  });
+
+  describe('render', () => {
+    describe('default', () => {
+      it('renders a button', () => {
+        const wrapper = render({ children: 'foo ' }, mount);
+        expect(wrapper.type()).toEqual(Button);
```
grabkowski: it will always be a `Button`

src/components/button/button.spec.js
grabkowski: don't you think that these tests should be rewritten, or removed when unnecessary?

src/components/button/button.style.js
```diff
@@ -0,0 +1,131 @@
...
+  display: inline-flex;
+  flex-direction: column;
+  flex-flow: wrap;
+  justify-content: center;
+  vertical-align: middle;
+
+  ${({ theme }) => theme.name !== 'classic' && stylingForType}
+  ${({ theme, size }) => theme.name !== 'classic' && buttonSizes(theme)[size]}
+
+  ${({ theme, renderAs }) => theme.name === 'classic' && classicRenderAs(renderAs) && stylingForClassic}
```
grabkowski: maybe it could be done in one method?
```
${({ theme, renderAs, size }) => ...
```

src/components/button/button-types.style.js
```diff
@@ -0,0 +1,62 @@
+export default ({ colors }) => ({
```
grabkowski: why not just return a string, that approach will save us a lot of unnecessary logic
```
primary: `
  background: ${colors.primary},
  color: ${colors.white}
  &:hover {
      background: ${colors.secondary},
      color: ${colors.white}
  }
`,
secondary: `
.
.
.
```

src/components/button/button.style.js
edleeks87: Yeah I've tidied this up a bitand separated it into a function that calls the correct style based on the theme context etc

src/components/button/button.spec.js
edleeks87: I want to try and leave them be wherever possible as it is clearer in demonstrating the backwards compatibility, I have refactored it so each one is in its own `describe` as it is odd the way it is

src/components/button/button-types.style.js
edleeks87: There's still the issue of having to then define the `disabled` style for each state the button is in, as it looks different when it is `primary` than when it is `secondary` etc

src/components/button/button-classic-config.style.js
```diff
@@ -0,0 +1,225 @@
...
+      &:focus {
+        box-shadow: 0 0 6px rgba(51,92,109,.6);
+        outline: none;
+      }
+    `
+  },
+  secondary: {
+    blue: `
+      background: transparent;
+      border: 1px solid #255bc7;
```
dquilter: You could declare variables for some of these repeated colours.

src/components/button/button.spec.js
```diff
@@ -0,0 +1,354 @@
...
+
+describe('Button', () => {
+  describe('when no props other than children are passed into the component', () => {
+    it('renders the default props and children to match the snapshot', () => {
+      const wrapper = render({ children: 'foo' });
+      expect(wrapper.props().renderAs).toEqual('secondary');
+      expect(wrapper.props().size).toEqual('medium');
+      expect(wrapper.props().disabled).toEqual(false);
+      expect(wrapper.props().iconPosition).toEqual('');
+      expect(wrapper.contains(<Icon type='filter' />)).toBeFalsy();
```
dquilter: Wonder whether you need the above lines if you're matching snapshot straight after?

dquilter: Or vice versa.

src/components/button/button.spec.js
```diff
@@ -0,0 +1,354 @@
...
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('renders the subtext if the size prop is "large"', () => {
+      const wrapper = render({ children: 'foo', size: 'large', subtext: 'foo' }, TestRenderer.create);
+      expect(wrapper).toMatchSnapshot();
+    });
+  });
+
+  describe('when the no props are passed except children', () => {
```
dquilter: `when ~the~ no props are passed except children`

src/components/button/button.spec.js
```diff
@@ -0,0 +1,354 @@
...
+    it('matches the expected style for a "blue" Button with default props', () => {
+      const wrapper = TestRenderer.create(<StyledButton theme={ classicTheme }>foo</StyledButton>);
+      assertStyleMatch({
+        background: 'transparent',
+        border: '1px solid #255bc7',
+        color: '#255bc7'
+      }, wrapper.toJSON());
+    });
+
+    it('matches the expected style default "diabled" Button', () => {
```
dquilter: Spelling: "disabled"

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+const defaultKnobs = () => {
+  const size = select('size', OptionsHelper.sizesRestricted, OptionsHelper.sizesRestricted[1]);
+
+  return {
+    as: select('as', OptionsHelper.themesBinary, OptionsHelper.themesBinary[1]),
+    children: text('children', 'Example Button'),
+    disabled: boolean('disabled', false),
+    iconPosition: select('iconPosition', [...OptionsHelper.buttonIconPositions, ''], ''),
+    iconType: select('iconType', [...OptionsHelper.icons, ''], ''),
+    onClick: action('click'),
```
dquilter: If you use `onClick: (evt) => action('click')(evt)`, you can get the event info in the action too.

src/components/button/button-classic-config.style.js
edleeks87: Good point, as the classic stuff is likely/ hopefully going to be fazed out down the line I've defined them in the same file

src/components/button/button.stories.js
edleeks87: Ah cool, changed it to that 😃

src/components/button/button.spec.js
```diff
@@ -75,8 +71,8 @@ describe('Button', () => {
     });
   });
 
-  describe('when the no props are passed except children', () => {
-    it('sets the default variants to "secondary" and "medium"', () => {
+  describe('when there are no props passed except children', () => {
+    it('matchwes the expect styles for a default button with variants "secondary" and "medium"', () => {
```
dquilter: Probs "matches the expected"...  😉

edleeks87: 🤦‍♂️

edleeks87: 🤦‍♂️

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { storiesOf } from '@storybook/react';
 import { text, select, boolean } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import OptionsHelper from '../../utils/helpers/options-helper';
-import notes from './notes.md';
-import Button from './button';
+import notes from './documentation/notes.md';
+import Info from './documentation/Info';
```
dawidzarzycki: in `documentation` we can create an `index.js` file and export both: `notes` and `info `so we can just import everything from one source
```js
//index.js
export { default as notes } from './notes.md'
export { default as info} from './info'
```
then we can just import 
```js
import { notes, info } from './documentation'
```

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+import Button from '.';
+
+const defaultKnobs = () => {
+  const size = select('size', OptionsHelper.sizesRestricted, OptionsHelper.sizesRestricted[1]);
+
+  return {
+    as: select('as', OptionsHelper.themesBinary, OptionsHelper.themesBinary[1]),
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.as`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
 import { action } from '@storybook/addon-actions';
 import OptionsHelper from '../../utils/helpers/options-helper';
-import notes from './notes.md';
-import Button from './button';
+import notes from './documentation/notes.md';
+import Info from './documentation/Info';
+import Button from '.';
+
+const defaultKnobs = () => {
+  const size = select('size', OptionsHelper.sizesRestricted, OptionsHelper.sizesRestricted[1]);
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.size`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+
+  return {
+    as: select('as', OptionsHelper.themesBinary, OptionsHelper.themesBinary[1]),
+    children: text('children', 'Example Button'),
+    disabled: boolean('disabled', false),
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.disabled`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+    iconPosition: select('iconPosition', [...OptionsHelper.buttonIconPositions, ''], ''),
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.iconPosition`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+    iconType: select('iconType', [...OptionsHelper.icons, ''], ''),
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.iconType`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+    onClick: ev => action('click')(ev),
+    size,
+    subtext: size === OptionsHelper.sizesRestricted[2] ? text('subtext', '') : undefined,
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.subText`

src/components/button/button.stories.js
```diff
@@ -1,32 +1,105 @@
...
+    theme: select('theme', OptionsHelper.buttonColors, OptionsHelper.buttonColors[0]),
```
dawidzarzycki: Can we take it from defaultProps? `Button.defaultProps.theme`

src/components/button/button.component.js
```diff
@@ -22,7 +22,6 @@ const Button = React.forwardRef((props, ref) => {
       iconType={ iconType }
       iconPosition={ iconPosition }
       variant={ theme }
-      as={ (props.href || props.to) ? 'a' : 'button' }
```
edleeks87: Hey @grabkowski thanks for carrying the work on while I was away, just curious as to why you made these changes? Might be missing something obvious as my mind still thinks it's on holiday 🍸

src/components/button/button.component.js
```diff
@@ -0,0 +1,122 @@
...
+  );
+
+  if (!disabled && (href || to)) {
+    buttonContent = <Link href={ href } to={ to }>{buttonContent}</Link>;
+  }
+
+  return (
+    <>
+      { iconType && iconPosition === 'before' && <Icon type={ iconType } /> }
+      {buttonContent}
```
andrewjtait: Does this mean the text within the button is the only clickable bit? The entire button should behave as the link.

andrewjtait: Screenshot 2019-05-09 at 15 02 10

^ not like this

src/components/button/button.component.js
```diff
@@ -0,0 +1,122 @@
...
+  as: 'secondary',
+  size: 'medium',
+  disabled: false,
+  iconPosition: 'before',
+  theme: 'blue',
+  subtext: ''
+};
+
+/** HOC created as a workaround for a Storybook rendering wrong PropTables and Story Source when forwarding a ref */
+const withForwardRef = () => {
```
dquilter: If this is just for Storybook, should it not live in the `stories.js` file? Or it's own file?

src/components/button/button.component.js
grabkowski: fixed

src/components/button/button.component.js
grabkowski: we need a `ForwardRef` for the `SplitButton` component. When forwarded directly it messes with Storybook. To get them both working we need it as a HO.C

src/components/button/button-classic.style.js
```diff
@@ -0,0 +1,342 @@
...
+      &:active {
+        background-color: ${blue.active};
+        border-color: ${blue.active};
+      }
+      &:hover {
+        background: ${blue.hover};
+        border-color: ${blue.hover};
+      }
+      &:focus {
+        box-shadow: 0 0 6px rgba(25,99,246,.6);
```
dquilter: Should these `box-shadow` colours be in your colour configs?

src/components/button/button.spec.js
```diff
@@ -0,0 +1,325 @@
...
+    });
+  });
+
+  describe('when a subtext prop is passed into the component', () => {
+    it('does not render the subtext if the size prop is not "large"', () => {
+      try {
+        const wrapper = render({ children: 'foo', subtext: 'bar' }).dive();
+
+        expect(wrapper.find('[data-element="subtext"]').exists()).toBe(false);
+      } catch (error) {} // eslint-disable-line no-empty
```
dquilter: Is it worth logging the error?

grabkowski: this code just catches that error so it won't be logged in the console (which we do not want).

src/components/button/button-classic.style.js
grabkowski: it's classic styling

PR #1980, opened 2019-03-22
integrating new validation with old validation

andrewjtait:
in progress...

This PR updates the `Form` component to use the new Validation context and API. In order to support backward compatibility, I have also updated the input-validation decorator to use this context object (to increment and decrement validation counts).

I have also refactored the new validation API to use a configuration object of validations, providing the ability to easily add or remove validation types.


src/components/validations/validation-types.constant.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
samhinton88: These should probably either be actual constants, or the file should be renamed.

src/components/form/form.js
```diff
@@ -519,6 +402,8 @@ class Form extends React.Component {
 
     if (valid && this.props.onSubmit) {
       this.props.onSubmit(ev, valid, this.enableForm);
+    } else if (valid) {
+      this._form.submit();
     }
```
grabkowski: this conditional block could be extracted into a separate method like `triggerFormSubmit`

grabkowski: could the `this.props.afterFormValidation` conditional be moved above the `valid` conditional and then reuse that `valid` conditional to submit the form?

src/components/validations/with-validation.hoc.js
```diff
@@ -12,36 +13,29 @@ const withValidation = (WrappedComponent) => {
...
       onBlur: PropTypes.func, // Custom function to be called when the component blurs
-      onChange: PropTypes.func // Custom function called when component value changes
+      onChange: PropTypes.func, // Custom function called when component value changes
+      ...Object.values(VALIDATION_TYPES).reduce((acc, type) => ({
+        ...acc,
+        [type]: PropTypes.oneOfType([ // The info validations that should be run against the value
+          PropTypes.func,
+          PropTypes.arrayOf(PropTypes.func)
+        ])
+      }), {})
```
grabkowski: I liked the previous solution better. I think that we shouldn't compromise readability for a few lines of code less.

src/utils/decorators/input-validation/input-validation.js
```diff
@@ -225,20 +229,13 @@ const InputValidation = (ComposedComponent) => {
...
-      if (this.context.form && (this._validations() || this.props.warnings || this.props.info)) {
-        // attach the input to the form so the form can track what it needs to validate on submit
-        this.context.form.attachToForm(this);
+    initValidationInput = (context) => {
+      if (!this.newContextIntegration && context) {
+        this.newContextIntegration = context;
+        if (this._validations() || this.props.warnings || this.props.info) {
+          // attach the input to the form so the form can track what it needs to validate on submit
+          context.addInput(this._guid, this.validateFromForm);
+        }
```
grabkowski: that block could be extracted into a separate method to avoid nesting conditionals

src/components/validations/with-validation.hoc.js
andrewjtait: It's not to reduce the lines of code, it is to try and address that we are supporting different kinds of validations and that code is spread across multiple components. Basing it from the same config allows us to better manage that going forward. One of the issues we had previously was trying to remember to update the code for all 3 variations.

andrewjtait: I agree that the previous code was more readable, but tools such as storybook means we shouldn't have developers looking at the code to see the defined prop types.

edleeks87: I also agree it was a little more readable but doing it this way should make it easier if we ever add or remove validation types

grabkowski: sounds reasonable :)

src/components/form/form.js
andrewjtait: done 👍

PR #1982, opened 2019-03-26
Fe 1225

dawidzarzycki:
# Description
create styled component for link component, create tests for the component, complete classic theme for the link case, create snapshot


src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    // eslint-disable-next-line jsx-a11y/anchor-is-valid
+    wrapper = shallow(<Link>test</Link>);
+  });
+
+  it('renders as expected', () => {
+    expect(render()).toMatchSnapshot();
+  });
+
+
+  it('if `disabled` prop has been passed to the component', () => {
```
edleeks87: The wording for these `it` blocks is a little confusing, it might be better to have them say something like:
```
it('matches the expected style if `disabled` prop has been passed to the component', () => ...
```

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  describe('should render a correct component based on prop', () => {
+    it('if prop `href` has been passed to the component it should render `<a>`', () => {
+      wrapper.setProps({ href: '#' });
+
+      expect(wrapper.find('a')).toHaveLength(1);
+    });
+
+    it('if prop `to` has been passed to the component it should render `<Link />`', () => {
```
edleeks87: Maybe switch the wording around for these to something like this
```
it('should render a `<a>` if prop the `href` has been passed to the component'
```

src/style/color-config/index.js
```diff
@@ -1,5 +1,7 @@
 // Base colors
 export default {
+  azul: '004B87',
```
edleeks87: This colour is already defined below as `navyBlue`

src/components/link/link.component.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return this.icon;
+  }
+
+  get iconRight() {
+    if (!this.props.icon || this.props.iconAlign !== 'right') {
+      return null;
+    }
+    return this.icon;
+  }
```
grabkowski: you could use one method to render that icon:
```
renderLinkIcon = (currentAlignment) => {
  const hasProperAlignment = (this.props.icon && this.props.iconAlign === currentAlignment);

  return hasProperAlignment ? this.icon : null;
}
```

src/components/link/link.component.js
```diff
@@ -0,0 +1,145 @@
...
+        className={ classes }
+        tooltipMessage={ this.props.tooltipMessage }
+        tooltipAlign={ this.props.tooltipAlign }
+        tooltipPosition={ this.props.tooltipPosition }
+      />
+    );
+  }
+
+  get tabIndex() {
+    return this.props.tabbable && !this.props.disabled ? '0' : '-1';
```
grabkowski: could you remove the negation?
`this.props.disabled ? '-1' : '0'`

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+const render = (props) => {
+  return TestRenderer.create(<Link { ...props }>test</Link>);
+};
+
+describe('Link', () => {
+  let wrapper;
+
+  beforeEach(() => {
+    // eslint-disable-next-line jsx-a11y/anchor-is-valid
+    wrapper = shallow(<Link>test</Link>);
```
grabkowski: you should use one render method in which you can pass props and a renderer

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+const render = (props) => {
+  return TestRenderer.create(<Link { ...props }>test</Link>);
+};
+
+describe('Link', () => {
+  let wrapper;
+
+  beforeEach(() => {
+    // eslint-disable-next-line jsx-a11y/anchor-is-valid
```
grabkowski: why do you disable eslint for the next line?

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  describe('should render a correct component based on prop', () => {
+    it('should render a `<a>` if prop the `href` has been passed to the component', () => {
+      wrapper.setProps({ href: '#' });
+
+      expect(wrapper.find('a')).toHaveLength(1);
+    });
+
+    it('should render a `<Link />` if prop the `to` has been passed to the component', () => {
```
grabkowski: should render a "RouterLink" component?

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+
+  it('matches the expected style if `disabled` prop has been passed to the component', () => {
+    assertStyleMatch({
+      cursor: 'not-allowed'
+    }, render({ disabled: true }).toJSON());
+  });
+
+  describe('should render a correct component based on prop', () => {
```
grabkowski: test descriptions should look like:
describe('when something is done with the component', ...
  it('should do something we expect', ...

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+  });
+
+  describe('should render an icon if prop `icon` has been passed to the component', () => {
+    beforeEach(() => {
+      wrapper.setProps({ icon: 'basket' });
+    });
+
+    it('the `Icon` rendered', () => {
+      expect(wrapper.find('Icon').props().type).toEqual('basket');
```
grabkowski: you're checking for a specific icon, the description should reflect that

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    it('if prop `to` has been passed to the component', () => {
+      const spy = jest.fn();
+      wrapper.setProps({ to: 'testRoute', onClick: spy });
+      wrapper.find(RouterLink).simulate('keydown', { which: 13 });
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('if prop `to` has been passed to the component but there was not `onClick` prop', () => {
```
grabkowski: typo: "there was **not** `onClick` prop"

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+    it('if prop `href` has been passed to the component', () => {
+      const spy = jest.fn();
+      wrapper.setProps({ href: '#', onClick: spy });
+      wrapper.find('a').simulate('keydown', { which: 13 });
+
+      expect(spy).not.toHaveBeenCalled();
+    });
+
+    it('if prop `to` has been passed to the component', () => {
```
grabkowski: this is a description not an expectation

src/components/link/link.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  describe('should generate correct events when `onKeyDown` prop has been passed to the component', () => {
+    it('if prop `onClick` has been passed to the component', () => {
+      const spy = jest.fn();
+      wrapper.setProps({ onClick: spy, to: 'testRoute', onKeyDown: spy });
+      wrapper.find(RouterLink).simulate('keydown', { keyCode: 13 });
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('if prop `href` has been passed to the component', () => {
```
grabkowski: this is a description not an expectation

src/components/link/link.style.js
```diff
@@ -0,0 +1,40 @@
...
+
+const LinkStyle = styled.div`
+  a {
+    font-size: 14px;
+    text-decoration: underline;
+    color: ${({ theme }) => theme.colors.primary} !important;
+
+    &:hover {
+      cursor: pointer;
+      color: ${({ theme }) => theme.colors.secondary} !important;
```
grabkowski: you have `!important` twice for the same property

src/components/link/link.style.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    &:focus {
+      color: ${({ theme }) => theme.text.color} !important;
+      background-color: ${({ theme }) => theme.colors.focus};
+      outline: none;
+    }
+
+    ${({ disabled, theme }) => disabled && css`
+      color: ${theme.disabled.text} !important;
```
grabkowski: another `!important`

src/style/color-config/index.js
```diff
@@ -1,5 +1,6 @@
 // Base colors
 export default {
+  blue: '255BC7',
```
grabkowski: @andrewjtait - do you think that should be a part of the color-config?

src/components/link/link.spec.js
dawidzarzycki: The above eslint rule force us to make `anchor` with `href` property, but we want to have that without href because we do not know what prop we will need

src/components/link/link.style.js
dawidzarzycki: This is for hover example

src/components/link/link.style.js
dawidzarzycki: This is for disabled example

src/components/link/link.component.js
dawidzarzycki: `this.icon` return a html tag so we need to have both (right and left) functions to set the icon correctly,

grabkowski: you use that method with an argument: `renderLinkIcon('left')`

src/components/link/link.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    describe('and component recevied an `iconAlign: right` prop', () => {
+      it('should render an `Icon` on the right side of the component', () => {
+        wrapper.setProps({ iconAlign: 'right' });
+
+        expect(wrapper.find('Icon').hasClass('carbon-link__icon--align-right')).toEqual(true);
+      });
+    });
+  });
+
+  describe('when component received an `onKeyDown` prop', () => {
```
grabkowski: when the onKeyDown event is triggered

src/components/link/link.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+  beforeEach(() => {
+    wrapper = renderLink({});
+  });
+
+  it('renders as expected', () => {
+    expect(render()).toMatchSnapshot();
+  });
+
+  it('matches the expected style if `disabled` prop has been passed to the component', () => {
```
grabkowski: please add a describe block here

src/components/link/link.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    describe('and component received a `to` prop but a `onClick` props is not available', () => {
+      it('should trigger `onKeyDown` prop but not `onClickFn`', () => {
+        wrapper.setProps({ to: 'testRoute', onKeyDown: onKeyDownFn });
+        wrapper.find(RouterLink).simulate('keydown', { which: 13 });
+
+        expect(onKeyDownFn).toHaveBeenCalled();
+        expect(onClickFn).not.toHaveBeenCalled();
+      });
```
grabkowski: you should split that into two `it` blocks

benwilsonsage:
Looking great - a few minor style points...

Links > Classic Theme
Expected: Links have font-weight 700.
Actual: Links have default weight.

Links > Classic Theme
Expected: Default state has no underline, and underline is applied on hover only.
Actual: All states have underline.

Links > Classic Theme
Expected: Focus state is browser default (no style).
Actual: Focus state has gold background.

Links > Small, Medium, Large themes
Expected: Focus state has background colour #FFDA7F (gold 50)
Actual: Focus state has gold background #FFB500
(There was some inconsistency in specification which will be resolved on the Design System site soon.)

Links > Classic Theme
Expected: Icons are positioned visually in the middle of the text X-height.
Actual: Icons appear a couple of px too low (for example, position: relative top: 2px would fix it)

src/style/themes/classic/classic-theme.config.js
```diff
@@ -6,21 +6,21 @@ export default (palette) => {
       name: THEMES.classic,
 
       colors: {
-        base: palette.productGreen,
-        primary: palette.productGreenShade(21),
-        secondary: palette.productGreenShade(41),
-        tertiary: palette.productGreenShade(61),
+        base: palette.blue,
```
samhinton88: These colours should not be changed or renamed.

src/style/themes/base/base-theme.config.js
```diff
@@ -20,7 +20,7 @@ export default (palette) => {
 
       // status
       error: palette.errorRed,
-      focus: palette.gold,
+      focus: palette.goldTint(50),
```
samhinton88: This should not be changed. What is this `focus` going to be used for?

src/components/link/link.component.js
```diff
@@ -0,0 +1,153 @@
...
+
+    return props;
+  }
+
+  renderLinkContent() {
+    return (
+      <span>
+        {this.renderLinkIcon()}
+
+        <span className='carbon-link__content'>{this.props.children}</span>
```
samhinton88: Is there a reason that we're keeping these class names?

src/style/themes/base/base-theme.config.js
dawidzarzycki: right, I added this to the small, medium and large theme instead

src/components/link/link.component.js
```diff
@@ -0,0 +1,153 @@
...
+  renderLinkIcon = (currentAlignment = 'left') => {
+    const hasProperAlignment = this.props.icon && (this.props.iconAlign === currentAlignment);
+
+    return hasProperAlignment ? this.icon : null;
+  }
+
+  get icon() {
+    const classes = classNames(
+      'carbon-link__icon',
+      `carbon-link__icon--align-${this.props.iconAlign}`
```
samhinton88: Is there a reason we're still including class names here?

dawidzarzycki: Yes, the problem is that `Icon` component is working on that, and changes need to be create also inside the `Icon` component but this is out of the ticket scope

src/style/color-config/index.js
samhinton88: This colour should not be added to the base colour config.

src/style/themes/large/large-theme.config.js
```diff
@@ -9,7 +9,10 @@ export default (palette) => {
         base: palette.amethyst,
         primary: palette.amethystTint(10),
         secondary: palette.amethystShade(10),
-        tertiary: palette.amethystShade(30)
+        tertiary: palette.amethystShade(30),
+
+        // status
+        focus: palette.goldTint(50)
```
samhinton88: Why have you added this here? What is the purpose of this colour?

dawidzarzycki: This is on design of the DLS

src/style/themes/base/base-theme.config.js
samhinton88: I would probably give this a more specific name, like `linkFocus`. Otherwise it could be describing a focus state for anything.

src/style/themes/classic/classic-theme.config.js
```diff
@@ -12,15 +12,15 @@ export default (palette) => {
         tertiary: palette.productGreenShade(61),
 
         // generic
-        baseBlue: '#255BC7',
         greyDarkBlue50: '#8099a4'
       },
 
       text: {
-        color: '#255bc7'
+        color: palette.blue
```
samhinton88: This `blue` should be removed from the `palette` object.

src/components/link/link.component.js
```diff
@@ -0,0 +1,153 @@
...
+import tagComponent from '../../utils/helpers/tags';
+import LinkStyle from './link.style';
+import './link.scss';
+
+class Link extends React.Component {
+  static safeProps = ['onClick'];
+
+  constructor() {
+    super();
+    this.onKeyDown = this.onKeyDown.bind(this);
```
samhinton88: Consider removing this and using a class method assignment for `onKeyDown` in the body of the `class`:

```
onKeyDown = (ev) => {
  // code
}
```

src/style/themes/base/base-theme.config.js
```diff
@@ -35,7 +35,8 @@ export default (palette) => {
     disabled: {
       border: palette.slateTint(80),
       disabled: blackWithOpacity(0.55),
-      input: palette.slateTint(95)
+      input: palette.slateTint(95),
+      text: blackWithOpacity(0.3)
```
samhinton88: Do you know what the `disabled.disabled` colour is used for?

src/components/link/link.component.js
```diff
@@ -0,0 +1,153 @@
...
+import { validProps } from '../../utils/ether';
+import Event from '../../utils/helpers/events';
+import tagComponent from '../../utils/helpers/tags';
+import LinkStyle from './link.style';
+import './link.scss';
+
+class Link extends React.Component {
+  static safeProps = ['onClick'];
+
+  constructor() {
```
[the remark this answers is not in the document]
dawidzarzycki: yup done

src/components/link/link.component.js
dawidzarzycki: Fixed

src/style/themes/base/base-theme.config.js
dawidzarzycki: I see this is used in `day-picker` files

PR #1984, opened 2019-03-27
FE-1355 input icon toggle classic theme

stevebutler2210:
Adds classic theme for the InputIconToggle component.

Sets dark blue colours on focus of the Input / hover of the InputIconToggle, and light blue colours on hover of the Input / Label components, replicating the functionality seen in the old component on the current demo site.


src/__experimental__/components/input-icon-toggle/input-icon-toggle-classic.style.js
```diff
@@ -0,0 +1,31 @@
...
+import InputPresentationStyle from '../input/input-presentation.style';
+import InputStyle from '../input/input.style';
+
+export default ({
+  theme, type
+}) => theme.name === THEMES.classic && css`
+  background-color: #e6ebed;
+  border-left: 1px solid #bfccd2;
+  box-sizing: border-box;
+  color: #003349;
```
grabkowski: why are these colors hardcoded?

src/style/themes/classic/classic-theme.config.js
```diff
@@ -9,7 +9,14 @@ export default (palette) => {
         base: palette.productGreen,
         primary: palette.productGreenShade(21),
         secondary: palette.productGreenShade(41),
-        tertiary: palette.productGreenShade(61)
+        tertiary: palette.productGreenShade(61),
+
+        iconBackground: '#e6ebed',
```
samhinton88: Is there a reason for declaring some colour hexes here and some in the palette? Might be better to be consistent one way or the other.

src/style/color-config/index.js
```diff
@@ -6,6 +6,7 @@ export default {
   genericGreen: '009900',
   productGreen: '00A376',
   productBlue: '0077C8',
+  lightBlue: '1963F6',
```
grabkowski: is it used in the classic theme only?

stevebutler2210:
> Screenshot 2019-04-01 at 13 29 02
> 
> 1px border missing on the right - does the input-label margin-right need to be `margin-right: -6px;` rather than `6.5`?

Hi @andrewjtait, what browser are you using? I saw you'd made that change previously, but for me that breaks the style, and it looks correct as it is currently:

Screenshot 2019-04-01 at 13 33 05

src/style/color-config/index.js
stevebutler2210: Yep, sorry updated now - put it there before realising it wasn't used elsewhere 👍

src/style/themes/classic/classic-theme.config.js
stevebutler2210: I've just moved the new one I added to the palette to a hard-coded definition in here as it wasn't being used elsewhere. The other ones I've used from the palette were already there, and the rest I hard-coded in here for now as they weren't being re-used.

stevebutler2210: I can move them all to one place or the other here if that would be preferred? Just seemed like the palette was meant only to be used for things which were being re-used across other styles.

Will take your lead on it though 👍 do you have a preference one way or the other?

stevebutler2210:
@andrewjtait margin corrected for viewing at 100% zoom. Thanks for helping get to the bottom of that!

stevebutler2210:
@marcinstachowicz has tested this functionality in `master` and confirmed that the required changes have already now been implemented elsewhere. As such I will close this PR and mark the ticket as "Done".

PR #1987, opened 2019-03-28
Extend the tooltip decorator

philearley:
# Description

* Allows users to interact with the tooltip. We do not hide the tooltip if the person driving the browser moves their cursor over into the tooltip, it will ensure the tooltip remains in view.

# Screenshots / Gifs

tooltip-decorator


src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -196,8 +198,12 @@ const TooltipDecorator = (ComposedComponent) => {
      * @return {void}
      */
     onHide = () => {
-      clearTimeout(this._tooltipTimeout);
-      this.setState({ isVisible: false });
+      clearTimeout(this._showTooltipTimeout);
+
+      this._hideTooltipTimeout = setTimeout(() => {
+        this.setState({ isVisible: false });
+        this.positionTooltip();
```
andrewjtait: I don't think you need to position the tooltip on hide

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -183,7 +183,9 @@ const TooltipDecorator = (ComposedComponent) => {
      * @return {void}
      */
     onShow = () => {
-      this._tooltipTimeout = setTimeout(() => {
+      clearTimeout(this._hideTooltipTimeout);
+
+      this._showTooltipTimeout = setTimeout(() => {
```
andrewjtait: as this property has been renamed, you should update or remove line 119

philearley:
Force pushed due to the original PR being off of master.

edleeks87:
will QA in staging environment with the tag

PR #1989, opened 2019-03-29
Cypress cucumber storybook

marcinstachowicz:
## Automation tests for Storybook components

**Added to testing scope:**
- Dialog
- Dialog Full Screen
- Draggable Context


cypress.env.json
```diff
@@ -5,5 +5,6 @@
   "Button": "http://localhost:9001/?path=/story/button--default",
   "Button Toggle":"http://localhost:9001/?path=/story/button-toggle--default",
   "Button Toggle Group": "http://localhost:9001/?path=/story/button-toggle-group--default",
-  "Dialog Full Screen": "http://localhost:9001/?path=/story/dialog-full-screen--default"
+  "Dialog Full Screen": "http://localhost:9001/?path=/story/dialog-full-screen--default",
+  "Dialog": "http://localhost:9001/?path=/story/dialog--default"
```
samhinton88: This file has a lot of repetition, and will keep growing as we add test suites. 

I would reorganise this so that there is only one `root` path: `expect(instance.formatValue()).toEqual('1,234,567.00')`, and have a function add the name of the story to the end.

cypress/integration/Dialog.feature
```diff
@@ -0,0 +1,116 @@
+Feature: Dialog component
+  I want to change Alert component properties
+
+  Background: Open Dialog component page
+    Given I open "Dialog" component page
+
+  @positive
+  Scenario Outline: Set height for Dialog dialog
```
samhinton88: I'm not sure how useful it is to test that a prop like this is set. 

It might be better to test that the element renders at the correct height?

cypress/integration/Dialog.feature
```diff
@@ -0,0 +1,116 @@
...
+      And I open component preview
+      And I hit ESC key
+    Then dialog is visible
+
+  @positive
+  Scenario: Enable escape key
+    When I uncheck disableEscKey
+      And I open component preview
+      And I hit ESC key
+    Then dialog is not visible
```
samhinton88: It would be good to see some tests that cover click actions, or some behaviour that mimics user interaction. 

Clicking outside the Dialogue should not, for example, close it - so that's something you could simulate with Cypress.

cypress.env.json
marcinstachowicz: fixed

cypress/integration/Dialog.feature
marcinstachowicz: When test will set height then it verifies if height is styled
here example
dialogPreview().should('have.attr', 'style').should('contain', `min-height: ${height}px`)
is this ok?

cypress/integration/Dialog.feature
marcinstachowicz: Thanks :) Scenarios for that cases was added

cypress/support/helper.js
```diff
@@ -1,6 +1,36 @@
...
-    cy.visit(Cypress.env(url))
-    knobsTab().click()
-}
\ No newline at end of file
+export function visitComponentUrl(component, suffix = 'default', iFrameOnly = false) {
+    cy.visit(prepareUrl(component, suffix, iFrameOnly))
+    if(!iFrameOnly) knobsTab().click()
+}
+
+export function dragAndDropForDraggableRecord(record, position) {
```
samhinton88: I'm wondering whether or not we could make this function more general, so it can be reused for other components / extended for dragging along the `X` axis.

marcinstachowicz: yes indeed, current sprint I am working on Configurable Items so I will make that method more general and reuse it

marcinstachowicz: I made general method for cypress/FE-1426 branch

PR #1992, opened 2019-04-02
Fe 1319

izakrzewska:
# Description
Message component refactored. Unit tests added.
Messages types in classic theme: info, deafult, maintenance, error, help, new, warning, success
Current message types: info, success, warning, error


src/style/themes/base/base-theme.config.js
```diff
@@ -21,9 +21,9 @@ export default (palette) => {
       // status
       error: palette.errorRed,
       focus: palette.gold,
-      info: palette.productBlueShade(3),
-      success: palette.brilliantGreenShade(20),
-      warning: palette.gold
+      info: palette.infoBlue,
```
samhinton88: These values, and the values in `style/colour-config/` should not be changed or renamed.

src/style/color-config/index.js
```diff
@@ -1,13 +1,16 @@
...
+  infoBlue: '0077C8',
   productGreen: '00A376',
   productBlue: '0077C8',
   navyBlue: '004B87',
   amethyst: '582C83',
   plum: '8246AF',
-  slate: '003349'
+  slate: '003349',
+  successGreen: '00B000',
+  warningOrange: 'E96400'
```
samhinton88: These base colours are used to create functions which build the palette that we use in our theming library and should not be changed or renamed.

customcodepl:
I think the `package.json` file should be removed f

src/components/message/MessageContent/message-content.component.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Content from '../../content/content';
+import MessageContentStyle from './message-content.style';
+
+const MessageContent = ({
+  as, title, transparent, children
+}) => {
+  return (
+    <MessageContentStyle type={ as } transparent={ transparent }>
```
samhinton88: We'll want this component to have a `data-component` attr., as per the original implementation. 

Our automation suites use them to locate nodes on the DOM.

https://github.com/Sage/carbon/blob/bb334d269d099c93955a4e931982d6f183a1726d/src/components/message/message.js#L142

src/components/message/CloseIcon/close-icon.style.js
```diff
@@ -0,0 +1,57 @@
...
+
+CloseIconContainerStyle.defaultProps = {
+  as: 'info',
+  roundedCorners: true,
+  theme: BaseTheme,
+  transparent: false
+};
+
+CloseIconContainerStyle.propTypes = {
+  as: PropTypes.string,
```
customcodepl: The same here. You can use `oneOf`

src/components/message/TypeIcon/type-icon.style.js
```diff
@@ -0,0 +1,82 @@
...
+  width: 30px;
+  text-align: center;
+  
+  ${({ roundedCorners }) => roundedCorners
+    && css`
+      border-radius: 3px 0 0 3px;
+    `}
+
+  ${({ theme, type, transparent }) => theme.name === THEMES.classic && stylingForClassic(type, transparent)}
+  ${({ theme, type, transparent }) => theme.name !== THEMES.classic && stylingForType(type, theme, transparent)}
```
customcodepl: `${({ theme, type, transparent }) => theme.name === THEMES.classic ? stylingForClassic(type, transparent) : stylingForType(type, theme, transparent);`

src/components/message/TypeIcon/type-icon.style.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+      }
+    `;
+  }
+
+  return css`
+    background-color: ${classicConfig[type].color};
+    span {
+      &:before {
+        color: #fff;
```
customcodepl: Hardcoded color

src/components/message/TypeIcon/type-icon.style.js
```diff
@@ -0,0 +1,82 @@
...
+  `;
+}
+
+function stylingForType(type, theme, transparent) {
+  if (transparent) {
+    return css`
+      background-color: #fff;
+      span {
+        &:before {
+          color: ${classicConfig[type].color};
```
customcodepl: Are you sure that `classicConfig`?

src/components/message/message-classic-config.style.js
```diff
@@ -0,0 +1,45 @@
...
+    color: '#ff7d00'
+  },
+  new: {
+    backgroundColor: '#f7f5fa',
+    borderColor: '#e0d6eb',
+    color: '#639'
+  },
+  transparent: {
+    backgroundColor: '#FFFFFF'
+  }
```
customcodepl: Hardcoded colors

src/components/message/message.style.js
```diff
@@ -0,0 +1,86 @@
...
+
+function stylingForClassic(type, transparent, border) {
+  if (transparent) {
+    return css`
+      border: none;
+      background-color: ${classicConfig.transparent.backgroundColor};
+    `;
+  }
+
+  if (!border) {
```
customcodepl: I think the better way is add border style when `border === true`.

benwilsonsage:
Hi Iga, looks great - just a few of points...

Message > Small/Medium/Large theme
Expected: Rounded corners option is only needed for 'Classic' theme.
Actual: Option present for Small/Medium/Large themes.

Message > Small/Medium/Large theme
Expected: Only success, error, warning, and info states are needed for Small/Medium/Large themes.
Actual: All legacy states available.

Message
Expected: Optional close icon has tap stop and focus state.
Actual: Unable to tab to close icon.

Message
Expected: If the container is very small, the dark coloured area containing icon remains fixed size, and the close icon has a fixed right margin.
Actual: Coloured area becomes smaller according to container size, and icon moves closer to container edge.
Screenshot 2019-04-02 at 18 06 47

Message
Expected: If filled with lots of content, the dark coloured area containing icon remains fixed size, and the close icon has a fixed right margin.
Actual: Coloured area becomes smaller according to container size, and icon moves closer to container edge.
Screenshot 2019-04-02 at 18 14 14

src/components/message/type-icon/type-icon.spec.js
```diff
@@ -0,0 +1,72 @@
...
+          expect(wrapper.toJSON()).toMatchSnapshot();
+        });
+      });
+    });
+
+    describe('when transparent prop is set to true', () => {
+      OptionsHelper.colors.forEach((messageType) => {
+        it(`applies white background and the type icon with the proper style applied for ${messageType}`, () => {
+          const wrapper = render({ transparent: true, type: messageType, theme: classicTheme });
+          expect(wrapper.toJSON()).toMatchSnapshot();
```
dquilter: Same again. You only need to test this once, I think? And, you could probably use the assertStyleMatch function, rather than a snapshot:

`import { assertStyleMatch } from '../../../__spec_helper__/test-utils';`

src/components/message/message.style.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable max-len */
```
dquilter: Is this needed?

izakrzewska: it's for 13 line (I kept getting maximum line length error)

dquilter: You can use the following just before the offending line:

`/* eslint-disable-next-line max-len */`

Which should just disable the rule on _that_ line.

src/components/message/type-icon/type-icon-classic.style.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default ({
+  theme, transparent, messageType, roundedCorners
+}) => theme.name === THEMES.classic
+  && css`
+    border-radius: ${roundedCorners ? '3px 0 0 3px' : '0px'};
+    background-color: ${transparent ? theme.colors.white : theme.colors[messageType][0]};
+    span {
+      &:before {
+        color: ${transparent ? theme.colors[messageType][0] : theme.colors.white};
```
samhinton88: We should probably avoid relying on index positions, I'm left wondering why theme.colors[whatever] is an array.

src/style/themes/classic/classic-theme.config.js
```diff
@@ -1,30 +1,37 @@
...
+      // generic
+      baseBlue: '#255BC7',
+      greyDarkBlue50: '#8099a4',
 
-        // generic
-        baseBlue: '#255BC7',
-        greyDarkBlue50: '#8099a4'
-      },
+      // messages status (main/background/border)
+      error: ['#c7384f', '#fdf5f5', '#f7d9d9'],
```
samhinton88: Why are these hex codes stored in an array like this?

src/components/message/close-icon/close-icon.style.js
```diff
@@ -0,0 +1,54 @@
...
+  background-color: ${({ messageType, theme }) => {
+    return theme.name === THEMES.classic ? theme.status[`${messageType}Background`] : theme.status.white;
+  }};
+  color: ${({ messageType, theme }) => {
+    return theme.name === THEMES.classic ? theme.status[messageType] : theme.status[messageType];
+  }};
+  span {
+    cursor: pointer;
+    &:before {
+      font-size: 16px;
```
customcodepl: Could you provide it from the theme?

src/components/message/message.style.js
```diff
@@ -0,0 +1,40 @@
...
+  ${({ theme, transparent, messageType }) => {
+    return theme.name !== THEMES.classic ? getBorderAndBackgroundStyles(theme, transparent, messageType) : null;
+  }}
+  ${({ theme }) => (theme.name === THEMES.classic ? messageClassicStyle : null)};
+`;
+
+function getBorderAndBackgroundStyles(theme, transparent, messageType) {
+  return `
+    border: ${transparent ? 'none;' : `1px solid ${theme.status[messageType]};`}
+    background-color: ${transparent ? `${theme.colors.white};` : `${theme.colors.white};`}
```
samhinton88: This ternary returns the same thing regardless of the result of the test expression (`transparent`).

src/style/themes/classic/classic-theme.config.js
```diff
@@ -1,30 +1,54 @@
...
+      // generic
+      baseBlue: '#255BC7',
+      greyDarkBlue50: '#8099a4'
+    },
 
-        // generic
-        baseBlue: '#255BC7',
-        greyDarkBlue50: '#8099a4'
-      },
+    status: {
```
samhinton88: Where have all of these colours come from? I feel like you should try bringing the classic theming config closer to the component.

src/components/message/message-classic.style.js
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components';
+import { THEMES } from '../../style/themes';
+import classicThemeColors from './message-classic-theme-colors';
+
+export default ({ theme, messageType }) => theme.name === THEMES.classic && css`
+  border: 1px solid ${classicThemeColors[messageType].borderColor};
+  background-color: ${classicThemeColors[messageType].backgroundColor};
+  border-radius: ${({ roundedCorners, border }) => (roundedCorners && border ? '3px;' : '0px;')};
+
+  ${({ border }) => !border && css`
```
dawidzarzycki: If we check the `false` we can rewrite to 
```js
${({ border }) => border || css
```

src/components/message/message.spec.js
```diff
@@ -0,0 +1,144 @@
...
+
+  describe('when onDismiss function is not provided', () => {
+    it('not renders the close icon when the onDismiss prop ', () => {
+      const wrapper = shallow(<Message>Message</Message>);
+      const closeIcon = wrapper.find(CloseIcon);
+      expect(closeIcon.length).toEqual(0);
+    });
+  });
+
+  describe('when onDismiss function is provided', () => {
```
dawidzarzycki: I think we can have here just one test about dissmiss and also instead of create everytime `shallow(<Message>Message</Message>)` we can create it in upper scope like
```js
let wrapper;
```
then use `beforeEach` to create the behavior for every test in the scope
```
beforeEach(() => {wrapper = shallow(<Message>Message</Message>)})
```
and if u need to set props then u can use `setProps` function
```js
wrapper.setProps({onDismiss})
```
Example:
```js
describe('Message', () => {
let wrapper;

beforeEach(()=>{
  wrapper = shallow(<Message>Message</Message>);
})

describe('when rendered', () => {
  it('should match the snapshot', () => {
  });
});

describe('when onDismiss function is not provided', () => {
  it('not renders the close icon when the onDismiss prop ', () => {
    const closeIcon = wrapper.find(CloseIcon);
      expect(closeIcon.length).toEqual(0);
    });
  });

describe('when onDismiss function is provided', () => {
it('should render the close icon', () => {
      const onDismiss = jest.fn();
      wrapper.setProps({onDissmiss})
      const closeIcon = wrapper.find(CloseIcon);
      expect(closeIcon.length).toEqual(1);
    });
  });
```

src/components/message/message-content/message-content.spec.js
```diff
@@ -0,0 +1,31 @@
...
+      expect(shallow(<MessageContent />)).toMatchSnapshot();
+    });
+  });
+
+  describe('when in classic theme', () => {
+    describe('when the title prop is provided', () => {
+      OptionsHelper.messages.forEach((messageType) => {
+        it('should render the title with a proper color applied', () => {
+          const wrapper = render({ theme: classicTheme, messageType, title: 'Example title' });
+          expect(wrapper.toJSON()).toMatchSnapshot();
```
grabkowski: you don't have to use `.toJSON` with the `TestRenderer`

src/components/message/message.spec.js
```diff
@@ -0,0 +1,143 @@
...
+      wrapper = shallow(<Message>Message</Message>);
+    });
+
+    it('should match the snapshot', () => {
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('does not render the close icon when onDismiss prop is not provided', () => {
+      const closeIcon = wrapper.find(CloseIcon);
+      expect(closeIcon.length).toEqual(0);
```
grabkowski: `expect(closeIcon.exists()).toEqual(flase)`

src/components/message/message.spec.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+
+    it('renders the close icon when onDismiss function is provided', () => {
+      const onDismiss = jest.fn();
+      wrapper.setProps({ onDismiss });
+      const closeIcon = wrapper.find(CloseIcon);
+      expect(closeIcon.length).toEqual(1);
```
grabkowski: `expect(closeIcon.exists()).toEqual(true)`

src/components/message/message.spec.js
```diff
@@ -0,0 +1,143 @@
...
+    });
```
grabkowski: we also need to test what happens when the `onDismiss` method is fired

src/components/message/message.spec.js
```diff
@@ -0,0 +1,143 @@
...
+      });
+    });
+  });
+
+  describe('when in classic mode', () => {
+    describe('when rendered', () => {
+      it('should match the snapshot', () => {
+        OptionsHelper.colors.forEach((messageType) => {
+          const wrapper = render({ theme: classicTheme, messageType });
+          expect(wrapper.toJSON()).toMatchSnapshot();
```
grabkowski: you don't have to use `.toJSON` with the `TestRenderer`

src/components/message/type-icon/type-icon.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      });
+    });
+  });
+
+  describe('when in classic theme', () => {
+    describe('when rendered', () => {
+      OptionsHelper.colors.forEach((messageType) => {
+        it(`should match the snapshot for ${messageType}`, () => {
+          const wrapper = render({ messageType, theme: classicTheme });
+          expect(wrapper.toJSON()).toMatchSnapshot();
```
grabkowski: you don't have to use `.toJSON` with the `TestRenderer`

PR #1894, opened 2019-01-11
FE-1187 add delete state to pill component

edleeks87:
# Description
Adds support for a 'close' Icon in the Pill component
Adds some specs to test new functionality

# Screenshots

- With no prop passed
screen shot 2019-01-15 at 14 31 30

- With prop passed and no cursor hover on icon
screen shot 2019-01-15 at 14 31 07

- With prop passed and cursor hover over icon
screen shot 2019-01-15 at 14 31 15

PR #1995, opened 2019-04-02
AC-4027 - re-introduces default line-height of 12px

stevebutler2210:
# Description
Adds default line-height style, which was removed in a previous PR (https://github.com/Sage/carbon/pull/1894/files), causing problems in some cases.


src/components/pill/pill.scss
```diff
@@ -6,6 +6,7 @@
   font-size: 12px;
   font-weight: 700;
   letter-spacing: 0.7px;
+  line-height: 12px;
```
andrewjtait: Although `12px` was the original value before the refactor, it looks like this is too low to resolve the requirement on the ticket of "Expected: Status indicator 'pills' are generally 21px in height.", it looks like `15px` works better.

I wonder if we should target this change against the release branch as well so we can ship it sooner.

PR #2000, opened 2019-04-05
Fe 1286 date input

grabkowski:
Experimental Date Input Component

Rewritten to work without Input Decorators


src/__experimental__/components/date/date-picker.component.js
```diff
@@ -0,0 +1,127 @@
...
+    top: inputRect.bottom + offsetY
+  };
+}
+
+/**
+ * Prevents propagation so date picker does not close on click inside the widget.
+ *
+ * @method handleWidgetClick
+ * @param {Object} ev event
+ * @return {void}
```
customcodepl: Can you remove these informations?

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -0,0 +1,127 @@
...
+import LocaleUtils from 'react-day-picker/moment';
+import Browser from '../../../utils/helpers/browser/browser';
+import DateHelper from '../../../utils/helpers/date/date';
+import Portal from '../../../components/portal/portal';
+import Navbar from './navbar';
+import StyledDayPicker from './day-picker.style';
+
+const DatePicker = (props) => {
+  const window = Browser.getWindow();
+  const [containerPosition, updateDatePickerPosition] = useState(getContainerPosition(window, props.inputElement));
```
andrewjtait: Just curious about what actually happens on each re-render if you use a variable as the initial value in `useState`? Does it ignore the `getContainerPosition` function on each re-render?

src/__experimental__/components/date/date-picker.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  it('should render a DayPicker component inside a Portal', () => {
+    expect(wrapper.find(Portal).contains(DayPicker)).toBe(true);
+  });
+
+  it('should pass proper disabledDays prop to the DayPicker component', () => {
+    const disabledDays = [{ before: moment('2019-02-02').toDate() }, { after: moment('2019-02-08').toDate() }];
+    expect(wrapper.find(DayPicker).props().disabledDays).toEqual(disabledDays);
+  });
+
+  describe('when an input field value is changed', () => {
```
samhinton88: Empty block.

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+
+  // Stores the document - allows us to override it different contexts, such as when running tests.
+  _document = document;
+
+  bridge = React.createRef() // this is a reference to the input decorator bridge component
+
+  blurBlocked = false; // stops the blur callback from triggering (closing the list) when we don't want it to
+
+  state = {
+    isDatePickerOpen: false,
```
samhinton88: Should be `datePickerIsOpen`.

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.blurBlocked && this.hasDatePickerValueChanged(prevProps)) {
+      this.unblockBlur();
+      this.handleBlur(); // TODO validate
+    }
+  }
+
+  hasDatePickerValueChanged = (prevProps) => {
```
samhinton88: Should be `datePickerValueHasChanged`

src/__experimental__/components/date/date-picker.component.js
grabkowski: the `useState` method sets the initial value of the state, so the `getContainerPosition` method is called only once. `getContainerPosition` returns a new object every time it's called, but even if we'll provide to the `setState` method an object defined as a variable it will deep copy that object, so further modifications should not change the original.

src/__experimental__/components/date/date-picker.spec.js
grabkowski: it's a draft, so tests are not yet completed

src/__experimental__/components/date/date.component.js
grabkowski: Do we have a name convention for that? I'm deliberately placing `is` or `has` as a prefix of the name to indicate that the value is a boolean.

samhinton88: Well it's not a convention really, it's the way it reads in plain english.

`isDatePickerOpen` suggests that the value will be a function:
```js
if( isDatePickerOpen ) { } // not good grammar

if( datePickerIsOpen ) { } // good grammar
```

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+    this.bridge.current.blockBlur = false; // this is to support the legacy behaviour in the bridge
+  }
+
+  handleBlur = (ev) => {
+    if (this.blurBlocked) return;
+    if (this.props.onBlur) this.props.onBlur(ev);
+  }
+
+  handleFocus = (ev) => {
+    if (this.blockFocus) {
```
samhinton88: I'm finding this a little hard to decode. Perhaps assign it in the same way that you are with `blurBlocked`, with a comment? 

I'm not clear from the name whether or not this would be a bool or a func., currently is reads like an instruction.

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+    const isoDate = convertToIsoDate(newValue);
+
+    this.setState({ selectedDate: isoDate });
+  };
+
+  emitOnChangeCallback = (stringDate) => {
+    this.props.onChange({ target: { value: stringDate } });
+  };
+
+  renderDatePicker = (passedProps) => {
```
samhinton88: Perhaps give this param a more descriptive name, something like `minMax`.

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+      };
+    }
+
+    return (
+      <StyledDateInput
+        onClick={ stopClickPropagation }
+        onKeyDown={ this.handleKeyDown }
+        role='presentation'
+      >
+        <InputDecoratorBridge
```
samhinton88: Do we want to remove the Bridge with this work?

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * To render the Date:
+ *
+ *   <Date name="myDate" />
+ *
+ * @class Date
+ * @constructor
+ * @decorators {Input,InputIcon,InputLabel,InputValidation}
+ */
+class Date extends React.Component {
```
samhinton88: Could this also be a candidate for a functional component?

src/__experimental__/components/date/date.component.js
grabkowski: I think that we should put more focus on code readability than the english gramatics. When you're reviewing code `isDatePickerOpen` on the first sight is more suggestive that's a boolean than `datePickerIsOpen`. 
If `isDatePickerOpen` would be a function (assuming that it returns a boolean and it has no side effects), then we could treat it as boolean and we'll not have to think about what that method is doing.

samhinton88: I commented on it because as I read the code I didn't think the term had good code readability, because when you read it on its own it looks like a function (which it isn't) and when you read it in the context of an `if` statement it doesn't read like English.

samhinton88: A compromise might be to leave out the `is` entirely, as `datePickerOpen` is more suggestive of a bool, and `if( datePickerOpen )` is more idiomatically English.

src/__experimental__/components/date/date.component.js
grabkowski: we could pass any prop later, for now it's only `minDate` and `maxDate`

src/__experimental__/components/date/date.component.js
grabkowski: you're right, it's a leftover from the previous Decorator based implementation, I'll change it.

src/__experimental__/components/date/date.component.js
samhinton88: Which other props will be passed?

src/__experimental__/components/date/date.component.js
grabkowski: Yes it could, but I've already extracted the `DatePicker` part from that component and I've opted to leave it as it was, because of the amount of work that should be done to rewrite it to functional.

src/__experimental__/components/date/date.component.js
grabkowski: It's a common convention to prefix booleans with `is` or `has`, I have used it in my previous projects and a lot of people is using that convention - for example the Enzyme team. 

As for the leaving out the `is` - `datePickerOpen` could be a string and it will be valid in a conditional.

src/__experimental__/components/date/date.component.js
grabkowski: we could pass other DatePicker config props needed to render a proper DatePicker, depending on the future requirements.

src/__experimental__/components/date/date.component.js
samhinton88: I think it is conventional for method names and attributes on an instance: `person.isAllve` or `expect(foo).toHaveBeenCalled()` are reasonable, `toHaveBeenCalled(foo)` is less reasonable. 

I don't see how `datePickerOpen` could imply a string type.

src/__experimental__/components/date/date.component.js
grabkowski: we need validations in this component

andrewjtait: Validations are almost finished https://github.com/Sage/carbon/pull/2002

Your Date component will inherit these automatically, we should avoid any further reliance on `InputDecoratorBridge`

src/__experimental__/components/date/date.component.js
samhinton88: Presumably that future requirement would require us to make changes to this file, though? For quick understanding of how the Component works currently, it might be better to give this a more descriptive name.

src/__experimental__/components/date/date.component.js
andrewjtait: (same applies to `Number`, which is already in master - we will need to remove it)

src/__experimental__/components/date/date.component.js
grabkowski: what do you think about `dateLimitProps`?

src/__experimental__/components/date/date.component.js
andrewjtait: We can QA the validations as part of the validations ticket, they should work across all input components.

grabkowski: ok, I'll swap it for the Textbox then

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,259 @@
...
+  assignInput = (input) => {
+    this.input = input.current;
+  };
+
+  blockBlur() {
+    this.blurBlocked = true;
+    this.bridge.current.blockBlur = true; // this is to support the legacy behaviour in the bridge
+  }
+
+  unblockBlur() {
```
samhinton88: You could refactor these two methods into `setBlurBlock`, and pass in a bool.

grabkowski: these methods will be removed with the bridge removal

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,242 @@
...
+
+    this.isBlurBlocked = true;
+    this.closeDatePicker();
+    this.updateVisibleValue(selectedDate);
+    this.emitOnChangeCallback(stringDateIso);
+  };
+
+  updateVisibleValue = (date) => {
+    const formattedDateString = formatVisibleValue(date);
+    this.setState({ visibleValue: formattedDateString });
```
dquilter: Little refector:

```
const visibleValue = formatVisibleValue(date);
this.setState({ visibleValue });
```

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,242 @@
...
+      ...dateRangeProps
+    };
+
+    return <DatePicker { ...datePickerProps } />;
+  }
+
+  render() {
+    const { minDate, maxDate, ...inputProps } = this.props;
+    let events = {};
+    delete inputProps.autoFocus;
```
dquilter: Seems like a lot of messing around with this `autoFocus` just for it to get deleted on the first render? Maybe I'm missing the point?

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+import { mount } from 'enzyme';
+import Date from './date.component';
+import DatePicker from './date-picker.component';
+import Textbox from '../textbox';
+
+describe('Date', () => {
+  let wrapper;
+
+  describe('when the Component is rendered', () => {
+    it("then the Textbox component should be rendered as it's descendant", () => {
```
dquilter: Reads a little better:

```
describe('when the Component is rendered', () => {
  it("should have the Textbox component rendered as it's descendant", () => {
```

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+    it("then component's input should be focused after render", () => {
+      wrapper = render({ autoFocus: true });
+      const input = wrapper.find('input');
+      const focusedElement = document.activeElement;
+      expect(input.instance()).toBe(focusedElement);
+    });
+  });
+
+  describe('when autoFocus prop is not defined', () => {
+    it('does not sets focus on the input', () => {
```
dquilter: Not sure this tests is needed? Does it add any value in terms of coverage?

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+      });
+    });
+  });
+
+  describe('when the "blur" event is triggered on the input', () => {
+    let onBlurFn;
+
+    beforeEach(() => {
+      onBlurFn = jest.fn();
+      wrapper = render({ onBlur: onBlurFn });
```
dquilter: Define onBlurFn as a `const` with name `onBlur` for `render({ onBlur })`.

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+    });
+  });
+
+  describe('when the Component is updated', () => {
+    const firstDate = '12/08/2019';
+    const secondDate = '17/08/2019';
+    let onBlurFn;
+
+    beforeEach(() => {
+      onBlurFn = jest.fn();
```
dquilter: As above.

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+    it('then the input element should be updated to reflect the passed date', () => {
+      expect(wrapper.update().find('input').prop('value')).toBe(getFormattedDate(mockDate));
+    });
+  });
+
+  describe('when the input value is changed', () => {
+    let onChangeFn;
+
+    beforeEach(() => {
+      onChangeFn = jest.fn();
```
dquilter: Same again.

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+    beforeEach(() => {
+      wrapper = render({});
+      simulateFocusOnInput(wrapper);
+      wrapper
+        .find(DatePicker)
+        .props()
+        .handleDateSelect(mockDate);
+    });
+
+    it('then the DatePicker component should be removed', () => {
```
dquilter: `it('should not contain the DatePicker component'`

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+      expect(
+        wrapper
+          .update()
+          .find(DatePicker)
+          .exists()
+      ).toBe(false);
+    });
+
+    it('then the input element should be updated to reflect the passed date', () => {
```
dquilter: `it(' should update the input element to reflect the passed date', () => {`

src/__experimental__/components/date/date.spec.js
```diff
@@ -0,0 +1,333 @@
...
+  describe.each(['disabled', 'readOnly'])('when the "%s" prop is set', (prop) => {
+    const validDate = '1 apr 2019';
+    let onChangeFn;
+    let onFocusFn;
+    let onBlurFn;
+
+    beforeEach(() => {
+      onChangeFn = jest.fn();
+      onFocusFn = jest.fn();
+      onBlurFn = jest.fn();
```
dquilter: And again (`onBlurFn` -> `const onBlur`).

src/__experimental__/components/date/date.stories.js
```diff
@@ -0,0 +1,65 @@
...
+import { StateDecorator, Store, State } from '@sambego/storybook-state';
+import {
+  boolean,
+  number,
+  text
+} from '@storybook/addon-knobs';
+import DateInput from './date.component';
+import Textbox from '../textbox';
+import getTextboxStoryProps from '../textbox/textbox.stories';
+import notes from './documentation/notes.md';
```
dquilter: There seems to be a push to use an `index.js` in the documentation for Storybook.

src/__experimental__/components/date/date-picker.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      const disabledDays = [{ before: moment(firstDate).toDate() }, { after: moment(secondDate).toDate() }];
+      expect(wrapper.find(DayPicker).props().disabledDays).toEqual(disabledDays);
+    });
+  });
+
+  describe('when the "onDayClick" prop have been triggered', () => {
+    let handleDateSelectFn;
+
+    beforeEach(() => {
+      handleDateSelectFn = jest.fn();
```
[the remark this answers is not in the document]
grabkowski: > Define with `const` here rather than `let` outside the `beforeEach`?
> 
> Doesn't look like you're redefining this anywhere?

if we'll define it with a `const`, then we'll share the same function between tests, which could lead to false positives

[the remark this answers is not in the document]
grabkowski: > Also, minor thing, but could rename to `handleDateSelect` for simpler syntax in your `render` call on line 75.

I like to emphasize that it's a jest mocked function

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,242 @@
...
+
+  state = {
+    isDatePickerOpen: false,
+    selectedDate: null,
+    visibleValue: formatVisibleValue(this.props.value)
+  };
+
+  componentDidMount() {
+    if (this.props.autoFocus) {
+      this.isAutoFocused = true;
```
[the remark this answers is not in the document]
grabkowski: it's needed to focus on the input field when the component is mounted. We don't need in in `state`, as it's evaluated only once on mount.

src/__experimental__/components/date/date.component.js
grabkowski: these props are passed down to the `Textbox` component, we don't need that property there as we are using it directly in the `Date` component

src/__experimental__/components/date/date.spec.js
grabkowski: 1.Yes, this test is needed, we are expecting the component to behave differently on mount depending on that prop.
2.Yes, coverage also includes `else` conditions.

src/__experimental__/components/date/date.spec.js
grabkowski: a `const` will be shared between tests and that could lead to false positives in assertions.

src/__experimental__/components/date/date.component.js
grabkowski: `isOpen` would suggest that the `Date` component have a state open, but this state is responsible for rendering of the `DatePicker` component.

src/__experimental__/components/date/date.stories.js
grabkowski: fixed

src/__experimental__/components/date/date.component.js
grabkowski: Is there a reason why it should be changed? That method could be treated as a question which makes more sense in my opinion, as the returned value is a boolean.

src/__experimental__/components/date/date.component.js
```diff
@@ -0,0 +1,225 @@
...
+
+    events = {
+      onBlur: this.handleBlur,
+      onChange: this.handleVisibleInputChange,
+      onFocus: this.handleFocus,
+      onKeyDown: this.handleKeyDown
+    };
+
+    return (
+      <StyledDateInput
```
dawidzarzycki: Shouldn't we have here a `data-component` prop?

src/__experimental__/components/date/date.component.js
edleeks87: I think the consensus is that we'll refactor components to functional in future tickets and focus on implementing the style changes for now

src/__experimental__/components/date/date-picker.component.js
andrewjtait: Looking into it it looks as though `getContainerPosition` is called on every new render (eg. when a prop or state changes). I think you need to wrap it in a callback to trigger only once

```js
const [containerPosition, updateDatePickerPosition] = useState(() => getContainerPosition(window, props.inputElement))
```

src/__experimental__/components/date/date.component.js
customcodepl: It can be helpful https://wiki.eclipse.org/Recommenders/CodingConventions#is_prefix_should_be_used_for_boolean_variables_and_methods.

PR #2006, opened 2019-04-05
FE-1256 split button dls

edleeks87:
# WORK IN PROGRESS
# Description
Updates component to be DLS compliant and adds new functionality such as key navigation of the listed additional buttons and the list opening on focus


src/components/button/button-sizes.style.js
```diff
@@ -0,0 +1,21 @@
...
+  `,
+  medium: `
+    font-size: ${text.size};
+    max-height: 40px;
+    height: 40px;
+    padding-left: 24px;
+    padding-right: 24px;
+  `,
+  large: `
+    font-size: 16px;
```
izakrzewska: isn't it possible to get this value from theme config?

src/components/button/button.component.js
```diff
@@ -0,0 +1,93 @@
...
+  theme: PropTypes.oneOf(OptionsHelper.buttonColors),
+  /** Used to transfrom button into anchor */
+  to: PropTypes.string
+};
+
+Button.defaultProps = {
+  as: 'secondary',
+  size: 'medium',
+  disabled: false,
+  iconPosition: '',
```
izakrzewska: why are all of these props empty strings by default?

PR #2016, opened 2019-04-16
Fe 1424 - Google Analytics Double Counting

izakrzewska:
# Description
Analytics is tracked only when React Router triggers the URL change (not a true page load, just a SPA page load).


change_log/next/FE-1424-google-analytics-bug.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "prvent Google Analytics double counting"
```
andrewjtait: typo on "Prevent"

PR #2018, opened 2019-04-16
Dropdown-filter empty visibleValue fix

blaky:
# Description
There was an issue in the dropdown-filter component: When the `visibleValue` property changed externally the displayed value remained unchanged because the `filter` state property is used for rendering (see `dropdown-filter.js:365`). So when the `visibleValue` property is changed, the filter value gets updated, just like in the constructor. 
# TODO
- [x] Release notes


src/components/dropdown-filter/__spec__.js
```diff
@@ -72,6 +72,26 @@ describe('DropdownFilter', () => {
...
+        instance = TestUtils.renderIntoDocument(
+          <DropdownFilter {...props} />
+        );
+
+        expect(instance.state.filter).toEqual('abc');
+        instance.componentWillReceiveProps({ ...props, visibleValue: 'def' });
+        expect(instance.state.filter).toEqual('def');
+      });
+    });
+  });
```
grabkowski: Component's internal implementation shouldn't be tested.
You could use Enzyme `shallow` or `mount` wrapper and the `setProps` method instead of directly triggering `componentWillReceiveProps` on an instance. You should also check something that depends on the state instead of the state directly on an instance.

grabkowski: Component's internal implementation shouldn't be tested.
You could use Enzyme `shallow` or `mount` wrapper and the `setProps` method instead of directly triggering `componentWillReceiveProps` on an instance. You should also check something that depends on the state instead of the state directly on an instance.

blaky:
Hi @grabkowski ,

The test is updated as you requested it. Please note that all other test in that particular test suite calls internal methods, checks internal properties.

grabkowski:
> Hi @grabkowski ,
> 
> The test is updated as you requested it. Please note that all other test in that particular test suite calls internal methods, checks internal properties.

Thank you.
I know about that issue, and a lot of refactor will be needed in the future. 
For now we should write proper tests instead of following these bad examples.

blaky:
Hi @andrewjtait, 

Can we get this merged and released soon please?

Thanks,

blaky:
> > Hi @grabkowski ,
> > The test is updated as you requested it. Please note that all other test in that particular test suite calls internal methods, checks internal properties.
> 
> Thank you.
> I know about that issue, and a lot of refactor will be needed in the future.
> For now we should write proper tests instead of following these bad examples.

@grabkowski  Yes, I completely agree, thanks for the careful review.

andrewjtait:
This should target release in order to release it faster, I've reopened it here https://github.com/Sage/carbon/pull/2047

PR #2022, opened 2019-04-17
FE-1315 - Confirm component (styles)

izakrzewska:
# Description
SCSS removed, styled components added.


src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,80 @@
...
+import Dialog from '../dialog/dialog.component';
+import Button from '../button';
+import StyledConfirmButtons from './confirm.style';
+
+class Confirm extends Dialog {
+  /**
+   * Returns main classes for the component combined with
+   * dialog main classes.
+   *
+   * @method mainClasses
```
dawidzarzycki: I think we dont need this anymore

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,80 @@
...
+  componentTags(props) {
+    return {
+      'data-component': 'confirm',
+      'data-element': props['data-element'],
+      'data-role': props['data-role']
+    };
+  }
+}
+
+Confirm.defaultProps = assign({}, Dialog.defaultProps, {
```
dawidzarzycki: Can we use spread operator here?

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,80 @@
...
+  size: 'extra-small',
+  showCloseIcon: false
+});
+
+Confirm.propTypes = assign({}, Dialog.propTypes, {
```
dawidzarzycki: Can we use spread operator here?

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,76 @@
...
+import Button from '../button';
+import StyledConfirmButtons from './confirm.style';
+
+class Confirm extends Dialog {
+  /**
+   * Returns main classes for the component combined with
+   * dialog main classes.
+   */
+  get mainClasses() {
+    return classNames(super.mainClasses);
```
samhinton88: I'm not sure we need these class names any more.

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,76 @@
...
+Confirm.defaultProps = {
+  ...Dialog.defaultProps,
+  size: 'extra-small',
+  showCloseIcon: false
+};
+
+Confirm.propTypes = {
+  ...Dialog.propTypes,
+  /**
+   * A custom event handler when a confirmation takes place
```
samhinton88: We can single-line these comments.

src/components/confirm/confirm.style.js
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+
+const StyledConfirmButtons = styled.div`
+  margin-top: 20px;
+  display: flex;
+  justify-content: flex-end;
+
+  .carbon-button {
```
samhinton88: I'm not sure we want to use these class-names here, as they may no longer be on the button.

src/components/confirm/confirm.component.js
stevebutler2210: @samhinton88 single-lining these prevents the descriptions being pulled in by StoryBook

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,76 @@
...
+   */
+  additionalContent() {
+    return (
+      <StyledConfirmButtons>
+        <Button
+          onClick={ this.props.onConfirm }
+          data-element='confirm'
+          as='primary'
+        >
+          {this.props.confirmLabel || I18n.t('confirm.yes', { defaultValue: 'Yes' })}
```
stevebutler2210: Could this be set in defaultProps?

src/components/confirm/confirm.component.js
```diff
@@ -0,0 +1,76 @@
...
+        </Button>
+        <Button
+          onClick={ this.props.onCancel }
+          data-element='cancel'
+          as='secondary'
+        >
+          {this.props.cancelLabel || I18n.t('confirm.no', { defaultValue: 'No' })}
```
stevebutler2210: Could this be set in defaultProps?

izakrzewska: I guess so but the scope of the ticket was to change the styling only. Changing that part will require changing spec files too so I guess it would be better to leave that for times when the whole component is refactored. What do you think?

src/components/confirm/confirm.component.js
izakrzewska: same as below

src/components/confirm/confirm.component.js
izakrzewska: The scope of the ticket was to change the styling only. I think that all of these classnames will be removed when the component itself is rewritten.

src/components/confirm/confirm.component.js
stevebutler2210: Makes sense, do we have a separate ticket on the backlog for the refactor?

stevebutler2210: If so I'll mark this as resolved 👍

izakrzewska: We do - that's the ticket - https://jira.sage.com/browse/FE-1119

src/components/confirm/confirm.style.js
izakrzewska: I changed the button to styled button. However, new implementation of Button component is not merged to master yet. I'll add 'DO NOT MERGE' label on this PR until then. Is that okay?

src/components/confirm/confirm.component.js
edleeks87: In general I think it's safe to remove any code that the addition of styled components will replace, whilst avoiding rewriting the implementation. However, as this is an inherited component I'd be inclined to leave it for the future ticket

PR #2025, opened 2019-04-18
adding data-element attribute to confirm dialog

DlgSHi:
# added data-element attribute to confirm dialog


src/components/confirm/confirm.stories.js
```diff
@@ -29,7 +29,7 @@ storiesOf('Confirm', module)
   })
   .add('default', () => {
     button('open', handleOpen);
-    const children = text('children', 'This is an example of a confirm.');
+    const children = <span data-element='confirm-children'>This is an example of a confirm.</span>;
```
grabkowski: could you move that directly to the JSX below?

DlgSHi: Thanks @grabkowski, fixed :)

PR #2030, opened 2019-04-23
Cypress/fe 1458

marcinstachowicz:
## Automation tests for Storybook components

**Added to testing scope:**
- Detail,
- Dropdown - not all test cases because of this component will be deprecated and replaced by Select

**Improvements:**
- move reusable locators to common locators
- move reusable steps to common steps
- minor readme update

### DIFF
--- a/cypress/integration/dropdown.feature
+++ b/cypress/integration/dropdown.feature
@@ -75,7 +75,8 @@ Feature: Dropdown component
       | fieldHelp                |
       | Example Test             |
       | 1!@#$%^*()_+-=~[];:.,?{} |
-      | 汉字                       |
+      | áéíóú¿¡üñ                |
+      | ÄÖÜßäöüß                 |
       | <>                       |
 
   @positive
--- a/cypress/integration/dropdown.feature
+++ b/cypress/integration/dropdown.feature
@@ -1,0 +1,89 @@
+# This componet will be depecated and replaced by Select,
+# that's why not all cases are covered and common methods are not isolated.
+Feature: Dropdown component
+  I want to change Dropdown component properties
+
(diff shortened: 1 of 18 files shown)


cypress/integration/dropdown.feature
```diff
@@ -0,0 +1,89 @@
...
+      | <>                      |
+
+  @positive
+  Scenario Outline: Change Dropdown component field help
+    When I set field help to "<fieldHelp>"
+    Then Field help on preview is set to "<fieldHelp>"
+    Examples:
+      | fieldHelp                |
+      | Example Test             |
+      | 1!@#$%^*()_+-=~[];:.,?{} |
```
DlgSHi: maybe we should use German / Spanish special characters for all input ?

cypress/locators/dropdown/index.js
```diff
@@ -0,0 +1,12 @@
+import { FORM } from '../locators';
+import {
+  DROPDOWN_INPUT, DROPDOWN_COMPONENT, DROPDOWN_LABEL,
+} from './locators';
+
+// knobs locators
+export const readOnlyCheckbox = () => cy.get(FORM).contains('readOnly').find('input');
```
DlgSHi: should we move readOnly to common staff ?

marcinstachowicz: readOnlyCheckbox is used only for Dropdown component, in my opinion current location is correct

DlgSHi: good point

cypress/integration/dropdown.feature
marcinstachowicz: fixed

PR #2041, opened 2019-04-29
FE-1196 - Fix timezone date format bug by adding conditional formatting

edleeks87:
# Description
The original attempt to format the date correctly based on locale caused issue elsewhere. Therefore, I have added a boolean condition to `formatDateString` that fixes the original bug and prevents the issues it attempted fix created

### DIFF
--- a/src/utils/helpers/date/date.js
+++ b/src/utils/helpers/date/date.js
@@ -62,12 +62,12 @@ const DateHelper = {
    * @method formatDateString
    * @param {String} value current value e.g. Wed Aug 23 2017 12:00:00 GMT+0100 (BST)
    * @param {String} formatTo Desired format e.g. YYYY-MM-DD
+   * @param {boolean} utcFormat set whether date should be utc formatted
    * @return {String} formatted date
    */
-  formatDateString: (value, formatTo) => {
-    return (
-      moment(new Date(value).getTime()).format(formatTo)
-    );
+  formatDateString: (value, formatTo, utcFormat = true) => {
+    const momentMethod = utcFormat ? moment.utc : moment;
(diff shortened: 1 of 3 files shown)


src/utils/helpers/date/date.js
```diff
@@ -62,12 +62,12 @@ const DateHelper = {
...
    * @param {String} value current value e.g. Wed Aug 23 2017 12:00:00 GMT+0100 (BST)
    * @param {String} formatTo Desired format e.g. YYYY-MM-DD
+   * @param {boolean} utcFormat set whether date should be utc formatted
    * @return {String} formatted date
    */
-  formatDateString: (value, formatTo) => {
-    return (
-      moment(new Date(value).getTime()).format(formatTo)
-    );
+  formatDateString: (value, formatTo, utcFormat = true) => {
```
dawidzarzycki: If we have two returns next to each other maybe it should be good to have something like 
```js
if (utcFormat) { 
  return moment(new Date(value).getTime()).format(formatTo)
};

return moment.utc(new Date(value).getTime()).format(formatTo);
```
What do you think about this?

edleeks87: Yeah it's more readable in this scenario I suppose, change made

src/utils/helpers/date/date.js
```diff
@@ -62,12 +62,14 @@ const DateHelper = {
...
+    if (utcFormat) {
+      return moment(new Date(value).getTime()).format(formatTo);
+    }
+    return moment.utc(new Date(value).getTime()).format(formatTo);
```
andrewjtait:
```
const momentMethod = utcFormat ? moment.utc : moment;
return momentMethod(new Date(value).getTime()).format(formatTo);
```

PR #2042, opened 2019-04-29
Extend table ajax for post

philearley:
# Description

Extends TableAjax to allow for post requests.

Followed pattern of mocking superagent for post and get based on another Sage repository. Pending comments from reviewers, I could change the mocks to match? Unless there is another way of testing `Request.post` was called.

### DIFF
--- a/src/components/table-ajax/table-ajax.js
+++ b/src/components/table-ajax/table-ajax.js
@@ -114,7 +114,15 @@ class TableAjax extends Table {
      * @property onAjaxError
      * @type {Function}
      */
-    onAjaxError: PropTypes.func
+    onAjaxError: PropTypes.func,
+
+    /**
+     * A prop to allow the override of the default get request and perform a post.
+     * @property postAction
+     * @type {Boolean}
+
+     */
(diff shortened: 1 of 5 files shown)


src/components/table-ajax/table-ajax.js
```diff
@@ -343,15 +351,27 @@ class TableAjax extends Table {
...
-      this._request = Request
-        .get(this.props.path)
-        .set(this.getHeaders())
-        .query(this.queryParams(element, options))
-        .end((err, response) => {
-          this._hasRetreivedData = true;
-          this.handleResponse(err, response);
-          if (resetHeight) { this.resetTableHeight(); }
-        });
+      if (this.props.postAction) {
```
samhinton88: This could be refactored, given that the only difference is the verb.

philearley:
Testing done with @emilybarr with Carbon code loaded into a local repository.

Confirmed existing usage of get in a TableAjax still works. Confirmed changing values returns success.

Confirmed in a new TableAjax using the new param, we can see the successful post request returning data.

Issue #1671, opened 2018-01-17
Missing translation in DropdownFilter with no options available


marioabreu:
Dropdown Filter is showing a missing translation when first click and there are no options available to show.

![36n2r6ulry](https://user-images.githubusercontent.com/763907/35041404-95ebcc28-fb7c-11e7-8c3b-fbd6d1fd43c3.gif)

Dropdown Filter should not show any message when there are no options available **and** the user doesn't type anything. The message should only appear after the user starts to type.

from the discussion, this could be the fix:
```
if (!items.length && typeof this.state.filter === 'string') {
```
**Source:** https://github.com/Sage/carbon/blob/v3.0.0-rc1/src/components/dropdown-filter/dropdown-filter.js#L303

PR #2043, opened 2019-04-30
S1T-1855: Fixed issue when no search term set

chriswaymansage:
# Description

Fixes an issue when no search term set.  An empty string is now returned when no search term is set as 'term' cannot be null for the i18n string interpolation.

https://jira.sage.com/browse/S1T-1855
Fixes https://github.com/Sage/carbon/issues/1671

# Screenshots / Gifs

## Before

image

## After

image

# Testing Instructions

When no search term has been entered in a dropdown-filter component so i18n string interpolation error should be apparent.


src/components/dropdown-filter/dropdown-filter.js
```diff
@@ -307,7 +307,7 @@ class DropdownFilter extends Dropdown {
           {
             I18n.t('dropdownlist.no_results', {
               defaultValue: 'No results match "%{term}"',
-              term: this.state.filter
+              term: this.state.filter || ''
```
philearley: I think there should be a unit test for this.

philearley:
Testing that was done for this component: (Taken from internal ticket)

Tested DropdownFilter component with no data being returned and confirmed the translation is not missing for term.

Tested DropdownFilter component with data being returned and confirmed the data is visible to select from.

PR #2044, opened 2019-04-30
Enforce a body with post requests

philearley:
# Description

This enforces when the TableAjax is used with a post request:
* It does not format the filter as it is not needed for a body.
* It sends the data instead of adding the filter as a query.

Extends: https://github.com/Sage/carbon/pull/2042

The above works for small data sets, but as this is designed to allow larger request payloads, superagent has `.send()` option to build a body separate to `.query()`.

# Testing Instructions

Confirm when doing a get request it builds the url.
Confirm when doing a post request it passes the query params into the body of the request.


philearley:
Tested with @emilybarr 

A TableAjax performing a get request is successfully returning data when interacting with values, no errors in console. The query params are still being serialized.

A TableAjax performing a post request is successfully returning data when interacting with values, no errors in console. The payload is correctly not serializing.

PR #2052, opened 2019-05-07
Fixed wrong styling in experimental datepickers

grabkowski:

### DIFF
--- a/src/__experimental__/components/date/date-picker.component.js
+++ b/src/__experimental__/components/date/date-picker.component.js
@@ -8,6 +8,7 @@ import Browser from '../../../utils/helpers/browser/browser';
 import DateHelper from '../../../utils/helpers/date/date';
 import Portal from '../../../components/portal/portal';
 import Navbar from './navbar';
+import Weekday from './weekday/weekday.component';
 import StyledDayPicker from './day-picker.style';
 
 const DatePicker = (props) => {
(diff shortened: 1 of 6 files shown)


src/__experimental__/components/date/day-picker-classic.style.js
```diff
@@ -2,11 +2,13 @@ import { css } from 'styled-components';
...
-  font-weight: 700;
-  margin-top: 0;
-  padding: 10px;
-  font-weight: 600;
-  color: ${theme.text.color};
+  .DayPicker {
+    font-weight: 700;
+    margin-top: 0;
+    padding: 10px;
+    font-weight: 600;
```
izakrzewska: double font-weight

grabkowski: removed

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -8,6 +8,7 @@ import Browser from '../../../utils/helpers/browser/browser';
 import DateHelper from '../../../utils/helpers/date/date';
 import Portal from '../../../components/portal/portal';
 import Navbar from './navbar';
+import Weekday from './weekday/weekday.component';
```
dawidzarzycki: If we have index file we can just simply 
`import Weekday from './weekday';`

PR #2058, opened 2019-05-08
Fe 1503 - Create component

dawidzarzycki:
# Description
Create component and Link component has been changed

# Testing Instructions
Link component should be re-tested also


src/components/create/create.component.js
```diff
@@ -0,0 +1,41 @@
...
+import CreateStyle from './create.style';
+
+class Create extends React.Component {
+  constructor(args) {
+    super(args);
+    this.linkProps = this.linkProps.bind(this);
+  }
+
+  /** Returns the props for the component. */
+  linkProps() {
```
dquilter: If you turn this into an arrow function, you can delete the class `constructor`.

src/components/create/create.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+  beforeEach(() => {
+    wrapper = render({});
+  });
+
+  it('should render correctly', () => {
+    expect(TestRenderer.create(<CreateStyle />)).toMatchSnapshot();
+  });
+
+  it('should render correct', () => {
```
dquilter: Spelling on "correctly" in incorrect. I assume you've done that on purpose because of the two similar tests?

I don't think you need both of those snapshots, do you?

src/components/create/create.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+  describe('when classic style has been provided', () => {
+    it('should apply custom styling ', () => {
+      wrapper.setProps({ theme: 'classic' });
+
+      expect(wrapper).toMatchSnapshot();
+    });
+  });
+
+  describe('when `custom class` is privded', () => {
```
dquilter: Spelling "provided".

src/components/create/create.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    it('component should contain correct className', () => {
```
dquilter: This might read better as:
describe - When custom class is provided to component
it - should contain correct class name

src/components/create/create.style.js
```diff
@@ -0,0 +1,37 @@
...
+    border: 1px dashed ${({ theme }) => theme.colors.border};
+    display: block;
+    padding: 12px 12px 10px;
+    text-align: center;
+    font-weight: 700;
+    text-decoration: none;
+    width: 100%;
+
+    :hover{
+        background-color: ${({ theme }) => theme.colors.white};
```
dquilter: Is your indentation/spacing incorrect in this file?

src/components/link/link.style.js
```diff
@@ -20,20 +18,28 @@ a {
     &:focus {
       color: ${theme.text.color};
       cursor: default;
-      background-color: ${theme.colors.focus};
+      background-color: ${theme.colors.focusedLink};
```
dquilter: Strange variable name, I'd expect `theme.colors.focusedLink` to affect text colour.

src/components/create/create.spec.js
dawidzarzycki: The first one is for the <CreateStyle>, the second one is for <Create>

dquilter: I think you probably just need to update the `it`s to `should render Create / CreateStyle correctly`.

src/components/create/create.component.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import tagComponent from '../../utils/helpers/tags/tags';
+import CreateStyle from './create.style';
+
+class Create extends React.Component {
```
stevebutler2210: Does this need to be a Class-based component?

src/components/create/create.component.js
```diff
@@ -3,26 +3,24 @@ import PropTypes from 'prop-types';
 import tagComponent from '../../utils/helpers/tags/tags';
 import CreateStyle from './create.style';
 
-class Create extends React.Component {
+const Create = (props) => {
   /** Returns the props for the component. */
-  linkProps = () => {
-    const { ...props } = this.props;
+  const linkProps = () => {
```
stevebutler2210: Sorry, looking at this again, does this even need to be a function? Looks like something like this could work:
```
const linkProps = {
  iconAlign: 'right',
  icon: 'add',
  ...props
}
```

dawidzarzycki: You're right, updated! :)

PR #2064, opened 2019-05-15
add form component automation tests

marcinstachowicz:
# add form component automation tests


cypress/integration/form.feature
```diff
@@ -0,0 +1,131 @@
...
+  Scenario: Disable show summary
+    When I check showSummary checkbox
+      And I uncheck showSummary checkbox
+      And I save form
+    Then summary is not visible
+
+  @positive
+  Scenario: Check validation
+    When I save form
+    Then input is validation field
```
DlgSHi: Then input is validated

cypress/support/step-definitions/form-steps.js
```diff
@@ -0,0 +1,75 @@
...
+Then('cancel button is visible', () => {
+  cancelButton().should('be.visible');
+});
+
+Then('cancel button is not visible', () => {
+  cancelButton().should('not.exist');
+});
+
+Then('buttons are aligned to {string}', (direction) => {
+  buttons().should('have.class', ALIGN_PREFIX + direction);
```
DlgSHi: maybe we should use more readable concatenation form `${ALIGN_PREFIX}${direction}` -> no + here ;)

PR #2069, opened 2019-05-16
Fe 1414 step sequence

dquilter:
Updates the `StepSequence` and `StepSequenceItem` components to use styled-components, and adds the DLS theme styles.

Screenshot 2019-05-16 at 12 09 28

Screenshot 2019-05-16 at 12 09 40


src/components/step-sequence/step-sequence-item/step-sequence-item.component.js
```diff
@@ -0,0 +1,39 @@
...
+StepSequenceItem.propTypes = {
+  /** Text content for the step item */
+  children: PropTypes.node.isRequired,
+  /** Value to be displayed before text for uncomplete steps */
+  indicator: PropTypes.string.isRequired,
+  /** Aria label */
+  ariaLabel: PropTypes.string,
+  /** Status for the step */
+  status: PropTypes.oneOf(['complete', 'current', 'incomplete']),
+  hiddenCompleteLabel: PropTypes.string,
```
edleeks87: Might be worth adding comments for the last two props? Just nitpicking as everything else looks ok to me

src/components/step-sequence/step-sequence.style.js
```diff
@@ -0,0 +1,29 @@
...
+    flex-direction: column;
+    
+    /* Avoids passing orientation prop down to child items of StepSequence */
+    ${StepSequenceItemVerticalStyle}
+  `};
+  
+  ${({ theme, orientation }) => theme.name === THEMES.classic && css`
+    ${StepSequenceClassicStyle}
+    
+    /* Avoids passing orientation prop down to child items of StepSequence */
```
samhinton88: Does this comment render inside of the css which is actually loaded in the browser? You might want to put it outside of the template string, and maybe be a little more explicit about what you're achieving here.

src/components/step-sequence/step-sequence-item/step-sequence-item.style.js
```diff
@@ -0,0 +1,57 @@
...
+  ${({ theme, status }) => status === 'complete' && css`
+    color: ${theme.colors.base};
+  
+    &::before {
+      background-color: ${theme.colors.base};
+    }
+  `};
+
+  ${({ theme }) => theme.name === 'classic' && css`
+    ${StepSequenceItemClassicStyle}
```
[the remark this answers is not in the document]
dquilter: I was a bit torn on some of these... 

I would tend to agree with you, but where I've done what you've suggested in previous PRs, I've been asked to move `classic` styles to a separate file...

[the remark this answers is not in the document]
grabkowski: there is also `step-sequence-item-vertical-classic.style.js` which has a lot of styles, if the orientation prop could be passed down

src/components/step-sequence/step-sequence.style.js
```diff
@@ -0,0 +1,29 @@
...
+const StepSequenceStyle = styled.ol`
+  display: flex;
+  margin: 0;
+  padding: 18px;
+  font-weight: bold;
+  
+  ${({ orientation }) => orientation === 'vertical' && css`
+    flex-direction: column;
+    
+    /* Avoids passing orientation prop down to child items of StepSequence */
```
grabkowski: why should we avoid passing that prop?

src/components/step-sequence/step-sequence.component.js
```diff
@@ -0,0 +1,25 @@
...
+);
+
+StepSequence.propTypes = {
+  /** Step sequence items to be rendered */
+  children: PropTypes.node,
+  /** The direction that step sequence items should be rendered */
+  orientation: PropTypes.oneOf([
+    'horizontal',
+    'vertical'
+  ])
```
grabkowski: `optionsHelper`

benwilsonsage:
Looking brilliant - just one thing I could see...
- The key line between a previously completed, and the current step, should be 55% of the accent colour (all prior keylines are 100%). This only applies to the keyline immediately before the current step. Opacity: 0.55 seems to do the trick.

src/components/step-sequence/index.js
```diff
@@ -0,0 +1,4 @@
+import StepSequence from './step-sequence.component';
+import StepSequenceItem from './step-sequence-item';
+
+export { StepSequence, StepSequenceItem };
```
dawidzarzycki: I think that we don't need to import and then export components. I think we can just export 
```
export { default as StepSequence } from './step-sequence.component'
export { default as StepSequenceItem } from './step-sequence-item' 
```
What do you think about this idea

PR #2071, opened 2019-05-17
adds ternaries to set start and end index for pagination

edleeks87:
# Description
Fixes issue where page sizes were all the same so records exceeded the total records value

### DIFF
--- a/src/components/table/table.stories.js
+++ b/src/components/table/table.stories.js
@@ -26,12 +26,13 @@ const handleChange = (e, tableOptions) => {
   action('change')(e, tableOptions);
 };
 
-const buildRows = (pageSizeFromKnobs) => {
-  const pageSize = pageSizeFromKnobs;
+const buildRows = (pageSize, totalRecords) => {
   const currentPage = store.get('currentPage');
+  const candidateIndex = pageSize * currentPage;
 
-  const endIndex = pageSize * currentPage;
-  const startIndex = endIndex - pageSize;
+  const endIndex = (candidateIndex <= totalRecords) ? candidateIndex : totalRecords;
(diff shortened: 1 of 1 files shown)


src/components/table/table.stories.js
```diff
@@ -26,12 +26,12 @@ const handleChange = (e, tableOptions) => {
...
 };
 
-const buildRows = (pageSizeFromKnobs) => {
+const buildRows = (pageSizeFromKnobs, totalRecords) => {
   const pageSize = pageSizeFromKnobs;
   const currentPage = store.get('currentPage');
 
-  const endIndex = pageSize * currentPage;
-  const startIndex = endIndex - pageSize;
+  const endIndex = pageSize * currentPage <= totalRecords ? pageSize * currentPage : totalRecords;
```
dquilter: Could create a variable for `pageSize * currentPage`

src/components/table/table.stories.js
```diff
@@ -26,12 +26,12 @@ const handleChange = (e, tableOptions) => {
...
+  const startIndex = totalRecords === endIndex ? totalRecords - (endIndex % pageSize) : endIndex - pageSize;
```
andrewjtait: Maybe this is a bit clearer?
```
let endIndex = pageSize * currentPage;
if (endIndex > totalRecords) endIndex = totalRecords;

const currentPageSize = (endIndex === totalRecords) ? (endIndex % pageSize) : pageSize;
const startIndex = endIndex - currentPageSize
```

src/components/table/table.stories.js
edleeks87: Yeah I've gone and added it

src/components/table/table.stories.js
edleeks87: Gone for a halfway between your way and mine

PR #2079, opened 2019-05-20
FE-1505 tables with inputs

edleeks87:
# Description
Adds the styling for inputs in a `Table` to match the DLS, the extra functionality such as hover states and key navigation will be part of a different ticket/ PR


src/components/table/table-row/classic-row.style.js
```diff
@@ -0,0 +1,49 @@
...
+    &:hover {
+      ${StyledTableCell} {
+        background-color: #E7F1FC;
+      }
+    }
+  `;
+}
+
+function applyClassicSelectedStyling(theme) {
+  const { colors } = theme;
```
dawidzarzycki: I'm not sure we need destructuring colors if we use  it only once

src/components/table/table-modern.style.js
```diff
@@ -45,27 +47,49 @@ function applyModernTableStyling({
...
+    }
+
+    && ${StyledInputPresentation} {
+      height: ${tableSizes[size].inputHeight};
+      min-height: ${tableSizes[size].inputHeight};
+    }
+
+    && ${StyledTableCell} div {
+      height: ${tableSizes[size].inputHeight};
+      min-height: ${tableSizes[size].inputHeight};
```
grabkowski: it could be simplified to:
```
const { font, inputHeight, padding } = tableSizes[size];
```
I would also consider renaming `font: fontSize` and `padding: paddingSize` during destructuration

src/components/table/table-row/modern-row.style.js
```diff
@@ -0,0 +1,36 @@
...
+    border-bottom-color: ${table.selected};
+  `;
+}
+
+function applyModernHighlightStyling({ table }) {
+  return `
+    background-color: ${table.selected};
+    border-bottom-color: ${table.selected};
+  `;
+}
```
grabkowski: this method does the same thing as the one above

src/components/table/table.stories.js
```diff
@@ -158,7 +223,73 @@ storiesOf('Table', module)
...
+      const highlightable = boolean('highlightable', false);
+      const shrink = boolean('shrink', false);
+      const caption = text('Caption', 'Country and Country Codes');
+      const totalRecords = text('totalRecords', '50');
+      const paginate = boolean('paginate', false);
+      const showPageSizeSelection = paginate && boolean('showPageSizeSelection', false);
+      const theme = select('theme', OptionsHelper.tableThemes, Table.defaultProps.theme);
+      const size = select('size', OptionsHelper.tableSizes, Table.defaultProps.size);
+      const isZebra = boolean('zebra striping', false);
+      const inputType = select(
```
grabkowski: common knobs could be extracted to a separate object and then reused with the rest operator

src/components/table/table-row/modern-row.style.js
edleeks87: Yeah I hadn't merged in the updates in the base, the duplication is removed now

src/components/table/table-modern.style.js
edleeks87: Yeah good call, I've made the changes

src/components/table/table-row/classic-row.style.js
edleeks87: Yeah,  good call

src/components/table/table-cell/table-cell.component.js
```diff
@@ -19,14 +20,41 @@ import OptionsHelper from '../../../utils/helpers/options-helper';
  * set styling options for the cell used for action such as delete.
  */
 class TableCell extends React.Component {
+  /**
+   * Returns if child element has type property.
+   */
+  childHasType(child) {
```
stevebutler2210: This name suggests to me a boolean return value, but actually returns the type if one exists.

Wonder if it might read more clearly if it was called `childType`, and returned `(child && child.type) || undefined`?

stevebutler2210: In fact, looking at where it's used, it might be better to return the `child.type.name` when it exists, as we don't seem to be using it for anything else?

src/components/table/table-cell/table-cell.component.js
```diff
@@ -19,14 +20,41 @@ import OptionsHelper from '../../../utils/helpers/options-helper';
...
+  }
+
+  /**
+   * Returns if child element is Textarea.
+   */
+  childrenHasTextArea() {
+    const { children } = this.props;
+
+    if (!Array.isArray(children)) {
+      return this.childHasType(children) && children.type.name === TextArea.name;
```
stevebutler2210: With the suggestion above this would be simplified to:
`return this.childType(children) === TextArea.name`

src/components/table/table-cell/table-cell.component.js
```diff
@@ -19,14 +20,41 @@ import OptionsHelper from '../../../utils/helpers/options-helper';
...
+
   /**
    * Returns props to be used on the TD element.
    */
   get tableCellProps() {
     const { ...props } = validProps(this);
-
     delete props.children;
     props.align = this.props.align;
+    props.size = this.props.size;
```
stevebutler2210: Aren't align and size already set? If not, is it because they need to be marked as safe when calling validProps? I.e. `validProps(this, ['align', 'size'])`.

src/components/table/table-cell/table-cell.component.js
```diff
@@ -19,14 +20,41 @@ import OptionsHelper from '../../../utils/helpers/options-helper';
...
+    return result;
+  }
+
   /**
    * Returns props to be used on the TD element.
    */
   get tableCellProps() {
     const { ...props } = validProps(this);
-
     delete props.children;
```
stevebutler2210: Could this be achieved with destructuring on the line above?
`const { children, ...props } = validProps(this)`

src/components/table/table-cell/table-cell.style.js
```diff
@@ -25,7 +31,9 @@ const StyledTableCell = styled.td`
       }
       
       ${action && applyActionStyling(colors)}
-  `;
+
+      ${!isClassic(theme) && size ? applyModernPresentationStyling(size, isTextArea) : ''}
```
stevebutler2210: Are we ok doing stuff like this? I'm totally fine with it, but I've had PRs rejected previously for including logic which refers to whether it's classic theme or not, outside of the actual classic style file - just checking so I know I can do this where it makes sense in future!

src/components/table/table-cell/table-cell.component.js
edleeks87: Yeah good point I'll make the change

src/components/table/table-cell/table-cell.style.js
edleeks87: There is only the single style file for TableCell so I think it's ok here, but I guess in scenarios where there are separate files it should be avoided. The style doesn't change when it's classic or modern as any overrides happen higher up based on the props received at that level.

src/components/table/table-cell/table-cell.component.js
edleeks87: Yeah good point, somehow forgot that `validProps` had been updated to allow that

sianford:
Tested Table > With inputs in Safari, Chrome, Firefox, IE11 and Edge browsers. Defects logged and have been resolved. Testing passed.

PR #2081, opened 2019-05-22
Cypress/fe 1608

marcinstachowicz:
# BDD Cypress

## Maintenance
- button
- button toggle

### DIFF
--- a/cypress/integration/button.feature
+++ b/cypress/integration/button.feature
@@ -2,34 +2,34 @@ Feature: Button component
   I want to change Button label, theme, size and subtext properties
 
   Background: Open Button component page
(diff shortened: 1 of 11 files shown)


cypress/integration/button.feature
```diff
@@ -65,24 +56,41 @@ Feature: Button component
     Then Button is enabled
 
   @positive
-  Scenario Outline: Set Button size to small, medium and large
+  Scenario Outline: Set Button size to <size>
     When I select size to "<size>"
-    Then Button size property on preview is "<size>"
+    Then Button height is "<height>"
```
DlgSHi: the height parameter is only changing ?

PR #2082, opened 2019-05-22
Button Toggle Group Validation

dquilter:
Adds the validation HOC to the `ButtonToggleGroup` component.

Setups up example in Storybook under `Validations`

Screenshot 2019-05-28 at 14 07 39


src/components/validations/validations.stories.js
```diff
@@ -179,4 +182,54 @@ storiesOf('Validations', module)
...
+            <ButtonToggle
+              name='button-toggle-group'
+              id='baz'
+              value='baz'
+              onChange={ (ev) => {
+                buttonToggleGroupStore.set({ value: ev.target.value });
+              } }
+            >
+              Baz
+            </ButtonToggle>
```
grabkowski: I think that these buttons could be rendered in a loop

src/components/button-toggle-group/button-toggle-group.spec.js
```diff
@@ -56,5 +56,9 @@ describe('ButtonToggleGroup', () => {
         width: '48%'
       }, wrapper.find(ButtonToggleGroupStyle));
     });
+    it('renders ButtonToggle correctly with validation errors', () => {
+      const wrapper = render({ theme: smallTheme, errorMessage: 'error' }, TestRenderer.create).toJSON();
+      expect(wrapper).toMatchSnapshot();
+    });
```
grabkowski: shouldn't this test have it's own `describe` method?

src/components/button-toggle-group/button-toggle-group.spec.js
```diff
@@ -56,5 +56,9 @@ describe('ButtonToggleGroup', () => {
         width: '48%'
       }, wrapper.find(ButtonToggleGroupStyle));
     });
+    it('renders ButtonToggle correctly with validation errors', () => {
+      const wrapper = render({ theme: smallTheme, errorMessage: 'error' }, TestRenderer.create).toJSON();
+      expect(wrapper).toMatchSnapshot();
```
grabkowski: I think that could be tested just by looking for a component generated by the validation instead of creating a snapshot of the whole component.

dquilter: Initially I had this written as an `assertStyleMatch` which always failed, I think because the test couldn't a nested styled component. More detail here:

https://github.com/styled-components/styled-components/issues/1985

I tried a number of suggested fixes, none of which worked. Then I used a snapshot instead, and the test passed (you can see the error styling being added).

I'm at a bit of an impasse here, so if you have any suggestions, I'm happy to hear them.

src/components/button-toggle-group/button-toggle-group.style.js
```diff
@@ -1,9 +1,23 @@
...
+      border-color: ${theme.colors.error};
+    }
+  `};
+
+  ${ValidationIconStyle} {
+    display: inline-block;
+    margin-left: 10px;
+    vertical-align: top;
+  }
 `;
```
dawidzarzycki: I think that we need to have defaultProps here when we use `theme` props

src/components/validations/validations.stories.js
```diff
@@ -179,4 +182,39 @@ storiesOf('Validations', module)
...
+  })
+  .add('ButtonToggleGroup', () => {
+    const test = value => new Promise((resolve, reject) => {
+      if (value === 'baz') return resolve(true);
+      return reject(Error('Baz is required!'));
+    });
+
+    return (
+      <Form
+        onSubmit={ (ev) => { ev.preventDefault(); } }
```
dawidzarzycki: I think would be good to have an action logs here

src/components/validations/validations.stories.js
```diff
@@ -179,4 +182,39 @@ storiesOf('Validations', module)
...
+            fieldHelp='Click save to run validation'
+            validations={ test }
+          >
+            {['foo', 'bar', 'baz'].map(value => (
+              <ButtonToggle
+                name='button-toggle-group'
+                id={ value }
+                value={ value }
+                onChange={ (ev) => {
+                  buttonToggleGroupStore.set({ value: ev.target.value });
```
dawidzarzycki: Is it possible to have actions logs here?

grabkowski: added

src/components/validations/validations.stories.js
grabkowski: logs added

harpalsingh:
Hey, after discussing around the original design to keep the error along the top, we would like to move the error icon to be center aligned based on the height if the button toggle please.

Currently it does:

Screenshot 2019-06-11 at 14 30 41

We would like that updated to:

Screenshot 2019-06-11 at 14 30 58

src/components/button-toggle-group/button-toggle-group.component.js
```diff
@@ -4,13 +4,23 @@ import FormField from '../../__experimental__/components/form-field';
...
-    </ButtonToggleGroupStyle>
-  </FormField>
-));
+const ButtonToggleGroup = withValidations((props) => {
+  const { inputWidth, errorMessage, onChange } = props;
+
+  return (
+    <FormField { ...props }>
+      <ButtonToggleGroupStyle
+        aria-label={ props.label }
```
dawidzarzycki: I see above that there is an object destructuring . Why wouldn't we add there label and children also?

src/components/button-toggle/button-toggle.component.js
```diff
@@ -10,19 +10,36 @@ import ButtonToggleIcon from './button-toggle-icon.component';
...
+    buttonIcon,
+    buttonIconSize,
+    onChange,
+    size
+  } = props;
   const inputGuid = guid();
-  const icon = props.buttonIcon ? <ButtonToggleIcon { ...props } /> : null;
+  let icon;
+
+  if (props.buttonIcon) {
```
dawidzarzycki: destructuring all props please

src/components/button-toggle/button-toggle.spec.js
```diff
@@ -1,34 +1,27 @@
...
   describe('Classic theme', () => {
     it('renders correctly with default settings', () => {
-      const wrapper = testRender({ theme: classicTheme });
+      const wrapper = renderWithTheme({ theme: ClassicTheme }, TestRenderer.create);
       expect(wrapper).toMatchSnapshot();
     });
     it('renders correctly with large buttonIcon and large size', () => {
-      const wrapper = render({
-        theme: classicTheme,
+      const wrapper = renderWithTheme({
```
dawidzarzycki: What do you think to create beforeEach in the case when we have same props provided?

grabkowski: that's a good idea, but we don't have the same prop sets here

PR #2085, opened 2019-05-23
FE-1518 - Preview Component

dawidzarzycki:
# Description
- Update `Preview` component to use `styled-components`

# Screenshots / Gifs
preview

# Testing Instructions
- There are no differences according to theme (classic, small, medium, large are the same)


src/components/preview/preview.spec.js
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
-import { shallow } from 'enzyme';
-import Preview from './preview';
+import TestRenderer from 'react-test-renderer';
+import 'jest-styled-components';
+import Preview from './preview.component';
 
 const renderShallow = (children, props) => {
```
grabkowski: it's not shallow now :)

src/components/preview/__snapshots__/preview.spec.js.snap
```diff
@@ -0,0 +1,163 @@
...
+      "height": undefined,
+      "width": undefined,
+    }
+  }
+/>
+`;
+
+exports[`Preview when given children will render the children 1`] = `"This is some text"`;
+
+exports[`Preview when given no children and a falsy loading prop will render nothing 1`] = `null`;
```
grabkowski: I think we shouldn't have snapshots like these

src/components/preview/__snapshots__/preview.spec.js.snap
```diff
@@ -0,0 +1,163 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Preview when given a width and height renders with the given width and height 1`] = `
+.c1 {
+  -webkit-animation: blNbyL 2s ease infinite;
+  animation: blNbyL 2s ease infinite;
```
grabkowski: wouldn't that be different on each snapshot generation?

dawidzarzycki: It should works good

src/components/preview/__snapshots__/preview.spec.js.snap
dawidzarzycki: Just copied, refactor isn't in the scope

PR #2087, opened 2019-05-24
Adds props to wrapper to calc min height

edleeks87:
# Description
Adds rowTotal and rowHeight props to table wrapper to calculate its minimum height

### DIFF
--- a/src/components/table/table-sizes.style.js
+++ b/src/components/table/table-sizes.style.js
@@ -22,5 +22,11 @@ export default {
     font: '16px',
     padding: '13px',
     inputHeight: '32px'
+  },
+  wrapper: {
+    compact: '24',
+    small: '32',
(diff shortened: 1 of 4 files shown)


src/components/table/table-sizes.style.js
```diff
@@ -22,5 +22,11 @@ export default {
     font: '16px',
     padding: '13px',
     inputHeight: '32px'
+  },
+  wrapper: {
+    compact: '24',
```
samhinton88: We probably want to keep consistency in the way we declare units; here we have a mixture. Probably better to specify the unit here, rather than in the style string.

edleeks87: I've changed it to return a function in the config now as the min-height needs to be calculated dynamically

PR #2089, opened 2019-05-24
FE-1256 update split button for DLS

grabkowski:
Updates component to be DLS compliant and adds new functionality such as key navigation of the listed additional buttons and the list opening on focus


src/components/split-button/split-button-toggle.style.js
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+      &:active {
+        background-color: ${theme.colors.tertiary};
+        border-color: ${theme.colors.tertiary};
+      }
+    ` : ''}
+
+    ${buttonType === 'primary' ? `border-left-color: ${theme.colors.secondary};` : 'border-left: none;'}
+    padding 0 ${horizontalPaddingSizes[size]};
```
izakrzewska: missing :

src/components/split-button/split-button-classic.style.js
```diff
@@ -0,0 +1,75 @@
...
+import StyledButton from '../button/button.style';
+
+const classicColors = {
+  active: '#1963f6',
+  primary: '#255bc7',
+  secondary: '#1e499f',
+  tertiary: '#163777'
+};
+
+const getSplitButtonChildrenClassicStyles = ({ theme }) => (theme.name === THEMES.classic ? css`
```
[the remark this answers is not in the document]
grabkowski: if I'll use `&&` operator, when the conditional results in false then a `"false"` string will be returned, which is not a valid CSS. It will be ignored by the browser and displayed correctly, but it could lead to an untraceable error during snapshot generation.

src/components/split-button/split-button-classic.style.js
```diff
@@ -0,0 +1,75 @@
...
+
+const getSplitButtonToggleClassicStyles = ({
+  theme,
+  disabled,
+  displayed,
+  buttonType
+}) => (theme.name === THEMES.classic ? css`
+  z-index: 20;
+
+  ${!disabled && displayed ? css`
```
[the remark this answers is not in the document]
grabkowski: no, we need to have it both enabled and displayed

src/components/split-button/split-button-toggle.style.js
grabkowski: good catch!

src/components/split-button/split-button-children.style.js
```diff
@@ -0,0 +1,52 @@
...
+
+  ${getSplitButtonChildrenClassicStyles}
+`;
+
+StyledSplitButtonChildrenContainer.propTypes = {
+  displayButtons: PropTypes.bool
+};
+
+StyledSplitButtonChildrenContainer.defaultProps = {
+  displayButtons: false,
```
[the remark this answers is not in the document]
grabkowski: removed

src/components/multi-action-button/multi-action-button.js
```diff
@@ -95,6 +74,46 @@ class MultiActionButton extends SplitButton {
...
+      <StyledMultiActionButton buttonType={ this.props.buttonType || this.props.as }>
+        { super.render() }
+      </StyledMultiActionButton>
+    );
+  }
 }
 
+MultiActionButton.propTypes = {
+  /**
+   * Customizes the appearance, can be set to 'primary', 'secondary' or 'transparent'.
```
dawidzarzycki: I think we can remove not necessary spaces and keep comments in one line
`/* Customizes the appearance, can be set to 'primary', 'secondary' or 'transparent'. */`

grabkowski: multi action button will be rewritten in a separate task, so I think we could leave it as it is now

sianford:
Tested Split Button in Safari, Chrome, Firefox, IE11 and Edge browsers. All defects logged, fixed and retested. All testing has passed.

harpalsingh:
There are a few issues that need resolving with the component.

**none** is using color that is not accessible.

Screenshot 2019-06-06 at 14 32 25

The hover color should be the same on secondary and primary on the chevron and matches the background of the split menu and therefore should not have a dark hover as it has no action on it.

color

harpalsingh:
Passed review, though we still need to review the "none" color at some point, but we can come back to that.

PR #2090, opened 2019-05-28
Fe 1510 styled components for modal

grabkowski:
Moves all modal related styles from SCSS to styled-component.

This component is a base for Dialog, DialogFullScreen and Sidebar components.


src/components/modal/modal.component.js
```diff
@@ -181,63 +92,35 @@ class Modal extends React.Component {
...
-          className='carbon-modal__background'
-        />
+        <StyledModalBackground data-element='modal-background' />
       );
     }
     return null;
   }
 
-  // Called after the modal opens
   get onOpening() { return null; }
```
dawidzarzycki: Can we remove brackets? 
`get get onOpening() return null `

src/components/modal/modal.component.js
```diff
@@ -5,65 +5,25 @@ import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
...
 import Portal from '../portal';
-import './modal.scss';
+import { StyledModal, StyledModalBackground } from './modal.style';
 
 const TIMEOUT = 500;
 /**
  * A Modal Component
  *
  * Abstract base class for all modals
- *
```
edleeks87: I think we want to remove the `@` doc stuff but want to leave the top level class comment so that when we finish the adding the Typescript definitions the developers get some helpful information about how to use the component from intellisense etc

PR #2091, opened 2019-05-28
Ensure all table data is sorted not just active page

edleeks87:
# Description
Fixes defect that only sorted data per page and not the whole data set

### DIFF
--- a/src/components/table/table.stories.js
+++ b/src/components/table/table.stories.js
@@ -35,6 +35,13 @@ const handleChange = (e, tableOptions) => {
   action('change')(e, tableOptions);
 };
 
+const recordsForActivePage = (start, end) => {
+  if (store.get('sortOrder') === 'desc' && store.get('sortedColumn').length) {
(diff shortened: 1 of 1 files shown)


src/components/table/table.stories.js
```diff
@@ -35,24 +35,28 @@ const handleChange = (e, tableOptions) => {
   action('change')(e, tableOptions);
 };
 
+const recordsForActivePage = (start, end) => {
+  if (store.get('sortOrder') === 'desc' && store.get('sortedColumn').length) {
+    return countriesList.reverse().slice(start, end).toJS();
+  }
+  return countriesList.slice(start, end).toJS();
```
grabkowski: I think it could be simplified:
```
  const records = countriesList;

  if (store.get('sortOrder') === 'desc' && store.get('sortedColumn').length) {
    records = records.reverse();
  }

  return records.slice(start, end).toJS();
```

edleeks87: Yeah I was leaning towards reducing the complexity

PR #2092, opened 2019-05-28
fix wrapper sizing when shrink prop set

edleeks87:
# Description
Table wrapper sizing was preventing shrink prop styling on table. Now passing the shrink prop to wrapper as well to allow it to toggle whether it calls the sizing function or not


src/components/table/table.spec.js
```diff
@@ -1026,6 +1028,41 @@ describe('Table', () => {
     });
   });
 
+  describe.each(OptionsHelper.tableSizes)(
+    'the table wrapper',
+    (size) => {
+      it('does not set the min-height of the wrapper when the shrin props is true', () => {
```
samhinton88: Should be 'shrink' not 'shrin'

PR #2093, opened 2019-05-28
Add width prop to header component

edleeks87:
# Description
Adds an optional width prop to the Header component to allow a developer to statically size the column

### DIFF
--- a/src/components/table/table-header/table-header.component.js
+++ b/src/components/table/table-header/table-header.component.js
@@ -156,12 +156,12 @@ class TableHeader extends React.Component {
       contents = this.props.children;
     }
 
-    const headerProps = {
+    const additionalProps = {
       align: this.props.align,
       sortable: this.props.sortable
     };
 
-    if (!Number.isNaN(Number(this.props.width))) headerProps.width = this.props.width;
+    if (/^\d+$/.test(this.props.width)) additionalProps.width = this.props.width;
 
     return (
       React.createElement(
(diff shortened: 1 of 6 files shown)


src/components/table/table-header/table-header.component.js
```diff
@@ -155,15 +155,22 @@ class TableHeader extends React.Component {
     } else {
       contents = this.props.children;
     }
+
+    const headerProps = {
+      align: this.props.align,
+      sortable: this.props.sortable
+    };
+
+    if (/^\d+$/.test(this.props.width)) headerProps.width = this.props.width;
```
stevebutler2210: Would `!isNaN(this.props.width)` work? If so that might read better

edleeks87: Yeah good point, the linter was complaining when I tried using `isNaN`, but just realised you can use `Number.isNaN`

stevebutler2210: 👌

src/components/table/table-header/table-header.component.js
```diff
@@ -155,15 +155,22 @@ class TableHeader extends React.Component {
     } else {
       contents = this.props.children;
     }
+
+    const headerProps = {
+      align: this.props.align,
+      sortable: this.props.sortable
+    };
+
+    if (!Number.isNaN(Number(this.props.width))) headerProps.width = this.props.width;
```
stevebutler2210: This should take the string just fine, it'll convert it

edleeks87: I've had to revert back to the regex implementation as I need it to prevent any hex coercions as well

PR #2095, opened 2019-05-29
S1T-1897 - Add withCredentials support to TableAjax

whomwah:
### Description

Adding `withCredentials` support to the TableAjax component.

> *withCredentials()* makes your browser include cookies and authentication headers in your XHR request. If your service depends on any cookie (including session cookies), it will only work with this option set.

### TODO

- [x] Release notes


whomwah:
@philearley I've not used the `definition.js` but it makes sense. I'll add it.

PR #2102, opened 2019-06-03
Fix table wrapper sizing

edleeks87:
# Description
Prevents the a story rendering a page that doesn't exist by resetting it to the CurrentPage to the first


src/components/table/table.stories.js
```diff
@@ -88,22 +112,20 @@ const buildRows = (pageSize, totalRecords) => {
   );
 };
 
+const setCurrentPage = ({ totalRecords, pageSize }) => {
+  if (totalRecords % pageSize !== 0) return store.get('currentPage');
+  return (store.get('currentPage') > (totalRecords / pageSize)) ? '1' : store.get('currentPage');
```
grabkowski: could you extract the conditional to a separate `const`, to increase readability?

edleeks87: Yeah good point, I had to refactor it slightly as I had missed an edge case

PR #2106, opened 2019-06-04
Adjust current page between table stories

edleeks87:
# Description
Fixes defect where pageSize caused incorrect currentPage value

### DIFF
--- a/src/components/table/table.stories.js
+++ b/src/components/table/table.stories.js
@@ -28,16 +28,7 @@ const getCommonKnobs = () => {
     caption: text('caption', 'Country and Country Codes'),
     totalRecords: text('totalRecords', '50'),
     paginate,
-    showPageSizeSelection: paginate && boolean('showPageSizeSelection', false),
-    theme: select(
-      'theme',
-      [
-        OptionsHelper.tableThemes[0],
-        OptionsHelper.tableThemes[1],
-        OptionsHelper.tableThemes[2]
-      ],
-      Table.defaultProps.theme
-    )
+    showPageSizeSelection: paginate && boolean('showPageSizeSelection', false)
   };
 };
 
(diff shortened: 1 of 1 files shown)


src/components/table/table.stories.js
```diff
@@ -112,8 +103,15 @@ const buildRows = ({ pageSize, totalRecords }) => {
...
+const roundPageValue = (value) => {
+  const roundedValue = Math.round(value);
+  if (!value || roundedValue < 1) return '1';
+  return roundedValue;
+};
+
 const setCurrentPage = ({ totalRecords, pageSize, paginate }) => {
-  const maxValidPage = Math.round(totalRecords / pageSize) + (totalRecords % pageSize);
+  const roundedValue = roundPageValue(totalRecords / pageSize);
+  const maxValidPage = roundedValue < totalRecords / pageSize ? roundedValue + 1 : roundedValue;
```
chrisbarber86: Might not an idea to put some brackets into this line to make it a bit clearer what is happening.

chrisbarber86: Total pages could be expressed as follows:

```
Math.max(Math.ceil(totalRecords / pageSize), 1)
```
you would want to check that `pageSize` is set and if it's not then return `1`.

grabkowski: or create a const containing `totalRecords / pageSize`

edleeks87: I went with a mix of both!

src/components/table/table.stories.js
```diff
@@ -185,7 +176,15 @@ storiesOf('Table', module)
       store.set({ currentPage: setCurrentPage(props) });
       props.size = select('size', OptionsHelper.tableSizes, Table.defaultProps.size);
       props.isZebra = boolean('zebra striping', false);
-
+      props.theme = select(
+        'theme',
+        [
```
stevebutler2210: Is there a reason you can't just pass in the whole `OptionsHelper.tableThemes` object?

edleeks87: Yeah haha no idea why I've done that here, the classic story only has two variants but the DLS has them all 🤦‍♂

PR #2108, opened 2019-06-04
FE-1418 Pill DLS

ianforster-sage:
# Description
Update Pill component to use style-components and new styling.

# TODO
- [x] Release notes


src/components/pill/pill-classic.style.js
```diff
@@ -0,0 +1,151 @@
...
+  },
+  success: {
+    color: '#50B848',
+    bgColor: '#50B848',
+    hoverBgColor: '#4782F7',
+    hoverColor: '#FFFFFF'
+  }
+};
+
+export default (props) => {
```
edleeks87: You could destructure the param here as you only need the `styledAs` prop:
```
export default ({ styledAs }) => {
  const colourSet = classicStyleConfig[styledAs];
...
```

src/components/pill/pill-modern.style.js
```diff
@@ -0,0 +1,84 @@
+import { css } from 'styled-components';
+import modernConfig from './pill-modern.config';
+import { THEMES } from '../../style/themes';
+
+export default (props) => {
```
edleeks87: Similar to above
```
export default ({ styledAs, theme,  inFill, isDeletable }) => {
...
```

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Callback function to delete the component, when the added Icon is clicked.
+   */
+  onDelete: PropTypes.func
+};
+
+Pill.defaultProps = {
+  as: 'default',
+  fill: false,
+  className: '',
```
edleeks87: Not sure `className` is needed as a prop anymore

src/components/pill/pill.spec.js
```diff
@@ -0,0 +1,348 @@
...
+  'neutral',
+  'negative',
+  'warning',
+  'positive'
+];
+
+const modernThemes = [
+  ['small', smallTheme]
+];
+describe('Pill', () => {
```
edleeks87: I'd probably add a blank line in between each new `describe` or `it` block so it's more readable

src/components/pill/pill.style.js
```diff
@@ -0,0 +1,35 @@
...
+import modernThemeForPill from './pill-modern.style';
+import { THEMES } from '../../style/themes';
+import OptionsHelper from '../../utils/helpers/options-helper';
+
+const PillStyle = styled.span`
+  ${props => (isClassic(props) ? classicThemeForPill : modernThemeForPill)}
+`;
+
+function isClassic({ theme, styledAs }) {
+  // handles incorrect default activeTheme in StoryBook AppWrapper
```
chrisbarber86: This shouldn't really be in the component itself, it should be part of the story book usage since this is a specific issue with the StoryBook AppWrapper, rather than the Pill component itself.

src/components/pill/pill-classic.style.js
```diff
@@ -0,0 +1,151 @@
+import { css } from 'styled-components';
+
+export const classicStyleConfig = {
+  disabled: {
+    color: '#CCD6DA',
```
edleeks87: Where the colour is duplicated iyou could probably get away with just defining it the once rather than having `color` `bgColor` for example

src/components/pill/pill.style.js
ianforster-sage: Thanks Chris, I agree but because the App Layout determines the theme and wraps the Pill with a Theme Provider before it renders the component it's already set before we try to style it. We know that there's a requirement for some middleware to better handle this, but for now this code is an unobtrusive workaround. When the middleware is available we'll revisit this and correct it.

src/components/pill/pill-modern.config.js
```diff
@@ -0,0 +1,15 @@
...
+  small: {
+    colors: {
+      neutral: pill.neutral,
+      negative: colors.error,
+      warning: colors.warning,
+      positive: colors.secondary
+    },
+    hoverColor: colors.white
+  }
+};
```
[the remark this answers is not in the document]
ianforster-sage: At the moment, yes. After discussion with UX we are implementing the small theme for all variations for now.

src/components/pill/pill-modern.style.js
```diff
@@ -0,0 +1,84 @@
...
+            `}
+  
+            &:hover,
+            &:focus {
+              color: ${styleSet.colors[styledAs]};
+            }
+          }
+      `}
+  `;
+};
```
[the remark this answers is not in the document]
ianforster-sage: What would you rename the existing `pill.style.js` as?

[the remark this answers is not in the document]
grabkowski: I would delete existing `pill.style.js`, rename this file and override styles with `pill.classic.style.js` for the `classic` theme

src/components/message/message.spec.js
```diff
@@ -77,7 +77,18 @@ describe('Message', () => {
...
+        const colors = [
+          'default',
+          'error',
+          'help',
+          'info',
+          'maintenance',
+          'new',
+          'success',
+          'warning'
+        ];
```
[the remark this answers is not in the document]
grabkowski: I think that `disabled` shouldn't be added to the colors helper, as it's used by multiple components through the Application. Adding it would allow to choose that color in stories in which we are using this helper.

src/components/pill/pill.style.js
```diff
@@ -0,0 +1,35 @@
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import baseTheme from '../../style/themes/base';
+import classicThemeForPill from './pill-classic.style';
+import modernThemeForPill from './pill-modern.style';
+import { THEMES } from '../../style/themes';
+import OptionsHelper from '../../utils/helpers/options-helper';
+
+const PillStyle = styled.span`
+  ${props => (isClassic(props) ? classicThemeForPill : modernThemeForPill)}
```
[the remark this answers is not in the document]
ianforster-sage: Was this agreed before I joined the team? I've just followed an existing pattern here.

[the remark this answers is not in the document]
grabkowski: Yes, it was agreed a few months ago. One on the examples:
https://github.com/Sage/carbon/blob/master/src/__experimental__/components/input/input-presentation-classic.style.js

src/utils/helpers/options-helper/options-helper.js
```diff
@@ -39,6 +39,7 @@ const OptionsHelper = {
 
   colors: [
     'default',
+    'disabled',
```
chrisbarber86: `disabled` is a colour?

src/components/pill/pill.style.js
ianforster-sage: @chrisbarber86 are you satisfied with this?

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,74 @@
...
+    );
+  }
+}
+
+Pill.propTypes = {
+
+  /**
+   * Sets the theme of the notification.
+   */
+  as: PropTypes.string,
```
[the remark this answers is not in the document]
grabkowski: what do you think about adding another prop that could have more meaningful name, e.g.: `colorVariant`?

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,75 @@
...
+        { onDelete && this.renderCloseIcon() }
+      </StyledPill>
+    );
+  }
+}
+
+Pill.propTypes = {
+
+  // Sets the theme of the notification [legacy].
+  as: PropTypes.string,
```
grabkowski: I think that should be `PropTypes.oneOf(...OptionsHelper.colors, 'disabled')`

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,75 @@
...
+
+  // Sets the theme of the notification.
+  colourVariant: PropTypes.string,
```
grabkowski: `PropTypes.oneOf(...OptionsHelper.pillColours)`

src/utils/helpers/options-helper/options-helper.js
```diff
@@ -48,6 +48,16 @@ const OptionsHelper = {
...
 
+  pillColours: [
+    'neutral',
+    'negative',
+    'positive'
+  ],
+
+  pillStates: [
+    'disabled'
+  ],
```
grabkowski: I don't think that we need that, as it's only one state

src/components/pill/pill.stories.js
```diff
@@ -1,56 +1,70 @@
...
-import { StoryHeader, StoryCode } from '../../../.storybook/style/storybook-info.styles';
+import classic from '../../style/themes/classic';
+import { THEMES } from '../../style/themes';
+import { notes, Info } from './documentation';
+
+const getKnobs = (theme) => {
+  // eslint-disable multiline-ternary
+  const styledAs = (theme === THEMES.classic)
+    ? select('as', [...OptionsHelper.colors, OptionsHelper.pillStates[0]], Pill.defaultProps.as)
+    : select('as', [...OptionsHelper.pillColours, OptionsHelper.colors[7]], OptionsHelper.pillColours[0]);
```
grabkowski: shouldn't you use `colourVariant` prop in modern themes?

src/__experimental__/components/select/select.component.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import SelectList from './select-list.component';
 import InputDecoratorBridge from '../input-decorator-bridge';
-import Pill from '../../../components/pill';
+import Pill from '../../../components/pill/pill.component';
```
edleeks87: This can just be `../../../components/pill` as you have and index file

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,75 @@
...
+
+  // Fills the pill background with colour. When fill is false only the border is coloured.
+  fill: PropTypes.bool,
+
+  // Callback function for when the pill is clicked.
+  onClick: PropTypes.func,
+
+  // Callback function to delete the component, when the added Icon is clicked.
+  onDelete: PropTypes.func
+};
```
grabkowski: I think that there should be a `disabled` prop and that we should override other colours in `pill.style.js` instead of passing a `disabled` colour as a `colourVariant`.

ianforster-sage: `disabled` is an attribute of some form controls like input, I think we'd be better to change the name and continue to pass it as a `colourVariant` as the behaviour we're looking for is the same as for other colours.

ianforster-sage: I checked with UX, their opinion is:

> Pills are just visual indicators with no interactivity, but sometimes they're used as links or tags added inside an input... I guess a disabled link is unusual, and tags can be deleted but not disabled. So I guess overall there may not be a use for a disabled pill, I think it's OK not to have a disabled state.

Therefore, as this is just a legacy feature that is deprecated I'd like to leave it as-is.

src/components/pill/pill.component.js
```diff
@@ -0,0 +1,76 @@
...
+      as,
+      onDelete,
+      colourVariant,
+      children
+    } = this.props;
+    return (
+      <StyledPill
+        { ...validProps(this) }
+        inFill={ fill }
+        styledAs={ colourVariant || as }
```
grabkowski: why not just `colorVariant` instead of `styledAs`?

benwilsonsage:
Looking good - just a few minor comments...

Expected: Classic theme > 'onDelete' hover state applies dark blue background behind 'X' icon regardless of 'as' setting.
Actual: Different coloured backgrounds behind 'X' icon for each different 'as' setting.

Expected: S/M/L theme > 'Warning' pill with 'filled' checked has text color black. (This is a unique exception compared to other colours due to accessibility contrast ratios).
Actual: Has text color white.

Expected: S/M/L theme > 'Warning' pill, with 'filled' checked has background color and border color 'Carrot Tint 20'. (This is a unique exception compared to other colours due to accessibility contrast ratios. If 'Filled is not checked, then border color is Carrot base.)
Actual: Background and border color are carrot base.

Expected: S/M/L theme > If keyboard focus is applied to a pill with 'onDelete', it has the same effect as hovering.
Actual: Keyboard focus applies browser default visuals, but doesn't trigger gold outline etc.

Expected: S/M/L theme > If I hover on the main body of a pill with 'onDelete', the area behind the 'X' checkbox takes a background colour, as well as the gold outline appearing around the whole element.
Actual: Gold outline appears, but no background colour behind icon.

src/components/pill/pill.spec.js
```diff
@@ -9,14 +9,19 @@ import { classicStyleConfig } from './pill-classic.style';
...
 import smallTheme from '../../style/themes/small';
+import mediumTheme from '../../style/themes/medium';
+import largeTheme from '../../style/themes/large';
 import classicTheme from '../../style/themes/classic';
 import baseTheme from '../../style/themes/base';
 import OptionsHelper from '../../utils/helpers/options-helper';
 
 const classicStyleTypes = [...OptionsHelper.colors, 'disabled'];
 const modernStyleTypes = [...OptionsHelper.pillColours, 'warning'];
+// const modernRoles = [...OptionsHelper.pillRoles];
```
chrisbarber86: to remove?

ianforster-sage: Yes.

src/components/pill/pill.component.js
```diff
@@ -70,7 +75,8 @@ Pill.defaultProps = {
   as: 'default',
   fill: false,
   onClick: null,
-  onDelete: null
+  onDelete: null,
+  role: 'tag'
```
grabkowski: I think that's not a valid ARIA role

src/components/pill/pill.style.js
```diff
@@ -12,26 +13,25 @@ import OptionsHelper from '../../utils/helpers/options-helper';
...
  ${({
-    colourVariant, theme, inFill, isDeletable
+    colourVariant, theme, inFill, isDeletable, role
   }) => {
     const { colors } = baseTheme;
-    const themeName = setTheme(theme);
-    const styleSet = styleConfig[themeName];
+    const correctedTheme = setTheme(theme);
+    const styleSet = styleConfig(correctedTheme)[role];
+    const colour = (role === 'status') ? colourVariant : 'primary';
```
stevebutler2210: We should probably change `colour` and `colourVariant` to Americanised spellings, for consistency with the rest of the code-base

src/components/pill/pill.stories.js
```diff
@@ -10,6 +10,16 @@ import classic from '../../style/themes/classic';
 import { THEMES } from '../../style/themes';
 import { notes, Info } from './documentation';
 
+const getStatusKnobs = () => {
+  const role = select('role', [...OptionsHelper.pillRoles], 'tag');
+
+  return {
+    role,
+    colourVariant: (role === 'status') ? select('colourVariant',
+      [...OptionsHelper.pillColours, OptionsHelper.colors[7]], OptionsHelper.pillColours[0]) : null
```
stevebutler2210: Same as comment below re `colour` and `colourVariant` - `pillColours` should be Americanised for consistency

stevebutler2210: `pillColours` occurrences still need updating

src/components/pill/pill-classic.style.js
```diff
@@ -46,8 +46,8 @@ export const classicStyleConfig = {
   }
 };
 
-export default (colourVariant, inFill, isDeletable) => {
-  const colourSet = classicStyleConfig[colourVariant];
+export default (colorVariant, inFill, isDeletable) => {
+  const colourSet = classicStyleConfig[colorVariant];
```
stevebutler2210: Also just noticed colourSet here which should be updated too (sorry, totally missed this one before!)

src/components/pill/pill.style.js
```diff
@@ -28,6 +28,11 @@ const PillStyle = styled.span`
     const { colors } = baseTheme;
     const correctedTheme = setTheme(theme);
     const styleSet = styleConfig(correctedTheme)[pillRole];
+    const { boxShadow, hoverColor } = styleConfig(correctedTheme);
```
stevebutler2210: As `correctedTheme` isn't used anywhere else, might it be worth instead of that, doing something like this on line 29:
`const correctedThemeConfig = styleConfig(setTheme(theme));`

Then you could use `correctedThemeConfig` rather than repeating the `styleConfig(correctedTheme)` call

src/components/pill/pill.style.js
```diff
@@ -0,0 +1,146 @@
...
+            background-color: ${styleSet[variety].buttonFocus};
+            color: ${(variety === 'warning') ? colors.black : styleSet.hoverColor};
+            cursor: pointer;
+          }
+ 
+          .carbon-icon {
+            font-size: 12px;
+            padding: 0 4px;
+
+            @supports (-ms-ime-align: auto) {
```
grabkowski: `@supports` is not compatible with the Internet Explorer, so it will work only on the Edge Browser

PR #2113, opened 2019-06-10
Fix show records in table pager

edleeks87:
# Description
Hooks up the pager to allow it to be used to adjust the number of  table records per page


src/components/pager/pager.js
```diff
@@ -22,7 +22,7 @@ class Pager extends React.Component {
     /**
      * Total number of records
      */
-    totalRecords: PropTypes.number.isRequired,
+    totalRecords: PropTypes.string.isRequired,
```
dawidzarzycki: Shouldn't we have also defaultProp to totalRecords if we have isRequired in propTypes :)?

edleeks87: When a prop isRequired providing a default value breaks the eslint rules we have set up in Carbon `eslint(react/default-props-match-prop-types)`

src/components/table/table.stories.js
```diff
@@ -113,96 +113,52 @@ const setCurrentPage = ({ totalRecords, pageSize, paginate }) => {
...
+    const Wrapper = () => {
+      const [pageSize, setPageSize] = useState('10');
+
+      const handlePagination = (size) => {
+        setPageSize(size);
+      };
+
+      const tableProps = getCommonKnobs();
+      store.set({ currentPage: setCurrentPage(tableProps) });
+      tableProps.pageSize = tableProps.showPageSizeSelection ? pageSize : text('pageSize', '5');
```
grabkowski: shouldn't that also be in `getCommonKnobs`?

src/components/table/table-story-helpers/table-story.hoc.js
```diff
@@ -0,0 +1,140 @@
...
+        >
+          <TableCell>{row.name}</TableCell>
+          <TableCell>{row.value}</TableCell>
+        </TableRow>
+      ))}
+  </>
+  );
+};
+
+const setCurrentPage = ({ totalRecords, pageSize, paginate }) => {
```
kyanmichael: The name `setCurrentPage` seems a bit misleading here – it sounds like executing the function will actually set the current page, which it doesn't.

Maybe a name like `calculateCurrentPage` would be better? (since all it's doing is calculating and returning the current page number, based on the supplied `tableProps` and the `store.get('currentPage')` value)

(I realise this isn't new code – it's simply been moved from `src/components/table/table.stories.js` – but maybe we could take this opportunity to improve the name?)

edleeks87: Yeah good point I'll make the change 👍

src/components/table/table.stories.js
edleeks87: Agreed, I've separated the knobs out differently

src/components/table/table-story-helpers/table-story.hoc.js
kyanmichael: @edleeks87: Great, many thanks :white_check_mark:

PR #2114, opened 2019-06-10
fixes styling defects

edleeks87:
# Description
Fixes the styling defects found in the base PR: removes the border on the table wrapper, sets a width for the date input when the table is `large` and adjusts the date picker icon so it is aligned vertically in the input

### DIFF
--- a/src/components/table/table-cell/table-cell.style.js
+++ b/src/components/table/table-cell/table-cell.style.js
@@ -105,8 +105,8 @@ function applyModernInputStyling(isTextArea, inputHeight, fontSize) {
       padding-bottom: 0px;
     }
     
-    .carbon-icon {
-      font-size: 13px !important;
+    && .carbon-icon {
+      font-size: 13px;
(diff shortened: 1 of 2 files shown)


src/components/table/table-cell/table-cell.style.js
```diff
@@ -103,6 +104,10 @@ function applyModernInputStyling(isTextArea, inputHeight, fontSize) {
       padding-top: 0px;
       padding-bottom: 0px;
     }
+    
+    .carbon-icon {
+      font-size: 13px !important;
+    }
```
grabkowski: could it be done with `&&.carbon-icon` instead of using the `!important` keyword?

edleeks87: Yeah good point I'll give it a try

edleeks87: Made the changes suggested 👍

PR #2120, opened 2019-06-11
FE-1549 [Profile component]

dawidzarzycki:
# Description
- Update `Profile` component to be style-component based with new DLS standard

- `Portrait` component

# Screenshots / Gifs
Zrzut ekranu 2019-06-11 o 15 01 50


src/components/profile/profile.component.js
```diff
@@ -0,0 +1,82 @@
...
+      >
+        { this.avatar }
+        { this.text }
+      </ProfileStyle>
+    );
+  }
+}
+
+Profile.propTypes = {
+  /** A custom class name for the component. */
```
grabkowski: we need to point out that this is a legacy prop

src/components/profile/profile.spec.js
```diff
@@ -0,0 +1,132 @@
...
+            email='foo@bar.com'
+          />
+        );
+        expect(instance.initials).toEqual('FBB');
+      });
+    });
+
+    describe('avatar', () => {
+      it('returns the portrait component', () => {
+        expect(ProfileAvatarStyle).toBeTruthy();
```
grabkowski: what this test is supposed to do?

src/components/profile/profile.component.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+
+  /** Returns the text portion of the profile. */
+  get text() {
+    return (
+      <ProfileDetailsStyle>
+        <ProfileNameStyle data-element='name'>
+          { this.props.name }
+        </ProfileNameStyle><br />
```
grabkowski: what do you think about using the `<strong>` tag instead of styling it with a styled-component? Looks like this element should be indicated as more important.

src/components/profile/profile.style.js
```diff
@@ -0,0 +1,44 @@
...
+    color: ${({ theme }) => theme.text.color};
+
+    ${({ large }) => large && css`
+        ${ProfileNameStyle} {
+            font-size: 20px;
+            font-weight: 400;
+            line-height: 21px;
+        }
+    `}
+    ${classicProfileStyle}
```
grabkowski: could you add a blank line to separate `classic` from the rest?

src/components/profile/profile-classic.style.js
```diff
@@ -0,0 +1,7 @@
+import { css } from 'styled-components';
+import { THEMES } from '../../style/themes';
+
+
+export default ({ theme }) => theme.name === THEMES.classic && css`
+    color: rgba(0, 0, 0, 0.85);
+`;
```
grabkowski: there is no need to create a separate file if it's only one line. You could just put it on the bottom of a styled-component.

src/components/profile/profile.spec.js
dawidzarzycki: Profile component uses Portrait component to show the avatar. In this text i wanted to check if the avatar rendered

src/components/profile/profile.component.js
dawidzarzycki: The name has different text color in the DLS

src/components/profile/profile-classic.style.js
dawidzarzycki: Ye, I agree with you

src/components/profile/profile.style.js
dawidzarzycki: There is no more classic import

src/components/profile/profile.spec.js
grabkowski: but it will always be truthy

src/components/profile/profile.stories.js
```diff
@@ -1,24 +1,44 @@
...
+    const initials = text('initials', 'JS');
+    const large = boolean('large', OriginalProfile.defaultProps.large);
+    const name = text('name', 'John Smith');
+
+    return (
+      <ThemeProvider theme={ classicTheme }>
+        <Profile
+          email={ email }
+          initials={ initials }
+          large={ large }
```
ianforster-sage: Is this still correct? Should it use `size` now?

dawidzarzycki: It's only for the classic

PR #2123, opened 2019-06-12
Fix table input defects

edleeks87:
# Description
Fixes issue with the width of date inputs when table is large, adds prop to isolate and toggle just the date width


src/components/table/table-cell/table-cell.spec.js
```diff
@@ -122,11 +123,19 @@ describe('TableCell', () => {
         });
 
         it('renders a textarea that matches the expected style', () => {
-          wrapper = mount(
+          const wrapper1 = mount(
```
grabkowski: why do we need another wrapper?

src/components/table/table-cell/table-cell.component.js
```diff
@@ -30,17 +31,26 @@ class TableCell extends React.Component {
...
    * Returns if child element is Textarea.
    */
-  childrenHasTextArea() {
+  get checkChildrenInputType() {
     const { children } = this.props;
 
     if (!Array.isArray(children)) {
-      return this.childName(children) === TextArea.name;
+      if (this.childName(children) === TextArea.name) {
+        return 'isTextArea';
```
ianforster-sage: Please could you define these as constants rather than magic strings

edleeks87: Yeah I'll add them to the options helper

src/components/table/table-cell/table-cell.spec.js
edleeks87: I need to run two tests to meet coverage, I'll remove the extra wrapper and just reassign it the new component

PR #2127, opened 2019-06-13
Fix pager defects

edleeks87:
# Description
Fixes styling defects in pager component


src/components/pager/pager.styles.js
```diff
@@ -90,10 +99,11 @@ const PagerLinkStyles = styled.button`
   &:focus {
     outline: none;
   }
+  color: rgba(0,0,0,0.90);
 
   ${({ isDisabled }) => isDisabled && css`
-    color: rgba(0,0,0,0.3);
-  `};
+    color: rgba(0,0,0,0.55);
```
izakrzewska: don't we have this color in a config?

edleeks87: Not as far as I can see, I'll add them to the config though so the colours aren't hardcoded

PR #2131, opened 2019-06-17
add cypress tests to pipeline

marcinstachowicz:
## Add cypress tests to pipeline
- add wait-on to wait for url
- impove babel path because travis fails


marcinstachowicz:
We reduce cypress test + env configuration (for cypress) from 4:40 min to 2:40

chrisbarber86:
Let's keep an eye on this and if the time starts to dramatically impact the build we can revisit.

PR #2133, opened 2019-06-17
Fixes accessibility defects in Pager component

edleeks87:
# Description
Adds label to input, sets cursor for disabled buttons, removes `withTheme()` HOC from Pager and uses callback instead to identify theme

### DIFF
--- a/src/components/pager/pager.component.js
+++ b/src/components/pager/pager.component.js
@@ -2,7 +2,6 @@ import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import I18n from 'i18n-js';
 import Immutable from 'immutable';
-import { withTheme } from 'styled-components';
 import PagerNavigation from './pager-navigation.component';
 import {
   PagerContainerStyles,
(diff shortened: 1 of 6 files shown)


src/components/pager/pager.component.js
```diff
@@ -57,6 +56,7 @@ const Pager = (props) => {
         { ...props }
         currentPage={ currentPage }
         setCurrentPage={ setCurrentPage }
+        currentTheme={ setCurrentTheme }
```
stevebutler2210: Shouldn't this use the getter?

edleeks87: No it's a callback to allow the navigation sub component to set the theme in the parent, it probably needs a better name for the prop though

edleeks87: What do you think if I rename the prop to `setcurrentThemeName` and pass in the hook setter as a callback to it?

stevebutler2210: Sorry, yep gotcha. Yeah that reads a bit more clearly 👍

PR #2135, opened 2019-06-17
Various CSS changes based on UX review

whomwah:
### S1T-1930 - Remove carbon-dropdown__input--filtered in line with UX review

> Text colour should be 85% black (rgb 0,0,0,0.85) throughout.
> Although the text colour of the original value of a dropdown is correct, if you change the value, then text colour turns to a dark slate colour (#335c6d).

This PR removes the colour change and keeps it as the default.

### S1T-1941 - Poor text contrast

Fix poor contrast on disabled inputs

### Dropdown text colour on filtering
dropdown

### Contrast colour on disabled input
input


whomwah:
Guys, I'm going to batch up a couple of these requested colour changes to save effort. I'll update this PR and push the updates.

whomwah:
OK, Now ready for re-review 👍 I've only added 1 extra change

chriswaymansage:
Curious at what the 'right use of important' actual is 🤔

edleeks87:
I just try to avoid it unless I've exhausted all other attempts to override the styles. As we move to styled-components we can use `&&` to set the priority for a given components style so I guess it's moot

PR #2141, opened 2019-06-19
FE-1411 inline inputs

grabkowski:
Updated the InlineInputs component to match new DLS requirements


src/components/inline-inputs/inline-inputs.stories.js
```diff
@@ -1,25 +1,64 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { text } from '@storybook/addon-knobs';
-import InlineInputs from './inline-inputs';
-import Textbox from '../textbox';
+import { State, Store } from '@sambego/storybook-state';
+import InlineInputs from './inline-inputs.component';
```
izakrzewska: `import InlineInputs from '.';`?

grabkowski: done

src/components/inline-inputs/inline-inputs.component.js
```diff
@@ -0,0 +1,68 @@
...
+  /** The id of the corresponding input control for the label */
+  htmlFor: PropTypes.string
+};
+
+InlineInputs.defaultProps = {
+  children: null,
+  className: ''
+};
+
+/* eslint-enable react/prop-types */
```
edleeks87: Why is this line here, I can't see where the prop-types were disabled?

grabkowski: I don't know, this file is just renamed (original is `inline-inputs.js`). I haven't toucher that part of code.

grabkowski: removed

harpalsingh:
There is an update we would like to do on the alignment of the inputs.

Currently, we have the inputs sit side by side, but that results in a 2px border:
Screenshot 2019-06-25 at 11 00 45

Could we update to have the borders feel like a single border, therefore each child div, would need a +1 offset (incremental) to make them feel aligned as a single border.

Screenshot 2019-06-25 at 11 00 30

PR #2147, opened 2019-06-21
FE-1326 - [Configurable Items]

dawidzarzycki:
# Description
- add new DLS version of configurable items

# Screenshots / Gifs
Configurable-items

# Testing instructions 
- Little changes in the `Checkbox` component so `Checkbox` need retesting


src/components/configurable-items/configurable-item-row/configurable-items-row.spec.js
```diff
@@ -0,0 +1,228 @@
...
+
+  describe('enabled', () => {
+    describe('when the row is enabled', () => {
+      beforeEach(() => {
+        wrapper = shallow(
+          <ConfigurableItemRowWithoutHOC enabled rowIndex={ 1 } />
+        );
+      });
+
+      it('sets the Checkbox value to true', () => {
```
ianforster-sage: If the value here can be only `true` or `false` it would be better to test for that Ie. `toBe(true)` rather than risking a incorrect value being coerced to a boolean.

src/components/configurable-items/configurable-item-row/configurable-items-row.spec.js
```diff
@@ -0,0 +1,228 @@
...
+
+    describe('when the row is not enabled', () => {
+      beforeEach(() => {
+        wrapper = shallow(
+          <ConfigurableItemRowWithoutHOC enabled={ false } rowIndex={ 1 } />
+        );
+      });
+
+      it('sets the Checkbox value to false', () => {
+        expect(wrapper.find(Checkbox).props().value).toBeFalsy();
```
ianforster-sage: If the value here can be only `true` or `false` it would be better to test for that Ie. `toBe(true)` rather than risking a incorrect value being coerced to a boolean.

src/components/configurable-items/configurable-item-row/configurable-items-row.spec.js
dawidzarzycki: fixed

src/components/configurable-items/configurable-item-row/configurable-items-row.spec.js
dawidzarzycki: fixed

PR #2148, opened 2019-06-21
FE-1561 textarea validation

grabkowski:
Added the validation HOC to the Textarea component


src/__experimental__/components/input/input-presentation.style.js
```diff
@@ -57,18 +57,29 @@ const InputPresentationStyle = styled.div`
...
-  return styling;
+function stylingForValidations({
+  theme,
+  hasError,
+  hasWarning,
+  hasInfo
+}) {
+  let validationColor;
+
+  if (hasError) {
```
dawidzarzycki: What do you think to use switch case if we have more than one else if?

PR #2151, opened 2019-06-25
revert for a sanity tests

DlgSHi:
reverting two methods for a click into accessibility tests and reRun tests for a sanity tests

### DIFF
--- a/cypress/features/sanity/sanity.feature
+++ b/cypress/features/sanity/sanity.feature
@@ -4,6 +4,8 @@ Feature: Sanity tests
   @sanity
   Scenario Outline: Component <component> page with button
     When I open "<component>" component page with button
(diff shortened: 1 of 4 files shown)


cypress/features/sanity/sanity.feature
```diff
@@ -13,6 +15,8 @@ Feature: Sanity tests
   @sanity
   Scenario Outline: Component <component> page and open preview
     When I open "<component>" component page
+      And click into Accessibility Tab
+      And click reRun tests button
```
marcinstachowicz: one step, no click word

DlgSHi: I leaved two steps, but changed the name of steps

PR #2118, opened 2019-06-11
FE-1330 - Tooltip Warning / Info DLS

stevebutler2210:
- re-write Tooltip as styled component
- update the TooltipDecorator to use the new component
- also updates the Icon spec to import the new Tooltip
- adds Story, documentation etc

Standalone:
Screenshot 2019-06-11 at 10 05 13

Warning:
Screenshot 2019-06-11 at 10 06 00

Info:
Screenshot 2019-06-11 at 10 05 50

PR #2155, opened 2019-06-26
DLS Tooltip - duplicate of PR #2118

stevebutler2210:
Identical to closed PR https://github.com/Sage/carbon/pull/2118, which was merged early in error before UX review had been conducted (reverted here https://github.com/Sage/carbon/pull/2154).


harpalsingh:
When using the "warning" style, we still see it as a dark tooltip?

What we see:
Screenshot 2019-06-26 at 11 03 58

Could the font weight be reverted to use 700 for all designs, same as classic so we see the bold weight:

currently:
Screenshot 2019-06-26 at 11 04 47

What we would like:
Screenshot 2019-06-26 at 11 04 44

In the design here http://designsystem.sage.com/patterns/help-tooltips/ the tooltips have a short fade in animation, does this apply to the tooltips? Triggering isVisible does not seem to show if this does that.

Could the padding please be updated to be 8px for top/bottom and 16px for left and right.

stevebutler2210:
Thanks for your feedback @harpalsingh 

> When using the "warning" style, we still see it as a dark tooltip?

This is as per the designs here:
http://designsystem.sage.com/components/text-inputs/

> Could the font weight be reverted to use 700 for all designs, same as classic so we see the bold weight

Can do - as this is a deviation from the current designs, will these be updated to reflect the change?

>In the design here http://designsystem.sage.com/patterns/help-tooltips/ the tooltips have a short fade in animation, does this apply to the tooltips? Triggering isVisible does not seem to show if this does that.

Yep, totally missed this - guessing this should be applied at all times for all tooltips using one of the new themes?

> Could the padding please be updated to be 8px for top/bottom and 16px for left and right.

Another good catch, will update, sorry!

harpalsingh:
Thanks @stevebutler2210 

I'll confirm on the warning, but for the rest, once we see it in a build, it helps to determine if there are any tweaks on the original design and where changes are suggested here, they will then be updated on the design system portal in short time, if they vary.

harpalsingh:
ok with leaving warning as it is 👍

PR #2173, opened 2019-07-05
Docker setup for running cypress test suite

chrisbarber86:
# Description

Docker setup for running cypress automation test suite.


package.json
```diff
@@ -71,7 +71,7 @@
     "core-js": "^3.1.4",
     "cpx": "^1.5.0",
     "cross-env": "^5.2.0",
-    "cypress": "^3.2.0",
+    "cypress": "3.3.0",
```
marcinstachowicz: good approach is to use 3.2.0 cypress version because 3.3.0 causes failing tests

chrisbarber86: I had to switch this to 3.3.0 because there was an issue in 3.2.0 that prevented this. Specifically https://github.com/cypress-io/cypress/issues/3742

chrisbarber86: If i switch back to 3.2.0 then the test suite doesn't run because it doesn't boot.

docker-compose-cypress.yml
```diff
@@ -0,0 +1,22 @@
...
+    command: bash -c 'cypress run'
+    environment:
+      cypress_localhost: http://storybook:9001/
+    depends_on:
+      - storybook
+    # Reason for the below `ipc: host` line:
+    # https://github.com/cypress-io/cypress/issues/350
+    # https://docs.cypress.io/guides/references/error-messages.html#The-Chromium-Renderer-process-just-crashed
+    ipc: host
+    # volumes:
```
stevebutler2210: If the volumes setup can these 2 lines be removed?

chrisbarber86: So that's really useful for when running this locally, which is why i left it in but commented out. We may also look at getting CI to upload these screenshots to S3 or something in future.

PR #2182, opened 2019-07-12
FE-1413 help component

grabkowski:
- Moved styles from SCSS to styled-component,
- Added DLS styling,


src/components/help/help.style.js
```diff
@@ -0,0 +1,32 @@
...
+    cursor: pointer;
+    text-decoration: none;
+
+    &:hover {
+      color: $blue;
+      text-decoration: underline;
+    }
+  `}
+
+  ${({ theme }) => theme.name === THEMES.classic && css`
```
izakrzewska: isClassic helper can be used here

src/components/help/help.component.js
```diff
@@ -0,0 +1,58 @@
...
+        tooltipPosition={ tooltipPosition }
+        tooltipAlign={ tooltipAlign }
+      />
+    </StyledHelp>
+  );
+};
+
+Help.propTypes = {
+  /** A custom class name for the component. */
+  className: PropTypes.string,
```
dawidzarzycki: should we add information about legacy prop?

harpalsingh:
Should all configs in the knobs work?

Using tooltipPosition: right and tooltipAlign: right, we get the following visual (same for left, but top and bottom position work fine with those alignments)

Screenshot 2019-07-16 at 10 56 05

grabkowski:
> Should all configs in the knobs work?
> 
> Using tooltipPosition: right and tooltipAlign: right, we get the following visual (same for left, but top and bottom position work fine with those alignments)
> 
> Screenshot 2019-07-16 at 10 56 05

The Component behaved the same way in the previous version so I assume that this is the correct behaviour.

harpalsingh:
@grabkowski the problem is more to do with left/right alignment would never really need any additional config, so its more of a config mix issue, so not a problem.

PR #2194, opened 2019-07-15
Date component - The visible value is empty if the value is empty

gusch:
# Description
When the value is string.Empty, the visible value is also empty (and not the date of today)
The datePickerValue is set to today so the datePicker does not crash.
With the prop showPickerOnTop, the datePicker will appear above the input

### DIFF
--- a/src/components/date/date.js
+++ b/src/components/date/date.js
@@ -109,7 +109,15 @@ const Date = Input(InputIcon(InputLabel(InputValidation(class Date extends React
      * @property value
      * @type {String}
      */
-    value: PropTypes.string
+    value: PropTypes.string,
+
+    /**
+     * Display the date of today if the value is empty
+     *
+     * @property readOnly
+     * @type {Boolean}
+     */
+    showTodayIfEmpty: PropTypes.bool
   };
 
   static defaultProps = {
@@ -129,7 +137,16 @@ const Date = Input(InputIcon(InputLabel(InputValidation(class Date extends React
     * @type {Array}
     * @default DateValidator
     */
-    internalValidations: [new DateValidator()]
+    internalValidations: [new DateValidator()],
+
+    /**
+     * Display the date of today if the value is empty
+     *
+     * @property showTodayIfEmpty
+     * @type {Boolean}
+     * @default true
+     */
+    showTodayIfEmpty: true
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/components/date/date.js
```diff
@@ -647,6 +647,10 @@ const Date = Input(InputIcon(InputLabel(InputValidation(class Date extends React
    * @return {String} formatted visible value
    */
   formatVisibleValue(value) {
+    if (value === '') {
```
LPTSage: why do you not also add a prop to the component to allow a consumer / developer to choose which behaviour they want ?

your change may break existing code using this component that which expects the date to automatically be `today` when there is no value

gusch: @LPTSage You're right. I updated the code.

src/components/date/date.js
```diff
@@ -129,7 +137,16 @@ const Date = Input(InputIcon(InputLabel(InputValidation(class Date extends React
...
+    internalValidations: [new DateValidator()],
+
+    /**
+     * Display the date of today if the value is empty
+     *
+     * @property showTodayIfEmpty
+     * @type {Boolean}
+     * @default true
+     */
+    showTodayIfEmpty: true
```
edleeks87: Whilst this work I think it's unnecessary to add the prop in this way, you could have a boolean prop called `allowEmptyValue` that would allow the date input to display an empty string. Doing this maintains backwards compatibility and without having to set a default prop to do so.

src/components/date/date.js
```diff
@@ -647,6 +664,10 @@ const Date = Input(InputIcon(InputLabel(InputValidation(class Date extends React
    * @return {String} formatted visible value
    */
   formatVisibleValue(value) {
+    if (!this.props.showTodayIfEmpty && value === '') {
```
edleeks87: This can then be changed to:
```
 if (this.props.allowEmptyValue && !value.length) {
      return '';
    }
```

edleeks87:
@gusch everything looks good, the build has failed as the tests need updating for the date component, happy to approve once they pass and complete code coverage is hit 👍

gusch:
Fix does not work

gusch:
I needed to set the datePickerValue to the date of today if the date is not valid and allowEmptyValue is true

sianford:
@gusch I tested this in the Carbon Demo Site but could not get the date picker to be activated when the showPickerOnTop prop was selected.

sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge. No further issues logged. The date picker is now appearing correctly when triggered. An issue regarding the date picker appearing at a distance from the date input when the showPickerOnTop prop was set to true has been resolved. Testing passed ✅

PR #2199, opened 2019-07-17
fixes warnings for labeAlign

edleeks87:
# Description
Adds a default value to the `labelAlign` knob that is passed as prop to `FormField`

### DIFF
--- a/src/__experimental__/components/select/select.stories.js
+++ b/src/__experimental__/components/select/select.stories.js
@@ -24,8 +24,10 @@ const commonKnobs = (store) => {
     errorMessage: text('errorMessage', ''),
     infoMessage: text('infoMessage', ''),
     label,
-    labelAlign: label !== '' ? select('labelAlign', OptionsHelper.alignBinary) : undefined,
-    labelInline: label !== '' ? boolean('labelInline', false) : undefined,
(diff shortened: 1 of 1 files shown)


src/__experimental__/components/select/select.stories.js
```diff
@@ -24,8 +24,10 @@ const commonKnobs = (store) => {
     errorMessage: text('errorMessage', ''),
     infoMessage: text('infoMessage', ''),
     label,
-    labelAlign: label !== '' ? select('labelAlign', OptionsHelper.alignBinary) : undefined,
-    labelInline: label !== '' ? boolean('labelInline', false) : undefined,
+    labelAlign: (
+      label.length ? select('labelAlign', OptionsHelper.alignBinary, OptionsHelper.alignBinary[0]) : undefined
```
ianforster-sage: I don't like setting a value as `undefined` it should remain a value only used by the engine, please can you use `null` instead.

edleeks87: Setting it to `null` will pass the prop down to the component and cause the warnings to appear again, setting it to `undefined` simply prevents the prop being passed. Instead I've just refactored it to remove the ternaries.

PR #2209, opened 2019-07-19
FE-1673 Portrait bug (transparent image backgrounds)

kyanmichael:
# Description

Bug: If a Gravatar or custom image has a transparent background, then the "fallback" initials / :bust_in_silhouette: user icon will incorrectly be visible behind the image.

This example uses [the Gravatar for `info@cloudinary.com`](https://www.gravatar.com/avatar/e3264cf16f34ecd3c7c564f5668cbc1e.png) which has a transparent background:

Bug

This PR fixes this bug, by only rendering the "fallback" initials / :bust_in_silhouette: user icon if the Gravatar or custom image fails to load (ie: if it triggers the ``'s [`onerror`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) event handler).

For Gravatar, this PR changes [the fallback option](https://en.gravatar.com/site/implement/images/#default-image) from `blank` (_"a transparent PNG image"_) to `404` (_"do not load any image if none is associated with the email hash, instead return an HTTP 404 File Not Found response"_).


kyanmichael:
@edleeks87: Ah yes, good point, thank you. Added in 3f42cb7 :thumbsup:

PR #2213, opened 2019-07-23
FE-1655 updated styles for small, medium and large themes. Updated sp…

damianOfKyan:
#Description

Implemented Medium and Large views for Pill component. Updated classic and base view styles. Updated specs.


src/components/pill/pill.style.js
```diff
@@ -113,6 +115,18 @@ const PillStyle = styled.span`
...
+        font-size: 12px;
+        border-radius: 16px;
+        padding-bottom: 4px;
+      `}
+
+      ${theme.name === 'large' && css`
+        font-size: 14px;
+        border-radius: 24px;
+        padding-bottom: 4px;
+      `}
```
grabkowski: Are you sure that font sizes for small business, medium business and large business themes should differ? I think that the size should be a separate prop.

stevebutler2210: ^^^ agreed

src/components/pill/pill.spec.js
```diff
@@ -362,16 +361,133 @@ describe('Pill', () => {
...
+        padding: '4px 8px 2px 8px',
         margin: '0px 8px 16px 0px'
       }, wrapper);
     });
   });
 
+  describe('classic theme', () => {
+    const renderBase = (props, renderer = mount) => {
+      return renderer(
+        <ThemeProvider theme={ classicTheme }>
```
stevebutler2210: This should be called renderClassic.

I'm also not sure any of these theme-specific methods are actually required, I would have thought you could just do:

```
const render = (props, renderer = mount) => {
  return renderer(
    <Pill { ...props } />
  )
}
```
and have `theme: classicTheme` or `theme: `baseTheme` in the props you pass. (Might not even need baseTheme for that one, as it may already be the default).

Then pull the function out a level so all your describe blocks can access it, and you'll save on a lot of repetition.

I would also check if the methods are ever currently being called _without_ passing `TestRenderer.create` as the renderer param - if they aren't, you can remove the param entirely and hard-code that, and if they are but only a few, then maybe worth looking at making `TestRenderer.create` the default param instead of `mount`.

src/components/pill/pill.spec.js
```diff
@@ -362,16 +361,133 @@ describe('Pill', () => {
...
+      assertStyleMatch({
+        borderRadius: '10px',
+        fontSize: '12px',
+        minWidth: '24px'
+      }, wrapper);
+    });
+  });
+
+  describe('base theme', () => {
+    const renderBase = (props, renderer = mount) => {
```
stevebutler2210: Same here. Also, there is already a describe block for base theme further up, so these tests should be moved under that.

src/components/pill/pill.spec.js
```diff
@@ -362,16 +361,133 @@ describe('Pill', () => {
...
+      assertStyleMatch({
+        borderRadius: '12px',
+        fontSize: '10px',
+        minWidth: '24px'
+      }, wrapper);
+    });
+  });
+
+  describe('small theme', () => {
+    const renderBase = (props, renderer = mount) => {
```
stevebutler2210: Same as other comment re separate render methods

src/components/pill/pill.spec.js
```diff
@@ -362,16 +361,133 @@ describe('Pill', () => {
...
+        top: '-1px',
+        padding: '4px 8px 2px 8px',
+        margin: '0px 8px 16px 0px',
+        textAlign: 'center'
+      }, wrapper);
+    });
+  });
+
+  describe('medium theme', () => {
+    const renderBase = (props, renderer = mount) => {
```
stevebutler2210: Same as other comment re separate render methods

src/components/pill/pill.spec.js
```diff
@@ -362,16 +361,133 @@ describe('Pill', () => {
...
+        borderRadius: '16px',
+        fontSize: '12px',
+        minWidth: '24px',
+        paddingBottom: '4px'
+      }, wrapper);
+    });
+  });
+
+  describe('large theme', () => {
+    const renderBase = (props, renderer = mount) => {
```
stevebutler2210: Same as other comment re separate render methods

damianOfKyan: resolved

src/components/pill/pill.spec.js
damianOfKyan: resolved

src/components/pill/pill.spec.js
damianOfKyan: resolved

src/components/pill/pill.spec.js
damianOfKyan: resolved

src/components/pill/pill.spec.js
damianOfKyan: stuck with one render method, indeed there was no need to have many.

src/components/pill/pill.style.js
damianOfKyan: agree, size is now separate property.

src/components/pill/pill-sizes.style.js
```diff
@@ -0,0 +1,47 @@
+export default (css, isClassic, isDeletable) => ({
```
stevebutler2210: I'm really struggling to work out what this file is / does? Is it meant to be in here? As far as I can tell, it isn't being imported anywhere, and wouldn't work because it has problems (uses methods like `css`, `isClassic` when they haven't been imported etc)? Feels like I'm missing something

src/components/pill/pill.style.js
```diff
@@ -99,20 +99,200 @@ const PillStyle = styled.span`
...
             &:focus {
               color: ${(variety === 'warning') ? colors.black : styleSet.hoverColor};
             }
-            &:before {
-              font-size: 12px;
+          }
+        }
+
+        ${size === 'small' && css`
+
```
stevebutler2210: This empty line can be deleted

src/components/pill/pill.style.js
```diff
@@ -99,20 +99,200 @@ const PillStyle = styled.span`
...
               color: ${(variety === 'warning') ? colors.black : styleSet.hoverColor};
             }
-            &:before {
-              font-size: 12px;
+          }
+        }
+
+        ${size === 'small' && css`
+        
+        min-width: 38px;
```
stevebutler2210: And this and line 114 should be indented an extra level

src/components/pill/pill.style.js
```diff
@@ -99,20 +99,200 @@ const PillStyle = styled.span`
...
+                  font-size: 11px;
+                }
+              }
+            }
+        `}
+      `}
+
+      ${!isClassic(theme, colorVariant) && !isDeletable && css`
+        ${size === 'small' && css`
+        font-size: 10px;
```
stevebutler2210: This and the 2 lines below it should be indented another level

src/components/pill/pill.style.js
```diff
@@ -99,20 +99,200 @@ const PillStyle = styled.span`
...
+                  font-size: 11px;
+                }
+              }
+            }
+        `}
       `}
 
       ${isClassic(theme, colorVariant) && classicThemeForPill(colorVariant, inFill, isDeletable)}
+      
+      ${isClassic(theme, colorVariant) && !isDeletable && css`
```
stevebutler2210: All the Classic styles should probably go in `pill-classic.style.js`

src/components/pill/pill-classic.style.js
```diff
@@ -87,6 +87,30 @@ export default (colorVariant, inFill, isDeletable) => {
...
+              font-size: 8px;
+            }
+          }
+        }
+      `}
+
+      ${size === 'medium' && css`
+        font-size: 12px;
+        min-width: 38px;
+        padding: 4px 8px 4px 8px;
```
dawidzarzycki: We can rewrite it to: `padding: 4px 8px`

src/components/pill/pill-classic.style.js
```diff
@@ -87,6 +87,30 @@ export default (colorVariant, inFill, isDeletable) => {
...
+      `}
+
+      ${size === 'large' && css`
+        font-size: 14px;
+        min-width: 46px;
+        padding: 4px 8px 4px 8px;
```
dawidzarzycki: Same as above

src/components/pill/pill.spec.js
```diff
@@ -308,6 +571,53 @@ describe('Pill', () => {
...
+          }, TestRenderer.create).toJSON();
+          assertStyleMatch({
+            fontSize: '12px',
+            minWidth: '38px',
+            padding: '4px 10px 4px 4px'
+          }, wrapper);
+        });
+      });
+
+      describe('when the component size is large', () => {
```
dawidzarzycki: I see that many times there is...
```
description(when the component size is large){
  it('test 1')
}

description(when the component size is large){
  it('test2')
}
```
I would like to ask is there a possbility to do something like

```
description(when the component size is large){
  it('test 1');
  it('test 2');
  it('test 3');
}

```

src/components/pill/pill.stories.js
```diff
@@ -25,8 +25,10 @@ const getKnobs = (theme) => {
     children: text('children', 'Pill'),
     fill: boolean('fill', Pill.defaultProps.fill),
     onDelete: boolean('onDelete', false),
-    theme
+    theme,
+    size: select('size', OptionsHelper.sizesRestricted, Pill.defaultProps.size)
   };
+
   if (theme === THEMES.classic) {
```
dawidzarzycki: I know this isn't in your changes but if you will change something in this file, can you rewrite it to use `isClassic()` method please?

src/components/pill/pill.spec.js
damianOfKyan: It is good point, unfortunately its separate due to permutations of "isDeletable".

src/components/pill/pill-classic.style.js
damianOfKyan: updated.

src/components/pill/pill-classic.style.js
damianOfKyan: updated.

src/components/pill/pill.stories.js
damianOfKyan: updated.

src/utils/helpers/style-helper/style-helper.js
```diff
@@ -1,6 +1,7 @@
 import { THEMES } from '../../../style/themes';
 
 // eslint-disable-next-line import/prefer-default-export
-export function isClassic({ name }) {
-  return name === THEMES.classic;
+export function isClassic(theme) {
+  if (theme === undefined) return false;
```
dawidzarzycki: Why do we need the condition?

src/utils/helpers/style-helper/style-helper.spec.js
```diff
@@ -11,3 +11,11 @@ describe.each([[THEMES.classic, Classic], [THEMES.small, Small]])(
     });
   }
 );
+
```
dawidzarzycki: If we remove condition from isClassic(), that test is not needed anymore

src/components/pill/pill-classic.style.js
```diff
@@ -52,12 +52,14 @@ export default (colorVariant, inFill, isDeletable, size) => {
   return css`
     border-radius: 10px; 
     display: inline-block;
-    font-size: 12px;
+    font-size: 10px;
     font-weight: 700;
-    line-height: 15px;
+    line-height: 10px;
```
edleeks87: Apologies in advance if I'm missing something glaringly obvious, has this change been okayed with UX @damianOfKyan? Just asking as I think we are trying to leave `classic` styling "as is" wherever possible to maintain compatibility with legacy uses of the components so I wouldn't expect to see changes to the style properties

src/components/pill/pill.style.js
```diff
@@ -99,20 +89,196 @@ const PillStyle = styled.span`
...
+              }
+            }
+          }
+        `}
       `}
 
-      ${isClassic(theme, colorVariant) && classicThemeForPill(colorVariant, inFill, isDeletable)}
+      ${isClassic(theme, colorVariant) && classicThemeForPill(colorVariant, inFill, isDeletable, size)}
+
+      ${size === 'S' && css`
```
stevebutler2210: Are all the styles on lines 259-281 the classic styles? If so these should go in `src/components/pill/pill-classic.style.js` (the call to `classicThemeForPill` above would then return the styles from that file if the theme is classic.

With this file written as it is at the moment, all of the styles appearing below line 257 will override the styles defined in `src/components/pill/pill-classic.style.js`. The call to import the classic styles should always be at the bottom of your styled components file so that it overrides the DLS styles defined above it in this file.

PR #2214, opened 2019-07-23
FE-1761 fixed bug with pages not switching.

damianOfKyan:
# Description

Fixed bug with pages not switching due to wrong page index being provided to onClick handleSlide event.


damianOfKyan:
due to slide/carousel being glitchy going to re-visit.

PR #2215, opened 2019-07-23
FE-1761 fixed bug with pages not switching.

damianOfKyan:
# Description

Page slides would glitch due to dependency of the 'name' property (that stores page/slide number) of Button which was being removed from DOM. Depending how quickly button element would be removed from DOM Carousel slider would work or not.


change_log/v9_0_0/FE-1761-pages-switch-fix.yaml
```diff
@@ -0,0 +1 @@
+Component Fix: Page slides would glitch due to dependency of the 'name' property (that stores page/slide number) of Button which was being removed from DOM. Depending how quickly button element would be removed from DOM Carousel slider would work or not.
```
grabkowski: That description suggests that there was an issue with a component code and the issue was with a story.

edleeks87: Agree with @grabkowski, would expect to see a description of the change made rather than what was wrong. The code changes look fine to me and I'll be happy to approve once you've updated

PR #2224, opened 2019-07-29
FE 1220 error validation dls

grabkowski:
- Added DLS styling for error validation,
- Fixed wrong tooltip pointer repositioning when position is set to "top" and the icon is close to the screen border,
- Added tooltip repositioning when tooltip position is set to "right" and the icon is close to the border,
- Added "extractProps" util to filter out props based on Prop Types of the provided component,
- Moved rendering of the ValidationIcon to InputIconToggle component to allow seamless display of the validation tooltip when hovering over the Textbox elements,
- Removed redundant props from the InputPresentation and InputIconToggle components.


src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -163,198 +122,145 @@ const TooltipDecorator = (ComposedComponent) => {
...
-            styleElement(tooltip, 'top', append(shifts[`side${startCase(alignment)}`], 'px'));
-            styleElement(tooltip, 'bottom', 'auto');
-            styleElement(tooltip, 'left', append(shifts[`${position}Horizontal`], 'px'));
-        }
+      if (!this.isVisible()) {
+        return;
+      }
+
+      const tooltip = this.getTooltip(),
+          target = this.getTarget();
```
dawidzarzycki: Just little one, would you like to create both const the same way?
I mean:
```
const tooltip = this.getTooltip();
const target = this.getTarget();
```
I think this is more readable

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -163,198 +122,145 @@ const TooltipDecorator = (ComposedComponent) => {
...
+
+      if (!tooltip || !target) {
+        if (this.state.isVisible) this.setState({ isVisible: false });
+        return;
+      }
+
+      const alignment = this.state.tooltipAlign || this.props.tooltipAlign || 'center',
+          position = this.state.tooltipPosition || this.props.tooltipPosition || 'top',
```
dawidzarzycki: Same as above

src/components/tooltip/tooltip.spec.js
```diff
@@ -215,10 +216,10 @@ describe('Tooltip', () => {
             it('applies the correct default &:before styles', () => {
               assertStyleMatch(
                 {
-                  borderTop: '7px solid transparent',
+                  borderTop: `${pointerSize}px solid transparent`,
```
edleeks87: These tests seem to almost repeat as they assert that the style is applied `before` the element based on the position. You could use `describe.each` for these as well, and could then assert that the style is applied to the border like so 
`border${position}Color: '#C7384F' `

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -163,198 +122,145 @@ const TooltipDecorator = (ComposedComponent) => {
...
+          shifts = this.calculatePosition(tooltip, target);
+
+      this.realignOffscreenTooltip(shifts, tooltip.offsetWidth);
+
+      styleElement(tooltip, 'bottom', 'auto');
+      styleElement(tooltip, 'right', 'auto');
+
+      switch (position) {
+        case 'top':
```
edleeks87: Aren't the `top` and `left` cases redundant and can be removed?

src/__experimental__/components/input/input-presentation.spec.js
```diff
@@ -180,7 +169,19 @@ describe('InputPresentation', () => {
...
+      }, renderWithTheme({ disabled: true }, classicTheme));
     });
   });
 });
+
+function render(props, renderer = mount) {
+  return renderer(<InputPresentation { ...props }>sample children</InputPresentation>);
+}
+
+function renderWithTheme(props, theme, renderer = mount) {
```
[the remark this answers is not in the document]
grabkowski: Yes, it's required to pass the theme to styled components. We are filtering unnecessary props, basing on PropTypes, to avoid passing `data` attributes and other html stuff to styled components. Theme is filtered out as it should be added with a provider.

src/components/tooltip/__snapshots__/tooltip.spec.js.snap
```diff
@@ -31,23 +31,22 @@ exports[`Tooltip visible and has children matches snapshot 1`] = `
...
   height: 0;
+}
+
+.c2:before {
   border-top: 8px solid #000000;
   border-right: 7px solid transparent;
-  border-bottom: 7px solid transparent;
-  border-left: 7px solid transparent;
   border-bottom: none;
+  border-left: 7px solid transparent;
```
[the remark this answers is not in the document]
grabkowski: yes, we are drawing the pointer as a border

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
grabkowski: they aren't redundant - top uses the same methods as bottom and left uses the same as right. There is no `break;` between them.

edleeks87: Yep good point 🤦‍♂, `note to self, don't do reviews before you've had coffee`

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -83,52 +85,17 @@ import { styleElement, append } from '../../ether';
...
-     * @property _hideTooltipTimeout
-     */
     _hideTooltipTimeout = null;
 
     /**
      * Cache the shifts calculations (used for positioning)
-     *
-     * @property _memoizedShifts
      */
     _memoizedShifts = null;
```
kyanmichael: Currently, this `_memoizedShifts` property doesn't seem to do anything useful (there's nowhere in the code where `_memoizedShifts` is set to a value other than `null`).

In my [`FE-1847-tooltip-positioning`](https://github.com/Sage/carbon/commits/FE-1847-tooltip-positioning) branch, I deleted `_memoizedShifts`, and instead I implemented an optimisation where `styleElement()` is only called if the CSS attribute's value has changed – see [`setPositionAttributeIfChanged()` in `tooltip-position.config.js`](https://github.com/Sage/carbon/blob/1a0afac/src/components/tooltip/tooltip-position.config.js#L11-L16).

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -163,198 +122,145 @@ const TooltipDecorator = (ComposedComponent) => {
...
 
-    /**
-     * Calculates position for tooltip, first result cached.
-     *
-     * @method calculatePosition
-     * @param {Object} tooltip
-     * @param {Object} target
-     * @return {Object} shifts calculated
-     */
     calculatePosition = (tooltip, target) => {
```
kyanmichael: To me, these functions – `calculatePosition()` and `positionTooltip()` – seem unnecessarily complicated.

In my [`FE-1847-tooltip-positioning`](https://github.com/Sage/carbon/commits/FE-1847-tooltip-positioning) branch, I simplified / cleaned up those functions and I moved them to a new file – see [`positionTooltip()` in `tooltip-position.config.js`](https://github.com/Sage/carbon/blob/1a0afac/src/components/tooltip/tooltip-position.config.js#L28-L80).

I think it'd be good to simplify / clean up the code in this branch in the same way.

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -163,198 +122,145 @@ const TooltipDecorator = (ComposedComponent) => {
...
-        verticalRight: (targetLeft + pointerDimension + pointerOffset) - tooltipWidth,
-        verticalLeft: targetLeft - (pointerDimension * 0.5),
-        rightHorizontal: targetRight + (0.5 * pointerDimension),
-        leftHorizontal: targetLeft - (pointerDimension * 0.5) - tooltipWidth,
-        sideTop: targetTop - pointerOffset,
-        sideBottom: (targetTop - tooltipHeight) + targetHeight + pointerOffset,
-        sideCenter: (targetTop + (targetHeight * 0.5)) - (tooltipHeight * 0.5)
+          targetRight = targetRect.right,
+          targetXCenter = targetWidth / 2,
+          targetYCenter = targetHeight / 2;
```
kyanmichael: The names `targetXCenter` and `targetYCenter` sound like coordinates, but they're not. Maybe renaming them to `targetHalfWidth` and `targetHalfHeight` would make their purpose clearer?

src/components/tooltip/tooltip.utils.js
```diff
@@ -0,0 +1,35 @@
+const oppositeDirections = {
+  left: 'right',
+  right: 'left',
+  top: 'bottom',
+  bottom: 'top'
+};
+
+function hasTheSameLine(tooltipPosition, pointerAlign) {
```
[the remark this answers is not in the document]
grabkowski: `direction` will suggest that they have the same direction, but we will have to check if they reside on the same line, e.g.: top/bottom or left/right

src/components/validations/validation-icon.component.js
```diff
@@ -1,31 +1,50 @@
...
-  </InputPresentationContext.Consumer>
-);
+const ValidationIcon = ({ theme, type, tooltipMessage }) => {
+  let tooltipPositionProps = {
+    tooltipPosition: 'right',
+    tooltipAlign: 'center'
+  };
+
+  if (isClassic(theme)) {
+    tooltipPositionProps = {};
```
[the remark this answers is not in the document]
grabkowski: we're just overriding the default values here

src/components/validations/with-validation.hoc.js
```diff
@@ -251,9 +251,9 @@ const withValidation = (WrappedComponent) => {
           { ...this.props }
           onBlur={ this.handleBlur }
           onChange={ this.handleChange }
+          { ...this.getValidationIconProps() }
         >
           { this.props.children }
-          { this.renderValidationMarkup() }
```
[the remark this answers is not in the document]
grabkowski: removed

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
grabkowski: Good poiunt, I have implemented a solution directly in the `styleElement` method

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
grabkowski: I have moved `calculatePosition` into a separate file and simplified `positionTooltip` method, unfortunately I had to add more logic, based on new requirements, so these methods could not be simplified in the same way.

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
grabkowski: good point, changed

src/components/tooltip/tooltip.utils.js
```diff
@@ -0,0 +1,35 @@
...
+  const isPositionVertical = isVertical(tooltipPosition);
+  const isAlignHorizontal = isHorizontal(pointerAlign);
+  const isAlignVertical = isVertical(pointerAlign);
```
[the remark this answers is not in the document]
grabkowski: comment added

src/components/tooltip/tooltip.spec.js
grabkowski: I agree, but refactoring of these tests was not in the scope of the task, so I have opted to just update them to test new functionalities.

edleeks87: Yeah sorry I've read the ticket and agree it's out of scope

src/utils/helpers/extract-props/extract-props.js
```diff
@@ -0,0 +1,10 @@
+export default function extractProps(props, reactComponent) {
+  const expectedKeys = Object.keys(reactComponent.propTypes);
```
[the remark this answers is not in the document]
grabkowski: changed to safeProps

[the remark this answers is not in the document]
kyanmichael: This solution of defining `safeProps` on relevant components looks good, thanks :white_check_mark:

src/components/validations/with-validation.hoc.js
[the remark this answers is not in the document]
kyanmichael: :white_check_mark:

src/components/tooltip/tooltip.utils.js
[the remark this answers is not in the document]
kyanmichael: Great, thanks :white_check_mark:

src/components/validations/validation-icon.component.js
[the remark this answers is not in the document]
kyanmichael: Many thanks for refactoring the code and adding that comment :white_check_mark:

src/components/tooltip/tooltip.style.js
```diff
@@ -46,152 +46,34 @@ const fadeIn = keyframes`
...
+  }
 
-function isVerticallyAligned(align) {
-  return ['top', 'bottom', 'center'].includes(align);
+  return textAlignment;
 }
 
-function isHorizontallyAligned(align) {
-  return ['left', 'right', 'center'].includes(align);
+function isHorizontal(property) {
```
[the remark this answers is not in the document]
kyanmichael: @grabkowski: Thanks :white_check_mark:

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
kyanmichael: Thanks :white_check_mark:

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
kyanmichael: OK. Thanks for moving `calculatePosition()` to the new file `src/utils/decorators/tooltip-decorator/calculate-position.js` :white_check_mark:

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
kyanmichael: Thanks for removing `_memoizedShifts`, and adding the optimisation in `styleElement()` in `src/utils/ether/ether.js` :white_check_mark:

src/components/tooltip/tooltip.utils.js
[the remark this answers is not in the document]
kyanmichael: Ah OK. So maybe `haveTheSameOrientation()` would be a better name? Since the function will return `true` for all these pairs of inputs:
```
position: top,    align: top

position: top,    align: bottom

position: bottom, align: top

position: bottom, align: bottom

position: left,   align: left

position: left,   align: right

position: right,  align: left

position: right,  align: right
```
For example, with `position: top, align: bottom`, they both have the same orientation (vertical).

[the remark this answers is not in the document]
grabkowski: I was looking for a proper word, looks like orientation will be fine :)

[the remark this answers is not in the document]
kyanmichael: Many thanks for renaming it :slightly_smiling_face:

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -206,6 +194,28 @@ const TooltipDecorator = (ComposedComponent) => {
       styleElement(tooltip, 'left', `${leftPosition}px`);
     };
 
+    isTooltipOffScreenRight = (position, alignment, tooltipWidth, shifts) => {
+      let pointerX;
+
+      if (position === 'right') {
+        pointerX = shifts.tooltipDistances.right;
+      } else {
```
kyanmichael: Don't we need a condition here like this?
```js
} else if (position === 'left') {
  return false;
} else {
```
Otherwise, when the position is `'left'`, then `pointerX = shifts.vertical[alignment]` may evaluate to `undefined`.

(When the position is `'left'`, then `alignment` will be one of `'top'`,`'center'`,`'bottom'`, but `shifts.vertical['top']` and `shifts.vertical['bottom']` will evaluate to `undefined`.)

grabkowski: and the return value will be evaluated as false, so that condition is not required here

src/components/validations/validation-icon.style.js
```diff
@@ -7,7 +7,7 @@ const ValidationIconStyle = styled.div`
   cursor: pointer;
   display: flex;
   align-items: center;
-  margin-left: 5px;
+  padding-left: 5px;
 
   ${({ theme }) => {
     if (theme.name === THEMES.classic) {
```
edleeks87: You could add the `isClassic` util here

grabkowski: changed

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -187,25 +172,68 @@ const TooltipDecorator = (ComposedComponent) => {
...
+
+      if (position === 'right') {
+        this.setState({ tooltipPosition: 'top', tooltipAlign: 'right' });
+      } else {
+        this.setState({ tooltipAlign: 'right' });
+      }
+    }
+
+    getTooltipDistanceOffset = (position) => {
+      if (!this.props.isThemeModern || !this.props.hasValidationTooltip || !this.props.size) return 0;
```
damianOfKyan: isn't `size` a default property? perhaps checking for it here is redundant?

grabkowski: it depends on the decorated component. Users could use a component that have no size property.

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
kyanmichael: Looking at this code in `isTooltipOffScreenRight()`:

```js
  pointerX = shifts.vertical[alignment];
}

return window.innerWidth < pointerX + tooltipWidth;
```

When `shifts.vertical['top']` or `shifts.vertical['bottom']` evaluates to `undefined`, then `pointerX` will have the value `undefined`.

`tooltipWidth` is a `number`, so `pointerX + tooltipWidth` will be `undefined + number`, which evaluates to [`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN).

`window.innerWidth < pointerX + tooltipWidth` will then be `number < NaN`.

This does evaluate to `false` for the return value, but I don't think it's good practice to be intentionally allowing `NaN` to be used in this confusing way :warning:

By adding this condition:

```js
} else if (position === 'left') {
  return false;
} else {
```

we avoid the use of `NaN`, and it makes it very clear for future developers about the behaviour of the code.

kyanmichael:
I've just unresolved [this conversation](https://github.com/Sage/carbon/pull/2224#discussion_r312092927) about `isTooltipOffScreenRight()` in `src/utils/decorators/tooltip-decorator/tooltip-decorator.js`. @grabkowski: Please could you take another look?

src/utils/decorators/tooltip-decorator/tooltip-decorator.js
grabkowski: sounds reasonable, added a separate condition for the left position.

kyanmichael: Many thanks for making this change in ea0c727 :white_check_mark:

harpalsingh:
The Icon size is too big, Icons should for the most part use the size the font provides, which is 16px, so the 20px override should be removed in this case.

**Currently:**
Screenshot 2019-08-09 at 13 07 30

**Should be:**
This also means moving from the right to 8px.
Screenshot 2019-08-09 at 13 09 06

harpalsingh:
Error tooltip in DLS themes should have no borders and shadow or border radius

Currently:
Screenshot 2019-08-09 at 13 13 43

PR #2225, opened 2019-07-29
FE-1318 form layout

edleeks87:
# Description

Updates Form to use styled-components, implements new style properties defined in DLS. Replaces Save and Cancel Button with a generic Form Button and updates the tests to match the new implementation and fulfil coverage requirements.


src/components/form/form-summary/form-summary.style.js
```diff
@@ -0,0 +1,49 @@
...
+export const StyledInternalSummary = styled.span`
+  padding: 0 3px;
+
+  carbon-form-summary__icon {
+    padding: 0 3px;
+    position: relative;
+    top: -2px;
+  }
+  ${({ type }) => type === 'warning' && css`color: #FF7D00;`}
+  ${({ type }) => type === 'error' && css`color: #D63F40;`}
```
grabkowski: why are these colors hardcoded?

src/components/form/form.style.js
```diff
@@ -0,0 +1,101 @@
...
+  `}
+`;
+
+const StyledForm = styled.form`
+  && ${StyledButton} {
+    margin-left: 16px;
+    ${({ theme }) => !isClassic(theme) && css`margin-left: 15px;`}
+  }
+
+  ${({ theme }) => !isClassic(theme) && css`
```
grabkowski: there is the `isClassic` helper to check for the `classic` theme

src/components/form/form-summary/form-summary.style.js
edleeks87: I'm waiting for some direction from UX, but if I'll try playing with the palette object and get as close as possible

src/components/form/form.style.js
edleeks87: Yeah that's what I'm using

src/components/form/form-button/form-button.spec.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+  });
+
+  describe.each(['save', 'cancel'])(
+    'the primary action buttons',
+    (name) => {
+      let id;
+
+      beforeEach(() => {
+        id = name === 'save' ? 'saveButtonProps' : 'cancelButtonProps';
```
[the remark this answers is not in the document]
edleeks87: Agreed it should be rewritten in a more readable way 👍

src/__experimental__/components/form/form.component.js
```diff
@@ -340,42 +341,29 @@ class FormWithoutValidations extends React.Component {
...
 
-  /** Main class getter */
-  get mainClasses() {
-    return classNames(
-      'carbon-form',
-      this.props.className, {
-        'carbon-form--sticky-footer': this.state.stickyFooter
-      }
-    );
-  }
```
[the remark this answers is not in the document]
edleeks87: Classes can still be passed into the Form, the `htmlProps` functions will still apply any `className` props applied. I will add a test case to ensure that this functions as it previously did.

src/components/form/form-button/form-button.component.js
```diff
@@ -0,0 +1,61 @@
...
+    ...propsForButton(props)[formButtonName],
+    'data-element': props.formButtonName,
+    ...props
+  };
+};
+
+const FormButton = (props) => {
+  return (
+    <StyledFormButtonWrapper { ...tagComponent(props.formButtonName, props) }>
+      <Button { ...addButtonProps(props) }>
```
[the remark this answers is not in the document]
edleeks87: I've removed this file all together as the styling can be applied in `form.style`

src/components/form/form-button/form-button.component.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import I18n from 'i18n-js';
+import PropTypes from 'prop-types';
+import tagComponent from '../../../utils/helpers/tags/tags';
+import Button from '../../button';
+
+const text = (props) => {
+  const { formButtonName } = props;
+  const defaultValue = formButtonName.charAt(0).toUpperCase() + formButtonName.slice(1);
```
kyanmichael: I think Lodash's [`startCase()`](https://lodash.com/docs#startCase) function could be used here:

```js
import { startCase } from 'lodash';

const defaultValue = startCase(formButtonName);
```

src/components/form/form-button/form-button.component.js
```diff
@@ -0,0 +1,60 @@
...
+    <div { ...tagComponent(props.formButtonName, props) }>
+      <Button { ...addButtonProps(props) }>
+        { text(props) }
+      </Button>
+    </div>
+  );
+};
+
+FormButton.propTypes = {
+  formButtonName: PropTypes.string,
```
kyanmichael: Maybe this could be strengthened like this?
```js
formButtonName: PropTypes.oneOf(['save', 'cancel'])
```

src/components/form/form-button/form-button.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        button = wrapper.find(Button);
+        const buttonName = name.charAt(0).toUpperCase() + name.slice(1);
+        expect(button.prop('children')).toEqual(buttonName);
+      });
+
+      it('renders custom text when provided', () => {
+        wrapper = renderFormButton({
+          formButtonName: name,
+          [`${name}Text`]: `This is a ${name} button`
+        }, shallow);
```
kyanmichael: I don't think `shallow` is needed here, is it? (Since the `render` parameter of `renderFormButton()` has a default value of `shallow`)

src/components/form/form-button/form-button.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        });
+      });
+
+      it('renders a custom class if provided', () => {
+        expect(wrapper.find('.my-custom-button-class').exists()).toBeTruthy();
+      });
+
+      it('renders the default text', () => {
+        button = wrapper.find(Button);
+        const buttonName = name.charAt(0).toUpperCase() + name.slice(1);
```
kyanmichael: I think Lodash's [`startCase()`](https://lodash.com/docs#startCase) function could be used here:

```js
import { startCase } from 'lodash';

const buttonName = startCase(name);
```

src/components/form/form-summary/form-summary.style.js
```diff
@@ -0,0 +1,65 @@
...
+  ${({ isInvalid, theme }) => isInvalid && css`
+    background-color: ${theme.form.invalid};
+    margin-left: 15px;
+  `}
+`;
+
+export const StyledInternalSummary = styled.span`
+  padding: 0 3px;
+
+  carbon-form-summary__icon {
```
kyanmichael: Since this is being used as a CSS class in `src/components/form/form-summary/form-summary.component.js`:
```jsx
<Icon className='carbon-form-summary__icon' type={ key } />
```
shouldn't its name start with a `.` here? (`.carbon-form-summary__icon`)

But shouldn't we be using StyledComponents instead of class names?

src/components/form/form-button/form-button.component.js
edleeks87: Agreed I've added them to the `OptionsHelper`

src/components/form/form-button/form-button.spec.js
edleeks87: Yep good spot

src/components/form/form-button/form-button.spec.js
edleeks87: I would normally use that but we are trying to limit the use `lodash` with the idea of removing it as a dependency altogether at some point

src/components/form/form-summary/form-summary.style.js
edleeks87: Eagle eyes, great spot!

src/components/form/form-button/form-button.spec.js
edleeks87: See above

src/components/form/form-button/form-button.component.js
edleeks87: I would normally use that but we are trying to limit the use lodash with the idea of removing it as a dependency altogether at some point

src/__experimental__/components/form/form.component.js
```diff
@@ -390,22 +378,30 @@ class FormWithoutValidations extends React.Component {
...
+        childOfForm: true,
+        addInputToFormState: this.addInputDataToState,
+        labelAlign: isLabelRightAligned ? 'right' : 'left'
+      });
     });
   }
 
   /** Renders the component. */
   render() {
+    const stickyFooter = !this.props.stickyFooter ? false : this.state.stickyFooter;
```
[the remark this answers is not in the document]
edleeks87: Good call

src/components/form/form-summary/form-summary.style.js
edleeks87: After discussing with UX I have implemented the colors as base error and warning already defined in the theme object

src/components/form/form.component.js
```diff
@@ -0,0 +1,511 @@
...
+   */
+  serialize = (opts) => {
+    return Serialize(this._form, opts);
+  }
+
+ /** Separates and returns HTML specific props */
+ htmlProps = () => {
+   const { onSubmit, fixedBottom, ...props } = validProps(this);
+   return props;
+ }
```
kyanmichael: This `htmlProps()` block of code is only indented by 1 space (should be 2 spaces).

The same indentation issue exists for `cancelButton()`, `saveButtonWithSummary()`, `formFooter()`, `orderFormButtons()`, `renderChildren()` and `render()`.

(I'm surprised this wasn't picked up by [ESLint's `indent` rule](https://eslint.org/docs/rules/indent) :thinking:)

src/components/form/form.component.js
```diff
@@ -0,0 +1,511 @@
...
+        saving={ this.props.isValidating || this.props.saving || this.state.submitted }
+      />
+    );
+  }
+
+  /** Returns a custom save button if passed in the default if not */
+  saveButton = () => {
+    if (!this.props.save) { return null; }
+
+    return this.props.customSaveButton ? this.props.customSaveButton : this.defaultSaveButton();
```
kyanmichael: I think this can be simplified to:
```js
return this.props.customSaveButton || this.defaultSaveButton();
```

src/components/form/form.component.js
```diff
@@ -0,0 +1,511 @@
...
+     </FormSummary>
+   );
+ }
+
+ /** Returns the footer for the form  */
+ formFooter = () => {
+   let padding = this.props.stickyFooterPadding;
+
+   if (padding && !padding.match(/px$/)) {
+     padding = `${padding}px`;
```
kyanmichael: I think this can be simplified to:
```js
padding += 'px';
```

src/__experimental__/components/form/form.component.js
```diff
@@ -390,22 +378,33 @@ class FormWithoutValidations extends React.Component {
...
     const childrenArray = Array.isArray(children) ? children : [children];
 
-    return childrenArray.map((child) => {
-      return React.cloneElement(child, { ...child.props, addInputToFormState: this.addInputDataToState });
+    return childrenArray.map((child, index) => {
+      if (typeof child.type !== 'function') return child;
+
+      return React.cloneElement(child, {
+        ...child.props,
+        key: String(index),
```
kyanmichael: It's generally [not recommended](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) to use an array index as a React key.

src/components/dialog/dialog.component.js
```diff
@@ -180,6 +180,24 @@ class Dialog extends Modal {
...
+    const { children } = this.props;
+
+    if (!children) return null;
+
+    if (typeof children !== 'object') return children;
+
+    const childrenArray = Array.isArray(children) ? children : [children];
+    return childrenArray.map((child, index) => {
+      return React.cloneElement(child, {
+        key: String(index),
```
kyanmichael: It's generally [not recommended](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) to use an array index as a React key.

src/components/form/form.component.js
```diff
@@ -0,0 +1,511 @@
...
+   if (!children) return null;
+
+   const childrenArray = Array.isArray(children) ? children : [children];
+
+   return childrenArray.map((child, index) => {
+     if (typeof child.type !== 'function') return child;
+
+     return React.cloneElement(child, {
+       ...child.props,
+       key: String(index),
```
kyanmichael: It's generally [not recommended](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) to use an array index as a React key.

src/components/form/form.component.js
```diff
@@ -0,0 +1,511 @@
...
+  cancel: true,
+  unsavedWarning: true,
+  save: true,
+  saving: false,
+  validateOnMount: false,
+  customSaveButton: null,
+  showSummary: true
+};
+
+FormWithoutValidations.childContextTypes = {
```
kyanmichael: `childContextTypes` and `contextTypes` are part of [React's legacy context API](https://reactjs.org/docs/legacy-context.html). Is it OK to be using them here?

src/components/form/form-button/form-button.component.js
kyanmichael: Understood :thumbsup: Maybe at some point we could create our own `toTitleCase()` helper function in `src/utils/helpers/text/text.js` ? Something like [this](https://stackoverflow.com/a/196991) maybe?
```js
function toTitleCase(str) {
  return str.replace(
    /\w\S*/g,
    function (txt) {
      return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
    }
  );
}
```

edleeks87: Yeah I like that idea, I'll add it in this PR

src/components/form/form-summary/form-summary.style.js
```diff
@@ -0,0 +1,65 @@
...
+    position: relative;
+    top: -2px;
+  }
+  ${({ type, theme }) => type === 'warning' && css`color: ${theme.form.summaryWarning};`}
+  ${({ type, theme }) => type === 'error' && css`color: ${theme.form.summaryError};`}
+`;
+
+export const StyledSummaryText = styled.span`
+  padding: 0 3px;
+  ${({ type }) => type === 'warning' && css`color: #c33e00;`}
```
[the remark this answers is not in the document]
edleeks87: Ah, missed that one somehow, will make the change

src/components/form/form.component.js
```diff
@@ -0,0 +1,507 @@
...
+    if (this.props.validateOnMount) {
+      this.props.validate();
+    }
+
+    if (this.props.unsavedWarning) {
+      this.addUnsavedWarningListener();
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
edleeks87: Yeah agreed I'll replace it

src/components/form/form.component.js
```diff
@@ -0,0 +1,507 @@
...
+    /** Tracks if the form is clean or dirty, used by unsavedWarning */
+    isDirty: false,
+
+    /** Tracks if the saveButton should be disabled */
+    submitted: false
+
+  }
+
+  /**  Returns form object to child components. */
+  getChildContext = () => {
```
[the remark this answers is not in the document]
edleeks87: I'll update it to use the new API

src/components/form/form.component.js
edleeks87: Yeah that is a strange one, I'll correct the indentation

src/__experimental__/components/form/form.component.js
edleeks87: Yeah this was to get see if it would remove the warnings, I'll use guids instead

src/components/form/form.component.js
```diff
@@ -0,0 +1,507 @@
...
+  /** enables a form which has been disabled after being submitted. */
+  enableForm = () => {
+    this.setState({ submitted: false });
+  }
+
+  /**
+   * Serializes the inputs in the form ready for submission via AJAX
+   * https://www.npmjs.com/package/form-serialize
+   */
+  serialize = (opts) => {
```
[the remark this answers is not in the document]
edleeks87: Yeah I was thinking the same thing, I'm going to leave it in the old Form as it may break some legacy stuff removing it. But I'll remove it form the experimental implementation

src/components/form/form-button/form-button.component.js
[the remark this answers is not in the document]
grabkowski: my question was do we need that wrapper at all?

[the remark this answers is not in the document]
edleeks87: Yeah I agree sorry should have been clearer I have deleted the wrapper and `form-button.style`

src/__experimental__/components/form/form.component.js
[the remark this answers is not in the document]
kyanmichael: Thanks :white_check_mark:

src/components/form/form-button/form-button.component.js
kyanmichael: Many thanks for creating `Text.titleCase()` :white_check_mark:

src/components/form/form-button/form-button.component.js
kyanmichael: Thanks :white_check_mark:

src/components/form/form-summary/form-summary.style.js
kyanmichael: :white_check_mark:

src/components/form/form-summary/form-summary.style.js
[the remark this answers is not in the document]
kyanmichael: Thanks :white_check_mark:

src/components/form/form.component.js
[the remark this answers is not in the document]
kyanmichael: Thanks :white_check_mark:

src/components/form/form.component.js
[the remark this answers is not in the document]
kyanmichael: Thanks :white_check_mark:

src/components/form/form.component.js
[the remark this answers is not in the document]
kyanmichael: OK, sounds good, thanks :white_check_mark:

src/components/form/form.component.js
kyanmichael: Thanks for fixing the indentation on those functions :white_check_mark:

Looking at the `.eslintrc` file, it extends `./node_modules/carbon-factory/.eslintrc`, which does configure the `indent` rule:
```
"indent": ["error", 2, {
  "VariableDeclarator": 2,
  "SwitchCase": 1
}],
```
So I wonder why it's not working here? :thinking: I wonder where else it's not working? :worried:

src/components/form/form.component.js
kyanmichael: :white_check_mark:

src/__experimental__/components/form/form.component.js
kyanmichael: Thanks :white_check_mark:

src/components/dialog/dialog.component.js
kyanmichael: :white_check_mark:

src/components/form/form.component.js
kyanmichael: :white_check_mark:

src/components/form/form.component.js
kyanmichael: :white_check_mark:

src/components/form/form.component.js
edleeks87: Yeah it will need some digging I suspect. In this case it seemed to be relating to how some preceding functions were aligned as it complained when i corrected the indentation until all the erroneous functions were aligned correctly 🤕

src/__experimental__/components/form/form.stories.js
```diff
@@ -7,6 +7,15 @@ import notes from './documentation/notes.md';
 import Info from './documentation/Info';
 import Form, { FormWithoutValidations } from '.';
 import Textbox from '../textbox';
+import Button from '../../../components/button';
+import Link from '../../../components/link';
+
+const additionalFormActions = (innerText) => {
+  return {
+    Button: <Button>{ innerText }</Button>,
+    Link: <Link href='./?path=/story/experimental-form--default'>{ innerText }</Link>
```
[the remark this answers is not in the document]
kyanmichael: @edleeks87: Many thanks for creating FE-1869 to address this :thumbsup:

src/components/form/form.component.js
kyanmichael: @edleeks87: Many thanks for creating FE-1870 to investigate this :thumbsup:

src/components/form/form.style.js
```diff
@@ -100,4 +103,32 @@ const StyledForm = styled.form`
...
+  theme: PropTypes.object,
+  stickyFooter: PropTypes.bool,
+  fixedBottom: PropTypes.bool
+};
+
+StyledFormFooter.defaultProps = {
+  theme: baseTheme
+};
+
+StyledFormFooter.propTypes = {
```
kyanmichael: It looks like the `StyledFormFooter` StyledComponent also supports these props: `stickyFooter`, `buttonAlign`. Should we add `propTypes` here for them?

edleeks87: Yep good spot I've corrected that now

kyanmichael: Thanks :white_check_mark:

lukeemmerson:
passed ux review after at desk with ed

src/__experimental__/components/form/form.component.js
```diff
@@ -128,16 +134,28 @@ class FormWithoutValidations extends React.Component {
...
 
   removeUnsavedWarningListener = () => {
+    this.unsavedListener = false;
     this._window.removeEventListener('beforeunload', this.checkIsFormDirty);
   }
 
+  /** Returns true if any input has value */
+  checkFormDataExists() {
+    const { formInputs } = this.state;
+    return Object.keys(formInputs).map(id => (!!formInputs[id].length)).includes(true);
```
kyanmichael: Just a minor comment: I think this can be simplified using [`Array.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some). Also, ESLint's [`no-implicit-coercion`](https://eslint.org/docs/rules/no-implicit-coercion) rule recommends using `Boolean(foo)` instead of `!!foo` (I think it'd be good to enable that rule).
```js
return Object.keys(formInputs).some(id => Boolean(formInputs[id].length));
```

kyanmichael: I think there's a potential risk of `Object.keys(formInputs)` throwing a `TypeError: Cannot convert undefined or null to object` if `this.state.formInputs` hasn't been initialised :warning:

I realise that currently there is a safety check of `this.state.formInputs` in `checkIsFormDirty()`:
```js
if (this.state.formInputs) {
  this.setState({ isDirty: this.checkFormDataExists() });
}
```
But to be completely safe, I think it'd be good to add a safety check here. Something like this:
```js
checkFormDataExists() {
  const { formInputs } = this.state;
  if (!formInputs) {
    return false;
  }
  …
```

src/components/form/form.component.js
```diff
@@ -12,9 +12,15 @@ import { withValidations } from '../validations';
 import ElementResize from '../../utils/helpers/element-resize/element-resize';
 import StyledForm, { StyledFormFooter, StyledAdditionalFormAction } from './form.style';
 
-const FormContext = React.createContext();
```
kyanmichael: I guess there's a reason why we need to revert to React's legacy context API here?

edleeks87: Yeah sorry I mentioned it in a message offline, it breaks a lot of the I had to revert some of the legacy context functions (`getChildContext` etc) in the legacy Form as they break a lot of the input decorators and the work to fix them is out of scope and unnecessary as they are soon to be deprecated by us anyway. The new/ experimental Form has the updated context API stuff

src/__experimental__/components/form/form.component.js
edleeks87: Yeah I was 50/50 about including the extra guard, happy to add it in

edleeks87: No problem I will make the change and I'll add the rule update to the es-lint ticket raised the other day

src/components/form/form.component.js
kyanmichael: @edleeks87: Thanks for refactoring this :white_check_mark:

src/__experimental__/components/form/form.component.js
kyanmichael: Many thanks for refactoring this and updating FE-1870 :white_check_mark:

kyanmichael: Many thanks :white_check_mark:

src/components/form/form.component.js
kyanmichael: Ah OK, thanks for clarifying about that :thumbsup:

PR #2235, opened 2019-08-02
FE-1868 fix prop tables

ianforster-sage:
# Description
Use the `react-docgen` CLI to generate a `__docgenInfo` object for each component and append this to the component in each story.js file. In some cases, such as components with input decorators, or those that extend other components, it will be necessary to build a simple component with all the PropTypes in one file for `react-docgen` to parse. These are not included in this PR as they were for dev purposes only.

# Background
Storybook `addon-info` module is failing to generate the Component property required to build the Prop Tables completely. The issue is caused by a bug in `babel-plugin-react-docgen`. If Component.__docgenInfo is 'undefined' @storybook/addon-info/src/components/makeTableComponent.js uses defaults when it can't find the actual values. Addon-Info is not being actively developed any more and is scheduled to be replaced by Addon Docs, which may fix the issue but it's still in Beta, with known bugs and isn't stable enough to use yet 
Tried a rollback of addon-info to 4.x but this fails because it uses an old version of core-js. 
Considered forking babel-plugin-react-docgen but after discussion within the team we decided that forking and maintaining our own version of a repo that is deprecated is not a path we want to go down.
So the option to pursue is to use the `react-docgen` CLI to generate a `__docgenInfo` object for each component and append this to the component in each story.js file.


src/components/dropdown/docgenInfo.json
```diff
@@ -0,0 +1,196 @@
+{
+  "src/components/dropdown/dropdown.dgi.js": [
```
chrisbarber86: this path doesn't look right?

src/components/validations/docgenInfo.json
```diff
@@ -0,0 +1,41 @@
+{
+  "src/components/validations/form-with-validations.spec.js": [
```
chrisbarber86: Should we be scoping out the `spec.js` files from the regex?

PR #2237, opened 2019-08-05
added autoFocus option to experimental input components

blaky:
# Description
New functionality to automatically focus experimental input fields when they get rendered. We need this capability when rendering new forms in dialogs/modals and we want the user to be able to continue without touching the mouse.

I also fixed a small bug on the storybook preview, a semi-column from the code was visible on the rendered page, please see the screenshot below.

# TODO
- [x] Release notes

# Screenshots / Gifs
image

# Testing Instructions
You can test the new functionality by flipping the `autoFocus` storybook property under `Knobs` or by setting a default value in the `*.story.js` files.
Furthermore to allow for testing on page-load added new story to relative components called `autoFocus`. 

Screenshot 2019-12-23 at 09 49 03


change_log/next/auto-focus-on-experimental-fields.yml
```diff
@@ -0,0 +1,2 @@
+New Features: "Adds `autoFocus` property to experimental input components"
+Bug Fixes: "Fixes a bug in storybook: visible semi-column removed from storybook preview page"
```
edleeks87: Hey @blaky assuming `semi-column` is a typoI believe this bug was fixed and merged into master the other day as part of this commit https://github.com/Sage/carbon/pull/2225/commits/b4ba1da772eb1a91ecbdf9c0131b5c3fbf082db0? If so this line in the change_log isn't needed

src/__experimental__/components/number/number.stories.js
```diff
@@ -29,6 +29,7 @@ storiesOf('Experimental/Number Input', module)
     const onChangeDeferredEnabled = boolean('Enable "onChangeDeferred" Action', false);
     const onKeyDownEnabled = boolean('Enable "onKeyDown" Action', false);
     const deferTimeout = onChangeDeferredEnabled ? number('deferTimeout') : undefined;
+    const autoFocus = boolean('autoFocus', false);
```
grabkowski: I think that's already imported in the `Textbox` story props

blaky:
@edleeks87 @grabkowski I fixed the issues that you highlighted but I don't have the rights to assign the PR back to you.

edleeks87:
@blaky no problem I will get one of our QAs to test it when @grabkowski approves 😄

src/utils/helpers/auto-focus/auto-focus.js
```diff
@@ -0,0 +1,24 @@
...
+      autoFocus: false
+    };
+  },
+
+  getKey: (autoFocus, { ...previous }) => {
+    if (autoFocus && previous && previous.autoFocus === autoFocus) {
+      return previous.key;
+    }
+
+    return guid();
```
kyanmichael: Since the second parameter of `AutoFocus.getKey()` isn't supplied in any call to this function in this PR, it looks like this function will always return a new unique ID from `guid()` every time it's called. (The first parameter `autoFocus` will have no effect if the second parameter isn't supplied.)

I'm not sure whether this is correct?

I know @jamime introduced similar functionality in #2540 to force the `Decimal` component to re-render when the `allowEmptyValue` Storybook knob is toggled.

Is the idea here to force the component to re-render when the `autoFocus` Storybook knob is toggled? (So that if `autoFocus` is now `true`, then the component will auto-focus when it's re-rendered?)

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -38,47 +39,52 @@ const previous = {
...
-    const allowEmptyValue = boolean('allowEmptyValue', false);
-
-    // When the allowEmptyValue knob changes we want to force the component to re-create
-    // allowEmptyValue is only used in the constructor and it is not currently supported to change during the lifetime
-    // of the component
-    if (previous.allowEmptyValue !== allowEmptyValue) {
-      previous.key = guid();
-    }
-    previous.allowEmptyValue = allowEmptyValue;
-    const { key } = previous;
```
kyanmichael: Is it OK to remove this code which @jamime added in #2540 ?

src/utils/helpers/auto-focus/auto-focus.js
```diff
@@ -0,0 +1,24 @@
...
+      return { key, autoFocus };
+    }
+
+    return {
+      key: guid(),
+      autoFocus: false
+    };
+  },
+
+  getKey: (autoFocus, { ...previous }) => {
```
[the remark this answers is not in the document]
damianOfKyan: it is in specs for auto-focus.

src/utils/helpers/auto-focus/auto-focus.js
damianOfKyan: Yes, toggle on knob should force re-render.

src/__experimental__/components/decimal/decimal.stories.js
damianOfKyan: well spotted @kyanmichael

src/utils/helpers/auto-focus/auto-focus.js
```diff
@@ -0,0 +1,24 @@
+import guid from '../guid';
+
+const AutoFocus = {
+  default: (key, autoFocus) => {
```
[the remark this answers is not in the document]
damianOfKyan: indeed, I now removed it since it is not used

change_log/next/auto-focus-on-experimental-fields.yml
```diff
@@ -0,0 +1 @@
+New Features: "Adds `autoFocus` property to experimental input components. Adds `autoFocus` story to relevant experimental components."
```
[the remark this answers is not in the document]
damianOfKyan: added components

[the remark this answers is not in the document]
kyanmichael: Thanks for adding that :white_check_mark:

src/utils/helpers/auto-focus/auto-focus.js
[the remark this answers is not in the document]
kyanmichael: Thanks for removing it :white_check_mark:

src/__experimental__/components/decimal/decimal.stories.js
kyanmichael: Thanks for restoring that code :white_check_mark:

src/utils/helpers/auto-focus/auto-focus.js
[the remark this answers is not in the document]
kyanmichael: Yes I realise that the second parameter is supplied in these specs:

* https://github.com/blaky/carbon/blob/0cb3002/src/utils/helpers/auto-focus/__spec__.js#L18
* https://github.com/blaky/carbon/blob/0cb3002/src/utils/helpers/auto-focus/__spec__.js#L22

But the specs don't actually affect the runtime behaviour/functionality of Carbon in production.

In all the places where `AutoFocus.getKey()` is called at runtime within the Carbon library, the second parameter is never passed, so it would have no effect on Carbon's runtime behaviour in production:

* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/checkbox/checkbox.stories.js#L91
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/select/select.stories.js#L39
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/textarea/textarea.stories.js#L62
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/textbox/textbox.stories.js#L150

So is that second parameter implemented for potential future use? (In other words: at some point in the future, maybe a developer might wish to supply that second parameter when writing some new code?)

src/utils/helpers/auto-focus/auto-focus.js
kyanmichael: OK. In that case, since the second parameter `{ ...previous }` is currently never supplied in any runtime/production code within the Carbon library, then this condition is currently always `false`:
```js
if (autoFocus && previous && previous.autoFocus === autoFocus) {
```
and so this `AutoFocus.getKey()` function is always behaving like this in runtime/production code:
```js
getKey: (autoFocus) => {
  return guid();
}
```
I'm guessing that the usage of `AutoFocus.getKey()` needs to be updated in some/all of these places:

* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/checkbox/checkbox.stories.js#L91
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/select/select.stories.js#L39
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/textarea/textarea.stories.js#L62
* https://github.com/blaky/carbon/blob/0cb3002/src/__experimental__/components/textbox/textbox.stories.js#L150

to pass a value for the second parameter, so that this `AutoFocus.getKey()` function can return `previous.key` when appropriate?

damianOfKyan: updated stories for those components.

src/utils/helpers/auto-focus/auto-focus.js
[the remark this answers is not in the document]
damianOfKyan: updated.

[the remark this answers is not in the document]
kyanmichael: Thanks for adding values for the second parameter :white_check_mark: b19a82d

src/utils/helpers/auto-focus/auto-focus.js
kyanmichael: Thanks for adding values for the second parameter :white_check_mark: b19a82d

PR #2238, opened 2019-08-05
FE-1845 responsive form footer

edleeks87:
# Description
Replaces the `AppWrapper` in the form footer with a new responsive wrapper with a smaller `min-width` to allow the buttons in the footer to scale better as the screen is adjusted.

Adds a style property to remove the `left-margin` from the first button in the footer (left most)


src/components/form/form.style.js
```diff
@@ -16,14 +19,49 @@ export const StyledAdditionalFormAction = styled.div`
...
+export const StyledResponsiveFooterWrapper = styled.div`
+  margin: 0 auto;
+  max-width: inherit;
+  min-width: inherit;
+  padding: 0;
+  align-items: center;
+  display: flex;
+  margin-top: 20px;
+  ${({ borderWidth }) => borderWidth && css`
+    border-width: ${borderWidth} !important;
```
kyanmichael: Is there an alternative to using `!important` here? Maybe [this StyledComponents suggestion](https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity) could be used?

> The way to override styles with a high specificity is to simply increase the specificity of your own styles. This could be done using `!important`, but that's error prone and generally not a good idea. We recommend the following technique ... Each `&` gets replaced with the generated class.

edleeks87: Good point I'll use `&&` I wrongly assumed it had to be applied to a styled-component but it clearly works for specific properties as well 👍

src/components/form/form.style.js
```diff
@@ -16,14 +19,49 @@ export const StyledAdditionalFormAction = styled.div`
       display: inline-block;
     `}
 
-    ${StyledButton} {
+    ${applyButtonLeftMargin}
+
```
damianOfKyan: white space.

src/components/form/form.style.js
```diff
@@ -16,14 +19,49 @@ export const StyledAdditionalFormAction = styled.div`
...
+      margin-left: 0px;
       &:first-child {
         margin-left: 0;
       }
     }
   `}
 `;
 
+export const StyledResponsiveFooterWrapper = styled.div`
+  margin: 0 auto;
```
damianOfKyan: you are using margin and margin-top, maybe just use `margin: 20px auto 0 auto`

src/components/form/form.style.js
edleeks87: In fact I  ended up having to use `&&&&` 😱

src/components/form/form.style.js
edleeks87: Yeah I wanted to leave it there so it's more readable

src/components/form/form.style.js
edleeks87: Good point will give it a go thanks!

src/components/form/form.style.js
kyanmichael: Wow, that's quite specific! :sweat_smile: Thanks for changing this :thumbsup:

sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge. Remaining issues with buttons overlapping in IE11 and disappearing off to the left in Chrome when all button/link options are added in the footer will not be fixed as there should never be as many actions in the footer. Further work will be done by UX to design a responsive small footer with many actions.

PR #2239, opened 2019-08-05
FE-1898 redundant knobs

ianforster-sage:
# Description
Fix filter of knobs for classic story

### DIFF
--- a/src/components/pill/pill.stories.js
+++ b/src/components/pill/pill.stories.js
@@ -6,7 +6,6 @@ import { text, select, boolean } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import Pill from './pill.component';
 import OptionsHelper from '../../utils/helpers/options-helper';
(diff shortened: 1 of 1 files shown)


src/components/pill/pill.stories.js
```diff
@@ -6,7 +6,6 @@ import { text, select, boolean } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import Pill from './pill.component';
 import OptionsHelper from '../../utils/helpers/options-helper';
-import { isClassic } from '../../utils/helpers/style-helper';
```
edleeks87: I think it would be better to remove `import { THEMES } from '../../style/themes';`, keeping `isClassic` and then just passing the theme object to the helper function

damianOfKyan: ^ agree with Ed, I should have done it like that first time around, great catch!

PR #2243, opened 2019-08-12
adds passive config prop to table context

edleeks87:
# Description
Adds prop that allows table data to be rendered passively with no hover styling applied to the rows

### DIFF
--- a/src/components/table/__spec__.js
+++ b/src/components/table/__spec__.js
@@ -1019,6 +1019,23 @@ describe('Table', () => {
     });
   });
 
+  describe('isPassive', () => {
+    it('returns false if isPassiveData and highlightable are true and selectable is false or undefined', () => {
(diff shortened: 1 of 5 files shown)


src/components/table/__spec__.js
```diff
@@ -1019,6 +1019,23 @@ describe('Table', () => {
     });
   });
 
+  describe('isPassive', () => {
```
stevebutler2210: Is it worth adding a case covering when both `highlightable` and `selectable` are true, or is that an invalid scenario?

edleeks87: No good point I'lll add for the sake of completeness 👍

sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge. Checked that rows did/did not have a hover state and could be highlighted and/or selected as required when the passive knob was set to true. Testing passed.

PR #2250, opened 2019-08-13
FE-1687 Removes icon from main button in component

edleeks87:
# Description
After discussion with UX, the main button should not have an additional icon other than the dropdown. This PR removes it and prevents one from being added to it in the future.

QA and UX reviews can take place once it is merged into FE-1291-icon


src/components/multi-action-button/multi-action-button.component.js
```diff
@@ -6,20 +6,29 @@ import StyledMultiActionButton from './multi-action-button.style';
 import Button from '../button';
 
 class MultiActionButton extends SplitButton {
+  /**
+   * Returns the props for the main button.
```
dawidzarzycki: We can think about changing multipleline comments into one line comments if it is possible

PR #2260, opened 2019-08-15
FE-1839 Tile DLS component

stevebutler2210:
- adds a new component, Tile
- this component will ultimately supersede the legacy Pod component
- used to display static content, Tile takes an arbitrary number of components and renders them into a simple Tile layout
- Tile has two themes, 'tile' and 'transparent', passed via the 'as' prop, which change the background color
- the Tile itself will take up 100% of the width of the parent element by default, and has an optional percentage-based width prop to manually specify this
- each child element is wrapped by a TileContent wrapper, and has the option of being passed a width prop to specify the percentage width of the Tile for the child to occupy (if unspecified, each child will take up an equal amount of space within the Tile)


stevebutler2210:
**Some examples of different width / orientation settings and the results:**

Default width settings - container fills 100% parent width, children equally spaced:
Screenshot 2019-08-15 at 13 45 30

Tile width set to 50%:
Screenshot 2019-08-15 at 13 45 41

First child width set to 50%:
Screenshot 2019-08-15 at 13 45 52

Orientation set to vertical, Tile width set to 10% of parent:
Screenshot 2019-08-15 at 13 54 39

src/__experimental__/components/tile/documentation/info.js
```diff
@@ -0,0 +1,40 @@
...
+import {
+  StoryHeader, StoryCode, StoryCodeBlock
+} from '../../../../../.storybook/style/storybook-info.styles';
+
+const info = (
+  <div>
+    <p>Tile component</p>
+    <StoryHeader> Implementation</StoryHeader>
+    <StoryCode>
+      {'import Tile from "carbon-react/lib/components/checkbox"'}
```
ianforster-sage: This path looks incorrect.

src/__experimental__/components/tile/tile.component.js
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <StyledTile
+      padding={ padding }
+      tileTheme={ as }
+      orientation={ orientation }
+      width={ width }
+      { ...rest }
+      { ...tagComponent('tile', props) }
+    >
+      {wrappedChildren}
```
ianforster-sage: Did this pass linting? It'll probably complain about spaces inside braces.

src/__experimental__/components/tile/documentation/info.js
stevebutler2210: Good spot, updated 👍

src/__experimental__/components/tile/tile.component.js
stevebutler2210: Yep, passed fine - I've noticed whenever I'm passing a child in curly braces like this it forces the formatting so that there are no spaces. Always thought it was odd, as it's at odds with the rule enforced elsewhere, but it automatically changes it for me if I try and add the spaces, so I've left it.

stevebutler2210:
Comments addressed, and I've also added an index.js for Tile, in line with what we've done for other components.

src/__experimental__/components/tile/tile.component.js
edleeks87: There's a ticket on the backlog to investigate some of the issues we seem to be having with eslint, I'll add this as an example to the ticket. I suspect it's the `"eslint.autoFixOnSave": true` rule in `settings.json` isn't picking up this edge case for some reason

src/__experimental__/components/tile/tile.component.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import tagComponent from '../../../utils/helpers/tags';
+import OptionsHelper from '../../../utils/helpers/options-helper';
+import { StyledTile, TileContent } from './tile.style.js';
+
+const Tile = (props) => {
```
edleeks87: You could destructure the props in as they're passed into the component but I'm probably being nit-picky so feel free to ignore me
```
const Tile = ({
 as,
 children,
 padding,
 orientation,
 width,
 ...props
}) => {
```

src/__experimental__/components/tile/tile.spec.js
```diff
@@ -0,0 +1,185 @@
...
+            },
+            wrapper,
+            { modifier: css`${`${TileContent}:not(:last-of-type)`}` }
+          );
+        });
+
+        it('sets border-left and padding-left for all but the first TileComponent', () => {
+          assertStyleMatch(
+            {
+              borderLeft: 'solid 1px #E5EAEC',
```
edleeks87: It might be better to use the colours from theme object wherever possible and assert that the border matches `solid 1px ${baseTheme.tile.separator}` etc. That way if the colours are ever changed in the theme object we won't have to update all the tests

src/__experimental__/components/tile/tile.style.js
```diff
@@ -0,0 +1,83 @@
...
+    padding: ${paddingSizes[padding]};
+    position: relative;
+    width: ${(width && width !== 0) ? `${width}` : '100'}%;
+
+    ${TileContent} {
+      box-sizing: border-box;
+
+      ${isHorizontal(orientation) && 'display: inline;'}
+
+      ${isVertical(orientation) && 'width: auto'}
```
edleeks87: Missing `;`?

ianforster-sage:
Once other thing Steve, I noticed that this is in `/__experimental__`, why did you decide to put it there?

stevebutler2210:
@ianforster-sage I did that as it was a brand new component, happy to move it though if that's not correct 👍

ianforster-sage:
Not sure it's necessary to put anything new in there now. Let's double check with @chrisbarber86

chrisbarber86:
@stevebutler2210 @ianforster-sage Since this is a brand new component, it can just go straight into the top level folder, doesn't need to go via `__experimental__`

src/__experimental__/components/tile/tile.spec.js
stevebutler2210: I left it like that intentionally, as I figured the test didn't seem very useful without the value being hard-coded, as it would continue to pass regardless?

Wouldn't the point of a test like this be to ensure that if somebody is changing values in the theme, they are fully aware of the places it impacts, so you'd want this to break?

src/__experimental__/components/tile/tile.component.js
stevebutler2210: Ah yeah, I did do it that way originally, but wasn't sure if we needed to be passing _all_ of the props to tagComponent (on line 38) - moved the destructuring in so that I could pass the complete props object there, but can change it if that's not actually required.

src/__experimental__/components/tile/tile.spec.js
edleeks87: Yeah fair point, tbh we could just use find and replace if we even need to update these, was just thinking of maintainability going forward but I'm being unnecessarily pedantic

src/__experimental__/components/tile/tile.component.js
edleeks87: It's not that big a deal but if you pass in the `...props` to tagComponent it would still work as it just looks for the `data-` based props. But either way works so I'm happy to approve it as is 👍

src/components/tile/tile.style.js
```diff
@@ -42,7 +42,11 @@ const StyledTile = styled.div`
 
       ${isHorizontal(orientation) && 'display: inline;'}
 
+<<<<<<< Updated upstream:src/components/tile/tile.style.js
```
DlgSHi: probably forgot to delete this :)

src/components/tile/tile.style.js
```diff
@@ -42,7 +42,11 @@ const StyledTile = styled.div`
...
       ${isVertical(orientation) && 'width: auto'};
+=======
+      ${isVertical(orientation) && 'width: auto;'}
+>>>>>>> Stashed changes:src/components/tile/tile.style.js
```
DlgSHi: probably forgot to delete also this :)

stevebutler2210:
@DlgSHi yep, conflict markers now removed, and I've also added the pixelWidth prop

src/components/tile/index.js
```diff
@@ -0,0 +1,3 @@
+import Tile from './tile.component';
```
dawidzarzycki: I see that we use shorter syntax in index.js files so in this case it should be
`export { default } from './tile.component'`
Can you change it to keep syntax clean please?

stevebutler2210: Done, will wait for QA to finish before requesting re-reviews though, in case any bugs come up.

harpalsingh:
Minor amendment, S = 16px padding and M = 24px padding.

src/components/tile/tile.style.js
```diff
@@ -0,0 +1,86 @@
+import styled, { css } from 'styled-components';
+import PropTypes from 'prop-types';
+import baseTheme from '../../style/themes/base';
+
+const paddingSizes = {
```
dawidzarzycki: Could you move it to the Options-Helper please? :) And then we can import it also in tests (https://github.com/Sage/carbon/pull/2260/files#diff-453b70e8319831a4c9f04c6bed98ac5fR168)

stevebutler2210: The values that the prop can take (`XS`, `S`, `M`, `L`, `XL`) are in the optionsHelper, via `OptionsHelper.sizesTile`. But I don't think the actual resulting padding values don't need to be stored in there as the optionsHelper is just for the values the prop takes?

And I'm using them in a slightly different format in the specs, so I couldn't just flat import the same thing to use in both.

PR #2264, opened 2019-08-19
Add passive functionality to dls table

edleeks87:
# Description
Refactors Table to support passive data option implemented in v8.1.1 using styled-component instead of sass


sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge browsers. When the passive knob was set to true, the table rows had no hover state. When the highlighted and/or selected knobs are selected with passive set to true or false these actions can still be done. The column headers can still be hovered and the content sorted. Data inputs can still be accessed and data entered and edited. Testing passed ✅ There may be need to add another option to make the column headers passive too in a later iteration.

PR #2268, opened 2019-08-21
FE-1915 - fix Configurable Items component in the storybook

dawidzarzycki:
# Description
- fix visibility of component in the storybook
- fix for styling of the component

# Important Information
- The issue with `Accessibility` (`ul` ,`li`) is familiar but we have to omit it because it is caused by DnD library specifications

### DIFF
--- a/change_log/next/FE-1915-configurable-items.yml
+++ b/change_log/next/FE-1915-configurable-items.yml
@@ -1,0 +1,1 @@
+Bug Fixes: "Fixes issue with component not being visible in the storybook"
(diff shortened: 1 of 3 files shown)


change_log/next/FE-1915-configurable-items.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "Fixes issue with component not being visible in the storybook"
```
edleeks87: Can you add that it was the `ConfigurableItems` component please

PR #2252, opened 2019-08-13
FE-1728 Checkbox with validation

joshdj:
# Description
Wrapped the checkbox component with the validation HOC.

Show errors, warnings and info.

PR #2278, opened 2019-08-27
FE-1728 Switch with validation

joshdj:
## Description

Implement validation on switch component.

* Unblock validation using new prop
* Added label colours for warning and info
* Updated spec
* Added validation story

## Screenshots
**Default Theme**
Screenshot 2019-09-12 at 13 04 57

**Classic Theme**
Screenshot 2019-09-10 at 13 24 28

**Label on text input after validation**

Having removed the validation colours from the label component the label will no longer show red text (etc) like what's is doing on the master branch.

Screenshot 2019-09-11 at 16 58 56


harpalsingh:
Was the requirement that clicking the help icon should trigger the switch toggle? While its normal for labels to activate the input, don't think the help icon itself in the validations should do that.

joshdj:
Blocked by changes needed from #2252

src/__experimental__/components/switch/switch.style.js
```diff
@@ -37,7 +38,20 @@ const StyledSwitch = styled.div`
       margin-bottom: 8px;
       margin-right: 100%;
 
-      ${labelWidth && `
+      ${({ hasError, hasWarning, hasInfo }) => css`
+        ${() => hasInfo && css`color: ${theme.colors.info};`}
```
dawidzarzycki: I guess it can be just
```
${hasInfo && css`color: ${theme.colors.info};`}
${hasWarning && css`color: ${theme.colors.warning};`}
${hasError && css`color: ${theme.colors.error};`}
```
We don't need additional abstract layer here

joshdj: This might become redundant. Waiting on PR #2252.

joshdj: Updated in commit 26f9d5007

joshdj:
This PR is no longer blocked.

src/components/validations/with-validation.hoc.js
```diff
@@ -278,5 +280,19 @@ const withValidation = (WrappedComponent) => {
   return WithValidation;
 };
 
-export { validationsPropTypes };
+function getValidationType({ hasError, hasWarning, hasInfo }) {
+  let type = '';
+
+  if (hasError) {
```
jamime: I think this function would befit from multiple returns instead of the `if-else` e.g. 

```
if(hasWarning){
    return 'warning';
}
```

src/__experimental__/components/switch/switch.style.js
```diff
@@ -37,7 +38,20 @@ const StyledSwitch = styled.div`
...
+      ${StyledValidationIcon} {
+        display: inline-block;
+        padding: 1px;
+        margin-top: 0;
+        margin-left: 8px;
+        margin-bottom: 0;
+        vertical-align: middle;
+      }
+
+      ${hasInfo && css`color: ${theme.colors.info};`}
```
jamime: You only need to use `css` if you have a function inside the template literal. It's safe to remove the call to `css` on the next 3 lines too.

https://www.styled-components.com/docs/api#css

src/components/validations/with-validation.hoc.js
joshdj: Updated in commit 02dd7594e

src/__experimental__/components/switch/switch.style.js
joshdj: Updated in commit 02dd7594e

joshdj:
> Hi @joshdj functionality looks good, there are a few minor changes required.
> 
>     * When the icon is a circle it moves up by ~1px when the validation kicks in
> 
>     * WCAG2 AA violation for the warning color `Element has insufficient color contrast of 3.33 (foreground color: #e96400, background color: #ffffff, font size: 10.5pt (14px), font weight: bold). Expected contrast ratio of 4.5:1`
> 
>     * Missing ARIA element `ARIA attribute element ID does not exist on the page: aria-describedby="5076d789-c919-ecc9-0407-bd8b3634c635-help"` to reproduce
>       
>       * Toggle validation on then off
>       * Rerun a11y test
> 
> 
> Thanks,
> Jamie

Thank you Jamie for the review.

Icon moving by 1 pixel has been fixed in commit 02dd7594e.
I've talked to Harpal about the colour violation. Hopeing to catch up on Monday about it.

joshdj:
>     * I tested the changes for aria-describedby hower I was unable to get voiceover to read the help text out using keyboard commands. @joshdj can you take a look?

Fixed in commit f7576b000

joshdj:
Awaiting feedback from Harpel for resolving WCAG2 AA violation for the warning colour.

joshdj:
Similar to the design on the text input (thanks to Harpal's feedback), the label won't have a different colour applied to it when validation fails, etc This resolves the colour violation issue mentioned above.
Resolved in commit fe0650972.

src/components/validations/validation-icon.component.js
```diff
@@ -31,7 +32,13 @@ const ValidationIcon = ({
     <InputPresentationContext.Consumer>
       {
         context => (
-          <ValidationIconStyle type={ type }>
+          <ValidationIconStyle
+            id={ iconId }
+            type='button'
+            data-validation-type={ type }
+            validationType={ type }
+            aria-label='additional validation information'
```
jamime: This should not be hard coded. The aria-label should be the contents of the tooltip, it also needs the role "tooltip".
To see a working a11y tooltip example check out https://van11y.net/downloads/simple-tooltip/demo/index.html using VoiceOver on your Mac.

jamime: In fact, aria-describedby should do the job, but I can't remember if the tooltip has to be visible.
docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute
example:
http://www.oaa-accessibility.org/examplep/tooltip1/

joshdj: @jamime Updated in commit 93fd9b571

joshdj: @jamime I believe I have resolved the issue with the input, where the input should now be linked with the label and not inherit the tooltip aria-label as well.
Commit 4d91dfbb4.

src/__experimental__/components/switch/switch.spec.js
```diff
@@ -171,6 +179,38 @@ describe('Switch', () => {
...
+    });
+
+    describe.each(validationTypes)('validation %s', (type) => {
+      it('check icon type', () => {
+        const propName = `has${text.titleCase(type)}`;
+        wrapper.setProps({
+          [propName]: true
+        });
+
+        expect(wrapper.find(StyledValidationIcon).prop('data-validation-type')).toEqual(type);
```
jamime: As we're trying to test the icon here I think we should be making assertions on the style and not the prop. The prop that we're asserting here isn't used in the styled component. So we're not really testing anything, from what I can see.

Can we update the `it` description so it says something along the lines of `validation error > displays error icon` instead of `validation error > check icon type`

src/components/validations/validation-icon.component.js
```diff
@@ -31,14 +32,21 @@ const ValidationIcon = ({
     <InputPresentationContext.Consumer>
       {
         context => (
-          <ValidationIconStyle type={ type }>
+          <ValidationIconStyle
+            id={ iconId }
+            data-validation-type={ type }
```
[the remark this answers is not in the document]
joshdj: Updated in commit ee2c98595

src/__experimental__/components/switch/switch.spec.js
joshdj: Updated in commit ee2c98595

PR #2281, opened 2019-08-27
FE-1924 update renogen options for deprecated

edleeks87:
# Description
Creates new entry in `renogen` for deprecation of old input components
Creates `__deprectated__` directory for housing inputs in transition period
Creates `warnOfDeprecation` function to add console warning message when used in development environments
Moves `Textbox` to `__deprecated__` and implements warning function as demo of how others need to be carried out (They should be done separately in stages as the PR will be huge otherwise)

# Screenshot
image
Screenshot of the warning, Textbox no longer needs one but leaving it as an example, ignore the `exprimental` typo 🤦‍♂


change_log/next/FE-1924-adds-deprecated-option-to-headings.yml
```diff
@@ -0,0 +1,4 @@
+New Features: Adds `utils/helpers/warn-as-deprecated` function to warn of deprecated components
+Other: |
+  Updates `.renogen` headings to include a `Deprecated Components` option
+  Creates a `__deprecated__` directory and moves and`Textbox` into it and updates paths
```
chrisbarber86:
```suggestion
  Creates a `__deprecated__` directory and moves and `Textbox` into it and updates paths
```

chrisbarber86: Maybe also need a line here that `Textbox` was deprecated - so if you are using that component you need to update your imports from `components/textbox` to `__depcated__/components/textbox`

PR #2282, opened 2019-08-27
FE-1939 removed min-height on table to resolve row height resize bug.

damianOfKyan:
# removed min-height from Table Component as it serves no purpose.


damianOfKyan:
closing this PR. Implemented solution does not satisfy the requirements and how component is expected to behave. Different solution will be put in place.

PR #2283, opened 2019-08-27
FE-1883 - fix for multiple date visible at the same time

dawidzarzycki:
# Description
- fixes for multiple date pickers visible at the same time

# Testing Instructions
- open storybook
- go to experimental/Date_Input
- open multiple date story

### DIFF
--- a/src/__experimental__/components/date/date.component.js
+++ b/src/__experimental__/components/date/date.component.js
@@ -41,7 +41,7 @@ class DateInput extends React.Component {
 
   isBlurBlocked = false;
 
-  isOpen= false;
+  isOpening = false;
 
   state = {
     isDatePickerOpen: false,
@@ -119,8 +119,8 @@ class DateInput extends React.Component {
   }
 
   closeDatePicker = () => {
-    if (this.isOpen) {
-      this.isOpen = false;
+    if (this.isOpening) {
+      this.isOpening = false;
       return;
     }
     document.removeEventListener('click', this.closeDatePicker);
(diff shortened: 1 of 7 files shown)


src/__experimental__/components/date/date.component.js
```diff
@@ -188,6 +197,12 @@ class DateInput extends React.Component {
     );
   }
 
+  handleClick = (ev) => {
+    if (ev.currentTarget === this.dateRef.current) {
+      this.isCurrentDateFocused = true;
+    }
```
grabkowski: I think that should work without these conditional

src/__experimental__/components/date/date.component.js
```diff
@@ -188,6 +197,12 @@ class DateInput extends React.Component {
     );
   }
 
+  handleClick = (ev) => {
```
grabkowski: could you change this method name to be more meaningful?

src/__experimental__/components/date/date.spec.js
```diff
@@ -214,6 +214,24 @@ describe('Date', () => {
     });
   });
 
+  describe('when there is more than one date input, and we switch between them ', () => {
+    it('we should see only one date picker visible ', () => {
```
damianOfKyan: Perhaps we could expand on this test so that it also checks if the other input has been focused. The 'it' block description does not go in pair with what is being tested.

dawidzarzycki: After changes I've made the tests are no longer necessary

PR #2293, opened 2019-08-29
FE-1944 table ajax errors

damianOfKyan:
# Fixed totalRecords prop type error. Fixed storybook unique key console warnings for table components.


change_log/v9_0_0/FE-1944-table-ajax-errors.yml
```diff
@@ -0,0 +1,2 @@
+Storybook Update: Fixed documentation 'info' and 'note' implementation to avoid duplication that was triggering not unique key console warning.
+Component Update: Fixed, totalRecords prop will now be accepted and passed as a Number not as a String.
```
chrisbarber86: Should really prefix this with the component it's relevant too.
```suggestion
Component Update: TableAjax - totalRecords prop will now be accepted and passed as a Number not as a String.
```

edleeks87:
Everything else looks good Damian, happy to approve once you've made the change Chris suggests 🙂

src/components/table-ajax/table-ajax.stories.js
```diff
@@ -69,7 +72,8 @@ storiesOf('Table Ajax', module)
 
       const pageSize = text('pageSize', '5');
       const paginate = boolean('paginate', TableAjax.defaultProps.paginate);
-      const getCustomHeaders = text('getCustomHeaders');
+      const customHeaders = text('customHeaders', '{ "Accept": "application/json" }');
+      const customHeadersObject = JSON.parse(customHeaders);
```
kyanmichael: After taking a closer look at the [`@storybook/addon-knobs`](https://www.npmjs.com/package/@storybook/addon-knobs) docs, I've discovered that we can use the [`object`](https://www.npmjs.com/package/@storybook/addon-knobs#object) knob here:

```js
const customHeadersObject = object('customHeaders', { Accept: 'application/json' });
```

Not only is this simpler, but more importantly, it has built-in error-handling for invalid JSON. If the user enters an invalid JSON string, the knob will display a red border to alert the user:

Error border

Otherwise, if we use a `text` knob and `JSON.parse()`, the story will crash and the user will get an unpleasant error screen:

Error message

damianOfKyan: updated.

kyanmichael: Thanks :white_check_mark:

PR #2298, opened 2019-08-30
FE-1973 updated css to resolved styling issue.

damianOfKyan:
# Updated css to resolved styling issue for Fieldset component.


damianOfKyan:
> @damianOfKyan can you confirm that we need the triple ampersand? Also, how can this bug be reproduced so I can check that it is fixed?
> 
> I've updated the release notes so they make sense when they are published [here](https://github.com/Sage/carbon/releases). Please feel free to adjust it.

You can remove the styling to see the difference, I think what @edleeks87 originally meant was that using &&& is safer here as it ensure it wont get accident overwritten if the element is deep nested.

jamime:
So this only happens when you have `Form > Fieldset > Textbox`. Which is not currently displayed in the storybook.

Internal conversation on JIRA.

jamime:
So we need the triple & because Form sets using double &. @damianOfKyan can you add a story under Form for Fieldset so we have a way to reproduce. Happy to merge once that has been added.

sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge browsers. Tested Classic, Small, Medium, Large and Generic (None) themes for both the default and fieldset > textbox stories. Testing passed ✅

PR #2302, opened 2019-09-04
FE-1990 - Radio Button component - restore files

dawidzarzycki:
# Description
- restore files inside radio-button component


src/components/radio-button/__spec__.js
```diff
@@ -0,0 +1,66 @@
...
+      <RadioButton
+        name='radiobutton'
+        label='radiobutton'
+      />
+    )
+  });
+
+  describe('render', () => {
+    it('renders a parent div with a pod CSS class', () => {
+      let radioButtonNode = TestUtils.scryRenderedDOMComponentsWithTag(instance, 'div')[0];
```
kyanmichael: (I realise this PR is just restoring the deleted files, so feel free to ignore this comment.)

I think every `let` in this file can be changed to `const` (except for `let instance` on line 8), because their values are never changed.

This should be picked up by https://eslint.org/docs/rules/prefer-const

src/components/radio-button/__spec__.js
```diff
@@ -0,0 +1,66 @@
...
+    instance = TestUtils.renderIntoDocument(
+      <RadioButton
+        name='radiobutton'
+        label='radiobutton'
+      />
+    )
+  });
+
+  describe('render', () => {
+    it('renders a parent div with a pod CSS class', () => {
```
kyanmichael: (I realise this PR is just restoring the deleted files, so feel free to ignore this comment.)

What is `pod` here? Should this be `carbon-radio-button` instead?

sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge. Clicking the help icon no longer activates the button. Testing passed ✅

PR #2308, opened 2019-09-06
 ActionPopover component FE-1253

jamime:
# Description
Initial implementation of the ActionPopover component from the DLS.
>  As a user I'd like to choose from a list of actions in the context of a table row.
# Screenshots
macos - opera
macos - firefox
macos - safari
macos - chrome
win10 - chrome
win10 - opera
win10 - firefox
win10 - ie11
win 10 - edge

# Testing Instructions
You can see this in the `Action Popover > default` story.


src/components/action-popover/action-popover.component.js
```diff
@@ -0,0 +1,251 @@
...
+      e.preventDefault();
+      setOpen(!isOpen);
+      if (isOpen) {
+        button.current.focus();
+      } else {
+        setFocusIndex(0);
+      }
+    } else if (Events.isUpKey(e)) {
+      e.preventDefault();
+      setOpen(true);
```
stevebutler2210: Looking at the examples in the w3 link commented above, shouldn't we be setting the focusIndex here to the last item in the menu?

https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html#kbd1_label

>Up Arrow | Opens menu and moves focus to last menuitem

src/components/action-popover/action-popover.component.js
```diff
@@ -0,0 +1,251 @@
...
+  const button = useRef();
+
+  // Keyboard commands implemented as reccomended by WAI-ARIA best practices
+  // https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-actions.html
+
+  const onButtonKeyDown = useCallback(((e) => {
+    if (Events.isSpaceKey(e) || Events.isDownKey(e) || Events.isEnterKey(e)) {
+      e.preventDefault();
+      setOpen(!isOpen);
+      if (isOpen) {
```
stevebutler2210: Shouldn't these conditions be the other way around?

src/components/action-popover/action-popover.component.js
jamime: Good spot! I was using https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/examples/menu-button/menu-button-1/menu-button-1.html previously where the wording is slightly different.

`Open menu and move focus to previous menuitem`
I interpreted this wrong, which is probably why they have changed the wording.

I'll make a fix.

src/components/action-popover/action-popover.component.js
jamime: So this is checking the current state, before the next render. If it's currently open and we've just pressed Enter we want it to focus the button. Enter is the only event that propagates from the menu item so this code wouldn't be called if the user pressed space/down.

If it's currently closed, and we're about to render the open state we want to set the focus on the first element.

src/components/action-popover/action-popover.component.js
jamime: Fixed in edaf0af030189a540afb12a5209a1ab61e4146bc

src/components/action-popover/action-popover.component.js
stevebutler2210: Ah yep, that makes sense. The actions are the opposite way round in the `onButtonClick` callback (lines 115-125) - does that mean those need changing?

jamime: I see what you mean, I'll refactor that now so they're consistent.

jamime: Resolved in https://github.com/Sage/carbon/pull/2308/commits/758bc43ea1dd974568b7269ec6332496b358b936

src/components/action-popover/action-popover.component.js
```diff
@@ -0,0 +1,248 @@
...
+
+const ActionPopover = ({ children, id }) => {
+  const [isOpen, setOpen] = useState(false);
+  const [focusIndex, setFocusIndex] = useState(0);
+  const [childrenWithRef, setChildrenWithRef] = useState();
+  const [items, setItems] = useState();
+  const [guid] = useState(createGuid());
+  const button = useRef();
+
+
```
stevebutler2210: Sorry, one last thing - an extra / unnecessary new line has crept in here! Once that's gone I'll approve 👍

harpalsingh:
UXQA all good on this, however there is an enhancement we need in a future ticket.

The active row should maintain hover color state when the corresponding action popover is open and the background for the ellipsis should also keep the white background in this state.

PR #2319, opened 2019-09-13
RC Hotfixes (GAC)

jamime:
# Description
- Dialog 
  - prevent warning about passing props to HTMLElement children
- Date
  - add spec helper for DateHelper to ensure that all times are returned as UTC
- Help `tabIndex` default prop had been set to `PropTypes.string` 
  - rename `tabIndexOverride` to be `tabIndex` to match DOM property naming convention
  - update `tabIndex` prop type to allow both strings and numbers
  - rename `tagTypeOverride` to be `as` to follow styled-component naming convention
- Link
  - prevent passing all props to Component's children
  - remove the topmost container to render just an anchor element
  - combine styled components to style only the anchor element
- Table
  - cleaned up warnings displayed in the Unit Test Console
- Form
  - add `innerRef` prop to pass ref back from `form` HTMLElement
- Fieldset
  - discard `childOfForm` and `leftAlign` prop passed from `Form`
# TODO
- [ ] Release notes

# Screenshots / Gifs

# Testing Instructions


package.json
```diff
@@ -116,6 +116,7 @@
     "react-dnd": "^2.4.0",
     "react-dnd-touch-backend": "0.2.7",
     "react-highcharts": "^15.0.0",
+    "react-is": "16.8.2",
```
ianforster-sage: is locking this down to a patch version a deliberate choice?

src/components/button/button.stories.js
```diff
@@ -1,18 +1,21 @@
 import React from 'react';
+// import styled from 'styled-components';
```
ianforster-sage: remove this comment

src/components/button/button.stories.js
```diff
@@ -1,18 +1,21 @@
...
+import Button from '.';
 import getDocGenInfo from '../../utils/helpers/docgen-info';
 
-OriginalButton.__docgenInfo = getDocGenInfo(
+Button.__docgenInfo = getDocGenInfo(
   require('./docgenInfo.json'),
   /button\.component(?!spec)/
 );
 
+// const StyledComponent = styled('div')``.render().type;
```
ianforster-sage: remove this commented code

package.json
edleeks87: The `react` dependency is locked so this one needs to be for now

src/__deprecated__/components/form/form.component.js
```diff
@@ -355,6 +356,10 @@ class FormWithoutValidations extends React.Component {
     };
   }
 
+  isHTMLElement(child) {
+    return isElement(child) && typeof child.type === 'string';
```
kyanmichael: What's the purpose of this condition?
```js
typeof child.type === 'string'
```
Maybe it'd be good to add a comment to the code to clarify its purpose?

src/__experimental__/components/select/select.spec.js
```diff
@@ -44,7 +45,7 @@ describe('Select', () => {
 
   // utility functions to fetch various elements from the wrapper
   const listOf = wrapper => wrapper.find('SelectList');
-  const textboxOf = wrapper => wrapper.find('Textbox');
+  const textboxOf = wrapper => wrapper.find(TextBox);
   const pillsOf = wrapper => wrapper.find('Pill');
```
kyanmichael: Is there a reason for the inconsistency in the types of parameters passed to this set of functions? (`'SelectList'` and `'Pill'` are strings, but `TextBox` is a component)

src/__experimental__/components/textbox/textbox.component.js
```diff
@@ -96,9 +96,9 @@ Textbox.propTypes = {
   warnings: validationsPropTypes,
   /** List of info validation functions */
   info: validationsPropTypes,
-
+  /** Flag to cofigure component when in a Form */
```
kyanmichael: Typo:
```suggestion
  /** Flag to configure component when in a Form */
```

src/components/button-toggle-group/button-toggle-group.component.js
```diff
@@ -29,9 +29,11 @@ const ButtonToggleGroup = withValidations((props) => {
   );
 });
 
-ButtonToggleGroup.propTypes = {
+BaseButtonToggleGroup.propTypes = {
   /** Children to be rendered (ButtonToggle). */
   children: PropTypes.node.isRequired,
+  /** Message displaye on error */
```
kyanmichael: Typo:
```suggestion
  /** Message displayed on error */
```

src/components/link/link.component.js
```diff
@@ -1,13 +1,11 @@
...
-import { assign } from 'lodash';
 import { Link as RouterLink } from 'react-router';
 import Icon from '../icon';
-import { validProps } from '../../utils/ether';
 import Event from '../../utils/helpers/events';
-import tagComponent from '../../utils/helpers/tags';
-import { LinkStyle, LinkStyleAnchor } from './link.style';
+import LinkStyle from './link.style';
 import OptionsHelper from '../../utils/helpers/options-helper';
+import tagComponent from '../../utils/helpers/tags/tags';
```
kyanmichael: I think this can be simplified to:
```suggestion
import tagComponent from '../../utils/helpers/tags';
```

src/components/validations/with-validation.hoc.js
```diff
@@ -272,8 +272,8 @@ const withValidation = (WrappedComponent) => {
     info: []
   };
 
-  const displayName = WrappedComponent.displayName || WrappedComponent.name || 'Component';
-  WithValidation.displayName = `WithValidation(${displayName})`;
+  const displayName = WrappedComponent.displayName || WrappedComponent.name || 'WithValidation(Unknown)';
+  WithValidation.displayName = displayName.replace('Base', '');
```
kyanmichael: I guess the substring `'Base'` will always occur at the beginning of `displayName`? For example:

* `BaseButtonToggleGroup`
* `BaseCarousel`
* `BaseDateInput`
* `BaseForm`

If so, maybe using a regex with a start-of-string anchor `^` would be a bit more robust and also make its behaviour clearer for future developers?

```suggestion
  WithValidation.displayName = displayName.replace(/^Base/, '');
```

src/utils/helpers/with-unique-name/with-unique-name.hoc.js
```diff
@@ -24,8 +24,8 @@ const withUniqueName = (WrappedComponent) => {
       );
     }
   }
-  const displayName = WrappedComponent.displayName || WrappedComponent.name || 'Component';
-  WithUniqueName.displayName = `WithUniqueName(${displayName})`;
+  const displayName = WrappedComponent.displayName || WrappedComponent.name || 'WithUniqueName(Unknown)';
+  WithUniqueName.displayName = displayName.replace('Base', '');
```
kyanmichael: Same suggestion as above:
```suggestion
  WithUniqueName.displayName = displayName.replace(/^Base/, '');
```

src/__deprecated__/components/form/form.component.js
kyanmichael: Thanks for adding that comment in 6342174 :thumbsup:

src/__experimental__/components/textbox/textbox.component.js
kyanmichael: Thanks for committing this fix in 6342174 :thumbsup:

src/components/button-toggle-group/button-toggle-group.component.js
kyanmichael: Thanks for committing this fix in 6342174 :thumbsup:

src/utils/helpers/with-unique-name/__snapshots__/with-unique-name.spec.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`withUniqueName HOC matches the snapshot when the it is passed an anonymous component with no name or display name 1`] = `
-<WithUniqueName(Component)
+<WithUniqueName(Unknown)
```
dawidzarzycki: Does the `Unknown` is correct?

src/__experimental__/components/select/select.spec.js
jamime: We're looking for the TextBox component type and the others are looking for components with the display name of `Pill`, `SelectList`. It was the minimum set of changes required to make the tests pass.

This worked previously because the HOC would append the component display name.
For example `Textbox` would become `WithUniqueName(WithValidations(Textbox)) > WithValidations(Textbox) > Textbox` but for v9 we have chosen to flatten the names so there will be multiple components with the same name (`Textbox > Textbox > Textbox`) due to the amount of GAC tests that are expecting 'Textbox' to return the top-level component. In a future version, we expect to re-instate the HOC display names.

When reviewing this comment I did notice the `Textbox` display name was `ForwardRef` in the snapshots so I fixed that in 2f027d8

* `.find('Textbox')` returns 3 elements
* `.find(TextBox)` returns 1 element

src/components/validations/with-validation.hoc.js
jamime: Fixed in 2f027d8

src/utils/helpers/with-unique-name/with-unique-name.hoc.js
jamime: Fixed in 2f027d8

src/utils/helpers/with-unique-name/__snapshots__/with-unique-name.spec.js.snap
jamime: Yes, this is intentional - when the HOC wraps a component without a display name.

src/__experimental__/components/select/select.spec.js
kyanmichael: @jamime: OK, many thanks for the detailed explanation :thumbsup:

src/components/link/link.component.js
kyanmichael: Thanks for committing this fix in 9b51a45 :thumbsup:

src/components/validations/with-validation.hoc.js
kyanmichael: Thanks :thumbsup:

src/utils/helpers/with-unique-name/with-unique-name.hoc.js
kyanmichael: Thanks :thumbsup:

PR #2323, opened 2019-09-17
FE-2009 deprecate checkable

edleeks87:
# Description
Deprecates checkable components: `checkbox`, `radio-button`, `switch`, `simple-color-picker`, moves them into `__deprecated__` directory, and updates path and cypress scenario accordingly

Hot-fix for Icons in `__deprecated__/components__/switch` as they were positioned incorrectly
Hot-fix alignment of label in `__deprecated__/components__/radio-button` and `__deprecated__/components__/checkbox`
Hot-fix of Icons in `__experimental__/components/switch`


src/__deprecated__/components/radio-button/radio-button.scss
```diff
@@ -19,6 +20,7 @@
 
   &.common-input--has-field-help {
     .common-input__label {
+      vertical-align: unset;
```
kyanmichael: Just wondering whether it's OK to use `unset` ? It's not supported by IE:

* https://developer.mozilla.org/en-US/docs/Web/CSS/unset#Browser_compatibility
* https://caniuse.com/#feat=css-unset-value

src/__deprecated__/components/simple-color-picker/simple-color-picker.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ColorOption from './color-option';
-import tagComponent from '../../utils/helpers/tags';
+import tagComponent from '../../../utils/helpers/tags/tags';
```
kyanmichael: I think this can be simplified to:
```suggestion
import tagComponent from '../../../utils/helpers/tags';
```

src/__deprecated__/components/radio-button/radio-button.scss
edleeks87: Ah yeah very good point, I'll make a safer change

kyanmichael: Thanks. `unset` is used in a couple of other places too:

* https://github.com/Sage/carbon/blob/5efe0b7/src/__deprecated__/components/checkbox/checkbox.scss#L26
* https://github.com/Sage/carbon/blob/5efe0b7/src/components/menu/menu-item/menu-item.scss#L158

edleeks87: Yeah I'll fix the checkbox style, but I'm going to leave the menu-item one (I'll make a jira ticket) as it's not in scope for this and makes the QA more fiddly on this PR 😅

kyanmichael: OK! :ok_hand:

PR #2337, opened 2019-09-19
FE-2033 - Animation Issue

dawidzarzycki:
# Description
Fixes for:
- switching between pages
- missing focus event on IE

### DIFF
--- a/src/components/heading/__spec__.js
+++ b/src/components/heading/__spec__.js
@@ -1,10 +1,16 @@
 import React from 'react';
 import TestUtils from 'react-dom/test-utils';
+import 'jest-styled-components';
 import { shallow, mount } from 'enzyme';
 import Heading from './heading';
 import Help from './../help';
 import Link from './../link';
 import { elementsTagTest, rootTagTest } from '../../utils/helpers/tags/tags-specs';
+import { assertStyleMatch } from '../../__spec_helper__/test-utils';
+import DefaultPages from '../pages/pages.component'
+import Page from '../pages/page/page.component'
+import { PagesContent } from '../pages/pages.style';
+import { LinkStyleAnchor } from '../link/link.style';
 
 describe('Heading', () => {
   let instance;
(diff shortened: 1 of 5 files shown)


src/components/heading/heading.js
```diff
@@ -120,6 +120,8 @@ class Heading extends React.Component {
 
     return (
       <Link
+        // this event allow en element to be focusable on cick event on IE
```
ianforster-sage:
```suggestion
        // this event allows an element to be focusable on click event on IE
```

src/components/heading/heading.js
```diff
@@ -120,6 +120,8 @@ class Heading extends React.Component {
...
+        onMouseDown={ e => e.currentTarget.focus() }
```
joshdj: Would using `:active` pseudo-class not achieve the same effect? Instead of faking a blur.

https://developer.mozilla.org/en-US/docs/Web/CSS/:active

dawidzarzycki: I tried the solution but Internet Explorer won't cooperate :/ 
This is the one solution I found for the Internet Explorer. All because there is an anchor and IE doesn't want to setup focus on it while clicking.

joshdj:
> I tried the solution but Internet Explorer won't cooperate :/
> This is the one solution I found for the Internet Explorer. All because there is an anchor and IE doesn't want to setup focus on it while clicking.

@dawidzarzycki 

Out of curiosity I did some research and I got it to work using the following code.
This isn't to say you need to do it this way, but I thought I'd share an alternative solution.

**src/components/pages/pages.style.js (line 18)**
```
&:active,
&:focus {
   outline: solid 3px ${theme.colors.focus};
}
```

**src/components/link/link.style.js (addded after line 24)**
```
span {
   pointer-events: none;
}
```

dawidzarzycki:
> > I tried the solution but Internet Explorer won't cooperate :/
> > This is the one solution I found for the Internet Explorer. All because there is an anchor and IE doesn't want to setup focus on it while clicking.
> 
> @dawidzarzycki
> 
> Out of curiosity I did some research and I got it to work using the following code.
> This isn't to say you need to do it this way, but I thought I'd share an alternative solution.
> 
> **src/components/pages/pages.style.js (line 18)**
> 
> ```
> &:active,
> &:focus {
>    outline: solid 3px ${theme.colors.focus};
> }
> ```
> 
> **src/components/link/link.style.js (added after line 24)**
> 
> ```
> span {
>    pointer-events: none;
> }
> ```

You are right, just tried...
The key is pointer-events in the link component...
I won't make changes in the link right now. As I know it is on refactor so I'd like to stay with current solution to don't affect the link

src/components/heading/__spec__.js
```diff
@@ -41,6 +47,24 @@ describe('Heading', () => {
...
+        }
+        >test
+          </Page>
+      </DefaultPages>)
+
+    const link = wrapper.find(Link);
+    link.find('a').simulate('mousedown');
+
+    assertStyleMatch({
+      outline: 'solid 3px #FFB500'
```
damianOfKyan: lets use theme here to make it more future/change proof 

`outline: 'solid 3px ' + classicTheme.colors.focus`

>

dawidzarzycki: Good point! 
Updated in https://github.com/Sage/carbon/pull/2337/commits/c3c4a51be627500c74d3912caa6aef3f15818c45

joshdj:
@dawidzarzycki Would it be worth adding `aria-label={ `${this.props.title} back button` }` to the Link component inside the Heading component?

PR #2338, opened 2019-09-20
Hot fix to allow default checked radio

edleeks87:
# Description
Currently setting a RadioButton to `checked` is ignored when it is in a RadioButtonGroup, I have added some code to ensure it is respected and applied when it is passed

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button.stories.js
+++ b/src/__experimental__/components/radio-button/radio-button.stories.js
@@ -20,6 +20,7 @@ function makeStory(name, themeSelector) {
       >
         <RadioButton
           id='input-1'
+          checked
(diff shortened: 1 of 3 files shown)


src/__experimental__/components/radio-button/radio-button.stories.js
```diff
@@ -20,6 +20,7 @@ function makeStory(name, themeSelector) {
       >
         <RadioButton
           id='input-1'
+          checked
```
damianOfKyan: maybe lets use checked="checked" ?

edleeks87: The PropType is a bool though and the html5 spec supports just using checked at level

PR #2339, opened 2019-09-23
Merge back for v8 releases

edleeks87:
# Description
Merges changes from `v8.1.4` and `v8.1.5` into master branch


edleeks87:
> Removing the prop `showPickerOnTop` should have been a breaking change.

Yeah agreed, can't do much about it now as it's been released already 😞 , the prop was only added as part of `v8.1.2` as well

PR #2340, opened 2019-09-23
FE-1569 fix table classic

ianforster-sage:
# Description
Fix styling of draggable table cell icon cursor.
Fix setting of base theme by moving it from component to style

### DIFF
--- a/src/components/table/table-row/__snapshots__/table-row.spec.js.snap
+++ b/src/components/table/table-row/__snapshots__/table-row.spec.js.snap
@@ -49,7 +49,9 @@ exports[`TableRow render if is not classic theme renders a row to match the snap
 }
 
 .c2 .c14 {
-  cursor: move;
+  cursor: -webkit-grab;
+  cursor: -moz-grab;
+  cursor: grab;
 }
 
 .custom-drag-layer .c2.c2 .c14 {
--- a/src/components/table/table-row/table-row.spec.js
+++ b/src/components/table/table-row/table-row.spec.js
@@ -593,7 +593,7 @@ describe('TableRow', () => {
         expect(cell.props().canDrag).toEqual(true);
 
         assertStyleMatch({
-          cursor: 'move'
+          cursor: 'grab'
         },
         draggableRow, { modifier: `${StyledIcon}` });
       });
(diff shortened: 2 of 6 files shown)


src/components/table/table-row/table-row.spec.js
```diff
@@ -593,7 +593,7 @@ describe('TableRow', () => {
         expect(cell.props().canDrag).toEqual(true);
 
         assertStyleMatch({
-          cursor: 'move'
+          cursor: 'grab'
```
kyanmichael: I made this identical change in 4bf6496 in [`FE-1569-react-dnd-and-dls-table-row`](https://github.com/Sage/carbon/commits/FE-1569-react-dnd-and-dls-table-row). Since the change is identical, it should merge cleanly without conflict.

src/components/table/table-row/__snapshots__/table-row.spec.js.snap
```diff
@@ -49,7 +49,9 @@ exports[`TableRow render if is not classic theme renders a row to match the snap
 }
 
 .c2 .c14 {
-  cursor: move;
+  cursor: -webkit-grab;
+  cursor: -moz-grab;
+  cursor: grab;
```
kyanmichael: I made this identical change in 4bf6496 in [`FE-1569-react-dnd-and-dls-table-row`](https://github.com/Sage/carbon/commits/FE-1569-react-dnd-and-dls-table-row). Since the change is identical, it should merge cleanly without conflict.

src/components/table/table-row/table-row.style.js
```diff
@@ -122,4 +124,13 @@ function dragRowStyling({ isDragged, isDragging, theme }) {
     `}
   `;
 }
+
+StyledTableRow.propTypes = {
+  theme: PropTypes.object
+};
+
+StyledTableRow.defaultProps = {
+  theme: baseTheme
```
kyanmichael: Interesting – so moving `.defaultProps = { theme: baseTheme }` from `table-row.component.js` to `table-row.style.js` resolves the issue of the Classic theme being lost on `<Table>` ?

Maybe this is something we need to document – that `.defaultProps = { theme: baseTheme }` must always be added to the `*.style.js` file (not the `*.component.js` file) ?

edleeks87: Yeah the default theme is only needed at the component level when we need to use the `withTheme` hoc for some conditional functionality, surprised it broke it though

PR #2346, opened 2019-09-24
FE-1569 and FE-1327: ReactDnD and DraggableContext / draggable table rows

kyanmichael:
# Description

* FE-1569: Upgraded `react-dnd-test-backend` to v2.6.0.
* Upgraded `src/components/table` from SCSS to StyledComponents.
* Upgraded `src/components/table/table-row` from SCSS to StyledComponents.
* Upgraded `src/components/table/draggable-table-cell` from SCSS to StyledComponents.
* Upgraded `src/components/drag-and-drop/custom-drag-layer` from SCSS to StyledComponents.
* Implemented required functionality/behaviour specified by FE-1327.


src/__deprecated__/components/checkbox/__spec__.js
```diff
@@ -12,6 +12,7 @@ describe('Checkbox', () => {
       <Checkbox
         name='checkbox'
         label='checkbox'
+        checked={ true }
```
ianforster-sage: Does this need to be assigned a value? When it's set to `true` isn't it preferable to just use `checked` without an assignment?

src/components/table/table-row/table-row.component.js
```diff
@@ -23,6 +23,7 @@ import ActionPopover from '../../action-popover';
  */
 class TableRow extends React.Component {
   state = {
+    rowNode: null,
```
ianforster-sage: Does this need comments to annotate the documentation?

src/components/table/table-row/table-row-modern.style.js
```diff
@@ -1,30 +1,79 @@
+import { css } from 'styled-components';
```
dawidzarzycki: Just noticed that we have `table-row-modern`
What I see we have two types of files:
- `component-classic.style.js` (for legacy styles)
- `component.style.js` (for dls styles)
Can you change the `table-row-modern.style.js` to `table-row.style.js` to keep convention please 😃

src/components/table/table-row/table-row-modern.style.js
```diff
@@ -1,30 +1,79 @@
...
-    background-color: ${table.selected} !important;
-    border-bottom-color: ${table.selected} !important;
+    background-color: ${table.selected};
+    border-bottom-color: ${table.selected};
+  `;
+}
+
+function applyModernDraggedStyling() {
+  return css`
+    border: 1px solid #000A0E;
```
dawidzarzycki: I'm not sure we should to use hardcoded values like this one `#000A0E` inside the dls styles. If we don't have this color  in the color themes maybe it should be good to add this there

src/__deprecated__/components/checkbox/__spec__.js
kyanmichael: Good point. I'll fix this.

This should've been picked up by the [`react/jsx-boolean-value`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) ESLint rule. This is yet another occurrence of the FE-1870 issue :disappointed:

src/components/table/table-row/table-row.component.js
kyanmichael: Yes, for consistency with the other `state` variables in this component (`highlighted` and `selected`). I'll fix this.

src/components/table/table-row/table-row.style.js
```diff
@@ -1,166 +1,148 @@
...
-          content: "";
-          height: 1px;
-          left: 0;
-          position: absolute;
-          top: -1px;
-          width: 100%;
+    ${isSelected && css`
+      &&&&,
+      &&&&:nth-child(odd),
+      &&&&:hover {
```
dawidzarzycki: Is there any other option than do `&&&&`?

src/__deprecated__/components/checkbox/__spec__.js
kyanmichael: Fixed in 42fda8d :thumbsup:

src/components/table/table-row/table-row.component.js
kyanmichael: Fixed in 42fda8d :thumbsup:

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.style.js
```diff
@@ -0,0 +1,18 @@
...
+  padding: 0;
+  pointer-events: none;
+  position: fixed;
+  top: 0;
+  width: 100%;
+  z-index: 100;
+  box-shadow: 0 10px 20px 0 rgba(0,20,29,0.2), 0 20px 40px 0 rgba(0,20,29,0.1);
+`;
+
+const StyledDragWrapper = styled.div``;
```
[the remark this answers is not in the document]
kyanmichael: I don't think we _have_ to use `styled.div` here. I guess we could probably use something like this instead:
```jsx
const StyledDragWrapper = (props) => <div>{props.children}</div>;
```
But I think using `styled.div` is cleaner, and also makes it clear that CSS styling can be easily added to the component in the future if needed.

src/components/table/table-row/table-row-modern.style.js
kyanmichael: That is indeed the naming convention we generally use.

But in this case, we actually have 3 `style` files:

* [`table-row.style.js`](https://github.com/Sage/carbon/blob/42fda8d/src/components/table/table-row/table-row.style.js)  –  This imports from `table-row-modern.style.js` and `table-row-classic.style.js`
* [`table-row-modern.style.js`](https://github.com/Sage/carbon/blob/42fda8d/src/components/table/table-row/table-row-modern.style.js)
* [`table-row-classic.style.js`](https://github.com/Sage/carbon/blob/42fda8d/src/components/table/table-row/table-row-classic.style.js)

So I can't rename the `table-row-modern.style.js` file to `table-row.style.js`, because we already have a file called `table-row.style.js`.

src/components/table/table-row/table-row.style.js
kyanmichael: I don't think so :disappointed:

@stevebutler2210 implemented this in 23a1e4a, and I believe he tried every possible alternative solution he could think of.

I think the problem is that the table row is styled with a high specificity — `StyledTable & StyledTableRow &:nth-child(2n)` — in `table.style.js` and `table-modern.style.js`, so we need to use [an even higher specificity here with `&&&&`](https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity) to make it work.

src/components/table/table-row/table-row-modern.style.js
kyanmichael: Good point.

After investigating this more carefully, I discovered that this `border` style isn't actually needed and has no effect. It looks like it was WIP code which @stevebutler2210 was working on. I've now removed it in 8491168.

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.style.js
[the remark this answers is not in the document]
dawidzarzycki: I just wonder should we use `StyledComponent` without any styling. I know it is better for the future if we want to add styles... but what if we won't? I think it's like adding `className` to the component without any styling. It can causes a confusion

src/components/table/__snapshots__/table.spec.js.snap
```diff
@@ -12,26 +12,26 @@ exports[`Table when the table size is compact renders a table to match the expec
...
   background-color: #F9FAFB;
 }
 
-.c0.c0 .carbon-table-row:hover .c2 {
+.c0 .c2:hover .c1 {
   background-color: #F2F4F5;
 }
 
-.c0.c0 .carbon-table-row:not(.carbon-table-row--passive):hover .c2 {
+.c0 .c2:not(.carbon-table-row--passive):hover .c1 {
```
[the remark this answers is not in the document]
kyanmichael: Good point. I agree that that class is obsolete now. It became obsolete when `mainClasses()` was removed from `src/components/table/table-row/table-row.component.js`:

* https://github.com/Sage/carbon/commit/23a1e4a#diff-a41f7b019054fbc700f428b1213f76c4L138

I've now removed it from `src/components/table/table-modern.style.js` in 8858e2e :thumbsup:

src/components/table/draggable-table-cell/draggable-table-cell.style.js
```diff
@@ -10,6 +10,10 @@ const StyledDraggableTableCell = styled(StyledTableCell)`
   ${StyledTableCell}:first-child {
     padding-left: 0;
   }
+
+  .draggable-table-cell__icon {
```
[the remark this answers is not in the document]
kyanmichael: Yes indeed. Refactored in 7195b51 :thumbsup:

src/components/table/draggable-table-cell/draggable-table-cell.spec.js
```diff
@@ -30,6 +32,13 @@ describe('DraggableTableCell', () => {
     expect(icon.props().type).toEqual('drag_vertical');
   });
 
+  it('styles the icon cursor', () => {
+    const component = TestRenderer.create(<StyledDraggableTableCell identifier='foo' />);
+    assertStyleMatch({
+      cursor: 'move'
+    }, component.toJSON(), { modifier: '.draggable-table-cell__icon' });
```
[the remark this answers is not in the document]
kyanmichael: Refactored in 7195b51 :thumbsup:

src/components/table/draggable-table-cell/draggable-table-cell.component.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import WithDrag from '../../drag-and-drop/with-drag';
 import Icon from '../../icon';
-import TableCell from '../table-cell';
+import StyledDraggableTableCell from './draggable-table-cell.style';
 
 const iconHTML = (
```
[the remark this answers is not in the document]
kyanmichael: Removed in 7195b51 :thumbsup:

src/components/drag-and-drop/custom-drag-layer/custom-drag-layer.style.js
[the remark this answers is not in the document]
kyanmichael: Fair point. It may be a case of [YAGNI](https://en.wikipedia.org/wiki/YAGNI) :slightly_smiling_face:

Refactored in d4f2a93 to remove `StyledDragWrapper` :thumbsup:

src/components/table/draggable-table-cell/draggable-table-cell.style.js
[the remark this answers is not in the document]
edleeks87: Thanks

src/components/table/__snapshots__/table.spec.js.snap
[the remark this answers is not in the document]
edleeks87: Thanks

src/components/table/draggable-table-cell/draggable-table-cell.spec.js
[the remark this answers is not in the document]
edleeks87: Thanks

src/components/table/draggable-table-cell/draggable-table-cell.component.js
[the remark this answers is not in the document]
edleeks87: Thanks

PR #2354, opened 2019-09-26
FE-2057 radio validation story refactor

edleeks87:
# Description
Groups the knobs so there is more granular control over each component
Add a help label to non-validation stories

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button.stories.js
+++ b/src/__experimental__/components/radio-button/radio-button.stories.js
@@ -35,7 +35,27 @@ const groupedKnobs = (type, themeName) => {
     value: text(`${type} value`, type, type),
     disabled: boolean(`${type} disabled`, false, type),
     reverse: boolean(`${type} reverse`, false, type),
-    size: themeName !== 'classic' ? select('size', OptionsHelper.sizesBinary, 'small', type) : undefined
+    size: themeName !== 'classic' ? select(`${type} size`, OptionsHelper.sizesBinary, 'small', type) : undefined,
+    fieldHelp: text(`${type} fieldHelp`, 'This text provides help for the input.', type),
+    fieldHelpInline: boolean(`${type} fieldHelpInline`, false, type),
+    inputWidth: number(`${type} inputWidth`, 0, {
+      range: true,
+      min: 0,
+      max: 100,
+      step: 1
+    }, type),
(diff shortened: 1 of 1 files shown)


edleeks87:
> Also `labelAlign` appears to have no effect.

`labelAlign` does have an effect when the label has a larger width applied. Have made the other changes you requested

src/__experimental__/components/radio-button/radio-button.stories.js
```diff
@@ -27,11 +27,45 @@ function makeStory(name, themeSelector, component) {
...
-  const knobs = defaultKnobs();
+const groupedKnobs = (type, themeName) => {
+  return {
+    key: type,
+    label: text(`${type} label`, `Example ${type} radio button`, type),
+    labelHelp: text(`${type} labelHelp`, 'This text provides more information for the label.', type),
+    value: text(`${type} value`, type, type),
+    disabled: boolean(`${type} disabled`, false, type),
+    reverse: boolean(`${type} reverse`, false, type),
+    size: themeName !== 'classic' ? select(`${type} size`, OptionsHelper.sizesBinary, 'small', type) : undefined,
```
kyanmichael: I recommend inverting the condition:

```suggestion
    size: themeName === 'classic' ? undefined : select(`${type} size`, OptionsHelper.sizesBinary, 'small', type),
```

IMO we should enable ESLint's [`no-negated-condition`](https://eslint.org/docs/rules/no-negated-condition) rule:

> Negated conditions are more difficult to understand. Code can be made more readable by inverting the condition instead.

src/__experimental__/components/radio-button/radio-button.stories.js
```diff
@@ -40,29 +74,27 @@ const radioComponent = () => {
...
-        value={ text('radioTwoValue', 'monthly') }
         { ...knobs }
+        { ...groupedKnobs('monthly', themeName) }
       />
       <RadioButton
         // id prop intentionally left off here, to demonstrate automatic GUID generation
         name='input-2'
-        label={ text('radioThreeLabel', 'Example Annual Radio Button') }
-        value={ text('radioThreeValue', 'annually') }
+        key='Radio Three'
```
jamime: I don't think we need this `key` as it's passed in from `groupedKnobs`

PR #2359, opened 2019-09-27
FE-2073 checkbox labels

ianforster-sage:


ianforster-sage:
> Are we still going with the approach of grouping knobs?

We are, but these changes are still valid. the label would be removed anyway when we group knobs, and the icon styling would still need to change.

PR #2363, opened 2019-09-30
Fe 2060 radio keyboard access

damianOfKyan:
# Description

Added keyboard accessibility and test coverage.
Navigation through keyboard shortcut as follows:

Tab
Moves keyboard focus to the checked radio button in a radiogroup.
Or, if no radio button is checked, focus moves to the first radio button in the group.
Further tab moves focus out of the radio group to the next component on the page.
Space
If the radio button with focus is unchecked, it's state will be changed to checked.
If radio is already checked, has no effect.
Down/Right arrows
Moves focus to next radio button in the group.
If focus is on the last radio button in the group, move focus to the first radio button.
Up/Left arrows
Moves focus to a previous radio button in the group.
If focus is on the first radio button in the group, move focus to the last radio button.


edleeks87:
@damianOfKyan really sorry I realised last night that I added `findIndex` isn't compatible for IE 🤦‍♂, I've fixed it and I'll push up the code

src/__experimental__/components/radio-button/radio-button-group.spec.js
```diff
@@ -309,7 +309,7 @@ describe('RadioButtonGroup', () => {
 
           describe('when down arrow key is pressed twice', () => {
             describe('on one down arrow key press', () => {
-              it('should change the focus to the last radio button', () => {
+              it('should change the focus to the second radio button', () => {
```
damianOfKyan: there are 3 radio buttons, [0, 1, 2]. radio button with key 2 is (3rd radio button aka last)

edleeks87: I didn't change them intentionally, I must have weird state locally, but all I did was pull origin and update the function above

jamime:
Fixed by #2378

PR #2368, opened 2019-10-02
FE-2077 Storybook - Tables -> downArrow has no theming color

mkrds:
# Description
Currently downArrow color does not change to the corresponding theme color when theme is changed.

# Screenshots / Gifs
ThemeLarge
ThemeMedium
ThemeNone
ThemeSmall


sianford:
Tested all themes in Safari, Chrome, Firefox, IE11 and Edge. The dropdown arrow is the correct themed colour now ✅

PR #2370, opened 2019-10-03
Improve rendering performance of validations and radio button FE-2059

jamime:
This change required a refactor to the validations hoc.

- Reduced the number of components that are re-rendered when the radio selection changes
- RadioButton is now dependent on RadioGroup
- Moved onChange callback from RadioButton to RadioGroup
- Fixed RadioButton/RadioGroup so it can be used as an uncontrolled component

# Testing/QA
Please pay particular attention to the validation functionality, as the HOC has been changed this could affect any of the inputs that provide validations. I have checked it myself and could not find any regressions.

# Breaking Changes
The validations HOC has been refactored to reduce the number of state updates. To do this some of the internal methods have been re-written. In summary, only `validate()` should be considered as a stable API. There are no indications that any of the internal methods have been used outside of carbon. Typically consumers have hooked in at the `form` level, not the `input` level validations.

# Screenshots / Gifs
The rendering performance of radio button was a likley cause to see a flash of the help icon as validation changes from `error` to running validations (`help`) to `warning`.

To record these screenshots I created a `RadioGroup` with three `RadioButtons`. The component was controlled with the state of `RadioGroup` being stored in `App`. To set up I focused the first radio button, started profiling, selected the second radio button, stopped profiling.

## Before
Selecting warning, re-renders everything
3 select warning
Validations reset, re-renders everything
4 errorMessage prop change
Validation complete, re-renders everything
5 warningMessage prop change
Scheduled re-render because react hook changed?
6 app re-rendered

## After
Select warning, third radio button, and label of radio button 1 and 2 skipped rendering
4 select warning
Validations reset, now a fast render <0.1ms
5 empty commit
Validations complete, only label renders.
6 warningMessage prop change

No scheduled re-render.


src/__experimental__/components/radio-button/__snapshots__/radio-button.spec.js.snap
```diff
@@ -99,863 +103,3344 @@ exports[`RadioButton styles base renders as expected 1`] = `
...
+        "color": "rgba(0,0,0,0.9)",
+        "placeholder": "rgba(0,0,0,0.3)",
+        "size": "14px",
+      },
+      "tile": Object {
+        "border": "#CCD6DA",
+        "separator": "#E5EAEC",
+      },
+    }
+  }
```
grabkowski: please make sure to not pass the theme config in snapshots

src/__experimental__/components/radio-button/radio-button-group.spec.js
```diff
@@ -180,4 +186,100 @@ describe('RadioButtonGroup', () => {
       );
     });
   });
+
+  describe('controlled', () => {
+    const Controller = (props) => {
+      const [value, setValue] = useState();
+      return (
+        <React.Fragment>
```
grabkowski: could you use shorter syntax here?

PR #2381, opened 2019-10-10
FE-2083-mouse-cursor-issue

dawidzarzycki:
- fix for cursor issue


harpalsingh:
Can test the UXQA this in the feature branch rather than just this specific issue

PR #2385, opened 2019-10-14
FE-2124 - Fix date input to allow it to be controlled using value prop and onChange 

edleeks87:
# Description
Sage 200 raised an issue relating to how the value prop did not allow the date input to be controlled after initial render. This PR adds a further condition to `componentDidUpdate` to allow for the input to be controlled as well as for the other value inputs to continue to work as expected.
These changes can be QA'd using this branch and https://github.com/Sage/carbon/tree/fixture

I have added a flag to check if the input has focus so that the value updates are not triggered when a user is typing in the input

Also added is support for using the component in controlled or uncontrolled and the format of the event emitted `onChange` and `onBlur`


src/__experimental__/components/date/date.component.js
```diff
@@ -116,9 +100,13 @@ class BaseDateInput extends React.Component {
 
   reformatVisibleDate = () => {
     const { visibleValue } = this.state;
-
     if (DateHelper.isValidDate(visibleValue)) {
-      this.setState({ visibleValue: formatDateToCurrentLocale(visibleValue) });
+      this.setState({ visibleValue: formatDateToCurrentLocale(visibleValue) }, () => {
+        if (this.props.onBlur) {
+          const dateWithSlashes = DateHelper.sanitizeDateInput(visibleValue);
+          this.props.onBlur({ target: this.input }, DateHelper.formatValue(dateWithSlashes));
```
kyanmichael: I see that this `this.props.onBlur()` call was originally in the `handleBlur()` method, but has been moved here.

The call is now conditional on `if (DateHelper.isValidDate(visibleValue)) {` so if that condition is `false` then `this.props.onBlur()` won't be called when a `blur` event occurs. Is that the intended behaviour?

src/__experimental__/components/date/date.component.js
```diff
@@ -133,19 +121,24 @@ class BaseDateInput extends React.Component {
...
+  updateVisibleValue = (date, pickerUsed) => {
     const visibleValue = formatDateToCurrentLocale(date);
 
     this.setState({
       selectedDate: date,
       visibleValue
+    }, () => {
+      if (pickerUsed) {
+        const event = {
+          target: this.input
```
kyanmichael: It looks like we create these kinds of "synthesised event objects" quite often in various components: `event = { target: … }`

Maybe it'd be good to have a helper/utility function to DRY this up and to clarify its purpose? eg:
```js
/** @see https://developer.mozilla.org/en-US/docs/Web/API/Event/target */
function makeEventFromTarget(targetElement) {
  return { target: targetElement };
}

this.emitOnChangeCallback(makeEventFromTarget(this.input), date);
```

src/__experimental__/components/date/date.spec.js
```diff
@@ -367,6 +360,26 @@ describe('Date', () => {
         expect(input.instance().value).toBe(mockDate);
       });
     });
+
+    describe('controlled vs uncontrolled input', () => {
+      it('supports being used as an controlled input via passing of a value prop', () => {
+        wrapper = render({ value: '27rd Feb 01' });
```
kyanmichael: Just wondering whether this `27rd` is intentional? :slightly_smiling_face: (instead of `27th`)

src/__experimental__/components/date/date.component.js
edleeks87: Yep agreed I'm currently doing such an approach in the refactor now the format for emitted events has been agreed upon

src/__experimental__/components/date/date.component.js
edleeks87: You're right I don't think the onBlur should be conditional on this, will refactor

kyanmichael: Thanks for moving the `this.props.onBlur()` call back into the `handleBlur()` method :thumbsup:

src/__experimental__/components/date/date.spec.js
kyanmichael: Thanks for fixing this :slightly_smiling_face:

src/__experimental__/components/date/date.component.js
```diff
@@ -179,17 +170,28 @@ class BaseDateInput extends React.Component {
...
-
-    if (this.props.onChange) this.props.onChange(changePayload);
+  emitOnChangeCallback = (ev, isoFormattedValue) => {
+    if (this.props.onChange) {
+      const event = this.buildCustomTarget(ev, isoFormattedValue);
+      this.props.onChange(event);
+    }
   };
 
+  buildCustomTarget = (ev, isoFormattedValue) => {
```
kyanmichael: Maybe `buildCustomEvent` might be a better name? Or `customiseEventTarget` ?

(The name `buildCustomTarget` sounds like it returns a custom target object, whereas it actually returns a custom event object.)

edleeks87: Yep agreed I had it as `buildCustomEvent` originally so will go with that

kyanmichael: Great, thanks :thumbsup:

src/__experimental__/components/date/date.stories.js
```diff
@@ -31,7 +31,7 @@ const store = new Store(
 
 const setValue = (ev) => {
   action('onChange')(ev);
-  store.set({ value: ev.target.value });
+  store.set({ value: ev.target.value.displayText });
```
kyanmichael: Where does `displayText` come from? :thinking:

edleeks87: Oh good spot I missed that when I refactored I will correct it thanks

kyanmichael: Thanks for fixing that in 8aa3379 :thumbsup:

src/__experimental__/components/date/date.component.js
```diff
@@ -62,7 +41,10 @@ class BaseDateInput extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (this.isBlurBlocked && this.hasValueChanged(prevProps)) {
+    if (this.isControlled && !this.inputHasFocus && this.hasValueChanged(prevProps)) {
```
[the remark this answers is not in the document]
kyanmichael: OK, thanks for creating FE-2262 for that :thumbsup:

src/__experimental__/components/date/date.component.js
```diff
@@ -13,45 +13,29 @@ import withUniqueName from '../../../utils/helpers/with-unique-name';
...
     /** Date object to pass to the DatePicker */
-    selectedDate: DateHelper.stringToDate(this.props.value),
+    selectedDate: DateHelper.stringToDate(DateHelper.formatValue(this.initialVisibleValue)),
     /** Displayed value, format dependent on a region */
-    visibleValue: formatDateToCurrentLocale(this.props.value)
+    visibleValue: formatDateToCurrentLocale(this.initialVisibleValue),
+    /** Stores last valid values to be emitted onBlur if current input is invalid */
+    lastValidEventValues: {
+      formattedValue: formatDateToCurrentLocale(this.initialVisibleValue),
+      rawValue: DateHelper.formatValue(this.initialVisibleValue)
```
kyanmichael: There's some discussion about whether nested objects in `state` are a good idea :point_right: https://stackoverflow.com/a/51136076

I think it's OK here, but it's worth bearing this in mind :point_right: [Speed up your React Components with flattened states and `React.PureComponent`](https://www.codementor.io/manasjayanth/speed-up-your-react-components-with-flattened-states-and-react-purecomponent-m3hv6mlrf)

src/__experimental__/components/date/date.component.js
```diff
@@ -162,34 +180,53 @@ class BaseDateInput extends React.Component {
...
 
+  buildCustomEvent = (ev, isoFormattedValue) => {
+    const { id, name, value } = ev.target;
+    const { lastValidEventValues } = this.state;
+    const validRawValue = DateHelper.isValidDate(isoFormattedValue);
+
+    ev.target = {
+      ...(name && { name }),
+      ...(id && { id }),
+      value:
```
damianOfKyan: this feels bit overwhelming, perhaps we could refactor that into separate helper function?

src/__experimental__/components/date/date.spec.js
```diff
@@ -178,11 +177,27 @@ describe('Date', () => {
...
           simulateFocusOnInput(wrapper);
           wrapper.setProps({ value: secondDate });
           expect(wrapper.find(DatePicker).exists()).toBe(true);
-          expect(onBlurFn).toHaveBeenCalled();
+          expect(onBlurFn).not.toHaveBeenCalled();
+        });
+      });
+
+      describe('and the rawValue is invalid', () => {
+        it('then onBlur prop should not have been called', () => {
```
damianOfKyan: it block text does not match the content.

src/__experimental__/components/date/date.component.js
```diff
@@ -132,7 +137,14 @@ class BaseDateInput extends React.Component {
     const visibleValue = formatDateToCurrentLocale(date);
     const newDate = this.getDateObject(date);
 
-    this.setState({ selectedDate: newDate, visibleValue },
+    this.setState({
+      selectedDate: newDate,
+      visibleValue,
+      lastValidEventValues: {
+        formattedValue: visibleValue,
+        rawValue: DateHelper.formatValue(visibleValue)
```
kyanmichael: There's a case of unfortunate naming here.

You'd assume that the return value of `DateHelper.formatValue()` should be assigned to `formattedValue`, so it looks like a mistake that it's being assigned to `rawValue` instead.

But I realise that `DateHelper.formatValue()` returns an ISO date string (`YYYY-MM-DD`), which is what we want in `rawValue`.

It'd be good if `DateHelper` had a method called `isoValue()` for use in situations like this.

src/__experimental__/components/date/date.component.js
kyanmichael: Maybe this would be better?
```js
value: {
  ...lastValidEventValues,
  ...(validRawValue && { formattedValue: formatDateToCurrentLocale(value) }),
  ...(validRawValue && { rawValue: isoFormattedValue })
}
```
Would that work?

src/__experimental__/components/date/date.spec.js
```diff
@@ -178,11 +177,27 @@ describe('Date', () => {
...
+          const event = {
+            target: {
+              name: 'foo',
+              id: 'foo',
+              value: '21/12/122'
+            }
+          };
+          expect(wrapper.find(BaseDateInput).instance()
+            .buildCustomEvent(event, 'foo').target.value)
+            .toEqual({ formattedValue: '12/08/2019', rawValue: '2019-08-12' });
```
kyanmichael: Maybe this would be clearer?
```suggestion
            .toEqual({ formattedValue: firstDate, rawValue: '2019-08-12' });
```

src/__experimental__/components/date/date.component.js
edleeks87: Yep good point, I had this as WIP whilst I tried it out, I'll refactor it thanks

src/__experimental__/components/date/date.component.js
```diff
@@ -110,15 +105,20 @@ class BaseDateInput extends React.Component {
...
     const { visibleValue } = this.state;
-
     if (DateHelper.isValidDate(visibleValue)) {
-      this.setState({ visibleValue: formatDateToCurrentLocale(visibleValue) });
+      this.setState({
+        visibleValue: formatDateToCurrentLocale(visibleValue),
+        lastValidEventValues: {
+          formattedValue: formatDateToCurrentLocale(visibleValue),
+          rawValue: DateHelper.formatValue(visibleValue)
+        }
```
kyanmichael: There are 3 places in this file where this identical object structure is created:
```js
{
  visibleValue: formatDateToCurrentLocale(visibleValue),
  lastValidEventValues: {
    formattedValue: formatDateToCurrentLocale(visibleValue),
    rawValue: DateHelper.formatValue(visibleValue)
  }
}
```
It'd be good to DRY that up, by creating a helper function, which takes a `visibleValue` parameter, and creates and returns that object :point_up:

src/__experimental__/components/date/date.spec.js
edleeks87: Yep good spot, will fix

src/__experimental__/components/date/date.component.js
edleeks87: Yeah agree, in this scenario I think it's fine as it's essentially just maintaining a snapshot of the previous valid values

src/__experimental__/components/date/date.component.js
```diff
@@ -162,34 +180,53 @@ class BaseDateInput extends React.Component {
...
+    const { id, name, value } = ev.target;
+    const { lastValidEventValues } = this.state;
+    const validRawValue = DateHelper.isValidDate(isoFormattedValue);
+
+    ev.target = {
+      ...(name && { name }),
+      ...(id && { id }),
+      value:
+        {
+          formattedValue: validRawValue ? formatDateToCurrentLocale(value) : lastValidEventValues.formattedValue,
```
damianOfKyan: this feels bit overwhelming, perhaps we could refactor that into separate helper function?

src/__experimental__/components/date/date.component.js
edleeks87: Yep agree I'll reformat the code 👍

src/__experimental__/components/date/date.component.js
edleeks87: Yeah I agree, I don't really want to rename it in the helper just in case end users have actually used it in the consumer apps, I will look to make it more readable in my implementation

src/__experimental__/components/date/date.spec.js
edleeks87: Yeah agreed, I'll make the change locally as I have other changes to make anyway

src/__experimental__/components/date/date.component.js
edleeks87: Yep agreed

src/__experimental__/components/date/date.component.js
kyanmichael: :white_check_mark:

src/__experimental__/components/date/date.component.js
kyanmichael: Thanks for refactoring with that new `isoFormattedValueString()` function :thumbsup:

src/__experimental__/components/date/date.spec.js
kyanmichael: :white_check_mark:

src/__experimental__/components/date/date.component.js
kyanmichael: Thanks for refactoring with that new `updateValidEventValues()` method :thumbsup:

src/__experimental__/components/date/date.component.js
```diff
@@ -110,48 +105,72 @@ class BaseDateInput extends React.Component {
...
     const visibleValue = formatDateToCurrentLocale(date);
+    const newDate = this.getDateObject(date);
 
     this.setState({
-      selectedDate: date,
-      visibleValue
+      selectedDate: newDate
+    },
+    () => {
+      this.updateValidEventValues(visibleValue);
```
kyanmichael: A couple of lines above this line, it updates the state – `this.setState({ selectedDate: newDate }` – and it passes a callback which waits for the state to update.

This `updateValidEventValues()` call also updates the state – `this.setState({ visibleValue, lastValidEventValues })` – but it doesn't pass a callback, so it doesn't wait for the state to update before proceeding to execute the rest of this function.

Does that matter? It may not matter, but I just wanted to check.

edleeks87: In this scenario it is fine, the callback is needed to ensure the selected date is updated, the rest of the stat update does not matter as it has updated accordingly by the time the `onBlur` triggers

kyanmichael: OK, thanks :ok_hand: :white_check_mark:

PR #2269, opened 2019-08-21
FE-2165 FE-1728 Radio button with validation

joshdj:
## Description

Wrapped the radio button component with the validation HOC.

#2278 to be merged in first

## Screenshots
Screenshot 2019-09-12 at 11 39 59

PR #2389, opened 2019-10-16
FE-2152 - [Radio Button] - should discard the name property

dawidzarzycki:
# Description
- Removed name prop from RadioButton. The prop has been replaced by groupName provided by RadioButtonGroup.

### DIFF
--- a/change_log/next/FE-2152-radio-button.yml
+++ b/change_log/next/FE-2152-radio-button.yml
@@ -1,0 +1,1 @@
+Breaking Changes: "Renamed `RadioButtonGroup` prop `groupName` to `name`(Component: RadioButtonGroup)"
--- a/src/__experimental__/components/radio-button/radio-button-group.component.js
+++ b/src/__experimental__/components/radio-button/radio-button-group.component.js
@@ -12,9 +12,8 @@ function checkedTabIndex(checked) {
 }
 
 const RadioButtonGroup = (props) => {
-  const { children, groupName, legend } = props;
+  const { children, name, legend } = props;
   const [selectedValue, setSelectedValue] = useState(null);
-
   const buttons = React.Children.map(children, (child, index) => {
     const isDefaultChecked = child.props.checked && !selectedValue;
     const checked = isDefaultChecked || selectedValue === child.props.value;
@@ -29,7 +28,7 @@ const RadioButtonGroup = (props) => {
       child,
       {
         checked,
-        inputName: groupName,
+        name,
         onChange: handleChange,
         tabindex
       }
@@ -47,7 +46,7 @@ RadioButtonGroup.propTypes = {
   /** The RadioButton objects to be rendered in the group */
   children: PropTypes.node.isRequired,
   /** Specifies the name prop to be applied to each button in the group */
-  groupName: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
   /** The content for the RadioGroup Legend */
   legend: PropTypes.string.isRequired
 };
--- a/src/__experimental__/components/radio-button/radio-button.component.js
+++ b/src/__experimental__/components/radio-button/radio-button.component.js
@@ -78,10 +78,7 @@ RadioButton.propTypes = {
   labelAlign: PropTypes.oneOf(OptionsHelper.alignBinary),
   /** Sets percentage-based label width */
   labelWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
-  /**
-   * The name of the group containing the RadioButton (can also be set via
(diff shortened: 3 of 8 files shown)


src/__experimental__/components/radio-button/radio-button-group.component.js
```diff
@@ -34,7 +36,8 @@ const RadioButtonGroup = (props) => {
         checked,
         inputName: groupName,
         onChange: handleChange,
-        tabindex
+        tabindex,
+        groupName
```
jamime: We don't need to pass down the `groupName` prop, its already passed as `inputName`

src/__experimental__/components/radio-button/radio-button.component.js
```diff
@@ -82,7 +82,15 @@ RadioButton.propTypes = {
    * The name of the group containing the RadioButton (can also be set via
    * the 'groupName' prop of the RadioButtonGroup component)
    */
```
mkrds: Not sure if that's enough, you can still pass the name prop to the RadioButton which gets passed on to CheckableInput even though it warns the user about not doing so

dawidzarzycki: Sure, you can pass the prop but you will get the console error that you shouldn't do it. The same situation is in other libraries

src/__experimental__/components/radio-button/radio-button-group.component.js
dawidzarzycki: Oh... how could I missed that one

src/__experimental__/components/radio-button/radio-button.component.js
jamime: I agree with @mkrds we should discard the property AND warn about it.

dawidzarzycki: I'm not big fan of the solution but I lost 2:1 :D 
Updated in https://github.com/Sage/carbon/pull/2389/commits/41792bfa5183738d2d45bf3ad66fe318edb8df1a

jamime: Yeah, delete is very slow. Can you not refactor it so we don't have three names for the same thing (`name`, `groupName`, `inputName`). Really we need to rename the `inputName` prop to be `name`?

mkrds: Not a fan of that solution either, but I guess that thanks to that we can free users of some hard to debug bugs. Anyway I'd rather not mutate the prop object and insted of deleting one property I'd just desctructure it like that:
`
const RadioButton = ({
  id, label, onChange, value, name, ...props
})
`

jamime: @mkrds that fails eslint rule, `no-unused-variables`. Why do you not want to rename `inputName`?
The intended interface is `<RadioGroup name="foobar"><RadioButton /></RadioGroup>`

change_log/next/FE-2152-radio-button.yml
```diff
@@ -0,0 +1 @@
+New Features: "Removed name prop from RadioButton. The prop has been replaced by groupName provided by RadioButtonGroup. (Component: RadioButton, RadioButtonGroup)"
```
jamime:
```suggestion
Breaking Changes: "Renamed `RadioButtonGroup` prop `groupName` to `name`(Component: RadioButtonGroup)"
```

src/__experimental__/components/radio-button/radio-button-group.component.js
```diff
@@ -47,9 +46,10 @@ RadioButtonGroup.propTypes = {
   /** The RadioButton objects to be rendered in the group */
   children: PropTypes.node.isRequired,
   /** Specifies the name prop to be applied to each button in the group */
-  groupName: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
   /** The content for the RadioGroup Legend */
   legend: PropTypes.string.isRequired
+
```
jamime:
```suggestion

```

jamime:
@edleeks87 It's because that branch that you used still used `groupName`. I've checked and I am happy with these changes. @dawidzarzycki please can you squash all of these changes into a single commit and rebase off master.

jamime:
I squashed the changes and assigned to @grabkowski to merge after https://github.com/Sage/carbon/pull/2269 has been merged.

dawidzarzycki:
> I squashed the changes and assigned to @grabkowski to merge after #2269 has been merged.

The squash looks good to me

jamime:
Closing this is now included in PR https://github.com/Sage/carbon/pull/2269

PR #2391, opened 2019-10-16
cypress/FE-2116 deprecated components

DlgSHi:
# Description
added new features for deprecated components


cypress/features/regression/deprecated/form.feature
```diff
@@ -1,32 +1,37 @@
-Feature: Form component
+Feature: Deprecated Form component
   I want to change Form component properties
```
KatarzynaQA: Please, add info that is 'deprecated'

DlgSHi: fixed

PR #2396, opened 2019-10-17
FE-2161 Switch does not render the the correct state when uncontrolled

mkrds:
# Description
Fixed an issue with switch component not rendering properly when in an uncontrolled state.

# UX Notes
Also fixed FE-2193, previously Switch has been only disabled while loading prop was passed as true only if classic theme has been selected. Fixed it so no matter which theme, loading prop forces disabled state of the component


src/__experimental__/components/switch/switch.component.js
```diff
@@ -8,20 +8,35 @@ import { isClassic } from '../../../utils/helpers/style-helper';
...
 }) => {
   const classicDisabled = props.theme && isClassic(props.theme) && props.loading;
 
+  const [isChecked, setIsChecked] = useState(checked || false);
+
+  useLayoutEffect(() => {
+    if (checked !== undefined && checked !== isChecked) {
+      setIsChecked(checked);
+    }
+  }, [checked]);
```
jamime:
```suggestion
  }, [checked, setIsChecked]);
```

jamime: Can you explain why we need `useLayoutEffect` and not `useEffect`?

src/__experimental__/components/switch/switch.spec.js
```diff
@@ -33,6 +35,62 @@ function render(props, renderer = TestRenderer.create) {
 }
 
 describe('Switch', () => {
+  describe('uncontrolled behaviour', () => {
+    it('sets proper default internal state', () => {
+      const wrapper = render({ defaultChecked: true }, mount);
+      expect(wrapper.find(CheckableInput).prop('checked')).toBe(true);
+    });
+
+    it('changes internal state and passess event to the provided onChange prop when change is triggered', () => {
```
ianforster-sage:
```suggestion
    it('changes internal state and passes event to the provided onChange prop when change is triggered', () => {
```

src/__experimental__/components/switch/switch.component.js
```diff
@@ -1,28 +1,49 @@
...
+  loading,
+  theme,
+  ...props
 }) => {
-  const classicDisabled = props.theme && isClassic(props.theme) && props.loading;
+  const isControlled = checked !== undefined;
+
+  const [checkedInternal, setCheckedInternal] = useState(defaultChecked || false);
+
+  const onChangeInternal = (e) => {
```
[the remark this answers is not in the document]
mkrds: Done

PR #2405, opened 2019-10-21
FE-2175 GroupedCharacter - added uncontrolled support

mkrds:
# Description
Added support for uncontrolled use of GroupedCharacter component.


src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -27,7 +39,10 @@ const GroupedCharacter = (props) => {
 
     const reProcessedInputValue = ev.target.value.split(separator).join('');
 
-    props.onChange({ target: { value: reProcessedInputValue } });
+    onChange({ target: { value: reProcessedInputValue } });
```
jamime: We need to include the whole `event` in the callback e.g
```js
const target = {
  ...ev.target,
value: reProcessedInputValue
}

ev.target = target;
onChange(ev);
```

mkrds: How about:
```js
const event= {
  ...ev,
  target: {
    ...ev.target,
    value: reProcessedInputValue
  }
}
onChange(event);
```
No mutations and imo looks more readable.

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -27,7 +39,10 @@ const GroupedCharacter = (props) => {
 
     const reProcessedInputValue = ev.target.value.split(separator).join('');
 
-    props.onChange({ target: { value: reProcessedInputValue } });
+    onChange(ev, reProcessedInputValue);
+    if (!isControlled) {
+      setInternalValue(reProcessedInputValue);
+    }
     setTimeout(() => eventRef.setSelectionRange(newCursorPos, newCursorPos));
```
jamime: `eventRef` is not defined.

src/__experimental__/components/grouped-character/grouped-character.component.js
```diff
@@ -4,12 +4,24 @@ import Textbox from '../textbox';
...
-const GroupedCharacter = (props) => {
+const GroupedCharacter = ({
+  groups,
+  value: externalValue,
+  defaultValue,
+  onChange,
+  separator: rawSeparator,
+  ...rest
+}) => {
   const [pressedKey, updatePressedKey] = useState(null);
```
jamime: This state is not used

change_log/next/FE-2175-grouped-character-uncontrolled-support.yml
```diff
@@ -0,0 +1,2 @@
+Bug Fixes: "Fixed issue where component did not supporter uncontrolled use. (Components: GroupedCharacter)"
```
damianOfKyan: Bug Fixes: "Fixed issue where component did not support uncontrolled use. (Components: GroupedCharacter)"

PR #2408, opened 2019-10-22
FE-2167 Decimal does not implement onBlur

kyanmichael:
# Description

Implemented a new `onBlur()` prop for the Experimental Decimal component. The callback receives two arguments:

1. The React `SyntheticEvent` object for the `blur` event.
2. The undelimited value (as a string). For example, if the component's displayed value is `"1,000.0"`, then the undelimited value is `"1000.0"`.


jamime:
I rebased off master after merging FE-2170

PR #2410, opened 2019-10-22
Fe 2169 decimal precision extra

damianOfKyan:
# Description

Decimal Input will now accept numbers beyond precision value.
Decimal Input will now always position a separator equal to precision calculating from right to left.


jamime:
Currently writing RFC to discuss Decimal formatting.

jamime:
Closed, more details about long term goal in JIRA.

PR #2415, opened 2019-10-25
[RFC] onChange & onBlur standardisation

jamime:
This RFC describes the proposed changes to the `onChange` and `onBlur` interface. 

---
This is our first RFC and it is important to note that this PR does not follow the typical PR process. Details of the RFC process can be found [here](https://github.com/Sage/carbon/blob/2163a70006e4e7853d6b4fe05703a2348a979f3e/rfcs/README.md).

Please use 👍 if you approve.


rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+      ]
+    };
+
+    onChange(e);
+  };
+```
+
+# Motivation
+
+Currently the `onChange` and `onBlur` handers in Carbon use a mix of custom objects `{value: "my-value"}` and React's `SyntheticEvent`.
```
kyanmichael:
```suggestion
Currently the `onChange` and `onBlur` handlers in Carbon use a mix of custom objects `{value: "my-value"}` and React's `SyntheticEvent`.
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+
+In the custom objects `name` or `id` are passed inconsistently.
```
kyanmichael:
```suggestion
In the custom objects, `name` or `id` are passed inconsistently.
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+
+The purpose of this RFC is to standardise the approach so users can rely upon the interface between Carbon their application.
```
kyanmichael:
```suggestion
The purpose of this RFC is to standardise the approach, so users can rely upon the interface between Carbon and their application.
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+The approach that we choose should be consistent across components so users familiar with the interface of one component can easily use another component.
+
+# Detailed design
+
+Users who implement this should be familiar with React's [SyntheticEvent](https://reactjs.org/docs/events.html).
+
+When writing a Carbon component the user should wrap the user provided `onChange`/`onBlur` callback with code similar to that in
+the basic example.
+
+* Components **MUST** replace `SyntheticEvent.target` to steer users from upon the component implementation details.
```
kyanmichael:
```suggestion
* Components **MUST** replace `SyntheticEvent.target`, to steer users away from depending upon the component's implementation details.
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+* `SyntheticEvent.target` **MUST** be an object with the properties `name`, `id`, `value`.
+* `SyntheticEvent.target` **MUST** include the `name` and `id` of the component.
```
kyanmichael:
```suggestion
* `SyntheticEvent.target` **MUST** include the `name` and `id` of the component (if the component has these props).
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+```js
```
kyanmichael: We'll get better syntax-highlighting if we use `jsx` here:
```suggestion
```jsx
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+<Select id="foo" name="foo">
+  <Option id="bar" name="bar" value="bar">Bar</Option>
+</Select>
+```
+> In the example above the `SyntheticEvent.target` **MAY** be the follow `{"id": "foo", "name":"foo", "value": {"optionValue": "bar", "optionText": "Bar"}}`
```
kyanmichael:
```suggestion
> In the example above, the `SyntheticEvent.target` **MAY** be the following object: `{"id": "foo", "name": "foo", "value": {"optionValue": "bar", "optionText": "Bar"}}`
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+
+* `SyntheticEvent.target` **MUST** have a value property, that property can be any value **UNLESS** the component supports
```
kyanmichael:
```suggestion
* `SyntheticEvent.target` **MUST** have a `value` property, which can be any value, **UNLESS** the component supports
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+multiple select, in that case the property must be an `Array` of values. If the component supports both single select
```
kyanmichael:
```suggestion
multiple select, in which case the property must be an `Array` of values. If the component supports both single select
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+and multiple select controlled by a prop the property **MUST** be an `Array`.
```
kyanmichael:
```suggestion
and multiple select (controlled by a separate prop), the `value` prop **MUST** be an `Array`.
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+* Components **MUST NOT** iterate over or replace props on the original `SyntheticEvent.target`
+* Components **MAY** reference values from the original `SyntheticEvent.target` or use `SyntheticEvent.nativeEvent.target` as they
+are synonymous.
+
+The approach suggested would modify the `SyntheticEvent` object that React's `EventPool` uses for performance optimisation.
+
+When validating this approach it was important to ensure that we didn't adversely affect the `EventPool`.
+
+* `SyntheticEvent` is constructed from a object in the `pool` [SyntheticEvent.js#L94](https://github.com/facebook/react/blob/ab1a4f249e61045d523ddbbfb840e868afbbf785/packages/react-dom/src/events/ChangeEventPlugin.js#L53)
```
kyanmichael: I think it'd be clearer if we use a tag (`v16.8.2`) instead of a commit hash for all these URLs:
```suggestion
* `SyntheticEvent` is constructed from a object in the `pool` [SyntheticEvent.js#L94](https://github.com/facebook/react/blob/v16.8.2/packages/react-dom/src/events/ChangeEventPlugin.js#L53)
```

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+* `SyntheticEvent.target` is a pointer to `SyntheticEvent.nativeEvent.target` [SyntheticEvent.js#L94](
+https://github.com/facebook/react/blob/b1a03dfdc8e42d075422556553ffe59868150e95/packages/legacy-events/SyntheticEvent.js#L94)
+
+* React "nullifies" the event properties when released back to the pool [EventBatching.js#L34](https://github.com/facebook/react/blob/b1a03dfdc8e42d075422556553ffe59868150e95/packages/legacy-events/EventBatching.js#L34) => [SyntheticEvent.js#L179](https://github.com/facebook/react/blob/b1a03dfdc8e42d075422556553ffe59868150e95/packages/legacy-events/SyntheticEvent.js#L179)
+
+I could see no references to `.target` in the [legacy-events](https://github.com/facebook/react/search?q=target+path%3Apackages%2Flegacy-events&unscoped_q=target+path%3Apackages%2Flegacy-events) package.
+
+When validating this approach I reviewed the code of another popular component library, MaterialUI. [This is the same
+approach that they have adopted](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Select/SelectInput.js#L130).
```
chrisbarber86: Might want to link to a tag since `master` may change.

rfcs/text/onChange.md
```diff
@@ -0,0 +1,128 @@
...
+    };
+    }
+  }
+  onChange(e);
+  };
+```
+
+# Drawbacks
+
+- It would be a breaking change and therefore would require a major version bump.
```
chrisbarber86: I'm assuming the suggestion here is to make this change as part of v9?

jamime: That's right. Tickets that are currently in progress have already started to adopt this process. We would need to have a cursory check of the other `_experimental_` components. If there are any changes required to `src/components` I propose we make those at a later date.

jamime: That's right. Tickets that are currently in progress have already started to adopt this proposal. I think we should have a cursory check of everything in `_experimental_`. We can do this using the `fixture` screens that already exist. I suggest that any changes to components in `src/components` are made at a later date.

jamime:
I've made a small modification in 316b505 to relax the rules. Previously it was suggested that we need to override all events. In reality, this is only needed when a component formats the users' input or returns multiple values.

jamime:
I plan to transition to final comments at the end of today.

rfcs/text/onChange.md
```diff
@@ -0,0 +1,131 @@
...
+Users who implement this should be familiar with React's [SyntheticEvent](https://reactjs.org/docs/events.html).
+
+When writing a Carbon component the user should wrap the user provided `onChange`/`onBlur` callback with code similar to that in
+the basic example.
+
+* Components that provide a formatted value **MUST** replace `SyntheticEvent.target`.
+* `SyntheticEvent.target` **MUST** be an object with the properties `name`, `id`, `value`.
+* `SyntheticEvent.target` **MUST** include the `name` and `id` of the component (if the component has these props).
+```jsx
+<Select id="foo" name="foo">
```
philearley: It would be great, to have values that are more in line with how the component could be used. 

foo, bar, baz has caused us so many problems when visiting a file written by someone else and you want to use it or make changes. 

Using something like, `name='Visible in the component'` and `value='id of visible value'`, make it easier to follow or maybe supplying context with, "when I have a set of data with, `{ key1: 'value1', key2: 'value2' }` and showing where those values would be used."

jamime: I've addressed this comment by giving a more realistic example in 4b01825.

jamime:
This RFC has now transitioned to `Final Comments` unless there are significant changes. This will be merged on Thursday evening.

rfcs/text/onChange.md
```diff
@@ -0,0 +1,132 @@
...
+
+When validating this approach I reviewed the code of another popular component library, MaterialUI. [This is the same
+approach that they have adopted](https://github.com/mui-org/material-ui/blob/v4.5.1/packages/material-ui/src/Select/SelectInput.js#L130).
+
+## Other considerations
+
+1. We currently have components that are controlled by a non-input field, for example `Date`. This is where a 3rd party date
+picker changes the underlying `value` of the `<input>`. There is an [outstanding bug in React](https://github.com/facebook/react/issues/13424) which means that `onChange` events attached to `<input>` won't trigger when changing the `value` with `setState`.  
+For this reason, when programmatically setting `value` the event **MUST** be a plain object and not a `SyntheticEvent`.
+It **MUST* follow all the other requirements above.
```
kyanmichael:
```suggestion
It **MUST** follow all the other requirements above.  
```

PR #2422, opened 2019-10-29
FE-2209/2210/2211/2272 Select component (onChange, value, defaultValue, uncontrolled mode)

kyanmichael:
# Description

This PR implements the following changes:

### General

* A new optional `enableMultiSelect` boolean prop has been implemented. When set to `true`, the `<Select>` component will operate in multi-select mode. When set to `false` (or omitted), the component will operate in single-select mode.
* The [`invariant`](https://www.npmjs.com/package/invariant) library has been added as a `dependency` in `package.json`. Note: `invariant` already exists in `package-lock.json` (since it's a subdependency of other existing dependencies such as `react-dnd`), so adding it to `package.json` doesn't cause any new packages to be added to `/node_modules/`.

### FE-2209
* In accordance with the RFC in PR #2415, the event object passed to the `onChange()` callback prop has been modified:
  * `name` and `id` have been added to `e.target`.
  * `e.target.value` is now an array of objects, where each object has this structure: `{ optionValue: "...", optionText: "..." }`
  * If the component is in single-select mode, then the `e.target.value` array has 1 element.
* Since `onChange()` can be triggered by various different DOM elements and user events (`click` event from `<Option>` element, `click` event from `<button>` of `<Pill>` element, `keydown` event from `<input>` element, etc), it doesn't make sense to expose any of those `SyntheticEvent` objects through the `onChange()` callback (since they represent private/internal/low-level implementation details of the `<Select>` component). Therefore, a custom-created event object is passed to the `onChange()` callback.

### FE-2210

* When the `<Select>` component is operating in controlled mode, the `value` prop is currently required to be an object with this structure: `{ value: "...", text: "..." }` (or an array of such objects, for multi-select mode).
* This PR modifies the `value` prop, so it's now required to be a string (or an array of strings, for multi-select mode), where each string matches a `value` in an `<Option value="...">` child of the `<Select>` component.

### FE-2211

* The `<Select>` component can now operate in uncontrolled mode (without a `value` prop).
* A new `defaultValue` prop has been implemented, to configure the initially-selected option when the component is operating in uncontrolled mode.

### FE-2272

* The documentation comments of the `propTypes` of the `<Option>` component have been improved.


src/__experimental__/components/select/select.component.js
```diff
@@ -70,17 +88,100 @@ class Select extends React.Component {
...
+      return this.props.value;
+    }
+    return this.state.value;
+  }
+
+  /**
+   * Gets this component's current single-select value.
+   * Should only be called when the component is in single-select mode.
+   */
+  getSingleSelectValue = () => {
```
mkrds: This method is not used anywhere

kyanmichael: True! Removed in f1b8abe :thumbsup:

src/__experimental__/components/select/select.component.js
```diff
@@ -70,17 +88,85 @@ class Select extends React.Component {
...
   handleClick = () => this.openList()
 
-  handleChange = (newValue) => {
-    let updatedValue = newValue;
+  /**
+   * Is this component currently controlled (by the `value` prop)?
+   */
+  isComponentControlled() {
+    return (this.props.value !== undefined && this.props.value !== null);
+  }
```
grabkowski: do we need to assume that this component could change from uncontrolled to controlled?
couldn't it be just `isComponentControlled = this.props.value !== undefined && this.props.value !== null;`?

src/__experimental__/components/select/select.component.js
```diff
@@ -41,6 +53,12 @@ class Select extends React.Component {
 
   input = {} // this will store a reference to the input html element
```
mkrds: Is the `componentWillReceiveProps` necessary in here? Only case when this `nextProps.value !== this.props.value` condition is met is when component is used in controlled way. And according to the `getValue` method it seems that `this.state.value` is not used in that case. Also I think we should use `UNSAFE_componentWillReceiveProps` because some of old lifecycle methods are deprecated. Or better dont use deprecated lifecycle methods at all.

src/__experimental__/components/select/select.component.js
```diff
@@ -41,6 +53,12 @@ class Select extends React.Component {
...
 
+  componentWillReceiveProps(nextProps) {
```
mkrds: Is the `componentWillReceiveProps` necessary in here? Only case when this `nextProps.value !== this.props.value` condition is met is when component is used in controlled way. And according to the `getValue` method it seems that `this.state.value` is not used in that case. Also I think we should use `UNSAFE_componentWillReceiveProps` because some of old lifecycle methods are deprecated. Or better dont use deprecated lifecycle methods at all.

src/__experimental__/components/select/select.component.js
kyanmichael: You're correct. In fact @edleeks87 created FE-2262 to cover this for Carbon in general. I've now modified `isComponentControlled` in 855908d :thumbsup:

src/__experimental__/components/select/select.component.js
kyanmichael: Good point. `componentWillReceiveProps()` isn't necessary. As you said, `state.value` is only used in `getValue()`, and only when the component is uncontrolled. I've removed `componentWillReceiveProps()` in 68512b0 :thumbsup:

src/__experimental__/components/select/select.component.js
```diff
@@ -32,7 +33,20 @@ class Select extends React.Component {
     this.listboxId = uniqueId('listbox-');
   }
```
mkrds: Not a blocking comment but I think that if we already define state and isComponentControlled class fields outside of constructor then listboxId could be initialized in the same way so we would get rid of constructor. Or the other way around, define state and isComponentControlled inside of a constructor for consistency.

kyanmichael: Good point. Nice to remove the constructor! Refactored in cc68577 :thumbsup:

kyanmichael:
:eight_pointed_black_star: When merging this PR, use the **"Squash and merge"** option :eight_pointed_black_star:

sianford:
Tested the Controlled and Uncontrolled Single-Select and Multi-Select components in the Carbon Fixture Testbed. All tests passed ✅ Outstanding issues with the onBlur event output object structure and "value" content with Multi-Select that are out of scope of these tickets will be fixed in ticket FE-2307.

PR #2428, opened 2019-10-30
feat: update decimal onChange and onBlur interface FE-2238

jamime:
# Description
Update the Decimal component to follow RFC guidelines.

# Testing Instructions
You can test this component in storybook or by running the fixture screens.


src/utils/helpers/createEvent/index.spec.js
```diff
@@ -0,0 +1,89 @@
...
+  name: 'username',
+  value: 'CarbonUser'
+};
+
+class Event {
+  constructor(target = mockInputTarget) {
+    this.target = target;
+  }
+}
+
```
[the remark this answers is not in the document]
jamime: `describe` is used to group a similar set of tests together and allows you to leverage `beforeEach` in this case there is no code shared between tests. The only benefit would be the jest report format, but it would increase the indentation of the entire file. Pros&Cons really. Happy to change it if you would like?

src/utils/helpers/createEvent/index.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Override a SyntheticEvent with a custom event.target
+ * This is intended to be used by Carbon components that provide a formatted value or those that need
+ * to provide a name or value that does no appear in the DOM e.g. a component that has two controls associated with one
+ * input.
+ */
+export default (event, overrides) => {
```
grabkowski: Could you change that helper to a named function? When it's anonymous it's difficult to find in an IDE, you also cannot auto import it in VSCode.

src/utils/helpers/createEvent/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+  const { name = event.target.name, id = event.target.id, value = event.target.value } = overrides;
+
+  event.target = {
+    ...(value !== undefined && { value }),
+    ...(name && { name }),
+    ...(id && { id })
+  };
+
+  return event;
```
grabkowski: I think that could be:
```
  return { target: { ...event.target, ...overrides } };
```
it will replace only defined values, or I'm missing something?

jamime: We only want `name`, `id` and `value` but we don't want the properties defined in the object if the values are undefined.

src/utils/helpers/createEvent/index.js
```diff
@@ -0,0 +1,25 @@
...
+};
```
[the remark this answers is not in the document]
grabkowski: we could create `index.js` file in the `helpers` folder and then export every helper in that file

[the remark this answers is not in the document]
mkrds: That's actually a good comment. We could have two files like `createEvent.js` and `index.js` in this `createEvent` folder (I think it's better than using `package.json`), where `index.js` purpose is only to export the `createEvent` so we have shorter import path.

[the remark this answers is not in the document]
mkrds: Or we could have an `index.js` file in `src\utils\helpers` so we could have named exports and import helpers like `import { SomeHelper } from `src\utils\helpers`. But that would require some folders restructuring.

src/utils/helpers/createEvent/index.js
grabkowski: then we could just delete them:

```javascript
const target = { ...event.target, ...overrides };

Object.keys(target).forEach(key => {
  if (target[key] === undefined) {
    delete target[key];
  }
});

return { target };
```

src/utils/helpers/createEvent/index.js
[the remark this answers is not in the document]
jamime: I've created the file `src/utils/helpers/index.js` so you can now `import { createEvent } from `/src/utils/helpers'`

src/utils/helpers/createEvent/index.js
jamime: Deleting Object properties is still a slow operation to discourage this we explicitly said not to iterate over `event.target` in the RFC (so spreading `event.target` is prohibited). I'm going to dismiss this comment. If the solution becomes an issue we will revisit it at a later date.

jamime: Deleting Object properties is still a slow operation to discourage this we explicitly said not to iterate over event.target in the RFC (so spreading event.target is prohibited). I'm going to dismiss this comment. If the solution becomes an issue we will revisit it at a later date

jamime:
This feature is included in https://github.com/Sage/carbon/pull/2513

PR #2429, opened 2019-10-30
FE-1987 - Remove Large Theme

chrisbarber86:
# Description

Breaking Change: Removes the Large Theme as an option

# TODO
- [x] Release notes

# Screenshots / Gifs

# Testing Instructions


change_log/next/FE-1987-remove-large-theme.yml
```diff
@@ -0,0 +1 @@
+Breaking Changes: "Removes the large theme."
```
jamime: Is it a `Breaking Change` if large was never released?

chrisbarber86: Good point - I wasn't sure if themes were added into the experimental stuff prior to v9 or not. If it's only v9 then technically my release notes for this PR should be deleting any mention of a `large` theme right?

jamime: Yeah, I'm not sure when `large` was introduced. I was under the assumption that v8 had no knowledge of themes. Yes, we should probably remove renogen entries that mention `large` too.

PR #2431, opened 2019-10-31
FE-2248/2249 CheckboxGroup (interface, uncontrolled operation)

damianOfKyan:
# Description

RadioButton now implements same interface as RadioGroup and groupName
Checkbox Group fix to work uncontrolled


src/__experimental__/components/checkbox/__snapshots__/checkbox-group.spec.js.snap
```diff
@@ -250,14 +254,16 @@ exports[`CheckboxGroup renders as expected 1`] = `
       </div>
     </div>
     <div
+      checked={false}
       className="c5 c6"
       data-component="checkbox"
-      name="check-optional"
+      name="my-checkbox-group"
```
ianforster-sage: We shouldn't be setting the `name` attribute of a `<div>`. Name should only be an attribute of a form input element 
https://www.w3schools.com/tags/att_name.asp

damianOfKyan: resolved in 007bc48b109edd5a4bea662ebcd953f4253334b3

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -7,26 +7,54 @@ import FormField from '../form-field';
...
+    if (checkedIndex === -1) {
+      checkedCheckboxes.push(e.target.value);
+    } else {
+      checkedCheckboxes.splice(checkedIndex, 1);
+    }
+
+    setCheckedValue(checkedCheckboxes);
+
+    const name = (id && { id });
+
```
mkrds: Why is this prop named `id` instead of `name` ?

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -7,26 +7,54 @@ import FormField from '../form-field';
...
+      ...name,
+      value: checkedCheckboxes
     };
 
+    onChange(e);
+  }, [onChange, checkedValue]);
+
+  const buttons = React.Children.map(children, (child) => {
+    const checked = (checkedValue.indexOf(child.props.value) !== -1);
+
```
mkrds: This solution does not support a `defaultChecked` prop passed to the one of elements in case of uncontrolled use of `CheckboxGroup`

damianOfKyan: added `defaultValue` prop.

src/__experimental__/components/checkbox/checkbox-group.component.js
damianOfKyan: this has now been removed as after updates it became obsolete

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -7,26 +7,53 @@ import FormField from '../form-field';
...
+    if (checkedIndex === -1) {
+      checkedCheckboxes.push(e.target.value);
+    } else {
+      checkedCheckboxes.splice(checkedIndex, 1);
+    }
+
+    setCheckedValue(checkedCheckboxes);
+
+    e.target = {
+      id,
```
jamime: I need to think about this more, but we've come to the consensus that we need to provide `name` in all events. This is complicated by the example above when there are two children with different names.

I think it is best to remove the `onChange` functionality from the `CheckboxGroup` component. We may re-introduce it for controlled components when we introduce `formik`. But for now, because the event object is inconsistent and I can't think of a way of determining the correct name. Let's remove the `onChange` helper from `CheckboxGroup`. This is not to say that this component does not have a `onChange` applied to the cloned child - we just don't want to call the user provided `onChange`.

I think having `onBlur` on the `CheckboxGroup` will be confusing, does it blur when all of them lose focus, what if you select one checkbox then another. To prevent it from being ambiguous, let's remove this.

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -7,26 +7,53 @@ import FormField from '../form-field';
...
+    onChange(e);
+  }, [onChange, checkedValue]);
+
+  const buttons = React.Children.map(children, (child) => {
+    const checked = (checkedValue.indexOf(child.props.value) !== -1);
+
     let childProps = {
-      inputName: groupName,
-      onChange: handleChange
+      name: child.props.name,
```
jamime:
```suggestion
```
This line does not do anything because we're accessing the prop from same child that we're re-applying it to.

jamime:
**Summary of Zoom on 15th Nov 10am GMT**
**Attendees**: @jamime, @damianOfKyan & @kyanmichael 
1. `CheckboxGroup` will not support uncontrolled validations.
1. `CheckboxGroup` will support controlled validations (where the user passes the `value` prop)
1. `CheckboxGroup` will not support user-defined `onBlur` or `onChange` until after we have added support for `formik`
1. `CheckboxGroup` will not use the `name` or `groupName` prop.
1. `CheckBox` will support being controlled and uncontrolled. 
At this time, uncontrolled means not passing a value. In the future, we should ensure that we pass refs to the appropriate child element too.

mkrds:
Controlled CheckboxGroup fixtures are not working.

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -2,49 +2,50 @@ import React from 'react';
...
-  const groupLabelId = `${groupName}-label`;
+  const defaultCheckedValue = (defaultValue !== undefined) ? defaultValue : [];
+  const checkboxGroupValue = (value !== undefined) ? value : defaultCheckedValue;
+  const groupLabelId = `${id}-label`;
 
   const buttons = React.Children.map(children, (child) => {
+    let checked = false;
+    let defaultChecked = false;
+
+    if (checkboxGroupValue.length > 0) {
```
jamime: We don't want to support a controlled checkbox-group at this time. We want users to control each checkbox instead. Please can you remove the code relating to`defaultChecked` and `value`

jamime: I think we can remove the `React.Children.map` entirely.

src/__experimental__/components/checkbox/checkbox-group.component.js
```diff
@@ -2,49 +2,50 @@ import React from 'react';
...
   });
 
   return (
     <StyledCheckboxGroup
       aria-labelledby={ groupLabelId }
       role='checkbox'
-      hasError={ hasError }
-      hasWarning={ hasWarning }
-      hasInfo={ hasInfo }
+      value={ checkboxGroupValue }
```
jamime: Why does the styled component need the value?

fixtures/src/components/uncontrolled/checkbox-group.js
```diff
@@ -46,25 +51,30 @@ const UncontrolledCheckbox = () => {
                 <li>value two is checked by default</li>
               </ul>
               <CheckboxGroup
-                name='uncontrolled_checkbox_group_default' label='Uncontrolled Example default'
-                onChange={ onChange }
-                onBlur={ onBlur }
+                id='uncontrolled_checkbox_group_default'
+                label='Uncontrolled Example default'
+                defaultValue={ ['value one'] }
```
mkrds: is having `defaultValue` appliable to `CheckboxGroup` is still really needed? Since we're applying directly `onChange` and `onBlur` on `Checkboxes` then why not just apply `defaultChecked` to the wanted `Checkboxes` ? Also it would mean that  mapping children in `CheckboxGroup` is not needed anymore.

src/__experimental__/components/checkbox/checkbox-group.component.js
kyanmichael: I don't think it does. `<StyledCheckboxGroup>` in `src/__experimental__/components/checkbox/checkbox.style.js` doesn't seem to use `value`. Removed in d418bfc

src/__experimental__/components/checkbox/checkbox-group.component.js
kyanmichael: Removed in 2c1a6c0 :thumbsup:

> We don't want to support a controlled checkbox-group at this time.

So does [`fixtures/src/components/controlled/checkbox-group.js`](https://github.com/Sage/carbon/blob/FE-2248-checkbox-group-interface/fixtures/src/components/controlled/checkbox-group.js) need to be deleted then?

fixtures/src/components/uncontrolled/checkbox-group.js
kyanmichael: I agree :thumbsup: Your comment aligns with @jamime's [comment](https://github.com/Sage/carbon/pull/2431#pullrequestreview-322898567) too:

> We don't want to support a controlled checkbox-group at this time. We want users to control each checkbox instead. Please can you remove the code relating to `defaultChecked` and `value`. I think we can remove the `React.Children.map` entirely.

So I definitely think the `defaultValue` prop needs to be removed from `<CheckboxGroup>` here :thumbsup:

@jamime: Once I've removed it, then what should we do about [the **"Default Checked"** section of the **"Uncontrolled Checkbox Group"** fixture](https://github.com/Sage/carbon/blob/47483ff/fixtures/src/components/uncontrolled/checkbox-group.js#L48-L80)? I don't think that section is needed, because the `<CheckboxGroup>` will not be supporting "default checked" functionality itself. That functionality is being provided/demonstrated by [the **"Default Checked"** section of the **"Uncontrolled Checkbox"** fixture](https://github.com/Sage/carbon/blob/f5d07c3/fixtures/src/components/uncontrolled/checkbox.js#L30-L44) for individual `<Checkbox>` components.

jamime: agreed, you can remove it.

kyanmichael: @jamime: Thanks, removed in 7bb7919 :thumbsup:

src/__experimental__/components/checkbox/checkbox-group.component.js
jamime: We can either keep it to show how to use a controlled checkbox in a checkbox-group.
OR
We can update `fixtures/src/components/controlled/checkbox.js ` to show with and without a group.

kyanmichael: Of those two options, I think this one seems better:

> We can update `fixtures/src/components/controlled/checkbox.js ` to show with and without a group.

I'll do that, and I'll also delete `fixtures/src/components/controlled/checkbox-group.js`.

kyanmichael: `fixtures/src/components/controlled/checkbox-group.js` removed in 99ebdf3 and ca02d94

CheckboxGroup example added to `fixtures/src/components/controlled/checkbox.js` in 5aee4e4 :thumbsup:

src/__experimental__/components/checkbox/checkbox.component.js
```diff
@@ -1,31 +1,45 @@
...
-import withValidations from '../../../components/validations/with-validation.hoc';
+import withValidation from '../../../components/validations/with-validation.hoc';
 
 const Checkbox = ({
-  id, label, onChange, value, ...props
+  id, name, label, value, onChange, defaultChecked, checked, ...props
 }) => {
+  const checkedValue = checked === undefined ? false : checked;
+  const checkedStatus = (checked === undefined && defaultChecked !== undefined) ? defaultChecked : checkedValue;
+
```
mkrds: This unfortunately doesn't work 100% as a controlled component, because it does not react to the `checked` prop changes. You can check it out by having two checkboxes linked to one state field. Clicking on one, should also automatically check the other one, but it does not happen.

kyanmichael: Many thanks for spotting that bug. Yes you're correct, `setIsChecked()` is currently only called by `handleOnChange()` (when the user toggles the checkbox). But `setIsChecked()` also needs to be called when the `checked` prop changes. Fixed in 3bcbdc2 :thumbsup:

mkrds: No worries :)

sianford:
More info is required about how to test these changes correctly.

jamime:
Can we get a `BREAKING CHANGE` release note added. We removed support for `groupName`.

PR #2437, opened 2019-11-01
[RFC] Proposal to add react-table as a dependency

Chris3y:
# Description
Recommended as basis for a grid component


ianforster-sage:
Closing as @Chris3y no longer wants this proposal to be considered.

PR #2443, opened 2019-11-04
Cypress/fe 2255 decimal fieldset

DlgSHi:
# Description
1. Decimal input
classic
2. Fieldset
default
classic


cypress/features/regression/experimental/decimalInputClassic.feature
```diff
@@ -0,0 +1,136 @@
...
+    When I select align to "<direction>"
+    Then input direction is "<direction>"
+    Examples:
+      | direction |
+      | left      |
+      | right     |
+
+  @positive
+  Scenario Outline: Change Decimal component label help to <label>
+    When I set label to "<label>"
```
KatarzynaQA: Maybe we should use 'Given' instead of 'When' ? I think that 'I hover mouse onto help icon' step is the proper action step before a result

cypress/features/regression/experimental/decimalInputClassic.feature
```diff
@@ -0,0 +1,136 @@
...
+      | <>                       |
+
+  @positive
+  Scenario Outline: Change Decimal component label inline
+    When I set label to "<label>"
+      And I check labelInline checkbox
+    Then label is set to inline
+    Examples:
+      | label                    |
+      | Sample text              |
```
KatarzynaQA: We have only one example here, so  better is use 'scenario' instead of 'scenario outline'

cypress/features/regression/experimental/decimalInputClassic.feature
```diff
@@ -0,0 +1,136 @@
...
+
+  @positive
+  Scenario Outline: Change Decimal component label width to <width>
+    When I set label to "<label>"
```
KatarzynaQA: I think better to use here 'Given'. 'Then' should be used for 'And I set label width slider to <width>'

cypress/features/regression/experimental/decimalInputClassic.feature
```diff
@@ -0,0 +1,136 @@
...
+    Examples:
+      | label       | width |
+      | Sample text |  0    |
+      | Sample text |  10   |
+      | Sample text |  50   |
+      | Sample text |  100  |
+
+  @positive
+  Scenario Outline: Change Decimal component input width to <width>
+    When I set label to "label"
```
KatarzynaQA: The same like above

cypress/features/regression/experimental/decimalInputClassic.feature
```diff
@@ -0,0 +1,136 @@
...
+    Examples:
+      | width |
+      |  1    |
+      |  10   |
+      |  50   |
+      |  100  |
+
+  @positive
+  Scenario Outline: Change Decimal component label align to <labelAlign>
+    When I set label to "label"
```
KatarzynaQA: The same like above

cypress/features/regression/experimental/fieldset.feature
```diff
@@ -0,0 +1,28 @@
...
+      | <>                      |
+
+  @negative
+  Scenario: Set legend in Fieldset to empty
+    When I set legend to empty
+    Then legend on preview not exists
+
+  @positive
+  Scenario: Verify Fieldset properties
+    # Given I open "Experimental Fieldset" component page
```
KatarzynaQA: Why is this step commented out?

cypress/features/regression/experimental/fieldsetClassic.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Experimental Fieldset component
```
KatarzynaQA: Add that it's related with 'classic' component

cypress/features/regression/experimental/fieldsetClassic.feature
```diff
@@ -0,0 +1,28 @@
...
+  I want to change Experimental Fieldset component properties
+
+  Background: Open Experimental Fieldset component page
```
KatarzynaQA: The same like above

cypress/features/regression/experimental/fieldsetClassic.feature
```diff
@@ -0,0 +1,28 @@
...
+      | <>                      |
+
+  @negative
+  Scenario: Set legend in Fieldset to empty
+    When I set legend to empty
+    Then legend on preview not exists
+
+  @positive
+  Scenario: Verify Fieldset properties
+    # Given I open "Experimental Fieldset" component page
```
KatarzynaQA: Why this step is commented out ? If it is needed. add some info

cypress/features/regression/experimental/decimalInputClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/decimalInputClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/decimalInputClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/decimalInputClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/decimalInputClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/fieldset.feature
DlgSHi: It's commented because we don't need to do any actions before we are making `Then` step

cypress/features/regression/experimental/fieldsetClassic.feature
DlgSHi: It's commented because we don't need to do any actions before we are making `Then` step

cypress/features/regression/experimental/fieldsetClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/fieldsetClassic.feature
DlgSHi: fixed

cypress/features/regression/experimental/fieldsetClassic.feature
```diff
@@ -25,4 +25,4 @@ Feature: Experimental Fieldset component
   @positive
   Scenario: Verify Fieldset properties
     # Given I open "Experimental Fieldset" component page
-    Then Fieldset component has proper field names
\ No newline at end of file
+    Then Fieldset component has proper field names
```
KatarzynaQA: Remove the last blank line, please

DlgSHi: fixed

PR #2452, opened 2019-11-05
FE-2216 fixes issue with input.focus and autoFocus behaviour across b…

damianOfKyan:
# Description

onClick will now trigger openDatePicker
resolved auto focus issue across browsers that was causing the date picker to be open when returning to the browser tab.


DlgSHi:
This issue fixed crashing of DateIpnut component and also IE issue with remaining opened dayPicker after picking up the date
 
The issue with autoFocus will be fixed in this one ticket -> https://jira.sage.com/browse/FE-2296

PR #2453, opened 2019-11-06
FE-2096 Fixed button icon wrong color

mkrds:
# Description
Fixes wrong icon color display when buttonType prop is not provided


src/components/button/button.component.js
```diff
@@ -7,27 +7,35 @@ import tagComponent from '../../utils/helpers/tags';
 import OptionsHelper from '../../utils/helpers/options-helper';
 
 const Button = React.forwardRef((props, ref) => {
-  const { disabled, to, iconType } = props;
+  const {
+    disabled, to, iconType, size, subtext
+  } = props;
+
+  const { as, buttonType, ...rest } = props;
```
edleeks87: Why are you destructuring from props twice?

mkrds: First destructuring is just to get the props used directly in this `Button` component (to avoid `props.disabled` etc)

Second destructuring is just to remove `as` prop and modify `buttonType` prop to be either `buttonProp` or `as`

Idea is to not pass `as` prop further below in component tree while still providing support for this legacy `as` prop.

edleeks87: Yeah I get that part, but can't you just destructure all at once like so?
```
 const {
    disabled, to, iconType, size, subtext, as, buttonType, ...rest
 } = props;
```

mkrds: I could, but then I'd have to define `propsWithoutAs` like that
```js
const propsWithoutAs = {
    ...rest,
    disabled,
    to,
    iconType,
    size,
    subtext,
    buttonType: buttonType || as
};
```

edleeks87: Ah yeah my bad I'll approve

PR #2456, opened 2019-11-06
FE-2296-date-focus-auto-open-issue

damianOfKyan:
…tabs due to auttoFocus issue.

# Description

Fixed issue where switching between browser tabs would trigger autoFocus on Date Input element and then consequent tab switches would trigger Open state on Date Picker.


damianOfKyan:
resolved on different PR. Closing as the approach I took would require massive rewrite of multiple components.

PR #2465, opened 2019-11-12
FE-2296 date picker opens when switching browser tabs

edleeks87:
# Description
Fixes focus behaviour where picker would open if input has focus and user switches between browser tabs.
Also fixes extra onBlur event that was firing when input icon was clicked


src/__experimental__/components/date/date-picker.component.js
```diff
@@ -44,10 +45,11 @@ const DatePicker = (props) => {
   };
 
   useEffect(() => {
-    if (props.selectedDate && monthOrYearHasChanged(datepicker, props.selectedDate)) {
-      datepicker.current.showMonth(props.selectedDate);
+    if (hasComponentUpdated(currentInputDate, isoFormattedValueString(props.inputDate))) {
```
grabkowski: there is no need to use that method arguments, as they could be accessed in it's body
```suggestion
    if (hasComponentUpdated()) {
```

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -44,10 +45,11 @@ const DatePicker = (props) => {
...
+      datepicker.current.showMonth(DateHelper.stringToDate(isoFormattedValueString(props.inputDate)));
```
grabkowski: `DateHelper.stringToDate(isoFormattedValueString(props.inputDate));` could be a const to increase readability

src/__experimental__/components/date/date.spec.js
```diff
@@ -227,6 +235,16 @@ describe('Date', () => {
     it('should update the input element to reflect the passed date', () => {
       expect(wrapper.update().find('input').prop('value')).toBe(getFormattedDate(mockDate));
     });
+
+    it('should return focus to the date input and the picker should not open', () => {
+      const instance = wrapper.find(BaseDateInput).instance();
+      expect(instance.inputFocusedViaPicker).toEqual(true);
+      const spy = spyOn(instance, 'updateSelectedDate');
+      const stateSpy = spyOn(instance, 'setState');
```
grabkowski: I think that you shouldn't spy on the instance of the component

src/__experimental__/components/date/date-picker.component.js
edleeks87: Good point will fix locally

src/__experimental__/components/date/date-picker.component.js
edleeks87: Agreed 👍

src/__experimental__/components/date/date.spec.js
edleeks87: Yeah I'll rewrite this as it's pretty much all implementation specific

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -44,15 +45,22 @@ const DatePicker = (props) => {
...
     }
-  }, [props.selectedDate]);
+  }, [props.selectedDate, props.inputDate]);
 
   function handleDayClick(selectedDate, modifiers) {
     if (!modifiers.disabled) props.handleDateSelect(selectedDate);
   }
 
+  function hasComponentUpdated() {
+    const propDate = isoFormattedValueString(props.inputDate);
```
grabkowski: what if props.inputDate is undefined?

edleeks87: It shouldn't ever be undefined as the value passed from the input is based of the value in state, I just added this guard so that the picker showMonth is never going to be passed undefined

edleeks87: Oh wait I realise what you mean now, I will move the guard 👍

grabkowski: then why do we have a check for that in the line below?

grabkowski: no need as it will just be undefined and falsy value will be returned by the guard below

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -70,9 +70,8 @@ const DatePicker = (props) => {
   if (!didMount) {
     return null;
   }
-
   return (
-    <Portal onReposition={ () => didMount && setContainerPosition(getContainerPosition(window, props.inputElement)) }>
+    <Portal>
```
grabkowski: I think removing that handler will introduce a bug in which after screen resizing (e.g. orientation change) the opened DatePicker won't reposition.

edleeks87: You are correct my good man 😞, will revert it

PR #2473, opened 2019-11-13
FE-2271 Allow empty Decimal Component value

grabkowski:
Added 'isEmptyOnInit' prop to allow empty initial value


jamime:
Feature included in https://github.com/Sage/carbon/pull/2513

PR #2475, opened 2019-11-13
FE-2309 stop additional onclick submitting form

DipperTheDan:
# Description
Add to type property to button component to prevent additional button submitting form.


src/__deprecated__/components/form/form-button/form-button.component.js
```diff
@@ -16,6 +16,7 @@ const propsForButton = (props) => {
   return ({
     save: {
       as: 'primary',
+      type: undefined,
```
jamime:
```suggestion
      type: 'submit',
```

jamime: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit

change_log/next/FE-2309-fix_child_button_submitting_form.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "Adds type property to button component to stop form submission on click. Removes type property from save button in form. (component: Button, FormButton)"
```
grabkowski: I think that could be split into two separate entries, e.g.:
```
Bug Fixes: "Adds type property to button component to stop form submission on click.  (component: Button)"
Bug Fixes: "Removes type property from save button in form. (component: FormButton)"
```

src/__deprecated__/components/form/form-button/form-button.component.js
DipperTheDan: Made the change locally as I had to update the snapshot. Thank you.

PR #2476, opened 2019-11-13
FE-2067 Fixes incorrect help icon colour in the Classic story

grabkowski:


sianford:
Tested the help icon colour was the correct #8099A4 in the classic stories of all components with a help icon in the deprecated, experimental and main component directories. Also checked the default stories in the same components were the correct rgba(0,0,0,0.65) and rgba(0,0,0,0.90) on hover. Tested in Safari, Chrome, Firefox, IE11 and Edge browsers. Testing passed ✅

PR #2479, opened 2019-11-14
FE-2294 - ActionPopover - added support for classic theme

mkrds:
# Description
Added support for classic theme


src/components/action-popover/action-popover.spec.js
```diff
@@ -2,11 +2,15 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { act } from 'react-dom/test-utils';
 import { ThemeProvider } from 'styled-components';
+import TestRenderer from 'react-test-renderer';
+import 'jest-styled-components';
```
edleeks87: This is imported globally in the jest config now

src/components/action-popover/__snapshots__/action-popover.spec.js.snap
```diff
@@ -0,0 +1,237 @@
...
+  box-shadow: inset 0px 0px 0px 2px #FFB500;
+}
+
+.c0:focus {
+  outline: none;
+  box-shadow: inset 0px 0px 0px 2px #255BC7;
+}
+
+<div
+  className="c0"
```
edleeks87: This snapshot has the theme object in it so will always need updating if the theme is altered

src/components/action-popover/action-popover.spec.js
jamime: This is already imported by `expect.js`

DlgSHi:
Removing `Pending QA` label until `Passed UX QA` will be gained

harpalsingh:
Could we add a white background like the new themes to the actual selector, as you can see its not visible here:

Screenshot 2019-11-18 at 09 52 41

harpalsingh:
Could we also fix the placement now that the trigger has the white background and also remove the border around the dropdown.

**Border to be removed and right: 0 in placement.**

Before:
Screenshot 2019-11-18 at 15 27 42

After those fixes are applied:
Screenshot 2019-11-18 at 15 27 31

PR #2481, opened 2019-11-15
docs: integrating Formik into Carbon FE-2283

kyanmichael:
# Description

WIP/draft RFC for integrating Formik into Carbon.

Also includes:

* **FE-2088**  –  Example of wiring up Formik's `dirty` state to React Router via `history.block()`.

### DIFF
--- a/rfcs/text/formik.md
+++ b/rfcs/text/formik.md
@@ -1,0 +1,695 @@
+- Start Date: 2019-11-15
+
+
+# Table of contents
+
+- [Table of contents](#table-of-contents)
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Detailed design](#detailed-design)
+  - [Overview of Carbon's form and validation functionality](#overview-of-carbons-form-and-validation-functionality)
+  - [Overview of Formik](#overview-of-formik)
+    - [Form validation using Formik](#form-validation-using-formik)
+  - [Code examples](#code-examples)
+    - [Checkbox and CheckboxGroup](#checkbox-and-checkboxgroup)
+      - [HTML and React](#html-and-react)
+      - [Checkbox usage with Formik](#checkbox-usage-with-formik)
+      - [Form-level validation examples](#form-level-validation-examples)
+      - [Simplification using `<Field>`](#simplification-using-field)
+      - [Field-level validation example](#field-level-validation-example)
+      - [CheckboxGroup example: uncontrolled](#checkboxgroup-example-uncontrolled)
+      - [CheckboxGroup example: controlled](#checkboxgroup-example-controlled)
+    - [Blocking navigation on unsaved changes](#blocking-navigation-on-unsaved-changes)
+- [Drawbacks](#drawbacks)
+  - [No support for non-blocking warning or info validations](#no-support-for-non-blocking-warning-or-info-validations)
+- [Alternatives](#alternatives)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+
+# Summary
+
+Consider replacing Carbon's form and validation functionality with Formik 2.0.
+
+
+<!-- # Basic example
+
+If the proposal involves a new or changed API, include a basic code example.
+Omit this section if it's not applicable. -->
+
+
+# Motivation
+
+React provides low-level support for HTML form elements (`<input>`, `<textarea>`, `<select>`, etc), but it doesn't
+provide any built-in functionality for form validation, handling form submission, etc.  Therefore, Carbon currently
+provides [`<Form>`](src/__experimental__/components/form/form.component.js) and
+[`<FormField>`](src/__experimental__/components/form-field/form-field.component.js) experimental components and
+[`withValidation`](src/components/validations/with-validation.hoc.js) HOC.
+
+However, much of that functionality is also provided by existing third-party open-source libraries (such as
+[React Form](https://github.com/tannerlinsley/react-form), [Formsy React](https://github.com/formsy/formsy-react),
+[Informed](https://joepuzzo.github.io/informed), [Unform](https://github.com/Rocketseat/unform), etc).  React's
+documentation on [fully-fledged form solutions](https://reactjs.org/docs/forms.html#fully-fledged-solutions) mentions:
+
+> If you're looking for a complete solution including validation, keeping track of the visited fields, and handling
+> form submission, [Formik](https://jaredpalmer.com/formik) is one of the popular choices.
+
+The purpose of this RFC is to discuss the advantages and disadvantages of deprecating Carbon's existing form and
+validation functionality and integrating [Formik](https://jaredpalmer.com/formik) as a replacement.
+
+There's plenty of general discussion online about "reinventing the wheel" (see
+[Stack Exchange](https://softwareengineering.stackexchange.com/questions/29513/is-reinventing-the-wheel-really-all-that-bad),
+[Quora](https://www.quora.com/Why-is-it-discouraged-to-reinvent-the-wheel-in-programming-Wont-it-give-me-a-deeper-and-better-understanding-of-how-things-work-underneath),
+[Jeff Atwood](https://blog.codinghorror.com/dont-reinvent-the-wheel-unless-you-plan-on-learning-more-about-wheels/),
+etc).  Some of the key advantages of "not reinventing the wheel" (ie: not building a custom solution, but using an
+existing third-party library) are:
+
+* **Less code to maintain**  –  Much time and effort can be saved by using a third-party library which is kept
+  well-maintained, thus freeing up one's development resources to focus on higher-priority project-unique functionality.
+* **Robustness**  –  A popular well-maintained library is likely to be thoroughly tested and debugged, especially for
+  edge cases which have been encountered and addressed by the library's community.
+* **Familiarity**  –  Users are potentially already familiar with popular third-party libraries.
+
+Considering the reasons why "reinventing the wheel" (ie: building a custom solution) can sometimes be beneficial in
+certain situations:
+
+* It can be a very educational experience for developers, but that isn't a relevant factor for this project.
+* A third-party library may be an unnecessarily heavy solution, if only a small percentage of its functionality will be
+  used by the application, and a custom-built solution may be much more lightweight.  But in this case, Formik is
+  [a relatively small library](https://jaredpalmer.com/formik/docs/overview)
+  ([14.7kB when minified and gzipped](https://bundlephobia.com/result?p=formik@2.0.4)), and we would be using a
+  significant percentage of its functionality.
+
+
+# Detailed design
+
+## Overview of Carbon's form and validation functionality
+
+Carbon provides the following components:
+
+* [**`<Form>`**](src/__experimental__/components/form/form.component.js)  –  This augments an HTML
+  [`<form>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) element with: a footer (containing "Save"
+  and "Cancel" buttons), form submission functionality, [CSRF](https://en.wikipedia.org/wiki/CSRF) functionality, a
+  [`beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) "unsaved" warning, etc.
+* [**`<FormField>`**](src/__experimental__/components/form-field/form-field.component.js)  –  This enables a
+  [`<Label>`](src/__experimental__/components/label/label.component.js) and
+  [`<FieldHelp>`](src/__experimental__/components/field-help/field-help.component.js) to be added to an individual
+  form field.
+* [**`withValidation`**](src/components/validations/with-validation.hoc.js)  –  This HOC wraps an individual form field,
+  enabling validation functions (for errors, warnings and info) to be executed when the component blurs or receives
+  updated props.
+
+
+## Overview of Formik
+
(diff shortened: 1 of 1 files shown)


rfcs/text/formik.md
```diff
@@ -0,0 +1,170 @@
...
+([`fn-name`](https://www.npmjs.com/package/fn-name),
+[`property-expr`](https://www.npmjs.com/package/property-expr),
+[`synchronous-promise`](https://www.npmjs.com/package/synchronous-promise) and
+[`toposort`](https://www.npmjs.com/package/toposort)).
+
+Some of the key disadvantages/risks of using a third-party library are:
+
+* **Risk of abandonment**  –  It's very common for open-source projects to eventually be abandoned by their
+  maintainers.  When this happens, any current or future bugs will never be fixed, and community-provided support (on
+  GitHub, Stack Overflow, etc) will gradually decline.
```
ianforster-sage: It may be useful to ask a question along the lines of 

> would we be comfortable maintaining the library ourselves if the current maintainer stopped doing so and community support failed to fill the void?

ianforster-sage: This of course is no worse than our current situation of maintaining our own attempt within Carbon.

rfcs/text/formik.md
```diff
@@ -0,0 +1,399 @@
...
+
+# Detailed design
+
+## Overview of Carbon's form and validation functionality
+
+Carbon provides the following components:
+
+* [**`<Form>`**](src/__experimental__/components/form/form.component.js)  –  This augments an HTML
+  [`<form>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) element with: a footer (containing "Save"
+  and "Cancel" buttons), form submission functionality, [CRSF](https://en.wikipedia.org/wiki/CRSF) functionality, a
```
edleeks87: Typo CRSF -> CSRF

rfcs/text/formik.md
```diff
@@ -0,0 +1,444 @@
...
+## Overview of Formik
+
+> Formik keeps track of your form's state, and then exposes it (plus a few reusable methods and event handlers) to your
+> form via props. [🔗](https://jaredpalmer.com/formik/docs/overview#the-gist)
+
+Formik provides a small set of React components and hooks, of which the key ones are:
+
+* [**`<Formik>`**](https://jaredpalmer.com/formik/docs/api/formik)  –  The central component of Formik, which wraps an
+  HTML form, and accepts props such as
+  [`initialValues`](https://jaredpalmer.com/formik/docs/api/formik#initialvalues-values),
```
edleeks87: We will probably want to document somewhere that Formik will only work for controlled inputs, highly unlikely that someone would try to use it with uncontrolled but better if we make it explicit IMO

rfcs/text/formik.md
kyanmichael: Wow, well spotted, thanks! :clap: Fixed in db9e307

rfcs/text/formik.md
kyanmichael: Good point :thumbsup: I suppose technically it's possible to use Formik with an uncontrolled input, by wiring up [`handleChange`](https://jaredpalmer.com/formik/docs/api/formik#handlechange-e-reactchangeevent-any-void) (for Formik's validation functionality) but not wiring up `value` to Formik. But that would cause complications, and it's not really how Formik is intended to be used.

edleeks87: Yeah good point, think we are best not encouraging it's use uncontrolled but we can't really force the hand of developers using Carbon

edleeks87:
As `Formik` only supports the single `error` state but we have multiple categories in some cases (such as `warnings` and so on) we will probably need to come up with a solution to support them. As the other lower categories do not block submission I don't think this blocks the us integrating, we will just need to add some handling for them and allow `Formik` to manage the errors

rfcs/text/formik.md
kyanmichael: Added documentation about uncontrolled components in 3516b7f :thumbsup:

jamime:
@kyanmichael did we come up with a way to display
- [ ] Info validations that do not prevent submit
- [ ] Warning validations that do not prevent submit

kyanmichael:
@edleeks87 @jamime: Regarding your comments :point_up: I've just added a new section to the RFC:

* https://github.com/Sage/carbon/blob/FE-2283-formik-rfc/rfcs/text/formik.md#no-support-for-non-blocking-warning-or-info-validations

edleeks87:
> @edleeks87 @jamime: Regarding your comments ☝️ I've just added a new section to the RFC:
> 
> * https://github.com/Sage/carbon/blob/FE-2283-formik-rfc/rfcs/text/formik.md#no-support-for-non-blocking-warning-or-info-validations

Thanks @kyanmichael, shame there is no support as it's a critical feature that we will need to support. As discussed offline with yourself and @jamime maybe we should explore how we might achieve this using `Yup` would be happy to pair up on this tomorrow if that works for you?

rfcs/text/formik.md
```diff
@@ -0,0 +1,624 @@
...
+      - [HTML and React](#html-and-react)
+      - [Checkbox usage with Formik](#checkbox-usage-with-formik)
+      - [Form-level validation examples](#form-level-validation-examples)
+      - [Simplification using `<Field>`](#simplification-using-field)
+      - [Field-level validation example](#field-level-validation-example)
+      - [CheckboxGroup example: uncontrolled](#checkboxgroup-example-uncontrolled)
+      - [CheckboxGroup example: controlled](#checkboxgroup-example-controlled)
+    - [Blocking navigation on unsaved changes](#blocking-navigation-on-unsaved-changes)
+- [Drawbacks](#drawbacks)
+  - [No support for non-blocking ("warning" or "info") validations](#no-support-for-non-blocking-%22warning%22-or-%22info%22-validations)
```
kyanmichael: @edleeks87: This change you made in 6f93191 seems to have broken the link:

```diff
- [No support for non-blocking ("warning" or "info") validations](#no-support-for-non-blocking-warning-or-info-validations)
+ [No support for non-blocking ("warning" or "info") validations](#no-support-for-non-blocking-%22warning%22-or-%22info%22-validations)
```

This link doesn't work for me:

* https://github.com/Sage/carbon/blob/FE-2283-formik-rfc/rfcs/text/formik.md#no-support-for-non-blocking-%22warning%22-or-%22info%22-validations

But the original link works for me:

* https://github.com/Sage/carbon/blob/FE-2283-formik-rfc/rfcs/text/formik.md#no-support-for-non-blocking-warning-or-info-validations

Please can we revert the change? :point_down:

```suggestion
  - [No support for non-blocking ("warning" or "info") validations](#no-support-for-non-blocking-warning-or-info-validations)
```

jamime: docs: revert link format

mkrds:
@kyanmichael @edleeks87 @jamime Hey guys, couldn't we just contact Jared Palmer (formik's creator), ask if he would accept our PR with validation mechanism extension and just add it to formik on our own?

rfcs/text/formik.md
jamime: Marking as resolved. I don't want the RFC to introduce more questions, I'm happy with the statement as it stands.

jamime:
@kyanmichael 
At the moment `Textbox` requires a minimum of 3 props to display a validation error.
1. `hasError={true}`
2. `inputIcon="error"`
3. `tooltipMessage="The error message"`

- [ ] We should document a simpler validation API that we will implement across our components.

I was thinking just `error`, `warning` or `info` would suffice. There's a question about what we do when a single field has more than one error. I _think_ that's out of the scope of this ticket.

jamime:
I have created this sandbox to show how we can use carbon components, it's not "reference architecture" but a sanity check that we're not tied to the current implementation.

https://codesandbox.io/s/formik-validation-example-pch92

We still need to 

- [ ] We should document a simpler validation API that we will implement across our components.
- [ ] I would like the code sandbox to be continued to test all of our components, including `Form` which shows the number of validation errors. We should save this to GitHub so we can showcase how to use `formik` when we have `storybook@5.2.0+` available.

- [ ] Document our intention, we should explicitly state that we are going to remove the `withValidationHOC`, `validate` prop and any inbuilt validations.

- [ ] Raise tickets for any components that don’t work out of the box with formik. I suspect that the `Form` component may rely on context and not props to set the validation counts. See if we can make a non-breaking change to support both in the interim.

edleeks87:
I've integrated the example Michael provided for achieving non-blocking validations using yup with the sandbox @jamime provided above to demonstrate that that it is possible to achieve it on a Carbon input level at least 
https://codesandbox.io/s/objective-wilson-km9hv

rfcs/text/formik.md
```diff
@@ -0,0 +1,695 @@
...
+
+The alternative to integrating Formik is to continue to maintain the current implementation of Form and Validations in Carbon. We would then continue to add additional functionality to meet any new requirements that may arise. This currently already adds a significant overhead to the team any will only grow as new functionality is implemented. A further issue is that it will continue to cause vendor lock-in, forcing consumers to use what is provided by Carbon out-of-the-box, and may hurt its adoption as a result.
+
+
+# Adoption strategy
+
+As discussed, we propose to remove the majority of the code from the `Carbon` `Form` and have it serve predominantly as a UI wrapper that renders either a `Formik` `Form` or an alterantive package chosen by the developers working on the consuming projects. Carbon's From will no longer be responsible for implementing the valididation interface that it currently does: the Form component will no longer be exported with the functionality added by the `withValidationHOC`.
```
chrisbarber86:
```suggestion
As discussed, we propose to remove the majority of the code from the `Carbon` `Form` and have it serve predominantly as a UI wrapper that renders either a `Formik` `Form` or an alternative package chosen by the developers working on the consuming projects. Carbon's Form will no longer be responsible for implementing the validation interface that it currently does: the Form component will no longer be exported with the functionality added by the `withValidationHOC`.
```

rfcs/text/formik.md
```diff
@@ -0,0 +1,695 @@
...
+
+As such, thevalidation interface will need to be simplified and updated as the use of the `withValidationHOC` is currently required to update the `FormSummary` sub-component and control the acceptance criteria for a successful form submission via a successful run of the `validate` function. The removal of a key aspect of the current validation API means carrying out the changes proposed in this RFC would be a breaking change.
```
chrisbarber86:
```suggestion
As such, the validation interface will need to be simplified and updated as the use of the `withValidationHOC` is currently required to update the `FormSummary` sub-component and control the acceptance criteria for a successful form submission via a successful run of the `validate` function. The removal of a key aspect of the current validation API means carrying out the changes proposed in this RFC would be a breaking change.
```

rfcs/text/formik.md
```diff
@@ -0,0 +1,695 @@
...
+
+
+# Alternatives
+
+There are a number of alternative packages available to developers looking for a solution to web forms and validation. It is our intention to provide an API that affords developers the flexibility to choose whatever solution they want to and integrate their choice with Carbon's Form and input components. 
+
+The alternative to integrating Formik is to continue to maintain the current implementation of Form and Validations in Carbon. We would then continue to add additional functionality to meet any new requirements that may arise. This currently already adds a significant overhead to the team any will only grow as new functionality is implemented. A further issue is that it will continue to cause vendor lock-in, forcing consumers to use what is provided by Carbon out-of-the-box, and may hurt its adoption as a result.
+
+
+# Adoption strategy
```
chrisbarber86: We will need to think about what this looks like to migrate from how forms are built today - to how forms are built tomorrow using Formik. I.e. if i'm a developer with a project using v12.0.0 of Carbon, what do i need to do to upgrade to the version with Formik support.

rfcs/text/formik.md
```diff
@@ -0,0 +1,695 @@
...
+      - [HTML and React](#html-and-react)
+      - [Checkbox usage with Formik](#checkbox-usage-with-formik)
+      - [Form-level validation examples](#form-level-validation-examples)
+      - [Simplification using `<Field>`](#simplification-using-field)
+      - [Field-level validation example](#field-level-validation-example)
+      - [CheckboxGroup example: uncontrolled](#checkboxgroup-example-uncontrolled)
+      - [CheckboxGroup example: controlled](#checkboxgroup-example-controlled)
+    - [Blocking navigation on unsaved changes](#blocking-navigation-on-unsaved-changes)
+- [Drawbacks](#drawbacks)
+  - [No support for non-blocking ("warning" or "info") validations](#no-support-for-non-blocking-%22warning%22-or-%22info%22-validations)
```
ianforster-sage: I think the entity characters are screwing with the skip-links here

edleeks87: @ianforster-sage  not sure what you mean, the links seem to navigate to the correct sections and the text seems fine, could you clarify for me as I may be missing something?

jamime:
I've just implemented deprecation warnings.

Going forward we should advise to use the props directly on the components to control the styling. `Form` and `inputs` with `validations`, `info` and `warnings` props are now deprecated.

Users should use `formik` and/or `yup, or a library of their choice, and use the props on the `input`s to add the error styling.

jamime:
Here is an updated example with formik and yup. We hope to make the validation interface simpler in the future. e.g. `error="My message` instead of the three props that are currently required.

https://codesandbox.io/s/formik-validation-example-200-9qps9

jamime:
New `Form` component for use with Formik or other Form libraries #2861

jamime:
New validation interface mentioned above: https://github.com/Sage/carbon/pull/2885

PR #2387, opened 2019-10-14
FE-2126 fixed problem with text label overflowing through components.

damianOfKyan:
# Description

fixed label formatting issue.
fixed styling for multi-line label text.

Screenshot 2019-11-06 at 15 03 25

PR #2492, opened 2019-11-21
Fixed checkbox alignment issue

mkrds:
# Description

Fixed an issue with wrong checkbox label and other inputs assignment.

# Testing instructions

This can be tested in `Checkbox` story - both default and classic, by switching reverse knob.
In normal state both label and fieldHelp have `padding-left: 6px` and in reverse state this padding is removed. Previously there's been no difference with this padding which caused the label misalignment as stated in the task description.


grabkowski:
issue fixed in https://github.com/Sage/carbon/pull/2387

PR #2501, opened 2019-11-22
FE-2168 labels now refer to id of input field and not name. focus is …

damianOfKyan:
# Description

label was not using id field.
focus now works as expected when label field is clicked thanks to for attribute matching with input id attribute.


sianford:
Tested in Chrome. All of the labels in all of the stories for all input components with labels have the `for=""`attribute. Selecting the label brings the input into focus and lands the cursor in the input or highlights existing content, as required. A screenreader identifies the label as being connected to the field. Testing passed ✅

PR #2503, opened 2019-11-25
feat: leftchildren property added to select

blaky:
# Description

# TODO
- [X] Release notes

# Screenshots / Gifs

# Testing Instructions


sianford:
Tested in Chrome, IE11 and Edge browsers. Tested by adding some text to each story in the select.stories.js file and by adding an image. The spacing around the image is equal around it. Confirmed that this has not affected the behaviour of the component. Testing passed ✅

PR #2505, opened 2019-11-25
FE-2394 add hidden input to date

edleeks87:
```
New Features:
  - Adds hidden input with raw value in experimental Date (Component: Date)
```


sianford:
Tested in Safari, Chrome, Firefox,, IE11 and Edge in both the Fixtures Testbed and Storybook. Tested all stories, controlled and uncontrolled. The rawValue is present in the onChange and onBlur containing either the date or no date as required. Testing passed ✅

PR #2506, opened 2019-11-26
FE-2217 Select list looping / Select home and end buttons support

mkrds:
# Description

This PR fixes incorrect Select list behaviour where traversing the list with `up` and `down` keyboard arrow buttons resulted in looping the whole list without stopping at the last or first item. However to enable this behaviour in case someone wants the Select to work like that there is a `isLoopable` prop available to be passed to the `Select` component.

Also in this PR I've implemented support for traversing the Select list using `home` and `end` button to select either first or the last selectable item.


change_log/next/FE-2217-select-list-loop.yml
```diff
@@ -0,0 +1,2 @@
+Bug Fixes: "Fixes issue with select list looping while traversing using up and down keyboard arrows (Old behaviour still achievable by passing isLoopable prop to the component). (Components: Select)"
```
jamime: This is a breaking change we need to communicate this in the release note.

PR #2508, opened 2019-11-26
FE-2405 move validation array out of render method

edleeks87:
# Description
The current implementation assigns the validation array on each render which causes it to infinite loop when in a form. To fix this the validation array is now stored in state unless it is part of a `date-range`. The array is only updated when a change is detected in `componentDidUpdate`

### DIFF
--- a/src/__experimental__/components/date/date.component.js
+++ b/src/__experimental__/components/date/date.component.js
@@ -46,7 +46,7 @@ class BaseDateInput extends React.Component {
       this.input.focus();
       this.openDatePicker(true);
     }
-    this.handleValidationUpdate(this.inputProps());
+    this.handleValidationUpdate();
   }
 
   componentDidUpdate(prevProps) {
@@ -55,7 +55,7 @@ class BaseDateInput extends React.Component {
     }
 
     if (this.hasValidationsChanged()) {
-      this.handleValidationUpdate(this.inputProps());
+      this.handleValidationUpdate();
     }
   }
 
@@ -77,16 +77,15 @@ class BaseDateInput extends React.Component {
       return true;
     }
 
-    for (let i = validationsArray.length; i--;) {
-      if (validationsArray[i] !== currentValidations[i]) {
-        return true;
-      }
+    if (validationsArray.some((val, index) => val !== currentValidations[index])) {
+      return true;
     }
 
     return false;
   }
 
-  handleValidationUpdate = (inputProps) => {
+  handleValidationUpdate = () => {
+    const inputProps = this.inputProps();
     this.setState({ validationsArray: concatAllValidations(inputProps) });
   }
 
@@ -391,7 +390,9 @@ BaseDateInput.propTypes = {
   /** The current date YYYY-MM-DD */
   value: PropTypes.string,
   /** Triggers textbox validation when it's boolean value changes */
-  forceUpdateTriggerToggle: PropTypes.bool
+  forceUpdateTriggerToggle: PropTypes.bool,
+  /** Temporary flag to indeicate if input is part of DateRange */
+  isDateRange: PropTypes.bool
 };
 
 BaseDateInput.defaultProps = {
--- a/src/__experimental__/components/date/date.component.js
+++ b/src/__experimental__/components/date/date.component.js
@@ -391,7 +391,7 @@ BaseDateInput.propTypes = {
   value: PropTypes.string,
   /** Triggers textbox validation when it's boolean value changes */
   forceUpdateTriggerToggle: PropTypes.bool,
-  /** Temporary flag to indeicate if input is part of DateRange */
+  /** Temporary flag to indicate if input is part of DateRange */
   isDateRange: PropTypes.bool
 };
 
--- a/src/__experimental__/components/date/date.component.js
+++ b/src/__experimental__/components/date/date.component.js
@@ -46,17 +46,49 @@ class BaseDateInput extends React.Component {
       this.input.focus();
       this.openDatePicker(true);
     }
+    this.handleValidationUpdate(this.inputProps());
   }
 
   componentDidUpdate(prevProps) {
     if (this.isControlled && !this.inputHasFocus && this.hasValueChanged(prevProps)) {
       this.updateSelectedDate(this.props.value);
     }
+
+    if (this.hasValidationsChanged()) {
+      this.handleValidationUpdate(this.inputProps());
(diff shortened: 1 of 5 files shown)


src/__experimental__/components/date/date.component.js
```diff
@@ -285,21 +317,24 @@ class BaseDateInput extends React.Component {
   }
 
   render() {
-    const { minDate, maxDate, ...inputProps } = this.props;
+    const {
+      minDate, maxDate, isDateRange, ...inputProps
```
kyanmichael: Should we add `isDateRange` to `BaseDateInput.propTypes` in this file?

src/__experimental__/components/date/date.component.js
```diff
@@ -46,18 +46,50 @@ class BaseDateInput extends React.Component {
...
+
+    if (validationsArray.length !== currentValidations.length) {
+      return true;
+    }
+
+    for (let i = validationsArray.length; i--;) {
+      if (validationsArray[i] !== currentValidations[i]) {
+        return true;
+      }
+    }
```
kyanmichael: I think this can be simplified using [`Array.prototype.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every):

```js
if (!validationsArray.every((val, index) => val === currentValidations[index])) {
  return true;
}
```

or alternatively using [`Array.prototype.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some):

```js
if (validationsArray.some((val, index) => val !== currentValidations[index])) {
  return true;
}
```

src/__experimental__/components/date/date.component.js
```diff
@@ -46,18 +46,50 @@ class BaseDateInput extends React.Component {
...
+
+    return false;
+  }
+
+  handleValidationUpdate = (inputProps) => {
+    this.setState({ validationsArray: concatAllValidations(inputProps) });
+  }
```
kyanmichael: Will there ever be a situation where `handleValidationUpdate()` is passed anything other than `this.inputProps()` ?

If not, then maybe `handleValidationUpdate()` could be rewritten like this?
```js
handleValidationUpdate = () => {
  const inputProps = this.inputProps();
  this.setState({ validationsArray: concatAllValidations(inputProps) });
}
```
and then it can be called like this (with no arguments):
```js
this.handleValidationUpdate();
```

edleeks87: Yeah very good point I'll refactor it 👍

src/__experimental__/components/date/date.component.js
edleeks87: Yep thanks I like this, will refactor now 👍

jamime: I believe `some` will return early where `every` will iterate over everything because it needs to check each value.

edleeks87: Yeah I've gone with `some` as i I like it to return as soon as it identifies an update has happened

src/__experimental__/components/date/date.component.js
edleeks87: I think I might try and refactor this as I don't think it's a great solution having `Date` know about `DateRange`

edleeks87: On further review I'm going to leave this as is and add a propType. It works as a hot-fix and then I'll raise another ticket to refactor how the validation is applied  by implementing `context` that can be overriden when part of a `DateRange`

src/__experimental__/components/date/date.component.js
```diff
@@ -355,7 +390,9 @@ BaseDateInput.propTypes = {
   /** The current date YYYY-MM-DD */
   value: PropTypes.string,
   /** Triggers textbox validation when it's boolean value changes */
-  forceUpdateTriggerToggle: PropTypes.bool
+  forceUpdateTriggerToggle: PropTypes.bool,
+  /** Temporary flag to indeicate if input is part of DateRange */
```
grabkowski:
```suggestion
  /** Temporary flag to indicate if input is part of DateRange */
```

src/__experimental__/components/date/date.component.js
kyanmichael: Thanks for adding the `propType` in 38d0e2a :white_check_mark:

src/__experimental__/components/date/date.component.js
kyanmichael: @jamime: Very good point!

@edleeks87: Thanks for refactoring that in 38d0e2a :white_check_mark:

src/__experimental__/components/date/date.component.js
kyanmichael: Thanks for refactoring that in 38d0e2a :white_check_mark:

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -110,10 +110,16 @@ function getDisabledDays(minDate, maxDate) {
...
+    if (!DateHelper.isValidDate(minDate)) {
+      return null;
+    }
     days.push({ before: DateHelper.stringToDate(minDate) });
   }
 
   if (maxDate) {
+    if (!DateHelper.isValidDate(maxDate)) {
+      return null;
+    }
```
grabkowski: if we have a valid `maxDate` and invalid `minDate` then the valid date will not be pushed. Is that the expected behaviour?

grabkowski: it could be simplified to:
```
const hasValidMinDate = minDate && DateHelper.isValidDate(minDate);
const hasValidMaxDate = maxDate&& DateHelper.isValidDate(maxDate);

if (hasValidMinDate) {
  days.push({ before: DateHelper.stringToDate(minDate) });
}
if (hasValidMaxDate ) {
  days.push({ after: DateHelper.stringToDate(maxDate) });
}
```

edleeks87: Yes you are correct 🤦‍♂ , I'm going to remove this as it doesn't actually fix the issue and raise it as a separate ticket

sianford:
Tested in Chrome to ensure the error, warning and info validations are working correctly. Checked the Controlled Form in the Carbon Fixtures Testbed to ensure the date input is rendering in the form and an error is flagged if an invalid date is entered. Testing passed ✅

PR #2511, opened 2019-11-27
Merge back v9.1.1

edleeks87:
# Description
Merge back updates from release to master


src/__experimental__/components/select/select.component.js
```diff
@@ -392,6 +392,10 @@ class Select extends React.Component {
       formattedValue: this.formattedValue(this.state.filter, value)
     };
 
+    if (leftChildren) {
+      props.leftChildren = props.leftChildren ? [leftChildren, ...props.leftChildren] : leftChildren;
+    }
```
grabkowski: could we simplify that to:
```
if (this.isMultiSelectEnabled()) {
  const multiValues = this.renderMultiValues(value);
  leftChildren = leftChildren ? [leftChildren, multiValues] : multiValues;
}
```
and then just add `leftChildren` to props, or I'm missing something?

PR #2512, opened 2019-11-27
FE-2414 ActionPopover menu button event propagation

mkrds:
# Description

Stopped further event propagation in ActionPopover


sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge browsers. When the Action Popover icon is selected, there is now no rowClick event. There is a rowClick event as expected when the top row is clicked. There is an event as expected when each action in the Action Popover is clicked as expected. Testing passed ✅

edleeks87:
I've restarted the build as it looks like a timeout issue that has been happening on other PRs, will merge if it passes

PR #2487, opened 2019-11-19
FE-2305 - Fixed decimal precision issue

mkrds:
# Description

Fixed decimal precision issue.
Also updated bignumber.js from `4.0.2` to `9.0.0`.

Issue #374, opened 2016-04-08
Decimal component should emit value on deletion of value


andrewjtait:
If I have a decimal value with the value "9.99", and I then delete the value to "", as soon as I blur the input  the visible value is set to "0.00" - however, this value is not emitted to the store, instead the store maintains the "" value. This could cause some inaccuracies, and instead it should ensure that "0.00" is always emitted to the store when the value is blank.

Alternatively, the decimal component could handle an empty value instead of forcing to reset to "0.00", however this is not the pattern we have used in the past.

PR #2513, opened 2019-11-27
feat: Decimal Improvements

### LINKED ISSUE #2485: Entering only '-' (negative sign) into Decimal and then tabbing out, changes value to NaN
If you type a hyphen ("-") as if to enter a negative value, and then tab off (before typing any digits), the entered value changes to "NaN". It's felt this should probably just revert to "0.00"

https://codesandbox.io/s/decimal-nan-y1245

jamime:
# Description
```
New Features:
- allow an empty value using the `allowEmptyValue` prop, this will default to '' instead of '0.00'
- entering a negative symbol only and blurring will revert to the default value ('' or '0.00')
- re-add hidden input that has the `name` and unformatted value

Bug Fixes:
- prevent the cursor from moving unless by the user, it would previously move then re-position itself
- allow the value of the component to be set by updating the `value` prop externally
- allow the user to modify a controlled value
- fire onChange when restoring a default value e.g. when the user deletes the value and it is defaulted to '0.00'

BREAKING CHANGE: The onChange and onBlur interfaces now dispatch an object instead of the SyntheticEvent. The decimal
value can be accessed from `e.target.value.rawValue` or `e.target.value.formattedValue`.

Invalid precision now throws an invariant error

Changing from controlled to uncontrolled (or vice versa) now throws an invariant error

Controlling the component with anything but a string now throws an invariant error

Event utils now prefer `key` over the deprecated `which` property. There should be no noticeable changes when called
with a SyntheticEvent.

Fixes #2485, #374, FE-2271, FE-2238, FE-282
```
In order to test this, I also needed to resolve the following issue
```
fix: resolve fixture logging issue FE-2386
```

Also included in this PR is https://github.com/Sage/carbon/pull/2487
```
Fixed decimal precision issue.
Also updated bignumber.js from 4.0.2 to 9.0.0.
```
# TODO
- [x] Release notes

#2485, #374, FE-2271, FE-2238, FE-282, FE-2386

# Screenshots / Gifs
image
image

# Testing Instructions
```
git pr 
npm install
npm test -- src/__experimental__/components/decimal
npm run fixtures
```
Please use the following fixture screens
[Controlled Decimal](http://localhost:9901/controlled/decimal)
[Uncontrolled Decimal](http://localhost:9901/controlled/decimal)

I have also implemented `i18n` support to the fixtures so it is possible for you to test the French, English, Swiss and default Decimal formatting.

## QA
The following issues raised on Slack have been fixed, please can you also test these as there are no JIRA tickets.
> * setting the value programatically
> * need onBlur event
> * needs to to be able to edit numbers containng thousands seperators & for simple string to be raised for onChange/onBlur event

## UX QA
- The user can now only enter the negative symbol at the start of the number.
- The user can now delete a value if `allowEmptyValue` is `true`. This will allow the user to enter an empty string.
- The users' cursor now remains static, it would previously move to the end then back the user presses a key that would make the input invalid.


jamime:
I've got some failing tests to address.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -1,130 +1,255 @@
...
+    // We're checking to see if the proposed change is valid, if not we block the user input
+    if (!this.isValidKeyPress(ev)) {
+      ev.preventDefault();
+    }
+    if (this.props.onKeyPress) {
+      this.props.onKeyPress(ev);
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) {
```
mkrds: Not really a blocking comment but I think it would be nice if we'd all follow airbnb's recommendation on ordering of lifecycle methods. It makes components definitely more readable and easier to grasp.
https://github.com/airbnb/javascript/tree/master/react#ordering
Could be also enforced by an ESLint rule.

fixtures/src/components/chrome.js
```diff
@@ -1,29 +1,92 @@
...
+      guid: guid(),
+      timestamp: new Date().toISOString(),
       name,
       id,
       value,
       checked
-    }]);
+    }];
+    setHistory(ref.current);
   }, [history, setHistory]);
```
damianOfKyan:
```suggestion
  }, [ref, setHistory]);
```

harpalsingh:
@jamime to verify, is the expected behaviour here that if you try to paste a number which has large precision than required, it simply does not or even tell the user any issue?

jamime:
@harpalsingh Yes the input will be blocked, we did discuss adding the wiggle animation, I'll create a separate ticket for that.

PR #2516, opened 2019-11-29
FE-2428 Enzyme Snapshot Helper

grabkowski:
When generating a snapshot of an Enzyme wrapper with an object passed as a prop, that object will be rendered in a snapshot, obfuscating the result.
With this helper we could render snapshots with these objects replaced by `[ <prop name> object ]`.

Previously we have avoided this issue by using `TestRenderer.Create`, this helper will allow us to not mix various test tools.

To remove theme object:
`expect(noThemeSnapshot(wrapper)).toMatchSnapshot()`

To remove a custom object/objects:
`expect(snapshotWithoutProps(wrapper, 'propName')).toMatchSnapshot()`
`expect(snapshotWithoutProps(wrapper, ['firstProp', 'secondProp'])).toMatchSnapshot()`

### DIFF
--- a/src/__spec_helper__/enzyme-snapshot-helper.js
+++ b/src/__spec_helper__/enzyme-snapshot-helper.js
@@ -1,0 +1,29 @@
+import toJson from 'enzyme-to-json';
+
+function snapshotWithoutProps(enzymeWrapper, propNames) {
+  let propNameList = propNames;
+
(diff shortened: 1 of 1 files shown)


grabkowski:
> Does this not need some tests writing for it?

this is an unit test util, so it is only used in unit tests

PR #2518, opened 2019-12-02
refactor: removes title attribute from Tab StyledDiv to remove obtrusive tooltip

ianforster-sage:
# Description
There is a tooltip on the child of a Tab that has the Tab title as content, which should not be visible. This refactor removes the title attribute from the StyledDiv which causes this behaviour.

# TODO
- [x] Release notes


sianford:
Tested in Safari, Chrome, Firefox, IE11 and Edge browsers. Checked all align and position options to confirm the tooltip is no longer appearing when the Content text is hovered over, selected or any tabs are brought into focus or hovered over. Testing passed ✅

PR #2523, opened 2019-12-03
FE-2218 Select control F6 copy action

mkrds:
# Description

Added possibility to pass onKeyDown prop callback which will allow developers to hook to the onKeyDown event and react properly on F6 press


src/__experimental__/components/select/select.component.js
```diff
@@ -148,6 +148,7 @@ class Select extends React.Component {
    *   - `value` - The option's invisible internal value.
    */
   handleChange = (optionProps) => {
+    console.log(optionProps);
```
damianOfKyan: is that left here on purpose?

PR #2527, opened 2019-12-03
FE-2433 - Select add a prop to prevent auto open on focus

mkrds:
# Description

Added `preventFocusAutoOpen` prop to `Select` which prevents the list from automatically opening when the component gets focused.

# Testing instructions

Can be tested using `preventFocusAutoOpen` knob in `Select` component stories


src/__experimental__/components/select/select.spec.js
```diff
@@ -316,6 +316,12 @@ describe('Select', () => {
       expect(onFocus).toHaveBeenCalled();
     });
 
+    it('does not open the list if `preventFocusAutoOpen` prop is passed', () => {
```
damianOfKyan:
```suggestion
    it('does not open the list if `preventFocusAutoOpen` prop is passed and set to true', () => {
```

PR #2531, opened 2019-12-04
FE-2421 no placeholder when select is read only

DipperTheDan:
# Description
Bug Fixes: "Sets placeholder to empty string when `readOnly` prop is true. (component: Select)"


src/__experimental__/components/select/select.spec.js
```diff
@@ -109,6 +109,35 @@ describe('Select', () => {
...
+    expect(wrapper.find(Input).props().placeholder).toEqual('');
+  });
+
+  it('does not render a placeholder when disabled is false and readOnly is true', () => {
+    const props = { disabled: false, readOnly: true };
+    const wrapper = renderWrapper({ props });
+    expect(wrapper.find(Input).props().placeholder).toEqual('');
+  });
+
+  describe.each(
```
mkrds: Doesn't this test case with `describe.each` actually covers the 3 test cases above?

DipperTheDan: Yeah. Ed showed me another way to do the tests and I left both ways in the code. My bad.

harpalsingh:
I'll pass UX review on this particular issue, however, we should look to adding the arrow indicator for disabled as well, so its clear its a disabled select and not a disabled textbox.

PR #2534, opened 2019-12-04
FE-2408 fixed console errors and story crashing during onChange event…

damianOfKyan:
# Description

changed onClick propType to PropTypes.func
changed storybook store value being passed to the component to be in line with propType.


change_log/next/FE-2408-inline-input-error.yml
```diff
@@ -0,0 +1,3 @@
+Bug Fixes:
+  - "Fixes issue were propType of onClick was expected to be an object rather than function. (Component: InputIconToggle)"
```
grabkowski: a typo: `were`

damianOfKyan: `were` is a correct form here.

damianOfKyan: apologies, changed to `where`.

jamime: shouldn't it be `when` or `where`?

PR #2535, opened 2019-12-04
FE-2449 - Fixed table header alignment and missing sort icon

mkrds:
# Description

Fixes issue with with incorrect alignment of header in sortable table component caused by missing icon. 

# Testing instructions

Check behaviour of `Table` sorting by clicking on the table header.


src/components/table/table-row/__snapshots__/table-row.spec.js.snap
```diff
@@ -252,7 +248,7 @@ exports[`TableRow render if is not classic theme renders a row to match the snap
 }
 
 .c10 .c3:first-child:not(['data-component=selectable-cell']),
-.c10 .c15:first-child:not(['data-component=selectable-cell']) {
+.c10 .sc-htoDjs:first-child:not(['data-component=selectable-cell']) {
```
grabkowski: we should not have generated styled-components classes in snapshots

mkrds: Seems that it's a bug caused by `jest-styled-components` as seen below. Generated className however is always the same both on my PC and on CI so it's not failing on each test run.

https://github.com/styled-components/jest-styled-components/issues/227

harpalsingh:
The sort icon needs fixing for the secondary and tertiary themes, as default for now, match text color in header.

As you can see the problem below in the visual:

Screenshot 2019-12-10 at 10 40 19

Screenshot 2019-12-10 at 10 40 29

harpalsingh:
Not sure if this is a storybook issue? but once you switch themes the color of the sort does not change until you reselect it.

So if you sort on primary, then switch the theme to secondary you get the following:

Screenshot 2019-12-11 at 13 16 07

If I re-sort, then the icon is fixed, if this is only a storybook issue and not a theme issue, thats ok. can you verify @mkrds

mkrds:
@harpalsingh Theme change resets the whole story, so after the theme change there's actually no sort icon because the table is not sorted yet. So yes, that's perfectly fine.

Issue #2324, opened 2019-09-17
Date input / Date range error icon misaligned


carlcheel-sage:
It looks as though the alignment of the error icon is out of alignment in `v9.0.0-rc3`. 

![image](https://user-images.githubusercontent.com/25386428/65035030-272e7080-d940-11e9-834b-8ca5b3e35955.png)

You can recreate this issue in the carbon storybook: https://carbon.sage.com/storybook/iframe.html?id=date-range--default

PR #2537, opened 2019-12-04
FE-2448 fixed validation icon alignment.

damianOfKyan:
# Description

Fixes issue where validation icon on date input was misaligned for depricated date and dateRange components

Issue reported:
https://github.com/Sage/carbon/issues/2324

before:
Screenshot 2019-12-10 at 11 00 41

after:
Screenshot 2019-12-10 at 11 01 03


harpalsingh:
When I go to Experimental - Textbox - Validations I see this?

Screenshot 2019-12-06 at 10 55 38

Having reviewed the designs, we should set the input padding at 12px (its 11px atm, thats ok) but the icon itself should sit next to that padding from the right, so where can I check the fix? and/or textbox still needs fixing?

harpalsingh:
Here the green area is 12px, to demonstrate:

padding

harpalsingh:
As per the visual, we see this:

Screenshot 2019-12-09 at 10 00 07

The alignment of the icon is great, but I would expect the tooltip arrow to be touch the red border.

harpalsingh:
I just ran this branch and looked at the date input, as you can see from the screenshot below did you just revert it back to the problem in the first place? the tooltip is next to the containers, but the padding and spacing of the icon is back to what it was before?

alignment

PR #2541, opened 2019-12-06
FE-2454 - Added wiggle animation to Decimal component

mkrds:
# Description
Added wiggle animation to Decimal component which is supposed to indicate that something is wrong

# Testing Instructions
Try typing or pasting something that will lead to invalid value in `Decimal` component


src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -20,7 +21,8 @@ class Decimal extends React.Component {
     this.state = {
       value,
       visibleValue: this.formatValue(value),
-      isControlled
+      isControlled,
+      isWiggling: false
```
dawidzarzycki: What about changing the name from `isWiggling` animation to `isAnimating`? It will be more flexible for the future if anyone will want to change the animation behavior

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -54,6 +56,10 @@ class Decimal extends React.Component {
     }
   }
 
+  startWiggleAnimation = () => this.setState({ isWiggling: true })
```
dawidzarzycki: According to the comment above. What about just `startAnimation` and `stopAnimation`

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -1499,7 +1540,7 @@ describe('Decimal', () => {
 
     it('has the correct automation selectors', () => {
       render();
-      expect(wrapper.current.childAt(0).getDOMNode().getAttribute('data-component')).toBe('decimal');
+      expect(wrapper.current.find(Textbox).prop('data-component')).toBe('decimal');
```
jamime: Previously this was making an assertion on the DOM (which is what we need for automation). Now, this only check's that the Textbox component has the prop, not that the Textbox component rendered into the DOM.

Could we revert the change please?

PR #2546, opened 2019-12-06
FE-2463 Fixes incorrect position of a label in relation to a text input when inline

grabkowski:
When an input based component has `labelInline` prop, the first row of a label should be positioned on the same level as an input text related to that label.

Issue fixed by adding padding to an inline label.


harpalsingh:
Most look fine, but the spacing from labels on radio and switch seem a lot more than they should be

Screenshot 2019-12-10 at 10 09 12

Screenshot 2019-12-10 at 10 09 55

PR #2548, opened 2019-12-09
FE-2468 - Fixed invalid animation behaviour on Decimal

mkrds:
# Description
Fixed an issue where wiggle animation started even when `readOnly` prop was passed as true to the component.

### DIFF
--- a/src/__experimental__/components/decimal/decimal.spec.js
+++ b/src/__experimental__/components/decimal/decimal.spec.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import TestRenderer from 'react-test-renderer';
 import { mount as enzymeMount } from 'enzyme';
 import I18n from 'i18n-js';
 import { assertStyleMatch } from '../../../__spec_helper__/test-utils';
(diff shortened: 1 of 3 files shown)


src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -173,6 +173,14 @@ describe('Decimal', () => {
       );
     });
 
+    it('is not triggered when Decimal has readOnly prop', () => {
+      render({ readOnly: true });
+      paste({ key: 'invalid' }, '0.|00');
```
jamime:
```suggestion
      paste({ key: 'a' }, '0.|00');
```

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -173,6 +173,14 @@ describe('Decimal', () => {
...
+      expect(wrapper.current.find(StyledWiggle).prop('isAnimating')).toBe(false);
+      press({ key: '-' }, '0.|00');
+      expect(wrapper.current.find(StyledWiggle).prop('isAnimating')).toBe(false);
```
jamime: Without triggering `onAnimationEnd` this assertion could never be `true`. Please, can you look to see if we can make these assertions without relying on the `isAnimating` prop?

PR #2551, opened 2019-12-09
FE-2430 invalid date prevent.

damianOfKyan:
# Description
resolved issue were error was thrown when an invalid min or max date was passed into the DatePicker component.


src/utils/helpers/date/date.js
```diff
@@ -34,6 +34,17 @@ const DateHelper = {
...
 
+  /**
+   * Determins if value length is valid
+   *
+   * @param {String} value - value to validate
+   * @param {Integer} valueLength - the number that determines valid length
+   * @return {Boolean}
+   */
+  isValidLength: (value, validLength = 0) => {
+    return (value.length >= validLength);
```
edleeks87: I think this could be updated to be just `>` rather than `>=` or you could update the getDisabledDays function to be 
```
if (minDate && DateHelper.isValidLength(minDate, 10) && DateHelper.isValidDate(minDate)) 
```
as the iso format (YYYY-MM-DD) is 10 digits.

Everything else looks good and I think it will still work as is but might be worth just tightening it up a little

damianOfKyan: we have `allowEmptyValue` prop on date component, if it is set to `true` and then isValidLength would fail when having `>` instead of `>=`. I think what I proposed is best option, hope that makes sense. Please correct me if I am wrong!

edleeks87: Ah yeah good point it does need to be `>=` I think it isn't essential to update  it to `DateHelper.isValidLength(minDate, 10)` then as the `isValidDate` catches anything that is only 9 chars long

change_log/next/FE-2430-prevent-invalid-date.yml
```diff
@@ -0,0 +1 @@
+Bug Fixes: "Resolved issue were moment() error was thrown when an incpmplete/invalid min or max date was passed. (Component: DateInput)"
```
edleeks87:
```suggestion
Bug Fixes: "Resolved issue where moment() error was thrown when an incomplete/invalid min or max date was passed. (Component: DateInput)"
```

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -109,17 +109,25 @@ function getDisabledDays(minDate, maxDate) {
...
 
   return days;
 }
 
+function checkIsoFormatLength(date) {
+  if (date.length !== 10 || !DateHelper.isValidDate(date, { defaultValue: 'YYYY-MM-DD' })) {
+    return false;
+  }
+  const array = date.split('-');
+  return array.length === 3 && array[0].length === 4 && array[1].length === 2 && array[2].length;
```
kyanmichael: The existing code is fine, but for extra clarity, I think we could make this improvement:
```suggestion
  return array.length === 3 && array[0].length === 4 && array[1].length === 2 && array[2].length === 2;
```

edleeks87: Good spot as always @kyanmichael, fixed in https://github.com/Sage/carbon/pull/2551/commits/bf2f8b513b1bbc673fd822391e033482513a09ed

src/__experimental__/components/date/date-picker.component.js
```diff
@@ -109,17 +109,25 @@ function getDisabledDays(minDate, maxDate) {
...
 
-  if (maxDate) {
+  if (maxDate && checkIsoFormatLength(maxDate)) {
     days.push({ after: DateHelper.stringToDate(maxDate) });
   }
 
   return days;
 }
 
+function checkIsoFormatLength(date) {
```
ianforster-sage: This function is doing two things, checking the length and verifying the length of each section. I'm wondering if it should be split into two functions that do one thing each? Or at least in would be better renamed as `verifyIsoFormat()`.

src/__experimental__/components/date/date-picker.component.js
kyanmichael: Thanks :white_check_mark:

PR #2553, opened 2019-12-10
chore: updates carbon-factory dependency

edleeks87:
BREAKING CHANGE: carbon-factory updated to v10 to support linting of
hooks and replace 'stage 0' flag in precompile script


jamime:
@edleeks87 I think this new rule will fail eslint in the build. Have you tried it?

PR #2555, opened 2019-12-10
feat: Add toggle open close to date FE-2219

DipperTheDan:
# Description
Implements functionality to toggle picker opening and closing when calendar icon is clicked.


src/__experimental__/components/date/date.component.js
```diff
@@ -191,15 +192,31 @@ class BaseDateInput extends React.Component {
...
   };
 
+  handleClick = () => {
+    this.setState(_prevState => ({
+      shouldPickerOpen: !_prevState.shouldPickerOpen
+    }), () => {
+      if (!this.state.shouldPickerOpen) {
+        this.inputFocusedViaPicker = false;
+        this.closeDatePicker();
+        console.log('here');
```
damianOfKyan: leftover debug code?

DipperTheDan: Yep. Jamie has pointed it out. I'll get it removed now.

src/__experimental__/components/date/date.component.js
```diff
@@ -191,15 +192,31 @@ class BaseDateInput extends React.Component {
...
         this.handleBlur();
       }
     });
   };
 
+  handleClick = () => {
+    this.setState(_prevState => ({
+      shouldPickerOpen: !_prevState.shouldPickerOpen
+    }), () => {
+      if (!this.state.shouldPickerOpen) {
```
damianOfKyan: just a thought.. I think it would be more readable if you re-aranged if/else statement here. You have check for Falsy in if and then drop into else. Maybe switch it around? Have a Thruthy IF?

src/__experimental__/components/date/date.spec.js
```diff
@@ -569,6 +570,16 @@ describe('Date', () => {
   });
 });
 
+describe('when the calendar icon is clicked and the picker is open', () => {
```
damianOfKyan: is there alternative to that? Maybe worth adding test the behaviour of two clicks. 1st click closes the picker, 2nd click re-opens the picker?

src/__experimental__/components/date/date.component.js
DipperTheDan: Good point. I'll re-write it so the IF statement is truthy and then the ELSE deals with the falsy. Thank you.

src/__experimental__/components/date/date.spec.js
```diff
@@ -480,6 +480,16 @@ describe('Date', () => {
       });
     });
 
+    it('short circuits when either "disabled" and/or "readOnly"', () => {
```
damianOfKyan: I think the correct implementation of that test would be to have separate `it` block for every permutation of these props.
1. disabled
2. readOnly
3. disabled and readOnly
4. neither (probably optional)

src/__experimental__/components/date/date.spec.js
```diff
@@ -480,6 +480,16 @@ describe('Date', () => {
...
+      wrapper = render({});
+      wrapper.find(InputIconToggle).props().onClick();
+      wrapper.setProps({ disabled: true, readOnly: true });
```
damianOfKyan: It would seem that calling .setProps is necessary here, simply set these props in render like so `wrapper = render({ disabled: true, readOnly: true });

DipperTheDan: This was to address an issue I noticed in Storybook, where you can focus the input and then set a `disabled` or `readOnly` knob and still interact with the picker after. The props need to be set in this way to test the scenario.

damianOfKyan: Ok that makes sense then :)

src/__experimental__/components/date/date.spec.js
```diff
@@ -479,6 +480,35 @@ describe('Date', () => {
       });
     });
 
+    // This test is for a specific bug found in Storybook where the component can be focused,
+    // the knobs for readOnly and/or disabled can then be selected and the date component can then
+    // still be interacted with.
+    it('short circuits when either "disabled" and/or "readOnly"', () => {
```
edleeks87: @DipperTheDan this test can be removed lines 486 -> 494

src/__experimental__/components/date/date.spec.js
```diff
@@ -479,6 +480,35 @@ describe('Date', () => {
...
+      wrapper.update();
+      expect(spyOn(wrapper.find(BaseDateInput).instance(), 'openDatePicker')).not.toBeCalled();
+    });
+
+    describe.each([
+      { disabled: true, readOnly: false },
+      { disabled: false, readOnly: true },
+      { disabled: true, readOnly: true }
+    ])('The date picker', (props) => {
+      it(`when disabled is ${props.disabled} and readOnly is ${props.readOnly}`, () => {
```
edleeks87:
```suggestion
      it(` does not call "openDatePicker" when disabled is ${props.disabled} and readOnly is ${props.readOnly}`, () => {
```

src/__experimental__/components/date/date.spec.js
DipperTheDan: Yep. Just realised that too. The purpose of the new test was to replace this.

mkrds:
Unfortunately DateInput does not work properly after applying these changes. Try opening the DatePicker, pick one of the dates, and then try opening it once again by clickin on the field.

PR #2558, opened 2019-12-12
FE-2451: fixes #2500 renders a button as Link when onClick passed

### LINKED ISSUE #2500: Link component when provided with a onClick function causes accessibility error
When using the Link component with an onClick function it causes accessibility errors to be raised.

Code specific error in our linter: Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead.eslint(jsx-a11y/anchor-is-valid)

This is also reported by Axe

Looks like this has been raised previous here and closed: https://github.com/Sage/carbon/issues/1915

edleeks87:
### Breaking Changes: 
* The link component now renders a `<button>` when an `onClick` prop is passed without a `to` prop to fix accessibility errors. (Component: Link)
### Styling Changes: 
* Removes text decoration from links in Heading and adds margin to fix regressed alignments. (Component: Heading, Pages)


src/components/heading/heading.style.js
```diff
@@ -1,10 +1,30 @@
...
+import styled, { css } from 'styled-components';
 import Icon from '../icon';
 import StyledIcon from '../icon/icon.style';
 import baseTheme from '../../style/themes/base';
+import StyledLink from '../link/link.style';
+import { isClassic } from '../../utils/helpers/style-helper';
 
 const StyledHeading = styled.div`
   width: 100%;
+  &&&& ${StyledLink} {
```
kyanmichael: `&&&&` – that's quite a lot of [specificity](https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity)!

src/components/heading/heading.style.js
```diff
@@ -1,10 +1,30 @@
...
+    ${({ divider, subheader, theme }) => !isClassic(theme) && css`
+      margin-top: ${(!divider ? '-16px' : '')};
+      margin-top: ${(!subheader ? '-22px' : '')};
+      margin-top: ${(!divider && !subheader ? '-12px' : '')};
+      margin-top: ${(divider && !subheader ? '-14px' : '')};
+    `}
+
+    ${({ divider, subheader, theme }) => isClassic(theme) && css`
+        margin-top: ${!divider && subheader ? '-14px' : ''};
+        margin-top: ${!divider && !subheader ? '-10px' : ''};
```
kyanmichael: Indentation needs to be reduced by 2 spaces.

mkrds:
No other comments than the ones pointed out by Michael

src/components/heading/heading.style.js
edleeks87: I'm a specific kind of guy 😉, good point was trying out something will scale it back to `&&`

edleeks87: I think it needs to be that specific as it doesn't override the styling as needed when the `Heading` is used in `Pages` as `margin-top: -8px;` is applied in the `.scss` file

kyanmichael: OK :thumbsup:

src/components/heading/heading.style.js
kyanmichael: :white_check_mark:

src/components/link/link.stories.js
```diff
@@ -49,7 +50,7 @@ function makeStory(name, themeSelector) {
         tooltipMessage={ tooltipMessage }
         tooltipPosition={ tooltipPosition }
         tooltipAlign={ tooltipAlign }
-        onClick={ action('click') }
+        onClick={ hasOnClick && action('click') }
```
kyanmichael: When `hasOnClick` is `false`, then `hasOnClick && action('click')` will evaluate to `false` which will be passed to the `onClick` prop of `<Link>`.

In `<Link>`, the `propType` of `onClick` is `func`:

* https://github.com/Sage/carbon/blob/180f948/src/components/link/link.component.js#L141

So will this cause a `propType` validation failure if we pass `false` when it's expecting a function?

edleeks87: Yeah good point have updated it to pass undefined if false

kyanmichael: Perfect, thanks :white_check_mark:

PR #2562, opened 2019-12-13
FE-2450 adds support for src prop in Profile

DipperTheDan:
# Description
adds support for src prop in Profile image

### DIFF
--- a/change_log/next/FE-2450-Add_src_prop_to_Profile_component.yml
+++ b/change_log/next/FE-2450-Add_src_prop_to_Profile_component.yml
@@ -1,0 +1,1 @@
+New Features: "adds support for src prop in Profile image. (component: Profile)"
--- a/src/components/profile/docgenInfo.json
+++ b/src/components/profile/docgenInfo.json
@@ -92,6 +92,13 @@
           "required": false,
           "description": "Allow to setup size for the component"
         },
+        "src": {
+          "type": {
+            "name": "string"
+          },
+          "required": false,
+          "description": "Import an image from a scr url"
+        },
         "theme": {
           "type": {
             "name": "object"
--- a/src/components/profile/docgenInfo.json
+++ b/src/components/profile/docgenInfo.json
@@ -97,7 +97,7 @@
             "name": "string"
           },
           "required": false,
-          "description": "Import an image from a scr url"
+          "description": "Import an image from a src url"
         },
         "theme": {
           "type": {
--- a/src/components/profile/profile.component.js
+++ b/src/components/profile/profile.component.js
@@ -86,13 +86,7 @@ Profile.propTypes = {
   /** [Legacy] A custom class name for the component */
   className: PropTypes.string,
   /* Custom source URL */
-  src: ({ src }) => {
-    if (src && typeof src !== 'string') {
-      throw new Error(
-        `Invalid prop \`src\` of type \`${typeof src}\` supplied to \`Profile\`, expected \`string\`.`
-      );
-    }
-  },
+  src: PropTypes.string,
   /** Define the name to display. */
   name: PropTypes.string.isRequired,
   /** Define the email to use (will check Gravatar for image). */
--- a/src/components/profile/profile.component.js
+++ b/src/components/profile/profile.component.js
@@ -85,7 +85,7 @@ class Profile extends React.Component {
 Profile.propTypes = {
   /** [Legacy] A custom class name for the component */
   className: PropTypes.string,
-  /* Custom source URL */
+  /** Custom source URL */
   src: PropTypes.string,
   /** Define the name to display. */
   name: PropTypes.string.isRequired,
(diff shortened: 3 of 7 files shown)


change_log/next/FE-2450-Add_src_prop_to_Profile_component.yml
```diff
@@ -0,0 +1 @@
+New Features: "adds support for src prop in Profile image. (component: Profile)"
```
ianforster-sage: The RFC for semantic release dictates that we use the the imperative, present tense, so please could you do this in the commit message and it makes sense to be consistent here.
```suggestion
New Features: "add support for src prop in Profile image. (component: Profile)"
```

src/components/profile/profile.stories.js
```diff
@@ -21,13 +21,15 @@ storiesOf('Profile', module)
...
     const name = text('name', 'John Smith');
+    const src = text('src', '');
 
     return (
       <ProfileWrapper
         email={ email }
         initials={ initials }
         size={ size }
         name={ name }
+        src={ src }
```
ianforster-sage: You need to regenerate docGenInfo.json and add it to this PR so the props table picks up the `src` prop.

src/components/profile/profile.component.js
```diff
@@ -15,11 +15,14 @@ import { isClassic } from '../../utils/helpers/style-helper';
 import baseTheme from '../../style/themes/base';
 
 class Profile extends React.Component {
+  /** Returns the hasSrc component */
```
kyanmichael: Maybe something like this would be better?
```suggestion
  /** Determines whether a `src` prop has been supplied */
```

src/components/profile/profile.component.js
```diff
@@ -30,6 +33,18 @@ class Profile extends React.Component {
 
   /** Returns the avatar portion of the profile. */
   get avatar() {
+    if (this.props.src) {
```
kyanmichael: I guess this could be replaced with:
```suggestion
    if (this.hasSrc) {
```

src/components/profile/profile.component.js
```diff
@@ -69,6 +85,14 @@ class Profile extends React.Component {
 Profile.propTypes = {
   /** [Legacy] A custom class name for the component */
   className: PropTypes.string,
+  /* Custom source URL */
+  src: ({ src }) => {
```
kyanmichael: What's the purpose of having this validation function? Couldn't we use a simple `string` check instead?
```js
src: PropTypes.string
```

src/components/profile/profile.spec.js
```diff
@@ -122,9 +120,35 @@ describe('ProfileAvatarStyle', () => {
     const wrapper = shallow(<Profile
       initials='AS'
       theme={ classicTheme }
+      email='foo'
+      name='foo'
+    />);
+
+    expect(wrapper.find(ProfileAvatarStyle).props().size).toEqual('medium-small');
+  });
+  it('should render an user image when a "src" prop is provided', () => {
```
kyanmichael:
```suggestion
  it('should render a user image when a "src" prop is provided', () => {
```

src/components/profile/profile.spec.js
```diff
@@ -170,4 +194,22 @@ describe('ProfileDetailStyle', () => {
       marginLeft: '14px'
     }, mount(<ProfileDetailsStyle theme={ classicTheme } />));
   });
+  describe.each(
+    [['XS', '1px'], ['S', '1px'], ['M', '4px'], ['ML', '8px'], ['L', '14px'], ['XL', '24px'], ['XXL', '32px']]
+  )(
+    'when an src prop is passed and the size is %s', (size, margin) => {
```
kyanmichael:
```suggestion
    'when a src prop is passed and the size is %s', (size, margin) => {
```

src/components/profile/profile.stories.js
DipperTheDan: Cheers for making me aware this needed doing. Understand why it needs doing and now done.

src/components/profile/docgenInfo.json
```diff
@@ -92,6 +92,13 @@
           "required": false,
           "description": "Allow to setup size for the component"
         },
+        "src": {
+          "type": {
+            "name": "string"
+          },
+          "required": false,
+          "description": "Import an image from a scr url"
```
ianforster-sage:
```suggestion
          "description": "Import an image from a src url"
```

src/components/profile/profile.component.js
```diff
@@ -69,6 +85,8 @@ class Profile extends React.Component {
 Profile.propTypes = {
   /** [Legacy] A custom class name for the component */
   className: PropTypes.string,
+  /* Custom source URL */
```
ianforster-sage:
```suggestion
  /** Custom source URL */
```

PR #2569, opened 2019-12-18
feat(Dialog, Sidebar): Create focus-trap util and adapt it to the components

dawidzarzycki:
# Description
- Add a new behavior called focus-trap. It allows keeping focus inside of the selected element and it focuses first focusable element. Focus-trap also has protection - if no focusable component occurs then dialog should not lose focus also.

# Testing Instructions
- go to storybook/dialog 
- open dialog with keyboard
- the focus should be placed on the first focusable element if elements exist
- tabbing or shift tabbing shouldn't leave the selected element (dialog)

### DIFF
--- a/src/components/dialog-full-screen/dialog-full-screen.component.js
+++ b/src/components/dialog-full-screen/dialog-full-screen.component.js
@@ -7,6 +7,8 @@ import StyledDialogFullScreen from './dialog-full-screen.style';
 import StyledContent from './content.style';
 import StyledIcon from './icon.style';
 import Browser from '../../utils/helpers/browser';
+import Events from '../../utils/helpers/events/events';
+import focusTrap from '../../utils/helpers/focus-trap';
 
 class DialogFullScreen extends Modal {
   constructor(props) {
@@ -16,6 +18,7 @@ class DialogFullScreen extends Modal {
      * Caches a reference to the document.
      */
     this.document = Browser.getDocument();
+    this.originalOverflow = undefined;
   }
 
   static state = {
@@ -74,19 +77,28 @@ class DialogFullScreen extends Modal {
   /**
    * Overrides the original function to disable the document's scroll.
    */
-  get onOpening() {
+  handleOpen() {
+    super.handleOpen();
+    this.removeFocusTrap = focusTrap(this._dialog);
+    this.originalOverflow = this.document.documentElement.style.overflow;
     this.document.documentElement.style.overflow = 'hidden';
-
-    return this.document.documentElement;
   }
 
   /**
    * Overrides the original function to enable the document's scroll.
    */
-  get onClosing() {
-    this.document.documentElement.style.overflow = 'auto';
-
+  handleClose() {
+    super.handleClose();
+    this.removeFocusTrap();
+    this.document.documentElement.style.overflow = this.originalOverflow;
     return this.document.documentElement;
+  }
+
+  onButtonKeyDown = (ev) => {
+    if (Events.isEnterKey(ev) || Events.isSpaceKey(ev)) {
+      ev.preventDefault();
+      this.props.onCancel();
+    }
   }
 
   /**
@@ -112,6 +124,9 @@ class DialogFullScreen extends Modal {
           data-element='close'
           onClick={ this.props.onCancel }
           type='close'
+          tabIndex='0'
+          role='button'
+          onKeyDown={ this.onButtonKeyDown }
         />
 
         { title }
--- a/src/components/dialog-full-screen/dialog-full-screen.spec.js
+++ b/src/components/dialog-full-screen/dialog-full-screen.spec.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { mount } from 'enzyme';
 import 'jest-styled-components';
 import DialogFullScreen from './dialog-full-screen.component';
 import FullScreenHeading from './full-screen-heading';
@@ -23,7 +23,7 @@ describe('DialogFullScreen', () => {
   const onCancel = jasmine.createSpy('cancel');
 
   beforeEach(() => {
-    wrapper = shallow(
+    wrapper = mount(
       <DialogFullScreen
         onCancel={ onCancel }
         className='foo'
@@ -41,37 +41,6 @@ describe('DialogFullScreen', () => {
   describe('default props', () => {
     it('sets enableBackgroundUI to true', () => {
       expect(instance.props.enableBackgroundUI).toBeTruthy();
-    });
-  });
-
-  describe('modalHTML', () => {
-    beforeEach(() => {
-      wrapper = mount(
-        <DialogFullScreen
-          open
-          className='foo'
-          title='my title'
-          onCancel={ onCancel }
-        >
-          <Button>Button</Button>
-          <Button>Button</Button>
-        </DialogFullScreen>
-      );
-      instance = wrapper.instance();
-    });
-
-    it('renders the dialog', () => {
-      expect(instance._dialog).toBeTruthy();
-    });
-
-    it('closes when the exit icon is click', () => {
-      const closeIcon = wrapper.find(Icon);
-      closeIcon.simulate('click');
-      expect(onCancel).toHaveBeenCalled();
-    });
-
-    it('renders the children passed to it', () => {
-      expect(wrapper.find(Button).length).toEqual(2);
     });
   });
 
@@ -95,14 +64,16 @@ describe('DialogFullScreen', () => {
 
   describe('onClosing', () => {
     beforeEach(() => {
-      wrapper = mount(<DialogFullScreen open={ false } />);
-      wrapper.setProps({ open: true });
+      wrapper = mount(<DialogFullScreen style={ { overflow: 'auto' } } open={ false } />);
     });
 
-    it('removes overflow hidden from the body', () => {
-      const html = wrapper.instance().document.documentElement;
+    it('recovers an original overflow', () => {
+      window.document.documentElement.style.overflow = 'auto';
+      expect(window.document.documentElement.style.overflow).toBe('auto');
+      wrapper.setProps({ open: true });
+      expect(window.document.documentElement.style.overflow).toBe('hidden');
       wrapper.setProps({ open: false });
-      expect(html.style.overflow).not.toMatch('hidden');
+      expect(window.document.documentElement.style.overflow).toBe('auto');
     });
   });
 
@@ -120,7 +91,7 @@ describe('DialogFullScreen', () => {
     describe('is an object', () => {
       beforeEach(() => {
         const titleHeading = <Heading title='my custom heading' />;
-        wrapper = shallow(
+        wrapper = mount(
           <DialogFullScreen
             onCancel={ onCancel }
             className='foo'
@@ -145,11 +116,11 @@ describe('DialogFullScreen', () => {
   describe('tags', () => {
     describe('on component', () => {
       it('include correct component, elements and role data tags', () => {
-        wrapper = shallow(
+        wrapper = mount(
           <DialogFullScreen
             open
-            onCancel={ () => {} }
-            onConfirm={ () => {} }
+            onCancel={ () => { } }
+            onConfirm={ () => { } }
             title='Test'
             data-role='baz'
             data-element='bar'
@@ -159,6 +130,89 @@ describe('DialogFullScreen', () => {
         expect(wrapper.instance().props['data-role']).toEqual('baz');
       });
     });
+  });
+});
+
+describe('closeIcon', () => {
+  let wrapper;
+  let onCancel;
+  let preventDefault;
+
+  beforeEach(() => {
+    jest.restoreAllMocks();
+    onCancel = jest.fn();
+    preventDefault = jest.fn();
+
+    wrapper = mount(
+      <DialogFullScreen
+        open
+        onCancel={ onCancel }
+        onConfirm={ () => { } }
+        title='Test'
+        data-role='baz'
+        data-element='bar'
+      />
+    );
+  });
+
+  it('should close Dialog if enter has been pressed on Close Icon', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 13, preventDefault });
+    expect(onCancel).toHaveBeenCalled();
+  });
+
+  it('should close Dialog if ESC key has been pressed', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ key: 'Escape' });
+  });
+
+  it('should not close Dialog with any key other than Enter or ESC', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 16 });
+    expect(onCancel).not.toHaveBeenCalled();
+  });
+});
+
+describe('modalHTML', () => {
+  let instance,
+      wrapper,
+      preventDefault;
+  const onCancel = jest.fn();
+
+  beforeEach(() => {
+    preventDefault = jest.fn();
+    wrapper = mount(
+      <DialogFullScreen
+        open
+        className='foo'
+        title='my title'
+        onCancel={ onCancel }
+      >
+        <Button>Button</Button>
+        <Button>Button</Button>
+      </DialogFullScreen>
+    );
+    instance = wrapper.instance();
+  });
+
+  it('renders the dialog', () => {
+    expect(instance._dialog).toBeTruthy();
+  });
+
+  it('closes when the exit icon is click', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.simulate('click');
+    expect(onCancel).toHaveBeenCalled();
+  });
+
+  it('closes when the exit icon is pressed with enter', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 13, preventDefault });
+    expect(onCancel).toHaveBeenCalled();
+  });
+
+  afterEach(() => {
+    wrapper.unmount();
   });
 });
 
--- a/src/components/dialog/dialog.component.js
+++ b/src/components/dialog/dialog.component.js
@@ -14,13 +14,14 @@ import {
   DialogInnerContentStyle
 } from './dialog.style';
 import tagComponent from '../../utils/helpers/tags';
+import Events from '../../utils/helpers/events/events';
+import focusTrap from '../../utils/helpers/focus-trap';
 
 class Dialog extends Modal {
   constructor(args) {
     super(args);
     this.componentTags = this.componentTags.bind(this);
     this.onDialogBlur = this.onDialogBlur.bind(this);
-    this.onCloseIconBlur = this.onCloseIconBlur.bind(this);
     this.document = Browser.getDocument();
     this.window = Browser.getWindow();
   }
@@ -29,42 +30,39 @@ class Dialog extends Modal {
     super.componentDidMount();
     if (this.props.open) {
       this.centerDialog();
-      if (this.props.autoFocus) {
-        this.focusDialog();
-      }
     }
   }
 
   onDialogBlur(ev) { } // eslint-disable-line no-unused-vars
 
-  onCloseIconBlur(ev) {
-    ev.preventDefault();
-    this.focusDialog();
-  }
-
-  get onOpening() {
+  handleOpen() {
+    super.handleOpen();
     this.document.documentElement.style.overflow = 'hidden';
     this.centerDialog(true);
     ElementResize.addListener(this._innerContent, this.applyFixedBottom);
     this.window.addEventListener('resize', this.centerDialog);
-
-    if (this.props.autoFocus) {
-      return this.focusDialog();
-    }
-
-    return null;
+    this.removeFocusTrap = focusTrap(this._dialog);
   }
 
-  get onClosing() {
+  handleClose() {
+    super.handleClose();
+    this.removeFocusTrap();
     this.appliedFixedBottom = false;
     this.document.documentElement.style.overflow = '';
     this.window.removeEventListener('resize', this.centerDialog);
     return ElementResize.removeListener(this._innerContent, this.applyFixedBottom);
   }
 
-  centerDialog = (animating) => {
-    if (!this._dialog) return;
+  onButtonKeyDown = (ev) => {
+    if (Events.isEnterKey(ev) || Events.isSpaceKey(ev)) {
+      ev.preventDefault();
+      this.props.onCancel();
+    }
 
+    return null;
+  }
+
+  centerDialog = (animating) => {
     const height = this._dialog.offsetHeight / 2,
         width = this._dialog.offsetWidth / 2,
         win = this.window;
@@ -110,11 +108,6 @@ class Dialog extends Modal {
       this.appliedFixedBottom = false;
       this.forceUpdate();
     }
-  }
-
-  focusDialog() {
-    if (!this._dialog) return;
-    this._dialog.focus();
   }
 
   shouldHaveFixedBottom = () => {
@@ -169,7 +162,8 @@ class Dialog extends Modal {
           onClick={ this.props.onCancel }
           type='close'
           tabIndex='0'
-          onBlur={ this.onCloseIconBlur }
+          role='button'
+          onKeyDown={ this.onButtonKeyDown }
         />
       );
     }
@@ -208,7 +202,6 @@ class Dialog extends Modal {
     }
 
     const dialogProps = {
-      tabIndex: 0,
       style: {
         minHeight: height
       },
@@ -273,13 +266,10 @@ Dialog.propTypes = {
   size: PropTypes.string,
   /** Determines if the close icon is shown */
   showCloseIcon: PropTypes.bool,
-  /** If true then the dialog receives focus when it opens */
-  autoFocus: PropTypes.bool,
   stickyFormFooter: PropTypes.bool
 };
 
 Dialog.defaultProps = {
-  autoFocus: true,
   size: 'medium',
   showCloseIcon: true,
   ariaRole: 'dialog'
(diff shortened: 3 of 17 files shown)


src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,49 @@
...
+
+  if (focusableElements.length <= 0) {
+    document.addEventListener('keydown', blockTabbing);
+    return;
+  }
+
+  const firstFocusableElement = focusableElements[0];
+  const lastFocusableElement = focusableElements[focusableElements.length - 1];
+
+  setTimeout(() => {
```
jamime: Why is this in setTimout?

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+};
+
+const setFocusTrap = (element) => {
+  // eslint-disable-next-line max-len
+  const ALL_FOCUSABLE_ELEMENTS = 'button, [href], input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])';
+  const focusableElements = element.querySelectorAll(ALL_FOCUSABLE_ELEMENTS);
+
+  if (focusableElements.length <= 0) {
+    document.addEventListener('keydown', blockTabbing);
```
jamime: What is the purpose of blocking tabbing on the document?

dawidzarzycki: If there was a dialog without focusable content you could press the tab key and the focus move out of the dialog. So you could not press escape to close the dialog.

src/utils/helpers/focus-trap/focus-trap.js
dawidzarzycki: without setTimeout focus won't land on the first focusable item when the dialog was opening. this async resolved the problem

jamime: This relies upon react's batched updates being applied before the next "tick". React now has low/high priority scheduling using rAF. I think this is prone to errors and another approach should be used. Im happy to be corrected though.

Instead, you should add a lifecycle method to the dialog.

Side note,
W3C recommends listening to the focus change on the document and restoring the focus instead. Did you try that approach?

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/js/dialog.js

This is reactive to the DOM update instead of trying to prevent the focus change.

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,49 @@
...
+    return;
+  }
+
+  const firstFocusableElement = focusableElements[0];
```
jamime: This doesn't account for element's that change from hidden to unhidden or unmount.

src/utils/helpers/focus-trap/focus-trap.js
dawidzarzycki: setTimouet has been removed

src/utils/helpers/focus-trap/focus-trap.js
dawidzarzycki: Refactored a bit. I think it should be good right now

src/components/dialog/dialog.component.js
```diff
@@ -49,22 +55,21 @@ class Dialog extends Modal {
     this.window.addEventListener('resize', this.centerDialog);
 
     if (this.props.autoFocus) {
-      return this.focusDialog();
+      this.focusDialog();
     }
 
-    return null;
+    return setFocusTrap(this._innerContent);
```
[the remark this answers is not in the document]
dawidzarzycki: I used return because eslint shout to me that get method need to have return value.
Maybe better will be to leave it as it was in the original version with return null and just before return use setFocusTrap?
```
setFocusTrap(this._innerContent);
return null;
```
What do you think about this?

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,59 @@
+let element;
+let focusableElements;
+let firstFocusableElement;
+let lastFocusableElement;
+
+const blockTabbing = (ev) => {
+  if (ev.keyCode === 9 || ev.keyCode === 16) {
```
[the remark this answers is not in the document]
dawidzarzycki: Good point, no need to block on shift

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,59 @@
...
+    ev.preventDefault();
+  }
+};
+
+// eslint-disable-next-line max-len
+const ALL_FOCUSABLE_ELEMENTS = 'button, [href], input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])';
```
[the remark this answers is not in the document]
dawidzarzycki: You are right, passing custom values is a very good idea

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,59 @@
...
+    document.addEventListener('keydown', blockTabbing);
+    return;
+  }
+
+  firstFocusableElement.focus();
+  document.addEventListener('keydown', ev => tabbing(ev));
+};
+
+const removeFocusTrap = () => {
+  document.activeElement.blur();
```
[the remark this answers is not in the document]
dawidzarzycki: I think some of the requirements are up to the developer that uses the component.

src/components/dialog/dialog.component.js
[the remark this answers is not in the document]
ianforster-sage: Maybe it would be better for setFocusTrap to return a reference to the element that has focus?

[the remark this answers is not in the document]
ianforster-sage: @dawidzarzycki Following our discussions I'm OK to leave this function as-is.

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -7,6 +7,8 @@ import StyledDialogFullScreen from './dialog-full-screen.style';
 import StyledContent from './content.style';
 import StyledIcon from './icon.style';
 import Browser from '../../utils/helpers/browser';
+import Events from '../../utils/helpers/events/events.js';
```
ianforster-sage: For consistency you can drop the .js extension here

src/components/dialog/dialog.component.js
```diff
@@ -14,13 +14,14 @@ import {
   DialogInnerContentStyle
 } from './dialog.style';
 import tagComponent from '../../utils/helpers/tags';
+import Events from '../../utils/helpers/events/events.js';
```
ianforster-sage: ...and here

src/components/dialog/dialog.spec.js
```diff
@@ -473,18 +422,28 @@ describe('Dialog', () => {
       });
 
       it('closes when the exit icon is click', () => {
-        wrapper.find('.carbon-dialog__close').simulate('click');
+        wrapper.find('.carbon-dialog__close').at(0).simulate('click');
         expect(onCancel).toHaveBeenCalled();
       });
+
+      it('closes when the exit icont is pressed with enter', () => {
```
ianforster-sage:
```suggestion
      it('closes when the exit icon is pressed with enter', () => {
```

src/components/dialog/dialog.spec.js
```diff
@@ -473,18 +422,28 @@ describe('Dialog', () => {
...
+        wrapper.find('.carbon-dialog__close').at(0).props().onKeyDown({ which: 13 });
+        expect(onCancel).toHaveBeenCalled();
+      });
+
+      it('closes when the exit icont is pressed with enter', () => {
```
ianforster-sage:
```suggestion
      it('closes when the exit icon is pressed with enter', () => {
```

src/components/sidebar/sidebar.component.js
```diff
@@ -5,6 +5,8 @@ import Icon from '../icon';
 import Modal from '../modal';
 import { SidebarStyle, SidebarCloseStyle } from './sidebar.style';
 import './sidebar.scss';
+import FocusTrap from '../../utils/helpers/focus-trap';
+import Events from '../../utils/helpers/events/events.js';
```
ianforster-sage: ...and here

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -112,6 +116,8 @@ class DialogFullScreen extends Modal {
           data-element='close'
           onClick={ this.props.onCancel }
           type='close'
+          tabIndex='0'
+          onKeyDown={ ev => (Events.isEnterKey(ev) ? this.props.onCancel() : null) }
```
jamime: This should have `role=button` and should listen for `Enter` and `Space`. All close icons should be the same.

src/utils/helpers/focus-trap/focus-trap.spec.js
```diff
@@ -0,0 +1,115 @@
...
+          </TestComponent>,
+          { attachTo: htmlElement }
+        );
+      });
+
+      afterEach(() => {
+        wrapper.unmount();
+      });
+
+      it('should block tabbing if `tab` pressed', () => {
```
jamime: Why?

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  setFocusableItems() {
+    this.focusableElements = this.element.querySelectorAll(this.focusableSelectors);
+    this.firstFocusableElement = this.focusableElements[0];
+    this.lastFocusableElement = this.focusableElements[this.focusableElements.length - 1];
+  }
+
+  setFocusToFirstFocusableElement() {
```
jamime: This breaks the `autoFocus` functionality provided by input components. 
E.g if you have a dialog, with a `Textbox` with `autoFocus` set, it will focus the close button instead of the `TextBox`.

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,80 @@ describe('DialogFullScreen', () => {
...
+    );
+  });
+
+  it('should close Icon with enter key', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 13 });
+    expect(spy).toHaveBeenCalled();
+  });
+
+  it('should not close Icon with other key than enter', () => {
```
damianOfKyan: What about ESC key, shouldn't we handle closing when either ESC and Enter are used?

src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+      } else if (document.activeElement === this.lastFocusableElement) {
+        this.firstFocusableElement.focus();
+        ev.preventDefault();
+      }
+    }
+  }
+
+  setFocusTrap(focusableSelectors = this.focusableSelectors) {
+    // this.element = el;
```
damianOfKyan: can remove?

src/utils/helpers/focus-trap/focus-trap.spec.js
```diff
@@ -0,0 +1,115 @@
+import React, { useEffect } from 'react';
+import { mount } from 'enzyme';
+import FocusTrap from './focus-trap';
+
+// eslint-disable-next-line
```
damianOfKyan: is this needed here?

src/utils/helpers/focus-trap/focus-trap.js
dawidzarzycki: You are right but new behavior is done because of the new acceptance criteria of this component
```When a dialog opens, place focus on the first focussable element within it.```

File: src/utils/helpers/focus-trap/focus-trap.spec.js (hunk 14 above)
Comment (dawidzarzycki) [reply src/utils/helpers/focus-trap/focus-trap.spec.js]: To don't let user tab out of the component when there is no focusable items inside

File: src/utils/helpers/focus-trap/focus-trap.spec.js (hunk 18 above)
Comment (dawidzarzycki) [reply src/utils/helpers/focus-trap/focus-trap.spec.js]: Yes, I didn't want to show eslint error about propTypes. This is a component created just for the test case

File: src/utils/helpers/focus-trap/focus-trap.js (hunk 17 above)
Comment (dawidzarzycki) [reply src/utils/helpers/focus-trap/focus-trap.js]: Good point

File: src/components/dialog-full-screen/dialog-full-screen.spec.js (hunk 16 above)
Comment (dawidzarzycki) [reply src/components/dialog-full-screen/dialog-full-screen.spec.js]: You are right. Probably the test is wrongly described. ESC works in the component very well :)

Comment (damianOfKyan) [reply src/components/dialog-full-screen/dialog-full-screen.spec.js]: please can you update that description then? Or/And maybe setup another it block testing to test that indeed ESC key closes (same way you have enter key tested on line 154)

File: src/components/dialog-full-screen/dialog-full-screen.component.js (hunk 13 above)
Comment (dawidzarzycki) [reply src/components/dialog-full-screen/dialog-full-screen.component.js]: Fixed

File: src/components/dialog/dialog.component.js [hunk 19]
```diff
@@ -112,150 +100,144 @@ class Dialog extends Modal {
...
+ get closeIcon() {
+   if (this.props.showCloseIcon) {
+     return (
+       <Icon
+         className='carbon-dialog__close'
+         data-element='close'
+         onClick={ this.props.onCancel }
+         type='close'
+         tabIndex='0'
+         onKeyDown={ ev => (Events.isEnterKey(ev) ? this.props.onCancel() : null) }
```
jamime: This should have role=button and should listen for Enter and Space. All close icons should be the same.

src/components/sidebar/sidebar.component.js
```diff
@@ -25,13 +27,31 @@ class Sidebar extends Modal {
             data-element='close'
             onClick={ this.props.onCancel }
             type='close'
+            tabIndex='0'
+            onKeyDown={ ev => (Events.isEnterKey(ev) ? this.props.onCancel() : null) }
```
jamime: This should have role=button and should listen for Enter and Space. All close icons should be the same.

jamime:
@dawidzarzycki UX has confirmed that focus should go to the first element unless a component has `autoFocus` set.

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -74,21 +76,32 @@ class DialogFullScreen extends Modal {
...
    * Overrides the original function to disable the document's scroll.
    */
-  get onOpening() {
-    this.document.documentElement.style.overflow = 'hidden';
+  onOpening() {
+    this.focusTrap = new FocusTrap(this._dialog);
+    this.focusTrap.setFocusTrap();
 
-    return this.document.documentElement;
+    this.document.documentElement.style.overflow = 'hidden';
```
jamime: We should store the original value of `overflow` in case someone has already changed it.

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -74,21 +76,32 @@ class DialogFullScreen extends Modal {
...
   }
 
   /**
    * Overrides the original function to enable the document's scroll.
    */
   get onClosing() {
+    this.focusTrap.removeFocusTrap();
     this.document.documentElement.style.overflow = 'auto';
```
jamime: We should restore the stored value of `overflow` here.

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
 });
 
+describe('closeIcon', () => {
+  let wrapper;
+  let spy;
+  let preventDefault;
+
+  beforeEach(() => {
+    jest.restoreAllMocks();
+    spy = jest.fn();
```
jamime:
```suggestion
    onCancel = jest.fn();
```

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
+    preventDefault = jest.fn();
+
+    wrapper = mount(
+      <DialogFullScreen
+        open
+        onCancel={ spy }
```
jamime:
```suggestion
        onCancel={ onCancel }
```

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
+        data-role='baz'
+        data-element='bar'
+      />
+    );
+  });
+
+  it('should close Dialog if enter has been pressed on Close Icon', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 13, preventDefault });
+    expect(spy).toHaveBeenCalled();
```
jamime:
```suggestion
    expect(onCancel).toHaveBeenCalled();
```

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
+
+  it('should close Dialog if ESC key has been pressed', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ key: 'Escape' });
+  });
+
+  it('should not close Icon with other key than enter or ESC', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ which: 16 });
+    expect(spy).not.toHaveBeenCalled();
```
jamime:
```suggestion
    expect(onCancel).not.toHaveBeenCalled();
```

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
+  });
+});
+
+describe('modalHTML', () => {
+  let instance,
+      wrapper,
+      preventDefault;
+  const onCancel = jasmine.createSpy('cancel');
```
jamime:
```suggestion
  const onCancel = jest.fn()
```

src/components/dialog/dialog.component.js
```diff
@@ -29,42 +30,42 @@ class Dialog extends Modal {
...
     this.window.addEventListener('resize', this.centerDialog);
-
-    if (this.props.autoFocus) {
-      return this.focusDialog();
-    }
-
-    return null;
+    this.focusTrap = new FocusTrap(this._dialog);
+    this.focusTrap.setFocusTrap();
+    this.focusTrap.setFocusToFirstFocusableElement(this.props.firstFocus);
```
jamime: I see the user now has to do `<Dialog focusFirst>` I know this was added for the case when a user wants to focus the first element. However, I think we should simplify this by keeping the existing `<Input autoFocus>` props.

At the moment if someone had 

```
<Dialog focusFirst>
<Input>
<Input autoFocus>
</Dialog>
```
 it would cause problems, I think the interface is simpler if we remove the focus functionality from `focusTrap`. What do you think?

dawidzarzycki: I have removed first focus

src/components/dialog-full-screen/dialog-full-screen.spec.js
```diff
@@ -162,6 +131,89 @@ describe('DialogFullScreen', () => {
...
+    closeIcon.props().onKeyDown({ which: 13, preventDefault });
+    expect(onCancel).toHaveBeenCalled();
+  });
+
+  it('should close Dialog if ESC key has been pressed', () => {
+    const closeIcon = wrapper.find(Icon);
+    closeIcon.props().onKeyDown({ key: 'Escape' });
+  });
+
+  it('should not close Icon with other key than enter or ESC', () => {
```
damianOfKyan:
```suggestion
  it('should not close Dialog with any key other than Enter or ESC', () => {
```

src/components/dialog/dialog.spec.js
```diff
@@ -473,18 +424,28 @@ describe('Dialog', () => {
       });
 
       it('closes when the exit icon is click', () => {
-        wrapper.find('.carbon-dialog__close').simulate('click');
+        wrapper.find('.carbon-dialog__close').at(0).simulate('click');
+        expect(onCancel).toHaveBeenCalled();
+      });
+
+      it('closes when the exit icon is pressed with enter', () => {
```
damianOfKyan:
```suggestion
      it('closes when exit icon is focused and Enter key is clicked', () => {
```

src/components/dialog/dialog.spec.js
```diff
@@ -473,18 +424,28 @@ describe('Dialog', () => {
...
+        wrapper.find('.carbon-dialog__close').at(0).props().onKeyDown({ which: 13, preventDefault });
         expect(onCancel).toHaveBeenCalled();
       });
+
+      it('closes when the exit icon is pressed with enter', () => {
```
damianOfKyan:
```suggestion
      it('closes when exit icon is focused and Enter key is clicked', () => {
```

src/components/sidebar/sidebar.component.js
```diff
@@ -25,13 +36,29 @@ class Sidebar extends Modal {
...
 
+  onOpening() {
+    this.focusTrap = new FocusTrap(this.sideBarRef);
+
+    if (!this.props.enableBackgroundUI) {
+      this.focusTrap.setFocusTrap();
+    }
+  }
+
+  get onClosing() {
```
jamime:
```suggestion
  onClosing() {
```

src/components/sidebar/sidebar.component.js
```diff
@@ -25,13 +36,29 @@ class Sidebar extends Modal {
...
+    this.focusTrap.removeFocusTrap();
+    return null;
```
jamime:
```suggestion
```

src/components/sidebar/sidebar.spec.js
```diff
@@ -68,10 +90,24 @@ describe('Sidebar', () => {
...
     });
+
+    describe('pressing Tab key on the close icon sidebar', () => {
+      it('closes the sidebar', () => {
+        const icon = wrapper.find('.carbon-sidebar__close-icon');
+        icon.at(0).props().onKeyDown({ which: 13, preventDefault });
+        expect(spy).toHaveBeenCalled();
+      });
+
+      it('does not close the sidebar', () => {
```
jamime: The describe says `pressing Tab key on the close icon sidebar` then there are two contradicting `it` blocks. Should it close or not?

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -74,21 +77,32 @@ class DialogFullScreen extends Modal {
   /**
    * Overrides the original function to disable the document's scroll.
    */
-  get onOpening() {
+  onOpening() {
+    this.focusTrap = new FocusTrap(this._dialog);
```
jamime: We always call the following two lines together and the only other method that we call is `removeFocusTrap`.
```js
this.focusTrap = new FocusTrap(this._dialog);
this.focusTrap.setFocusTrap()
```

Could we change it so the interface is

```js
onOpening(){
  this.removeFocusTrap = focusTrap(this._dialog);
}

onClosing(){
  this.removeFocusTrap();
}
```
e.g.

```js
focusTrap(el){
  const Cls = new FocusTrap(cls);
  return Cls.removeFocusTrap.bind(cls);
}
```

What do you think?

src/components/dialog/dialog.component.js
```diff
@@ -273,13 +265,10 @@ Dialog.propTypes = {
   size: PropTypes.string,
   /** Determines if the close icon is shown */
   showCloseIcon: PropTypes.bool,
-  /** If true then the dialog receives focus when it opens */
-  autoFocus: PropTypes.bool,
   stickyFormFooter: PropTypes.bool
 };
 
 Dialog.defaultProps = {
-  autoFocus: true,
```
jamime: If people used to have `autoFocus=false` this would no longer work. It's important that we communicate this in a `BREAKING CHANGE` commit.

src/components/modal/modal.component.js
```diff
@@ -63,7 +63,7 @@ class Modal extends React.Component {
   handleOpen() {
     this.listening = true;
     this.updateDataState();
-    this.onOpening; // eslint-disable-line no-unused-expressions
+    this.onOpening(); // eslint-disable-line no-unused-expressions
```
jamime: Instead of defining an empty `onOpening` method we should override `handleOpen` in the subclass and call the `super()`.

src/components/dialog-full-screen/dialog-full-screen.component.js
dawidzarzycki: After I have removed the method `setFocusToFirstElement` there is nonsense to force people to use `focusTrap and then setFocusTrap`. I'll change this

change_log/next/FE-1455-dialog_FE-1648-sidebar.yml
```diff
@@ -0,0 +1,2 @@
+New Features: "Add trap-focus. It allows to keep focus inside of the element while tabbing.
```
jamime: Please remove this file and add the detail into the commit message. We no longer use renogen.

dawidzarzycki: done

harpalsingh:
@dawidzarzycki Testing this, the focus trap works and keeps focus inside the dialog, however in storybook, after using keyboard to open the dialog, it takes 2 tabs get get focus inside the dialog, is this a bug or behaviour expected in storybook?

dawidzarzycki:
> @dawidzarzycki Testing this, the focus trap works and keeps focus inside the dialog, however in storybook, after using keyboard to open the dialog, it takes 2 tabs get get focus inside the dialog, is this a bug or behaviour expected in storybook?

@harpalsingh This is expected behavior. The focus when the dialog is open is up to the developer that uses the component. So if the developer wants to have focused an item when the dialog is open he should manage it by himself

PR #2571, opened 2019-12-18
Update lint script

dawidzarzycki:
It allows running linter on `./src` and `./fixtures` directory


jamime:
I've updated codeship to use the new command.

PR #2583, opened 2020-01-08
chore: update storybook to 5.3.3 & implement addon-docs - FE-1998

dawidzarzycki:
### Proposed behaviour
Add the new version of a storybook that allows use addon-docs to create better documentation

### Current behaviour
Older version of a storybook that doesn't support addon-info

### Checklist

- [x] Release notes
<del>- [ ] Unit tests
<del>- [ ] Theme support

### Additional context
FE-1998

### Testing instructions
- Go to storybook
- pick any of component
- go to DOCS tab
- error screen shouldn't be visible
- screen with information about DOCS should be visible


change_log/next/FE-1998-storybook-update.yml
```diff
@@ -0,0 +1 @@
+Dependency Updates: Updates storybook to 5.2.8v. Adds storybook addon-docs.
```
ianforster-sage:
```suggestion
Dependency Updates: Updates storybook to v5.2.8. Adds storybook addon-docs.
```

src/components/button-toggle/button-toggle.stories.js
```diff
@@ -33,7 +33,7 @@ function makeStory(name, themeSelector) {
     };
 
     return (
-      [
+      <>
```
ianforster-sage: Using React Fragments like this Ie. declaring what look like empty elements causes the Storybook addon to output this message.
```
"Unknown" Component
No propTypes defined!
```
It's better to use a `<div>` here, as there's no problem for us to add the extra element to the DOM and it cleans up that Unknown Component output.

src/components/card/card.stories.js
```diff
@@ -115,7 +115,11 @@ function makeStory(name, themeSelector) {
   const component = () => {
     const knobs = cardKnobs();
 
-    return ([getCard(knobs), getSmallCard(knobs)]);
+    return (
+      <>
```
ianforster-sage: Same as my earlier comment about empty elements.

edleeks87:
Do we still need a change_log now file now semantic releasing has been implemented?

jamime:
@edleeks87 we do not, it is generated by `@semantic-release/changelog`.

src/components/button-toggle/button-toggle.stories.js
ianforster-sage: @damianOfKyan this issue has reappeared. Please could you resolve it.

src/components/card/card.stories.js
ianforster-sage: This has reappeared too.

src/components/button-toggle/button-toggle.stories.js
damianOfKyan: fixed in d1a866b

src/components/card/card.stories.js
damianOfKyan: fixed in d1a866b

damianOfKyan:
> There is a message in the console `Please use '@storybook/addon-docs/preset' instead of '@storybook/addon-docs/react/preset'.`
fixed in 8115344

PR #2586, opened 2020-01-10
added id prop

igor-strelkov:
### Proposed behavior
ID prop is required on the form against the panel that is created that houses the sticky buttons

### Current behavior
There is currently no such ID

### Related Issues / Pull Requests
FE-2486

### DIFF
--- a/src/__experimental__/components/form/form.component.js
+++ b/src/__experimental__/components/form/form.component.js
@@ -485,6 +485,8 @@ function calculateCsrfValues(doc) {
 }
 
 FormWithoutValidations.propTypes = {
+  /** Unique Identifier for the form */
+  id: PropTypes.string,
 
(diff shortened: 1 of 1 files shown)


jamime:
@igor-strelkov please use the issue template provided. In this case, it would have identified that there have been no unit tests added and the typescript definition has not been updated.

I've just looked at the ticket description and asked @ianforster-sage to check, but it looks like the required functionality can be achieved without adding more ID's.

jamime:
@igor-strelkov I've confirmed that the previous commit can be discarded, and we can add `data-element="sticky-footer"` to `StyledFormFooter` in `src/_experimental_/components/form/form.component.js`

igor-strelkov:
the specifications have been modified and this commit is no longer necessary

PR #2588, opened 2020-01-10
FE-2503 adds support for value or defaultValue being an object

edleeks87:
### Proposed behaviour
Proposed to add support for allowing the Select to be passed either a string, an object formatted like
`{ value: '', text: '' }`, an array of strings or an array of the above object format when the component is in multi-select mode

### Current behaviour
Currently the  Select component only supports being passed string or array of strings as a value or defaultValue

### Checklist

- [ /] Release notes (renogen)
- [ /] Unit tests
<del>- [ /] Cypress automation tests</del>
<del>- [ /] Storybook added or updated</del>
<del>- [ /] Theme support</del>
- [ /] Typescript `d.ts` file added or updated

### Additional context
The proposed change is to support projects who do not populate the option list unless the a user performs an action to indicate they wish to change the initial value, at that point the values or lazy loaded via a call to an api.

### Testing instructions

I have created `test-branch-for-FE-2503` to allow it to be tested, the fixtures for the Select component have been adjusted to simulate the component being initialised with values but no options and then a button can be clicked to simulate an api call being made
https://github.com/Sage/carbon/tree/test-branch-for-FE-2503

### DIFF
--- a/change_log/next/FE-2503-update-selct-to-support-object-value.yml
+++ b/change_log/next/FE-2503-update-selct-to-support-object-value.yml
@@ -1,0 +1,3 @@
+New Features: "Loosens the prop-typing for `value` and `defaultValue`. The component now supports being initialized\n
+ with string and object (`{ value '', text: ''}`), to support projects that only populate the option list when the\n
+ component is interacted with. (Component: Select)"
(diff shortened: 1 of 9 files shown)


edleeks87:
Closed as no longer necessary to support object values

PR #2599, opened 2020-01-17
test(alert, animated button): add new tests for Alert and Alert Classic files

DlgSHi:
### Additional context
- add missing tests for Alert feature;
- add tests to Alert Classic feature;
- animated button tests cover all component functionality;
- refactor for label function;
- update for few scenario descriptions.


cypress/locators/configurable-items/locators.js
```diff
@@ -1,3 +1,2 @@
 // component preview locators
-export const COMMON_INPUT_LABEL = 'label[data-element="label"]';
 export const ITEMS_WRAPPER = 'ol[data-element="configurable-items-wrapper"]';
```
KatarzynaQA: If we can remove 'ol' for ITEMS_WRAPPER, please remove it

```suggestion
export const ITEMS_WRAPPER = '[data-element="configurable-items-wrapper"]';
```

DlgSHi: updated

PR #2601, opened 2020-01-17
feat(search): create search component - FE-2332

DipperTheDan:
### Proposed behaviour

Creates a new Search component.

### Current behaviour

There is no Search component.
### Checklist

- [x] Unit tests
- [x] Theme support (Only one theme)

### Additional context

Fixes FE-2332
### Testing instructions

Look at the `experimental/Search` stories and the fixture for search (No. 15).

### DIFF
--- a/cypress/features/regression/test/search.feature
+++ b/cypress/features/regression/test/search.feature
@@ -1,0 +1,90 @@
+Feature: Search component
+  I want to change Search component properties
+
+  Background: Open Search component page
+    Given I open "Test Search" component page basic
+
+  @positive
+  Scenario Outline: Set placeholder to <placeholder>
+    When I set placeholder to "<placeholder>"
+    Then Search component placeholder is set to "<placeholder>"
+    Examples:
+      | placeholder             |
+      | Sample text             |
+      | áéíóú¿¡üñ               |
+      | !@#$%^*()_+-=~[];:.,?{} |
+      | ÄÖÜßäöüß                |
+      | <>                      |
+
+  @positive
+  Scenario Outline: Verify inner text is set to <text> and threshold to <threshold> and search icon has golden border
+    Given I set threshold to "<threshold>"
+      And Type "<text>" text into search input
+    When I click onto search icon
+    Then search icon has golden border
+    Examples:
+      | text | threshold |
+      | t    | 1         |
+      | te   | 2         |
+      | test | 4         |
+
+  @positive
+  Scenario: searchButton property is enabled
+    When Type "S" text into search input
+    Then search icon as button is visible
+
+  @positive
+  Scenario: searchButton property is disabled
+    Given I uncheck searchButton checkbox
+    When Type "S" text into search input
+    Then search icon as button is not visible
+
+  @positive
+  Scenario: Verify inner elements in Search component
+    # commented because of BDD default scenario Given - When - Then
+    # When I open "Experimental Search" component page
+    Then Search component has input and "search" as icon
+
+  @positive
+  Scenario: Verify inner elements in Search component after change icon
+    When Type "Search" text into search input
+    Then Search component has input and "cross" as icon
+      And Search component input has golden border
+
+  @positive
+  Scenario: Verify golden outline for search icon
+    Given Type "Sea" text into search input
+    When I click onto search icon
+    Then search icon has golden border
+
+  @positive
+  Scenario: Check the change event for Search component
+    Given clear all actions in Actions Tab
+    When Type "Search" text into search input
+    Then change action was called in Actions Tab
+
+  @positive
+  Scenario: Check the blur event for Search component
+    Given clear all actions in Actions Tab
+      And I click inside input
+    When I click outside of the component in DLS directory
+    Then blur action was called in Actions Tab
+
+  @positive
+  Scenario: Verify proper color for search icon button
+    Given Type "Sea" text into search input
+    When I click onto search icon
+    Then search icon has proper inner color
+
+  @positive
+  Scenario: Verify inner elements text is cleared after click on cross icon
+    Given Type "Search" text into search input
+    When I click on cross icon
+    Then search input is empty
+
+  @positive
+  Scenario: Click event for Search icon
+    Given Type "Search" text into search input
+      And clear all actions in Actions Tab
+    When I click onto search icon
+    Then click action was called in Actions Tab
--- a/src/__experimental__/components/search/search.component.js
+++ b/src/__experimental__/components/search/search.component.js
@@ -1,0 +1,151 @@
+import React, { useState, useRef, useLayoutEffect } from 'react';
+import PropTypes from 'prop-types';
+import invariant from 'invariant';
+import StyledSearch, { StyledSearchButton, StyledButtonIcon } from './search.style';
+import Icon from '../../../components/icon';
+import Textbox from '../textbox';
+import Button from '../../../components/button';
+
+const Search = ({
+  defaultValue, onChange, onClick, value, id, name, threshold, searchButton, placeholder, ...rest
+}) => {
+  const isControlled = value !== undefined;
+  const initialValue = isControlled ? value : defaultValue;
+  invariant(
+    typeof initialValue === 'string',
+    'This component has no initial value'
+  );
+  let inputRef = useRef(null);
+  const [searchValue, setSearchValue] = useState(initialValue);
+  const [isActive, setIsActive] = useState(inputRef.current === document.activeElement);
+  const [searchIsActive, setSearchIsActive] = useState(initialValue.length >= threshold);
+  const [iconType, setIconType] = useState(null);
+
+  const handleChange = (e) => {
+    if (onChange) {
+      onChange(e);
+    }
+    setSearchValue(e.target.value);
+  };
+
+  const handleOnFocus = () => {
+    setIsActive(true);
+  };
+
+  let buttonProps = {};
+  if (searchButton && onClick) {
+    buttonProps = {
+      onClick: (ev) => {
+        onClick({
+          target: {
+            name: ev.target.name,
+            id: ev.target.id,
+            value: searchValue
+          }
+        });
+      }
+    };
+  }
+
+  const handleIconClick = () => {
+    setSearchValue('');
+    if (onChange) {
+      onChange({
+        target: {
+          ...(name && { name }),
+          ...(id && { id }),
+          value: ''
+        }
+      });
+    }
+  };
+
+  const handleBlur = () => {
+    setIsActive(false);
+  };
+
+  useLayoutEffect(() => {
+    setSearchIsActive(searchValue.length >= threshold);
+    if (searchValue.length > 0) {
+      setIconType('cross');
+    } else if ((searchButton && isActive)) {
+      setIconType('');
+    } else if ((!isActive && !searchIsActive) || threshold === 0) {
+      setIconType('search');
+    }
+  }, [isActive, searchButton, searchIsActive, searchValue, threshold]);
+
+  return (
+    <StyledSearch
+      onFocus={ handleOnFocus }
+      onClick={ handleOnFocus }
+      onBlur={ handleBlur }
+      onChange={ handleChange }
+      isActive={ isActive }
+      searchIsActive={ searchIsActive }
+      id={ id }
+      data-component='search'
+      name={ name }
+      searchHasValue={ searchValue && searchValue.length }
+      hasSearchButton={ Boolean(searchButton) }
+    >
+      <Textbox
+        { ...rest }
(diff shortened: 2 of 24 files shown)


src/__experimental__/components/search/search.stories.js
```diff
@@ -0,0 +1,62 @@
...
+        onBlur={ handleBlur }
+        value={ store.get('value') }
+        name='search_name'
+        id='search_id'
+      />
+    </State>
+  );
+};
+
+storiesOf('Experimental/Search', module)
```
edleeks87: I think we should avoid using `storiesOf` as its deprecated in favour of `Component Story Format (CSF)`

cypress/features/regression/experimental/search.feature
```diff
@@ -0,0 +1,71 @@
+Feature: Search component
+  I want to change Search component properties
+
+  Background: Open Search component page
+    Given I open "Experimental Search" component page
```
ianforster-sage:
```suggestion
    Given I open "Search" component page
```

src/__experimental__/components/search/search.stories.js
DipperTheDan: Addressed in commit 76f58cf4d7d015bfb7090389cd84b9f24edcdbb0

cypress/features/regression/experimental/search.feature
DipperTheDan: This has been addressed in commit 4ff060a5bf6bac1f57c0b3fab9e4daafe1322a0d

cypress/features/regression/experimental/search.feature
```diff
@@ -0,0 +1,71 @@
...
+    Examples:
+      | placeholder             |
+      | Sample text             |
+      | áéíóú¿¡üñ               |
+      | !@#$%^*()_+-=~[];:.,?{} |
+      | ÄÖÜßäöüß                |
+      | <>                      |
+
+  @positive
+  Scenario Outline: Set inner text to <text> and threshold to <threshold>
```
KatarzynaQA: You check that `Then search icon has golden border`, change the name of `Scenario`

cypress/features/regression/experimental/search.feature
```diff
@@ -0,0 +1,71 @@
...
+    Then Search component has input and "search" as icon
+
+  @positive
+  Scenario: Verify inner elements in Search component after change icon
+    When Type "Search" text into search input
+    Then Search component has input and "cross" as icon
+      And Search component input has golden border
+
+  @positive
+  Scenario: Verify golden outline for cross icon
```
KatarzynaQA: name of `Scenario` should indicate what we check in TC

cypress/features/regression/experimental/search.feature
```diff
@@ -0,0 +1,71 @@
+Feature: Search component
```
KatarzynaQA: 1. Please add TC which check the color of search icon when input is focused
2. Please add TC which check whether the input is empty after clicking on the cross icon

cypress/features/regression/experimental/search.feature
DlgSHi: this change brakes the build feature

DlgSHi: This change won't work

cypress/features/regression/experimental/search.feature
DlgSHi: fixed

cypress/features/regression/experimental/search.feature
DlgSHi: fixed

cypress/features/regression/experimental/search.feature
DlgSHi: added

cypress/features/regression/experimental/search.feature
```diff
@@ -0,0 +1,71 @@
...
+  Scenario: Verify inner elements in Search component
+    # commented because of BDD default scenario Given - When - Then
+    # When I open "Experimental Search" component page
+    Then Search component has input and "search" as icon
+
+  @positive
+  Scenario: Verify inner elements in Search component after change icon
+    When Type "Search" text into search input
+    Then Search component has input and "cross" as icon
+      And Search component input has golden border
```
damianOfKyan: Instead of stating golden border, it might be better to use wording that does not describe the color in case in future the colour changes. Instead we can simply say  "And Search component input has correct border"

src/__experimental__/components/search/search.component.js
```diff
@@ -0,0 +1,145 @@
...
+      onChange(e);
+    }
+    setSearchValue(e.target.value);
+  };
+
+  const handleOnFocus = () => {
+    setIsActive(true);
+  };
+
+  let buttonProps = {};
```
damianOfKyan:
```suggestion
  let buttonProps;
```

src/__experimental__/components/search/search.spec.js
```diff
@@ -0,0 +1,214 @@
...
+  describe('styles', () => {
+    it('matches the expected styles', () => {
+      assertStyleMatch({
+        borderBottom: '1px solid #CCD6DA',
+        display: 'inline-flex',
+        fontSize: '14px',
+        fontWeight: '700'
+      }, renderWrapper({ value: '' }, mount));
+    });
+
```
damianOfKyan: This can be in its own describe caluse.

`describe('matches the expected styles', () => {`

and then it blocks 

`it('when the input is focused', () => {`

src/__experimental__/components/search/search.style.js
```diff
@@ -0,0 +1,103 @@
...
+import StyledInputPresentation from '../input/input-presentation.style';
+import StyledIcon from '../../../components/icon/icon.style';
+import StyledButton from '../../../components/button/button.style';
+import { baseTheme } from '../../../style/themes';
+import StyledFormField from '../form-field/form-field.style';
+
+const StyledSearch = styled.div`
+  ${({ theme }) => `border-bottom: 1px solid ${theme.search.passive};`}
+  ${({ isActive, searchHasValue }) => css`
+    ${!isActive && !searchHasValue && css`
```
damianOfKyan: `!isActive` is nested in inside `isActive`.

src/__experimental__/components/search/search.component.js
DipperTheDan: Couldn't this potentially lead to spreading `{ ...undefined }` in `<Button>`? I also felt that the original way was more readable.

src/__experimental__/components/search/search.stories.js
edleeks87: Thanks mate

src/__experimental__/components/search/search.spec.js
DipperTheDan: I take onboard your suggestion however, I felt that having the styles all in one `describe` block makes it easier to read and see the tests in the terminal.

cypress/features/regression/experimental/search.feature
DlgSHi: fixed

cypress/features/regression/experimental/search.feature
DlgSHi: we need to be precise :) so golden border is a good step name

cypress/features/regression/test/search.feature
```diff
@@ -0,0 +1,83 @@
...
+  Scenario: Verify golden outline for search icon
+    Given Type "Sea" text into search input
+    When I click onto search icon
+    Then search icon has golden border
+
+  @positive
+  Scenario: Check the change function call for Search component
+    Given clear all actions in Actions Tab
+    When Type "Search" text into search input
+    Then change action was called in Actions Tab
```
KatarzynaQA: we have similar steps in cypress tests, so we should keep convention related to the name
```suggestion
    Then change event was called in Actions Tab
```

cypress/features/regression/test/search.feature
```diff
@@ -0,0 +1,83 @@
...
+
+  @positive
+  Scenario: Check the blur function call for Search component
+    Given clear all actions in Actions Tab
+      And I click inside input
+    When I click outside of the component in DLS directory
+    Then blur action was called in Actions Tab
```
KatarzynaQA:
```suggestion
    Then blur event was called in Actions Tab
```

cypress/features/regression/test/search.feature
```diff
@@ -0,0 +1,83 @@
...
+    Examples:
+      | placeholder             |
+      | Sample text             |
+      | áéíóú¿¡üñ               |
+      | !@#$%^*()_+-=~[];:.,?{} |
+      | ÄÖÜßäöüß                |
+      | <>                      |
+
+  @positive
+  Scenario Outline: Set inner text to <text> and threshold to <threshold> and verify golden border for search icon
```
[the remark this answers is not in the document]
DlgSHi: udpated

cypress/features/regression/test/search.feature
```diff
@@ -0,0 +1,83 @@
...
+    Then search icon has golden border
+
+  @positive
+  Scenario: Check the change function call for Search component
+    Given clear all actions in Actions Tab
+    When Type "Search" text into search input
+    Then change action was called in Actions Tab
+
+  @positive
+  Scenario: Check the blur function call for Search component
```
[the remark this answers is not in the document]
DlgSHi: udpated

cypress/features/regression/test/search.feature
```diff
@@ -0,0 +1,83 @@
...
+      And Search component input has golden border
+
+  @positive
+  Scenario: Verify golden outline for search icon
+    Given Type "Sea" text into search input
+    When I click onto search icon
+    Then search icon has golden border
+
+  @positive
+  Scenario: Check the change function call for Search component
```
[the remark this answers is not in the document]
DlgSHi: udpated

cypress/features/regression/test/search.feature
DlgSHi: it's a common step, won't be fixing it

cypress/features/regression/test/search.feature
DlgSHi: it's a common step, won't be fixing it

harpalsingh:
@DipperTheDan Might have been missed in a spec, but all buttons show behave the same, so only thing missing is the button to have the darker value on hover.

src/__experimental__/components/search/search.style.js
```diff
@@ -81,16 +81,23 @@ export const StyledSearchButton = styled.div`
...
+    `}
     width: 43px;
     height: 43px;
     margin: 0px 0px;
     padding-bottom: 3px;
     transition: border 0.2s ease, background 0.2s ease;
     :focus {
      z-index: 10;
-   }
+    }
```
damianOfKyan: looks like a wrong indentation

edleeks87: agree with @damianOfKyan  and you can remove the empty lines as well

src/__experimental__/components/search/search.component.js
```diff
@@ -0,0 +1,145 @@
+import React, { useState, useRef, useLayoutEffect } from 'react';
```
chrisbarber86: If this is a new component, why is it in `__experimental__`?

edleeks87: It's built using the `__experimental__ Textbox` so I advised @DipperTheDan  to have it in the same directory

chrisbarber86: Ok - we probably need to start thinking about when we drop the deprecated components so we can move these into the main components folder

src/__experimental__/components/search/search.component.js
```diff
@@ -134,8 +134,14 @@ Search.propTypes = {
...
   name: PropTypes.string,
-  /** Prop for active search threshold */
-  threshold: PropTypes.number,
+  /** Prop for active search threshold. This must be a positive number. */
+  threshold (props, propName) {
+    let error;
+    if (props[propName] && typeof props[propName] === 'number' && props[propName] < 0) {
+      error = new Error('Threshold must be a positive number.');
+    }
+    return error;
```
damianOfKyan: Wouldn't that work, just bit cleaner:

```suggestion
   if (props[propName] && typeof props[propName] === 'number' && props[propName] < 0) {
      return new Error('Threshold must be a positive number.');
    }
```

DipperTheDan: I'm not sure it would work. With the syntax of the suggested change, `return` is inside of the `if` statement. The method of threshold will expect the `return` outside of the `if` statement. For this reason, I've assigned `new Error('Threshold must be a positive number.')` inside the `if` statement and `return error;` outside the `if` statement.

PR #2607, opened 2020-01-20
ci: add trigger cron, add all regression tests

jamime:
### Proposed behaviour

Regression suite should be running as Travis build  triggered by cron.

### Checklist

<del> - [ ] Release notes (Conventional Commits)  </del> 
<del>  - [ ] Unit tests </del> 
- [x] Cypress automation tests on CI

<del>  - [ ] Storybook added or updated </del> 
<del>  - [ ] Theme support </del> 
<del>  - [ ] Typescript `d.ts` file added or updated </del> 

### Additional context

add regression suite for Travis build triggered by cron.


jamime:
@DlgSHi to progress this I think we should build on a nightly cron. That way we do not need teamcity. Do you agree?

package.json
```diff
@@ -19,12 +19,13 @@
...
-    "cypress-group-regression-common": "cypress run --spec './cypress/features/regression/*.feature' --reporter mocha-allure-reporter --browser chrome",
-    "cypress-group-regression-experimental": "cypress run --spec './cypress/features/regression/experimental/*.feature' --reporter mocha-allure-reporter --browser chrome",
-    "cypress-group-regression-deprecated": "cypress run --spec './cypress/features/regression/deprecated/*.feature' --reporter mocha-allure-reporter --browser chrome",
-    "cypress-group-regression-validations": "cypress run --spec './cypress/features/regression/validations/*.feature' --reporter mocha-allure-reporter --browser chrome",
-    "cypress-group-regression-themes": "cypress run --spec './cypress/features/themes/themes.feature' --reporter mocha-allure-reporter --browser chrome",
+    "test-cypress-regression": "npm run test-cypress-accessibility; npm run cypress-group-regression-validations; npm run cypress-group-regression-themes; npm run cypress-group-regression-common; npm run cypress-group-regression-experimental; npm run cypress-group-regression-deprecated",
```
jamime: `test-cypress-regression` and `cypress-group-regression-common` both run the same tests. 

I prefer the `cypress-group-regression-common` but could we rename it to `cypress-group-regression-all`and remove `test-cypress-regression`. I prefer it because of the glob matching all files.

package.json
```diff
@@ -19,12 +19,13 @@
...
+    "cypress-group-regression-common": "cypress run --spec './cypress/features/regression/*.feature' --reporter spec --browser chrome",
+    "cypress-group-regression-experimental": "cypress run  --spec './cypress/features/regression/experimental/*.feature' --reporter spec --browser chrome",
```
jamime: `cypress-group-regression-*` are not used by anything, we should remove these. Even running locally you can use the vscode launcher to run a specific cypress file.

package.json
DlgSHi: I've divided the scripts that way to have possibility run npm script from Terminal

package.json
DlgSHi: `test-cypress-regression` is using all those `cypress-group-regression-*` scripts

PR #2616, opened 2020-01-24
feat(textbox): add icon onclick prop to textbox

DipperTheDan:
### Proposed behaviour

A new iconOnClick prop has been added to allow for onClick events on the Icon to be handled separately from the onClick on the input.

### Current behaviour

Currently Textbox only supports a single onClick handler, which is consumed by the Input and Icon. 

### Checklist

- [x] Unit tests
<del>- [ ] Storybook added or updated </del>
<del>- [ ] Theme support </del>
<del>- [ ] Typescript `d.ts` file added or updated </del>

### Additional context

This work is needed to unblock development of new search component

### Testing instructions

Test that current behaviour of select component on master branch still remains.

### DIFF
--- a/cypress/support/step-definitions/textbox-steps.js
+++ b/cypress/support/step-definitions/textbox-steps.js
@@ -1,4 +1,6 @@
-import { textbox, textboxByPosition, textboxDataComponent } from '../../locators/textbox';
+import {
+  textbox, textboxByPosition, textboxDataComponent, textboxIcon, textboxInput,
+} from '../../locators/textbox';
 import {
   label, labelByPosition, commonDataElementInputPreview, commonInputPreview,
   fieldHelpPreviewByPosition, tooltipPreviewByPosition,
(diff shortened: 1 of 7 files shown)


src/__experimental__/components/textbox/textbox.spec.js
```diff
@@ -13,4 +14,24 @@ describe('Textbox', () => {
     ).dive().dive().dive();
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('supports a seprate onClick handler passing for the icon', () => {
```
edleeks87:
```suggestion
  it('supports a separate onClick handler passing for the icon', () => {
```

sianford:
Tested the new iconOnClick event in Chrome, Safari and Edge. Also checked that Textbook and Select components had not regressed. Testing passed ✅  However, it requires some Cypress tests adding for Textbox to cover events and icon testing.

cypress/support/step-definitions/textbox-steps.js
```diff
@@ -229,3 +229,20 @@ Then('Multiple label Align on preview is {string}', (direction) => {
   labelByPosition(FIRST_ELEMENT).should($element => expect($element).to.have.css(TEXT_ALIGN, `${direction}`));
   labelByPosition(SECOND_ELEMENT).should($element => expect($element).to.have.css(TEXT_ALIGN, `${direction}`));
 });
+
+Then('I click on icon inside of Textbox', () => {
+  icon().click();
+});
+
+Then('icon on preview is {string} and is visible', (iconName) => {
+  icon().should('have.attr', 'data-element', iconName)
```
DlgSHi: we should use new locator to find the icon inside the Textbox input

cypress/support/step-definitions/textbox-steps.js
```diff
@@ -229,3 +229,20 @@ Then('Multiple label Align on preview is {string}', (direction) => {
...
+    .and('be.visible');
+});
+
+When('I click on Textbox', () => {
+  textbox().click();
+});
+
+Then('Textbox input has golden border on focus', () => {
+  textbox().should('have.css', 'outline', 'rgb(255, 181, 0) solid 3px');
```
DlgSHi: we need to verify cy.focused() element, not just the textbox

cypress/support/step-definitions/textbox-steps.js
KatarzynaQA: fixed

cypress/support/step-definitions/textbox-steps.js
KatarzynaQA: I can't use cy.focused() because I click on the input, and the parent element has a golden border

PR #2621, opened 2020-01-27
fix(form): sticky footer form retains focus on page when tabbing FE-2498

damianOfKyan:
### Proposed behaviour
User should be able to Tab/Shift+Tab through focusable elements on the page including Form Component and child elements while Form Component has enabled sticky footer.
User should be able to trigger form validation through clicking Save/Submit button and regain focus using Tab/Shift+Tab keys.

### Current behaviour
For Form Component with sticky footer enabled, the focus is lost when tabbing beyond last form element (for example submit/save button). Also focus is lost when Submit/Save button is clicked but form is not submitted due to validation. Focus is never regain using keyboard alone.

### Checklist

- [ ] Release notes (Conventional Commits) 
<del>- [ ] Unit tests
<del>- [ ] Theme support

### Testing instructions
Storybook -> Experimental -> Form -> default (story)
Select sticky footer (knob)
Select form element to gain focus, then use Tab Key to change focus until Save button is focused, use Enter key, this should cause Save button to loose focus, then either, pressing Tab or Shift+Tab should focus element on the page.


src/__experimental__/components/form/form.component.js
```diff
@@ -207,9 +207,16 @@ class FormWithoutValidations extends React.Component {
       this.triggerSubmit(ev, valid);
     } else {
       this.setState({ submitted: false });
+      this.regainFocusOnSubmit();
     }
   }
 
+  regainFocusOnSubmit() {
```
jamime: Why is this needed, it would be better if we prevent `focus` from changing in the first place. Screen readers will read the element that is briefly focused otherwise.

harpalsingh:
@damianOfKyan Unless I'm missing something, just tried this and after pressing enter there appears to be no re-focus on elements?

jamime:
I've created FE-2549 to address the fact that focus is moving when it shouldn't be. As a team we decided that this ticket should continue as overall the experience is still better.

jamime:
> Unless I'm missing something, just tried this and after pressing enter there appears to be no re-focus on elements?

Did you press enter on the submit button? I tried and it worked for me.

sianford:
Tested in Chrome, Safari, Firefox, Edge and IE11 browsers. Tested all stories, with and without a sticky footer. Focus is not lost when tabbing through and out of the form nor when an attempt to submit the form is made when it has failed validation. Testing passed ✅

PR #2632, opened 2020-02-04
fix(help): allow to set different icon on help component

bepud:
### Proposed behaviour
We would like to have the ability to choose another icon for the help icon on a textbox.
So I added a "type" props in the help component.

image

### Current behaviour
There is a way to pass the icon we want through the labelHelpIcon props.
From the textbox through the form-field componant to the label componant :https://github.com/Sage/carbon/blob/0241195de9006711b110a500831372bae278e131/src/__experimental__/components/form-field/form-field.component.js#L56
Then from the label to the help component :https://github.com/Sage/carbon/blob/0241195de9006711b110a500831372bae278e131/src/__experimental__/components/label/label.component.js#L66
But in the help component, the icon is hard coded :https://github.com/Sage/carbon/blob/0241195de9006711b110a500831372bae278e131/src/components/help/help.component.js#L70
I just completed the chain.

### Checklist
- [x] Unit tests
- [ ] Theme support

### Additional context
That modification will be usefull for Client Management : 

image

### Testing instructions
On an help component, we can add the props type with the value 'info'. That should display the icon for information.
It can be used on a TextBox using the "labelHelpIcon" props .
For both components, the default behavior must remain the help icon if the props are not defined to avoid side effect.

### DIFF
--- a/src/components/help/help.component.js
+++ b/src/components/help/help.component.js
@@ -18,7 +18,8 @@ const Help = (props) => {
     as,
     tooltipPosition,
     tooltipAlign,
-    isFocused
+    isFocused,
+    type
   } = props;
 
   useEffect(() => {
@@ -67,7 +68,7 @@ const Help = (props) => {
       aria-label={ children }
     >
       <Icon
-        type='help'
+        type={ type }
         tooltipMessage={ children }
         tooltipPosition={ tooltipPosition }
         tooltipAlign={ tooltipAlign }
@@ -95,13 +96,16 @@ Help.propTypes = {
   /** A path for the anchor */
   href: PropTypes.string,
   /** A boolean recived from IconWrapper */
-  isFocused: PropTypes.bool
+  isFocused: PropTypes.bool,
+  /** Icon to display, can be received from label component */
+  type: PropTypes.oneOf(OptionsHelper.icons)
 };
 
 Help.defaultProps = {
   tooltipPosition: 'top',
   tooltipAlign: 'center',
-  tabIndex: 0
+  tabIndex: 0,
+  type: 'help'
 };
 
 export default Help;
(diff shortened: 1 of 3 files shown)


jamime:
Thanks for the PR, we will review this in our triage session on Tuesday.

src/components/help/help.stories.js
```diff
@@ -26,12 +26,14 @@ function makeStory(name, themeSelector) {
       Help.defaultProps.tooltipAlign
     ) : undefined;
     const href = text('href', '');
+    const type = text('type', 'help');
```
jamime:
```suggestion
    const type = select('type', OptionsHelper.icons, 'help');
```

src/components/help/help.component.js
```diff
@@ -95,13 +96,16 @@ Help.propTypes = {
   /** A path for the anchor */
   href: PropTypes.string,
   /** A boolean recived from IconWrapper */
-  isFocused: PropTypes.bool
+  isFocused: PropTypes.bool,
+  /** Icon to display, can be received from label component */
+  type: PropTypes.string
```
jamime:
```suggestion
  type: PropTypes.oneOf(OptionsHelper.icons)
```

bepud:
You're right @sianford  and @jamime it would be better, thanks.

sianford:
Tested the default add classic stories in Safari, Chrome, Firefox, Edge and IE11. The icon can now be changed to any available icon in the dropdown selector. Providing no further changes are required by UX, testing has passed ✅

jamime:
I [cleanly](https://github.com/Sage/carbon/compare/e9e5c74068f1c4b98cc35c92e53b3a84f47d7a35..487be1bed194f0810442ed42328474d487c2143b) squashed the commits for the release notes.

I'll merge once CI passes.

PR #2652, opened 2020-02-19
fix(pager): add singular/plural translation FE-2510

jamime:
### Proposed behaviour

1. Add tests for existing I18n translations in `Pager`.
1. Add singular/plural translations fo records, the default translation will now display `0 items`, `1 item`, `10 items`.
1. Use the page size value for the translation of page size instead of total records.

image
image
image

### Current behaviour

1. There are no tests for translations.
1. If they are using the default translations and `totalRecords=1` it will say `1 items`.
1.  If
    1. `totalRecords=1 pageSize=10` it will say `Show 10 item`.
    1. `totalRecords=100 pageSize=1` it will say `Show 1 items`.

image
image

### Checklist

- [x] Unit tests
<del>- [ ] Theme support

### Additional context

The singular/plural translations currently work if the user defines their own translation. However, the page size and total records both use the `totalRecords` prop to work out how may to display. This means you can end up with `Show 10 item` and `Showing 1 item` instead of `Showing 10 items` and `Showing 1 item`.

### Testing instructions
* Go to http://localhost:9001/?path=/story/pager--default
* Check `showPageSizeSelection`
* Use `pageSize` and `totalRecords` knobs.
* Compare to https://carbon.sage.com/storybook/?path=/story/pager--default

image

### DIFF
--- a/src/components/pager/pager.component.js
+++ b/src/components/pager/pager.component.js
@@ -22,10 +22,10 @@ const Pager = (props) => {
   }, [props.currentPage]);
 
   /** Term used to describe table data */
-  const descriptor = I18n.t(
+  const records = count => I18n.t(
     'pager.records',
     {
-      count: Number(props.totalRecords),
+      count: Number(count),
       defaultValue: isClassic(currentTheme) ? {
         one: 'record',
         zero: 'records',
@@ -50,9 +50,10 @@ const Pager = (props) => {
   }
 
   function pageSizeOptions() {
+    const show = I18n.t('pager.show', { defaultValue: 'Show' });
     const elem = (
       <PagerSizeOptionsInnerStyles>
-        <span>{ I18n.t('pager.show', { defaultValue: 'Show' }) }</span>{ sizeSelector() }<span>{ descriptor }</span>
+        <span>{ show }</span>{ sizeSelector() }<span>{ records(props.pageSize) }</span>
       </PagerSizeOptionsInnerStyles>
     );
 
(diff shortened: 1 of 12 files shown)


src/components/pager/pager.stories.js
```diff
@@ -80,7 +80,12 @@ TableComponent.propTypes = {
 function makeStory(name, themeSelector) {
   const component = () => {
     const totalRecords = number('totalRecords', 100);
-    const pageSize = select('pageSize', OptionsHelper.pageSizes, Pager.defaultProps.pageSize);
+    const pageSize = select('pageSize', {
+      one: 1,
```
jamime: _Note_: It's not possible to use `1` as a value. As far as I can tell there's a bug in `@storybook/addon-knobs` which makes it the default, we want `10` to be the default.

src/components/pager/pager.component.js
```diff
@@ -22,11 +22,19 @@ const Pager = (props) => {
...
+  const records = count => I18n.t(
     'pager.records',
     {
-      count: Number(props.totalRecords),
-      defaultValue: isClassic(currentTheme) ? ' records' : ' items'
+      count: Number(count),
+      defaultValue: isClassic(currentTheme) ? {
+        one: 'record',
+        zero: 'records',
+        other: 'records'
```
edleeks87: @jamime is there a reason why you have to define both `zero` and `other`?

jamime: Removed.

PR #2656, opened 2020-02-20
fix(card): add data role prop for automation

edleeks87:
### Proposed behaviour

Allow consumer to pass data-role prop

### Current behaviour

data-role can not be passed to Card
### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>

### DIFF
--- a/src/components/card/card.component.js
+++ b/src/components/card/card.component.js
@@ -12,7 +12,8 @@ const Card = ({
   cardWidth,
   interactive,
   draggable,
-  spacing
+  spacing,
+  dataRole
 }) => {
   const handleClick = (ev) => {
     if (!draggable && action) {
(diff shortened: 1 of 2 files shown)


src/components/card/card.component.js
```diff
@@ -36,6 +37,7 @@ const Card = ({
       type='button'
       onClick={ onClickHandler }
       tabIndex={ 0 }
+      data-role={ dataRole || '' }
```
jamime: It's better to conditionally add this prop otherwise we'll end up with `<div data-role></div>`

src/components/card/card.spec.js
```diff
@@ -158,6 +158,14 @@ describe('Card', () => {
       }, wrapper.toJSON());
     });
   });
+
+  describe('when the "draggable" prop is set to true', () => {
```
jamime: Incorrect `describe` block(draggable is not set to true in the test), the `it` block is sufficient.
Also needs a negative test for when data role is not provided.

PR #2660, opened 2020-02-20
feat: add new anchor navigation component - FE-2335

mkrds:
### Proposed behaviour
This PR implements new AnchorNavigation component

### Checklist

- [x] Unit tests
- [ ] Theme support

### Testing instructions
Test/AnchorNavigation storybook stories

Also contains fixes for FE-2581 and FE-2582

### DIFF
--- a/cypress/features/regression/test/anchorNavigation.feature
+++ b/cypress/features/regression/test/anchorNavigation.feature
@@ -1,0 +1,29 @@
+Feature: Anchor Navigation component
+  I want to change Anchor Navigation component properties
+
+  Background: Open Anchor Navigation component page
+    Given I open in full screen Test "AnchorNavigation" component page in noIframe
+      And I open component preview in noIFrame
+
+  @positive
+  Scenario Outline: Press on <tab> tab and scroll to the the <index> anchor navigation
+    When I click onto "<tab>" tab
+      And I wait 250
+    Then "<index>" anchor navigation section is visible
+    Examples:
+      | tab                                  | index          |
+      | First                                | First section  |
+      | Second                               | Second section |
+      | Third                                | Third section  |
+      | Navigation item with very long label | Fourth section |
+      | Fifth                                | Fifth section  |
+
+  @positive
+  Scenario Outline: Scroll <index> and verify that proper <index> anchor navigation row is visible
+    Given I click onto "<tab>" tab
+    When I scroll window to the "<index>" position
+    Then "<index>" anchor navigation section is visible
+    Examples:
+      | tab    | index         |
+      | Fifth  | First section |
+      | First  | Fifth section |
--- a/cypress/locators/anchor-navigation/index.js
+++ b/cypress/locators/anchor-navigation/index.js
@@ -1,0 +1,13 @@
+import {
+  ANCHOR_NAVIGATION,
+  ANCHOR_NAVIGATION_STICKY_NAVIGATION,
+} from './locators';
+
+// component preview locators
+export const anchorNavigation = () => cy.get(ANCHOR_NAVIGATION);
+export const anchorNavigationStickyNavigation = text => anchorNavigation()
+  .find(ANCHOR_NAVIGATION_STICKY_NAVIGATION)
+  .contains(text);
+export const anchorNavigationStickyMainPage = text => anchorNavigation()
+  .find('div > h2')
+  .contains(text);
--- a/src/components/anchor-navigation/anchor-navigation.component.js
+++ b/src/components/anchor-navigation/anchor-navigation.component.js
@@ -1,0 +1,160 @@
+import React, {
+  useState, useRef, useEffect, useCallback
+} from 'react';
+import PropTypes from 'prop-types';
+import throttle from 'lodash/throttle';
+
+import Event from '../../utils/helpers/events';
+import { StyledAnchorNavigation, StyledNavigation, StyledContent } from './anchor-navigation.style';
+import AnchorNavigationItem from './anchor-navigation-item.component';
+
+const SECTION_VISIBILITY_OFFSET = 200;
+const SCROLL_THROTTLE = 100;
+
+const AnchorNavigation = ({ children, stickyNavigation, styleOverride }) => {
+  const [selectedIndex, setSelectedIndex] = useState(0);
+
+  const sectionRefs = useRef(React.Children.map(stickyNavigation.props.children, child => child.props.target));
+
+  const anchorRefs = useRef(Array.from({
+    length: React.Children.count(stickyNavigation.props.children)
+  }, () => React.createRef()));
+
+  const contentRef = useRef();
+
+  const navigationRef = useRef();
+
+  const isUserScroll = useRef(true);
+
+  const isUserScrollTimer = useRef();
+
+  const setSelectedAnchorBasedOnScroll = useCallback(() => {
+    const sectionsTopOffsets = sectionRefs.current.map(({ current }) => current.getBoundingClientRect().top);
+    const { top: navTopOffset } = navigationRef.current.getBoundingClientRect();
+
+    const indexOfSmallestNegativeTopOffset = sectionsTopOffsets.reduce((currentTopIndex, sectionTopOffset, index) => {
+      if (sectionTopOffset - SECTION_VISIBILITY_OFFSET > navTopOffset) return currentTopIndex;
+      return sectionTopOffset > sectionsTopOffsets[currentTopIndex] ? index : currentTopIndex;
+    }, 0);
+
+    setSelectedIndex(indexOfSmallestNegativeTopOffset);
+  }, []);
+
+  const scrollHandler = useCallback(throttle(() => {
+    if (isUserScroll.current) {
+      setSelectedAnchorBasedOnScroll();
+    } else {
+      window.clearTimeout(isUserScrollTimer.current);
+      isUserScrollTimer.current = setTimeout(() => { isUserScroll.current = true; }, SCROLL_THROTTLE + 50);
+    }
+  }, SCROLL_THROTTLE), []);
+
+  useEffect(() => {
+    window.addEventListener('scroll', scrollHandler, true);
+    return () => window.removeEventListener('scroll', scrollHandler, true);
+  }, [scrollHandler]);
+
+  const focusFirstFocusableChild = (section) => {
+    // eslint-disable-next-line max-len
+    const defaultFocusableSelectors = 'button, [href], input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])';
+    const firstFocusableElement = section.querySelector(defaultFocusableSelectors);
+    if (firstFocusableElement) {
+      firstFocusableElement.focus({ preventScroll: true });
+    }
+  };
+
+  const scrollToSection = (section) => {
+    section.scrollIntoView({ block: 'start', inline: 'nearest', behavior: 'smooth' });
+    isUserScroll.current = false;
+  };
+
+  const handleClick = (event, index) => {
+    event.preventDefault();
+    const sectionToScroll = sectionRefs.current[index].current;
+
+    scrollToSection(sectionToScroll);
+    focusFirstFocusableChild(sectionToScroll);
+    setSelectedIndex(index);
+  };
+
+  const focusNavItem = (event, index) => {
+    event.preventDefault();
+    let newIndex = index;
+    if (index === -1) {
+      newIndex = anchorRefs.current.length - 1;
+    } else if (index === anchorRefs.current.length) {
+      newIndex = 0;
+    }
+    anchorRefs.current[newIndex].current.focus();
+  };
+
+  const handleKeyDown = (event, index) => {
+    if (Event.isUpKey(event)) {
+      focusNavItem(event, index - 1);
+    } else if (Event.isDownKey(event)) {
+      focusNavItem(event, index + 1);
+    } else if (Event.isEnterKey(event) || Event.isSpaceKey(event)) {
+      handleClick(event, index);
+    }
+  };
+
+  return (
+    <StyledAnchorNavigation
+      ref={ contentRef }
+      data-component='anchor-navigation'
+      styleOverride={ styleOverride.root }
+    >
+      <StyledNavigation
+        ref={ navigationRef }
+        data-element='anchor-sticky-navigation'
+        styleOverride={ styleOverride.navigation }
+      >
+        {React.Children.map(
+          stickyNavigation.props.children,
+          (child, index) => React.cloneElement(child, {
+            isSelected: index === selectedIndex,
+            tabIndex: index === selectedIndex ? 0 : -1,
+            onClick: ev => handleClick(ev, index),
+            onKeyDown: ev => handleKeyDown(ev, index),
+            ref: anchorRefs.current[index]
+          })
+        )}
+      </StyledNavigation>
+      <StyledContent styleOverride={ styleOverride.content }>
+        {children}
+      </StyledContent>
+    </StyledAnchorNavigation>
+  );
+};
+
+AnchorNavigation.propTypes = {
+  children: PropTypes.node,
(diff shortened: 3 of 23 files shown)


src/components/anchor-navigation/anchor-section-divider.d.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export interface AccordionSectionDividerProps {
```
edleeks87: `AccordionSectionDivider...`?

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,133 @@
...
+import Event from '../../utils/helpers/events';
+import { StyledAnchorNavigation, StyledNavigation, StyledContent } from './anchor-navigation.style';
+import AnchorNavigationItem from './anchor-navigation-item.component';
+
+const AnchorNavigation = ({ children, stickyNavigation, styleOverride }) => {
+  const [selectedIndex, setSelectedIndex] = useState(0);
+
+  const sectionRefs = useRef(React.Children.map(stickyNavigation.props.children, child => child.props.target));
+
+  const anchorRefs = useRef(Array.from({
```
edleeks87: This will need a polyfill or different implementation while we still support `IE` 😢  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,133 @@
...
+    length: React.Children.count(stickyNavigation.props.children)
+  }, () => React.createRef()));
+
+  const contentRef = useRef();
+
+  const navigationRef = useRef();
+
+  const setSelectedAnchorBasedOnScroll = useCallback(throttle(() => {
```
edleeks87: Ideally we'd not use `lodash` here as the aim is to hopefully remove it as a dependency in the future. You could replace it with your `throttle` own implementation like what's discussed here https://stackoverflow.com/a/55974034/10894881

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,133 @@
...
+
+  useEffect(() => {
+    window.addEventListener('scroll', setSelectedAnchorBasedOnScroll, true);
+    return () => window.removeEventListener('scroll', setSelectedAnchorBasedOnScroll, true);
+  }, [setSelectedAnchorBasedOnScroll]);
+
+  const scrollToItem = (event, index) => {
+    event.preventDefault();
+    const sectionToScroll = sectionRefs.current[index].current;
+    sectionToScroll.scrollIntoView({ block: 'start', inline: 'nearest', behavior: 'smooth' });
```
edleeks87: Passing `scrollIntoView` the `scrollIntoViewOptions` isn't supported in `IE` or `Safari`, is the degradation in behaviour acceptable or do we need some kind of alternative for those browsers here?
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: True, I forgot about IE 11 :)

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: I thought only `IE11` doesn't support it, didn't even check for Safari. I think that getting immediately scrolled to the wanted section is fine for `IE11` as it covers bare minimum of the functionality. But for Safari it might be not enough. I've looked for quick solutions and it seems that installing a polyfill would do the job.
https://github.com/iamdustan/smoothscroll - would work, but I'm not sure what is our stance on installing additional packages like this

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: @edleeks87 It seems that our library depends on the lodash pretty extensively. Are planning to rewrite methods like `merge` or `deepClone` in the future just to get rid of the `lodash`?

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: @harpalsingh Confirmed that it's ok that on Safari it will just immediately jump to the wanted section. So I'm leaving it as it is.

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: It seems that we dont need to support IE11 at all anymore

edleeks87: 🎉

src/components/anchor-navigation/anchor-navigation.component.js
edleeks87: Good point, I think we need to review how we import lodash as discussed on slack, but that is out of scope for this

src/components/anchor-navigation/anchor-navigation.component.js
edleeks87: Thanks for clearing that up @mkrds

harpalsingh:
@mkrds Could we change the background used in the demo to match the Dialog color? , it currently uses the same color the hover of the tabs do, so appears like nothing happens on hover.

Screenshot 2020-02-24 at 15 59 16

The background used here is not the dialog background color.

Could we also update that there is no hover effect change on the active anchor point, so it remains with the white background and theme highlight.

src/components/anchor-navigation/anchor-navigation.component.js
harpalsingh: Sorry, was meant to comment that on the ticket, yes that was ok about Safari.

src/components/anchor-navigation/anchor-navigation.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+<Meta
+  title="Design System/Anchor Navigation"
+  parameters={{ info: { disable: true }}}
+/>
+
+# Anchor Navigation
+Easily navigate through long scrollable content.
+
+## Overview
+
```
ianforster-sage: If there is no content in these sections you can remove the headings.

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,133 @@
...
+        {children}
+      </StyledContent>
+    </StyledAnchorNavigation>
+  );
+};
+
+AnchorNavigation.propTypes = {
+  children: PropTypes.node,
+  /** The AnchorNavigationItems components to be rendered in the sticky navigation */
+  stickyNavigation: (props, propName, componentName) => {
```
damianOfKyan: Unless you need that custom error message, maybe we can consider implementing shorter version of children.
```
children: PropTypes.shape({
    type: PropTypes.oneOf([AnchorNavigationItem, PropTypes.element])
  }).isRequired
```

src/components/anchor-navigation/anchor-navigation.d.ts
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+export interface AnchorNavigationProps {
+  children?: React.ReactNode;
+  /** The AnchorNavigationItems components to be rendered in the sticky navigation */
+  stickyNavigation?: React.ReactNode;
+  /** Allows to override existing component styles */
+  styleOverride?: {
+   root?: () => object | object;
```
damianOfKyan: So in component we allow `styleOverride` props to be either `func` or `object`. Can we do it here as well maybe?

src/components/anchor-navigation/anchor-navigation.style.js
```diff
@@ -0,0 +1,104 @@
...
+  ${({ styleOverride }) => styleOverride}
+`;
+
+const StyledAnchorDivider = styled.div.attrs({ 'data-element': 'anchor-navigation-divider' })`
+  background-color: ${({ theme }) => theme.anchorNavigation.divider};
+  height: 1px;
+
+  ${({ styleOverride }) => styleOverride}
+`;
+
```
damianOfKyan: white space.

src/components/anchor-navigation/anchor-section-divider.d.ts
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+export interface AnchorSectionDividerProps {
+  /** Allows to override existing component styles */
+  styleOverride?: () => object | object;
```
damianOfKyan: same as previous comment, func | object maybe?

src/components/anchor-navigation/anchor-navigation.style.js
mkrds: I don't see any unnecessary white spaces here

src/components/anchor-navigation/anchor-navigation.d.ts
mkrds: `root?: () => object | object` means that it is either a function returning an object, or just an object

src/components/anchor-navigation/anchor-section-divider.d.ts
mkrds: Same as above<
`styleOverride?: () => object | object` means that it is either a function returning an object, or just an object

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: Problem is that I'm checking the children of the <> (React.fragment) component here, so I'm unable to check it with PropTypes.shape here

src/components/anchor-navigation/anchor-navigation.stories.mdx
mkrds: I actually left it there because I thought it would be good to add it in the future - so I left it as a reminder of work needed to be done. I can remove though if you think that's better

ianforster-sage: I'd say either add the content now or remove the headings.

cypress/features/regression/test/anchorNavigation.feature
```diff
@@ -0,0 +1,29 @@
...
+      | First                                | First section  |
+      | Second                               | Second section |
+      | Third                                | Third section  |
+      | Navigation item with very long label | Fourth section |
+      | Fifth                                | Fifth section  |
+
+  @positive
+  Scenario Outline: Scroll <index> and verify that proper <index> anchor navigation row is visible
+    Given I click onto "<tab>" tab
+      And I scroll window to the "<index>" position
```
KatarzynaQA:
```suggestion
      When I scroll window to the "<index>" position
```

cypress/locators/anchor-navigation/index.js
```diff
@@ -0,0 +1,14 @@
...
+} from './locators';
+
+// component preview locators
+export const anchorNavigation = () => cy.get(ANCHOR_NAVIGATION);
+export const anchorNavigationStickyNavigation = text => anchorNavigation()
+  .find(ANCHOR_NAVIGATION_STICKY_NAVIGATION)
+  .contains(text);
+export const anchorNavigationStickyMainPage = text => anchorNavigation()
+  .find('div')
+  .find('h2')
```
KatarzynaQA: I think we can write just only one `find` as `.find('div h2')`

cypress/features/regression/test/anchorNavigation.feature
DlgSHi: updated

cypress/locators/anchor-navigation/index.js
DlgSHi: updated

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -21,35 +24,59 @@ const AnchorNavigation = ({ children, stickyNavigation, styleOverride }) => {
...
-      return topOffset > topEdgeOffsets[currentTopIndex] ? index : currentTopIndex;
+
+    const indexOfSmallestNegativeTopOffset = sectionsTopOffsets.reduce((currentTopIndex, sectionTopOffset, index) => {
+      if (sectionTopOffset - SECTION_VISIBILITY_OFFSET > navTopOffset) return currentTopIndex;
+      return sectionTopOffset > sectionsTopOffsets[currentTopIndex] ? index : currentTopIndex;
     }, 0);
 
     setSelectedIndex(indexOfSmallestNegativeTopOffset);
-  }, 200), []);
+  }, []);
```
damianOfKyan: you should decide to either drop useCallback or specify the dependencies: https://reactjs.org/docs/hooks-reference.html#usecallback

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -21,35 +24,59 @@ const AnchorNavigation = ({ children, stickyNavigation, styleOverride }) => {
...
+
+  const scrollHandler = useCallback(throttle(() => {
+    if (isUserScroll.current) {
+      setSelectedAnchorBasedOnScroll();
+    } else {
+      window.clearTimeout(isUserScrollTimer.current);
+      isUserScrollTimer.current = setTimeout(() => { isUserScroll.current = true; }, SCROLL_THROTTLE + 50);
+    }
+  }, SCROLL_THROTTLE), []);
```
damianOfKyan: same as above.

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,160 @@
...
+    const sectionsTopOffsets = sectionRefs.current.map(({ current }) => current.getBoundingClientRect().top);
+    const { top: navTopOffset } = navigationRef.current.getBoundingClientRect();
+
+    const indexOfSmallestNegativeTopOffset = sectionsTopOffsets.reduce((currentTopIndex, sectionTopOffset, index) => {
+      if (sectionTopOffset - SECTION_VISIBILITY_OFFSET > navTopOffset) return currentTopIndex;
+      return sectionTopOffset > sectionsTopOffsets[currentTopIndex] ? index : currentTopIndex;
+    }, 0);
+
+    setSelectedIndex(indexOfSmallestNegativeTopOffset);
+  }, []);
```
damianOfKyan: you should decide to either drop useCallback or specify the dependencies: https://reactjs.org/docs/hooks-reference.html#usecallback

src/components/anchor-navigation/anchor-navigation.component.js
```diff
@@ -0,0 +1,160 @@
...
+
+  const scrollHandler = useCallback(throttle(() => {
+    if (isUserScroll.current) {
+      setSelectedAnchorBasedOnScroll();
+    } else {
+      window.clearTimeout(isUserScrollTimer.current);
+      isUserScrollTimer.current = setTimeout(() => { isUserScroll.current = true; }, SCROLL_THROTTLE + 50);
+    }
+  }, SCROLL_THROTTLE), []);
```
damianOfKyan: same as above.

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: It is not necessary to specify dependencies here because `indexOfSmallestNegativeTopOffset` only takes `sectionRefs` and `navigationRef` from the outer scope and both of these are created by useRef meaning that they keep their references through whole lifecycle of the component.

Adding `[sectionRefs, navigationRef]` wont change anything in this case

src/components/anchor-navigation/anchor-navigation.component.js
mkrds: Same here

src/components/anchor-navigation/anchor-navigation.component.js
damianOfKyan: in fact I was wrong, due to how ref callbacks work we can have empty array of dependencies in this case: https://reactjs.org/docs/hooks-faq.html

PR #2669, opened 2020-02-25
fix(multi-action-button): change button text position (FE-2081)

grabkowski:
Dropdown icon was removed from the multi action button.
Rendering of the Icon moved directly into the button to keep consistency with other buttons.

### Proposed behaviour
Render the main button in Multi Action Button by passing the "iconType" prop to the Button Component to avoid style inconsistencies and needless style overrides.

### Current behaviour
Multi Action main button icon is rendered as a child of the button.

### Checklist
- [x] Unit tests
- [x] Theme support


jamime:
@grabkowski is this a breaking change for our users, what will they have to do differently?

grabkowski:
> @grabkowski is this a breaking change for our users, what will they have to do differently?

I think that I may have misunderstood what should be a `BREAKING CHANGE`. I'll amend that commit message.

sianford:
Tested in Safari, Chrome, Firefox, Edge and IE11 browsers. The buttons and text within in all default and classic stories are aligned in all browsers. Testing passed ✅

PR #2670, opened 2020-02-25
fix: experimental textarea rows not being applied

basalter:
### Proposed behaviour
Textarea rows property defines the height of the rendered textarea

### Current behaviour
Textarea rows property is overriden by the css height value

### Checklist

- [ ] Unit tests
- [ ] Theme support

### DIFF
--- a/src/__experimental__/components/textarea/textarea.style.js
+++ b/src/__experimental__/components/textarea/textarea.style.js
@@ -8,6 +8,7 @@ const StyledTextarea = styled.div`
     min-height: 40px;
     margin-top: 5px;
     margin-bottom: 5px;
+    height: unset;
   }
 
   ${({ labelInline }) => labelInline && css`


basalter:
Hi @jamime,

The PR is failing due to snapshots being outdated. I couldn't update them since I edited the files directly from Github.

About the height, it is set by some dynamic classes I have no idea where are coming from. I'd by happy to remove the proprety but I don't know how to do it.

Regards,
Carles

jamime:
@basalter In order to fix this, you will need to pull the branch down and edit it locally. You will need to include updated tests and snapshots when doing so.

IMO the correct fix is to remove the IE fix here. We no longer support IE11
https://github.com/Sage/carbon/blob/c93b3d39ed93fc481d8c2c4cc525c6bba914abd1/src/__experimental__/components/input/input-presentation.style.js#L24L27

If you would like to make these changes we will review your PR.

basalter:
> @basalter In order to fix this, you will need to pull the branch down and edit it locally. You will need to include updated tests and snapshots when doing so.
> 
> IMO the correct fix is to remove the IE fix here. We no longer support IE11
> https://github.com/Sage/carbon/blob/c93b3d39ed93fc481d8c2c4cc525c6bba914abd1/src/__experimental__/components/input/input-presentation.style.js#L24L27
> 
> If you would like to make these changes we will review your PR, also please include all fields in the PR. I see you removed the section on how to test your changes.
> 
> Thanks,
> Jamie

Getting rid of IE11 compatibility is always a good idea 😄 In deed, removing the height property is the right thing to do. However, I am not familiar enough with the project to be confident to make such changes. I'll close this PR and let you guys work on it in another occasion, given removing this property is a somehow critical change.

Cheers!

jamime:
No problem, thanks anyway.

We have a ticket on our backlog to address this FE-2538

PR #2671, opened 2020-02-25
test: integrate applitools

aaron66:
Add Applitools to project with some example tests

### Proposed behaviour
Add Applitools to project
Applitools hidden behind env var CYPRESS_APPLITOOLS, set to false by default
Applitools started before each feature that uses the @applitools tag
Button.feature screen capture of iframe
configurableItems.feature capture of element within the iframe
APPLITOOLS_API_KEY needs setting on host

Run Cypress `./node_modules/.bin/cypress open --env CYPRESS_APPLITOOLS=true`

Plan is to leave Applitools behind the env var until we are satisfied with the results and that they are consistent 

I will be adding visual checks to a range of components and then create an additional nightly build to run them


cypress/features/visual/alert.feature
```diff
@@ -0,0 +1,26 @@
...
+      | !@#$%^*()_+-=~[];:.,?{} |
+      | ÄÖÜßäöüß                |
+  # @ignore because of FE-1447
+  # | <>                       |
+
+  @applitools
+  Scenario: Check open click event
+    When clear all actions in Actions Tab
+      And I open component preview
+    Then Element displays correctly
```
sianford: Scenario should be:
Given clear all actions in Actions Tab (setting pre-requisites)
When I open component preview (Test point)
Then Element displays correctly (Result)

sianford: I think the Feature section for these tests should specifically state these are visual tests to distinguish them from those in the /regression suite, e.g. "Feature: I want to check Alert component displays correctly when properties are changed". Similarly for Scenario Outline and Scenario.

cypress/features/visual/button.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Button component
+  I want to check Button component properties
```
sianford: As above with Alert Feature comment.

cypress/support/index.js
```diff
@@ -62,3 +62,48 @@ before(() => {
 });
 
 Cypress.Screenshot.defaults({ screenshotOnRunFailure: DEBUG_FLAG });
+
+// Configure custom commands eyes-cypress
+import '@applitools/eyes-cypress/commands'
```
DlgSHi: import should be declared at the beginning of this file

cypress/support/index.js
```diff
@@ -62,3 +62,48 @@ before(() => {
...
+
+const {
+  Before,
+  After
+} = require("cypress-cucumber-preprocessor/steps");
+
+
+Before({ tags: "@applitools" }, () => {
+  // debugger
```
DlgSHi: remove this line please, if it isn't necessary

cypress/support/index.js
```diff
@@ -62,3 +62,48 @@ before(() => {
...
+
+  cy.eyesOpen({
+    appName: 'Carbon Sage',
```
DlgSHi: please create `applitools.config.js` in the same location where we have `cypress.json` file and put all additional setup there using json format as it is mentioned [here](https://www.npmjs.com/package/@applitools/eyes-cypress#configure-plugin-and-commands) 
```
module.exports = {
  appName: 'My app',
  showLogs: true,
  batchName: 'My batch'
  ...
  // all other configuration variables apply
}

File: cypress/support/index.js [hunk 6]
```diff
@@ -62,3 +62,48 @@ before(() => {
...
+    matchLevel: 'Strict', // Strict, Exact, Layout and Content - https://help.applitools.com/hc/en-us/articles/360007188591-Match-Levels
+    envName: 'Development',
+    testName: cy.state('ctx').test.title, // Assign the current test name to Applitools test name
+    batchName: cy.state('ctx').test.parent.title, // Assign the feature name to Applitools batch name
+    browser: [            
+              // {width: 800, height: 600, name: 'firefox'},
+              // {width: 1024, height: 768, name: 'chrome'},
```
DlgSHi: we should remove all that commented lines of code if those aren't necessary

cypress/support/step-definitions/applitools-steps.js
```diff
@@ -0,0 +1,11 @@
+When('Element displays correctly', () => {
```
DlgSHi: This step is used as Then in bdd
```suggestion
Then('Element displays correctly', () => {
```

cypress/features/visual/alert.feature
```diff
@@ -0,0 +1,26 @@
...
+  I want to change Alert component properties
+
+  Background: Open Alert component page
+    Given I open "Alert" component page with button
+
+  @applitools
+  Scenario Outline: Change Alert component title to <title>
+    When I set title to "<title>"
+      And I open component preview
+    Then Element displays correctly
```
DlgSHi: this step is not checking title, just the correct visibility of inner content of iFrame, so we need to remove from here this step `When I set title to "<title>"` and `Examples`

package.json
```diff
@@ -19,6 +19,7 @@
     "test-cypress": "cypress open",
     "test-cypress-build": "cypress run --spec ./cypress/features/build/build*.feature --browser chrome --reporter spec",
     "test-cypress-accessibility": "cypress run --spec ./cypress/features/accessibility/accessibility.feature --browser chrome --reporter spec",
+    "test-cypress-visual": "cypress run --spec './cypress/features/visual/*.feature' --browser chrome --reporter spec",
```
DlgSHi: here we have the configuration for `--browser chrome` but in the setup we are using `Edge`

package.json
```diff
@@ -62,21 +63,22 @@
     "i18n-js": "^3.0.0"
```
DlgSHi: do we need those all updates for devDependencies?

cypress/features/visual/alert.feature
aaron66: Agree, these are just demo tests which have been copied from regression

aaron66: We are going to change approach for visual tests, using existing ones rather than writing a whole new suite

cypress/features/visual/button.feature
aaron66: as above

package.json
aaron66: The test runs in chrome but then we use applitools to test edge, dosnt matter anyway as it has been decided to only test chrome

package.json
aaron66: no

cypress/support/index.js
```diff
@@ -63,47 +68,23 @@ before(() => {
...
-              //   name: 'chrome' // optional, just to make it explicit this is browser emulation and not a real device. Only chrome is supported for device emulation.
-              // }
-            ]
-  
-  }); 
-});
+if ( Cypress.env('CYPRESS_APPLITOOLS') ) {
+  Before({ tags: "@applitools" }, () => {
+    applitools_settings.testName = cy.state('ctx').test.title
+    applitools_settings.batchName = cy.state('ctx').test.parent.title
```
DlgSHi: why we can't set those settings in applitools  config ?

package.json
```diff
@@ -14,7 +14,7 @@
     "test-cypress": "cypress open",
     "test-cypress-build": "cypress run --spec ./cypress/features/build/build*.feature --browser chrome --reporter spec",
     "test-cypress-accessibility": "cypress run --spec ./cypress/features/accessibility/accessibility.feature --browser chrome --reporter spec",
-    "test-cypress-visual": "cypress run --spec './cypress/features/visual/*.feature' --browser chrome --reporter spec",
+    "test-cypress-visual": "cypress run --spec './cypress/features/regression/*.feature' --env TAGS='@applitools',CYPRESS_APPLITOOLS=true --browser chrome --reporter spec",
```
DlgSHi: if we are running only one `*.feature` file we should pass it in `--spec` parameter, it will speed up those tests

DlgSHi:
Also needs to be checked why the build is failing

cypress/support/index.js
aaron66: `cy.state('ctx').test.title` is out of context in applitools config file

package.json
aaron66: isn't it already doing that? `--spec './cypress/features/visual/*.feature'` ??

cypress/support/index.js
aaron66: Could only get this working with in the Before

package.json
DlgSHi: nope, because the file is in the `cypress/features/regression/button.feature` instead of `'./cypress/features/visual/`
Visual folder doesn't exist

aaron66: Sorry miss types my last comment, the command is running all feature files in the 
features/regression directory with the tag applitools

aaron66: ` "test-cypress-visual": "cypress run --spec './cypress/features/regression/*.feature' --env TAGS='@applitools',CYPRESS_APPLITOOLS=true --browser chrome --reporter spec",`

DlgSHi: yes, but cypress will scan all feature files inside

aaron66: OK have removed until i know which files to run

DlgSHi:
we are missing `"test-cypress-visual": "cypress run --spec './cypress/features/visual/*.feature' --browser chrome --reporter spec"` in package.json

DlgSHi:
There is no possibility to run applitools tests without APPLITOOLS_KEY

jamime:
Converting to draft to prevent merge.

sianford:
Closing this PR as this work has continued in another branch.

PR #2682, opened 2020-02-27
chore: update babel config

edleeks87:
### Proposed behaviour

The new config should allow the `lib` directory to be built without extending the config in `carbon-factory`

### Current behaviour

Currently the building of the `lib` relies on extending the config from `carbon-factory`

### Checklist

 <del>- [ ] Unit tests <del>
 <del>- [ ] Cypress automation tests <del>
 <del>- [ ] Storybook added or updated <del>
 <del>- [ ] Theme support <del>
 <del>- [ ] Typescript `d.ts` file added or updated <del>

### Additional context

A step in removing `carbon-factory` dependency. The required babel dependencies will be added when `carbon-factory` is removed

### Testing instructions

The build will fail if it does not transpile the code

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,22 @@
-const path = require('path');
-
 module.exports = {
-  extends: path.resolve(process.cwd(), './node_modules/carbon-factory/babel.config.js')
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'usage',
+        targets: {
+          browsers: 'IE 11'
+        },
+        corejs: 3
+      }
+    ],
+    '@babel/preset-react',
+    '@babel/preset-typescript'
+  ],
+  plugins: [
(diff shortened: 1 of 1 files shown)


babel.config.js
```diff
@@ -1,5 +1,22 @@
...
-
 module.exports = {
-  extends: path.resolve(process.cwd(), './node_modules/carbon-factory/babel.config.js')
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'usage',
+        targets: {
+          browsers: 'IE 11'
```
chrisbarber86: Since we're dropping IE11 support going forwards - do we need this?

edleeks87: Good point will remove

edleeks87: Good point I'll remove

jamime:
We decided to keep IE11 and remove it in a later PR with the other IE11 workarounds.

PR #2699, opened 2020-03-04
feat(select): add value to custom filter function

sagedemdreisbusch:
Current behaviour
The customFilter function has got following params: (text, filter)

Desired behaviour
And we request (text, filter, value)
actually the order should be another one but that would be a breaking change

### Proposed behaviour
This PR adds value support to the customFilter-Function of the select component

### Current behaviour
There is no chance to access the value object in the customFilter-function

### Checklist
- [X] Release notes (Conventional Commits) 
- [X] Unit tests
- [X] Cypress automation tests
- [X] Storybook added or updated
- [X] Theme support
- [X] Typescript `d.ts` file added or updated


dawidzarzycki:
@sagedemdreisbusch please, remember about squash some commits before merge to remove some of them like eg. `linters`

src/__experimental__/components/select/option.component.js
```diff
@@ -13,7 +13,7 @@ Option.propTypes = {
   /** Optional: alternative rendered content, displayed within <SelectList> of <Select> (eg: an icon, an image, etc) */
   children: PropTypes.node,
   /** The option's invisible internal value */
-  value: PropTypes.string.isRequired,
+  value: PropTypes.any.isRequired,
```
grabkowski: I think that we should avoid using `any` type, you could use `oneOfType`, e.g.: `PropTypes.oneOfType([PropTypes.object, PropTypes.string])`

sagedemdreisbusch: Ok, just fixed that

sagedemdreisbusch:
Have got a codesandbox-sample here: https://codesandbox.io/s/elegant-cartwright-7rz0v?fontsize=14&hidenavigation=1&theme=dark

DlgSHi:
Correct [sandbox](https://codesandbox.io/s/compassionate-dan-n8tke)

PR #2700, opened 2020-03-04
fix(icon-button): stop handling esc key - FE-2599 

damianOfKyan:
### Proposed behaviour
Only Enter and Space key should trigger onAction callback.

### Current behaviour
Esc key triggers onAction callback.

### Checklist

- [ ] Release notes (Conventional Commits) 
<del>- [ ] Unit tests
<del>- [ ] Storybook added or updated
<del>- [ ] Theme support

### Testing instructions
Opening dialog:
http://localhost:9001/?path=/story/dialog--default

When "X" (close icon) is focused, ESC key press should not close Dialog when dialog prop/knob "disableEscKey" is set to true.

### DIFF
--- a/cypress/features/regression/dialog.feature
+++ b/cypress/features/regression/dialog.feature
@@ -6,6 +6,24 @@ Feature: Dialog component
       And I open component preview
 
   @positive
+  Scenario: Esc key pressed on Close Icon does not close Dialog with disableEscKey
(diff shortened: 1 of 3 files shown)


damianOfKyan:
After UX feedback from Ben, this implementation is no longer needed. We do want to handle Esc key on close icon.

PR #2701, opened 2020-03-05
chore(stale): adds configuration for github stale app

ianforster-sage:
### Proposed behaviour
The Stale App has been enabled in Github, this configuration sets the threshold for labelling issues and PRs at 365 days. It will not close anything and it will not set to stale, anything with labels `pinned` or `security`

### Current behaviour
Currently we use manual review for managing old issues or PRs.

### Checklist


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had recent activity. 
+  It will be reviewed by the core Carbon team and may be closed if no further activity occurs. Thank you for your contributions.
```
chrisbarber86: Perhaps need to add here that if someone doesn't think it's stale, then they should add a comment

ianforster-sage: updated

Issue #2567, opened 2019-12-17
Select component : Error if only one option child is set


ardvi:
https://codesandbox.io/s/carbon-quickstart-d104r

to reproduce:
Put a Select component with only one Option in child, Set a value to Select.
> a error 'optionsComponents.find is not a function' throwed.

jamime:
Still an issue on 12.4.0.
On backlog FE-2547

ianforster-sage:
Fixed in https://github.com/Sage/carbon/pull/2735

PR #2735, opened 2020-03-18
fix(select): resolved error when only one option child is set - FE-2547

damianOfKyan:
Fixes FE-2547
Fixes GitHub issue #2567

### Proposed behaviour
Select can accept one or more `<Option />` as children.

### Current behaviour
Select breaks when one `<Option />` is given as child.

### Checklist

- [x] Unit tests
<dev>- [ ] Cypress automation tests
<dev>- [ ] Storybook added or updated
<dev>- [ ] Theme support
<dev>- [ ] Typescript `d.ts` file added or updated

### Testing instructions
To experience breaking behaviour, on master branch modify the `defaultStory` story in `select.stories.js`.

```diff
const defaultComponent = () => {
   return (
     <State store={ singleSelectStore }>
       <Select ariaLabel='singleSelect' { ...commonKnobs(singleSelectStore) }>
-        { selectOptions }
+        <Option
+          key="a"
+          text="a"
+          value="1"
+        />
       </Select>
     </State>
   );
);
```

browser to:
http://localhost:9001/?path=/story/experimental-select--default
click on select, it should raise the error in the browser console.

Checkout this branch and repeat the steps. There should be no errors.


src/__experimental__/components/select/select.spec.js
```diff
@@ -462,4 +462,35 @@ describe('Select', () => {
       assertStyleMatch({ color: '#FFFFFF' }, styleWrapper.toJSON(), { modifier: `&:hover ${StyledIcon}` });
     });
   });
+
+  describe('renders select with one option', () => {
+    it('renders correctly', () => {
+      const wrapper = mount(
+        <Select>
```
mkrds: Maybe use `shallow` here because the generated snapshot is 3k LOC

damianOfKyan: fixed.

src/__experimental__/components/select/__snapshots__/select.spec.js.snap
```diff
@@ -99,6 +99,39 @@ exports[`Select renders only the Textbox when closed 1`] = `
 </styled.div>
 `;
```
mkrds: I think you forgot to remove this :)

damianOfKyan: yep, fixed :)

PR #2739, opened 2020-03-19
feat(textbox): add override styles functionality (FE-2631)

grabkowski:
### Proposed behaviour
Textbox component should have a functionality to override it's styles, e.g.: set fixed width of the input or change it's background. 

### Current behaviour
Textbox styles cannot be overriden by Carbon user.

### Checklist

- [x] Unit tests
- [x] Theme support

### Additional context
Required to set fixed width of post code field in Address Fieldset

### Testing instructions
* npm start
* Go to http://localhost:9001/?path=/story/test-textbox--styles-overriden

### DIFF
--- a/src/__experimental__/components/label/label.component.js
+++ b/src/__experimental__/components/label/label.component.js
@@ -22,7 +22,8 @@ const Label = (props) => {
     tooltipMessage,
     useValidationIcon,
     htmlFor,
-    tabIndex
+    tabIndex,
+    styleOverride
   } = props;
   const labelProps = filterByProps(props, [
     'theme',
@@ -78,6 +79,7 @@ const Label = (props) => {
       { ...labelProps }
       id={ labelId }
       htmlFor={ htmlFor }
+      styleOverride={ styleOverride }
     >
       {children}
       {icon()}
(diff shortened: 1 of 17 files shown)


src/__experimental__/components/label/label.component.js
```diff
@@ -109,12 +111,15 @@ Label.propTypes = {
   /** Set focus possibilities to an <IconWrapperStyle /> element.
    *  More information: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
   */
-  tabIndex: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+  tabIndex: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Allows to override existing component styles */
+  styleOverride: PropTypes.oneOfType([PropTypes.func, PropTypes.object])
```
jamime: Why does this include `PropTypes.func`?  It looks like only an object would work


```js
${({ styleOverride }) => styleOverride};
```

grabkowski: when you pass a function, then you'll have access to styled component props in that function arguments (theme object for example). The function is automatically invoked, It's counterintuitive, but that is how it works in Styled Components.

jamime: Thanks for clarifying 🙂

PR #2743, opened 2020-03-24
fix(button-toggle): add id prop support

ianforster-sage:
### Proposed behaviour
The id prop passed to ButtonToggleGroup should be in the DOM.
The id prop passed to ButtonToggle should be exposed in the onChange/onBlur event it should also be in the DOM.
The checked prop of the ButtonToggle radio button is available to be set.

### Current behaviour
The id prop is not available in the DOM
Within a ButtonToggleGroup, you can't set the checked property of the ButtonToggle radio button.

### Checklist
- [ ] Unit tests
- [ ] Theme support


ianforster-sage:
Closing. Fix applied in v10.

PR #2757, opened 2020-03-27
fix(icon-button): hover background over focus outline (FE-2634)

grabkowski:
### Proposed behaviour
Change z-index of focused Icon Button to render over it's siblings.
image

### Current behaviour
When there is multiple icon buttons next to each other, one is focused, hovering over adjacent buttons their background will cover the focus outline.
image

### Checklist
~~- [ ] Unit tests~~
~~- [ ] Theme support~~

### Testing instructions
* npm start
* Go to http://localhost:9001/?path=/docs/design-system-batch-selection--basic
* Focus on one of the Icon Buttons and hover over another

### DIFF
--- a/src/components/batch-selection/__snapshots__/batch-selection.spec.js.snap
+++ b/src/components/batch-selection/__snapshots__/batch-selection.spec.js.snap
@@ -50,6 +50,7 @@ exports[`BatchSelection component should render correctly 1`] = `
   color: rgba(0,0,0,0.9);
   background-color: transparent;
   outline: solid 3px #FFB500;
+  z-index: 1;
 }
 
 .c3:hover {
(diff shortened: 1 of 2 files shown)


jamime:
https://codesandbox.io/s/confident-morse-50mqo

I noticed that the outline now overlaps the link, should we increase the padding or leave this to the developer?

image

grabkowski:
> https://codesandbox.io/s/confident-morse-50mqo
> 
> I noticed that the outline now overlaps the link, should we increase the padding or leave this to the developer?
> 
> image

This PR fixes only the issue with Icon Button Component when placed next to each other, and in my opinion, overlapping of other components should not be taken into consideration in the scope of this component.
Link and buttons are rendered as children in Batch Selection, developer could just wrap these buttons in a styled div to fix the overlapping outline.

PR #2765, opened 2020-03-30
feat(action-popover): support for aligning menu to right of button

edleeks87:
### Proposed behaviour

Passing the `rightAlignMenu` prop will allow the `Menu` to align to the right of the `MenuButton`, by default the `Menu` will continue to open to the left

### Current behaviour

Currently the Menu opens to the left of the button
### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>

### Additional context

Further PR will allow a custom `MenuButton` to be passed into the Popover by a consumer, I thought it best to separate the work.

Default behaviour with no prop set:
image

Custom behaviour with prop set:
image

### Testing instructions

https://github.com/Sage/carbon/pull/2765#issuecomment-606117514

### DIFF
--- a/src/components/action-popover/action-popover.spec.js
+++ b/src/components/action-popover/action-popover.spec.js
@@ -209,6 +209,16 @@ describe('ActionPopover', () => {
     }).not.toThrow();
   });
 
+  it('opens with the menu left position assigned so it is to right of the button if the "rightAlignMenu" prop is true',
+    () => {
+      render({ rightAlignMenu: true });
+      const { menu } = getElements();
+      assertStyleMatch({
+        right: undefined,
+        left: '0'
+      }, menu);
+    });
+
   describe.each([
     ['Click handlers', 'Clicking', item => item.simulate('click')],
     ['Keypress handlers', 'Pressing Enter', item => simulate.keydown.pressEnter(item)]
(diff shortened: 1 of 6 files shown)


src/components/action-popover/action-popover.spec.js
```diff
@@ -209,6 +209,14 @@ describe('ActionPopover', () => {
     }).not.toThrow();
   });
 
+  it('renders with the menu aligned to the right if the "rightAlignMenu" prop is true', () => {
+    render({ rightAlignMenu: true });
+    const { menu } = getElements();
+    assertStyleMatch({
+      right: undefined
```
damianOfKyan: I think this should be `right: 0`

damianOfKyan: I think this should be `right: 0`

edleeks87:
> affectionate-surf-0wu7u

Testable demo
https://codesandbox.io/s/awesome-night-ud8g1

src/components/action-popover/action-popover.spec.js
edleeks87: The assertion is correct but my description is vague, I've updated the test and add more to the assertion to hopefully make it clearer 😄

PR #2767, opened 2020-03-31
fix(search): close icon persists when field is empty (FE-2638)

grabkowski:
### Proposed behaviour
Close icon should be replaced by the search icon when the text is cleared.
image

### Current behaviour
When the text field is cleared using the close button, the close button is not replaced by the search icon.
image

### Checklist

~~- [ ] Unit tests~~
~~- [ ] Theme support~~

### Testing instructions
1. run: npm start
2. go to: http://localhost:9001/?path=/story/test-search--basic
3. uncheck the `searchButton` knob
4. type `test` in the text field
5. click on the `close` button in the field

### DIFF
--- a/src/__experimental__/components/search/search.component.js
+++ b/src/__experimental__/components/search/search.component.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useMemo } from 'react';
+import React, { useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import invariant from 'invariant';
 import StyledSearch, { StyledSearchButton, StyledButtonIcon } from './search.style';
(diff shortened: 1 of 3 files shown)


src/__experimental__/components/search/search.component.js
```diff
@@ -17,9 +17,25 @@ const Search = ({
   );
   let inputRef = useRef(null);
   const [searchValue, setSearchValue] = useState(initialValue);
-  const [isActive, setIsActive] = useState(inputRef.current === document.activeElement);
+  const [isFocused, setIsFocused] = useState(inputRef.current === document.activeElement);
```
mkrds: I think there's no need for this `inputRef.current === document.activeElement` comparison as `inputRef.current` is undefined when `useState` is initialized anyway

src/__experimental__/components/search/search.component.js
```diff
@@ -17,9 +17,25 @@ const Search = ({
   );
   let inputRef = useRef(null);
```
mkrds: Is seems that this ref is not used anywhere

grabkowski: removed

src/__experimental__/components/search/search.component.js
grabkowski: you're right, removed

PR #2771, opened 2020-03-31
docs(pod): add address pod example (FE-2632)

grabkowski:
### Proposed behaviour
Address pod mdx example
image

### Checklist

~~- [ ] Unit tests~~
~~- [ ] Theme support~~

### Testing instructions
1. run: npm start
2. go to: http://localhost:9001/?path=/docs/design-system-pod--address-example

### DIFF
--- a/src/components/pod/pod.stories.mdx
+++ b/src/components/pod/pod.stories.mdx
@@ -1,0 +1,44 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+import Pod from './';
+import Heading from '../heading';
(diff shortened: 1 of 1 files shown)


harpalsingh:
Can we update the padding to 16px, rather than 15px.If you also move the position of the button 8px top and 8px right, it will align the actual text with the other content.

grabkowski:
@harpalsingh position of the edit button changed as suggested.
I can't update the Pod padding as it's not part of this task and it will affect all currently existing Pods which will be a breaking change.

PR #2773, opened 2020-04-01
feat(action-popover): add support for custom menu button override FE-2329

edleeks87:
### Proposed behaviour

A new renderButton prop has been added to allow the menu button to be overriden with a custom
component passed in by the consumer.

### Current behaviour

The component does not support a custom button, it renders only the Icon

### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>

### Additional context

I have built a demo of how to use the `ActionPopover` component to deliver a `ServicesColumn`

### Testing instructions

The demo is available in `test-action-popover--services-column` in storybook.
I will demonstrate the custom button functionality using the codesandbox demo link create as part of the build

Services Column closed with custom button:
image

Services Column open:
image

### DIFF
--- a/src/components/action-popover/action-popover.spec.js
+++ b/src/components/action-popover/action-popover.spec.js
@@ -45,16 +45,32 @@ describe('ActionPopover', () => {
     const defaultProps = {
       children: [
         <ActionPopoverItem
+          key='item-1'
           icon='pdf'
           { ...{ onClick: onClickWrapper('pdf') } }
           disabled
         >
           Download PDF
         </ActionPopoverItem>,
-        <ActionPopoverItem icon='email' { ...{ onClick: onClickWrapper('email') } }>Email Invoice</ActionPopoverItem>,
-        <ActionPopoverItem icon='print' { ...{ onClick: onClickWrapper('print') } }>Print Invoice</ActionPopoverItem>,
-        <ActionPopoverDivider />,
-        <ActionPopoverItem icon='csv' { ...{ onClick: onClickWrapper('csv') } }>Download CSV</ActionPopoverItem>,
+        <ActionPopoverItem
+          key='item-2'
+          icon='email' { ...{ onClick: onClickWrapper('email') } }
+        >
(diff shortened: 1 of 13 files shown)


src/components/action-popover/action-popover.spec.js
```diff
@@ -888,4 +899,33 @@ describe('ActionPopover', () => {
...
+          </ActionPopover>
+        </ThemeProvider>
+      ).find(ActionPopover);
+
+      const button = popover.find(Button);
+
+      expect(popover.find(Icon).first().exists()).toBeFalsy();
+      expect(button.exists()).toBeTruthy();
+      expect(button.props().tabIndex).toEqual(-1);
+      expect(button.props()['data-element']).toEqual('action-popover-menu-button');
```
ianforster-sage, line 918: Would this be a good candidate for a snapshot test? Seems to be a case of "given a renderButton prop I expect the markup to be like this".

edleeks87: I considered it here but I had to add a fair few snapshots when I was doing the submenu work and felt that adding another might be excessive. I decided against it here as I found an alternative way of achieving the same assertion but change it to a snapshot if the consensus is that it's preferred

edleeks87:
> Hi, looks good to me except this error
> 
> ```
>     Warning: Each child in a list should have a unique "key" prop.    
>     Check the render method of `ActionPopover`. See https://fb.me/react-warning-keys for more information.
>         in WithTheme(ActionPopoverItem)
>         in ActionPopover
>         in ThemeProvider
> ```
> 
> It occurs during `npm test src\components\action-popover\action-popover.spec.js`.
> The `key` error also show up in the story `Test/Action Popover/ servieces column` and in the `Design System/ActionPopover` in the DOCS tab. But I think this one is connected with the `services column` also

Good spot fixed in https://github.com/Sage/carbon/pull/2773/commits/3f31cf548379076be2df01af84dfeff9e9b0c676

PR #2814, opened 2020-04-03
docs(button): add a story with all button options

jamime:
### Proposed behaviour

- Convert to CSF format
- Add new story with all button variations for visual regression testing

### Current behaviour

- Using deprecated`storiesOf` format
- Only posible to see each variation by changing the knobs

### Checklist

<del>- [ ] Unit tests
<del>- [ ] Theme support

### Additional context
@sianford @aaron66 will use this to complete their applitools investigation. 

### Testing instructions

- `npm start`
- http://localhost:9001/iframe.html?id=button--all-buttons&theme=mint
- http://localhost:9001/iframe.html?id=button--all-buttons&theme=aegean
- http://localhost:9001/iframe.html?id=button--all-buttons&theme=none

_Note: these pages take a while to load due to the SVG icons._


src/components/button/button.stories.js
```diff
@@ -38,45 +30,128 @@ const getKnobs = () => {
...
-    return (
-      <div>
-        <Button { ...props }>{ children }</Button>
-        <Button { ...props }>{ children }</Button>
-      </div>
-    );
-  };
+export const allButtons = () => {
+  const types = ['primary', 'secondary', 'tertiary'];
+  const sizes = ['small', 'medium', 'large'];
```
grabkowski, line 76: you could use `OptionsHelper.sizesRestricted` here

src/components/button/button.stories.js
```diff
@@ -38,45 +30,128 @@ const getKnobs = () => {
...
-    const { children } = props; // eslint-disable-line react/prop-types
-    return (
-      <div>
-        <Button { ...props }>{ children }</Button>
-        <Button { ...props }>{ children }</Button>
-      </div>
-    );
-  };
+export const allButtons = () => {
+  const types = ['primary', 'secondary', 'tertiary'];
```
grabkowski, line 75: don't we also have a `darkBackground` type?

src/components/button/button.stories.js
```diff
@@ -38,45 +30,128 @@ const getKnobs = () => {
...
+                </Button>
+
+                {size === 'large' && (
+                  <Button
+                    iconPosition={ iconPosition }
+                    iconType={ icon }
+                    size={ size }
+                    disabled
+                    buttonType={ type }
+                    subtext='line two'
```
grabkowski, line 146: I think that you could extract `iconPosition, iconType, size and buttonType` to an object and just add these props as `{ ...objectName }`

src/components/button/button.stories.js
```diff
@@ -38,45 +30,128 @@ const getKnobs = () => {
...
-        <Button { ...props }>{ children }</Button>
-        <Button { ...props }>{ children }</Button>
-      </div>
-    );
-  };
+export const allButtons = () => {
+  const types = ['primary', 'secondary', 'tertiary'];
+  const sizes = ['small', 'medium', 'large'];
+  return <>
+    {['before', 'after'].map(iconPosition => (
```
edleeks87, line 78: Could use `buttonIconPositions` from OptionsHelper?

src/components/button/button.stories.js
grabkowski: if yes, then you could use `OptionsHelper.buttonTypes`

src/components/button/button.stories.js
jamime: I've done it in a new commit, can you let me know what you think. I'm not sure if it helps the readability of the code. I did the other OptionHelper too, but again I think it makes the code harder to read.

PR #2824, opened 2020-04-03
fix(select): allow passing custom validation icon type prop to component - FE-2639

mkrds:
### Proposed behaviour
Select should pass `hasError`, `hasWarning`, `hasInfo`, `tooltipMessage` and `inputIcon` to the `Textbox` component

### Current behaviour
Currently `inputIcon` prop get's overwritten internally to the 'dropdown' type. 

### Checklist

- [x] Unit tests
- [ ] Theme support

### Testing instructions
Check `external validations` story on Storybook


src/__experimental__/components/select/select.spec.js
```diff
@@ -525,4 +525,29 @@ describe('Select', () => {
...
+      const props = { hasWarning: true, inputIcon: 'warning', tooltipMessage: 'Warning' };
+      const wrapper = renderWrapper({ props });
+      expect(wrapper.find(Textbox).props()).toMatchObject(props);
+    });
+
+    it('should pass info props to the Textbox component', () => {
+      const props = { hasInfo: true, inputIcon: 'info', tooltipMessage: 'Info' };
+      const wrapper = renderWrapper({ props });
+      expect(wrapper.find(Textbox).props()).toMatchObject(props);
+    });
```
grabkowski, line 546: you could use `describe.each` here:
```
describe.each([
  ['error', { hasError: true }],
  ['warning', { hasWarning: true }],
  ['info', { hasInfo: true }]
])(
  'should pass %s props to the Textbox component',  (validationType, validationProp) => {
```

PR #2840, opened 2020-04-06
test(grid): update test to remove internal styled comp from snapshot

edleeks87:
### Proposed behaviour

Snapshots for both components should not have internal styled components as they our unit tests should only focus on carbon components

### Current behaviour

Both components have snapshots that contain internal styled-components that are subject to break every time the dependency version is updated

### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>

### Additional context

Related to https://github.com/Sage/carbon/pull/2775


src/components/grid/grid.spec.js
```diff
@@ -148,8 +148,15 @@ describe('Grid', () => {
...
-      expect(elem).toMatchSnapshot();
+      expect(
+        shallow(
+          <GridContainer id='testContainer'>
+            <GridItem responsiveSettings={ [item11500, item11300, item1900] }>1</GridItem>
+            <GridItem responsiveSettings={ [item21500, item21300, item2900] }>2</GridItem>
+            <GridItem responsiveSettings={ [item31500, item3900, item31300] }>3</GridItem>
+          </GridContainer>
+        )
+      ).toMatchSnapshot();
```
grabkowski, line 159: I think that the main reason of this snapshot was to render CSS styles, which are missing on shallow render

edleeks87: Yeah fair point, I've updated to use `TestRenderer.create` as it will maintain the css

PR #2861, opened 2020-04-08
feat(form): create new form component - FE-2530

mkrds:
### Proposed behaviour
This PR implements new Form component which is meant to be only visual component without any validation handling mechanism implemented.

### Current behaviour
Current forms, both experimental and deprecated one, are bloated with unnecessary code.

### Checklist

- [x] Unit tests
<del>- [ ] Theme support</del>

### Testing instructions
Check `Design System/Form` docs to see different Form variations.

Also stories `Dialog` and `DialogFullScreen` should be checked out if the implemented changes didn't break currently used components.


src/components/form/form.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+/>
+
+# Form
+Represents a document section containing interactive controls for submitting information.
+
+## Quick Start
+
+To use Form, import the `Form` and pass the content, usually various inputs, as children.
+
+Save button is meant to be passed as `saveButton` prop -
```
jamime, line 20: I think this is too detailed.
```diff
- Save button is meant to be passed as `saveButton` prop - 
- it will be wrapped internally by validation summary if `showSummary` prop is passed as `true`.

+ You can provide a save button using the `saveButton` prop.
+ You can hide the validation summary by setting the `showSummary` prop to false.
```

src/components/form/form.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+```jsx
+import { Form } from "carbon-react/lib/components/form"
+
+const MyComponent = () => (
+  <Form
+    onSubmit={ handleSubmit }
+    leftSideButtons={ <Button onClick={ handleCancel }>Cancel</Button> }
+    saveButton={ <Button buttonType='primary' type="submit">Save</Button> }
+    showSummary
```
jamime, line 31:
```suggestion
```

`showSummary` defaults to `true`

src/components/form/form.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+    >
+      <Textbox label="Textbox" />
+      <Textbox label="Textbox" />
+      <Textbox label="Textbox" />
+    </Form>
+  </Story> 
+</Preview>
+
+### In Dialog
+It is a good idea to render `Form` as a content of `Dialog` component.
```
jamime, line 103:
```suggestion
It is possible to render `Form` as a content of `Dialog` component.
```

src/components/form/form.stories.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+<Preview>
+  <Story id="test-form--in-dialog" name="In Dialog" />
+</Preview>
+
+### In DialogFullScreen
+It is also a good idea to render `Form` as a content of `DialogFullScreen` component.
```
jamime, line 110:
```suggestion
It is possible to render `Form` as a content of `DialogFullScreen` component.
```

src/components/form/form-summary.component.js
```diff
@@ -0,0 +1,82 @@
...
+Summary.propTypes = {
+  type: PropTypes.oneOf(['error', 'warning'])
+};
+
+const FormSummary = props => (
+  <StyledFormSummary
+    showSummary={ props.errors > 0 || props.warnings > 0 }
+    data-element='form-summary'
+  >
+    <Summary type='error' { ...props } />
```
jamime, line 70: Instead of pluralizing the type, I think we should pass `errors` here. We should update `translation`, `translationKey` and `StyledInternalSummary` to use the plural here too.

We'd have to then only map the plural back to singular for the `icon`.

What do you think?

src/components/form/form.component.js
```diff
@@ -0,0 +1,163 @@
...
+  }, [checkStickyFooter]);
+
+  const removeStickyFooterListeners = useCallback(() => {
+    ElementResize.removeListener(formRef.current, checkStickyFooter);
+    window.removeEventListener('resize', checkStickyFooter, true);
+    window.removeEventListener('scroll', checkStickyFooter, true);
+  }, [checkStickyFooter]);
+
+  useEffect(() => {
+    if (stickyFooter && !stickyListenersAddedRef.current) {
```
jamime, line 65: I see that the cleanup of this effect will call `removeStickyFooterListeners` but won't reset the guard `stickyListenersAddedRef`.

> This is why React also cleans up effects from the previous render before running the effects next time
> https://reactjs.org/docs/hooks-effect.html

So if the component is rendered again, the listener will be removed and won't be added again.

src/components/form/form.component.js
```diff
@@ -0,0 +1,163 @@
...
+      ref={ formRef }
+      stickyFooter={ stickyFooter && isFooterSticky }
+      onSubmit={ onSubmit }
+      data-component='form'
+      { ...rest }
+    >
+      { children }
+      <StyledFormFooter
+        data-element='form-footer'
+        className={ isFooterSticky ? 'isSticky' : '' }
```
jamime, line 115: can we do this with a prop or do we need `.isSticky` for Dialog?

src/components/form/form.stories.js
```diff
@@ -0,0 +1,157 @@
...
+export const Basic = () => {
+  const { defaultProps } = Form;
+
+  const buttonAlignment = select(
+    'buttonAlignment',
+    OptionsHelper.alignBinary,
+    defaultProps.buttonAlignment
+  );
+
+  const showSummary = boolean('showSummary', defaultProps.showSummary);
```
edleeks87, line 35: As we are hoping to streamline our regression testing (see link below), it might be worth moving away from using `addon-knobs`, as it adds a lot of overhead, in favour of adding multiple demos to display the various configurations of the new Form.

https://github.com/Sage/carbon/diffs/1?base_sha=5dcd7ef085344df9c974a4232859cc4fbc4fe165&commentable=true&pull_number=2774&sha1=5dcd7ef085344df9c974a4232859cc4fbc4fe165&sha2=95389c39933f7e1723181501d5c3d621155b8ca1&short_path=2c80563&unchanged=expanded#motivation

src/components/form/form.component.js
mkrds: I need to override `StyledFormFooter` style when `Form` is rendered in `Dialog` component but only when the form footer is in sticky state. Adding `isSticky` class to `StyledFormFooter` is the easiest way to achieve it since I don't have access to its props in Dialog styles.

src/components/dialog/dialog.style.js
```diff
@@ -36,24 +38,35 @@ const DialogStyle = styled.div`
...
+      margin-left: -35px;
+      left: auto;
+      width: ${dialogSizes[size]};
+      position: fixed;
+    }
+  `}
+
   ${({ fixedBottom }) => fixedBottom && css`
       bottom: 0;
       min-height: 0px !important;
```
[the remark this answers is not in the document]
mkrds: It seems that it is required to override inline style `minHeight` which is applied to the dialog in line 194 of `dialog.component.js`

src/components/form/form.stories.js
mkrds: @edleeks87 I've addressed this comment and removed `form.stories.js` - everything is in `form.stories.mdx` now

PR #2873, opened 2020-04-14
fix(row): allow number prop type for columns, columnSpan, columnOffset

blaky:
### Proposed behaviour
Allow numeric value types for properties that are clearly meant to be numeric.

### Current behaviour
It only allows string type values for these properties, but the values actually handled as numbers. When a numeric value is provided, a notification is displayed.

### Checklist

- [X] Release notes (Conventional Commits) 
- <del>Unit tests</del>
- <del>Cypress automation tests</del>
- <del>Storybook added or updated</del>
- <del>Theme support</del>
- [X] Typescript `d.ts` file added or updated

### Additional context
This PR also brings TypeScript declarations for the Row and Column components.

### Testing instructions
It is a very technical change, the only testing I can think of is automated regression


blaky:
Yes, I was in doubt which one to use. Please see the commit message updated.

Issue #2748, opened 2020-03-25
TextBox: Validation icon wraps when rendered within InlineInputs container


simon-johnstone:
### Current behaviour

When rendering a TextBox, with validation, within an InlineInputs container, the validation icon 'wraps' causing the TextBox to be rendered with increased height.

### Expected behaviour

A TextBox with validation should be rendered the same way irrespective of whether it's a child of an InlineInputs container or not.

### Reproducible example

Sample that reproduces the problem...
https://codesandbox.io/s/textbox-with-validation-within-inlineinputs-ijk06

### Suggested solution(s)

(See expected behaviour)

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 13.4.0
| carbon-factory   | 
| react-scripts    | 3.3.0
| React            | 16.12.0
| Browser          | Chrome  (80.0.3987.149)
| npm              | 6.12.1
| Operating System | Win10

ianforster-sage:
On backlog, ref: FE-2671

PR #2875, opened 2020-04-16
fix(inline-input): fix to input width that affects input icon - FE-2671

damianOfKyan:
Fixes FE-2671
Fixes GitHub Issue: #2748

### Proposed behaviour
InlineInputs should display a validation icon correctly by default aligned to the right hand side of the input field.

inline-input-fixed

### Current behaviour
InlineInputs are using up 100% of the width pushing a validation icon to the new line. (wrap)

Screenshot 2020-04-16 at 13 40 02

### Checklist

- [x] Unit tests
- [x] Theme support

### Testing instructions
1. checkout branch
2. navigate to http://localhost:9001/?path=/story/inlineinputs--validations
3. Investigate first inline input that should have a icon to the right hand side

Inspecting element in the browser and changing input width: auto; to width: 100% will cause the issue to re-appear confirming that width: auto; resoles the issue.


src/components/inline-inputs/inline-inputs.spec.js
```diff
@@ -67,9 +67,9 @@ describe('Inline Inputs', () => {
       }, wrapper, { modifier: '.carbon-row' });
     });
 
-    it('then all inputs should have 100% width', () => {
+    it('then all inputs should have auto width', () => {
```
grabkowski, line 70: I think it's no longer `auto`

damianOfKyan: fixed.

sianford:
Tested in Safari, Chrome, Firefox and Edge (for Mac) browsers. All issues resolved. The validation icon no longer wraps. All components work correctly. Testing passed ✅

PR #2882, opened 2020-04-17
feat(select): add max height prop to select to control height of list FE-2700

edleeks87:
### Proposed behaviour

Adds `maxHeight` prop which can be passed into the list to control the height of the options

### Current behaviour

There is no way of controlling the max height of the list

### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>
<del>- [] Typescript `d.ts` file added or updated<del>

### Additional context

image

image


edleeks87:
Closing as we have decided to avoid using style overriding props

PR #2884, opened 2020-04-21
fix(flat-table): update hover color for flat table row and cell

edleeks87:
### Proposed behaviour

Updates hover colour to be aligned with DLS, added `flatTable` object and a `default` `hover` property to theme

### Current behaviour

Colour is not what is described in DLS

### Checklist

- [x] Unit tests
<del>- [ ] Theme support<del>

### Additional context

image

### DIFF
--- a/src/style/themes/base/base-theme.config.js
+++ b/src/style/themes/base/base-theme.config.js
@@ -79,6 +79,13 @@ export default (palette) => {
       inactiveSelectorBackground: palette.slateTint(80)
     },
 
+
+    flatTable: {
+      default: {
+        hover: palette.slateTint(95)
+      }
+    },
+
     help: {
       color: blackWithOpacity(0.65),
       hover: blackWithOpacity(0.9)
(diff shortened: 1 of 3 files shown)


src/style/themes/base/base-theme.config.js
```diff
@@ -130,6 +130,7 @@ export default (palette) => {
       tertiary: palette.slateTint(10),
       header: palette.slateTint(20),
       hover: palette.slateTint(90),
+      flatTableHover: palette.slateTint(95),
```
grabkowski, line 133: shouldn't we have a separate object for the flatTable colours? Please keep in mind that we'll have multiple themes for the flat table.

edleeks87: Yeah I was debating whether to do that, I wanted to avoid potentially creating conflicts for @dawidzarzycki's work on `Sort`, I'll just create a separate object for it and try to avoid creating too many conflicts for him 🤞

PR #2899, opened 2020-04-28
fix(date/select): remove portal from date and select components - FE-2673/FE-3062

mkrds:
### Proposed behaviour
DatePicker/SelectList should be positioned absolutely in relation to Date/Select component, meaning it's in the same stacking context.

### Current behaviour
DatePicker/SelectList is rendered in portal using carbon's Portal component meaning that it's in entirely different stacking context which does not allow to display the DatePicker/SelectList under Dialog's header or Form's sticky footer

### Checklist

- [x] Unit tests
- [ ] Theme support

### Testing instructions
Storybook -> Design System -> Form -> In DialogFullScreen with sticky footer


damianOfKyan (dismissed):

edleeks87 (dismissed):

grabkowski (changes requested):
by placing the popup inside the input component we'll break the functionality of these components in every container in which the overflow attribute is set, e.g. drawer or accordion. Depending on the value of the overflow we'll have to either scroll to see that popup or it will be partially hidden.

PR #2902, opened 2020-04-29
fix(search): fix height, width and background style bugs - FE 2596 

DipperTheDan:
### Proposed behaviour
- The height of the page remains the same as the border of the textbox changes. 
- Background of the component is transparent until focus/active state when then textbox background-color is white.
- Width has a prop to either be 100% or a defined pixel width.

Search-fixed

### Current behaviour
- Focus the textbox, hover over, the height of the page adjusts as the border of the textbox changes. 
- Background is permanently white.
- Width is fixed to 375px.

Search-broken

### Checklist

- [x] Unit tests
<del>- [ ] Theme support

### Additional context
N/A

### Testing instructions

- Open Storybook
- Navigate to `Design Systems` folder
- Navigate to the `Search` component
- Navigate to `Docs` tab
- Clicking on any of the demo components should not alter the height of the demo container


src/__experimental__/components/search/search.spec.js
```diff
@@ -30,27 +30,44 @@ describe('Search', () => {
       }, renderWrapper({ value: '' }, mount));
     });
 
+    it('applies the default width when the user does not specify a width', () => {
```
edleeks87, line 33: Neither of these tests asserts anything about the width? Something like below might work

```
assertStyleMatch({
        borderBottom: '2px solid #CCD6DA',
        display: 'inline-flex',
        fontSize: '14px',
        fontWeight: '700',
       width: '100%'
      }, renderWrapper({ value: '' }, mount));
```

src/__experimental__/components/search/search.spec.js
```diff
@@ -30,27 +30,44 @@ describe('Search', () => {
...
+      assertStyleMatch({
+        borderBottom: '2px solid #CCD6DA',
+        display: 'inline-flex',
+        fontSize: '14px',
+        fontWeight: '700'
+      }, renderWrapper({ value: '', searchWidth: null }, mount));
+    });
+
+    it('applies the correct width specified by the user', () => {
```
edleeks87, line 42: assertStyleMatch({
        borderBottom: '2px solid #CCD6DA',
        display: 'inline-flex',
        fontSize: '14px',
        fontWeight: '700',
        width: '400px'
      }, renderWrapper({ value: '', searchWidth: 400 }, mount));

src/__experimental__/components/search/search.spec.js
```diff
@@ -30,27 +30,44 @@ describe('Search', () => {
...
       wrapper = renderWrapper({ value: '' }, mount);
       const input = wrapper.find('input');
       input.simulate('focus');
       assertStyleMatch({
-        borderBottom: '2px solid transparent'
+        borderBottom: '2px solid #CCD6DA'
       }, wrapper);
     });
 
     it('matches the expected styles when the search is active', () => {
```
edleeks87, line 60:
```suggestion
    it('matches the expected styles when there is active and has a value', () => {
```

src/__experimental__/components/search/search.spec.js
DipperTheDan: Done.

src/__experimental__/components/search/search.spec.js
DipperTheDan: Done.

src/__experimental__/components/search/search.spec.js
DipperTheDan: Done. This has been changed to `'matches` the expected styles when the search is active and has a value'`

src/__experimental__/components/search/search.style.js
```diff
@@ -6,106 +6,104 @@ import { baseTheme } from '../../../style/themes';
...
-    color: #668592;
-    width: 20px;
-    height: 20px;
-    cursor: pointer;
-    :hover {
-      color: rgba(0, 0, 0, 0.65);
-    };
-  }
+      ${StyledButton} { 
+        background-color: #255BC7;
```
damianOfKyan, line 61: this should be theme color instead of hardcoded hex value, i think the  theme.colors.search.button  is what you looking for.

damianOfKyan: In fact please can you make sure that any hardcoded value is present in theme and use that instead?

DipperTheDan: I can indeed.

DipperTheDan: Done.

louisescott:
@harpalsingh  @DipperTheDan  can you confirm what color the background should be once the textbox is in focus?  The ticket states white but the ux review/fix is still transparent. 
cc @jamime

PR #2911, opened 2020-05-05
chore: make styled-component a peer dependency

jamime:
### Proposed behaviour
It's not possible to use two instances of the same version of `styled-components` (it removes all styles). Moving `styled-components` to a peer-dependency resolves this.

### Current behaviour
We install a specific version of `styled-components` as a dependency.

- Users can not use the same version of `styled-components` in their project.
- It's not possible to have two versions of `carbon-react` on the page at the same time because of `styled-components`. This may be required when adopting micro frontends.

### Checklist

<del>- [ ] Carbon implementation and Design System documentation are congruent
<del>- [ ] Release notes (Conventional Commits) 
<del>- [ ] Unit tests
<del>- [ ] Storybook added or updated
<del>- [ ] Theme support

### Additional context
:warning: BREAKING CHANGE: styled-components is now a peer dependency.

To upgrade you should install styled-components@^4.4.1 in your project.

### Testing instructions

Check that the codesandbox still works, the link will be in the comment below by the codesandbox bot.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -19,8 +19,13 @@ Carbon is available as an [npm package](https://www.npmjs.com/package/carbon-rea
 npm install carbon-react
 ```
 
-
 ### Peer Dependencies
+You will need to install the following dependencies in your project, these are peer-dependencies of carbon-react and are required.
+
+```sh
+npm install react@^16.12.0 react-dom@^16.12.0 i18n-js@^3.0.0 styled-components@^4.4.1
+```
+
 #### Fonts
 ##### CDN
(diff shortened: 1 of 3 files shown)


jamime:
> please could you reflect this change in the [codesandbox template](https://codesandbox.io/embed/carbon-quickstart-xi5jc) and [getting started guide](https://github.com/Sage/carbon/blob/master/docs/getting-started.md)

Will do, currently just a draft so I can use the codesandbox package in another project.

jamime:
Looks like the codesandbox updated the old template, here is the new one working.

https://codesandbox.io/s/infallible-sara-1yu9l

jamime:
Removed pending QA. There is no way to QA this apart from in a separate project.

PR #2913, opened 2020-05-06
feat: add new SimpleSelect Component (FE-2646)

grabkowski:
### Proposed behaviour
Add new Simple Select and Filterable Select components that with other components like SelectAsync and MultiSelect will replace the Experimental Select in the future

Screenshot 2020-05-06 at 14 24 33

FilterableSelect:

Screenshot 2020-06-18 at 15 32 09

### Checklist

- [x] Unit tests
- [x] Theme support

### Testing instructions
1. run npm start locally.
2. open http://localhost:9001/?path=/docs/design-system-simple-select--basic
3. open http://localhost:9001/?path=/docs/design-system-select-filterable--basic

### Codesandbox
These examples will be forked, please see the comment by `codesandbox[bot]` below for the latest version.
https://codesandbox.io/s/filterable-select-review-f4nox
https://codesandbox.io/s/select-review-ofjs3

### DIFF
--- a/src/components/select/filterable-select/filterable-select.component.js
+++ b/src/components/select/filterable-select/filterable-select.component.js
@@ -1,0 +1,326 @@
+import React, {
+  useState,
+  useEffect,
+  useRef,
+  useCallback
+} from 'react';
+import PropTypes from 'prop-types';
+import invariant from 'invariant';
+import SelectTextbox, { formInputPropTypes } from '../select-textbox/select-textbox.component';
+import guid from '../../../utils/helpers/guid';
+import withFilter from '../utils/with-filter.hoc';
+import SelectList from '../select-list/select-list.component';
+
+const FilterableSelectList = withFilter(SelectList);
+
+const FilterableSelect = React.forwardRef(({
+  value,
+  defaultValue,
+  id,
+  name,
+  children,
+  onOpen,
+  onChange,
+  onClick,
+  onKeyDown,
+  noResultsMessage,
+  ...textboxProps
+}, inputRef) => {
+  const selectListId = useRef(guid());
+  const labelId = useRef(guid());
+  const containerRef = useRef();
+  const listboxRef = useRef();
+  const isControlled = useRef(value !== undefined);
+  const [textboxRef, setTextboxRef] = useState();
+  const [isOpen, setOpenState] = useState(false);
+  const [textValue, setTextValue] = useState('');
+  const [selectedValue, setSelectedValue] = useState('');
+  const [highlightedValue, setHighlightedValue] = useState('');
+  const [filterText, setFilterText] = useState('');
+
+  const setOpen = useCallback((newValue) => {
+    setOpenState((isAlreadyOpen) => {
+      if (!isAlreadyOpen && newValue && onOpen) {
+        onOpen();
+      }
+
+      return newValue;
+    });
+  }, [onOpen]);
+
+  const createCustomEvent = useCallback((newValue) => {
+    const customEvent = {
+      target: {
+        ...(name && { name }),
+        ...(id && { id }),
+        value: newValue
+      }
+    };
+
+    return customEvent;
+  }, [name, id]);
+
+  const updateValues = useCallback((newFilterText, isDeleteEvent) => {
+    setFilterText(newFilterText);
+
+    setSelectedValue((previousValue) => {
+      const match = findElementWithMatchingText(newFilterText, children);
+
+      if (!match || isDeleteEvent) {
+        setTextValue(newFilterText);
+
+        return previousValue;
+      }
+
+      if (onChange) {
+        onChange(createCustomEvent(match.props.value));
+      }
+
+      if (match.props.text.toLowerCase().startsWith(newFilterText.toLowerCase())) {
+        setTextValue(match.props.text);
+      } else {
+        setTextValue(newFilterText);
+      }
+
+      if (isControlled.current) {
+        return previousValue;
+      }
+
+      setHighlightedValue(match.props.value);
+
+      return match.props.value;
+    });
+  }, [children, createCustomEvent, onChange]);
+
+  const setMatchingText = useCallback((newValue) => {
+    const matchingOption = React.Children.toArray(children).find(option => (option.props.value === newValue));
+
+    if (matchingOption) {
+      setTextValue(matchingOption.props.text);
+      setFilterText(matchingOption.props.text);
+    }
+  }, [children]);
+
+  const handleTextboxChange = useCallback((event) => {
+    const newValue = event.target.value;
+    const isDeleteEvent = event.nativeEvent.inputType === 'deleteContentBackward'
+      || event.nativeEvent.inputType === 'deleteContentForward'
+      || event.nativeEvent.inputType === 'delete';
+
+    updateValues(newValue, isDeleteEvent);
+    setOpen(true);
+  }, [setOpen, updateValues]);
+
+  const fillLastFilterCharacter = useCallback((key) => {
+    setFilterText((previousFilterText) => {
+      if (previousFilterText.length === textValue.length - 1 && key === textValue.slice(-1)) {
+        return textValue;
+      }
+
+      return previousFilterText;
+    });
+  }, [textValue]);
+
+  const handleTextboxKeydown = useCallback(((event) => {
+    const { key } = event;
+
+    if (onKeyDown) {
+      onKeyDown(event);
+    }
+
+    if (!event.defaultPrevented && isNavigationKey(key)) {
+      event.preventDefault();
+      setOpen(true);
+    }
+
+    fillLastFilterCharacter(key);
+  }), [fillLastFilterCharacter, onKeyDown, setOpen]);
+
+  const handleGlobalClick = useCallback((event) => {
+    const notInContainer = containerRef.current && !containerRef.current.contains(event.target);
+    const notInList = listboxRef.current && !listboxRef.current.contains(event.target);
+
+    if (notInContainer && notInList) {
+      setMatchingText(selectedValue);
+      setOpen(false);
+    }
+  }, [setMatchingText, setOpen, selectedValue]);
+
+  useEffect(() => {
+    const newValue = value || defaultValue;
+    const modeSwitchedMessage = 'Input elements should not switch from uncontrolled to controlled (or vice versa). '
+    + 'Decide between using a controlled or uncontrolled input element for the lifetime of the component';
+    const onChageMissingMessage = 'onChange prop required when using a controlled input element';
+
+    invariant(isControlled.current === (value !== undefined), modeSwitchedMessage);
+    invariant(!isControlled.current || (isControlled.current && onChange), onChageMissingMessage);
+
+    setSelectedValue(newValue);
+    setHighlightedValue(newValue);
+  }, [value, defaultValue, onChange]);
+
+  useEffect(() => {
+    setMatchingText(value || defaultValue);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    const clickEvent = 'click';
+
+    document.addEventListener(clickEvent, handleGlobalClick);
+
+    return function cleanup() {
+      document.removeEventListener(clickEvent, handleGlobalClick);
+    };
+  }, [handleGlobalClick]);
+
+  useEffect(() => {
+    const textStartsWithFilter = textValue.toLowerCase().startsWith(filterText.toLowerCase());
+
+    if (textboxRef && textValue.length > filterText.length && textStartsWithFilter) {
+      textboxRef.selectionStart = filterText.length;
+    }
+  }, [textValue, filterText, textboxRef]);
+
+  function handleTextboxClick(event) {
+    if (onClick) {
+      onClick(event);
+    }
+  }
+
+  function handleDropdownIconClick(event) {
+    if (onClick) {
+      onClick(event);
+    }
+
+    setOpenState((isAlreadyOpen) => {
+      if (isAlreadyOpen) {
+        return false;
+      }
+
+      if (onOpen) {
+        onOpen();
+      }
+
+      return true;
+    });
+  }
+
+  const onSelectOption = useCallback((optionData) => {
+    const { text, value: newValue, selectionType } = optionData;
+
+    if (!isControlled.current) {
+      setSelectedValue(newValue);
+    }
+
+    setTextValue(text);
+
+    if (onChange) {
+      onChange(createCustomEvent(newValue));
+    }
+
+    setHighlightedValue(newValue);
+
+    if (selectionType !== 'navigationKey') {
+      setOpen(false);
+      setFilterText(text);
+      textboxRef.focus();
+      textboxRef.select();
+    }
+  }, [createCustomEvent, onChange, setOpen, textboxRef]);
+
+  const onSelectListClose = useCallback(() => {
+    setOpen(false);
+    setMatchingText(selectedValue);
+  }, [selectedValue, setMatchingText, setOpen]);
+
+  function findElementWithMatchingText(textToMatch, list) {
+    return list.find((child) => {
+      const { text } = child.props;
+
+      return text && text.toLowerCase().indexOf(textToMatch.toLowerCase()) !== -1;
+    });
+  }
+
+  function assignInput(input) {
+    setTextboxRef(input.current);
+
+    if (inputRef) {
+      inputRef.current = input.current;
+    }
+  }
+
+  function getTextboxProps() {
+    return {
+      id,
+      name,
+      inputRef: assignInput,
+      selectedValue,
+      formattedValue: textValue,
+      onClick: handleTextboxClick,
+      iconOnClick: handleDropdownIconClick,
+      onKeyDown: handleTextboxKeydown,
+      onChange: handleTextboxChange,
+      ...textboxProps
+    };
+  }
+
+  function isNavigationKey(key) {
+    return key === 'ArrowDown' || key === 'ArrowUp'
+    || key === 'Home' || key === 'End';
+  }
+
+  return (
+    <div
+      data-component='filterable-select'
+      aria-expanded={ isOpen }
+      aria-haspopup='listbox'
+      ref={ containerRef }
+    >
+      <SelectTextbox
+        aria-controls={ isOpen ? selectListId.current : '' }
+        type='text'
+        labelId={ labelId.current }
+        { ...getTextboxProps() }
+      />
+      { isOpen && (
+        <FilterableSelectList
+          ref={ listboxRef }
+          id={ selectListId.current }
+          labelId={ labelId.current }
+          role='listbox'
+          anchorElement={ textboxRef.parentElement }
+          onSelect={ onSelectOption }
+          onSelectListClose={ onSelectListClose }
+          filterText={ filterText }
+          highlightedValue={ highlightedValue }
+          noResultsMessage={ noResultsMessage }
+        >
+          { children }
+        </FilterableSelectList>
+      ) }
+    </div>
+  );
+});
+
+FilterableSelect.propTypes = {
+  ...formInputPropTypes,
+  /** The selected value(s), when the component is operating in controlled mode */
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object
+  ]),
+  /** The default selected value(s), when the component is operating in uncontrolled mode */
+  defaultValue: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.object
+  ]),
+  /** Child components (such as Option) for the SelectList */
+  children: PropTypes.node.isRequired,
+  /** A custom callback for when the dropdown menu opens */
+  onOpen: PropTypes.func,
+  /** A custom message to be displayed when any option does not match the filter text */
+  noResultsMessage: PropTypes.string
+};
+
+export default FilterableSelect;
--- a/src/components/select/filterable-select/filterable-select.stories.mdx
+++ b/src/components/select/filterable-select/filterable-select.stories.mdx
@@ -1,0 +1,143 @@
+import { Meta, Props, Preview, Story } from '@storybook/addon-docs/blocks';
+import { action } from '@storybook/addon-actions';
+import { useState } from 'react';
+import { FilterableSelect, Option } from '../';
+import SelectTextbox from '../select-textbox/select-textbox.component';
+
+<Meta title="Design System/Select/Filterable" />
+
+# Filterable Select
+
+Select one of available options from the drop-down menu using filter
+
+## Overview 
+
+Filterable Select is a Carbon styled implementation of WAI-ARIA Combobox with Inline Autocomplete.
+
+## Guidance, hints and tips 
+
+Always insert `Option` Components inside the `FilterableSelect`, analogous to the `<select>` and `<option>` HTML Elements.
+
+If you type printable characters in the Textbox,
+you can filter through the existing options leaving only those that match the text you typed.
+
+```javascript
+import { FilterableSelect, Option } from "carbon-react/lib/components/select";
+```
+
+### Basic usage:
+
+<Preview>
+  <Story name="basic" parameters={{ info: { disable: true }}} >
+    <FilterableSelect
+      name='simple'
+      id='simple'
+      label='label'
+      labelInline
+      onOpen={ action('onOpen') }
+      onChange={ action('onChange') }
+      onClick={ action('onClick') }
+      onFocus={ action('onFocus') }
+      onBlur={ action('onBlur') }
+      onKeyDown={ action('onKeyDown') }
+    >
+      <Option text='Amber' value='1' />
+      <Option text='Black' value='2' />
+      <Option text='Blue' value='3' />
+      <Option text='Brown' value='4' />
+      <Option text='Green' value='5' />
+      <Option text='Orange' value='6' />
+      <Option text='Pink' value='7' />
+      <Option text='Purple' value='8' />
+      <Option text='Red' value='9' />
+      <Option text='White' value='10' />
+      <Option text='Yellow' value='11' />
+    </FilterableSelect>
+  </Story>
+</Preview>
+
+### Controlled Usage:
+
+<Preview>
+  <Story name="controlled" parameters={{ info: { disable: true }}} >
+    {() => {
+      const [value, setValue] = useState('');
+      function onChangeHandler(event) {
+        setValue(event.target.value);
+        action('value set');
+      };
+      return (
+        <FilterableSelect
+          id='controlled'
+          name='controlled'
+          value={ value }
+          onChange={ onChangeHandler }
+        >
+          <Option text='Amber' value='1' />
+          <Option text='Black' value='2' />
+          <Option text='Blue' value='3' />
+          <Option text='Brown' value='4' />
+          <Option text='Green' value='5' />
+          <Option text='Orange' value='6' />
+          <Option text='Pink' value='7' />
+          <Option text='Purple' value='8' />
+          <Option text='Red' value='9' />
+          <Option text='White' value='10' />
+          <Option text='Yellow' value='11' />
+        </FilterableSelect>
+      );
+    }}
+  </Story>
+</Preview>
+
+### Disabled:
+
+<Preview>
+  <Story name="disabled" parameters={{ info: { disable: true }}} >
+    <FilterableSelect name='disabled' id='disabled' defaultValue='3' disabled>
+      <Option text='Amber' value='1' />
+      <Option text='Black' value='2' />
+      <Option text='Blue' value='3' />
+      <Option text='Brown' value='4' />
+      <Option text='Green' value='5' />
+      <Option text='Orange' value='6' />
+      <Option text='Pink' value='7' />
+      <Option text='Purple' value='8' />
+      <Option text='Red' value='9' />
+      <Option text='White' value='10' />
+      <Option text='Yellow' value='11' />
+    </FilterableSelect>
+  </Story>
+</Preview>
+
+### Read Only:
+
+<Preview>
+  <Story name="readonly" parameters={{ info: { disable: true }}} >
+    <FilterableSelect name='readonly' id='readonly' defaultValue='4' readOnly>
+      <Option text='Amber' value='1' />
+      <Option text='Black' value='2' />
+      <Option text='Blue' value='3' />
+      <Option text='Brown' value='4' />
+      <Option text='Green' value='5' />
+      <Option text='Orange' value='6' />
+      <Option text='Pink' value='7' />
+      <Option text='Purple' value='8' />
+      <Option text='Red' value='9' />
+      <Option text='White' value='10' />
+      <Option text='Yellow' value='11' />
+    </FilterableSelect>
+  </Story>
+</Preview>
+
+## Props:
+
+<Props of={ FilterableSelect } />
+
+### Props derived from the Textbox Component
+
+<Props of={ SelectTextbox } />
+
+### Props of the Option Component
+
+<Props of={ Option } />
--- a/src/components/select/filterable-select/index.d.ts
+++ b/src/components/select/filterable-select/index.d.ts
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import Option from '../option';
+
+export interface FilterableSelectProps {
+  /** Id attribute of the input element */
+  id?: string;
+  /** Name attribute of the input element */
+  name?: string;
+  /** If true the Component will be read-only */
+  readOnly?: boolean;
+  /** If true the Component will be disabled? */
+  disabled?: boolean;
+  /** If true the Component will be focused when rendered */
+  autoFocus?: boolean;
+  /** Label */
+  label?: string;
+  /** Text applied to label help tooltip */
+  labelHelp?: string;
+  /** When true, label is placed in line with an input */
+  labelInline?: boolean;
+  /** Width of a label in percentage. Works only when labelInline is true */
+  labelWidth?: number;
+  /** Width of an input in percentage. Works only when labelInline is true */
+  inputWidth?: number;
+  /** Size of an input */
+  size?: 'small' | 'medium' | 'large';
+  /** Placeholder string to be displayed in input */
+  placeholder?: string;
+  /** The selected value(s), when the component is operating in controlled mode */
+  value?: string | object;
+  /** The default selected value(s), when the component is operating in uncontrolled mode */
+  defaultValue?: string | object;
+  /** Child components (such as Option) for the SelectList */
+  children: Array<typeof Option>;
+  /** If true, prevents opening the menu on focus */
+  preventFocusAutoOpen?: boolean;
+  /** If true the component input has no border and is transparent */
+  transparent?: boolean;
+  /** If true the Component has filter functionality (filtering starts when two or more characters are typed) */
+  filterable?: boolean;
+  /** A custom callback for when the dropdown menu opens */
+  onOpen?: () => void;
+  /** A custom callback for when changes occur */
+  onChange?: () => void;
+  /** Callback function for when the Select Textbox is clicked. */
+  onClick?: () => void;
+  /** Callback function for when the Select Textbox is focused. */
+  onFocus?: () => void;
+  /** Callback function for when the Select Textbox loses it's focus. */
+  onBlur?: () => void;
+}
+
+declare const FilterableSelect: React.ComponentType<FilterableSelectProps>;
+
+export default FilterableSelect;
--- a/src/components/select/index.js
+++ b/src/components/select/index.js
@@ -1,0 +1,2 @@
+export { default as Option } from './option/option.component';
+export { default as Select } from './simple-select/simple-select.component';
--- a/src/components/select/index.js
+++ b/src/components/select/index.js
@@ -1,2 +1,3 @@
 export { default as Option } from './option/option.component';
 export { default as Select } from './simple-select/simple-select.component';
+export { default as FilterableSelect } from './filterable-select/filterable-select.component';
--- a/src/components/select/option/index.d.ts
+++ b/src/components/select/option/index.d.ts
@@ -1,0 +1,14 @@
+import * as React from 'react';
+
+export interface OptionProps {
+  /** The option's visible text, displayed within <Textbox> of <Select>, and used for filtering */
+  text: string;
+  /** Optional: alternative rendered content, displayed within <SelectList> of <Select> (eg: an icon, an image, etc) */
+  children?: React.FunctionComponent | React.ComponentClass;
+  /** The option's invisible internal value */
+  value: string | object;
+}
+
+declare const Option: React.ComponentType<OptionProps>;
+
+export default Option;
--- a/src/components/select/option/option.component.js
+++ b/src/components/select/option/option.component.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import StyledOption from './option.style';
+
+const Option = ({
+  text,
+  children,
+  onSelect,
+  value,
+  isHighlighted
+}) => {
+  function handleClick() {
+    onSelect({ text, value });
+  }
+
+  return (
+    <StyledOption
+      aria-selected={ isHighlighted }
+      data-component='option'
+      onClick={ handleClick }
+      isHighlighted={ isHighlighted }
+      role='option'
+    >
+      { children || text }
+    </StyledOption>
+  );
+};
+
+Option.propTypes = {
+  /** The option's visible text, displayed within Textbox of Select, and used for filtering */
+  text: PropTypes.string.isRequired,
+  /** Optional: alternative rendered content, displayed within SelectList of Select (eg: an icon, an image, etc) */
+  children: PropTypes.node,
+  /** The option's invisible internal value */
+  value: PropTypes.oneOfType([PropTypes.object, PropTypes.string]).isRequired,
+  /**
+   * @private
+   * @ignore
+   * Callback to return value when the element is selected (prop added by the SelectList component) */
+  onSelect: PropTypes.func,
+  /**
+   * @private
+   * @ignore
+   * True if the option is highlighted (prop added by the SelectList component) */
+  isHighlighted: PropTypes.bool
+};
+
+export default Option;
--- a/src/components/select/option/option.spec.js
+++ b/src/components/select/option/option.spec.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import { shallow, mount } from 'enzyme';
+import Option from './option.component';
+import { baseTheme } from '../../../style/themes';
+
+describe('Option', () => {
+  it('renders properly', () => {
+    const props = { value: '1', text: 'foo', children: 'bar' };
+    expect(renderOption(props, TestRenderer.create)).toMatchSnapshot();
+  });
+
+  describe('when no children are provided', () => {
+    it('renders with the text prop as children', () => {
+      const props = { value: '1', text: 'foo' };
+      expect(renderOption(props, TestRenderer.create)).toMatchSnapshot();
+    });
+  });
+
+  describe('when isHighlighted prop is set', () => {
+    it('then it should have expected background', () => {
+      const props = { value: '1', text: 'foo', isHighlighted: true };
+      expect(renderOption(props, mount)).toHaveStyleRule('background-color', baseTheme.select.selected);
+    });
+  });
+
+  describe('when the element is hovered over', () => {
+    it('then it should have expected background', () => {
+      const props = { value: '1', text: 'foo' };
+      expect(renderOption(props, mount))
+        .toHaveStyleRule('background-color', baseTheme.select.selected, { modifier: ':hover' });
+    });
+  });
+
+  describe('when the element is clicked', () => {
+    it('then onSelect should be called with text and value', () => {
+      const onSelect = jest.fn();
+      const props = { value: '1', text: 'foo', onSelect };
+      const wrapper = renderOption(props, mount);
+      wrapper.simulate('click');
+
+      expect(onSelect).toHaveBeenCalledWith({ text: props.text, value: props.value });
+    });
+  });
+});
+
+function renderOption(props, renderer = shallow) {
+  return renderer(
+    <Option { ...props } />
+  );
+}
--- a/src/components/select/select-list/select-list.component.js
+++ b/src/components/select/select-list/select-list.component.js
@@ -1,0 +1,176 @@
+import React, {
+  useEffect,
+  useState,
+  useCallback
+} from 'react';
+import PropTypes from 'prop-types';
+import StyledSelectList from './select-list.style';
+import updateListScrollTop from './update-list-scroll';
+import getNextChildByText from '../utils/get-next-child-by-text';
+import Portal from '../../portal/portal';
+import { getNextIndexByKey } from '../utils/get-next-child-by-key';
+
+const overhang = 4;
+
+const SelectList = React.forwardRef(({
+  id,
+  labelId,
+  children,
+  onSelect,
+  onSelectListClose,
+  filterText,
+  anchorElement,
+  highlightedValue
+}, listRef) => {
+  const [currentOptionsListIndex, setCurrentOptionsListIndex] = useState(-1);
+
+  const highlightItem = useCallback((newIndex, optionList) => {
+    const { text, value } = optionList[newIndex].props;
+
+    setCurrentOptionsListIndex(newIndex);
+    updateListScrollTop(newIndex, listRef.current);
+    onSelect({ text, value, selectionType: 'navigationKey' });
+  }, [listRef, onSelect]);
+
+  const handleGlobalKeydown = useCallback((event) => {
+    const { key } = event;
+    const optionList = React.Children.toArray(children);
+    const lastIndex = optionList.length - 1;
+
+    if (key === 'Tab' || key === 'Escape') {
+      onSelectListClose();
+    } else if (key === 'Enter') {
+      const currentOption = optionList[currentOptionsListIndex];
+
+      if (!currentOption) {
+        onSelectListClose();
+
+        return;
+      }
+
+      const { text, value } = currentOption.props;
+
+      onSelect({ text, value, selectionType: 'enterKey' });
+    } else if (isNavigationKey(key)) {
+      const nextIndex = getNextIndexByKey(key, currentOptionsListIndex, lastIndex);
+
+      highlightItem(nextIndex, optionList);
+    }
+  }, [currentOptionsListIndex, children, onSelectListClose, onSelect, highlightItem]);
+
+  const repositionList = useCallback(() => {
+    if (anchorElement) {
+      const inputBoundingRect = anchorElement.getBoundingClientRect();
+      const top = `${window.pageYOffset + inputBoundingRect.top + inputBoundingRect.height}px`;
+      const width = `${inputBoundingRect.width + 2 * overhang}px`;
+      const left = `${window.pageXOffset + inputBoundingRect.left - overhang}px`;
+
+      listRef.current.setAttribute('style', `top: ${top}; width: ${width}; left: ${left}`);
+    }
+  }, [anchorElement, listRef]);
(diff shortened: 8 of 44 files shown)


src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -0,0 +1,144 @@
+import { Meta, Props, Preview, Story } from '@storybook/addon-docs/blocks';
+import SimpleSelect from '.';
+import Option from '../select-option';
+
+<Meta title="Design System/Simple Select" />
+
+# Simple Select
+
+
```
DipperTheDan, line 9: I know we haven't got a template in place yet but do you think it would be worth adding a `Quick Start` section here with an example of how to import the component in to their project?

grabkowski: good point, looks like I've forgot about that :)

grabkowski: quick start section added

DipperTheDan: Excellent - thank you. I've also noticed you've added a couple of other helpful sections to the MDX file too. Good thinking.

src/components/select/select-list/select-list.spec.js
```diff
@@ -0,0 +1,141 @@
...
+      const onSelectOption = jest.fn();
+      const wrapper = renderSelectList({ onSelectOption });
+
+      expect(wrapper.find(Option).first().prop('onSelectOption')).toBe(onSelectOption);
+      expect(wrapper.find(Option).first().prop('index')).toBe(0);
+      expect(wrapper.find(Option).first().prop('selectedIndex')).toBe(-1);
+    });
+  });
+
+  describe('when the hasTypeToSearch prop and filterText nas more that two characters', () => {
```
damianOfKyan, line 107:
```suggestion
  describe('when the hasTypeToSearch prop and filterText have more than two characters', () => {
```

src/components/select/select-list/select-list.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    describe('and one of the options text contains the filterText', () => {
+      it('then there should be only one option selected', () => {
+        const wrapper = renderSelectList({ hasTypeToSearch: true, filterText: 'gre' });
+
+        expect(wrapper.find(Option)).toHaveLength(1);
+      });
+    });
+  });
+
+  describe('when the hasTypeToSearch prop and filterText nas more that two characters', () => {
```
damianOfKyan, line 117:
```suggestion
  describe('when the hasTypeToSearch prop and filterText have more than two characters', () => {
```

src/components/select/select-list/select-list.style.js
```diff
@@ -0,0 +1,28 @@
...
+  list-style-type: none;
+  max-height: ${props => `${props.maxHeight}`};
+  margin: 0;
+  overflow-x: hidden;
+  overflow-y: scroll;
+  padding: 0;
+  width: 100%;
+  z-index: 1;
+
+  ${({ styleOverride }) => styleOverride};
```
damianOfKyan, line 20: shouldn't that be removed?

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,153 @@
...
+    if (Events.isTabKey(event) || Events.isEscKey(event)) {
+      onSelectListClose();
+    } else if (Events.isEnterKey(event)) {
+      if (isNoOptionSelected) {
+        return;
+      }
+
+      const { text, value } = optionsList[currentOptionsListIndex].props;
+
+      onSelectOption({ text, value });
```
edleeks87, line 42: This may be undefined if no `onSelectOption` prop has been passed in

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,153 @@
...
+  const [currentOptionsListIndex, setCurrentOptionsListIndex] = useState(-1);
+
+  const handleGlobalKeydown = useCallback((event) => {
+    let newIndex;
+    const isNoOptionSelected = currentOptionsListIndex === -1;
+    const optionsList = React.Children.toArray(children);
+    const lastIndex = optionsList.length - 1;
+
+    if (Events.isTabKey(event) || Events.isEscKey(event)) {
+      onSelectListClose();
```
edleeks87, line 34: This could be undefined if no `onSelectListClose` is passed in

src/components/select/simple-select/index.d.ts
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import Option from '../option';
+
+export interface SimpleSelectProps {
+  /** The selected value(s), when the component is operating in controlled mode */
+  value: string | object;
```
edleeks87, line 6: Should these have `?` as they're not required

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,260 @@
...
+  const handleGlobalClick = useCallback((event) => {
+    if (!Events.composedPath(event).includes(containerRef.current)) {
+      setMatchingText(selectedValue);
+      setOpen(false);
+      setFilterValue('');
+    }
+  }, [setMatchingText, setOpen, selectedValue]);
+
+  useEffect(() => {
+    const newValue = value || defaultValue;
```
edleeks87, line 101: Might be worth adding an invariant for switching between controlled/ uncontrolled like we have for other inputs

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -0,0 +1,161 @@
...
+        <Option text='Purple' value='8' />
+        <Option text='Red' value='9' />
+        <Option text='White' value='10' />
+        <Option text='Yellow' value='11' />
+      </SimpleSelect>
+    </div>
+  </Story>
+</Preview>
+
+### Opens on action:
```
jamime, line 49: Please can we add more detail here.
I think the prop `opensOnFocus` should be `openOnFocus` and it should default to `false` just like the HTML component.

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -0,0 +1,161 @@
...
+    </div>
+  </Story>
+</Preview>
+
+### With type to search:
+
+<Preview>
+  <Story name="typetosearch" parameters={{ info: { disable: true }}} >
+    <div style={{ height: '250px' }}>
+      <SimpleSelect name='typetosearch' id='typetosearch' defaultValue='4' hasTypeToSearch>
```
jamime, line 142: I think the `hasTypeToSearch` should be `filter`. We've not used the `has` prefix on props and I think people will confuse the word `search` when we come to implement the component that does a async search.

src/components/select/simple-select/simple-select.stories.mdx
grabkowski: I agree with the name change, but if that prop defaults to false then the select will be not backward compatible with the old one

src/components/select/simple-select/simple-select.stories.mdx
grabkowski: actually we've already used `has` prefix, e.g.: 'hasSearchButton' in search or `hasAdditionalActions` in form. I could change it to `hasFilter` as it suggests it's a boolean.

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      const { text, value } = optionsList[currentOptionsListIndex].props;
+
+      onSelectOption({ text, value });
+    } else if (Events.isDownKey(event)) {
+      if (currentOptionsListIndex === lastIndex || isNoOptionSelected) {
+        newIndex = 0;
+      } else {
+        newIndex = currentOptionsListIndex + 1;
```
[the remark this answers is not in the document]
grabkowski: you're right I must have refactored that part too many times

src/components/select/select-list/select-list.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    describe("and it's the Enter key", () => {
+      describe('without an highlighted element', () => {
+        it('then the onSelectOption prop should not be called', () => {
+          domNode.dispatchEvent(enterKeyDownEvent);
+          expect(onSelectOption).not.toHaveBeenCalled();
+        });
+      });
+
+      describe('with an highlighted element', () => {
+        it('then the onSelectOption prop should be called', () => {
```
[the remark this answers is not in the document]
grabkowski: when I change that to your suggestion, the resulting test will be described:
`when a key is pressed and it's the Enter key with an highlighted element should call onSelectOption`
vs
`when a key is pressed and it's the Enter key with an highlighted element then the onSelectOption prop should be called`

src/components/select/select-list/select-list.style.js
grabkowski: looks like I've missed one

src/components/select/option/index.d.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export interface OptionProps {
+  /** The option's visible text, displayed within <Textbox> of <Select>, and used for filtering */
+  text: string;
+  /** Optional: alternative rendered content, displayed within <SelectList> of <Select> (eg: an icon, an image, etc) */
+  children?: React.FunctionComponent | React.ComponentClass;
+  /** The option's invisible internal value */
+  value: string | object;
+  /** Callback to return value when the element is selected */
+  onSelectOption: () => void;
```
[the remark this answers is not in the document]
grabkowski: I'll remove it from the d.ts file ans it's a prop added by the parent component and should not be revealed to the user

src/components/select/option/option.component.js
```diff
@@ -0,0 +1,50 @@
...
+  index,
+  selectedIndex
+}) => {
+  const [isHighlighted, setIsHighlighted] = useState();
+  useEffect(() => {
+    setIsHighlighted(index === selectedIndex);
+  }, [index, selectedIndex]);
+
+  function handleClick() {
+    onSelectOption({ text, value });
```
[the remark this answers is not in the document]
grabkowski: it's a prop that is added by the parent component. The component is not designed to work outside of the SelectList, so we could assume that this prop will be there.

src/components/select/select-list/select-list.component.js
grabkowski: I'll set the `onSelectOption` prop as `isRequired`, this component is only presentational and is not designed to work outside of select components

edleeks87: Yeah I think that is the better approach

src/components/select/option/option.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+  describe('when no children are provided', () => {
+    it('renders with the text prop as children', () => {
+      const props = { value: '1', text: 'foo' };
+      expect(renderOption(props, TestRenderer.create)).toMatchSnapshot();
+    });
+  });
+
+  describe('when isHighlighted prop is set', () => {
+    it('then it should have proper background', () => {
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/option/option.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      const props = { value: '1', text: 'foo' };
+      expect(renderOption(props, mount)).toHaveStyleRule('background-color', baseTheme.select.selected);
+    });
+  });
+
+  describe('when the element is hovered over', () => {
+    it('then it should have proper background', () => {
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/select-list/highlight-part-of-text.spec.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import highlightPartOfText from './highlight-part-of-text';
+import MatchingText from './matching-text.style';
+
+describe('highlightPartOfText', () => {
+  describe('when the text contains no matching parts', () => {
+    it('then the returned text should be not changed', () => {
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/select-list/highlight-part-of-text.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      const text = 'abcdefghi';
+      const partToBeHighlighted = 'xyz';
+
+      expect(highlightPartOfText(text, partToBeHighlighted)).toBe(text);
+    });
+  });
+
+  describe('when the partToBeHighlighted is an empty string', () => {
+    it('then the returned text should be not changed', () => {
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/select-list/highlight-part-of-text.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      const text = 'abcdefghi';
+      const partToBeHighlighted = '';
+
+      expect(highlightPartOfText(text, partToBeHighlighted)).toBe(text);
+    });
+  });
+
+  describe('when the text contains matching part', () => {
+    it('the proper part of text should be highlighted', () => {
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,153 @@
...
+      return (<StyledOption>{ noResultsText }</StyledOption>);
+    }
+
+    return getChildrenWithListProps(filteredElements);
+  }
+
+  function getChildrenWithListProps(filteredElements) {
+    return React.Children.map(filteredElements, (child, index) => {
+      const newProps = {
+        onSelectOption,
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/select-list/select-list.spec.js
grabkowski: changed

src/components/select/select-list/select-list.spec.js
grabkowski: changed

src/components/select/simple-select/simple-select.component.js
grabkowski: added

src/components/select/simple-select/simple-select.stories.mdx
grabkowski: changed to `preventFocusAutoOpen` to keep backward compatibility

src/components/select/simple-select/index.d.ts
grabkowski: fixed

src/components/select/simple-select/simple-select.stories.mdx
grabkowski: changed to `filterable` to keep backward compatibility

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,160 @@
...
+  /** The ID for the parent <div> */
+  id: PropTypes.string,
+  /** Child components (such as <Option>) for the <ScrollableList> */
+  children: PropTypes.node,
+  /** DOM element to position the dropdown menu list relative to */
+  anchorElement: PropTypes.object,
+  /** A callback for when a child is selected */
+  onSelect: PropTypes.func.isRequired,
+  /** A callback for when the list should be closed */
+  onSelectListClose: PropTypes.func,
```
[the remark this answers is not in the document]
grabkowski: you're right, changed to `isRequired`

src/components/select/simple-select/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  /** [Form Field] Id attribute of the input element */
+  id?: string;
+  /** [Form Field] Name attribute of the input element */
+  name?: string;
+  /** [Form Field] If true the Component will be read-only */
+  readOnly?: boolean;
+  /** [Form Field] If true the Component will be disabled? */
+  disabled?: boolean;
+  /** [Form Field] If true the Component will be focused when rendered */
+  autoFocus: boolean;
```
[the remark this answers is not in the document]
grabkowski: added

src/components/select/simple-select/simple-select.spec.js
```diff
@@ -74,18 +74,18 @@ describe('SimpleSelect', () => {
     });
   });
 
-  describe('when the preventFocusAutoOpen prop is not set', () => {
+  describe('when the openOnFocus prop not set', () => {
```
edleeks87, line 77: Think the description needs updating
```suggestion
  describe('when the openOnFocus prop is set', () => {
```

src/components/select/simple-select/simple-select.stories.mdx
jamime: I meant that `has` is only used within our internal components. `filterable` is fine.

src/components/select/simple-select/simple-select.stories.mdx
jamime: This was kept as `openOnFocus` as it better describes the new default behaviour. `preventFocusAutoOpen` ends up being a double negative when you want it to open on focus `preventFocusAutoOpen=false`.

src/components/select/select-list/select-list.spec.js
```diff
@@ -0,0 +1,246 @@
...
+        });
+        expect(wrapper.update().find(Option).last()).toHaveStyleRule('background-color', baseTheme.select.selected);
+      });
+    });
+
+    // coverage filler
+    it('does nothing if the other key is pressed', () => {
+      act(() => {
+        domNode.dispatchEvent(new KeyboardEvent('keydown', { which: 32, bubbles: true }));
+      });
```
edleeks87, line 151: I realise this test is merely to hit coverage but maybe you could iterate over the `Options` and `expect(option).not.toHaveStyleRule('background-color')`, then at least that way there's is an assertion at the end of it

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,363 @@
...
+      clearTimeout(filterTimer.current);
+    };
+  }, []);
+
+  function handleTextboxClick(event) {
+    if (onClick) {
+      onClick(event);
+    }
+
+    setOpenState((isAlreadyOpen) => {
```
edleeks87, line 195: I think you need to add a guard here `!openOnFocus` or you might toggle the open state inadvertently when `handleTextboxFocus` is called

src/components/select/utils/get-next-child-by-text.spec.js
```diff
@@ -0,0 +1,67 @@
...
+describe('getNextChildByText', () => {
+  let wrapper;
+  let children;
+
+  beforeEach(() => {
+    wrapper = renderList();
+    children = wrapper.find('li').props().children;
+  });
+
+  describe('when there is only one character passed in textToMath', () => {
```
edleeks87, line 14:
```suggestion
  describe('when there is only one character passed in textToMatch', () => {
```

src/components/select/utils/get-next-child-by-text.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    });
+
+    describe('and that character does not match the first letter in text of any element', () => {
+      it('then undefined should be returned', () => {
+        expect(getNextChildByText('x', children)).toBe(undefined);
+      });
+    });
+  });
+
+  describe('when there are multiple characters passed in textToMath', () => {
```
edleeks87, line 28:
```suggestion
  describe('when there are multiple characters passed in textToMatch', () => {
```

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,363 @@
...
+    </StyledSimpleSelect>
+  );
+};
+
+const valuePropType = PropTypes.oneOfType([
+  PropTypes.string,
+  PropTypes.object
+]);
+
+const formInputPropTypes = {
```
edleeks87, line 312: It might be worth extracting this into a separate file and then importing it in if it's going to be done for every `Select` variant

grabkowski: extracted to SelectTextbox

src/components/select/simple-select/simple-select.component.js
grabkowski: good point, added

src/components/select/select-list/select-list.spec.js
grabkowski: added new assertion

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  function onSelectListClose() {
+    setOpen(false);
+    setMatchingText(selectedValue);
+  }
+
+  function findElementWithMatchingText(textToMatch, list) {
+    return list.find((child) => {
+      return child.props.text && child.props.text.toLowerCase().indexOf(textToMatch.toLowerCase()) !== -1;
```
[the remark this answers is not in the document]
grabkowski: changed

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -0,0 +1,293 @@
...
+  function onSelectOption(optionData) {
+    const { text, value: newValue, selectionType } = optionData;
+
+    if (!isControlled.current) {
+      setSelectedValue(newValue);
+      setTextValue(text);
+    }
+
+    if (onChange) {
+      onChange(createCustomEvent(newValue));
```
jamime, line 206: I know we raised an RFC about this, primarily for difficult components that store multiple values.

In this example, is the value the same that is in the DOM? If so we can just pass the ref `onChange(textboxRef.current)`

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -0,0 +1,293 @@
...
+  const listboxRef = useRef();
+  const isControlled = useRef(value !== undefined);
+  const [textboxRef, setTextboxRef] = useState();
+  const [isOpen, setOpenState] = useState(false);
+  const [textValue, setTextValue] = useState('');
+  const [selectedValue, setSelectedValue] = useState('');
+  const [highlightedValue, setHighlightedValue] = useState('');
+  const [filterText, setFilterText] = useState('');
+
+  const assignInput = useCallback((input) => {
```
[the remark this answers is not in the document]
grabkowski: 1. I agree, I'll implement a forward ref.
2. I need to get the parent element of the input, to do it I need it to be set in the state. When you have your method wrapped in `useCallback` hook it will always reference proper htmlElement as it's happening here, so I can't see any issues with that. This is a ref to the main element of the component and I can't see a case when it's replaced with another one during the life of that component.

In their docs they are saying that the ref shouldn't be defined as an inline function, and it's not an inline function so the mentioned caveat does not concerns us here.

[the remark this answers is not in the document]
jamime: 1. ✅ 
2. In the past it's been when a user has tried to access a ref within a callback, e.g. accessing the ref in "onBlur". I haven't been able to reproduce the example, so it looks like we're fine 😄 
We do still however need to handle callback refs, because we're setting the ref manually, we don't get that for free.

src/components/select/filterable-select/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  /** [Form Field] Name attribute of the input element */
+  name?: string;
+  /** [Form Field] If true the Component will be read-only */
+  readOnly?: boolean;
+  /** [Form Field] If true the Component will be disabled? */
+  disabled?: boolean;
+  /** [Form Field] If true the Component will be focused when rendered */
+  autoFocus?: boolean;
+  /** [Form Field] A callback to retrieve the input reference */
+  inputRef?: (inputReference: React.RefObject<HTMLInputElement>) => void;
```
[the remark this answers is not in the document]
grabkowski: looks like I've missed this one

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+        <Option text='White' value='10' />
+        <Option text='Yellow' value='11' />
+      </FilterableSelect>
+    </div>
+  </Story>
+</Preview>
+
+## Props:
+
+<Props of={FilterableSelect} />
```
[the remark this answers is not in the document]
grabkowski: added Textbox derived props and Option props

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+## Guidance, hints and tips 
+
+Always insert `Option` Components inside the `FilterableSelect`, analogous to the `<select>` and `<option>` HTML Elements.
+
+If you type printable characters in the Textbox,
+you can filter through the existing options leaving only those that match the text you typed.
+
+## Quick Start 
+
+To use the `FilterableSelect` component you have to import two components: `<FilterableSelect />` and `<Option />`
```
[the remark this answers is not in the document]
grabkowski: added to both simple and filterable one

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+        <Option text='Pink' value='7' />
+        <Option text='Purple' value='8' />
+        <Option text='Red' value='9' />
+        <Option text='White' value='10' />
+        <Option text='Yellow' value='11' />
+      </FilterableSelect>
+    </div>
+  </Story>
+</Preview>
+
```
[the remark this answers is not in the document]
grabkowski: example included in test folder

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -0,0 +1,134 @@
...
+## Props:
+
+<Props of={ FilterableSelect } />
+
+### Props derived from the Textbox Component
+
+<Props of={ SelectTextbox } exclude={ ['onKeyDown', 'formattedValue', 'selectedValue'] }/>
```
jamime, line 130: Is there a way to actually remove these props from the component? Or is this the only way because we still have the `__experimental__` select component? I'm wary that we could add more props to Textbox and have them surfaced here unintentionally.

I think it would be best if we specifically defined the prop types for the select component and passed the props by key/value instead of spreading the props.

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -0,0 +1,175 @@
...
+
+Simple Select is a Carbon styled equivalent of HTML Select Element.
+
+## Guidance, hints and tips 
+
+Always insert `Option` Components inside the `SimpleSelect`, analogous to the original `<select>` and `<option>` HTML Elements.
+
+## Quick Start 
+
+To use the `SimpleSelect` component you have to import two components: `<SimpleSelect />` and `<Option />`
```
jamime, line 22: We use the term `SimpleSelect` in this document but we changed the public interface to be `Select`. We need to update this document to reflect that change.

e.g. `import {SimpleSelect} from 'carbon-react/lib/components/select'` is actually `import {Select} from 'carbon-react/lib/components/select'`

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,170 @@
...
+        onSelectListClose();
+
+        return;
+      }
+
+      const { text, value } = currentOption.props;
+
+      onSelect({ text, value, selectionType: 'enterKey' });
+    } else if (isNavigationKey(event)) {
+      const nextIndex = getNextIndexByKey(event.which, currentOptionsListIndex, lastIndex);
```
jamime, line 52: `event.which` is deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which

You probably want `event.key` instead.

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,331 @@
...
+    clearTimeout(filterTimer.current);
+
+    filterTimer.current = setTimeout(() => {
+      isTimerCounting.current = false;
+      filterText.current = '';
+    }, 500);
+  }, [selectValueStartingWithText]);
+
+  const handleTextboxKeydown = useCallback(((event) => {
+    const keyCode = event.which;
```
jamime, line 126: `event.which` is deprecated.

src/components/select/filterable-select/filterable.stories.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react';
+import FilterableSelect from '.';
+import Option from '../option';
+
+export default {
+  title: 'Test/Select/FilterableSelect',
```
jamime, line 6: The controlled example should be in the mdx docs.

src/components/select/filterable-select/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './filterable-select.component';
```
jamime, line 1: We're exposing two ways of using filterable-select.

- `import {FilterableSelect} from 'carbon-react/lib/components/select`
- `import FilterableSelect from 'carbon-react/lib/components/select/filterable-select`

We should have only one way to do this.

src/components/select/simple-select/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './simple-select.component';
```
jamime, line 1: We're exposing two ways of using Select.

- `import {Select} from 'carbon-react/lib/components/select`
- `import Select from 'carbon-react/lib/components/select/simple-select`

We should have only one way to do this.

src/components/select/select-list/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './select-list.component';
```
jamime, line 1: If this isn't a component our consumers should be using, we shouldn't have an export for it.

e.g. we don't want to support

- `import SelectList from 'carbon-react/lib/components/select/select-list`

When using it internally we should use the `select-list.component` directly.

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,170 @@
...
+  }, [currentOptionsListIndex, children, onSelectListClose, onSelect, highlightItem]);
+
+  const repositionList = useCallback(() => {
+    if (anchorElement) {
+      const inputBoundingRect = anchorElement.getBoundingClientRect();
+      const top = `${inputBoundingRect.top + inputBoundingRect.height + window.pageYOffset}px`;
+      const width = `${inputBoundingRect.width}px`;
+      const left = `${inputBoundingRect.left}px`;
+
+      listRef.current.setAttribute('style', `top: ${top}; width: ${width}; left: ${left}`);
```
jamime, line 65: I've noticed if you resize the window, the select-list is positioned without respect to scroll position.

image

src/components/select/select-list/select-list.component.js
```diff
@@ -0,0 +1,170 @@
...
+
+  return (
+    <Portal onReposition={ repositionList }>
+      <StyledSelectList
+        id={ id }
+        data-element='select-list'
+        role='listbox'
+        ref={ listRef }
+      >
+        { getChildrenWithListProps() }
```
jamime, line 147: We're currently cloning the children every time we render, do we need to?

src/components/select/select-textbox/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './select-textbox.component';
```
jamime, line 1: I don't think we should be exporting this.

src/components/select/select-textbox/select-textbox.component.js
```diff
@@ -0,0 +1,120 @@
...
+  readOnly,
+  placeholder,
+  onClick,
+  onFocus,
+  onBlur,
+  selectedValue,
+  ...restProps
+}) => {
+  const defaultPlaceholder = I18n.t('select.placeholder', {
+    defaultValue: 'Please Select...'
```
jamime, line 19: This should be a `defaultProp`
```suggestion
    defaultValue: placeholder
```

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,331 @@
...
+  children,
+  transparent,
+  openOnFocus,
+  onOpen,
+  onChange,
+  onClick,
+  onFocus,
+  ...textboxProps
+}, inputRef) => {
+  const selectListId = guid();
```
jamime, line 32: This guid will change on each render.

src/components/select/simple-select/simple-select.stories.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react';
+import SimpleSelect from '.';
+import Option from '../option';
+
+export default {
+  title: 'Test/Select/SimpleSelect',
+  component: SimpleSelect
+};
+
+const Wrapped = () => {
```
jamime, line 10: IMO this should be in MDX.

src/components/select/utils/get-next-child-by-key-code.js
```diff
@@ -0,0 +1,36 @@
...
+  const lastIndex = childrenList.length - 1;
+
+  return childrenList[getNextIndexByKey(keyCode, currentSelectedIndex, lastIndex)];
+}
+
+export function getNextIndexByKey(keyCode, currentIndex, lastIndex) {
+  const isNoOptionSelected = currentIndex === -1;
+  let newIndex = currentIndex;
+
+  if (keyCode === 36) { // home
```
jamime, line 17: You used the `Event` class in the other files, e.g. `Events.isHomeKey` `Events.isEndKey` `Events.isDownKey` and `Events.isUpKey`. Should we use them here also?

src/components/select/utils/with-filter.hoc.js
```diff
@@ -0,0 +1,64 @@
...
+  }) => {
+    function getFilteredChildren() {
+      let filteredElements = children;
+
+      if (filterText) {
+        filteredElements = filterChildren({ value: filterText })(children);
+
+        if (!filteredElements) {
+          const noResultsText = I18n.t('select.no_results', {
+            defaultValue: 'No results'
```
jamime, line 24: We should have a prop for this if we're going to support `placeholder`.

src/components/select/utils/with-filter.hoc.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import I18n from 'i18n-js';
+import filterChildren from '../../../utils/filter-children';
+import StyledOption from '../option/option.style';
+import highlightPartOfText from './highlight-part-of-text';
+
+/** Filters wrapped component children based on provided filter text and highlights matching content */
+const withFilter = (WrappedComponent) => {
```
jamime, line 9: What is the benefit of having the `with-filter` hoc, could we re-use this or is it only ever going to be used in `FilterableSelect`?

src/components/select/utils/with-filter.hoc.js
```diff
@@ -0,0 +1,64 @@
...
+  const FilteredComponent = ({
+    children,
+    filterText,
+    forwardRef,
```
jamime, line 13: Please can we use `React.forwardRef` instead so we can use the regular `ref` prop.

src/components/select/option/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './option.component';
```
jamime, line 1: We're exposing two ways of using option.

- `import {Option} from 'carbon-react/lib/components/select`
- `import Option from 'carbon-react/lib/components/select/option`

We should have only one way to do this.

src/components/select/option/option.component.js
```diff
@@ -0,0 +1,50 @@
...
+  /** Optional: alternative rendered content, displayed within SelectList of Select (eg: an icon, an image, etc) */
+  children: PropTypes.node,
+  /** The option's invisible internal value */
+  value: PropTypes.oneOfType([PropTypes.object, PropTypes.string]).isRequired,
+  /** Callback to return value when the element is selected (prop added by the SelectList component) */
+  onSelect: PropTypes.func,
+  /** Index of the option in the SelectList (prop added by the SelectList component) */
+  index: PropTypes.number,
+  /** Currently selected index (prop added by the SelectList component) */
+  selectedIndex: PropTypes.number
```
jamime, line 47: We should use the `@ignore` and `@private` jsdoc annotation for the internal prop types. This has the added benefit that they won't be included in the prop tables, no need to ignore them in MDX.

src/components/select/option/option.component.js
```diff
@@ -0,0 +1,50 @@
...
+      { children || text }
+    </StyledOption>
+  );
+};
+
+Option.propTypes = {
+  /** The option's visible text, displayed within Textbox of Select, and used for filtering */
+  text: PropTypes.string.isRequired,
+  /** Optional: alternative rendered content, displayed within SelectList of Select (eg: an icon, an image, etc) */
+  children: PropTypes.node,
```
jamime, line 39: Having the option of `text` and `children` causes this weird bug where you can select an option that does not appear in the list.

https://codesandbox.io/s/select-review-vbx73

src/components/select/filterable-select/filterable-select.stories.mdx
grabkowski: all these props are not taken directly from the Textbox, so you should not be worried about new ones surfacing. If we specify them then we'll have to add 40 lines of code per every select based component (SimpleSelect, FilterableSelect, MultiSelect).

src/components/select/filterable-select/index.js
grabkowski: it's against our `Prefer default export` linting rule. We need to either remove that rule to avoid use of `disable rule` comments or we leave it as it is. I think that it's pretty clear in the documentation how this component should be imported.

src/components/select/select-textbox/select-textbox.component.js
grabkowski: it's the default value of the translation, if we change it then the user's placeholder wouldn't be replaced with default translation?

src/components/select/utils/get-next-child-by-key-code.js
grabkowski: Event class works on events, I'd have to pass down whole event and then import that again there.

I've removed the dependency on the `Event` util as it's using deprecated `which` event property.

src/components/select/utils/with-filter.hoc.js
grabkowski: a placeholder in the option list?

src/components/select/utils/with-filter.hoc.js
grabkowski: it would be used in the `MultiSelect` and `SelectAsync`. It also allows us to not implement these functionality directly in the SelectList.

jamime: Great, thanks for explaining.

src/components/select/option/option.component.js
grabkowski: This is an expected behaviour, please look at the description above the `children` prop. We have allowed to do it in the past version of the component.

src/components/select/filterable-select/index.js
jamime: Sorry for the confusion, I mean we shouldn't allow the component to be imported from two different locations.

We should either delete the file `src/components/select/filterable-select/index.js`.
Or we should remove the named import `FilterableSelect` from `carbon-react/lib/components/select`.

src/components/select/utils/with-filter.hoc.js
jamime: At the moment we have to use I18n to change the "No results" but we can change the placeholder by using a prop or I18n. We should be able to change the "No results" text by using a prop too.

src/components/select/utils/get-next-child-by-key-code.js
jamime: keyCode is also deprecated
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

We should probably use `e.key === 'Home`. Some of the methods in the Events util already do this.

src/components/select/select-textbox/select-textbox.component.js
jamime: You're right, your implementation allows for a user to override the translation with a placeholder. That's more beneficial.

src/components/select/utils/get-next-child-by-key-code.js
grabkowski: keyCode was the name of a variable, it's not related to the event. I have changed that variable name to key as it's no longer keyCode based.

src/components/select/filterable-select/index.js
grabkowski: file deleted

src/components/select/select-list/select-list.component.js
grabkowski: I cannot reproduce it, could you provide the steps you have taken?

mkrds:
I have a question regarding opening the popup. I've noticed that popup does not open when I tab onto the Select and that's correct according to AC's but I've also noticed that when I click on the Select it also does not open. Shouldn't clicking on the Select field always open it?

src/components/select/select-list/select-list.component.js
grabkowski: I think that we cannot memoize this, so we cannot do it better

grabkowski:
> I have a question regarding opening the popup. I've noticed that popup does not open when I tab onto the Select and that's correct according to AC's but I've also noticed that when I click on the Select it also does not open. Shouldn't clicking on the Select field always open it?

it depends on which Select component you are talking about. If it's a SimpleSelect the it should open on click (and it opens right now), if it's Filterable then it should not.

mkrds:
Apart of Jamie's comments only thing that I'd change are the propTypes descriptions. I don't think that it's relevant for the consumer that some of these are `[Form Field]`, also instead of using `exclude` prop on `<Props .../>` maybe we should start using the JSDoc annotations.

src/components/select/option/option.component.js
mkrds: +1 from my side on this being a weird behaviour

src/components/select/select-textbox/select-textbox.component.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+const formInputPropTypes = {
+  /** [Form Field] Id attribute of the input element */
+  id: PropTypes.string,
+  /** [Form Field] Name attribute of the input element */
+  name: PropTypes.string,
+  /** [Form Field] If true the Component will be read-only */
+  readOnly: PropTypes.bool,
+  /** [Form Field] If true the Component will be disabled? */
```
mkrds, line 76: I don't think there's supposed to be a question mark here :D

grabkowski: nice catch ;)

src/components/select/option/option.component.js
jamime: Can we get confirmation from UX that they want this behaviour. Now is our chance to remove legacy defects. cc @harpalsingh

src/components/select/select-list/select-list.component.js
jamime: • Go to https://codesandbox.io/s/select-review-vbx73
• Scroll the preview pane to the right so it looks like this
image
• Open the dropdown using the mouse
image

src/components/select/utils/get-next-child-by-key.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+const homeKey = 'Home';
```
jamime, line 3: These variables are only used in one place. I think having them defined as constants instead of using the strings makes the code more difficult to read.

src/components/select/select-textbox/select-textbox.component.js
```diff
@@ -0,0 +1,120 @@
...
+  onFocus: PropTypes.func,
+  /** Callback function for when the Select Textbox loses it's focus. */
+  onBlur: PropTypes.func,
+  /** Callback function for when the key is pressed when focused on Select Textbox. */
+  onKeyDown: PropTypes.func
+};
+
+SelectTextbox.propTypes = {
+  ...formInputPropTypes,
+  /** @private Value to be displayed in the Textbox */
```
jamime, line 110: `@private` correctly tells developer that this prop is internal, but it doesn't hide it in storybook.
We also need to use `@internal`.

src/components/select/option/option.component.js
jamime: We need `@ignore`.

src/components/select/filterable-select/index.js
jamime: The file is still here.

src/components/select/option/index.js
jamime: Unresolved.

src/components/select/utils/with-filter.hoc.js
jamime: Unresolved.

src/components/select/utils/with-filter.hoc.js
jamime: Unresolved.

src/components/select/option/index.js
jamime: Unresolved.

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -0,0 +1,293 @@
...
+    return {
+      id,
+      name,
+      inputRef: assignInput,
+      selectedValue,
+      formattedValue: textValue,
+      onClick: handleTextboxClick,
+      onKeyDown: handleTextboxKeydown,
+      onChange: handleTextboxChange,
+      ...textboxProps
```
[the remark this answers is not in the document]
jamime: Unresolved.

src/components/select/filterable-select/index.js
jamime: Unresolved.

src/components/select/simple-select/index.js
jamime: Unresolved.

src/components/select/select-textbox/select-textbox.component.js
grabkowski: you mean `@ignore`?

src/components/select/select-list/select-list.component.js
grabkowski: Looks like it's an issue specific to content being embedded in an iframe in the Codesandbox

it affects every carbon component rendered in a portal, for example the Date Component:

Screenshot 2020-07-06 at 16 38 26

Screenshot 2020-07-06 at 16 38 16

src/components/select/utils/with-filter.hoc.js
grabkowski: added a prop with default value

src/components/select/simple-select/index.js
grabkowski: removed index file

src/components/select/filterable-select/index.js
grabkowski: I'm sorry, I must have missed it

src/components/select/filterable-select/filterable-select.component.js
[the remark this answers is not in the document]
grabkowski: added `event.defaultPrevented` check

src/components/select/utils/with-filter.hoc.js
grabkowski: there was a forward ref on the bottom of the file, moved it to the top for clarity

src/components/select/option/index.js
grabkowski: index file removed

src/components/select/option/option.component.js
grabkowski: added `@ignore`

src/components/select/simple-select/simple-select.component.js
```diff
@@ -0,0 +1,335 @@
...
+  }, [selectValueStartingWithText]);
+
+  const handleTextboxKeydown = useCallback(((event) => {
+    const { key } = event;
+
+    if (onKeyDown) {
+      onKeyDown(event);
+    }
+
+    if (isNavigationKey(key)) {
```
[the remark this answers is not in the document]
grabkowski: added a check

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -0,0 +1,302 @@
...
+    updateValues(newValue, isDeleteEvent);
+    setOpen(true);
+  }, [setOpen, updateValues]);
+
+  const handleTextboxKeydown = useCallback(((event) => {
+    if (onKeyDown) {
+      onKeyDown(event);
+    }
+
+    if (isNavigationKey(event.key)) {
```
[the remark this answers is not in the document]
grabkowski: added a check

src/components/select/utils/get-next-child-by-key.js
grabkowski: removed

src/components/select/utils/with-filter.hoc.js
```diff
@@ -21,7 +21,7 @@ const withFilter = (WrappedComponent) => {
 
         if (!filteredElements) {
           const noResultsText = I18n.t('select.no_results', {
-            defaultValue: 'No results'
+            defaultValue: noResultsMessage
```
jamime, line 24: I noticed the old implementation used to have a translation for no_results_for_term, is this no longer supported by the DLS?

Also, in this conversation https://github.com/Sage/carbon/pull/2913/files#r448837015
we chose to allow placeholder to override the translation. We should be doing the same here.
In this case, if they have a translation provided globally, there is no way to override a specific instance.

src/components/select/select-textbox/select-textbox.component.js
jamime: Yes, sorry for the confusion.

src/components/select/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as Option } from './option';
```
jamime, line 1: The import for option is broken.
e.g. `import { Select, Option } from "carbon-react/lib/components/select";` does not work.

src/components/select/utils/with-filter.hoc.js
grabkowski: yup, you're right. Changed that to `no_results_for_term` and moved override out of the translation

src/components/select/option/option.component.js
benwilsonsage: Hello - thanks for pointing me at this thread.
I'd imagine that an <option> would always have some kind of text label (mandatory), and other optional elements (thumbnail image, icon...)
I can't think of an example where an <option> would not have a text label, and I agree that it isn't expected behaviour to appear to choose an item from the list, but then the selected value is then different.
I may be missing some context or detail coming into this fresh, but that's my suggestion anyway.
Happy to discuss if you would like?

src/components/select/index.js
grabkowski: fixed

src/components/select/utils/with-filter.hoc.js
jamime: It actually has both, do we need "no results" and "new results for term"? I think "no results" is displayed if you type a value, then delete it.

src/components/select/select-list/select-list.component.js
jamime: OK, please can you create a JIRA ticket and we'll fix it separately.

src/components/select/utils/with-filter.hoc.js
grabkowski: if there you delete the text then all values are displayed so we don't need that here

jamime: OK no problem.

src/components/select/option/option.component.js
grabkowski: @benwilsonsage what do you propose to be done here?

benwilsonsage: Discussed in detail with Kuba and considered various options.
The tricky part is backwards compatibility with existing implementations - which could have a variety of formats.
We discussed possible checks to prevent the symptom above, but can't be sure this wouldn't result in a breaking change, or prevent a valid implementation. For example, perhaps there may be an implementation that implements text and sub-text, where both option and children are required.
Ultimately it could be argued that the developer can use the flexibility wisely, or implement unexpected behaviour, but that's a factor of their design and implementation.

Given that we couldn't find a practical solution right now, I'd suggest we accept it as a legacy feature, although should be used wisely. 
Although that's my suggestion, caveat that my knowledge of code conventions and best practice is very limited!

src/components/select/select-list/select-list.component.js
jamime: Issue was fixed for this component.

PR #2920, opened 2020-05-11
fix(inline-input): resolve accessibility violations - FE-2785

damianOfKyan:
### Proposed behaviour
Storybook demo should not have any accessibility violations

Screenshot 2020-05-11 at 09 28 15

### Current behaviour
There are number of violations due to incorrect story implementation.

InlineInputs_missing_aria_labels

### Checklist

<del>- [ ] Unit tests
<del>- [ ] Theme support

### Testing instructions
http://localhost:9001/?path=/story/design-system-inline-inputs--default-story


damianOfKyan:
> I don't think that wrapping everything in a `<Label>` component is a good solution, as this fixes only the issue that we have in our storybook. There is a label in the component and the issue should be fixed there by passing proper aria attributes to the component children.
I have addressed the issue by implementing aria-labelledby na removed unnecessary <Label> wrap.

src/components/inline-inputs/inline-inputs.component.js
```diff
@@ -32,17 +33,33 @@ const InlineInputs = (props) => {
     gutter
   } = props;
 
+  const labelId = guid();
```
mkrds, line 36: This will lead to a new id on each render

damianOfKyan: fixed 07f8227

src/components/inline-inputs/inline-inputs.component.js
```diff
@@ -32,17 +33,33 @@ const InlineInputs = (props) => {
...
 
+  const labelId = useRef(createGuid());
+
   function renderLabel() {
     if (!label) return null;
 
-    return <Label inline htmlFor={ htmlFor }>{ label }</Label>;
+    return (
+      <Label
+        labelId={ labelId }
```
grabkowski, line 43: you have to access this value by `labelId .current`, `labelId` is an object

src/components/inline-inputs/inline-inputs.component.js
```diff
@@ -32,17 +33,33 @@ const InlineInputs = (props) => {
...
+      >
+        { label }
+      </Label>
+    );
+  }
+
+  function renderChildren() {
+    if (!label) return children;
+
+    return React.Children.map(children, child => React.cloneElement(child, { 'aria-labelledby': labelId }));
```
grabkowski, line 55: `labelId.current`

damianOfKyan: fixed d7500ff

src/components/inline-inputs/inline-inputs.component.js
damianOfKyan: fixed d7500ff

PR #2943, opened 2020-05-18
feat(accordion): allow passing data-role prop to the root element - FE-2800

### LINKED ISSUE #2933: Accordion component: Unable to pass in data-roles using the component
### Current behaviour
It seems that we're unable to pass in and set data-roles on the Accordion component - https://github.com/Sage/carbon/blob/master/src/components/accordion/accordion.component.js#L21

### Expected behaviour
To be able to pass in and set data-roles on the Accordion component

### Reproducible example
https://codesandbox.io/s/blissful-cache-4he9v

Thanks to @philearley for the provided example

mkrds:
### Proposed behaviour
It should be possible to pass all props including `data-role` to the root element of the Accordion component

### Current behaviour
Currently it's not possible to pass any props that are not listed in prop-types.

### Checklist

### Additional context
Closes #2933
### Testing instructions
Apply `data-role` prop using React DevTools and see if it's rendered on the Accordion


grabkowski:
this change will allow to pass any prop into the underlaying element, do we want to do this?

src/components/accordion/accordion.component.js
```diff
@@ -78,6 +79,7 @@ const Accordion = React.forwardRef(({
       data-component='accordion'
       accordionType={ type }
       styleOverride={ styleOverride.root }
+      { ...rest }
```
damianOfKyan, line 82: maybe its worth using
```
import tagComponent from '../../utils/helpers/tags/tags';
```

sianford:
Tested that props data-role, data-whatever and any-other-prop could be passed into Accordion from React Developer Tools although any-other-prop that was not valid was not rendered, which is expected. Tests passed ✅

PR #2981, opened 2020-06-09
fix(tooltip): offscreen positioning - FE-2811

damianOfKyan:
### Proposed behaviour

When the tooltip goes off the screen to the left it should reposition above the element using the tooltip. 
When the gap to the edge of the screen on either left or right is less than 16px, the tooltip should reposition, rather than only when it's actually off the screen.
image
image

### Current behaviour

The tooltip does not reposition when it goes off the screen to the left.
image

It also doesn't reposition early enough on the right.
image

### Checklist

- <del>[ ] Cypress automation tests added or updated</del>
- <del>[ ] Typescript `d.ts` file added or updated</del>

### Testing instructions
A sandbox using the Textbox component to test is available here:
https://codesandbox.io/s/beautiful-water-xxxqd?file=/src/index.js


src/utils/decorators/tooltip-decorator/tooltip-decorator.js
```diff
@@ -173,51 +173,77 @@ const TooltipDecorator = (ComposedComponent) => {
...
-        pointerX = shifts.tooltipDistances.right;
+      if (tooltipPosition === 'left') {
+        pointerX = shifts.tooltipDistances.left;
       } else {
-        pointerX = shifts.vertical[alignment];
+        pointerX = shifts.vertical[tooltipAlign];
       }
 
-      return window.innerWidth < pointerX + tooltipWidth;
+      return pointerX < 16;
```
mkrds, line 223: Why specifically `16`?

sjohnson-sage: UX asked for the tooltip to reposition when the gap between the edge of the tooltip and its container gets to less than 16px (on both right and left). I've added it to the offscreen right calculation below as well

PR #3026, opened 2020-06-26
docs: storybook examples

jamime:
DO NOT MERGE.

### DIFF
--- a/src/components/example/example.component.js
+++ b/src/components/example/example.component.js
@@ -1,0 +1,21 @@
+/* eslint-disable import/prefer-default-export */
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '../button';
+
+export const Example = ({ children, onClick, enableSuperPower }) => {
+  return <Button buttonType={ enableSuperPower ? 'secondary' : 'primary' } onClick={ onClick }>{children}</Button>;
+};
+
+Example.propTypes = {
+  children: PropTypes.string.isRequired,
+  onClick: PropTypes.func.isRequired,
+  /**
+   * Apply the super power button styling
+   */
+  enableSuperPower: PropTypes.bool
+};
+
+Example.defaultProps = {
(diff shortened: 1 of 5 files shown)


src/components/example/example.component.js
```diff
@@ -0,0 +1,21 @@
...
+  onClick: PropTypes.func.isRequired,
+  /**
+   * Apply the super power button styling
+   */
+  enableSuperPower: PropTypes.bool
+};
+
+Example.defaultProps = {
+  enableSuperPower: false
+};
```
dawidzarzycki, line 21: What about moving defaultProps to the top?
```js
export const Example = ({ children, onClick, enableSuperPower = false }) => {
```

src/components/example/example.test.stories.js
```diff
@@ -0,0 +1,27 @@
...
+      disable: true
+    }
+  }
+};
+
+export const chromatic = () => (
+  <> <p>Story can have any name</p>
+    <Example>An example story for chromatic automation</Example>
+  </>
+);
```
dawidzarzycki, line 21:
```suggestion
  <> 
    <p>Story can have any name</p>
    <Example>An example story for chromatic automation</Example>
  </>
);
```

src/components/example/example.test.stories.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const cypress = () => (
+  <> <p>Story can have any name</p>
+    <Example>An example story for cypress automation</Example>
+  </>
+);
```
dawidzarzycki, line 27:
```suggestion
export const cypress = () => (
  <> 
    <p>Story can have any name</p>
    <Example>An example story for cypress automation</Example>
  </>
);
```

src/components/example/example.component.js
jamime: yep, we should do this.

PR #3035, opened 2020-06-30
docs(rfc): propose internal interfaces

jamime:
Introduce `__internal__` folders to distinguish when components/utilities are for internal use only.

Introduce `@private` and `@ignore` JSDoc annotations to document prop types.

---

Please use 👍  if you approve.


philearley:
I take it private items would not be visible on carbon.sage.com ?

jamime:
> I take it private items would not be visible on carbon.sage.com ?

Correct the `__internal__` components wouldn't have dedicated stories. Although we might have some development only stories that aren't published.

The internal props wouldn't be displayed in the prop tables either.

jamime:
This RFC has entered the "Final Comments" stage. This will be accepted unless there are significant changes requested before Thursday 9 July 2020 1330 BST.

jamime:
I've added some minor changes to explain "internal use" and added further detail to the alternative approaches.

PR #3031, opened 2020-06-30
feat(validations): show tooltip when input is hovered or focused - FE-2812

mkrds:
### Proposed behaviour

Show validations icon tooltip message when either input or it's label is focused or hovered. Same behaviour is appliable when validation icon is displayed on the grouping component label i.e RadioButtonGroup

### Current behaviour

Currently validation icon tooltip message is shown only when validation icon is focused

### Checklist

- [ ] Screenshots are included in the PR
- [x] Carbon implementation and Design System documentation are congruent
- [x] All themes are supported
- [x] Commits follow our style guide
- [x] Unit tests added or updated
- [ ] Cypress automation tests added or updated
- [x] Storybook added or updated
- [ ] Typescript `d.ts` file added or updated

### Additional context
This PR also delivers FE-2813

### Testing instructions
All `validations` stories

PR #3055, opened 2020-07-08
fix(dialog): auto focus dialog when opened

sjohnson-sage:
### Proposed behaviour

Dialog will now by default focus the first focusable element when it is open. This can be disabled using the `disableAutoFocus` prop should the consumer wish to. 

image

It is now possible to specify a element to focus first using the `focusFirstElement` prop by passing in a function which focuses that element. The focus trap can also be disabled using the `disableFocusTrap` prop so the consumer can write their own focus logic if they wish. 

Also includes changes to the advanced color picker component which used the incorrect version of focus-trap. The behaviour of this component hasn't changed but it now handles it's own focus logic rather than focus-trap having specific code for this component.

### Current behaviour

Dialog does not focus the first element by default. If you specify a different element to focus first, this makes it not possible to tab to the first focusable element at all. 

image

### Checklist

- <del>[ ] Typescript `d.ts` file added or updated</del>

### Testing instructions
A sandbox will be provided to assist with testing different scenarios
https://codesandbox.io/s/focused-shtern-9b0dw


src/utils/helpers/focus-trap/focus-trap.js
```diff
@@ -5,48 +5,55 @@ function blockTabbing(ev) {
...
       ev.preventDefault();
     }
   }
 }
+
 // eslint-disable-next-line max-len
 const defaultFocusableSelectors = 'button, [href], input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])';
 
-const focusTrap = (element, focusFirstElement) => {
+const focusTrap = (element, focusFirstElement, bespokeTrap) => {
```
jamime, line 38: I noticed that `focusFirstElement` can have three values, `true`, `false` and `function(){}` I think we should make this two separate arguments for clarity.

```suggestion
const focusTrap = (element, autoFocus = true, focusFirstElement, bespokeTrap) => {
```
`focusFirstElement` can be a function or a DOMElement.

Then this

```js
if (focusFirstElement) {
    if (typeof focusFirstElement === 'boolean') {
      setElementFocus(firstFocusableElement);
    } else {
      setElementFocus(focusFirstElement);
    }
  }
```
would become
```
if (autoFocus) {
     setElementFocus(focusFirstElement || focusableElements[0]);
}
```

src/components/dialog/dialog.component.js
```diff
@@ -256,13 +265,22 @@ Dialog.propTypes = {
   showCloseIcon: PropTypes.bool,
   stickyFormFooter: PropTypes.bool,
   /** function runs when user click close button */
-  onCancel: PropTypes.func
+  onCancel: PropTypes.func,
+  focusFirstElement: PropTypes.func,
```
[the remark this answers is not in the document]
sjohnson-sage: Done!

src/components/advanced-color-picker/advanced-color-picker.component.js
```diff
@@ -50,8 +50,20 @@ const AdvancedColorPicker = ({
...
+    if (e.key === 'Tab' || e.keyCode === 9) {
+      /* istanbul ignore else */
+      if (e.shiftKey) {
+        /* istanbul ignore else */
+        if (document.activeElement === selectedColorRef.current.input.current) {
+          lastFocusableElement.focus();
+          e.preventDefault();
+        }
+      } else if (document.activeElement === lastFocusableElement) {
+        selectedColorRef.current.input.current.focus();
```
mkrds, line 63: This won't work after this https://github.com/Sage/carbon/pull/3031 PR get's merged

src/components/advanced-color-picker/advanced-color-picker.component.js
```diff
@@ -50,8 +50,20 @@ const AdvancedColorPicker = ({
     }
   }, [colors, currentColor, dialogOpen, isOpen]);
 
-  const handleFocus = useCallback(() => {
-    selectedColorRef.current.input.current.focus();
+  const handleFocus = useCallback((e, firstFocusableElement, lastFocusableElement) => {
+    if (e.key === 'Tab' || e.keyCode === 9) {
```
mkrds, line 54: `keyCode` is deprecated, isn't `e.key === 'Tab'` enough?
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

src/components/advanced-color-picker/advanced-color-picker.component.js
sjohnson-sage: No problem, I'll wait until that's merged and sort it out

src/components/advanced-color-picker/advanced-color-picker.component.js
sjohnson-sage: It would be yes, I didn't realise `keyCode` was deprecated - just copied the existing logic in the component, will change that

src/components/advanced-color-picker/advanced-color-picker.component.js
mkrds: Hopefully my PR gets resolved soon. If not then we'll let this one go, and I'll solve this issue on my branch then, just wanted to leave a note to not forget about it later on :)

jamime:
Currently blocked by #3031

src/components/advanced-color-picker/advanced-color-picker.component.js
sjohnson-sage: Note we have discussed and are happy for this PR to be merged first

cypress/features/regression/alert.feature
```diff
@@ -7,7 +7,6 @@ Feature: Alert component
   @positive
   Scenario: CloseIcon has the border outline
     Given I open component preview
```
DlgSHi, line 9: Should be always `Given` - `When` - `Then` of `When` - `Then`
```suggestion
    When I open component preview
```
We need to change `Given` to `When` here - because we removed `When` step

cypress/features/regression/dialogFullScreen.feature
```diff
@@ -7,7 +7,6 @@ Feature: Dialog Full Screen component
   @positive
   Scenario: CloseIcon has the border outline
     Given I open component preview
```
DlgSHi, line 9: Should be always `Given` - `When` - `Then` of `When` - `Then`
```suggestion
    When I open component preview
```
We need to change `Given` to `When` here - because we removed `When` step

cypress/features/regression/sidebar.feature
```diff
@@ -7,7 +7,6 @@ Feature: Sidebar component
 
   @positive
   Scenario: CloseIcon has the border outline
```
DlgSHi, line 9: Should be always `Given` - `When` - `Then` of `When` - `Then`
``` suggestion
Scenario: CloseIcon has the border outline
# commented because of BDD default scenario Given - When - Then
# When I check open checkbox
```
We need to comment `When` here - because the step was actually performed in `Background` and add a comment why it was commented out

PR #3062, opened 2020-07-10
chore(carousel): remove classic theme

DipperTheDan:
removal of classic theme from the component and the classic theme example from storybook.

BREAKING CHANGE: The classic theme has been removed from carousel. To upgrade
please use a DLS theme.

BREAKING CHANGE: The `transition` prop has been removed from carousel. This was used in the classic theme.

fixes FE-2908

### Proposed behaviour
Removal of classic theme from the Carousel component and removal of the classic theme storybook example.

### Current behaviour
Component currently supports classic theme and classic theme examples in storybook.

### Checklist

<del>- [ ] Screenshots are included in the PR</del>
<del>- [ ] Carbon implementation and Design System documentation are congruent</del>
<del>- [ ] Cypress automation tests added or updated</del>

### Additional context
N/A

### Testing instructions
Pull down the branch.
Open storybook using `npm start`
Check the carousel component was as it previously did.
Check that the classic story for Carousel is not present.


src/components/carousel/carousel.component.js
```diff
@@ -160,20 +153,18 @@ class BaseCarousel extends React.Component {
...
-
     return React.cloneElement(visibleSlide, assign({}, visibleSlide.props, additionalProps));
   }
 
   visibleSlides() {
-    const arrayWithKeys = this.props.children.map((element, key) => {
+    const arrayWithKeys = React.Children.map(this.props.children, (element, key) => {
       return React.cloneElement(element, {
-        key: `slide-${key}`,
+        key: `slide-${guid()}`,
```
jamime, line 167: guid's generally shouldn't be used as react keys. To properly fix this, the child needs to have it's own unique identifier, which might be a guid.

Is this something thats broken, if not let's just leave it as the array index, which is the `key` variable here.

src/components/carousel/carousel.spec.js
```diff
@@ -680,7 +628,6 @@ describe('SlideStyle', () => {
     wrapper = mount(<Slide
       transitionName={ () => {} }
```
jamime, line 629:
```suggestion
```
I think `transitionName` is a classic prop

DipperTheDan: It is. Well spotted. Thank you.

src/components/carousel/carousel.component.js
DipperTheDan: It was throwing up warnings when running tests and also in the console when the component was loaded. I figured seen as I here doing all this work, I might as well rectify it. I thought it would be something that would be picked up and kicked back to me in QA.

src/components/carousel/carousel.component.js
```diff
@@ -124,11 +119,9 @@ class BaseCarousel extends React.Component {
 
   /** Re-enables the next and previous buttons after timeout */
   enableButtonsAfterTimeout() {
-    if (isClassic(this.props.theme)) {
-      setTimeout(() => {
-        this.setState({ disabled: false });
-      }, TRANSITION_TIME);
-    }
+    setTimeout(() => {
```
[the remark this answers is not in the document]
DipperTheDan: When I attempted to remove this from the component, the whole component broke. I have left it in for this reason. I plan to add this component to the Tech Debt doc for a refactor anyway. Most of the code seems to have not been touched in four years and uses the old API.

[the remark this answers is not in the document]
DipperTheDan: When I attempted to remove this from the component, the whole component broke. I have left it in for this reason. I plan to add this component to the Tech Debt doc for a refactor anyway. Most of the code seems to have not been touched in four years and uses the old API.

src/components/carousel/carousel.component.js
jamime: We don't get any value from index or guid TBH. React uses them to work out when an element has changed position. If the guid changes on each render, they'll always be replaced. If the index in the array changes, they'll also all be replaced.

Just leave it as guid, it's one less eslint warning and we'll fix it when it's refactored to be a function component.

src/components/carousel/carousel.component.js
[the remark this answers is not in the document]
jamime: It looks like theres nothing setting `disabled: true`. I've ran this locally and we can safely remove the `enableButtonsAfterTimeout` function.

PR #3066, opened 2020-07-13
feat(note): create component and tooltip status FE-2825

edleeks87:
### Proposed behaviour

Component supports some features not present in `Tile`, such as a composable inline control, title and footer content including a status with tooltip. The component uses the `draftjs` editor in `readOnly` mode to support rendering of both plain and rich text content.

The meta content has intentionally been omitted as it may be possible to simply render that as part of the `draft-js` `Editor`'s content

### Current behaviour

This is a new component
### Checklist

### Additional context

With plain text:
image

With Rich text:
image

With title:
image

With inline control:
image
image

With footer:
image
image

### Testing instructions

Can be tested on storybook
`http://localhost:9001/?path=/docs/design-system-note--`...
https://codesandbox.io/s/carbon-quickstart-hwssk?file=/src/index.js

### DIFF
--- a/src/components/note/__internal__/note.component.js
+++ b/src/components/note/__internal__/note.component.js
@@ -1,0 +1,108 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Editor, EditorState } from 'draft-js';
+import {
+  StyledNote,
+  StyledNoteContent,
+  StyledInlineControl,
+  StyledTitle,
+  StyledFooter,
+  StyledFooterContent
+} from './note.style.js';
+import StatusWithTooltip from './status-with-tooltip';
+
+const Note = ({
+  noteContent = EditorState.createEmpty(),
+  width,
+  inlineControl,
+  title,
+  name,
+  createdDate,
+  status,
+  ...rest
+}) => {
+  const [showTooltip, setShowTooltip] = useState(false);
+
+  const renderStatus = () => {
+    if (!status || !status.text || !status.text.length) {
+      return null;
+    }
+
+    const { text, timeStamp } = status;
+
+    return (
+      <StyledFooterContent data-component='note-status'>
+        <StatusWithTooltip { ...tooltipProps(timeStamp, showTooltip, setShowTooltip) }>
+          { text }
+        </StatusWithTooltip>
+      </StyledFooterContent>
+    );
+  };
+
+  return (
+    <StyledNote
+      width={ width }
+      { ...rest }
+      data-component='note'
+    >
+      { title && (
+        <StyledTitle>
+          { title }
+        </StyledTitle>
+      )}
+
+      { inlineControl && (
+        <StyledInlineControl>
+          { inlineControl }
+        </StyledInlineControl>
+      )}
+
+      <StyledNoteContent>
+        <Editor readOnly editorState={ noteContent } />
+      </StyledNoteContent>
+
+      {(name && createdDate) && (
+        <StyledNoteContent>
+          <StyledFooter>
+            <StyledFooterContent>{ name }</StyledFooterContent>
+            <StyledFooterContent>{ createdDate }</StyledFooterContent>
+            { renderStatus() }
+          </StyledFooter>
+        </StyledNoteContent>
+      )}
+    </StyledNote>
+  );
+};
+
+function tooltipProps(id, showTooltip, setShowTooltip) {
+  return {
+    tooltipMessage: id,
+    tooltipPosition: 'top',
+    tooltipAlign: 'center',
+    tooltipVisible: showTooltip,
+    onMouseOver: () => setShowTooltip(true),
+    onMouseLeave: () => setShowTooltip(false)
+  };
+}
+
+Note.propTypes = {
+  /**  The rich text content to display in the Note */
+  noteContent: PropTypes.object,
+  /**
+   * Set a percentage-based width for the whole Note component, relative to its parent.
+   * If unset or zero, this will default to 100%.
+   */
+  width: PropTypes.number,
+  /** renders a control for the Note */
+  inlineControl: PropTypes.node,
+  /** Adds a Title to the Note */
+  title: PropTypes.string,
+  /** Adds a name to the Note footer */
+  name: PropTypes.string,
+  /** Adds a created on date to the Note footer */
+  createdDate: PropTypes.string,
+  /** Adds a status and tooltip to the Note footer */
+  status: PropTypes.shape({ text: PropTypes.string, timeStamp: PropTypes.string })
+};
+
+export default Note;
--- a/src/components/note/__internal__/note.spec.js
+++ b/src/components/note/__internal__/note.spec.js
@@ -1,0 +1,152 @@
+import React from 'react';
+import { ThemeProvider } from 'styled-components';
+import { mount } from 'enzyme';
+import Note from './note.component';
+import baseTheme from '../../../style/themes/base';
+import {
+  StyledNoteContent,
+  StyledInlineControl,
+  StyledTitle,
+  StyledFooter,
+  StyledFooterContent
+} from './note.style';
+import StatusWithTooltip from './status-with-tooltip';
+import { assertStyleMatch } from '../../../__spec_helper__/test-utils';
+import Tooltip from '../../tooltip';
+
+function render(props = {}) {
+  return mount(
+    <ThemeProvider theme={ baseTheme }>
+      <Note { ...props } />
+    </ThemeProvider>
+  );
+}
+
+describe('Note', () => {
+  describe('Styling', () => {
+    it('matches the expected', () => {
+      const wrapper = render();
+
+      assertStyleMatch({
+        backgroundColor: `${baseTheme.colors.white}`,
+        border: `1px solid ${baseTheme.tile.border}`,
+        display: 'flex',
+        flexDirection: 'column',
+        padding: '24px',
+        position: 'relative',
+        width: '100%',
+        minWidth: '314px'
+
+      }, wrapper);
+
+      const content = wrapper.find(StyledNoteContent);
+
+      assertStyleMatch({
+        position: 'relative',
+        width: '100%'
+      }, content);
+
+      assertStyleMatch({
+        paddingBottom: '24px'
+      }, content, { modifier: ':not(:last-of-type)' });
+
+      assertStyleMatch({
+        borderTop: `solid 1px ${baseTheme.tile.separator}`
+      }, content, { modifier: `+ ${StyledNoteContent}` });
+    });
+
+    it('supports dynamic sizing by setting the "width" prop', () => {
+      assertStyleMatch({
+        width: '75%'
+      }, render({ width: 75 }));
+    });
+  });
+
+  describe('StyledTitle', () => {
+    const title = 'title';
+
+    it('does not render the "title" when prop is undefined', () => {
+      expect(render().find(StyledTitle).exists()).toBeFalsy();
+    });
+
+    it('renders the "title" with expected styling when prop has value', () => {
+      const wrapper = render({ title });
+
+      assertStyleMatch({
+        fontWeight: '900',
+        fontSize: '16px',
+        lineHeight: '21px',
+        paddingBottom: '16px'
+      }, wrapper.find(StyledTitle));
+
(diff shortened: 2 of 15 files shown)


src/components/note/__internal__/status-with-tooltip/status.component.js
```diff
@@ -0,0 +1,34 @@
...
+import PropTypes from 'prop-types';
+import StyledStatusIconWrapper from './status.style';
+import TooltipDecorator from '../../../../utils/decorators/tooltip-decorator/tooltip-decorator';
+
+class StatusIcon extends React.Component {
+  render() {
+    return (
+      <>
+        <StyledStatusIconWrapper
+          data-component='content-with-tooltip'
```
DlgSHi, line 11: Should it probably be like:
```suggestion
          data-component='status-with-tooltip'
```

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,110 @@
...
+  const renderStatus = () => {
+    if (!status || !status.text || !status.text.length) {
+      return null;
+    }
+
+    const { text, timeStamp } = status;
+
+    return (
+      <FooterContent>
+        <StatusWithTooltip { ...tooltipProps(timeStamp, showTooltip, setShowTooltip) }>
```
DlgSHi, line 33: will we have here a `data-component` tag for cypress tests?

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,110 @@
...
+  const {
+    noteContent,
+    width,
+    inlineControl,
+    title,
+    name,
+    createdDate,
+    status,
+    ...rest
+  } = props;
```
dawidzarzycki, line 20: Why this
```js
 const {
    noteContent,
    width,
    ...rest
} = props
```
instead of this
```js
const Note = ({noteContent, width, ...rest})
```

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,110 @@
...
+        </StatusWithTooltip>
+      </FooterContent>
+    );
+  };
+
+  return (
+    <StyledNote
+      width={ width }
+      { ...rest }
+      { ...tagComponent('note', props) }
```
dawidzarzycki, line 44: what do you think about replace this above with 
```js
<StyledNote data-component="note" />
```

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,110 @@
...
+    tooltipAlign: 'center',
+    tooltipVisible: showTooltip,
+    onMouseOver: () => setShowTooltip(true),
+    onMouseLeave: () => setShowTooltip(false)
+  };
+}
+
+Note.defaultProps = {
+  noteContent: EditorState.createEmpty()
+};
```
dawidzarzycki, line 88: is it possible to replace 
```js 
Note.defaultProps = {} 
``` 
with
```js 
noteContent = EditorState.createEmpty()
```
at top of the component?

src/components/note/__internal__/note.component.js
edleeks87: Yeah good point I missed that

src/components/note/__internal__/status-with-tooltip/status.style.js
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components';
+
+const StyledStatusIconWrapper = styled.div`
+  :before {
+    content: "";
+    margin-right: -8px;
+  }
+  list-style: disc;
+  display: list-item;
```
dawidzarzycki, line 9:
```suggestion
  :before {
    content: "";
    margin-right: -8px;
  }
  
  list-style: disc;
  display: list-item;
```

src/components/note/__internal__/note.stories.mdx
```diff
@@ -0,0 +1,197 @@
...
+import Note from "carbon-react/lib/components/note";
+import { EditorState, ContentState, convertFromHTML } from 'draft-js';
+```
+
+## Quick Start
+To use `Note`, use the import path above and pass the content via the `noteContent` prop by utilising the static 
+methods provided by the `draft-js` (see the import above) framework https://draftjs.org/docs/api-reference-content-state#static-methods.
+
+
+### Basic Usage:
```
dawidzarzycki, line 30:
```suggestion
## Quick Start
To use `Note`, use the import path above and pass the content via the `noteContent` prop by utilising the static 
methods provided by the `draft-js` (see the import above) framework https://draftjs.org/docs/api-reference-content-state#static-methods.

### Basic Usage:
```

src/components/note/__internal__/note.component.js
edleeks87: I was originally passing the props on to another component, that isn't needed anymore so I'll destructure in the top level as you suggest

src/components/note/__internal__/note.component.js
edleeks87: Good point, not sure why I didn't do that in the first place 🤦

sjohnson-sage: We should be doing this as `defaultProps` is getting deprecated for functional components

src/components/note/__internal__/note.spec.js
```diff
@@ -0,0 +1,149 @@
...
+import Note from './note.component';
+import baseTheme from '../../../style/themes/base';
+import {
+  NoteContent, InlineControl, Title, Footer, FooterContent
+} from './note.style';
+import StatusWithTooltip from './status-with-tooltip';
+import { assertStyleMatch } from '../../../__spec_helper__/test-utils';
+import Tooltip from '../../tooltip';
+
+function render(props, renderer = mount) {
```
sjohnson-sage, line 13: You've not got any other renderers in this file so you can just get rid of this `renderer` parameter and use `mount` below

src/components/note/__internal__/note.style.js
```diff
@@ -0,0 +1,152 @@
+import styled, { css } from 'styled-components';
+import PropTypes from 'prop-types';
+import baseTheme from '../../../style/themes/base';
+
+const NoteContent = styled.div`
```
sjohnson-sage, line 5: On every other component I've looked at so far all the styled components are prefixed with `Styled`, i.e. here would be `StyledNoteContent` - that seems like a good idea so it's immediately obvious when looking at the component file, which bits are styled components.
I'd say worth changing these to be the same

src/components/note/__internal__/note.style.js
```diff
@@ -0,0 +1,152 @@
...
+  position: relative;
+  flex-grow: 0;
```
sjohnson-sage, line 7: `flex-grow` is 0 by default isn't it?

src/components/note/__internal__/note.style.js
```diff
@@ -0,0 +1,152 @@
...
+`;
+
+const InlineControl = styled.div`
+  position: absolute;
+  top: 24px;
+  right: 16px; 
+  z-index: 100
+`;
+
+const Title = styled.div`
```
sjohnson-sage, line 67: Seems strange that title is a `div` and not a header, is there a reason for that?

src/components/note/__internal__/status-with-tooltip/status.component.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import StyledStatusIconWrapper from './status.style';
+import TooltipDecorator from '../../../../utils/decorators/tooltip-decorator/tooltip-decorator';
+
+class StatusIcon extends React.Component {
```
sjohnson-sage, line 6: Why is this a class and no a functional component? I'm very against creating new class components unless it's absolutely necessary

src/components/note/__internal__/status-with-tooltip/status.spec.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import StatusWithTooltip from './status.component';
+import Tooltip from '../../../tooltip';
+import { assertStyleMatch } from '../../../../__spec_helper__/test-utils';
+
+const onMouseOver = jest.fn();
+const onMouseLeave = jest.fn();
+
+const render = (props = {}, renderer = mount) => {
```
sjohnson-sage, line 10: `mount` is the only renderer in this file so you can get rid of this `renderer` parameter and just use `mount below

src/components/note/__internal__/note.style.js
edleeks87: Fair point will update the styled components to be consistent

src/components/note/__internal__/status-with-tooltip/status.component.js
edleeks87: Yeah completely agree, but the tooltip decorator requires a class component so it can hook into `this`

src/components/note/__internal__/note.spec.js
edleeks87: Fair point I usually set them up like this just in case future test cases are written that use different `renderer`s, but in this component I think that is very unlikely so happy to remove

src/components/note/__internal__/status-with-tooltip/status.spec.js
edleeks87: see above reply

src/components/note/__internal__/note.style.js
edleeks87: Good point, will update the styles

src/components/note/__internal__/note.style.js
edleeks87: Yeah good point, the Title was originally just a container but I went a different way and didn't update the styles

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,108 @@
...
+
+  const renderStatus = () => {
+    if (!status || !status.text || !status.text.length) {
+      return null;
+    }
+
+    const { text, timeStamp } = status;
+
+    return (
+      <StyledFooterContent data-component='note-status'>
```
DlgSHi, line 34: Should it be like: 
```suggestion
      <StyledFooterContent data-component="note">
```

src/components/note/__internal__/note.component.js
```diff
@@ -0,0 +1,108 @@
...
+   */
+  width: PropTypes.number,
+  /** renders a control for the Note */
+  inlineControl: PropTypes.node,
+  /** Adds a Title to the Note */
+  title: PropTypes.string,
+  /** Adds a name to the Note footer */
+  name: PropTypes.string,
+  /** Adds a created on date to the Note footer */
+  createdDate: PropTypes.string,
```
DlgSHi, line 103: will it be only a string ? and not a date ?

src/components/note/__internal__/note.component.js
edleeks87: No the parent component already has that data tag I think `note-status` is okay, there is also a tag on the component itself so should be enough for cypress to target

src/components/note/__internal__/note.component.js
edleeks87: Yeah it should be a string IMO, otherwise it has to be formatted internally

src/components/note/__internal__/note.spec.js
sjohnson-sage: Yeah YAGNI principle!

src/components/note/__internal__/status-with-tooltip/status.component.js
sjohnson-sage: Ah fair enough

kylemayne:
Passed UX QA 👍

src/components/note/__internal__/note.style.js
```diff
@@ -0,0 +1,125 @@
...
+StyledNote.defaultProps = {
+  theme: baseTheme
+};
+
+StyledFooter.defaultProps = {
+  theme: baseTheme
+};
+
+export {
+  StyledNote, StyledNoteContent, StyledInlineControl, StyledTitle, StyledFooter, StyledFooterContent
```
sjohnson-sage, line 124: Don't have to change for this PR but next time I suggest putting all these exports on their own lines - much easier to read! Once I've added prettier in on commit that will be done automatically

PR #3068, opened 2020-07-14
docs(rfc): propose i18n upgrade

jamime:
Introduce first class support for internationalisation and localisation within `carbon-react`.

---

Please use 👍  if you approve.

### DIFF
--- a/rfcs/text/i18n.md
+++ b/rfcs/text/i18n.md
@@ -1,0 +1,305 @@
+- Start Date: 2020-07-14
+
+# Summary
+
+Introduce first class support for internationalisation and localisation within `carbon-react`.
+
+# Basic example
+
+The proposal is to introduce a new component `I18nProvider` that allows our consumers to provide a 
+[`react-i18next`](https://github.com/i18next/react-i18next) compatible<sup>[1] [2]</sup> `t` function.
+
+We will need to check each usage of our current `i18n-js` implementation to ensure the:
+* key makes sense;
+* translations are provided in each of our supported languages; and
+* interpolation/plural/formatting works correctly.
+
+`react-i18next` also uses [`Suspense`](https://reactjs.org/docs/concurrent-mode-suspense.html) to allow for async
+translation loading. We don't need to support this within `carbon-react` to adopt this approach, but we will use this in
+the future when we add `loading` states to our components.
+
+We will deprecate our `I18n` component. This component consumes `i18n-js` directly and formats any Markdown content.
+There are better alternatives for Markdown, such as [`react-markdown`](https://github.com/rexxars/react-markdown) which 
+avoids `dangerouslySetInnerHTML`. Our consumers can use `react-markdown` and `react-i18next` in tandem. Removing the
+`I18n` from carbon allows our consumers to patch any Markdown parsing vulnerabilities without requiring a new 
+`carbon-react` release.
+
+We will also deprecate all [`i18n` helpers][3]. Continuing to support these is beyond the scope of a component library
+and is better handled by a dedicated library such as [`numeral.js`](http://numeraljs.com/). Removing these will further
+reduce the complexity of this project and allow for more advanced configurations by our consumers. Where we use these
+helpers internally, we will use a dedicated library if appropriate. We will ensure that any new i18n helpers are
+considered as private methods i.e. marked as `@private` or not exported.
+
+Finally, each translation will be overridable by our consumers by providing a string prop to the component. This string 
+can also be provided by `react-i18next`.
+
+[1]: https://react.i18next.com/latest/usetranslation-hook
+[2]: https://www.i18next.com/overview/api#t
+[3]: https://github.com/Sage/carbon/blob/a59afc80fd7ac954b58da5cd9ef278239b5756c0/src/utils/helpers/i18n/i18n.js
+# Motivation
+
+We currently only publish english translations. Each time a product launches in a non-english market they have to set up
+translations, these are typically non-contextual and will be the same in each product.
+
+Having `i18n-js` as a peer-dependency means that we are encouraging the use of this, unintentionally, for content
+translations. The current recommended setup is `i18next`, so consumers will have to configure both libraries, increasing
+the bundle size and reducing performance.
+
+# Detailed design
+
+```js
+// carbon-react/lib/components/__internal__/I18nContext/index.js
+import {createContext} from "react";
+
+export default createContext(() => {throw new Error("No I18nProvider exists.")});
+```
+
+```js
+// carbon-react/lib/components/I18nProvider/index.js
+import React from "react";
+import Context from "../__internal__/I18nContext";
+export default ({t, children}) => <Context.Provider value={t}>{children}</Context.Provider>
+```
+
+```js
+// carbon-react/lib/components/Example/index.js
+import React, { Suspense, useContext } from "react";
+import Context from "../__internal__/I18nContext";
+export default ({nextLabel}) => {
+  const t = useContext(Context);
+  return <span>{nextLabel || t("example.next")}</span>;
+};
+```
+
+Below is an example of how someone could configure lazy loading of carbon translations.
+This file is accessible from `/locales/fr-FR/carbon.json`
+
+```json
+{
+  "example": {
+    "next": "Suivant",
+    "previous": "Précédent"
+  }
+}
+```
+
+This file is accessible from `/locales/fr-FR/app.json`
+```json
+{
+  "example": {
+    "next": "Dernière page"
+  }
+}
+```
+
+```js
+// App.js
+import React, { Suspense } from "react";
+import HttpApi from "i18next-http-backend";
+import i18n from "i18next";
+import { initReactI18next, useTranslation } from "react-i18next";
+import Example from "carbon-react/lib/components/Example";
+import I18nProvider from "carbon-react/lib/components/I18nProvider";
+
+i18n
+  .use(initReactI18next)
+  .use(HttpApi)
+  .init({
+    interpolation: {
+      escapeValue: false,
+    },
+    lng: "fr-FR",
+    ns: ["carbon", "app"],
+    backend: {
+      loadPath: "/locales/{{lng}}/{{ns}}.json",
+    }
+  });
+
+const Page = () => {
+  const { t } = useTranslation('carbon');
+  const { t: appT } = useTranslation('app');
+  return (
+    <I18nProvider t={t}>
+      <Example />
+      <Example nextLabel={appT('example.next')} />
+    </I18nProvider>
+  );
+};
+
+function App() {
+  return (
+    <Suspense fallback="Loading...">
+      <Page />
+    </Suspense>
+  );
+}
+
+export default App;
+
+```
+
+Consumers may also choose to bundle the translations.
+
+```js
+// App.js
+import React, { Suspense } from "react";
+import i18n from "i18next";
+import { initReactI18next, useTranslation } from "react-i18next";
+import Example from "carbon-react/lib/components/Example";
+import I18nProvider from "carbon-react/lib/components/I18nProvider";
+import frFr from "carbon-react/lib/translations/fr-FR/carbon.json";
+import AppFrFr from "carbon-react/lib/translations/fr-FR/app.json";
+i18n
+  .use(initReactI18next)
+  .init({
+    resources: {
+      'fr-FR': {
+        carbon: frFr,
+        app: AppFrFr
+      }
+    },
+    lng: "fr-FR",
+    ns: ["carbon", "app"],
+    interpolation: {
+      escapeValue: false
+    }
+  });
+
+const Page = () => {
+  const { t } = useTranslation('carbon');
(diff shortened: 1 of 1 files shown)


rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+# Summary
+
+Introduce first class support for internationalisation and localisation within `carbon-react`.
+
+# Basic example
+
+The proposal is to introduce a new component `I18nProvider` that allows our consumers to provide a 
+[`react-i18next`](https://github.com/i18next/react-i18next) compatible<sup>[1] [2]</sup> `t` function.
+
+We will need to check each instance that we use our current `i18n-js` implementation and ensure that the
```
DlgSHi, line 12: I think it will look better
```suggestion
We will need to check each instance that we use our current `i18n-js` implementation and ensure that the:
```

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+* key makes sense
```
DlgSHi, line 13:
```suggestion
* key makes sense;
```

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+* translations are provided in each of our supported languages
```
DlgSHi, line 14:
```suggestion
* translations are provided in each of our supported languages;
```

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+* interpolation/plural/formatting works correctly
```
DlgSHi, line 15:
```suggestion
* interpolation/plural/formatting works correctly.
```

rfcs/text/i18n.md
sjohnson-sage: I also don't think this sentence makes sense looking at it - maybe "We will need to check each instance where our current `i18n-js` implementation is used" ?

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+import {createContext} from "react";
+
+export default createContext(() => {throw new Error("No I18nProvider exists.")});
+```
+
+```js
+// carbon-react/lib/components/I18nProvider/index.js
+import React from "react";
+import Context from "../__internal__/I18nContext";
+export default ({t, children}) => <Context.Provider value = {t}>{children}</Context.Provider>
```
sjohnson-sage, line 43: Very minor but it should be `value={t}` on this line, without spaces

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+  .use(initReactI18next)
+  .use(HttpApi)
+  .init({
+    interpolation: {
+      escapeValue: false,
+    },
+    lng: "fr-FR",
+    ns: ["carbon", "app"],
+    backend: {
+      loadPath: "/locales/{{lng}}/{{ns}}.json",
```
sjohnson-sage, line 87: More of a question but I've never seen this method of string interpolation in js, how does this work?

rfcs/text/i18n.md
jamime: Changed to
> We will need to check each usage of our current `i18n-js` implementation to ensure the:

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
+- Start Date: 2020-06-14
```
DlgSHi, line 1: Should the start date be like 
```suggestion
- Start Date: 2020-07-14
```

rfcs/text/i18n.md
jamime: change to `; and`

rfcs/text/i18n.md
jamime: 🤦 fixed

rfcs/text/i18n.md
jamime: fixed

ianforster-sage:
This document focusses on translations, should we discuss here formats for date, currency, number etc. that are also within scope of internationalisation and localisation?

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+  .use(initReactI18next)
+  .use(HttpApi)
+  .init({
+    interpolation: {
+      escapeValue: false,
+    },
+    lng: "fr-FR",
+    ns: ["carbon", "app"],
+    backend: {
+      loadPath: "/locales/{{lng}}/{{ns}}.json",
```
thehiggster, line 87: src/i18n/translations/{{lng}}.json

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+import I18nProvider from "carbon-react/lib/components/I18nProvider";
+
+i18n
+  .use(initReactI18next)
+  .use(HttpApi)
+  .init({
+    interpolation: {
+      escapeValue: false,
+    },
+    lng: "fr-FR",
```
thehiggster, line 84: May seem petty, but the format will be frFR.
File names, weirdly, being fr-FR.json

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+import frFr from "carbon-react/lib/translations/fr-FR/carbon.json";
+i18n
+  .use(initReactI18next)
+  .init({
+    resources: {
+      'fr-FR': {
+        carbon: frFr
+      }
+    },
+    lng: "fr-FR",
```
thehiggster, line 133: as above, frFR

rfcs/text/i18n.md
```diff
@@ -0,0 +1,270 @@
...
+import i18n from "i18next";
+import { initReactI18next, useTranslation } from "react-i18next";
+import Example from "carbon-react/lib/components/Example";
+import I18nProvider from "carbon-react/lib/components/I18nProvider";
+import frFr from "carbon-react/lib/translations/fr-FR/carbon.json";
+i18n
+  .use(initReactI18next)
+  .init({
+    resources: {
+      'fr-FR': {
```
thehiggster, line 129: frFR

rfcs/text/i18n.md
jamime: The format used (`fr-FR`) is  is the IEFT BCP 47 tag. The first part is a language tag in ISO 639-1 alpha-2 format. The second part is the region tag in ISO 3166-1 alpha-2. The import name can’t be `fr-FR` hence `frFr` for the variable that references the import file.

rfcs/text/i18n.md
jamime: The load path is the URL rather than a file path. It will also need a namespace so we can keep carbon/app translations separate.

jamime:
> This document focusses on translations, should we discuss here formats for date, currency, number etc. that are also within scope of internationalisation and localisation?

My main priority is to remove the `i18n-js` peer dependency.
Our current `formatDecimal` function can be upgraded, it's used internally by the `Decimal` component.

Our `formatCurrency` uses `i18n-js`'s `toCurrency` which is not implemented in `i18next`. We don't actually use that function in any of our components and it's probably better handled by a dedicated library such as [`numeral.js`](http://numeraljs.com/#locales).

I'll add this to the adoption strategy.

jamime:
There should be no need to use the `I18n`component, we should deprecate it. I'll update the RFC to reflect this.

jamime:
Posting on our internal Slack for feedback from other teams within Sage.

jamime:
> Do our consumers need to ability to override a specific translation in one instance, or should we enforce consistency?
>
> E.g. Can consumers change the label on `Pager` to say `Next Page` on one instance, but display `Next` on another instance.

@benwilsonsage @harpalsingh what do you think?

benwilsonsage:
> > Do our consumers need to ability to override a specific translation in one instance, or should we enforce consistency?
> > E.g. Can consumers change the label on `Pager` to say `Next Page` on one instance, but display `Next` on another instance.
> 
> @benwilsonsage @harpalsingh what do you think?

Hello! Certainly we'd need to handle translated text strings for different countries.
I'd imagine that the label (previous/next) would usually be the same in all instances - but I wonder if there might be a case to have something like 'Finish' on the final step, or maybe 'Agree and Continue' if asking the user to accept terms as one step in the middle of the process.
If it's easy to make it customisable, that's great. As a middleground, a different string for the last step would satisfy most cases I reckon.

jamime:
> > > Do our consumers need to ability to override a specific translation in one instance, or should we enforce consistency?
> > > E.g. Can consumers change the label on `Pager` to say `Next Page` on one instance, but display `Next` on another instance.
> > 
> > 
> > @benwilsonsage @harpalsingh what do you think?
> 
> Hello! Certainly we'd need to handle translated text strings for different countries.
> I'd imagine that the label (previous/next) would usually be the same in all instances - but I wonder if there might be a case to have something like 'Finish' on the final step, or maybe 'Agree and Continue' if asking the user to accept terms as one step in the middle of the process.
> If it's easy to make it customisable, that's great. As a middleground, a different string for the last step would satisfy most cases I reckon.

I've updated the RFC to show how consumers can override the default translations that we provide.

jamime:
Entering the final comment period which ends on 2020-08-01 10:22.

rfcs/text/i18n.md
```diff
@@ -0,0 +1,298 @@
...
+We will deprecate our `I18n` component. This component consumes `i18n-js` directly and formats any Markdown content.
+There are better alternatives for Markdown such as (`react-markdown`)[https://github.com/rexxars/react-markdown] which avoids `dangerouslySetInnerHTML`.
+Our consumers can use `react-markdown` and `react-i18next` in tandem. Removing the `I18n` from carbon allows our consumers to patch any 
+Markdown parsing vulnerabilities without requiring a new `carbon-react` release.
+
+We will also deprecate all [`i18n` helpers](https://github.com/Sage/carbon/blob/a59afc80fd7ac954b58da5cd9ef278239b5756c0/src/utils/helpers/i18n/i18n.js). Continuing to support these is beyond the scope of a component library and is better handled by a dedicated library such as [`numeral.js`](http://numeraljs.com/).
+Removing these will further reduce the complexity of this project and allow for more advanced configurations by our consumers.
+
+Finally, each translation will be overridable by our consumers by providing a string prop to the component. This string 
+be also be provided by `react-i18next`.
```
sjohnson-sage, line 28: Typo here on `be also be`, I think it should have been `can also be`

rfcs/text/i18n.md
```diff
@@ -0,0 +1,298 @@
...
+
+[1]: https://react.i18next.com/latest/usetranslation-hook
+[2]: https://www.i18next.com/overview/api#t
+
+# Motivation
+
+We currently only ship english translations. Each time a product launches in a non-english market they have to set up
```
sjohnson-sage, line 35: Should be capital E on English really

rfcs/text/i18n.md
```diff
@@ -0,0 +1,298 @@
...
+  .init({
+    interpolation: {
+      escapeValue: false,
+    },
+    lng: "fr-FR",
+    ns: ["carbon", "app"],
+    backend: {
+      loadPath: "/locales/{{lng}}/{{ns}}.json",
+    },
+    interpolation: {
```
sjohnson-sage, line 109: Why does it need the interpolation object twice?

rfcs/text/i18n.md
```diff
@@ -0,0 +1,298 @@
...
+* interpolation/plural/formatting works correctly.
+
+`react-i18next` also uses [`Suspense`](https://reactjs.org/docs/concurrent-mode-suspense.html) to allow for async translation loading. We don't need to support this within `carbon-react` to adopt this approach, but we will use this in the future when we add `loading` states to our components.
+
+We will deprecate our `I18n` component. This component consumes `i18n-js` directly and formats any Markdown content.
+There are better alternatives for Markdown such as (`react-markdown`)[https://github.com/rexxars/react-markdown] which avoids `dangerouslySetInnerHTML`.
+Our consumers can use `react-markdown` and `react-i18next` in tandem. Removing the `I18n` from carbon allows our consumers to patch any 
+Markdown parsing vulnerabilities without requiring a new `carbon-react` release.
+
+We will also deprecate all [`i18n` helpers](https://github.com/Sage/carbon/blob/a59afc80fd7ac954b58da5cd9ef278239b5756c0/src/utils/helpers/i18n/i18n.js). Continuing to support these is beyond the scope of a component library and is better handled by a dedicated library such as [`numeral.js`](http://numeraljs.com/).
```
chrisbarber86, line 24: What's the plan for the ones we use internally right now?  Like `formatDecimal` in the Decimal Component

https://github.com/Sage/carbon/blob/c7a51a2e53542c31790966c68d7e4d381640e0dd/src/__experimental__/components/decimal/decimal.component.js#L229

jamime: We'll also update our components to use a dedicated library, in this case numeral.js. When we do this we'll move this helper to a `__internal__` folder, or even inline it within the component itself.

rfcs/text/i18n.md
jamime: Good spot :)

PR #3079, opened 2020-07-20
feat: add new props to accordion

sjohnson-sage:
### Proposed behaviour

Add 5 new props to Accordion component as per new designs: 
- Size: "large" or "small"   -> changes padding and text size, large size can have a sub title
- Scheme: "transparent" or "white"  -> changes background colour only
- Borders: "default" or "full"  -> toggles left and right borders
- Custom Padding: value in px  -> adds extra top and bottom padding
- Width: string e.g. "50%" or "500px"  -> changes with of whole accordion

Remove "type" prop, as its functionality can be achieved with a combination of the new "scheme" and "borders" props. Note it is no longer possible to have no top and bottom border, these will always be present (by design).

Aligning the icon left will now rotate the icon correctly.
image

### Current behaviour

These props do not currently exist, and most functionality achieved by theme is currently not possible in the component.

Previously aligning the icon left would rotate the icon the same as aligning it right, which for arrow icons didn't look right.
image

### Checklist


src/components/accordion/accordion.spec.js
```diff
@@ -158,33 +159,93 @@ describe('Accordion', () => {
...
-        border: `1px solid ${baseTheme.accordion.border}`
+        border: `1px solid ${baseTheme.accordion.border}`,
+        borderLeft: 'none',
+        borderRight: 'none'
       }, wrapper.find(StyledAccordionContainer));
     });
 
-    it('renders accordion with transparent background and no border when type is set to "secondary" (default)', () => {
-      render({ type: 'secondary' });
+    it('renders accordion with transpart background and border when scheme is set to "transparent"', () => {
```
edleeks87, line 171: `transpart` -> `transparent`

harpalsingh:
@sjohnson-sage There is one thing to fix, on hover there is no "theme" colors used on the title, the text should stay black-90.

harpalsingh:
@sjohnson-sage Thanks all good! approved, I have some thoughts about the animation, but I will take a look at that myself a bit later to see if we can make it "feel" smoother, but will review that on what we might want to do based on a cubic bezier.

PR #3099, opened 2020-07-27
feat(hr): add new hr component

sjohnson-sage:
### Proposed behaviour

Add new `Hr` component. This is primarily to be used to make a distinct separation between fields in a form but it could have other uses as well.
image

It has 4 props, one for each margin to enable the consumer to display it in alignment with their specific use case. The top and bottom margin props are integer values that will be multiplied by the theme spacing constant (8px). The left and right margins can be any valid CSS margin string in order to make alignment easy for the consumer.

### Current behaviour

Component does not exist.

### Checklist

- <del>[ ] Cypress automation tests added or updated</del>

### Testing instructions
Divider testable at https://codesandbox.io/s/carbon-quickstart-chix8


src/components/divider/divider.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface DividerProps {
+  /** Margin top, this value will be multiplied by the theme spacing constant (8) */
+  marginTop: FormFieldSpacing;
+
+  /** Margin bottom, this value will be multiplied by the theme spacing constant (8) */
+  marginBottom: FormFieldSpacing;
+
+  /** Margin left, any valid css value */
+  marginLeft: string;
```
edleeks87, line 12: Should `marginLeft` and `marginRight` not have `?` as they don't have default values and aren't required?

src/components/divider/divider.stories.mdx
```diff
@@ -0,0 +1,84 @@
+
```
edleeks87, line 1: Empty line here

src/components/divider/divider.d.ts
sjohnson-sage: Good point yes they should!

src/components/divider/divider.component.js
```diff
@@ -0,0 +1,33 @@
...
+
+const Divider = ({
+  marginTop = 3,
+  marginBottom = 3,
+  marginLeft,
+  marginRight
+}) => {
+  return (
+    <StyledDivider
+      marginTop={ marginTop }
```
DlgSHi, line 14: could we add some `data-component='divider'` here? We will need it further in `cypress` tests

src/components/divider/divider.d.ts
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import { FormFieldSpacing } from '../../utils/helpers/options-helper/options-helper';
+
+export interface DividerProps {
+  /** Margin top, this value will be multiplied by the theme spacing constant (8) */
+  marginTop: FormFieldSpacing;
+
```
DlgSHi, line 7: do we need white lines here ?

sjohnson-sage: All the other components seem to so I just followed that convention, happy to remove though

PR #3101, opened 2020-07-27
 feat: add new MultiSelect component (FE-2791)

grabkowski:
### Proposed behaviour
Add new MultiSelect Component to replace the `enableMultiSelect` functionality of the Experimental Select

Screenshot 2020-07-27 at 17 28 26

Screenshot 2020-07-27 at 17 31 17

### Checklist

### Testing instructions
1. run npm start locally.
2. open http://localhost:9001/?path=/story/design-system-select-multiselect--basic


src/components/select/multi-select/multi-select.component.js
```diff
@@ -0,0 +1,416 @@
...
+    if (!isControlled.current && onChange) {
+      onChange(createCustomEvent(selectedValue));
+    }
+
+    setRepositionTrigger((previousValue) => {
+      return !previousValue;
+    });
+  }, [createCustomEvent, onChange, selectedValue]);
+
+  function handleTextboxClick(event) {
```
mkrds, line 199: I'd stick to defining component methods using either `const method = () => {}` or `function method() {}`. Pick one and stick with it.

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -288,7 +288,6 @@ const FilterableSelect = React.forwardRef(({
           ref={ listboxRef }
           id={ selectListId.current }
           labelId={ labelId.current }
-          role='listbox'
```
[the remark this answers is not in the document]
grabkowski: it's already in the list component and this one is not being passed down anyway

src/components/select/multi-select/multi-select.component.js
grabkowski: i'm defining all of these methods with `function`, only these that have to rely on `useCallback` are defined using `const` here, as other solution that I could think of is:
```
function handleSomething() {
  return useCallback(() => {
  .
  .
  .
  });
}
```
which is less readable in my opinion

src/components/select/multi-select/index.d.ts
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react';
+import Option from '../option';
+
+export interface MultiSelectProps {
+  /** [Form Field] Id attribute of the input element */
```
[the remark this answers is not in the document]
grabkowski: right, looks like I've missed this one

src/components/select/multi-select/multi-select.component.js
mkrds: I'd just define everything using arrow functions, but I leave it up to you

grabkowski: when using const you have to think where you will place your method as it cannot be below anything that uses it.
In my opinion it leads to a messy code.

PR #3114, opened 2020-08-03
feat(menu): add keyboard navigation - FE-2927

dawidzarzycki:
### Proposed behaviour
Add keyboard navigation to menu component
image

### Current behaviour
keyboard navigation is not available

### Checklist

### Testing instructions
storybook/menu


src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+        }
+        return index - 1;
+      }
+      return index;
+    };
+    const tempIndex = findIndex();
+    let newIndex = tempIndex;
+    actualIndex = tempIndex;
+
+    if (index === submenuItemsRefs.current.length) {
```
jamime, line 50: it would be great to define a variable that describes the condition e.g.

```js
const isFocusOnLastMenuItem = index === submenuItemsRefs.current.length;
if(isFocusOnLastMenuItem){...}
```

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+      newIndex = 0;
+    } else if (index === -1) {
```
jamime, line 52: We should add a variable for this condition also.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+    if (submenu && isOpen) {
+      if (!openByArrowUp) {
+        submenuItemsRefs.current[0].current.focus();
+      } else {
+        submenuItemsRefs.current[submenuItemsRefs.current.length - 1].current.focus();
+      }
+    }
+  }, [isOpen, openByArrowUp, submenu, submenuItemsRefs]);
+
+  const onCloseSubmenu = () => {
```
jamime, line 69:
```suggestion
  const onCloseSubmenu = useCallback(() => {
```

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+    setIsOpen(false);
+    setOpenByArrowUp(false);
+  };
+
+  const onKeyDownSubmenu = (ev, index) => {
```
jamime, line 74:
```suggestion
  const onKeyDownSubmenu = useCallback((ev, index) => {
```

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+      if (Events.isDownKey(ev)) {
+        setFocusToSubmenuElement(ev, index + 1);
+      }
+      if (Events.isUpKey(ev)) {
+        setFocusToSubmenuElement(ev, index - 1);
+      }
+      if (
+        Events.isLeftKey(ev)
+        || Events.isRightKey(ev)
+        || (Events.isShiftKey(ev) && Events.isTabKey(ev))
```
jamime, line 95:
```suggestion
```

This line can be removed, but keep the test.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -61,7 +187,8 @@ const MenuItem = ({
     icon,
     hasSubmenu: Boolean(submenu),
     selected,
-    menuType
+    menuType,
+    tabbable: !!isFirstElement
```
jamime, line 191: `isFirstElement` is a bool, no need to coerce the value.
```suggestion
    tabbable: isFirstElement
```

src/components/menu/menu.component.js
```diff
@@ -1,8 +1,74 @@
...
+  const menuItemsRefs = useRef(React.Children.map(children, child => child.ref || React.createRef()));
+  let actualIndex;
+  const setFocusToElement = (event, index) => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    let newIndex = index;
+    actualIndex = index;
+    if (index === menuItemsRefs.current.length) {
```
jamime, line 16: As before, please add more descriptive variables for the `if` condition and the `index` variables.

src/components/menu/menu.component.js
```diff
@@ -1,8 +1,74 @@
...
+      newIndex = 0;
+    } else if (index === -1) {
+      newIndex = menuItemsRefs.current.length - 1;
+    }
+
+    menuItemsRefs.current[newIndex].current.focus();
+  };
+
+  const handleKeyDown = (event, index, isOpen) => {
```
jamime, line 25:
```suggestion
  const handleKeyDown = useCallback((event, index, isOpen) => {
```

src/components/menu/menu.component.js
```diff
@@ -14,12 +80,18 @@ const Menu = ({ menuType = 'light', children }) => {
...
               return (
                 <StyledMenuItem>{
                   React.cloneElement(
                     child,
-                    { menuType },
+                    {
+                      menuType,
+                      ref: menuItemsRefs.current[index],
+                      isFirstElement,
+                      handleKeyDown: (ev, isOpen) => handleKeyDown(ev, index, isOpen)
```
jamime, line 93: Changing the function each time will cause a re-render. I think we can update the `handleKeyDown` function to use `isOpen` from state then we can memoize the index.
```suggestion
                      handleKeyDown: useCallback(ev => handleKeyDown(ev,index), [index])
```

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+      if (
+        Events.isLeftKey(ev)
+        || Events.isRightKey(ev)
+        || (Events.isShiftKey(ev) && Events.isTabKey(ev))
+        || Events.isTabKey(ev)
+      ) {
+        onCloseSubmenu();
+      }
+      if (Events.isHomeKey(ev)) {
+        submenuItemsRefs.current[0].current.focus();
```
jamime, line 101: We need to ensure for every keypress that we handle that we `e.preventDefault()`. On a submenu, we might also need `e. stopPropagation()` to prevent the `Menu` key handler from processing the event too.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+    const findIndex = () => {
+      if (submenuItemsRefs.current[index] && submenuItemsRefs.current[index].current.tagName === 'DIV') {
+        if (Events.isDownKey(event)) {
+          return index + 1;
+        }
+        return index - 1;
+      }
+      return index;
+    };
+    const tempIndex = findIndex();
```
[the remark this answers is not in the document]
dawidzarzycki: I think there is no need to have `tempIndex`, `newIndex` and `actualIndex`. I replaced it all with just one `actualFocusedItemIndex`

src/components/menu/menu.component.js
dawidzarzycki: I removed `newIndex` and `actualIndex` and replaced it all with `actualFocusedItemIndex`

src/components/link/link.component.js
```diff
@@ -156,12 +157,14 @@ Link.propTypes = {
   /** A routing component to render when the to prop is set */
   routerLink: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
   /** Target property in which link should open ie: _blank, _self, _parent, _top */
-  target: PropTypes.string
+  target: PropTypes.string,
+  /** Ref to be forwarded */
+  innerRef: PropTypes.object
```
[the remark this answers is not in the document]
dawidzarzycki: Link component breaks because of story.js and Component.defaultProps, I have added forwardRef to Link what causes problems with storybook and actions, but if you type props without Component.defaultProps then it works good

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,152 @@
...
+  const [openByArrowUp, setOpenByArrowUp] = useState(false);
+  const submenuItemsRefs = submenu && useRef(React.Children.map(children, child => child.ref || React.createRef()));
+  const ifRef = () => !submenu && { ref };
+  let actualIndex;
+  const setFocusToSubmenuElement = (event, index) => {
+    if (event) {
+      event.preventDefault();
+    }
+
+    const findIndex = () => {
```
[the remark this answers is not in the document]
jamime: ah yes, didn't realise it was nested.

src/components/link/link.component.js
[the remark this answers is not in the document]
jamime: I think we still need to mark `innerRef` as `@private` and `@ignore`.

src/components/menu/menu.component.js
jamime: Unresolved

src/components/link/link.component.js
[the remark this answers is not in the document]
jamime: I think we still need to mark `innerRef` as `@private` and `@ignore`.

src/components/menu/menu.component.js
jamime: Unresolved

src/components/link/link.component.js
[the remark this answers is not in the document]
dawidzarzycki: It is updated, you will need to choose "show all changes" option probably to see it

sjohnson-sage:
I think there's an issue, if you click on a dropdown menu, you can't immediately use the arrow key to navigate down into it, you have to go across to a different menu item and back before you can go down into the sub menu:

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+      actualFocusedItemIndex.current = 0;
+    } else if (isFocusOnFirstMenuItem) {
+      actualFocusedItemIndex.current = submenuItemsRefs.current.length - 1;
+    }
+
+    submenuItemsRefs.current[actualFocusedItemIndex.current].current.focus();
+  }, [submenuItemsRefs]);
+
+  const onCloseSubmenu = useCallback(() => {
+    setIsOpenSubmenu(null);
```
jamime, line 66: The events in here are triggered by a document event listener, this means they are not batched.
https://github.com/facebook/react/issues/14259#issuecomment-439632622

This will cause three updates

- setIsOpenSubmenu
- setOpenByMouseEnter
- setOpenByArrowUp

this might cause some unexpected behaviour and will cause unnecessary re-renders.

We can avoid this by using a state object and updating the state in one go. You might want to use a state reducer to do this. https://reactjs.org/docs/hooks-reference.html#usereducer

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+  handleKeyDown,
+  isOpen,
+  setIsOpenSubmenu,
+  menuItemIndex
+}, ref) => {
+  const [didOpenByArrowUp, setOpenByArrowUp] = useState(false);
+  const [didOpenByMouseEnter, setOpenByMouseEnter] = useState(false);
+  const [didOpenByClick, setOpenByClick] = useState(false);
+  const submenuItemsRefs = submenu && useRef(React.Children.map(children, child => child.ref || React.createRef()));
+  const ifRef = () => !submenu && { ref };
```
jamime, line 34: Why is the ref conditional based on submenu? The `ref` prop should be accessible by the user and should be consistent.

e.g. 

```js
const myRef = useRef();
<Menu><MenuItem ref={myRef}>Option One</MenuItem></Menu>
```

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+      }
+      if (Events.isAlphabetKey(ev)) {
+        // A-Za-z: focus the next item on the list that starts with the pressed key
+        // selection should wrap to the start of the list
+        ev.stopPropagation();
+        let firstMatch;
+        let nextMatch;
+
+        React.Children.forEach(children, ({ props }, i) => {
+          if (props.children && props.children.toString().toLowerCase().startsWith(ev.key.toLowerCase())) {
```
jamime, line 155: There is a small bug where the order is not correct on first load.
In this example pressing `i` once should move the focus down to the second item. Currently the first two presses of `i` has no effect.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -70,14 +256,17 @@ const MenuItem = ({
 
   return (
     <StyledMenuItemWrapper
+      { ...ifRef() }
```
jamime, line 259: Why is this conditional?

src/components/menu/menu.component.js
```diff
@@ -1,8 +1,80 @@
...
+    if (isFocusedOnLastMenuItem) {
+      actualFocusedItemIndex.current = 0;
+    } else if (isFocusedOnFirstMenuItem) {
+      actualFocusedItemIndex.current = menuItemsRefs.current.length - 1;
+    }
+
+    menuItemsRefs.current[actualFocusedItemIndex.current].current.focus();
+  }, []);
+
+  const handleKeyDown = useCallback((event, index) => {
```
jamime, line 27: We should call the users `onKeyDown` we also need to check to see if they did `e.preventDefault()` in their `onKeyDown` if they did we should not run our `handleKeyDown`.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+      if (!didOpenByArrowUp && !didOpenByMouseEnter && !didOpenByClick) {
+        submenuItemsRefs.current[0].current.focus();
+      } else if (didOpenByArrowUp && !didOpenByMouseEnter && !didOpenByClick) {
+        submenuItemsRefs.current[submenuItemsRefs.current.length - 1].current.focus();
+      }
+    }
+  // eslint-disable-next-line max-len
+  }, [didOpenByArrowUp, submenu, submenuItemsRefs, detectClickOutside, isOpen, didOpenByMouseEnter, onCloseSubmenu, setIsOpenSubmenu, menuItemIndex, ref, didOpenByClick]);
+
+  const onKeyDownSubmenu = useCallback((ev, index) => {
```
jamime, line 92: See comment below about user `onKeyDown`

src/components/link/link.component.js
```diff
@@ -156,12 +157,18 @@ Link.propTypes = {
...
 };
 
 Link.defaultProps = {
   iconAlign: 'left',
   tabbable: true
 };
 
-export default Link;
+export default React.forwardRef((props, ref) => (<Link innerRef={ ref } { ...props } />));
```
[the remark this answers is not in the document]
mkrds: Is `forwardRef` even needed here? Class based components forward refs themselves.

src/components/menu/menu.component.js
```diff
@@ -1,8 +1,80 @@
...
+      event.preventDefault();
+      setFocusToElement(event, index + 1);
+    }
+
+    if (Events.isLeftKey(event)) {
+      event.preventDefault();
+      setFocusToElement(event, index - 1);
+    }
+
+    if (!isOpenSubmenu) {
```
mkrds, line 38: Is this condition necessary here?

src/components/menu/menu.component.js
```diff
@@ -1,8 +1,80 @@
...
+      } else if (Events.isEndKey(event)) {
+        event.preventDefault();
+        menuItemsRefs.current[menuItemsRefs.current.length - 1].current.focus();
+      } else if (Events.isAlphabetKey(event)) {
+        // A-Za-z: focus the next item on the list that starts with the pressed key
+        // selection should wrap to the start of the list
+        event.stopPropagation();
+        let firstMatch;
+        let nextMatch;
+        const hasSubmenu = (element) => {
```
mkrds, line 51: `hasSubmenu` - it should be renamed to something more descriptive like `getMenuText` or `getMenuLabel`

src/components/menu/menu-item/menu-item.component.js
mkrds: +1, looks super confusing, if the conditional is necessary here just write it directly on the component
`<StyledMenuItemWrapper  { ...(!submenu && { ref } />`

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+  const onCloseSubmenu = useCallback(() => {
+    setIsOpenSubmenu(null);
+    setOpenByMouseEnter(false);
+    setOpenByArrowUp(false);
+  }, [setIsOpenSubmenu]);
+
+  const detectClickOutside = useCallback((e) => {
+    if (!Events.composedPath(e).includes(submenuWrapperRef.current)) {
+      onCloseSubmenu();
+      document.removeEventListener('click', detectClickOutside);
```
mkrds, line 74: I think that this eventListener should be also removed when component unmounts.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,211 @@
...
+        }
+
+        if (Events.isUpKey(ev)) {
+          setOpenByArrowUp(true);
+          ev.preventDefault();
+          setOpenByClick(false);
+          setFocusToSubmenuElement(ev, submenuItemsRefs.current.length - 1);
+        }
+      }
+      if (Events.isDownKey(ev) && !didOpenByClick) {
```
mkrds, line 120: Shouldnt this `if (Events.isDownKey(ev) && !didOpenByClick) {`
be put in 
`
if (didOpenByClick) {}
else { HERE } 
`

src/components/menu/menu-item/menu-item.component.js
dawidzarzycki: I moved the conditional to the component without additional function `ifRef`, I need to have the conditional here because if there is a `submenu` prop we need to push the ref deeper to another one and also we don't want to put `ref` in a `div` element

src/components/menu/menu-item/menu-item.component.js
dawidzarzycki: I need to have the conditional here because if there is a submenu prop we need to push the ref deeper to another one and also we don't want to put ref in a div element

src/components/menu/menu.component.js
dawidzarzycki: You are right, I think that `getMenuText` is fine

src/components/menu/menu.component.js
dawidzarzycki: Yes, we need this conditional because our MenuItem is in use as a MenuItem adn SubmenuItem at the same time

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,242 @@
...
+  const initialState = {
+    didOpenByArrowUp: false,
+    didOpenByMouseEnter: false,
+    didOpenByClick: false
+  };
+
+  // eslint-disable-next-line consistent-return
+  function reducer(state, action) {
+    // eslint-disable-next-line default-case
+    switch (action.type) {
```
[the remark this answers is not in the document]
dawidzarzycki: @sjohnson-sage ye, @jamime is right, I misunderstood. It should be ok right now

sjohnson-sage (approved):

src/components/menu/menu-item/menu-item.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { IconTypes } from '../../../utils/helpers/options-helper/options-helper';
+
+export interface MenuItemProps {
+  children?: React.ReactNode;
+  onClick?: () => void;
```
jamime, line 6: The onClick needs to have the argument type. Without it it will cause issues for Typescript projects https://github.com/Sage/carbon/issues/3143

src/components/menu/menu.component.js
jamime: `onKeyDown` has not been resolved in this file.

src/components/menu/menu.component.js
jamime: I _think_ @mkrds is suggesting this because the event will already be handled, we're probably just missing a `e.stopPropagation()` in the sub-`MenuItem`.

src/components/menu/menu-item/menu-item.component.js
```diff
@@ -1,33 +1,243 @@
...
+          dispatch({ type: 'setOpenByArrowUp', payload: true });
+          setIsOpenSubmenu(menuItemIndex);
+        }
+      }
+
+      if (isOpen) {
+        if (state.didOpenByClick) {
+          if (Events.isDownKey(ev)) {
+            ev.preventDefault();
+            dispatch({ type: 'setOpenByClick', payload: false });
```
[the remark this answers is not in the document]
dawidzarzycki: We have to remove it to avoid some bugs because there is some conditions based on it

jamime (commented):
All issues resolved. Commits need squashing before being merged.

jamime (approved):

sjohnson-sage (approved):

sjohnson-sage (dismissed):

mkrds (dismissed):

jamime:
*Please squash commits before merging*

DlgSHi (approved):

mkrds (approved):

sjohnson-sage (approved):

PR #3117, opened 2020-08-04
feat(numeral-date): add label and fix onchange method in numeral-date component - FE-2957/FE-3022

mkrds:
### Proposed behaviour
This PR adds `label` to the NumeralDate component. Other props like `labelHelp`, `labelInline`, `labelAlign`, `labelWidth` and `fieldHelp` were also added so that the NumeralDate behaves similarly to other textbox based inputs.

This PR also changes the arguments passed to the `onChange` prop. `event.target.value` will now have the same structure as the `value` prop passed to this component. Also props `name` and `id` will be available as `event.target.name` and `event.target.id`

### Current behaviour
1. Not possible to add a label
2. Not possible to distinguish which of the partial date inputs is being changed because, there were no `name` or `id` assigned. Also `event.target.value` was a value of currently active input - meaning it was just a raw string like `12`.

### Checklist


src/__experimental__/components/numeral-date/numeral-date.component.js
```diff
@@ -37,86 +53,91 @@ const NumeralDate = ({
...
+    </FormField>
   );
 };
 
 NumeralDate.propTypes = {
-  /** Prop array string to define custom layout. Use a comma to seperate values. */
+  /** Array of strings to define custom input layout. I.e ['dd', 'mm'] */
   dateFormat: PropTypes.arrayOf(PropTypes.string),
-  /** Prop for `uncontrolled` use */
+  /** Default value for use in 'uncontrolled` mode  */
```
edleeks87, line 138: A mix of apostrophe and back tick here

src/__experimental__/components/numeral-date/numeral-date.component.js
```diff
@@ -15,18 +17,32 @@ const NumeralDate = ({
...
+  labelInline,
+  labelWidth,
+  labelAlign,
+  labelHelp,
+  fieldHelp
 }) => {
+  const { current: uniqueId } = useRef(id || guid());
   const isControlled = value !== undefined;
   const initialValue = isControlled ? value : defaultValue;
-  invariant(
```
edleeks87, line 22: It might be worth having an invariant to cover switching between controlled and uncontrolled

jamime: Why did we remove this invariant?

src/__experimental__/components/numeral-date/numeral-date.component.js
```diff
@@ -37,86 +53,91 @@ const NumeralDate = ({
...
-      name: ev.target.name,
-      id: ev.target.id,
-      value: { ...dateValue }
-    };
+    if (newValue.length <= datePart.length) {
+      const newDateValue = { ...dateValue, [datePart]: newValue };
+      setDateValue(newDateValue);
 
-    setIsActive(false);
+      /* istanbul ignore else */
```
jamime, line 63: IMO we shouldn't be using `/* istanbul ignore else */` because all tests will still pass if we remove the `if` condition.

src/__experimental__/components/numeral-date/numeral-date.component.js
mkrds: It was not possible to have this component work without any value provided.

src/__experimental__/components/numeral-date/numeral-date.component.js
mkrds: On the other hand we have to write dummy tests like 'does nothing if onChange is not provided'

PR #3123, opened 2020-08-10
fix(note): remove min width to allow note to size dynamically

edleeks87:
### Proposed behaviour

The missing `box-sizing: border-box` and has
a `min-width` have been added, this means it aligns with a `fullWidth` `Button` for example. The footer has also been updated to allow the content to wrap at smaller screen sizes.

### Current behaviour

Currently the `Note` component overflows its parent as it is missing `box-sizing: border-box` and has
a `min-width`. The footer does not wrap as there is a min-width on the component

### Checklist

<del>- [ ] Cypress automation tests added or updated<del>

### Additional context

Example of mis-align with Button:
image

Overflowing parent: 
image

No longer overflowing:
image

codesandbox of changes:
https://codesandbox.io/s/carbon-quickstart-5q11p?file=/src/index.js

### DIFF
--- a/src/components/note/__internal__/note.style.js
+++ b/src/components/note/__internal__/note.style.js
@@ -45,24 +45,21 @@ const StyledTitle = styled.header`
 `;
 
 const StyledFooterContent = styled.div`
-  position: relative;
   line-height: 21px;
-
-  &:first-of-type {
-    font-weight: bold;
-    font-size: 14px;
-    top: 16px;
-    height: 21px;
-  }
-
+  align-items: baseline
   ${({ theme }) => `
+    &:first-of-type {
+      font-weight: bold;
+      font-size: 14px;
+      margin-top: ${2 * theme.spacing}px;
+    }
+
     &:nth-of-type(2) {
       font-weight: bold;
       font-size: 12px;
       color: ${theme.note.timeStamp};
-      left: 16px;
-      top: 17px;
-      height: 21px;
+      margin-left: ${2 * theme.spacing}px;
+      margin-top: ${2 * theme.spacing}px;
     }
 
     &:last-of-type:not(:nth-of-type(2)) {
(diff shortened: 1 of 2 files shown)


src/components/note/__internal__/note.style.js
```diff
@@ -45,42 +45,37 @@ const StyledTitle = styled.header`
...
+
     &:nth-of-type(2) {
       font-weight: bold;
       font-size: 12px;
       color: ${theme.note.timeStamp};
-      left: 16px;
-      top: 17px;
-      height: 21px;
+      margin-left: ${2 * theme.spacing}px;
+      margin-top: ${1 + (2 * theme.spacing)}px;
```
sjohnson-sage, line 61: Why is the extra 1px being added to the spacing multiplier here? Unless UX specifically want it as 17px I would argue for just using `2 * theme.spacing` so we can be consistent with using the spacing multiplier. 

On the other hand if it just needs to be 17px no matter what, I don't think you should use the spacing multiplier here

src/components/note/__internal__/note.style.js
```diff
@@ -45,42 +45,37 @@ const StyledTitle = styled.header`
...
 
     &:last-of-type:not(:nth-of-type(2)) {
       font-weight: bold;
       font-size: 12px;
       color: ${theme.note.timeStamp};
       cursor: pointer;
-      top: 17px;
-      height: 21px;
-      left: 40px;
+      margin-top: ${1 + (2 * theme.spacing)}px;
```
sjohnson-sage, line 69: Same question as above, I think we should either use a multiple of `theme.spacing` or just use a px value

src/components/note/__internal__/note.style.js
edleeks87: The alignment is off, I assume because the font size is smaller for the last to content blocks. I agree I'll go with hardcoding

sjohnson-sage: Ah I see, just pulled it locally and had a look. You can leave this as 16px, all you need to do is set `align-items: baseline` on the div surrounding the content blocks, it's already a flex container so that will align the text correctly

PR #3133, opened 2020-08-13
test(app-wrapper/button): apply new url approach

DlgSHi:
### Proposed behaviour

Apply new `url` approach in tests for:
`app wrapper` (timing: was `00:20` -> `00:07`),
`button` (timing: was  `00:46` -> `00:15`).

### Current behaviour

Currently we are passing all parameters via `knobs`.

### Checklist

- [ ] Unit tests added or updated


cypress/support/step-definitions/button-steps.js
```diff
@@ -1,88 +1,101 @@
...
+  buttonSubtextPreviewIframe,
+} from '../../locators/button';
 import { icon } from '../../locators';
 import { positionOfElement } from '../helper';
 
 Then('Button label on preview is {word}', (label) => {
   buttonDataComponent().should('have.text', label);
 });
 
+Then('Button label on preview is {word} in IFrame', (label) => {
```
KatarzynaQA, line 14: Can we use here `label on preview is {word}` from `common-steps.js` step which is a reusable step in our code

DlgSHi: this step is looking for a `[data-element="label"]`, but we are looking for a `[data-component="button"]` in this step. So no, we can't

PR #3136, opened 2020-08-13
feat(validations): display validations when field is readOnly

### LINKED ISSUE #3097: Validation highlights should be possible on disabled and read-only controls.
### Current behaviour
On disabled/read-only fields there is limited highlighting available for validation errors. The tool-tip goes away completely.

### Desired behaviour
We have many cases where showing a validation tooltip in a disabled/read-only field is necessary. Perhaps we want to show an info to explain *why* a field is disabled. 

More commonly, we want to highlight warnings and errors in computed fields which are appropriately read-only. Perhaps the total value of an order which due to user defined settings, can not exceed a certain amount and be posted back to the server.

### Suggested solution
Allow us to over-ride the default behaviour with props and still highlight validation issues on disabled fields for the many use-cases we have.

### Describe alternatives you've considered
In our grid, we're using a custom (carbon-looking) renderer control in order to highlight validation errors in disabled fields. We have no work around for general form fields but might have to craft a similar one to our grid as this use-case isn't something we can ignore and is built into the design of our desktop application which customers are familiar with.

### Additional context
I appreciate the source of this issues "Why would a field be invalid if you can't type into it?" but as mentioned, we have many cases where that's true. Most commonly, where it is a calculated field that relies on other input. Those inputs themselves aren't inherently invalid but business logic depends on this computed field being within a certain value range. 

### Your environment
N/A

### Codesandbox
https://codesandbox.io/s/carbon-3097-b9wkt

nicktitchmarsh:
#1810 # Proposed behaviour

Issue #3097 was raised in regards to validation highlighting on readOnly and disabled fields. A discussion in UX determined that disabled fields should not be used to show validation, instead readOnly fields should be used. Styling should be consistent with other error stylings on normal inputs just with the grey background for readOnly.

Additionally I have added custom prop validation to form-field components that will give a console error if disabled is used i conjunction with error, warning or info props.

Fixes FE-3007, Fixes #3097

Screenshot 2020-08-14 at 10 20 39

Screenshot 2020-09-04 at 14 39 05
Screenshot 2020-09-04 at 15 37 28

### Current behaviour
Currently incorrect styling on readOnly with the wrong border thickness and disabled fields are incorrectly having error styles applied to them

Screenshot 2020-08-13 at 11 26 04

### Checklist

<del> - [ ] Carbon implementation and Design System documentation are congruent
<del> - [ ] All themes are supported

<del>- [ ] Cypress automation tests added or updated

<del>- [ ] Typescript `d.ts` file added or updated</del>

### Testing instructions

Put inputs into a readOnly or disabled state and set an error message

### Codesandbox
https://codesandbox.io/s/happy-sammet-gkp5r?file=/src/index.js

See comment below for latest build


jamime (approved):

ianforster-sage (approved):

lukeemmerson:
@nicktitchmarsh should all the validation states have the same thickness of border? Looks like some have 1px and others 2px currently.

In fact via the DLS, it looks like only error overrides the border. I believe for warning, info etc we should just be using iconography. http://designsystem.sage.com.s3-website.eu-central-1.amazonaws.com/components/text-inputs/

DlgSHi (approved):

jamime (approved):

edleeks87 (approved):

DlgSHi (approved):

PR #3147, opened 2020-08-18
test(button-as-sibling/buttonToggle/card): apply new url approach

DlgSHi:
### Proposed behaviour

Apply new `url` approach in tests for:
`buttonAsSiblings` (timing: was `01:47` -> `00:07`),
`buttonToggle` (timing: was  `00:46` -> `00:25`),
`card` (timing  `00:50` -> `00:18`).

### Current behaviour

Currently we are passing all parameters via `knobs`.

### Checklist

- [ ] Unit tests added or updated


cypress/features/regression/card.feature
```diff
@@ -1,50 +1,47 @@
...
-      | 999px |
+      | width | nameOfObject |
+      | 550px | width550     |
+      | 999px | width999     |
 
   @positive
   Scenario Outline: Set the width of Card component to outOfScope <width>
-    When I set width to "<width>"
+    When I open default "Card" component in noIFrame with "card" json from "commonComponents" using "<nameOfObject>" object name
     Then Card component has not set width to "<width>"
```
KatarzynaQA, line 26:
```suggestion
    Then Card component has not set width to <width>
```

cypress/features/regression/card.feature
```diff
@@ -1,50 +1,47 @@
...
     Then Card component is interactive
 
   @positive
   Scenario: Verify the shadow whithout interactive card
-    # When I open "Card" component page
+    When I open default "Card" component in noIFrame with "card" json from "commonComponents" using "interactiveCardFalse" object name
     Then Card component has non-interactive shadow
 
   @positive
   Scenario: Verify the interactive card shadow
```
KatarzynaQA, line 44: I think that we can delate this TC and move step `Then Card component has interactive shadow` to the `Enable interactive card checkbox` test case as `And` step

cypress/features/regression/card.feature
DlgSHi: no, we can't do that, because the input accepts `string`

cypress/features/regression/card.feature
DlgSHi: No, because we need to perform one more step before that (hovering) and this is as another TC

PR #3161, opened 2020-08-20
fix(help): navigate to URL when href is set

### LINKED ISSUE #3009: Help component not opening destination link specified in href
### Current behaviour
Clicking on the help icon does not open the URL in the href unless you right click on the icon and choose to open link in another tab/window.

### Expected behaviour
The destination URL is opened.

### Reproducible example
https://codesandbox.io/s/wizardly-diffie-lffkd

### Suggested solution(s)
As in the expected behaviour.

### Your environment
Please see code sandbox (latest versions).

nicktitchmarsh:
Fixes FE-2906, Fixes #3009

### Proposed behaviour
Should open links when href is set in props

### Current behaviour
Does not open as preventDefault is set on the click handler, this is no longer needed

### Checklist

### Testing instructions
Click on the help icon should open a new tab, open the codesandbox view in the popout window otherwise the href wont navigate inside the iframe. Known issue when tooltip is rendered outside visible window, this will cause the href to not work. Ticket has been raised to address this.


src/components/help/help.spec.js
```diff
@@ -82,11 +82,11 @@ describe('Help', () => {
     });
   });
 
-  it('calls preventDefault to prevent clicking interacting with the input', () => {
+  it('do not prevent clicking interacting with the input', () => {
```
sjohnson-sage, line 85: Very minor obviously but should probably read "it does not prevent clicking..."

sjohnson-sage (approved):

src/components/help/help.spec.js
edleeks87: Could we not just delete this test instead?

jamime: We could, but I think it's best to keep it to prevent regressions. If it was a new component and new functionality, I wouldn't write a negative assertion;

nicktitchmarsh: Changed inline with Sam's suggestion

jamime (changes requested):
This doesn't seem to work. I tried it in codesandbox and copied the HTML of the component. The HTML works fine, so it must be something to do with the event handler in the component still.

https://codesandbox.io/s/lucid-feather-94449

src/components/help/help.component.js
```diff
@@ -40,8 +40,7 @@ const Help = (props) => {
   }
 
   function handleFocusBlur(bool) {
-    return (ev) => {
-      ev.stopPropagation();
+    return () => {
```
jamime, line 43: I think we should also add a negative assertion test for this method too.

jamime (changes requested):
Let's add a test that covers the focus/blur to prevent any regressions.

jamime (approved):

sjohnson-sage (approved):

edleeks87:
@nicktitchmarsh I'm still not able to get this to work following the instructions you've added for testing, I have the same issue when running the sandbox code locally as well unfortunately but it works when @jamime tests it

sjohnson-sage:
> @nicktitchmarsh I'm still not able to get this to work following the instructions you've added for testing, I have the same issue when running the sandbox code locally as well unfortunately but it works when @jamime tests it

I only checked this in storybook where it works, but it doesn't work in the sandbox for me either actually

sjohnson-sage:
> > @nicktitchmarsh I'm still not able to get this to work following the instructions you've added for testing, I have the same issue when running the sandbox code locally as well unfortunately but it works when @jamime tests it
> 
> I only checked this in storybook where it works, but it doesn't work in the sandbox for me either actually

On further inspection it only doesn't work in the sandbox as it blocks you from opening links. This works in storybook and pulling your branch locally for me, so all looks good from my point of view

jamime (changes requested):
This is currently failing on `Mojave 10.14` but passing on `Catalina 10.15`.

jamime (commented):
Confirmed not an OS issue. It's caused when the tooltip displays partially off-screen. @nicktitchmarsh is going to raise a separate issue for all Tooltips as its unrelated to the Help component.

jamime (approved):

DlgSHi (approved):

PR #3162, opened 2020-08-20
feat: add styled system to carbon

sjohnson-sage:
### Proposed behaviour
Use `styled-system` to give all components the same spacing options. This can be used in any component like this: 

```
// form.style.js
import styled from 'styled-components';
import { space } from 'styled-system';
const StyledForm = styled.div`
  ${space}
  ... other styles
`;
```

In that example `Form` would then have the following props:

image

The following spacing options (in px) have been added to the theme: 
`space: [0, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80]`

All of the props above can either be a number, which is an index of the theme space array, i.e. `ml={1}` relates to `8` in the array (so effectively behaves as a multiplier of 8), or a string, which can be any valid CSS string.

### Current behaviour
Spacing is currently handled individually in components, and varies from component to component. This is both hard to maintain and prone to bugs/inconsistent components.

### Checklist

- <del>[ ] Unit tests added or updated if required</del>
- <del>[ ] Cypress automation tests added or updated if required</del>
- <del>[ ] Storybook added or updated if required</del>
- <del>[ ] Typescript `d.ts` file added or updated if required</del>

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button.style.js
+++ b/src/__experimental__/components/radio-button/radio-button.style.js
@@ -17,6 +17,7 @@ const RadioButtonStyle = styled(CheckboxStyle)`
     theme,
     inline
   }) => css`
+    padding-top: 8px;
     margin-bottom: 12px;
 
     ${FieldHelpStyle} {
--- a/package-lock.json
+++ b/package-lock.json
@@ -9429,6 +9429,116 @@
         }
       }
     },
+    "@styled-system/background": {
+      "version": "5.1.2",
+      "resolved": "https://registry.npmjs.org/@styled-system/background/-/background-5.1.2.tgz",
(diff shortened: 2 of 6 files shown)


src/__experimental__/components/radio-button/radio-button.style.js
```diff
@@ -17,6 +17,7 @@ const RadioButtonStyle = styled(CheckboxStyle)`
     theme,
     inline
   }) => css`
+    padding-top: 8px;
```
mkrds, line 20: How about adding the same `pt` prop defaulted to `1` (as in in `Checkbox` now) to `RadioButton` instead of adding this line?

sjohnson-sage: I could have, but only wanted to add styled-system to one component in this PR just as an example to people so I decided just to do this for now. If you really want me to change it I can? It will be getting updated soon anyway as I'll be going through all the input components and adding styled-system

edleeks87 (approved):

mkrds (approved):

Issue #3143, opened 2020-08-17
Button incorrect typescript definition


damianOfKyan:
### Current behaviour
Currently (carbon v38.0.1) when using the Button component, the typescript linter issues a problem when providing a `onClick` prop.

<img width="489" alt="Screenshot 2020-08-17 at 08 19 15" src="https://user-images.githubusercontent.com/47245766/90363124-612d5880-e062-11ea-95ca-78d485eaa82d.png">

```
No overload matches this call.
  Overload 1 of 2, '(props: ButtonProps | HTMLProps<HTMLButtonElement>, context?: any): ReactElement<any, any> | Component<ButtonProps | HTMLProps<...>, any, any> | null', gave the following error.
    Type '{ children: TFunctionResult; buttonType: "tertiary"; size: "medium"; "data-role": string; onClick: (event: MouseEvent<HTMLButtonElement, MouseEvent>) => void; }' is not assignable to type '(IntrinsicAttributes & ButtonProps) | (IntrinsicAttributes & HTMLProps<HTMLButtonElement>)'.
      Types of property 'onClick' are incompatible.
        Type '(event: React.MouseEvent<HTMLButtonElement>) => void' is not assignable to type '() => void'.
  Overload 2 of 2, '(props: PropsWithChildren<ButtonProps | HTMLProps<HTMLButtonElement>>, context?: any): ReactElement<any, any> | Component<...> | null', gave the following error.
    Type '{ children: TFunctionResult; buttonType: "tertiary"; size: "medium"; "data-role": string; onClick: (event: MouseEvent<HTMLButtonElement, MouseEvent>) => void; }' is not assignable to type '(IntrinsicAttributes & ButtonProps & { children?: ReactNode; }) | (IntrinsicAttributes & HTMLProps<HTMLButtonElement> & { ...; })'.
      Types of property 'onClick' are incompatible.
        Type '(event: React.MouseEvent<HTMLButtonElement>) => void' is not assignable to type '() => void'.
```

### Expected behaviour
No linter errors should be present.

### Reproducible example
https://codesandbox.io/s/buttonevent-h5kyk?file=/src/index.tsx

### Suggested solution(s)
ButtonProps interface for Button component needs to be updated and address the following:
1. onClick?: () => void;
2. Button: React.ComponentType<ButtonProps | React.HTMLProps<HTMLButtonElement>>

jamime:
FE-3039

jamime:
Same issue in NumeralDate `onChange`.

PR #3185, opened 2020-08-31
fix(button): update typescript definitions - FE-3039

hegou33:
### Proposed behaviour

Resolves issue (https://github.com/Sage/carbon/issues/3143) by including typescript definitions.

### Current behaviour

Currently (carbon v38.0.1) when using the Button component, the typescript linter issues a problem when providing a `onClick` prop.

Screenshot 2020-08-17 at 08 19 15

```
No overload matches this call.
  Overload 1 of 2, '(props: ButtonProps | HTMLProps<HTMLButtonElement>, context?: any): ReactElement<any, any> | Component<ButtonProps | HTMLProps<...>, any, any> | null', gave the following error.
    Type '{ children: TFunctionResult; buttonType: "tertiary"; size: "medium"; "data-role": string; onClick: (event: MouseEvent<HTMLButtonElement, MouseEvent>) => void; }' is not assignable to type '(IntrinsicAttributes & ButtonProps) | (IntrinsicAttributes & HTMLProps<HTMLButtonElement>)'.
      Types of property 'onClick' are incompatible.
        Type '(event: React.MouseEvent<HTMLButtonElement>) => void' is not assignable to type '() => void'.
  Overload 2 of 2, '(props: PropsWithChildren<ButtonProps | HTMLProps<HTMLButtonElement>>, context?: any): ReactElement<any, any> | Component<...> | null', gave the following error.
    Type '{ children: TFunctionResult; buttonType: "tertiary"; size: "medium"; "data-role": string; onClick: (event: MouseEvent<HTMLButtonElement, MouseEvent>) => void; }' is not assignable to type '(IntrinsicAttributes & ButtonProps & { children?: ReactNode; }) | (IntrinsicAttributes & HTMLProps<HTMLButtonElement> & { ...; })'.
      Types of property 'onClick' are incompatible.
        Type '(event: React.MouseEvent<HTMLButtonElement>) => void' is not assignable to type '() => void'.
```

### Checklist

### Testing instructions

Compare https://github.com/Sage/carbon/issues/3143 issue codesandbox and props applied. They should no longer raise the error when applied in this PR's codesandbox.

### DIFF
--- a/src/__experimental__/components/numeral-date/numeral-date.d.ts
+++ b/src/__experimental__/components/numeral-date/numeral-date.d.ts
@@ -1,5 +1,27 @@
 import * as React from 'react';
 
+interface DayMonthDate {
+   dd: string;
+   mm: string;
+}
+
+interface MonthYearDate {
+   mm: string;
+   yyyy: string;
+}
+
+interface FullDate extends DayMonthDate {
+   yyyy: string;
+}
+
+interface NumeralDateEvent {
+  target: {
+    name: string,
+    id: string,
+    value: DayMonthDate | MonthYearDate |  FullDate,
+  };
+}
+
 export interface NumeralDateProps {
   /* Array of strings to define custom input layout.
   Allowed formats:
@@ -26,9 +48,9 @@ export interface NumeralDateProps {
   Pass true boolean to only display blue border */
   info?: boolean | string;
   /** Blur event handler  */
-  onBlur?: () => void;
+  onBlur?: (ev: NumeralDateEvent) => void;
   /** Change event handler */
-  onChange?: () => void;
+  onChange?: (ev: NumeralDateEvent) => void;
   /** `id` for events */
   id?: string;
   /** `name` for events */
--- a/src/components/button/button.d.ts
+++ b/src/components/button/button.d.ts
@@ -14,7 +14,7 @@ export interface ButtonProps {
   children?: React.ReactNode;
   renderRouterLink?: (args: object) => React.ReactNode;
   forwardRef?: () => void;
-  onClick?: () => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement | HTMLLinkElement>) => void;
 }
 declare const Button: React.ComponentType<ButtonProps | React.HTMLProps<HTMLButtonElement>>;
 export default Button;


grabkowski (commented):
shouldn't the commit message be `docs(button): update typescript definitions`?

jamime:
> shouldn't the commit message be `docs(button): update typescript definitions`?

I'm happy for `fix` as we want it to trigger a release. Docs is better suited for README's and mdx files that do not fix any bugs.

src/__experimental__/components/numeral-date/numeral-date.d.ts
```diff
@@ -26,9 +26,9 @@ export interface NumeralDateProps {
   Pass true boolean to only display blue border */
   info?: boolean | string;
   /** Blur event handler  */
-  onBlur?: () => void;
+  onBlur?: (ev: React.SyntheticEvent<HTMLElement>) => void;
```
jamime, line 29: Numeral date has a custom `ev` object. See https://github.com/Sage/carbon/blob/35cf0e369a982e63b5a271f22fcc5eb027b50852/src/__experimental__/components/numeral-date/numeral-date.component.js#L50

src/__experimental__/components/numeral-date/numeral-date.d.ts
```diff
@@ -26,9 +26,9 @@ export interface NumeralDateProps {
...
   /** Change event handler */
-  onChange?: () => void;
+  onChange?: (ev: React.ChangeEvent<HTMLElement>) => void;
```
jamime, line 31: Numeral date has a custom `ev` object. See https://github.com/Sage/carbon/blob/35cf0e369a982e63b5a271f22fcc5eb027b50852/src/__experimental__/components/numeral-date/numeral-date.component.js#L50

src/components/button/button.d.ts
```diff
@@ -14,7 +14,7 @@ export interface ButtonProps {
   children?: React.ReactNode;
   renderRouterLink?: (args: object) => React.ReactNode;
   forwardRef?: () => void;
-  onClick?: () => void;
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```
jamime, line 17: Button can sometimes be an `a` element instead of `button`.
https://github.com/Sage/carbon/blob/35cf0e369a982e63b5a271f22fcc5eb027b50852/src/components/button/button.component.js#L26

jamime (changes requested):

grabkowski (approved):

jamime (changes requested):
Thank's for the PR @hegou33 

Please can you change the `object` to be an `interface` it has a well defined structure. You can declare it inline in the same file, and there is no need to export it. See my previous review for the object structure.

jamime (dismissed):

src/__experimental__/components/numeral-date/numeral-date.d.ts
```diff
@@ -1,5 +1,17 @@
...
+  target: {
+    name: string,
+    id: string,
+    value: {
+     dd: string,
+     mm: string,
+     yyyy?: string,
+   },
+  };
+}
```
grabkowski, line 13: we have multiple allowed formats that could be passed as that value:
```
  Allowed formats:
  ['dd', 'mm', 'yyyy'],
  ['mm', 'dd', 'yyyy'],
  ['dd', 'mm'],
  ['mm', 'dd'],
  ['mm', 'yyyy']
```

we need to add interfaces for them

```suggestion
interface DayMonthDate {
   dd: string;
   mm: string;
}

interface MonthYearDate {
   mm: string;
   yyyy: string;
}

interface FullDate extends DayMonthDate {
   yyyy: string;
}

interface NumeralDateEvent {
  target: {
    name: string,
    id: string,
    value: DayMonthDate | MonthYearDate |  FullDate,
  };
}
```

grabkowski, line 13: we have multiple allowed formats that could be passed as that value:
```
  Allowed formats:
  ['dd', 'mm', 'yyyy'],
  ['mm', 'dd', 'yyyy'],
  ['dd', 'mm'],
  ['mm', 'dd'],
  ['mm', 'yyyy']
```

we need to add interfaces for them

```suggestion
interface DayMonthDate {
   dd: string;
   mm: string;
}

interface MonthYearDate {
   mm: string;
   yyyy: string;
}

interface FullDate extends DayMonthDate {
   yyyy: string;
}

interface NumeralDateEvent {
  target: {
    name: string,
    id: string,
    value: DayMonthDate | MonthYearDate |  FullDate,
  };
}
```

dawidzarzycki (approved):

grabkowski (approved):

PR #3216, opened 2020-09-14
fix(note): update inline control variant check constructor FE-3134

edleeks87:
### Proposed behaviour

Updates the invariant for the `inlineControl` prop to check if  the passed component's constructor,
rather than its string name, is `ActionPopover`. The previous implementation throws an error in production build environments when the code is minified.

### Current behaviour

Invariant checks `inlineControl.type.name` which works in development but throws in production environment.

### Checklist

### DIFF
--- a/src/components/note/__internal__/note.component.js
+++ b/src/components/note/__internal__/note.component.js
@@ -11,6 +11,7 @@ import {
   StyledFooterContent
 } from './note.style.js';
 import StatusWithTooltip from './status-with-tooltip';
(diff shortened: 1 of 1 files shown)


jamime (approved):

jamime (approved):
Confirmed working on built version using CRA.

* Does not throw when using `ActionPopover`
* Throws error for any other component

sjohnson-sage (approved):

PR #3230, opened 2020-09-22
fix(defition-list): change white-space to normal - FE-3155

arvig33:
### Proposed behaviour

The label or `<Dt>` is intruding in to the space that the content or `<Dd>` should occupy when a long string is passed to `<Dt>`.
image

### Current behaviour

image

### Checklist

### Testing instructions

1. npm start
2. http://localhost:9001/?path=/story/design-system-tile--tile-with-definition-list-with-custom-width

### DIFF
--- a/src/components/definition-list/definition-list.style.js
+++ b/src/components/definition-list/definition-list.style.js
@@ -18,20 +18,21 @@ export const StyledDtDiv = styled.div`
   ${({ w, dtTextAlign }) => css`
     text-align: ${dtTextAlign};
     width: ${w}%;
-    white-space: nowrap;
   `}
 `;
 
 export const StyledDdDiv = styled.div`
   ${({ ddTextAlign }) => css`
     text-align: ${ddTextAlign};
-    width: auto;
-    white-space: nowrap;
+    width: inherit;
+    width: -moz-available;
+    width: -webkit-fill-available;
   `}
 `;
 
 export const StyledDt = styled.dt`
   ${space}
+  margin-bottom: 0px;
   ${({ theme }) => css`
     font-size: 14px
     font-weight: 700;
@@ -45,6 +46,7 @@ StyledDt.defaultProps = {
 
 export const StyledDd = styled.dd`
   ${space}
+  margin-bottom: 0px;
   ${({ theme }) => css`
     font-size: 14px
     font-weight: 700;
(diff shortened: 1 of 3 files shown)


src/components/definition-list/definition-list.style.js
```diff
@@ -18,15 +18,15 @@ export const StyledDtDiv = styled.div`
   ${({ w, dtTextAlign }) => css`
     text-align: ${dtTextAlign};
     width: ${w}%;
-    white-space: nowrap;
+    white-space: normal;
```
jamime, line 21: `normal` is the default value for `white-space`. Could we simply remove this rule?

sjohnson-sage:
Unfortunately this change can't go in like this because it means the terms and descriptions (`Dt` and `Dd` tags) move out of line on a smaller screen as in the below screenshot. This was the reason that `nowrap` was added in the first place:

src/components/definition-list/definition-list.style.js
```diff
@@ -18,20 +18,18 @@ export const StyledDtDiv = styled.div`
...
   ${({ ddTextAlign }) => css`
     text-align: ${ddTextAlign};
-    width: auto;
-    white-space: nowrap;
+    width: -webkit-fill-available;
   `}
 `;
 
 export const StyledDt = styled.dt`
-  ${space}
```
sjohnson-sage, line 34: This is CSS just specific to your application and can't just be overwritten in Carbon like this, and `${space}` is used to add the styled-system spacing props so it can't just be taken out!

src/components/definition-list/definition-list.style.js
```diff
@@ -44,7 +42,6 @@ StyledDt.defaultProps = {
 };
 
 export const StyledDd = styled.dd`
-  ${space}
```
sjohnson-sage, line 47: Same here this can't be removed

src/components/definition-list/definition-list.style.js
```diff
@@ -18,20 +18,19 @@ export const StyledDtDiv = styled.div`
...
-    white-space: nowrap;
   `}
 `;
 
 export const StyledDdDiv = styled.div`
   ${({ ddTextAlign }) => css`
     text-align: ${ddTextAlign};
-    width: auto;
-    white-space: nowrap;
+    width: -webkit-fill-available;
```
jamime, line 27: This would only work for browsers that use `webkit`.

edleeks87:
@arvig33 @hegou33 I have a ticket in flight to update this component to support it's use outside of the `Tile`, I think it might be better if I also include the fix for the label wrapping in that work as well so we don't end of creating conflicts

edleeks87:
closing as fix applied as part of https://github.com/Sage/carbon/pull/3242

PR #3242, opened 2020-09-25
fix(definition-list): display as grid to enable content wrapping FE-3160

edleeks87:
### Proposed behaviour

Updates `DefinitionList` to `display: grid` in order to support the component being both responsive whilst also maintain alignment between  associated `Dt` and `Dd` components when the content wraps.

Adds an override to ensure the `Accordion` component has dynamic height when rendering a `DefinitionList` as a child.

Adds some examples of the component used inside an `Accordion` and in a `Grid`.
### Current behaviour

The original implementation used a single dimension to align the `Dt` and `Dd` children in columns
which caused issues with aligning them when some content wrapped. 

### Checklist

### Testing instructions

http://localhost:9001/?path=/story/design-system-grid--subgrid
http://localhost:9001/?path=/story/design-system-tile--tile-with-definition-list-default
http://localhost:9001/?path=/story/design-system-tile--tile-with-definition-list-with-custom-width
http://localhost:9001/?path=/story/design-system-accordion--with-a-definition-list

### DIFF
--- a/src/components/accordion/accordion.spec.js
+++ b/src/components/accordion/accordion.spec.js
@@ -140,6 +140,49 @@ describe('Accordion', () => {
       wrapper.update();
       isCollapsed(wrapper);
     });
+
+    describe('when window resizes', () => {
+      beforeEach(() => {
+        wrapper = mount(
+          <AccordionGroup>
+            <Accordion title='Title_1' defaultExpanded>
+              <div>Foo</div>
+            </Accordion>
+          </AccordionGroup>
+        );
+      });
+
+      it('recalculates the content height', () => {
+        act(() => render({ expanded: true }));
+        wrapper.update();
+        isExpanded(wrapper);
+
+        assertStyleMatch({
+          maxHeight: `${contentHeight}px`
+        }, wrapper.find(StyledAccordionContentContainer));
+
+        const newContentHeight = 400;
+
+        jest.spyOn(
+          wrapper.find(StyledAccordionContent).getDOMNode(), 'scrollHeight', 'get'
+        ).mockImplementation(() => newContentHeight);
+        act(() => {
+          global.innerWidth = 500;
+          global.innerHeight = 500;
+
+          global.dispatchEvent(new Event('resize'));
+        });
+        wrapper.update();
+
+        assertStyleMatch({
+          maxHeight: `${newContentHeight}px`
+        }, wrapper.find(StyledAccordionContentContainer));
+      });
+
+      afterEach(() => {
+        wrapper.unmount();
+      });
+    });
   });
 
   describe('layout', () => {
--- a/src/components/accordion/accordion.style.js
+++ b/src/components/accordion/accordion.style.js
@@ -82,9 +82,14 @@ const StyledAccordionContentContainer = styled.div`
   box-sizing: border-box;
   overflow: hidden;
   transition: all 0.3s;
-  max-height: ${({ maxHeight }) => maxHeight}px;
-  height: ${({ maxHeight }) => maxHeight}px;
-  ${({ isExpanded }) => !isExpanded && 'visibility: hidden'};
+  ${({ maxHeight, isExpanded }) => css`
+    max-height: ${isExpanded ? `${maxHeight}px` : '0px'};
+    height: ${isExpanded ? `${maxHeight}px` : '0px'};
+    
+    ${!isExpanded && `
+      visibility: hidden;
+    `}
+  `}
 `;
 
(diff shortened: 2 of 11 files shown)


jamime (commented):
I've not reviewed the code, but it appears that the accordion no longer animates in the definition list example.

edleeks87:
> I've not reviewed the code, but it appears that the accordion no longer animates in the definition list example.

So it does good spot, I'll come up with a different approach

grabkowski (approved):

src/components/accordion/accordion.style.js
```diff
@@ -82,9 +82,15 @@ const StyledAccordionContentContainer = styled.div`
   box-sizing: border-box;
   overflow: hidden;
   transition: all 0.3s;
-  max-height: ${({ maxHeight }) => maxHeight}px;
-  height: ${({ maxHeight }) => maxHeight}px;
-  ${({ isExpanded }) => !isExpanded && 'visibility: hidden'};
+  ${({ maxHeight }) => css`
```
jamime, line 85: Please can we change this. Currently a closed accordion will output two CSS rules for `max-height` and `height`.

src/components/definition-list/dd.component.js
```diff
@@ -5,11 +5,13 @@ import { StyledDd } from './definition-list.style';
...
 const Dd = ({
   mb = 2,
-  children
+  children,
+  align = 'right'
 }) => {
   return (
     <StyledDd
       mb={ mb }
+      align={ align }
```
jamime, line 14: This doesn't appear to be used anywhere. Please remove.

src/components/definition-list/dl.component.js
```diff
@@ -4,27 +4,48 @@ import PropTypes from 'prop-types';
...
+      ddContent.push(child);
+    }
 
-      <StyledDtDiv w={ w } dtTextAlign={ dtTextAlign }>
-        {dtLabels}
-      </StyledDtDiv>
+    if (dtLabel && (index === (listChildren.length - 1) || listChildren[index + 1].type === Dt)) {
+      const key = child.props.key || `dl_row_${createGuid()}}`;
+      dlComponent.push(
+        <React.Fragment key={ key }>
```
jamime, line 31: `createGuid` has the same problem as using `index` in this case, as the order doesn't change, it's fine to use `index`.

https://reactjs.org/docs/lists-and-keys.html#keys

```suggestion
        <React.Fragment key={ child.props.key || index }>
```

jamime (changes requested):

src/components/definition-list/dl.component.js
edleeks87: ok thanks

src/components/definition-list/dd.component.js
edleeks87: Consider it done

src/components/accordion/accordion.style.js
edleeks87: Good point will update

src/components/accordion/accordion.spec.js
```diff
@@ -140,6 +140,44 @@ describe('Accordion', () => {
...
+            </Accordion>
+          </AccordionGroup>
+        );
+      });
+
+      it('recalculates the content height', () => {
+        act(() => render({ expanded: true }));
+        wrapper.update();
+        isExpanded(wrapper);
+
```
jamime, line 159: lets add an assertion here for the height before resize.

src/components/definition-list/dl.component.js
```diff
@@ -11,20 +11,39 @@ const Dl = ({
...
+      dtLabel = child;
+    }
+    if (child.type === Dd) {
+      ddContent.push(child);
+    }
 
-      <StyledDtDiv w={ w } dtTextAlign={ dtTextAlign }>
-        {dtLabels}
-      </StyledDtDiv>
+    if (dtLabel && (index === (listChildren.length - 1) || listChildren[index + 1].type === Dt)) {
```
jamime, line 27:
```suggestion
    const isLastChild = index === (listChildren.length - 1);
    const nextItemIsDt = listChildren[index + 1].type === Dt;
    if (dtLabel && (nextItemIsDt || isLastChild)) {
```

jamime (changes requested):

grabkowski (dismissed):

jamime (dismissed):

DlgSHi (approved):

DlgSHi (approved):

jamime (approved):

Issue #3225, opened 2020-09-17
Control the margin/padding of somes components


jerlau:
## Feature request
We want to align text in several component so we need to have the controle of the margin or padding.
We also want to control the margin in the DialogFullScreen component between the heading and the children.

### Current behaviour

We don't have the control of the padding/margin of somes component.

> in red: what we want to remove
> in green/.blue: what we want to align

![image](https://user-images.githubusercontent.com/17121053/93493458-373eae80-f90c-11ea-84bf-acc31e25a330.png)

### Desired behaviour

1. Control the padding/margin of the content of a dialog full screen
2. Control the padding/margin of the title of the drawer to align with the title of the dialog full screen
3. Control maybe the padding/margin of the title of a dialog full screen.

### Suggested solution

add parameters to the concerned component

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

sjohnson-sage:
FE-3152

PR #3247, opened 2020-09-29
feat(tabs, tab): add style-system - FE-3152

dawidzarzycki:
### Proposed behaviour
Add possibility to manage paddings and margins on Tab and Tabs components.
It won't cover whole github (Dialog Full Screen, Drawer) components that was listed in the issue but it fixes things that will allow to create correct design (Tab, Tabs)

### Current behaviour
There is no possibility to manage margins and paddings

### Checklist

- [ ] <del> Unit tests added or updated if required

### Additional context
Github issue: https://github.com/Sage/carbon/issues/3225


edleeks87:
@dawidzarzycki I haven't done a full review yet but I'm not sure we want to pass the spacing props any further down then the main Tabs component (ie not to the `Tab`, `TabTitle` or `TabsHeader`) as that will effect the internal alignment rather than the alignment of the whole component

PR #3255, opened 2020-10-06
fix(validation): renders tooltip in 'top' position when validation is on label and label is inline

### LINKED ISSUE #3191: Validation tooltip position
### Current behaviour
Validation tooltips are positioned to the right of the validation icon, when using `labelInline` and `validationOnLabel` this means the tooltip obscures the input.
image

### Desired behaviour
Tooltip should dynamically position itself. 

### Suggested solution
Should we use `validationOnLabel={false}`?

### Describe alternatives you've considered
`validationOnLabel` would not match the designs provided.

### Additional context
N/A

### Your environment

| carbon-react     |39.0.0

nicktitchmarsh:
Fixes #3191, Fixes FE-3100

### Proposed behaviour
Tooltip now renders in position top when validationOnLabel prop is true and labelInline prop is also true

image

### Current behaviour
Tooltip renders over the top of an input when validationOnLabel prop is true and labelInline prop is also true
image

https://codesandbox.io/s/suspicious-kirch-zpv6q?file=/src/index.js

### Checklist

### Testing instructions
Storybook now includes a examples of inline label validation in the textbox component story


src/__experimental__/components/label/label.component.js
```diff
@@ -14,6 +14,12 @@ const shouldDisplayValidationIcon = ({
   return disabled ? false : typeof validation === 'string';
 };
 
+const shouldDisplayInlineTooltip = ({
+  error, warning, info, inline
+}) => {
+  return (error || warning || info) && inline === true;
```
jamime, line 20: We're returning a boolean here to then convert it to `top`/`right` below. We can simplify this by returning the tooltip position here.

```js
const tooltipPosition = ({
  error, warning, info, inline
}) => {
  return (error || warning || info) && inline === true ? 'top' :'right';
};

```

src/__experimental__/components/label/label.spec.js
```diff
@@ -235,6 +235,14 @@ describe('Label', () => {
       expect(icon.exists()).toEqual(false);
     });
   });
+
+  describe.each(validationTypes)('when %s prop is passed as string and label is inline', (vType) => {
+    it('tooltip should render in position "top"', () => {
+      const wrapper = render({ [vType]: 'Message', useValidationIcon: true, inline: true }, mount);
+      const icon = wrapper.find(ValidationIcon);
+      expect(icon.prop('tooltipPosition')).toBe('top');
```
jamime, line 243: As there is no test in this suite that checks for `tooltipPosition: right` this suite would pass if the tooltip incorrectly was always displayed on top.

We should add one that asserts when `inline:false` that `tooltipPosition: right`.

jamime (changes requested):

src/__experimental__/components/label/label.component.js
nicktitchmarsh: Simplified this so the function returns the tooltip position.

src/__experimental__/components/label/label.spec.js
nicktitchmarsh: Updated to include negative test case

jamime (approved):

jamime (approved):

DipperTheDan (approved):

DlgSHi (approved):

PR #3285, opened 2020-10-13
feat(dialogfullscreen): allow content in headers

sjohnson-sage:
### Proposed behaviour
Add left and right adaptive paddings to DialogFullScreen.

Add new `headerChildren` prop to allow the user to pass in components be be rendered to the right of the heading in `DialogFullScreen`. This content will wrap underneath the heading when there isn't enough room on the screen.

image

### Current behaviour
Above functionality does not exist.

### Checklist

### Testing instructions
Responsive paddings: Look at the default story on DialogFullScreen and resize the window.

Header children: see the "with header children" story in storybook


src/components/pages/pages.stories.js
```diff
@@ -111,9 +111,10 @@ function makeStory(name, themeSelector, disableChromatic = false) {
...
             open={ store.get('open') }
             onCancel={ handleCancel }
+            pagesStyling
           >
             <PageState>
-              <DefaultPages
+              <Pages
                 pageIndex={ handleSlide(null, pageIndex()) }
               >
                 <Page title={ <Heading title='My First Page' /> }>
```
edleeks87, line 120: Will `onCancel` prop not need to be passed as it's required in `IconButton`?

src/__internal__/full-screen-heading/full-screen-heading.component.js
```diff
@@ -0,0 +1,50 @@
...
+    onCancel,
+    ...otherProps
+  } = props;
+
+  const closeIcon = () => {
+    if (!showCloseIcon || !onCancel) return null;
+    return (
+      <IconButton
+        data-element='close'
+        onAction={ onCancel }
```
edleeks87, line 21: This prop is required in `IconButton` but you might end up passing `undefined` here

src/__internal__/full-screen-heading/full-screen-heading.component.js
```diff
@@ -0,0 +1,50 @@
...
+      >
+        <Icon type='close' />
+      </IconButton>
+    );
+  };
+
+  return (
+    <StyledFullScreenHeading
+      { ...otherProps }
+      { ...tagComponent('full-screen-heading', props) }
```
dawidzarzycki, line 31: Do we need this `tagComponent` function instead of just `data-component`?

sjohnson-sage: I didn't write this, just moved it but I think it does in case `data-element` or `data-role` are ever passed as props to `FullScreenHeading`

src/components/pages/pages.stories.js
sjohnson-sage: There's where the issue with Pages that I mentioned is, it doesn't actually render the close button itself - the `DialogFullScreen` wrapping this has `onCancel` set and that's where it can be closed.
`Pages` doesn't even have an `onCancel` prop, which is part of what's going to have to change when this component is re-written IMO

src/__internal__/full-screen-heading/full-screen-heading.component.js
sjohnson-sage: I'll put something defensive in here to stop that from happening

sjohnson-sage: Actually the line above `if (!showCloseIcon || !onCancel) return null;` already does that, it's not possible to pass `undefined` through here

edleeks87: Looks like it's checking the prop type erroneously then, as this component is going to be refactored anyway we might as well just look to fix it then as the issue is fairly trivial

edleeks87 (approved):

dawidzarzycki (dismissed):

dawidzarzycki (dismissed):

edleeks87 (dismissed):

KatarzynaQA (approved):

KatarzynaQA (approved):

edleeks87 (approved):

edleeks87 (approved):

KatarzynaQA (approved):

dawidzarzycki (approved):

PR #3299, opened 2020-10-16
fix(tabs): tabs position left in sidebar and remove set target callback

### LINKED ISSUE #3214: Tabs inside Drawer component doesn't work correctly on v40.0.0
### Current behaviour
After update carbon version to 40.0.0.
Tabs inside Drawer component doesn't work correctly :
we got the error _setTarget is not a function when we call handleChange.
We don't have this problem in the version 39.2.1
You can test that in the sandbox  https://codesandbox.io/s/busy-chebyshev-fkbnc?file=/src/index.js.

### Expected behaviour
Tabs inside Drawer component has no regression : no "_setTarget  error" raised.

### Reproducible example
https://codesandbox.io/s/busy-chebyshev-fkbnc?file=/src/index.js
Use this sandbox and change carbon version : 
v39.2.1 : ok
v40.0.0 : ko

### Your environment

| carbon-react     | 40.0.0
| React            | 16.13.1
| Browser          | Chrome
| npm              | 6.14.2
| Operating System | Windows

edleeks87:
### Proposed behaviour

Removes the `setTarget` callback from the context as same functionality can be achieved using
`onTabChange` callback, defaults `Tabs` to position left when in sidebar

BREAKING CHANGE: `setTarget` callback is no longer available, `onTabChange` should be used instead

fix #3214

### Current behaviour

Currently we have both a `setTarget` and `onTabChange` callbacks that achieve the same functionality
 
### Checklist

### Testing instructions

`https://carbon.sage.com/?path=/story/design-system-drawer--with-tab-controls` behaviour should be the same as seen here


src/components/tabs/__internal__/tab-title/tab-title.component.js
```diff
@@ -25,7 +25,7 @@ const TabTitle = React.forwardRef(({
   noLeftBorder = false,
   noRightBorder = false,
   customLayout,
-  hasCustomTarget,
+  isInSidebar,
```
sjohnson-sage, line 28: Just double-checking - can this only be used in Sidebar, or could it potentially be in other components as well?

sjohnson-sage (approved):

src/components/tabs/__internal__/tab-title/tab-title.component.js
edleeks87: Yeah it is just to apply a style override specific to when it is in the Drawer's sidebar

nicktitchmarsh (approved):

DlgSHi (approved):

PR #3347, opened 2020-11-02
feat(navigation-bar): add support for custom spacing props

edleeks87:
### Proposed behaviour

Adds support for the `styled-system` spacing props to be used to apply custom paddings and margins
to the `NavigationBar`.

### Current behaviour

No support for custom spacing in `NavigationBar`

### Checklist

### Testing instructions

A basic example has been added here http://localhost:9001/?path=/story/design-system-navigation-bar--with-custom-spacing

Codesandbox:
https://codesandbox.io/s/carbon-quickstart-forked-rlez8?file=/src/index.js

### DIFF
--- a/src/components/navigation-bar/navigation-bar.component.js
+++ b/src/components/navigation-bar/navigation-bar.component.js
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import propTypes from "@styled-system/prop-types";
 import AppWrapper from "../app-wrapper/app-wrapper";
 import StyledNavigationBar from "./navigation-bar.style";
 
@@ -8,6 +9,7 @@ const NavigationBar = ({
   isLoading = false,
   children,
   ariaLabel,
+  ...props
 }) => {
   return (
     <StyledNavigationBar
@@ -15,6 +17,7 @@ const NavigationBar = ({
       aria-label={ariaLabel}
       navigationType={navigationType}
       data-component="navigation-bar"
+      {...props}
     >
       <AppWrapper className="carbon-navigation-bar__content">
         {!isLoading && children}
@@ -24,6 +27,8 @@ const NavigationBar = ({
 };
 
 NavigationBar.propTypes = {
+  /** Styled system spacing props */
+  ...propTypes.space,
   children: PropTypes.node,
   ariaLabel: PropTypes.string,
   /** color scheme of navigation component */
--- a/src/components/navigation-bar/navigation-bar.d.ts
+++ b/src/components/navigation-bar/navigation-bar.d.ts
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import NavigationBar from './navigation-bar.component';
+import { SpacingProps } from '../../utils/helpers/options-helper';
 
-export interface NavigationBarProp {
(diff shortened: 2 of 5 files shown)


grabkowski (dismissed):

src/components/navigation-bar/navigation-bar.component.js
```diff
@@ -1,29 +1,42 @@
...
+    }
+  });
+
   return (
     <StyledNavigationBar
       role='navigation'
       aria-label={ ariaLabel }
       navigationType={ navigationType }
       data-component='navigation-bar'
+      { ...(Object.keys(spacingProps).length && spacingProps) }
```
jamime, line 28: Because `StyledNavigationBar` is the final component I think we could simply spread `...props` into the component like we have elsewhere. What do you think?

If we choose to keep these props separate, there is no need to check the length of `spacingProps`, we can always spread it because it's always an object.

src/components/navigation-bar/navigation-bar.component.js
```diff
@@ -1,29 +1,42 @@
...
   navigationType = 'light',
   isLoading = false,
   children,
-  ariaLabel
+  ariaLabel,
+  ...props
 }) => {
+  const spacingProps = {};
+
+  Object.keys(props).forEach((propName) => {
```
jamime, line 16: See my comment below.
If we choose to keep these props separate I think we can convey the intention better by writing this as a reducer.

```js
const spacingProps = Object.keys(props).reduce((acc, propName) => {
      acc[propName] = props[propName];
      return acc;
    }, {});
```

jamime (changes requested):
Code coverage is currently failing, the easiest way to fix this would be to simplify the logic and spread the props into `StyledNavigationBar`.

src/components/navigation-bar/navigation-bar.component.js
edleeks87: The main reason I was against just spreading `...props` was to avoid inadvertently providing a mechanism for consumers passing in something like `style={{...}}`

src/components/navigation-bar/navigation-bar.component.js
edleeks87: Yeah I think separating them is worthwhile for the above reason, but using a reducer does convey it a little better so I'll update it

src/components/navigation-bar/navigation-bar.component.js
jamime: Ah, we use that so users can add any `aria-*`, `data-*`, `as` props that they might require. If we're going to start conditionally adding props maybe we should make a util that caters for this (or stick with the current convention of passing everything).

edleeks87: Yeah that's a fair point, as we've already adopted the spreading of all props into the root I'll just go with that as well to be consistent

jamime (dismissed):

jamime (approved):

jamime (approved):

grabkowski (approved):

KatarzynaQA (dismissed):

grabkowski (dismissed):

KatarzynaQA (approved):

grabkowski (approved):

PR #3354, opened 2020-11-05
feat(select): add option group headers

sjohnson-sage:
### Proposed behaviour
Add ability to add option headers to the Select component: 
image

They have an optional icon, and are slightly greyed out to indicate that they're not options

### Current behaviour
It is not currently possible in Carbon to have option headers in Select

### Checklist

### Testing instructions
Go to story http://localhost:9001/?path=/docs/design-system-select--option-groups


src/components/select/simple-select/simple-select.component.js
```diff
@@ -50,6 +50,12 @@ const SimpleSelect = React.forwardRef(
     const [textValue, setTextValue] = useState("");
     const [selectedValue, setSelectedValue] = useState("");
 
+    const getChildOptions = (allChildren) => {
+      return React.Children.toArray(allChildren).filter(
+        (child) => !child.props.label
```
mkrds, line 55: `(child) => child.type === Option`

src/components/select/option-group-header/option-group-header.style.js
```diff
@@ -0,0 +1,45 @@
...
+
+const StyledOptionGroupHeader = styled.div`
+  height: 26px;
+  margin: 16px 0 0;
+  padding-left: 16px;
+  display: flex;
+  align-items: center;
+
+  &:focus {
+    background-color: green;
```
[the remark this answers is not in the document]
sjohnson-sage: Ah oops that was me testing something and forgot to take it out

src/components/select/select-list/select-list.component.js
```diff
@@ -34,9 +34,33 @@ const SelectList = React.forwardRef(
...
+          childOptionIndex
+        ];
+        let currentOptionValue;
+
+        if (currentSelectedOption) {
+          currentOptionValue = currentSelectedOption.props.value;
+        }
+
+        return React.Children.toArray(children).findIndex(
+          (child) => child.props.value === currentOptionValue
```
edleeks87, line 55: Could `child.props.value` not be an object?

edleeks87: Although I see below there are other places where it's being treated as if it's only a string so maybe the work to update for objects as well should be done separately

src/components/select/select-list/select-list.component.js
```diff
@@ -34,9 +34,33 @@ const SelectList = React.forwardRef(
     const listRef = useRef();
     const listActionButtonRef = useRef();
 
+    const getChildOptions = (allChildren) => {
+      return React.Children.toArray(allChildren).filter(
+        (child) => !child.props.label
+      );
+    };
+
```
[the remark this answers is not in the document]
sjohnson-sage: I tried this previously and it doesn't work with all of the `useCallback` functions in this file, even when passing in `childOptions` as a dependancy of the callbacks, I've tried to find a way around that but I can only seem to get it to work by calling the `getChildOptions` function within the callbacks directly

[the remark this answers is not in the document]
sjohnson-sage: Unfortunately this doesn't work with all of the `useCallback` functions, even when `childOptions` is added to their dependancy lists. I'll do this for the other places though

src/components/select/select-list/select-list.component.js
sjohnson-sage: Yeah like you say, it's already being treated as a string, I think it's on Kuba's radar to make all this stuff work with it being an object as well, so this can be changed at the same time as everything else

src/components/select/select-list/select-list.component.js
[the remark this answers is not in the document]
edleeks87: You could save it using `usRef` and access it via `foo.current` but it's not a dealbreaker for me

src/components/select/select-list/select-list.component.js
edleeks87: Yeah makes sense

edleeks87 (approved):

mkrds (approved):

src/components/select/select-list/select-list.component.js
```diff
@@ -193,10 +230,30 @@ const SelectList = React.forwardRef(
...
+          const newProps = {
+            onSelect: handleSelect,
+            isHighlighted: currentOptionValue === child.props.value,
+          };
+
+          return React.cloneElement(child, newProps);
+        });
+      }
+
```
grabkowski, line 256: you don't have to create a separate list for children other than options, what you need is to use recurrence to check if the value under an index is an Option:
```
    const getNextHighlightableItemIndex = useCallback((key, indexOfHighlighted, lastIndex) => {
      let nextIndex = getNextIndexByKey(key, indexOfHighlighted, lastIndex);
      const nextElement = React.Children.toArray(children)[nextIndex];

      if (nextElement && nextElement.type !== Option) {
        nextIndex = getNextHighlightableItemIndex(key, nextIndex, lastIndex)
      }

      return nextIndex;
    }, [children]);
```
you only need to be sure that there is at least one Option to avoid infinite loop.

Doing that you could use the original index instead of relying on the option list.

grabkowski (changes requested):

grabkowski (changes requested):

edleeks87:
I haven't got anything else to add past what Kuba has raised already

src/components/select/select-list/select-list.component.js
```diff
@@ -34,11 +38,20 @@ const SelectList = React.forwardRef(
...
+    const getChildOptions = (allChildren) => {
+      return React.Children.toArray(allChildren);
+    };
+
+    const getNextHighlightableItemIndex = useCallback(
+      (key, indexOfHighlighted, lastIndex) => {
+        let nextIndex = getNextIndexByKey(key, indexOfHighlighted, lastIndex);
+        const nextElement = getChildOptions(children)[nextIndex];
+
+        if (nextElement && nextElement.type !== Option) {
```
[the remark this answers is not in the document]
sjohnson-sage: No it isn't, this check is for when the next element at `nextIndex` isn't an option, then it'll find the next option. `nextElement` could be any of the child as it just uses `getChildOptions`

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -191,82 +193,123 @@ You can use the `required` prop to indicate if the field is mandatory.
...
+  <Story name="option groups" parameters={{ info: { disable: true } }}>
+    <Select
+      name="optGroups"
+      id="optGroups"
+      onOpen={action("onOpen")}
+      onChange={action("onChange", { depth: 2 })}
+      onClick={action("onClick", { depth: 2 })}
+      onFocus={action("onFocus", { depth: 2 })}
+      onBlur={action("onBlur", { depth: 2 })}
+      onKeyDown={action("onKeyDown", { depth: 2 })}
```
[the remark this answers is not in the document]
sjohnson-sage: I meant to take those out but forgot!

edleeks87 (approved):

src/components/select/simple-select/simple-select.component.js
```diff
@@ -50,6 +54,10 @@ const SimpleSelect = React.forwardRef(
     const [textValue, setTextValue] = useState("");
     const [selectedValue, setSelectedValue] = useState("");
 
+    const getChildOptions = useCallback(() => {
+      return React.Children.toArray(children);
+    }, [children]);
+
```
mkrds, line 60: const childOptions = useMemo(() => getChildOptions(), [children])

grabkowski (approved):

mkrds (approved):

sianford:
Tested in Safari, Chrome, Firefox and Edge Chromium browsers. No issues found. All passed ✅

mkrds (approved):

edleeks87 (approved):

sianford (approved):

grabkowski (approved):

PR #3366, opened 2020-11-09
feat(select): async select loader (FE-2647)

grabkowski:
### Proposed behaviour
Screenshot 2020-11-09 at 18 26 59
Add a loader to handle async requests on select open for Simple and Filterable Selects.
Fix warning for incorrect PropTypes when an object is passed as Option value (FE-3266).
Add exports for Select PropTypes.

### Checklist

### Testing instructions
http://localhost:9001/?path=/story/design-system-select--asynchronous
http://localhost:9001/?path=/story/design-system-select-filterable--asynchronous
http://localhost:9001/?path=/story/design-system-select--with-object-as-value


src/components/select/filterable-select/index.d.ts
```diff
@@ -57,6 +57,8 @@ export interface FilterableSelectProps {
   onListAction?: () => void;
   /** Flag to configure component as mandatory */
   required?: boolean;
+  /** If true the loader animation is displayed in the option list */
+  isLoading: boolean;
```
edleeks87, line 61:
```suggestion
  isLoading?: boolean;
```

src/components/select/simple-select/index.d.ts
```diff
@@ -55,6 +55,8 @@ export interface SimpleSelectProps extends SpacingProps {
   onKeyDown?: () => void;
   /** Flag to configure component as mandatory */
   required?: boolean;
+  /** If true the loader animation is displayed in the option list */
+  isLoading: boolean;
```
edleeks87, line 59:
```suggestion
  isLoading?: boolean;
```

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -203,12 +204,15 @@ const FilterableSelect = React.forwardRef(
         onChageMissingMessage
```
edleeks87, line 204: `onChageMissingMessage` -> `onChangeMissingMessage`

src/components/select/simple-select/simple-select.component.js
```diff
@@ -425,6 +427,8 @@ SimpleSelect.propTypes = {
   transparent: PropTypes.bool,
   /** A custom callback for when the dropdown menu opens */
   onOpen: PropTypes.func,
+  /** If true the loader animation is displayed in the option list */
+  isLoading: PropTypes.bool,
```
DlgSHi, line 431:
```suggestion
  isLoading?: PropTypes.bool,
```

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -210,6 +210,59 @@ A custom `Button` Component could be passed as the `listActionButton` value.
   </Story>
 </Preview>
 
+### Asynchronous:
+
+<Preview>
+  <Story name="asynchronous" parameters={{ chromatic: { disable: true }}} > 
+    {() => {
+      let isInitial = useRef(true);
```
jamime, line 218: I think we can simplify this by using `useState` instead of `useRef`? Same in the simple story.

src/components/select/filterable-select/filterable-select.stories.mdx
```diff
@@ -210,6 +210,59 @@ A custom `Button` Component could be passed as the `listActionButton` value.
   </Story>
 </Preview>
 
+### Asynchronous:
+
```
jamime, line 214: Please an we add a text description explaining when the use this pattern. There's always a lot of confusion around lazy-loading/infinite loading.

We should make it very clear which use case it is that we're supporting. I assume this is only for loading the items on open of dropdown, but not to load more options when the input changes? Equally, it shouldn't be used to load more items when the user scrolls?

Whatever the intended usage is, we should document it, please do the same for the simple story.

src/components/select/select-list/select-list.component.js
```diff
@@ -196,20 +199,38 @@ const SelectList = React.forwardRef(
...
 
         return React.cloneElement(child, newProps);
       });
-    }
 
-    function handleSelect(optionData) {
-      onSelect({ ...optionData, selectionType: "click" });
-    }
+      if (isLoading) {
+        optionsList.push(
```
jamime, line 225: If we return here instead of pushing to the `optionList` we don't need to support `hidden` on the option component.

If we do that we can:
- skip the cloning of all the children elements that are set to `display:none`.
- remove the code that allows an option to be hidden.

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -127,6 +127,101 @@ You can use the `required` prop to indicate if the field is mandatory.
   </Story>
 </Preview>
 
+### With object as value:
+
```
jamime, line 131: We should also add a description here explaining why you would want to use an option.

jamime (changes requested):

src/components/select/select-list/select-list.component.js
grabkowski: by doing that I wanted to avoid deselection of previously set option, you're right, it will not affect the children prop, so it won't trigger a re-render of the parent.

grabkowski: the loader will have to be pushed as the last element when the lazy loading is implemented. I could remove the `hidden` prop and clear the list when that prop is set to true.

src/components/select/simple-select/simple-select.component.js
grabkowski: it's not TypeScript

src/components/select/filterable-select/filterable-select.stories.mdx
grabkowski: State should be used only when something is being passed down through a prop, as modification of it's value will cause re-render. I'll leave it as it is, using state here might be a bad practice.

src/components/select/filterable-select/filterable-select.stories.mdx
grabkowski: right now the lazy loading is not yet implemented, it's part of another task. When the other task done I'll add a description on which one should be used.

src/components/select/simple-select/simple-select.stories.mdx
```diff
@@ -127,6 +127,105 @@ You can use the `required` prop to indicate if the field is mandatory.
...
+        </div>
+      );
+    }}
+  </Story>
+</Preview>
+
+### With isLoading prop:
+When `isLoading` prop is passed, a loader will be appended at the end of the Select List. That functionality could be used to load the options asynchronously.
+
+This prop should only be used when the dropdown is expanded.
```
jamime, line 176:
```suggestion
```

jamime (changes requested):

src/components/select/simple-select/simple-select.stories.mdx
grabkowski: if the dropdown is not expanded this prop will do nothing

edleeks87 (dismissed):

edleeks87 (dismissed):

jamime (dismissed):

jamime (dismissed):

edleeks87 (approved):

edleeks87 (approved):

jamime (approved):

DlgSHi (approved):

PR #3371, opened 2020-11-11
refactor: fix storybook warnings

sjohnson-sage:
### Proposed behaviour
Fix storybook warnings shown below. A couple of incorrect imports from `styled-system` and `Link` was being imported from the wrong package

The `Link` export was moved from `react-router` to `react-router-dom` from v4 onwards

### Current behaviour
image

### Checklist

### Testing instructions
`npm start` and check for any warnings

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "react-dnd-test-utils": "^10.0.2",
     "react-dom": "^16.12.0",
     "react-highlight": "briancappello/react-highlight#react-v16-compiled",
-    "react-router": "^5.2.0",
+    "react-router-dom": "^5.2.0",
     "react-test-renderer": "^16.12.0",
     "rimraf": "^2.6.3",
     "sass-loader": "7.0.3",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -158,6 +158,7 @@
     "react-dnd-html5-backend": "^10.0.2",
     "react-dnd-legacy": "npm:react-dnd@^2.6.0",
     "react-dnd-touch-backend": "0.2.7",
+    "react-router-dom": "^5.2.0",
```
edleeks87, line 161: I think we can remove `react-router` as well

sjohnson-sage: Very good point

sjohnson-sage: I've also accidentally added it to dependancies rather than devDependancies as well so I'll move that

edleeks87: I'm also not sure if we should have `react-router-dom` as a dependency of carbon just so we can demo it in storybook but happy to be convinced otherwise

edleeks87 (approved):

DipperTheDan (approved):

PR #3388, opened 2020-11-16
 feat(action-popover): add support to place menu above button  FE-2956

### LINKED ISSUE #3072: Flat Table: We cannot mix hasStickyHead and pagination
### Current behaviour
An empty line appears (in yellow) at the end of the list

image

### Expected behaviour
This empty line must be removed

### Reproducible example
https://codesandbox.io/s/cranky-cloud-j5rpt?file=/src/index.js:3915-3928

### Your environment

| carbon-react     | 33.3.0
| carbon-factory   | 11.0.1 (none in sandbox)
| react-scripts    | 3.0.1
| React            | 16.8.6(sandbox) and 16.13.1 (client managment)
| Browser          | Chrome Version 83.0.4103.116 (Build officiel) (64 bits)
| npm              | 6.13.7
| Operating System | windows 10

edleeks87:
### Proposed behaviour

Allow `ActionPopoverMenu` to open above the menu button when the `placement` prop is set to "top"

### Current behaviour

There is no prop to support opening the menu above the button
### Checklist

### Additional context

fix #3072

The menu's are positioned absolutely which causes the overflow to account for the space they take up.

### Testing instructions

http://localhost:9001/?path=/story/design-system-flat-table--paginated-with-sticky-header

image

### DIFF
--- a/src/components/action-popover/action-popover-menu.component.js
+++ b/src/components/action-popover/action-popover-menu.component.js
@@ -19,6 +19,7 @@ const ActionPopoverMenu = React.forwardRef(
       setOpen,
       setFocusIndex,
       setItems,
+      placement = "bottom",
       ...rest
     },
     ref
@@ -148,6 +149,7 @@ const ActionPopoverMenu = React.forwardRef(
             // callback and index to update focusIndex if item hovered and has submenu
             const itemWithRef = React.cloneElement(child, {
               ref: React.createRef(),
+              placement: child.props.submenu ? placement : undefined,
             });
             itemsWithRef.push(itemWithRef);
 
@@ -160,7 +162,15 @@ const ActionPopoverMenu = React.forwardRef(
       // so we use a smaller array which makes keyboard navigation easier as we don't have to filter it on every
       // keypress
       setItems(itemsWithRef);
-    }, [children, focusIndex, isOpen, menuID, setFocusIndex, setItems]);
+    }, [
+      children,
+      placement,
+      focusIndex,
+      isOpen,
+      menuID,
+      setFocusIndex,
+      setItems,
+    ]);
 
     useEffect(() => {
       if (isOpen && focusIndex !== null) {
@@ -179,6 +189,8 @@ const ActionPopoverMenu = React.forwardRef(
         aria-labelledby={parentID}
         role="menu"
         ref={ref}
+        placement={placement}
+        buttonHeight={getButtonHeight(button)}
         {...rest}
       >
         {childrenWithRef}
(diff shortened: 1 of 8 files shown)


src/components/action-popover/action-popover-menu.component.js
```diff
@@ -187,6 +199,22 @@ const ActionPopoverMenu = React.forwardRef(
   }
 );
 
+function checkRef(ref) {
+  return Boolean(ref && ref.current);
+}
+
+function getButtonHeight(buttonRef) {
+  // if no ref return height of default button
+  if (!checkRef(buttonRef)) {
```
jamime, line 208: I think we need to look for a long term solution, we have a lot of similar code.

Something like this https://reach.tech/rect/ what do you think?

src/components/action-popover/action-popover-menu.component.js
```diff
@@ -179,6 +189,8 @@ const ActionPopoverMenu = React.forwardRef(
         aria-labelledby={parentID}
         role="menu"
         ref={ref}
+        placement={placement}
+        buttonHeight={getButtonHeight(button)}
```
jamime, line 193: `buttonHeight` isn't a prop  on the `Menu` component, we'll need to document it.

jamime (changes requested):

src/components/action-popover/action-popover-menu.component.js
edleeks87: Ah yeah good point I missed that

edleeks87: The prop is internal to the `Menu` which is a `styled-component`

src/components/action-popover/action-popover-menu.component.js
edleeks87: Yeah I think something like this would definitely be worth it as we have a number of examples where we've had to default when the ref is null on first render

src/components/flat-table/flat-table.stories.mdx
```diff
@@ -865,6 +865,194 @@ Balance performance concerns with the convenience of the user being able to see
   </Story>
 </Preview>
 
+### Paginated with sticky header
```
ianforster-sage, line 868: Please could you add a paragraph here that explains the ActionPopover behaviour in row 5. As it is, you'd have to really look for it to see that this is possible.

ianforster-sage (changes requested):

ianforster-sage (approved):

jamime (dismissed):

ianforster-sage (dismissed):

ianforster-sage (dismissed):

jamime (dismissed):

ianforster-sage (dismissed):

sianford:
Tested in Safari, Chrome, Firefox and Edge Chromium. No gap exists between the last row of the table and the pagination area. Action Popover now appears above the row when the row is at or near the bottom of the table. Testing passed ✅

ianforster-sage (approved):

jamime (approved):

sianford (approved):

PR #3393, opened 2020-11-17
chore: add performance Profiler to the project

DlgSHi:
### Proposed behaviour

- add `<Profiler>` to the carbon;
- create a `decorator` to wrap all `<Story>` with the `performance` (add new Storybook 6+ version also commented until we will upgrade the Storybook to the proper version)
- all data is logged into the console.

Additionally:
- add the cypress-terminal-output to make it possible to store the values in the proper format (`.json`);
- add `performance.spec.js` 
- data is stored in the `./cypress/.logs/performance/performance.spec.json`;

[this](https://github.com/marketplace/actions/upload-file-action) could be used for uploading the file to Sumo

Alert
Alert

ButtonToggle
ButtonToggle

### Current behaviour

We need to consider of implementation for `performance` data for all project to track all `timing` for `components`/`stories`

### Checklist


nicktitchmarsh (commented):
We might want to look at how we are measuring performance on re-renders, if we should make a specific tests for performance. We should also look at recording the results in a Cypress Reporter to allow us to compare results between runs. But this is a good first version

https://docs.cypress.io/guides/tooling/reporters.html#Examples

.storybook/preview.js
```diff
@@ -59,4 +60,21 @@ if (!isChromatic()) {
...
+    console.log(`Start time: ${startTime}`);
+    console.log(`Commit time: ${commitTime}`);
+    console.log(`End of gathering performance data for ${id} component`);
+  };
+  return (
+    <Profiler id={id} onRender={callback}>
+      <Story />
+    </Profiler>
+  );
+});
```
mkrds, line 80: This code in `preview.js` won't work after storybook is updated to v6

jamime (commented):
Performance profiling is tricky, this PR does expose the metrics, but it's not possible to say if something has improved or got worse because theres no baseline.

I'm sure we can come up with some static values that we can use as a maximum or minimum but I think we're most interested in the trend over time and identifying when I changed component X, component Y degraded.

Changing `react-dom` and `scheduler` will have a negative performance impact for users visiting `carbon.sage.com`. Ideally we want to change this only when doing a performance profile.

I'd like us to investigate our options further and merge in a complete solution.

.storybook/preview.js
jamime: New API https://storybook.js.org/docs/react/writing-stories/decorators#global-decorators

PR #3408, opened 2020-11-18
chore(row): remove classic theme support - FE-3120/FE-3075

mkrds:
BREAKING CHANGE: Row and Column components no loger support classic theme, also `.carbon-row` and `.carbon-column` classNames no longer exist, adjust your code accordingly before updating if your code relies on these classes.

This PR removes classic theme support from Row and Column components, refactors stories to use new CSF/MDX formats and replaces `scss` files with `styled-components`

### Checklist


src/components/row/column/column.d.ts
```diff
@@ -1,36 +1,20 @@
 import * as React from 'react';
 
 export interface ColumnProps {
-    /**
-     * Classes to apply to the component.
-     */
+    /** Children. */
+    children: React.ReactNode;
```
sjohnson-sage, line 5: `children` is optional in the propTypes in the component but required here, one or the other needs to change

src/components/row/column/column.d.ts
```diff
@@ -1,36 +1,20 @@
...
+    /** Total column count. */
+    columns?: number | string;
+    /** Classes applied by row component to affect all rows */
     columnClasses?: string;
-
-    /**
-     * Alignment of content within column.
-     */
+    /** Alignment of content within column. */
     columnAlign?: 'left' | 'right';
```
sjohnson-sage, line 13: This can be any string in the propTypes but limited to "left" or "right" here, I think probably the propTypes one needs to change to reflect this

sjohnson-sage: I see that these aren't really your changes but worth doing IMO while you changing them

src/components/row/row.component.js
```diff
@@ -0,0 +1,68 @@
...
+
+const Row = ({
+  columns,
+  children,
+  columnClasses,
+  columnDivide,
+  gutter,
+  className,
+}) => {
+  const columnsCount = columns || React.Children.toArray(children).length;
```
sjohnson-sage, line 14: There's a `React.Children.count(children)` that could be used here instead

src/components/row/row.d.ts
```diff
@@ -1,39 +1,21 @@
 import * as React from 'react';
-import Column from './column';
 
 export interface RowProps {
-
-    /**
-     * Classes to apply to the component.
-     */
+    /** This component supports children of type Column. */
+    children: React.ReactNode;
```
sjohnson-sage, line 5: This is required here but optional in propTypes

sjohnson-sage (commented):
Just a few minor comments but all looks good!

src/components/row/row.component.js
mkrds: It looks that it has to stay that way because there is an unit test that checks if passing `null` as a children counts it toward the column count. Looks like using `React.Children.count` actually counts the `null`, while `React.Children.toArray(children).length` does not (as intended).

sjohnson-sage (approved):

edleeks87 (dismissed):

DlgSHi (approved):

src/components/inline-inputs/inline-inputs.style.js
```diff
@@ -20,83 +33,25 @@ const StyledInlineInputs = styled.div`
...
+    ${StyledRow} {
       margin-bottom: 0;
-      padding-left: 8px;
-    }
-  }
-
-  .carbon-row.carbon-row--gutter-small {
-    margin-bottom: 0;
-    margin-left: -16px;
+      margin-left: -${spacings[gutter]}px;
```
edleeks87, line 47: Could gutter not equal "none" which would cause some invalid CSS in `InlineInputs` or `Row`?

sjohnson-sage (approved):

src/components/row/row.style.js
```diff
@@ -4,6 +4,7 @@ import baseTheme from "../../style/themes/base";
 import StyledColumn from "./column/column.style";
 
 const sizes = {
+  none: 0,
```
Alec-Walker-Sage, line 7: none should be added to the MDX gutter example to show what none looks like, the test story knobs and to the row component so it shows it as an option in the docs.

Alec-Walker-Sage (approved):

PR #3419, opened 2020-11-23
test(row): add performance testing tool

KatarzynaQA:
### Proposed behaviour

1. Add `react-component-benchmark` to the carbon 
2. data is logged into local console

result

### Current behaviour

We need to consider of implementation for performance data for all project to track all timing for components/stories

### Checklist

### Additional context

To see the data in the console  first of all run 'npm start' and then `npm test ./src/components/row/__spec__.js`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35399,6 +35399,11 @@
         }
       }
     },
+    "react-component-benchmark": {
+      "version": "0.0.4",
+      "resolved": "https://registry.npmjs.org/react-component-benchmark/-/react-component-benchmark-0.0.4.tgz",
+      "integrity": "sha1-7uNYXiIDZF7ZaKQAIPYH2JvIL2Q="
+    },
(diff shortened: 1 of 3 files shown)


jamime (commented):
I left this comment on #3393 and I think the same applies here

> Performance profiling is tricky, this PR does expose the metrics, but it's not possible to say if something has improved or got worse because theres no baseline.
>
>I'm sure we can come up with some static values that we can use as a maximum or minimum but I think we're most interested in the trend over time and identifying when I changed component X, component Y degraded.
>
> ...
>
>I'd like us to investigate our options further and merge in a complete solution.

KatarzynaQA:
React-component-benchmark uses `enzyme` to render the components instead of realtime rendering of components as it is in `Profiler`. So the better for performance testing is `Profiler` which we choose to further investigation

PR #3465, opened 2020-12-03
fix: prevent accidental form submission - FE-3336

mkrds:
### Proposed behaviour
Change `type` of all the buttons to be `button` instead of the default `submit` value.

### Current behaviour
Some of the buttons that we use internally does not have a `type` defined meaning that they default to `type="submit"`. If a component that uses this type of button will be placed inside of `<form>` tag this will lead to unintentional form submission.

### Checklist

### DIFF
--- a/src/components/text-editor/__internal__/toolbar/toolbar-button/toolbar-button.style.js
+++ b/src/components/text-editor/__internal__/toolbar/toolbar-button/toolbar-button.style.js
@@ -3,7 +3,7 @@ import { isDLS } from "../../../../../utils/helpers/style-helper";
 import baseTheme from "../../../../../style/themes/base";
 import StyledIcon from "../../../../icon/icon.style";
 
-const StyledToolbarButton = styled.button`
+const StyledToolbarButton = styled.button.attrs({ type: "button" })`
   background-color: inherit;
   border: none;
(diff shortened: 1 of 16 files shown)


jamime (approved):

src/components/text-editor/__internal__/toolbar/toolbar-button/toolbar-button.style.js
```diff
@@ -3,7 +3,7 @@ import { isDLS } from "../../../../../utils/helpers/style-helper";
 import baseTheme from "../../../../../style/themes/base";
 import StyledIcon from "../../../../icon/icon.style";
 
-const StyledToolbarButton = styled.button`
+const StyledToolbarButton = styled.button.attrs({ type: "button" })`
```
edleeks87, line 6: This already has type button in the component file, I think it should be removed from there and implemented like this so we are consistent

https://github.com/Sage/carbon/blob/adbe79f619dab19eb147f00868246066c6abcc47/src/components/text-editor/__internal__/toolbar/toolbar-button/toolbar-button.component.js#L31

jamime (approved):

edleeks87 (approved):

PR #3467, opened 2020-12-03
fix(dialog): fix styling for pill close icon

### LINKED ISSUE #3397: MultiSelect component - The icon is displayed outside the Pill component into Dialog component
### Current behaviour

When we add a `MultiSelect` into a dialog, the close icon of the Pill is displayed outside the `Pill`. See picture:
image
 

### Expected behaviour

We expect that the closed icon is displayed into the `Pill`. See Picture:
image

### Reproducible example

https://codesandbox.io/s/laughing-wilson-4lvbp?file=/src/index.js
### Your environment

| carbon-react     | 48.7.1
| carbon-factory   |
| react-scripts    | 
| React            | 16.8.6
| Browser          | Chrome
| npm              |
| Operating System | Windows

DipperTheDan:
Fixes an issue raised where the close icon for pill is rendered outside of the pill component. This
only happens when pill is used inside of a dialog.

fixes FE-3328
fixes #3397

### Proposed behaviour

Screenshot 2020-12-04 at 13 03 12

### Current behaviour

image

### Checklist

### Additional context
N/A

### Testing instructions
https://codesandbox.io/s/jolly-currying-kcjus?file=/src/index.js


edleeks87 (approved):

src/components/dialog/dialog.style.js
```diff
@@ -72,7 +73,6 @@ const DialogStyle = styled.div`
       min-height: 0px !important;
     `}
 
-
   ${StyledIconButton} {
```
sjohnson-sage, line 76: I think a better fix would be to just move this `StyledIconButton` css into `DialogTitleStyle` then it'll only apply to the Dialog close icon as intended. It shouldn't be applying to anything inside the Dialog content

Then they'll be no need for the overriding CSS specifically for Pill below, we'll be removing styles from Pill rather than adding even more

sjohnson-sage (changes requested):

src/components/dialog/dialog.style.js
DipperTheDan: Good idea. I did try moving the code around to avoid having to add more code to the styles but couldn't achieve what was required in the ticket. I'll give your suggestion a shot and see what is what. Thanks.

edleeks87 (approved):

Issue #3343, opened 2020-11-02
Tabbing behaviour is inconsistent in Menu


edleeks87:
### Current behaviour

The tabbing behaviour in the Menu component is inconsistent, if you press `tab` whilst a `MenuItem` without a submenu it focuses the first item, but if you navigate to a submenu item and press `tab` it moves focus away from the `Menu`. This also effects the alpha key navigation as a result, after you have tabbed away from the submenu item if you then refocus the menu and press the `m` key it will jump to the item that was next before you originally tabbed away.

[example of weird keyboard behavior in menu.mov.zip](https://github.com/Sage/carbon/files/5474485/example.of.weird.keyboard.behavior.in.menu.mov.zip)

### Expected behaviour

I would expect it to behave like this https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

### Reproducible example

Issue is present in latest version and can be seen here https://carbon.sage.com/?path=/story/design-system-menu--default-story

### Suggested solution(s)

Update the `actualFocusedItemIndex` ref when tabbing occurs

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     |
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

nicktitchmarsh:
FE-3304 has been raised for this

Issue #3286, opened 2020-10-13
Menu/MenuItem - Error setIsOpenSubmenu is not a function


jlongridge:
### Current behaviour
Release 39.1.0 added keyboard navigation which causes an error when menu items are split between components.  

```js
Error: Uncaught [TypeError: setIsOpenSubmenu is not a function]
```

```
const App = () => {
  return (
    <Menu>
      <MenuItem>Menu Item 1</MenuItem>
      <MySubMenu />
    </Menu>
  );
};

const MySubMenu = () => {
  return (
    <MenuItem submenu="Foo">
      <MenuItem>Bar</MenuItem>
      <MenuItem>Baz</MenuItem>
    </MenuItem>
  );
};
```
Menu passes setIsOpenSubmenu to the MenuItem if you have an extra component inbetween that could be the cause.
### Expected behaviour
No JS errors when MenuItems are
Menu passes setIsOpenSubmenu to the MenuItem, it should be possible to have an extra component in between without causing js errors.

### Reproducible example
https://codesandbox.io/s/patient-fog-xj9qx?file=/src/App.js

### Your environment
Carbon >39.10

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | > 39.1.0
| react-scripts    | 3.4.3 
| React            | 16.13.1
| Browser          | Chrome
| Operating System | Mac

jamime:
I think we'll look to solve this using [context](https://reactjs.org/docs/context.html) to pass the required props to the children.

DipperTheDan:
FE-3243

PR #3471, opened 2020-12-04
refactor(menu): rewrite component

sjohnson-sage:
Fixes: #3286
Fixes: #3343 

### Proposed behaviour
A rewrite of the `Menu` and `MenuItem` components. The submenu functionality of `MenuItem` has been split out into another component, `Submenu` without changing the interface for consumers and therefore avoiding a breaking change.

Rather than using props to pass things down from `Menu` to `MenuItem` it has now been refactored to use context instead, which means that `MenuItem` no longer needs to be a direct child of `Menu` in order for keyboard navigation etc to work. 

I have also split out as much of the keyboard navigation as I could into a separate file in an effort to simply the logic in these components.

Both `Menu` and `MenuItem` now extend the `Box` component, so all of the `Box` props can be added for alignment of children etc. An example of this is shown at the new `design-system-menu--full-navbar-alignment-example` story.

A "submenu direction left` story has also been added

### Current behaviour
See issue #3286 for current behaviour using props rather than context.

### Checklist

### Testing instructions
A nice example to test keyboard navigation etc can be found here at the PR sandbox in the codesandbox bot comment below 
https://codesandbox.io/s/quizzical-rain-xns67?file=/src/index.js


src/components/menu/menu.component.js
```diff
@@ -1,132 +1,81 @@
...
-    } else if (isFocusedOnFirstMenuItem) {
-      actualFocusedItemIndex.current = menuItemsRefs.current.length - 1;
-    }
+const Menu = ({ menuType = "light", children, ...rest }) => {
+  const [focusedItemIndex, setFocusedItemIndex] = useState(undefined);
+  const [openSubmenuIndex, setOpenSubmenuIndex] = useState(undefined);
 
-    menuItemsRefs.current[actualFocusedItemIndex.current].current.focus();
-  }, []);
+  const reset = () => {
```
edleeks87, line 14: This could be moved inside the scope of `onClickOutside` as it's not used anywhere else

edleeks87 (commented):
You'll need to update or add some TypeScript definitions as a new component has been added and some prop types have been tightened

src/components/menu/submenu-block/submenu-block.component.js
```diff
@@ -1,27 +1,38 @@
...
+let deprecatedWarnTriggered = false;
+
+const SubmenuBlock = ({ children, variant = "default" }) => {
+  const menuContext = useContext(MenuContext);
+
+  if (!deprecatedWarnTriggered) {
+    deprecatedWarnTriggered = true;
+    // eslint-disable-next-line max-len
+    Logger.deprecate(
+      "`styleOverride` that is used in the `Fieldset` component is deprecated and will soon be removed."
```
edleeks87, line 16: As mentioned in private conversation, this needs updating

cypress/locators/menu/index.js
```diff
@@ -1,13 +1,15 @@
...
+  SUBMENU,
 } from './locators';
 
 // component preview locators
 export const menuPreview = () => cy.get(MENU_PREVIEW);
 export const menuListItems = () => cy.get(MENU_PREVIEW)
   .find(MENU_ITEM);
-export const submenuBlock = () => cy.get(MENU_ITEM)
+export const submenu = () => cy.get(SUBMENU);
+export const submenuBlock = () => cy.get(SUBMENU)
```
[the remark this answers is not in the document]
sjohnson-sage: Ah yes, I've got lazy now prettier has been added in so I don't even bother typing one anymore, forgot it doesn't run over these files

sjohnson-sage:
> You'll need to update or add some TypeScript definitions as a new component has been added and some prop types have been tightened

The new component is just an internal one so I don't think we need to provide a ts interface for it, I could be wrong about that though

src/components/menu/menu.stories.mdx
```diff
@@ -367,12 +312,85 @@ is rendered.
...
+              </MenuItem>
+            </Menu>
+          </Navbar>
+        </div>
+      );
+    }}
+  </Story>
+</Preview>
+
+## Split submenu into separate component
```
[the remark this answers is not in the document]
sjohnson-sage: I've forgotten to add a `href` or `onClick` to those `MenuItem`s - I'll fix

src/components/menu/__internal__/submenu/submenu.component.js
```diff
@@ -0,0 +1,267 @@
...
+            setSubmenuOpen(true);
+            setSubmenuFocusIndex(numberOfChildren - 1);
+          }
+
+          if (!event.defaultPrevented) {
+            menuContext.handleKeyDown(event);
+          }
+        }
+
+        if (submenuOpen) {
```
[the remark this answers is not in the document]
sjohnson-sage: It could be but it's cleaner IMO to have this, easier to someone to look through and immediately see what's going on

src/components/menu/menu.component.js
```diff
@@ -1,132 +1,81 @@
...
+      if (submenuOpen) {
+        setOpenSubmenuIndex(newIndex);
       }
     },
-    [filteredChildren, setFocusToElement]
+    [children]
   );
 
-  let index = 0;
+  /* istanbul ignore next */
```
jamime, line 36: You can test this, see ActionPopover spec for example.
```suggestion
```

src/components/menu/menu.component.js
```diff
@@ -1,132 +1,81 @@
...
-            </StyledMenuItem>
-          );
-        })}
-      </StyledMenuItemsWrapper>
+        return (
+          <MenuContext.Provider
+            value={{
+              menuType,
+              isFirstElement,
+              handleKeyDown: (ev, submenuOpen) =>
```
jamime, line 69: Can this be simplified? Seems like we're passing it down, using `openSubmenu` but then passing it back as an argument to this function.
```suggestion
              handleKeyDown: ev =>
```

src/components/menu/submenu-block/submenu-block.component.js
jamime: Do we need this warning at all?

jamime (changes requested):
Good work sam.

One thing I noticed was there are two props `variant` and `menuType`, given we already have a codemod that can change a prop. I think it would be a good time to consolidate these by removing `menuType`.

src/components/menu/submenu-block/submenu-block.component.js
edleeks87: The component may have been used even though it is stripped away inside `MenuItem`

edleeks87:
> > You'll need to update or add some TypeScript definitions as a new component has been added and some prop types have been tightened
> 
> The new component is just an internal one so I don't think we need to provide a ts interface for it, I could be wrong about that though. Also I think the only other props I've changed are removing private ones, which weren't in the ts files to begin with

Ah yeah good point about the `__internal__`, I'd also missed you were updating `children` in `Menu` to be required as they are defined in the ts file

cypress/locators/menu/index.js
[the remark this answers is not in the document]
sjohnson-sage: Actually no I didn't, that's not the end of the line 😉

src/components/menu/__internal__/submenu/submenu.component.js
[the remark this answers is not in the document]
edleeks87: I don't think an `if/else` is so complex that someone couldn't work it out, but as there are a lot of nested ifs I can see your point about legibility so happy to defer to you here

sjohnson-sage:
> Good work @sjohnson-sage.
> 
> One thing I noticed was there are two props `variant` and `menuType`, given we already have a codemod that can change a prop. I think it would be a good time to consolidate these by removing `menuType`.

Thanks @jamime! 

Do we want to do that though? `menuType` is set only on `Menu` to set the overall colour scheme, then `MenuItem` can override that, so we could have some menu items just the light theme, and some alternate as in the dropdown on this story: https://carbon.sage.com/?path=/story/design-system-menu--default-theme-alternate-colour 

I might be missing something but I think it makes sense as it is

src/components/menu/menu.component.js
sjohnson-sage: I can't see a way - this is specifically for when you have a submenu item focus and you press the left or right key onto another submenu. Somehow one submenu needs to communicate to open the second submenu through Menu, and this was the cleanest way I could fine. 
Happy to change if there's a better way I can't see though!

jamime:
> > Good work @sjohnson-sage.
> > One thing I noticed was there are two props `variant` and `menuType`, given we already have a codemod that can change a prop. I think it would be a good time to consolidate these by removing `menuType`.
> 
> Thanks @jamime!
> 
> Do we want to do that though? `menuType` is set only on `Menu` to set the overall colour scheme, then `MenuItem` can override that with the `variant` prop, so we could have some menu items just the light theme, and some alternate as in the dropdown on this story: https://carbon.sage.com/?path=/story/design-system-menu--default-theme-alternate-colour
> 
> I might be missing something but I think it makes sense as it is

We've discussed and we're going to push this down the road to avoid the breaking change. We'll revisit this, likely when we do the DLS alignment exercise.

jamime (dismissed):

edleeks87 (approved):

jamime (dismissed):

edleeks87 (approved):

sianford:
Tested the refactoring work and regression tested all of Menu in Safari, Chrome, Firefox and Edge Chromium in the Chromatic published Storybook. Testing passed ✅ Awaiting completion of the associated MDX example task before merging.

jamime (approved):

edleeks87 (dismissed):

Alec-Walker-Sage (dismissed):

edleeks87 (approved):

Alec-Walker-Sage (approved):

jamime (approved):

PR #3472, opened 2020-12-07
test: add improvements for accessibility

DlgSHi:
### Proposed behaviour

- improve our step for `accessibility` tests, add new assertion to check `No Preview` story;
- fix/update few accessibility tests.

### Current behaviour

- We need to improve our step for `accessibility`. Because when we open a non-existing story we have `No Preview` story and `cy.axe` just confirming that we have no accessibility errors on that story (what we actually don't need to). We need to check does the `No Preview` story is not visible before we will check `accessibility` tests.
- To make it possible we need to fix/update few accessibility tests.

### Checklist


KatarzynaQA (dismissed):

DlgSHi:
Needs to be done as a separate task

PR #3478, opened 2020-12-08
feat(box): extend box to support scrollbar style

nicktitchmarsh:
Fixes FE-3398

### Proposed behaviour
Extend box component to allow an override of scrollbar styles. A new prop will allow a light or dark theme to be shown. 

image

### Current behaviour
Scroll boxes are un-styled.

### Checklist


src/components/box/box.stories.mdx
```diff
@@ -63,11 +63,37 @@ Basic component that gives access to a number of styling options. Should be used
   </Story>
 </Preview>
 
-## Props
+## Scroll
+Use the `scrollType` prop to indicate which theme scrollbar you would like to display. Use `light` or `dark` values to display as below. This will still require the use of the `overflow` prop and only displays in Chrome or Safari
```
jamime, line 67: We use a `variant` in most of new components to change the theme, should this be `scrollVariant`?

src/components/flat-table/__snapshots__/flat-table.spec.js.snap
```diff
@@ -62,31 +62,35 @@ exports[`FlatTable when rendered with proper table data should have expected str
...
+.c1 .c3 .c13,
+.c1 .c8,
+.c1 .c3 .c6 {
   background-color: #335C6D;
   border-right: 1px solid #668592;
   color: #FFFFFF;
 }
 
-.c10 {
+.c0::-webkit-scrollbar {
```
sjohnson-sage, line 89: Shouldn't this only be added if a `Box` is used with the `scrollVariant` prop? Otherwise it should be browser default?

src/components/box/box.component.js
```diff
@@ -9,6 +10,38 @@ const Box = styled.div`
   ${layout}
   ${flexbox}
   ${color}
+
+  ${({ scrollVariant, theme }) => css`
+    &::-webkit-scrollbar {
```
sjohnson-sage, line 15: As my comment below, do we not just want this added when the `scrollVariant` prop has been added?

jamime: When we make this change, would this be useful to move into a helper, like I did for `${color}`, that would allow other components to add it without extending `Box`? Or should we do this at a later date?

src/components/box/box.component.js
```diff
@@ -9,6 +10,40 @@ const Box = styled.div`
   ${layout}
   ${flexbox}
   ${color}
+
+  ${({ scrollVariant, theme }) => css`
+    ${scrollVariant === "light" &&
+    css`
+      scrollbar-color: ${theme.scrollbar.light.thumb}
```
jamime, line 17: Theres a lot of duplication between the light and dark variants. We could simplify this by using the prop like this
```suggestion
      scrollbar-color: ${theme.scrollbar[scrollVariant]thumb}
```

src/components/box/box.spec.js
```diff
@@ -1,16 +1,74 @@
...
 import Box from "./box.component";
+import BaseTheme from "../../style/themes/base";
 
 describe("Box", () => {
   testStyledSystemSpacing((props) => <Box {...props} />);
   testStyledSystemColor((props) => <Box {...props} />);
   testStyledSystemLayout((props) => <Box {...props} />);
   testStyledSystemFlexBox((props) => <Box {...props} />);
+
+  it("has styles applied when scrollVariant is set to light", () => {
```
jamime, line 19: These tests could also be simplified by using `it.each`

```jsx
it.each(['light', 'dark], "has styles applied when scrollVariant is set to %s", (scrollVariant) => {
  const wrapper = mount(<Box scrollVariant={scrollVariant} />);
  // ...
   assertStyleMatch(
      {
        backgroundColor: BaseTheme.scrollbar[scrollVariant]track,
      },
      wrapper,
      { modifier: "::-webkit-scrollbar-track" }
    );
  // ...
})
```

src/components/box/box.stories.mdx
```diff
@@ -63,11 +63,37 @@ Basic component that gives access to a number of styling options. Should be used
   </Story>
 </Preview>
 
-## Props
+## Scroll
+Use the `scrollVariant` prop to indicate which theme scrollbar you would like to display. Use `light` or `dark` values to display as below. This will still require the use of the `overflow` prop and only displays in Chrome or Safari
+<Preview>
+  <Story name="scroll" parameters={{ info: { disable: true }}} >
```
jamime, line 69: I checked chromatic and the scroll bar isn't visible in the preview. I think we can disable the story unless we can find a way to fix it. We should contact chromatic using the in app chat, they normally reply within a few hours.

jamime (changes requested):

src/components/box/box.stories.mdx
jamime: I contacted them, I'll report back when they reply.

jamime: > We hide scrollbars when taking snapshots so that screenshots capture the content only. Unfortunately there's no workaround for capturing scrollbars right now. I can see your use case though and will pass it on as a feature request to the team. 

Lets disable chromatic on this story.

nicktitchmarsh: Disabled chromatic on that story

src/components/box/box.spec.js
nicktitchmarsh: Yeah this is better. Ive changed it over

src/components/box/box.component.js
nicktitchmarsh: This is definitely more efficient

src/components/box/box.component.js
nicktitchmarsh: Moved this to only be added when scrollVariant isnt undefined

src/components/flat-table/__snapshots__/flat-table.spec.js.snap
nicktitchmarsh: Same as below. This is just snapshots

src/components/box/box.component.js
```diff
@@ -9,6 +10,24 @@ const Box = styled.div`
   ${layout}
   ${flexbox}
   ${color}
+
+  ${({ scrollVariant, theme }) => css`
+    ${typeof scrollVariant !== "undefined" &&
```
sjohnson-sage, line 15: Could you not just have done `${scrollVariant && css...` ?

src/components/box/box.stories.mdx
```diff
@@ -63,11 +63,37 @@ Basic component that gives access to a number of styling options. Should be used
...
   flexBox
   layout
-/>
\ No newline at end of file
+/>
+<PropsTable rows={[
+  {
+    name: "scrollVariant",
+    type: { summary: 'dark | light' },
+    description: 'String to denote which theme style to use for scrollbars, can be light theme or dark theme',
```
jamime, line 95:
```suggestion
    description: 'String to denote which theme style to use for scrollbars,
```

I think this is confusing, the prop table will say `light|dark` but the comment says it can be `light theme` or `dark theme`. I think this is better omitted.

jamime (commented):
Happy to approve once outstanding comment by Sam has been resolved. I've left a minor comment which about the mdx prop table too.

sjohnson-sage (dismissed):

jamime (approved):

jamime (approved):

sjohnson-sage (approved):

harpalsingh:
Passed UX review, but maybe later would be better to use a darker background on the demo for the dark variant.

src/style/themes/base/base-theme.config.js
```diff
@@ -272,6 +272,18 @@ export default (palette) => {
       errorButtonFocus: palette.errorRedShade(20),
     },
 
+    scrollbar: {
+      light: {
+        thumb: palette.slateTint(40),
+        track: palette.slateTint(90),
```
Alec-Walker-Sage, line 278: Talked the Harpal and this should actually be `slate tint 95`

Alec-Walker-Sage (changes requested):
If you are changing this tint, it might be worth adding the dark background mentioned by Harpal in a previous comment

src/style/themes/base/base-theme.config.js
edleeks87: I'll update the theme config while Nick focuses on the navigation work

jamime (approved):

sjohnson-sage (approved):

Alec-Walker-Sage (approved):

PR #3487, opened 2020-12-10
test: add improvements for accessibility

DlgSHi:
### Proposed behaviour

- improve our step for `accessibility` tests, add new assertion to check `No Preview` story;
- fix/update few accessibility tests.

### Current behaviour

- We need to improve our step for `accessibility`. Because when we open a non-existing story we have `No Preview` story and `cy.axe` just confirming that we have no accessibility errors on that story (what we actually don't need to). We need to check does the `No Preview` story is not visible before we will check `accessibility` tests.
- To make it possible we need to fix/update few accessibility tests.

### Checklist


tempertemper:
A quick note after a conversation with @DlgSHi, the accessibility tag we're working towards is `wcag21aa` (WCAG 2.1 Level AA). Achieving 2.1 AA will also satisfy 2.1 A, as AA is the level higher than A. 2.2 isn't a recommendation yet, so we don't have to worry about that until it is (likely late next year).

mkrds (approved):

KatarzynaQA (approved):

PR #3504, opened 2020-12-16
docs(action-popover): add flat table story

grabkowski:
### Proposed behaviour
Add a story that shows implementation in the FlatTable

### Checklist

### Testing instructions
1. run `npm start`
2. open http://localhost:9001/?path=/docs/design-system-action-popover--in-flat-table


jamime (commented):
Story looks good but I think this should be in the `FlatTable` MDX instead of `ActionPopover`.

grabkowski:
> Story looks good but I think this should be in the `FlatTable` MDX instead of `ActionPopover`.

We are planning to remove the `Table`, so the `ActionPopover` will be placed only in the `FlatTable`.
I think that people will be looking for an example of it's implementation under `ActionPopover` not the `FlatTable`.

jamime (approved):

mkrds (approved):

PR #3513, opened 2020-12-17
fix(navbar): remove old styling to allow full width

sjohnson-sage:
### Proposed behaviour
Remove min and max width and auto margin, leaving the nav bar to be the full width of the screen.

Add an MDX example to show users how to set a max-width on their nav bar content using `Box`: 

image

### Current behaviour
The nav bar has a min and max width, and margin auto for left and right which is causing problems with the max width (I.e. the nav bar not spanning the whole width of the page)

### Checklist

### Testing instructions
http://localhost:9001/?path=/story/design-system-navigation-bar--content-max-width-using-box
and test the other nav bar stories


edleeks87 (dismissed):

jamime (approved):
I checked this in one of our projects and it works as expected.

jamime (approved):

edleeks87 (approved):

PR #3515, opened 2020-12-17
feat(menu): add scrollable block component

sjohnson-sage:
### Proposed behaviour
New `ScrollableBlock` component for use in submenus. This component extends `Box` so any box prop can be used. 

image

image

image

image

### Current behaviour
There is no way to add a scrollable block in a submenu.

### Checklist

### Testing instructions
http://localhost:9001/?path=/story/design-system-menu--scrollable-submenu
http://localhost:9001/?path=/story/design-system-menu--scrollable-submenu-dark

### DIFF
--- a/cypress/features/regression/designSystem/menu.feature
+++ b/cypress/features/regression/designSystem/menu.feature
@@ -1,0 +1,15 @@
+Feature: Design Systems Menu component
+	I want to test the Design Systems Menu component
+
+	Background: Open Menu on scrollable menu story
+		Given I open "Design System Menu" component page "scrollable-submenu" in no iframe
+
+  @positive
+  Scenario Outline: Test that the scroll block within a submenu is scrollable 
+    When I open the "<position>" submenu
+      And I scroll to the bottom of the block
+    Then The last element is visible
+    Examples:
+    | position |
+    | first    |
+    | second   |
--- a/cypress/locators/menu/locators.js
+++ b/cypress/locators/menu/locators.js
@@ -2,3 +2,4 @@
 export const MENU_PREVIEW = '[data-component="menu"]';
 export const MENU_ITEM = '[data-component="menu-item"]';
 export const SUBMENU = '[data-component="submenu-wrapper"]';
+export const SCROLL_BLOCK = '[data-component="submenu-scrollable-block"]';
--- a/cypress/support/step-definitions/menu-steps.js
+++ b/cypress/support/step-definitions/menu-steps.js
@@ -3,6 +3,8 @@ import {
   submenuBlock,
   innerMenu,
   submenu,
+  scrollBlock,
+  lastSubmenuElement,
 } from '../../locators/menu';
 import { positionOfElement } from '../helper';
 
@@ -20,4 +22,16 @@ Then('Menu third expandable element has inner elements', () => {
   innerMenu(positionOfElement('third')).should('have.attr', 'data-component', 'menu-divider');
   innerMenu(positionOfElement('fourth')).should('have.attr', 'data-component', 'link');
   innerMenu(positionOfElement('fifth')).should('have.attr', 'data-component', 'link');
+});
+
+When('I open the {string} submenu', (position) => {
+  submenu().eq(positionOfElement(position)).trigger('mouseover');
+});
+
+When('I scroll to the bottom of the block', () => {
+  scrollBlock().scrollTo('bottom');
+});
+
+Then('The last element is visible', () => {
+  lastSubmenuElement().should('be.visible');
 });
--- a/src/components/menu/__internal__/submenu/submenu.component.js
+++ b/src/components/menu/__internal__/submenu/submenu.component.js
@@ -17,8 +17,8 @@ import Events from "../../../../utils/helpers/events";
 import { MenuContext } from "../../menu.component";
 import MenuItem from "../../menu-item";
 import { characterNavigation } from "../keyboard-navigation";
-
-const SubmenuContext = React.createContext({});
+import ScrollableBlock from "../../scrollable-block";
+import SubmenuContext from "./submenu.context";
 
 const Submenu = React.forwardRef(
   (
@@ -41,9 +41,19 @@ const Submenu = React.forwardRef(
     const [submenuFocusIndex, setSubmenuFocusIndex] = useState(undefined);
     const submenuRef = useRef();
     const menuContextOpen = menuContext.openSubmenu;
-    const numberOfChildren = useMemo(() => React.Children.count(children), [
-      children,
-    ]);
+
+    const formattedChildren = React.Children.map(children, (child) => {
+      if (child.type === ScrollableBlock) {
+        return [...child.props.children];
+      }
+
+      return child;
+    });
+
+    const numberOfChildren = useMemo(
+      () => React.Children.count(formattedChildren),
+      [formattedChildren]
+    );
 
     const closeSubmenu = useCallback(() => {
       setSubmenuOpen(false);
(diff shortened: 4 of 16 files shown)


src/components/menu/__internal__/submenu/submenu.component.js
```diff
@@ -17,8 +17,8 @@ import Events from "../../../../utils/helpers/events";
 import { MenuContext } from "../../menu.component";
 import MenuItem from "../../menu-item";
 import { characterNavigation } from "../keyboard-navigation";
-
-const SubmenuContext = React.createContext({});
+import ScrollableBlock from "../../scrollable-block";
+import SubmenuContext from "./submenu.context";
```
nicktitchmarsh, line 21: Do we really need to separate out the context into a different file?

nicktitchmarsh: Stops circular dependency. resolved

nicktitchmarsh (dismissed):

src/components/menu/__internal__/submenu/submenu.component.js
```diff
@@ -214,7 +227,11 @@ const Submenu = React.forwardRef(
               <SubmenuContext.Provider
                 value={{
                   isFocused: submenuFocusIndex === index,
+                  focusIndex: submenuFocusIndex,
                   handleKeyDown,
+                  blockIndex: React.Children.toArray(children).findIndex(
```
edleeks87, line 232: Would this not only support a single `ScrollableBlock` within a `Submenu`?

src/components/menu/__internal__/submenu/submenu.spec.js
```diff
@@ -748,4 +749,28 @@ describe("Submenu component", () => {
       );
     });
   });
+
```
edleeks87, line 752: Would be good to add a test for keyboard nav with a scrollable block just so we can catch it if there's a regression

src/components/menu/__internal__/submenu/submenu.component.js
sjohnson-sage: It would break it if there was multiple, my assumption here is what you said there, that there will only be a single scrollable block in the submenu - which I think is a fair assumption to make!

src/components/menu/__internal__/submenu/submenu.spec.js
sjohnson-sage: I did think about it but the keyboard nav just loops over `MenuItems` - so the test below proving that the scrollable block renders all of it's child `MenuItems` is enough I think, otherwise we're just testing the keyboard nav functionality twice

edleeks87: Yeah that's a fair point, as long as any regression in that behaviour is caught then all good

src/components/menu/__internal__/submenu/submenu.component.js
sjohnson-sage: Added a description to the MDX as discussed elsewhere

edleeks87 (approved):

nicktitchmarsh (approved):

cypress/support/step-definitions/menu-steps.js
```diff
@@ -21,3 +22,11 @@ Then('Menu third expandable element has inner elements', () => {
   innerMenu(positionOfElement('fourth')).should('have.attr', 'data-component', 'link');
   innerMenu(positionOfElement('fifth')).should('have.attr', 'data-component', 'link');
 });
+
+When('I open {int}', (index) => {
+  submenu().eq(index - 1).trigger('mouseover');
```
DlgSHi, line 27: we could use here the function `positionOfElement()` and pass here a `{string}` as `first`/`second` instead of passing `{int}`  and subtracting some `magic numbers`.

cypress/features/regression/designSystem/menu.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Design Systems Menu component
+  I want to test the Design Systems Menu component
+    
+  @positive
+  Scenario Outline: Scenario name: Test that all child link elements are part of a scrollable block
+      Given I open "Design System Menu" component page "scrollable-submenu" in no iframe
```
DlgSHi, line 6: 1. We can move this step to the `Background:` above the TC;
2. Pass here `scrollable submenu` as a parameter to make it more readable.

cypress/features/regression/designSystem/menu.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Design Systems Menu component
+  I want to test the Design Systems Menu component
+    
+  @positive
+  Scenario Outline: Scenario name: Test that all child link elements are part of a scrollable block
```
DlgSHi, line 5: Please delete this - probably left when you've created a step with auto completion - `Scenario name:`

cypress/locators/menu/locators.js
```diff
@@ -2,3 +2,4 @@
 export const MENU_PREVIEW = '[data-component="menu"]';
 export const MENU_ITEM = '[data-component="menu-item"]';
 export const SUBMENU = '[data-component="submenu-wrapper"]';
+export const SCROLL_BLOCK = 'div[overflow="auto"]';
```
DlgSHi, line 5: are you sure this is a correct locator ?

cypress/features/regression/designSystem/menu.feature
```diff
@@ -0,0 +1,12 @@
...
+      Given I open "Design System Menu" component page "scrollable-submenu" in no iframe
+      When I open <submenu number>
+      Then The submenu is scrollable
```
DlgSHi, line 8: We just scroll here, but we actually need to be sure that submenu was scrolled to the proper element. So we need to add something that some element became visible after scroll.

cypress/locators/menu/locators.js
Alec-Walker-Sage: Yep, it finds the right element. There is probably a more unique way to locate it though

cypress/support/step-definitions/menu-steps.js
```diff
@@ -23,10 +24,14 @@ Then('Menu third expandable element has inner elements', () => {
   innerMenu(positionOfElement('fifth')).should('have.attr', 'data-component', 'link');
 });
 
-When('I open {int}', (index) => {
-  submenu().eq(index - 1).trigger('mouseover');
+When('I open the {string} submenu', (index) => {
```
DlgSHi, line 27: it is rather a position instead of `index`

cypress/features/regression/designSystem/menu.feature
```diff
@@ -1,12 +1,15 @@
...
-      Then The submenu is scrollable
-      Examples:
-      | submenu number |
-      | 1              |
-      | 2              |
\ No newline at end of file
+    I want to test the Design Systems Menu component
+
+    Background: Open Menu on scrollable menu story
+    Given I open "Design System Menu" component page "scrollable-submenu" in no iframe
```
DlgSHi, line 5: paste `scrollable submenu` as a parameter here :)

DlgSHi, line 5: 2 spaces indent should be here

cypress/features/regression/designSystem/menu.feature
```diff
@@ -1,12 +1,15 @@
...
+
+    @positive
+    Scenario Outline: Test that all child link elements are part of a scrollable block
+        When I open the "<position>" submenu
+            And I scroll to the bottom of the block
```
DlgSHi, line 10: 2 spaces (1 tab) indent should be here instead of 4 spaces (2 tabs)

cypress/features/regression/designSystem/menu.feature
```diff
@@ -1,12 +1,15 @@
...
-      | 1              |
-      | 2              |
\ No newline at end of file
+    I want to test the Design Systems Menu component
+
+    Background: Open Menu on scrollable menu story
+    Given I open "Design System Menu" component page "scrollable-submenu" in no iframe
+
+    @positive
+    Scenario Outline: Test that all child link elements are part of a scrollable block
```
DlgSHi, line 8: I think we need more precise the decription

nicktitchmarsh (approved):

edleeks87 (dismissed):

Alec-Walker-Sage (dismissed):

Alec-Walker-Sage (approved):

nicktitchmarsh (approved):

edleeks87 (approved):

PR #3529, opened 2020-12-30
test(cypress): remove flash accessibility test

sjohnson-sage:
### Proposed behaviour
Re-enable some of the accessibility tests in Cypress and remove the `Flash` test case, as the `Flash` component no longer exists in Carbon

### Current behaviour
A selection of the accessibility tests are not run with an `@ignore` tag

### Checklist

### Testing instructions
The Cypress builds passing should be enough

### DIFF
--- a/cypress/features/regression/accessibility/accessibilityCommon.feature
+++ b/cypress/features/regression/accessibility/accessibilityCommon.feature
@@ -14,7 +14,6 @@ Feature: Accessibility tests - Common list
       | Content             |
       | Detail              |
       | Dialog              |
-      | Dialog-full-screen  |
       | Heading             |
       | Help                |
       | I18nComponent       |
@@ -22,7 +21,6 @@ Feature: Accessibility tests - Common list
       | Message             |
       | MenuList            |
       | Mount In App        |
-      # | Pages             |
       | Portrait            |
       | Preview             |
       | Profile             |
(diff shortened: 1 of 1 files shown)


cypress/features/regression/accessibility/accessibilityCommon.feature
```diff
@@ -36,7 +36,7 @@ Feature: Accessibility tests - Common list
     When I open "Button-Toggle-Group" component page "basic" in no iframe
     Then "Button Toggle Group" component has no accessibility violations
 
-  @ignore
+  @accessibility
```
Alec-Walker-Sage, line 39: Could be worth removing this all together as it is no longer testing anything? I think anyway

sjohnson-sage: It's looping over a load of components with the `component` example I think - I will remove the comments about being ignored though, missed those

sjohnson-sage: Actually no you're right, maybe it isn't doing anything now

Alec-Walker-Sage (commented):
What about the commented out pages on line 25?

sjohnson-sage:
> What about the commented out pages on line 25?

That's being done in a separate ticket 👍🏻

DlgSHi (approved):

Alec-Walker-Sage (approved):

PR #3544, opened 2021-01-06
chore(deps): upgrade jest

sjohnson-sage:
### Proposed behaviour
Upgrade `jest` to fix a dependabot alert for `node-notifier`

### Current behaviour
`node-notifier` version is below `8.0.1` and therefore has a security issue

### Checklist


ianforster-sage (approved):

edleeks87 (dismissed):

jamime (commented):
We've had issues in the past when we had different versions of `babel-jest` and `jest`. Can we update them both, and `@types/jest`, if theres a matching version available.

**Update**: Maybe it was `jest`/`jest-cli` https://github.com/Sage/carbon/pull/2909 that I was thinking of.

sjohnson-sage:
> We've had issues in the past when we had different versions of `babel-jest` and `jest`. Can we update them both, and `@types/jest`, if theres a matching version available.

Yeah sure, just out of interest what issues have we had with that?

edleeks87 (approved):

jamime (approved):

jamime:
> > We've had issues in the past when we had different versions of `babel-jest` and `jest`. Can we update them both, and `@types/jest`, if theres a matching version available.
> 
> Yeah sure, just out of interest what issues have we had with that?

When we were using `npm install` the tests would fail.
But when we were running `npm ci` the tests would pass.

After a lot of debugging, IIRC the child process was using the `jest` defined in the package json but the parent process was using the `jest` bundled with `jest-cli` or vice-versa. I think this was largely because we were using both `jest` and `jest-cli`.

I'd like to keep `babel-jest` and `jest` with the same version, they're developed in the same monorepo and are released at the same time.

`@types/babel` is developed in a different repo, but we should ensure we use a version that corresponds to our version of `jest`.

ianforster-sage (approved):

PR #3545, opened 2021-01-06
feat(search): added keyboard navigation to clear text button (FE-3449)

### LINKED ISSUE #3485: [ Search ] lack of possibility to clear value via keyboard
### Current behaviour

Using a `Search` component we have no possibility to clear value via keyboard.

Steps to reproduce:
1. Open `Design System` -> `Search` -> `default`
2. Type `test` inside the search input
3. Press Tab - to focus `cross` icon

Actual Result:
Cross icon is not focusable and there is no possibility to clear value via keyboard.

### Expected behaviour

Cross icon should be focusable and there should be possibility to clear value via keyboard pressing `Space`/`Enter` over the hovered `cross` icon.

### Reproducible example

[story](https://carbon.sage.com/?path=/story/design-system-search--default-story)
### Your environment

| carbon-react     | 53.0.0 |
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          | All |
| npm              |
| Operating System | macOS |

nicktitchmarsh:
Fixes FE-3449, Fixes #3485

### Proposed behaviour
Allow the clear input icon to be focusable via keyboard navigation when it is visible. 

image

### Current behaviour
When typing in the search input, the clear input icon is not keyboard accessible. 

### Checklist

### Testing instructions

http://localhost:9001/?path=/docs/design-system-search--default-story
Enter some text, press Tab


DipperTheDan:
Good day. I can see you've added the issue number in the commit message but it doesn't appear that the PR has linked to the issue, which means it won't automatically close the issue when the PR is merged. You might need to check the syntax you have used in your commit message. I think you may need to do two `fixes` if I remember rightly. 
`
Fixes FE-3449
Fixes #3485

`

DipperTheDan (dismissed):

src/__experimental__/components/input-icon-toggle/input-icon-toggle.component.js
```diff
@@ -22,14 +22,15 @@ const InputIconToggle = ({
...
+  iconTabIndex,
 }) => {
   if (
     useValidationIcon &&
     !disabled &&
     shouldDisplayValidationIcon({ error, warning, info })
   ) {
     return (
-      <InputIconToggleStyle size={size}>
+      <InputIconToggleStyle size={size} tabIndex={iconTabIndex}>
```
dawidzarzycki, line 33: I suggest to use prop `tabbable` which will be a `boolean`. It will be more friendly for the user in my opinion so the user won't think what number to use but he will could decide how it could works. 
`tabbable` prop is also used in our other components so it will be more consistent.

nicktitchmarsh: I think this is OK, input-toggle-icon isnt referenced in the docs anywhere, it seems to just be used internally. We should be able to pass tabIndex as its a normal HTML attribute.

dawidzarzycki (dismissed):

dawidzarzycki (approved):

DipperTheDan (dismissed):

Alec-Walker-Sage (approved):

DlgSHi (approved):

dawidzarzycki (approved):

edleeks87 (approved):

Issue #3475, opened 2020-12-08
Select component clears selected item even if the options contains the item


philearley:
### Current behaviour

When the select component is being interacted, if the array of options happens to change, even if the value still exists in the returned options.

### Expected behaviour

If I make an api call on open of the Select component and choose the first option in the response and then open the select element again and the api call happens again. If, the initially selected value is returned in the response, it should remain as the selected item.

### Reproducible example

https://codesandbox.io/s/compassionate-bose-b9u4m
Choose Amber, click the dropdown again wait for loading to finish.
The setTimeout is in place of our api call.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 52.0.0

grabkowski:
to fix this issue we could enforce Option's object value to have the `value` key, so we're comparing only the values instead of whole objects.

FE-3421 on backlog

grabkowski:
Issue fixed with #3593 merge. I have also corrected your code in the sandbox so it works as expected, you could check it there:
https://codesandbox.io/s/carbon-quickstart-forked-lpvbi?file=/src/MySelect.js

PR #3593, opened 2021-01-20
feat(select): object option comparison tweak (FE-3421)

grabkowski:
### Proposed behaviour
When the `id` property is set on an Option object value, objects will be compared based on that property (could be used when the list is recreated after an API call). If there is no `id` prop specified on an object, then the exact objects will be compared.

### Current behaviour
When Option values are defined as objects, the exact objects are compared looking for the selected value.

### Checklist

### Testing instructions
1. run `npm start`.
2. open http://localhost:9001/?path=/docs/design-system-select--with-object-as-value
3. open http://localhost:9001/?path=/docs/design-system-select-filterable--with-object-as-value
4. open http://localhost:9001/?path=/docs/design-system-select-multiselect--with-object-as-value

### DIFF
--- a/cypress/features/regression/designSystem/simpleSelect.feature
+++ b/cypress/features/regression/designSystem/simpleSelect.feature
@@ -77,3 +77,16 @@ Feature: Design System Select component
     When I scroll to the "bottom" of Select List
     Then Lazy loading is visible
       And Select list "Lazy Loaded A1" option is visible
+
+  @positive
+  Scenario: Check that using an object as a value displays the correct options
+    Given I open "Design System Select" component page "with object as value" in no iframe
+    When I click on dropdown button
+    Then visible options on Select list are "Amber", "Black", "Blue"
+
+  @positive
+  Scenario: Check that options can be selected and displayed correctly when using an object as a value
+    Given I open "Design System Select" component page "with object as value" in no iframe
+     And I click on dropdown button
+    When I click on "first" option on Select list
+    Then Select input has "Amber" value
--- a/src/components/select/multi-select/multi-select.component.js
+++ b/src/components/select/multi-select/multi-select.component.js
@@ -11,6 +11,8 @@ import StyledMultiSelect, {
   StyledSelectPillContainer,
 } from "./multi-select.style";
 import Pill from "../../pill";
+import isExpectedOption from "../utils/is-expected-option";
+import isExpectedValue from "../utils/is-expected-value";
 
 const FilterableSelectList = withFilter(SelectList);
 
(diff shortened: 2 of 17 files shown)


src/components/select/utils/is-expected-option.js
```diff
@@ -0,0 +1,17 @@
+import Option from "../option/option.component";
+
+export default function isExpectedOption(element, expectedValue) {
```
edleeks87, line 3: If you pass the `value` here instead of an `element` you could use this function instead of writing `isExpectedValue`

src/components/select/multi-select/multi-select.component.js
```diff
@@ -349,9 +357,22 @@ const MultiSelect = React.forwardRef(
           return valueList;
         });
       },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
       [createCustomEvent, onChange, textboxRef]
     );
 
+    function isExpectedValue(singleValue, expectedValue) {
```
edleeks87, line 364: See other comment, I don't think we need to have this duplicate code if you filter `(type === Option)` and pass `child.props.value` instead of the `child`

src/components/select/utils/is-expected-option.js
grabkowski: that would not work as I have to check if it's an Option first.
I will extract part of the code into isExpectedValue method that could be reused elsewhere.

src/components/select/multi-select/multi-select.component.js
grabkowski: moved to a separate file

src/components/select/utils/is-expected-option.js
edleeks87: Yeah probably more readable than filtering the array anyway

edleeks87 (approved):

nicktitchmarsh (approved):

Alec-Walker-Sage:
The issue https://github.com/Sage/carbon/issues/3475. This code sandbox shows how it has fixed the issue https://codesandbox.io/s/carbon-quickstart-forked-7oq6v

cypress/features/regression/designSystem/simpleSelect.feature
```diff
@@ -76,4 +76,17 @@ Feature: Design System Select component
...
+      And Select list "Lazy Loaded A1" option is visible
+
+  @positive
+  Scenario: Check that using an object as a value displays the correct options
+    Given I open "Design System Select" component page "with object as value" in no iframe
+    When I click on dropdown button
+    Then visible options on Select list are "Amber", "Black", "Blue"
+
+  @positive
+  Scenario: Check that options can be selected and isplayed correctly when using an object as a value
```
DlgSHi, line 88:
```suggestion
  Scenario: Check that options can be selected and displayed correctly when using an object as a value
```

cypress/locators/select/locators.js
```diff
@@ -8,3 +8,4 @@ export const SELECT_LIST = '[data-element="select-list"]';
 export const SELECT_BASIC = '[id="simple"]';
 export const OPEN_ON_FOCUS_ID = '[id="openOnFocus"]';
 export const IS_LOADING_ID = '[id="isLoading"]';
+export const SELECT_INPUT = '[data-element="input"]';
```
DlgSHi, line 11: we don't need to make this as a separate locator, we can use `getDataElementByValue()` function

Alec-Walker-Sage (dismissed):

edleeks87 (approved):

Alec-Walker-Sage (approved):

sjohnson-sage (approved):

PR #3598, opened 2021-01-20
feat(toast): add ability to change default max width - FE-3504

bulaj:
### Proposed behaviour
- Max width of a toast is customizable
- Exposed "maxWidth" prop for setting custom max width of a toast.
image

### Current behaviour
- Max width of a toast is 300px.
- No ability to set custom max width of a toast.
image

### Checklist

- [ ] <del> Cypress automation tests added or updated if required

### Additional context
First PR!

### Testing instructions
See custom-width and Custom Width stories within Toast directory.


src/components/toast/toast.stories.js
```diff
@@ -55,6 +55,32 @@ Basic.story = {
   },
 };
 
+export const CustomWidth = () => {
```
DlgSHi, line 58: I don't think we need to create a story here - let's make it in the `mdx`

src/components/toast/toast.component.js
```diff
@@ -139,13 +148,16 @@ Toast.propTypes = {
...
+  /** Custom Toast width in pixels. */
+  width: PropTypes.number,
 };
 
 Toast.defaultProps = {
   as: "warning",
   onDismiss: null,
   open: true,
   isCenter: false,
+  width: null,
```
dawidzarzycki, line 160: I think its not necessary here

src/components/toast/toast.spec.js
```diff
@@ -189,6 +190,14 @@ describe("Toast", () => {
       elementsTagTest(wrapper.find(IconButton).first().find("span"), ["close"]);
     });
   });
+
+  describe("when toast has specified width", () => {
+    it("should render ToastStyle with correct width and no maxWidth", () => {
+      const wrapper = enzymeFind(mount(<Toast width={200} />), ToastStyle);
+      expect(wrapper).toHaveStyleRule("max-width", undefined);
+      expect(wrapper).toHaveStyleRule("width", "200px");
```
dawidzarzycki, line 198: I would suggest to use our `assertStyleMatch` function that can be imported from `__spec_helper__/test-utils`
And then the matcher would look like this
```js
assertStyleMatch({
  maxWidth: undefined,
  width: '200px'
})
```

src/components/toast/toast.stories.js
dawidzarzycki: I agree with @DlgSHi the `MDX` is enough for this example

src/components/toast/toast.style.js
```diff
@@ -25,9 +25,16 @@ const ToastStyle = styled(MessageStyle)`
   box-shadow: 0 10px 30px 0 rgba(0,20,29,.1), 0 30px 60px 0 rgba(0,20,29,.1);
   line-height: 22px;
   margin-top: 30px;
-  max-width: 300px;
   position: relative;
   margin-right: ${({ isCenter }) => (isCenter ? "auto" : "30px")};
+  ${({ customWidth }) =>
```
dawidzarzycki, line 30: I would suggest to move props to the top of the component and list them like
```js
const ToastStyle = styled(MessageStyle)`
 ${({isCenter, customWidth}) => {
  box-shadow: 0 10px 30px 0 rgba(0,20,29,.1), 0 30px 60px 0 rgba(0,20,29,.1);
  line-height: 22px;	  line-height: 22px;
  margin-top: 30px;	  margin-top: 30px;
  max-width: 300px;	
  position: relative;	  position: relative;
  margin-right: ${isCenter ? 'auto' : '30px'}	
  //... code  
 }}

```

src/components/toast/toast.style.js
```diff
@@ -25,9 +25,16 @@ const ToastStyle = styled(MessageStyle)`
...
+    customWidth
```
dawidzarzycki, line 31: I think I would keep it in one line 
```js
${customWidth => customWidth ? css`width: ${customWidth}px` : css`max-width: 600px`}
```

src/components/toast/toast.component.js
```diff
@@ -72,12 +72,21 @@ class Toast extends React.Component {
...
+      onDismiss,
+      children,
+      width,
+    } = this.props;
 
     const toastProps = {
       isCenter,
       variant: variant || as,
       id,
+      customWidth: width,
```
grabkowski, line 89: is there a reason to not pass it as `width`?

bulaj: Styled-components filter out custom props only, those that have same name as html attributes are passed down to the element, resulting in:
`<div width="200"/>`

jamime (changes requested):
Welcome @bulaj 👋  

I had a bit more background on this than most, it appears the JIRA ticket didn't capture the requirements fully. 

I spoke to @harpalsingh to confirm:

- `max-width` should be overridable with the default being `300px`. We don't want our existing messages which currently wrap at `300px` to now wrap at `600px`.
- No need for the `width` prop, it should automatically stretch up to the `max-width`.

So the final interface should be `<Toast maxWidth="">`. 

We don't need to restrict the value of the `maxWidth` prop, we should accept any string. It'll be up to the consumers to ensure they use a value that adheres to the DLS.

This will give us a sensible default but allow some customisation where needed. Sorry we didn't catch this in refinement.

jamime (dismissed):

src/components/toast/toast.spec.js
```diff
@@ -189,14 +190,21 @@ describe("Toast", () => {
       elementsTagTest(wrapper.find(IconButton).first().find("span"), ["close"]);
     });
   });
+
+  describe("when toast has specified maximum width", () => {
+    it("should render ToastStyle with correct maxWidth", () => {
+      const wrapper = enzymeFind(mount(<Toast maxWidth="200px" />), ToastStyle);
```
dawidzarzycki, line 196: I think that we don't need enzymeFind

jamime (approved):

dawidzarzycki (approved):

KatarzynaQA (approved):

PR #3599, opened 2021-01-21
refactor(heading): replace sass with styled component (FE-3082)

nicktitchmarsh:
Fixes FE-3082

### Current behaviour
Uses sass for styling instead of Styled Components

### Checklist

### Testing instructions

http://localhost:9001/?path=/docs/heading--without-divider#default
http://localhost:9001/?path=/story/design-system-card--default-story
http://localhost:9001/?path=/story/dialog--default-story
http://localhost:9001/?path=/story/dialog-full-screen--default-story
http://localhost:9001/?path=/story/pages--default-story
http://localhost:9001/?path=/story/settingsrow--default-story


sjohnson-sage:
The `__spec__.js` file has been deleted but not replaced with any other tests, there should now be a `heading.spec.js` file right?

src/components/heading/heading.style.js
```diff
@@ -64,4 +61,71 @@ StyledHeadingIcon.defaultProps = {
   theme: baseTheme,
 };
 
-export { StyledHeadingIcon, StyledHeading };
+const StyledSubHeader = styled.div`
+  margin-top: 5px;
+`;
+
+StyledSubHeader.defaultProps = {
+  theme: baseTheme,
```
sjohnson-sage, line 69: This isn't needed because `StyledSubHeader` doesn't use the theme

src/components/heading/heading.style.js
```diff
@@ -64,4 +61,71 @@ StyledHeadingIcon.defaultProps = {
...
+};
+
+const StyledHeader = styled.div`
+  position: relative;
+`;
+
+StyledHeader.defaultProps = {
+  theme: baseTheme,
```
sjohnson-sage, line 77: This isn't needed because `StyledHeader` doesn't use the theme

src/components/heading/heading.component.js
```diff
@@ -82,15 +92,16 @@ class Heading extends React.Component {
...
-      <Help
-        className="carbon-heading__help"
-        data-element="help"
-        tooltipAlign="center"
-        tooltipPosition="right"
-        href={this.props.helpLink}
-      >
-        {this.props.help}
-      </Help>
+      <StyledHeaderHelp>
```
sjohnson-sage, line 95: `StyledHeaderHelp` should be a `styled(Help)` so there's no need for the extra wrapping div here

src/components/heading/heading.component.js
```diff
@@ -114,15 +125,16 @@ class Heading extends React.Component {
...
-      <Link
-        // this event allows an element to be focusable on click event on IE
-        onMouseDown={(e) => e.currentTarget.focus()}
-        className="carbon-heading__back"
-        data-element="back"
-        {...props}
-      >
-        <StyledHeadingIcon type="chevron_left" />
-      </Link>
+      <StyledHeaderLink>
```
sjohnson-sage, line 128: `StyledHeaderLink` should be a `styled(Link)` so there's no need for the extra wrapping div here

src/components/heading/heading.component.js
```diff
@@ -185,32 +187,31 @@ class Heading extends React.Component {
...
       >
-        <div className="carbon-heading__header">
+        <StyledHeader>
           {this.back}
 
-          <div className="carbon-heading__headers">
-            <div className="carbon-heading__main-header">
-              <h1
-                className="carbon-heading__title"
+          <StyledHeaders back={this.props.backLink}>
```
sjohnson-sage, line 197: `StyledHeaders` should be a `styled(Box)` so we don't have another unnecessary wrapping div

src/components/heading/heading.component.js
```diff
@@ -185,32 +187,31 @@ class Heading extends React.Component {
...
+            <Box>
```
sjohnson-sage, line 198: Does this need to be a `Box`? Could probably just be a div saying as we're not using any `Box` props

src/components/heading/heading.style.js
```diff
@@ -64,4 +61,71 @@ StyledHeadingIcon.defaultProps = {
...
+};
+
+const StyledDivider = styled.hr`
```
sjohnson-sage, line 80: Why aren't we using our `Hr` component here?

sjohnson-sage: In fact this `StyledDivider` isn't needed at all, I'll comment above with the solution

src/components/heading/heading.component.js
```diff
@@ -138,40 +150,30 @@ class Heading extends React.Component {
...
-   * @method separator
+   * @method divider
    * @return {Object} JSX
    */
-  get separator() {
-    return this.props.separator ? (
-      <hr className="carbon-heading__separator" />
-    ) : null;
+  get divider() {
+    return this.props.divider ? <StyledDivider /> : null;
```
sjohnson-sage, line 176: We should probably be using our `Hr` component, then we wouldn't need to duplicate the styles. So rather than `<StyledDivider />` - `<Hr m="15px 0 20px" />`

src/components/heading/heading.style.js
```diff
@@ -64,4 +61,71 @@ StyledHeadingIcon.defaultProps = {
...
+StyledSeparator.defaultProps = {
+  theme: baseTheme,
+};
+
+const StyledHeaders = styled.div`
+  margin-left: ${({ back }) => (back ? "30px" : "")};
+`;
+
+const StyledHeaderLink = styled.div`
+  margin-top: -14px;
```
sjohnson-sage, line 108: Do we really need this? I'm against using negative margins where possible

DipperTheDan:
Good day. Seems to be odd styling with `Alert` on this branch. Looks to have two horizontal dividers.
Screenshot 2021-01-25 at 11 43 36

src/components/heading/heading.style.js
DipperTheDan: Without this line of the code that backlink arrow looks like this ⬇️ 

Screenshot 2021-01-25 at 11 48 49

So if it were to be removed, we'd need to find a way of correcting the position.

sjohnson-sage: Yeah that's what I meant really sorry, my comment was is there a better way of doing this positioning than using a negative margin? I haven't spend any time looking into it, but usually there is a better way

DipperTheDan:
Good day. It appears that Dialog has a new styling issue. 

Screenshot 2021-01-26 at 16 54 34

Screenshot 2021-01-26 at 16 52 08

DipperTheDan:
Also, Dialog Full Screen has an additional underline.

src/components/heading/heading.component.js
grabkowski: wrapper removed

src/components/heading/heading.component.js
grabkowski: wrapper removed

src/components/heading/heading.component.js
grabkowski: good point, changed

src/components/heading/heading.component.js
grabkowski: Box replaced with div

src/components/heading/heading.style.js
grabkowski: removed

src/components/heading/heading.style.js
grabkowski: removed

src/components/heading/heading.style.js
grabkowski: replaced with Hr

src/components/heading/heading.style.js
grabkowski: changed display to flexbox

sjohnson-sage (commented):
Code looks good to me, happy to approve once you've addressed the visual points @DipperTheDan found

sjohnson-sage (approved):

src/components/pages/pages.stories.mdx
```diff
@@ -3,7 +3,7 @@ import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
 import DefaultPages from './pages.component';
 import Page from './page/page.component';
 import DialogFullScreen from '../dialog-full-screen';
-import Heading from '../heading/heading';
+import Heading from '../heading';
 import Button from '../button';
```
DipperTheDan, line 8: The other visual defects have been rectified but this visual defect has appeared with `Pages` now.

Screenshot 2021-01-28 at 11 19 28

DipperTheDan (changes requested):

DipperTheDan (dismissed):

src/components/dialog-full-screen/dialog-full-screen.component.js
```diff
@@ -81,6 +81,7 @@ class DialogFullScreen extends Modal {
           titleId="carbon-dialog-title"
           subheader={this.props.subtitle}
           subtitleId="carbon-dialog-subtitle"
+          divider={false}
```
sjohnson-sage, line 84: Do we still need this then?

src/components/pages/pages.stories.js
```diff
@@ -69,22 +69,32 @@ export const Basic = () => {
       <Button onClick={handleOpen}>Open Preview</Button>
       <DialogFullScreen pagesStyling open={isOpen} onCancel={handleCancel}>
         <Pages initialpageIndex={initialpageIndex} pageIndex={pageIndex}>
-          <Page title={<Heading title="My First Page" />}>
+          <Page title={<Heading title="My First Page" divider={false} />}>
```
sjohnson-sage, line 72: and do we need this?

src/components/dialog-full-screen/dialog-full-screen.component.js
grabkowski: we have two options:
1. set the `divider={false}` in the component, which will result in not rendering the divider.
2. set the display on the divider to `none`, which will result in hiding the divider.

IMO 1. is a better option

sjohnson-sage (approved):

DipperTheDan (dismissed):

DipperTheDan (approved):

sjohnson-sage (dismissed):

sjohnson-sage (dismissed):

DipperTheDan (dismissed):

sjohnson-sage (dismissed):

DipperTheDan (approved):

DipperTheDan (dismissed):

sjohnson-sage (dismissed):

lukeemmerson-finch:
As discussed with Kuba, i'm happy to pass UXQA on this and for us to use HR everywhere instead of a border style. However, we will need to create a separate bug fix ticket as Kuba mentioned that the style of the HR changes across windows and mac. Please could QA generate a ticket and add to the backlog for this?

DipperTheDan (approved):

DlgSHi (approved):

PR #3604, opened 2021-01-22
fix(focus-trap): add focus trap component and remove util function FE-3586

edleeks87:
BREAKING CHANGE: the focus trap util function has been removed

### Proposed behaviour

The `focusTrap` util does not support the correct tab order if a disabled component becomes active,
it has been removed and the functionality has been moved into an internal component that is composed
into the `Modal` based components

### Current behaviour

The `focusTrap` util exists there is no component

### Checklist

### Testing instructions

The modal based stories will need to be checked to ensure they function as before, the `editable` `Dialog` story has been updated to show that it now supports tab ordering when disabled components are enabled

https://codesandbox.io/s/carbon-quickstart-forked-eeo0w?file=/src/index.js


edleeks87:
> Good day, There appears to be bug in the Dialog -> editable MDX example.
> 
> Dialog_editable_bug

Good spot I'll make sure the radio buttons are wired up

DipperTheDan (dismissed):

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -0,0 +1,118 @@
...
+  }, [firstElement, lastElement, focusableElements, bespokeTrap]);
+
+  return <div ref={ref}>{children}</div>;
+};
+
+FocusTrap.propTypes = {
+  children: PropTypes.node.isRequired,
+  autoFocus: PropTypes.bool,
+  focusFirstElement: PropTypes.func,
+  bespokeTrap: PropTypes.func,
```
mkrds, line 115: I'd write prop descriptions here, as I'm not really sure what `bespokeTrap` does unless I check out the `FocusTrap` code

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -0,0 +1,118 @@
...
+  }, [children, hasNewInputs]);
+
+  useEffect(() => {
+    if (autoFocus && firstOpen.current && (focusFirstElement || firstElement)) {
+      setElementFocus(focusFirstElement || firstElement);
+      firstOpen.current = false;
+    }
+  }, [autoFocus, firstElement, focusFirstElement]);
+
+  useEffect(() => {
```
mkrds, line 63: Is it necessary to run `ev.preventDefault()` and `setElementFocus` in cases when its not a last/first element?
I think when we're let's say on 3rd input out of 10 in `Dialog`, and press `Tab` then the default browser bevahiour would be just fine wouldnt it?

edleeks87: It is necessary to run it unfortunately, as when a disabled element becomes enabled it is not included in the tab order as in the gif in the PR description. Even when you manually set a tabIndex of 0 when it becomes enabled it still doesn't include it until the Modal is closed and re-opened

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -0,0 +1,118 @@
...
+    };
+  }, [firstElement, lastElement, focusableElements, bespokeTrap]);
+
+  return <div ref={ref}>{children}</div>;
+};
+
+FocusTrap.propTypes = {
+  children: PropTypes.node.isRequired,
+  autoFocus: PropTypes.bool,
+  focusFirstElement: PropTypes.func,
```
mkrds, line 114: Is it always meant to be a function? I can't pass a ref to the element that I want to be focused first?

edleeks87: Discussed privately, the components that use it currently use callback refs but I have expanded the prop type to support object refs as well

DipperTheDan (approved):

mkrds (approved):

DipperTheDan (dismissed):

mkrds (dismissed):

DipperTheDan (dismissed):

Alec-Walker-Sage (dismissed):

DipperTheDan (approved):

Alec-Walker-Sage (approved):

Issue #1756, opened 2018-06-01
Button as a Link accessibility


toni-sharpe:
`<Button href="/">My Link</Button>` will look like a button but render an `<a>` tag.

This pen demonstrates how the interactions are different:

https://codepen.io/tonileigh303/pen/LrVbmK

A link doesn't respond to a space bar so using it as the tag under something that looks like a button is an accessibility issue.

There's also some further stuff about how links should navigate and buttons should cause events, so some UX input here might be good to make sure we're not facilitating unexpected accessibility issues with our design.

andrewjtait:
Accessibility wise though, does it necessarily matter what it looks like? For screen readers or other things that parse the HTML it will recognise it for it is, not what it looks like.

Also as we move to more dialog based UI, or "single page app" style, we will have less page navigation from clicking buttons and they will trigger events instead. Having the `Button` seamlessly transition from an anchor to a button but still looking the same could help us by not providing an unfamiliar interface when it changes, but providing the correct accessibility when it changes.

toni-leigh:
I think it matters for keyboard users, who could assume that the thing that looks like a button will respond to the same keypresses as the other thing that looks like a button (this was raised at FENE conference as part of a general 'think about your HTML' message).

A solution would probably be to make anything that looks like a button respond to the `space` bar like a `<button>`

andrewjtait:
> A solution would probably be to make anything that looks like a button respond to the space bar like a <button>

👍 

From [MDN's docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role#Keyboard_and_focus)

> Warning: Be careful when marking up links with the button role. Buttons are expected to be triggered using the Space or Enter key, while links are expected to be triggered using the Enter key. In other words, when links are used to behave like buttons, adding role="button" alone is not sufficient. It will also be necessary to add a key event handler that listens for the Space key in order to be consistent with native buttons.

DipperTheDan:
FE-1788

DipperTheDan:
@tempertemper - Just tagging you in this issue as it is in relation to accessibility. Have you got anything to add to this issue? As the issue from 2018 and the solution was also suggested in 2018, the world of accessibility may well have moved on. Your input would be greatly appreciated so we can get the best solution possible.

Thanks 😄

sjohnson-sage:
[...]

PR #3606, opened 2021-01-25
docs(routing): add usage with routing docs

sjohnson-sage:
Fixes: #1756 

### Proposed behaviour
Remove Carbon support for rendering router links in `Button`, `Link` and `MenuItem`. 

Instead provide a documented example of how users can use Carbon components as links in React Router versions 5 and 6. This can be seen at: http://localhost:9001/?path=/docs/documentation-usage-with-routing--page

Also the `onClick` prop for `Link` will no longer render it as a `button` element, it will always render an `a` element now.

Fixed the accessibility for a `Button` rendering as a link. It now has `role="button"` and `draggable="false"` on the `<a>` element and pressing the space bar will follow the link rather than scroll the page

### Current behaviour
`Button`, `Link` and `MenuItem` have custom implementations on rendering router links via props.

### Checklist

### Testing instructions
http://localhost:9001/?path=/docs/documentation-usage-with-routing--page

To test `Button` rendering as a link http://localhost:9001/?path=/story/design-system-button--as-a-link

### DIFF
--- a/docs/usage-with-routing.stories.mdx
+++ b/docs/usage-with-routing.stories.mdx
@@ -1,0 +1,133 @@
+<Meta title="Documentation/Usage with routing" />
+
+# Using Carbon components with routing libraries
+
+## Contents
+
+- [Usage with React Router v5](#usage-with-react-router-v5)
+- [Usage with React Router v6](#usage-with-react-router-v6)
+- [Supported components](#supported-components)
+
+## Introduction
+
+There is no code in Carbon to implement a specific routing library, we aim to let the user decide which library to use.
+
+React Router is one of the most popular routing libraries so we have provided examples below to help with implementing Carbon components as links using both v5 and v6 of this library. There are small but important differences in the implementation of each version.
+
+## Usage with React Router v5
+
+An example of how to use the Carbon `Button` and `Link` components as links with React Router v5 is shown below, this can be applied to any of the supported Carbon components.
+Note the use of the `useHistory` hook from `react-router-dom`.
+
+```jsx
+import React, { useCallback } from "react";
+import Button from "carbon-react/lib/components/button";
+import Link from "carbon-react/lib/components/link";
+
+import { BrowserRouter, Switch, Route, useHistory } from "react-router-dom";
+
+const CarbonButtonWithReactRouter = ({ to, children }) => {
+  const history = useHistory();
+  const onClick = useCallback(() => {
+    history.push(to);
+  }, [to, history]);
+
+  return <Button onClick={onClick}>{children}</Button>;
+};
+
+const CarbonLinkWithReactRouter = ({ to, children }) => {
+  const history = useHistory();
+  const onClick = useCallback(() => {
+    history.push(to);
+  }, [to, history]);
+
+  return <Link onClick={onClick}>{children}</Link>;
+};
+
+const Home = () => (
+  <div>
+    <CarbonButtonWithReactRouter to="settings">
+      Settings
+    </CarbonButtonWithReactRouter>
+  </div>
+);
+
+const Settings = () => (
+  <div>
+    <CarbonLinkWithReactRouter to="/">Home</CarbonLinkWithReactRouter>
+  </div>
+);
+
+const App = () => (
+  <BrowserRouter>
+    <Switch>
+      <Route exact path="/" element={<Home />} />
+      <Route path="/settings" element={<Settings />} />
+    </Switch>
+  </BrowserRouter>
+);
+```
+
+## Usage with React Router v6
+
+An example of how to use the Carbon `Button` and `Link` components as links with React Router v6 is shown below, this can be applied to any of the supported Carbon components.
+Note the use of the `useNavigate` hook from `react-router-dom`.
+
+```jsx
+import React, { useCallback } from "react";
+import Button from "carbon-react/lib/components/button";
+import Link from "carbon-react/lib/components/link";
+
+import { BrowserRouter, Routes, Route, useNavigate } from "react-router-dom";
+
+const CarbonButtonWithReactRouter = ({ to, children }) => {
+  const navigate = useNavigate();
+  const onClick = useCallback(() => {
+    navigate(to);
+  }, [to, navigate]);
+
+  return <Button onClick={onClick}>{children}</Button>;
+};
+
+const CarbonLinkWithReactRouter = ({ to, children }) => {
+  const navigate = useNavigate();
+  const onClick = useCallback(() => {
+    navigate(to);
+  }, [to, navigate]);
+
+  return <Link onClick={onClick}>{children}</Link>;
+};
+
+const Home = () => (
+  <div>
+    <CarbonButtonWithReactRouter to="settings">
+      Settings
+    </CarbonButtonWithReactRouter>
+  </div>
+);
+
+const Settings = () => (
+  <div>
+    <CarbonLinkWithReactRouter to="/">Home</CarbonLinkWithReactRouter>
+  </div>
+);
+
+const App = () => (
+  <BrowserRouter>
+    <Routes>
+      <Route path="/" element={<Home />} />
+      <Route path="settings" element={<Settings />} />
+    </Routes>
+  </BrowserRouter>
+);
+```
+
+## Supported components
+
+There are many components that you can use this method with, including:
+
+- `ActionPopoverItem`
+- `Button`
+- `Link`
+- `MenuItem`
+- `Tab`
--- a/src/components/button/button.component.js
+++ b/src/components/button/button.component.js
@@ -14,17 +14,11 @@ const renderStyledButton = (buttonProps) => {
     iconType,
     theme,
     forwardRef,
-    href,
     px,
     size,
     noWrap,
     ...styleProps
   } = buttonProps;
-
-  // added to support legacy link buttons
-  if (href) {
-    styleProps.href = href;
-  }
 
   let paddingX;
 
@@ -41,7 +35,6 @@ const renderStyledButton = (buttonProps) => {
 
   return (
     <StyledButton
-      as={!disabled && href ? "a" : "button"} // legacy link button feature
       buttonType={buttonType}
       disabled={disabled}
       role="button"
@@ -60,7 +53,7 @@ const renderStyledButton = (buttonProps) => {
 };
 
 const Button = (props) => {
-  const { disabled, to, iconType, renderRouterLink, size, subtext } = props;
+  const { size, subtext } = props;
 
   const { as, buttonType, ...rest } = props;
 
@@ -71,15 +64,6 @@ const Button = (props) => {
 
   if (subtext.length > 0 && size !== "large") {
     throw new Error("subtext prop has no effect unless the button is large");
-  }
-
-  // added to support legacy link buttons
-  if (!disabled && to && renderRouterLink) {
-    return renderRouterLink({
-      to,
-      type: iconType,
-      children: renderStyledButton(propsWithoutAs),
-    });
   }
 
   return renderStyledButton(propsWithoutAs);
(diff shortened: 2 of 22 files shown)


docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
+<Meta title="Documentation/Usage with routing" />
```
jamime, line 1: We can ignore this story from chromatic.

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Contents
+
+- [Usage with React Router v5](#usage-with-react-router-v5)
+- [Usage with React Router v6](#usage-with-react-router-v6)
+- [Supported components](#supported-components)
+
+## Introduction
+
+There is no code in Carbon to implement a specific routing library, we aim to let the user decide which one they want to use and allow some Carbon components to be used as links (listed in the [Supported components](#supported-components) section below).
```
jamime, line 13:
```suggestion
There is no code in Carbon to implement a specific routing library, we aim to let the user decide which library to use.
```

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+React Router is one of the most popular routing libraries so we have provided examples below to help with implementing Carbon components as links using both v5 and v6 of this library. There are small but significant differences in the implementation of each version.
```
jamime, line 15:
```suggestion
React Router is one of the most popular routing libraries so we have provided examples below to help with implementing Carbon components as links using both v5 and v6 of this library. There are small but important differences in the implementation of each version.
```

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Usage with React Router v5
+
+An example of how to use the Carbon `Button` and `Link` components as links with React Router v5 is shown below, this can be applied to any of the suuported Carbon components.
+Note the use of the `useHistory` hook from `react-router-dom` and the `exact` prop on each `Route`
```
jamime, line 20:
```suggestion
Note the use of the `useHistory` hook from `react-router-dom`.
```

The exact prop is only required for the first route and is not unique to this situation.

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+  <div>
+    <CarbonLinkWithReactRouter to="/">Home</CarbonLinkWithReactRouter>
+  </div>
+);
+
+const App = () => (
+  <BrowserRouter>
+    <Switch>
+      <Route exact path="/" element={<Home />} />
+      <Route exact path="/settings" element={<Settings />} />
```
jamime, line 65:
```suggestion
      <Route path="/settings" element={<Settings />} />
```

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+  <BrowserRouter>
+    <Routes>
+      <Route path="/" element={<Home />} />
+      <Route path="settings" element={<Settings />} />
+    </Routes>
+  </BrowserRouter>
+);
+```
+
+## Supported components
```
jamime, line 125:
```suggestion
## Supported components

There are many components that you can use this method with, including:
```

src/components/button/button.component.js
```diff
@@ -155,10 +146,6 @@ Button.propTypes = {
   as: PropTypes.oneOf(OptionsHelper.themesBinary),
   /** Legacy - used to transform button into anchor */
   href: PropTypes.string,
```
jamime, line 148: Should we remove `href` too?

src/components/link/link.component.js
```diff
@@ -96,9 +89,7 @@ class InternalLink extends React.Component {
   createLinkBasedOnType = () => {
```
jamime, line 89: I think this would also be a good time to make Links always return an anchor element.

jamime (changes requested):

docs/usage-with-routing.stories.mdx
sjohnson-sage: I don't think these docs ones get added to chromatic, at least I didn't see one getting added in the UI review check, and none of the others have ignore on either

src/components/button/button.component.js
sjohnson-sage: Yeah good plan, will do

docs/usage-with-routing.stories.mdx
jamime: Ah yeah, because it doesn't have `<Story>`. Sorry about the confusion.

src/components/heading/__spec__.js
```diff
@@ -74,7 +74,7 @@ describe("Heading", () => {
     );
 
     const link = wrapper.find(Link);
-    link.find("button").simulate("mousedown");
+    link.find("a").simulate("mousedown");
```
jamime, line 77: This is interesting, so the back arrow will now render as an `a`. We should check if we should be using a button here or an anchor.

image

src/components/link/link.component.js
```diff
@@ -120,7 +97,11 @@ class InternalLink extends React.Component {
         hasContent={Boolean(children)}
         {...tagComponent("link", this.props)}
       >
-        {this.createLinkBasedOnType()}
+        {React.createElement(
```
jamime, line 100: No need for `React.createElement` we can replace this with JSX now that the element doesn't change.

jamime (changes requested):

src/components/heading/__spec__.js
sjohnson-sage: As discussed elsewhere we should be using an `a` here

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,133 @@
...
+      <Route exact path="/" element={<Home />} />
+      <Route path="/settings" element={<Settings />} />
+    </Switch>
+  </BrowserRouter>
+);
+```
+
+## Usage with React Router v6
+
+An example of how to use the Carbon `Button` and `Link` components as links with React Router v6 is shown below, this can be applied to any of the suuported Carbon components.
```
edleeks87, line 73:
```suggestion
An example of how to use the Carbon `Button` and `Link` components as links with React Router v6 is shown below, this can be applied to any of the supported Carbon components.
```

docs/usage-with-routing.stories.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+## Introduction
+
+There is no code in Carbon to implement a specific routing library, we aim to let the user decide which library to use.
+
+React Router is one of the most popular routing libraries so we have provided examples below to help with implementing Carbon components as links using both v5 and v6 of this library. There are small but important differences in the implementation of each version.
+
+## Usage with React Router v5
+
+An example of how to use the Carbon `Button` and `Link` components as links with React Router v5 is shown below, this can be applied to any of the suuported Carbon components.
```
edleeks87, line 19:
```suggestion
An example of how to use the Carbon `Button` and `Link` components as links with React Router v5 is shown below, this can be applied to any of the supported Carbon components.
```

edleeks87 (approved):

src/components/link/link.stories.mdx
```diff
@@ -1,146 +1,146 @@
...
-    </Link>
+  <Story name="with onClick" parameters={{ chromatic: { disable: true } }}>
+    <Link onClick={() => {}}>This is a link</Link>
   </Story>
 </Preview>
 
 ### Target
-Use the `target` prop to modify the behaviour when the Link component is clicked. In this example, target has been assigned '_blank'.
+
+Use the `target` prop to modify the behaviour when the Link component is clicked. In this example, target has been assigned '\_blank'.
```
jamime, line 134: I think this should be a code snippet.
```suggestion
Use the `target` prop to modify the behaviour when the Link component is clicked. In this example, target has been assigned `_blank`.
```

src/components/link/link.stories.mdx
```diff
@@ -1,146 +1,146 @@
...
 
-<Meta title="Link" parameters={{ info: { disable: true }}} />
+<Meta title="Link" parameters={{ info: { disable: true } }} />
 
 # Link
+
 Navigates the user to another location.
 
-Avoid using links for commands (performing an action like saving a form) - use them for navigating the user to a new location. 
+Avoid using links for commands (performing an action like saving a form) - use them for navigating the user to a new location.
```
jamime, line 11: Should we link to the routing documentation here? We don't want people that are using react-router to accidentally use `href` because it'll cause a full page refresh.

jamime (commented):
I think we can now remove `react-router-dom` from `package.json`.

jamime (dismissed):

edleeks87 (approved):

sianford (dismissed):

sianford:
Tested no Safari and Chrome. Link `onClick` works in an `a` element as expected and the `onClick` action event is recorded correctly. Documentation has been updated as described. Button and Link components work as expected. No regressions, no automation errors. Testing passed ✅

edleeks87 (dismissed):

src/components/button/button.component.js
```diff
@@ -41,97 +96,46 @@ const renderStyledButton = (buttonProps) => {
 
   return (
     <StyledButton
-      as={!disabled && href ? "a" : "button"} // legacy link button feature
+      as={!disabled && href ? "a" : "button"}
```
jamime, line 99: This means that disabled links would render as a button. Should we be handling this on the `a` element instead by setting the cursor and blocking the click event?

Not blocking for this PR.

src/components/button/button.component.js
```diff
@@ -41,97 +96,46 @@ const renderStyledButton = (buttonProps) => {
...
+      type={href ? undefined : "button"}
       iconType={iconType}
       size={size}
       px={px || paddingX}
       noWrap={noWrap}
       {...tagComponent("button", buttonProps)}
-      {...styleProps}
-      ref={forwardRef}
+      {...rest}
+      ref={forwardRef || linkRef}
```
jamime, line 112: If you provide a forwardRef I think this would break the `linkRef.current.click`.

src/components/button/button.component.js
```diff
@@ -35,17 +96,20 @@ const renderStyledButton = (buttonProps) => {
 
   return (
     <StyledButton
+      as={!disabled && href ? "a" : "button"}
+      onKeyDown={href && handleLinkKeyDown}
+      draggable={false}
```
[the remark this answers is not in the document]
sjohnson-sage: Normally yes, but if you look at Martin's comment on this: https://github.com/Sage/carbon/issues/1756 - it shouldn't in this particular scenario because we want it to behave as a button, not a link

src/components/button/button.component.js
sjohnson-sage: Good point I'll change that

src/components/button/button.component.js
sjohnson-sage: Sorted above by checking for `forwardRef` when space is pressed and if it exists, do `forwardRef.current.click();` instead

src/components/button/button.component.js
```diff
@@ -41,97 +98,46 @@ const renderStyledButton = (buttonProps) => {
...
+  const { size, subtext } = props;
+
+  const linkRef = useRef(null);
 
   const { as, buttonType, ...rest } = props;
 
   const propsWithoutAs = {
     ...rest,
     buttonType: buttonType || as,
+    linkRef,
```
jamime, line 131: We can simplify `renderStyledButton` if we move the conditional into this function.

```js
  const { as, buttonType, forwardRef ...rest } = props;
  const propsWithoutAs = {
    ...rest,
    buttonType: buttonType || as,
    ref: forwardRef || ref
  };
```

jamime (commented):
Looks good to me, I left one minor comment but happy to approve as is.

src/components/button/button.component.js
sjohnson-sage: Changed!

edleeks87 (approved):

jamime (dismissed):

jamime (approved):

edleeks87 (dismissed):

DlgSHi (approved):

edleeks87 (approved):

Issue #3589, opened 2021-01-19
Selected tab overlaps menu dropdown items on hovering


SivalenSivalingam:
### Current behaviour
We have 3 tab component under a navigation menu item and list of reports. When a tab is selected and you try and hover over the menu item to show the list of reports. The tab selected element overlaps the dropdown menu item.

### Expected behaviour
When you select a tab and try and hover over a menu drop down to show a list of reports, the drop down list should take preference over the selected tab element

### Reproducible example
https://codesandbox.io/s/great-cache-t5fm1

### Additional context
This only happen when you hover over the menu item when a tab is selected

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 48.3.0
| react-scripts    | 3.4.3
| React            | 16.13.1
| react-router-dom | 5.2.0
| react-dom        | 16.13.1

sjohnson-sage:
FE-3598

Issue #3588, opened 2021-01-18
Menu - When a submenu overlaps the search component it is no accessible with the mouse


philearley:
### Current behaviour

We have a use case where the Navigationbar component has a Menu with a sub menu.
We also have a Table that is searchable.

When you expand the Menu showing its contents, if it covers the Search component it is not accessible with the mouse.

An example of this behaviour can be found here:  https://codesandbox.io/s/elegant-forest-8096m

### Expected behaviour

The Menu to have a priority in position over any elements the sub menu is rendered on top of.

### Additional context

I have not tested with other components, but I have seen in demos where parts of the elements within the Table are hiding the bottom of the sub menu.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 58.1.0

sjohnson-sage:
FE-3598

PR #3617, opened 2021-01-28
fix(menu): update zIndex - FE-3598

dawidzarzycki:
### Proposed behaviour
update `z-index` to fix overlays bugs related to menu component
fixes: #3588
fixes: #3589

### Current behaviour
Current z-indexes causes issues listed above

### Checklist

- [ ] <del> Unit tests added or updated if required
- [ ] <del> Cypress automation tests added or updated if required


src/components/portal/portal.style.js
```diff
@@ -4,7 +4,7 @@ import Portal from "./portal";
 const StyledPortal = styled(Portal)`
   position: absolute;
   top: 0;
-  z-index: 1002;
+  z-index: 3000;
```
mkrds, line 7: Shouldnt we use a value defined in `theme` here as well?

src/components/menu/menu.style.js
```diff
@@ -11,6 +11,7 @@ const StyledMenuWrapper = styled(Box).attrs({ as: "ul" })`
   list-style: none;
   margin: 0;
   padding: 0;
+  z-index: ${({ theme }) => theme.zIndex.popover};
```
sjohnson-sage, line 14: I think you should be adding this to the submenu, not the whole `Menu`. If you look in `src/menu/__internal__/submenu/submenu.style.js` you'll see `z-index: 10` being applied to the `StyledSubmenu` - I think that's what you should be changing here

src/components/portal/portal.style.js
sjohnson-sage: Why are you even changing this here? The submenus don't use `Portal` so this is unrelated to the bugs you're fixing

dawidzarzycki: but without the changes if you will use Menu and Dialog, then Menu will appear over the Dialog background

dawidzarzycki: @sjohnson-sage You are completely right! We don't need to set zIndex for whole menu, submenu is enough and because of that I don't need to change portal's zIndex!

mkrds (approved):

sjohnson-sage (approved):

src/__experimental__/components/search/search.style.js
```diff
@@ -194,7 +194,7 @@ export const StyledSearchButton = styled.div`
     margin: 0px 0px;
     padding-bottom: 3px;
     :focus {
-      z-index: ${({ theme }) => theme.zIndex.overlay};
+      z-index: ${({ theme }) => theme.zIndex.smallOverlay};
```
sjohnson-sage, line 197: I think you should change the other `z-index` in this file to be `smallOverlay` as well

mkrds (approved):

sjohnson-sage (approved):

KatarzynaQA (approved):

PR #3618, opened 2021-01-29
docs(styled-system): prop generator now supports margin and padding subsets FE-3619

edleeks87:
### Proposed behaviour

Update the `space` test util to support when only a subset (margin/ padding) of the props are used for a component.
Update the `StyledSystemProps` to support rendering only `margin`/ `padding` subset of `space` props

### Current behaviour

We only support applying all the `space` props to a component, the test-util and `StyledSystemProps` are not configurable for the subsets

### Checklist

### Additional context

Margin only prop table:

image

Padding only prop table:

image

### Testing instructions

Check prop tables render as before


src/__spec_helper__/test-utils.js
```diff
@@ -364,6 +507,8 @@ export {
   carbonThemesJestTable,
   mockMatchMedia,
   testStyledSystemSpacing,
```
sjohnson-sage, line 509: I think that `testStyledSystemSpacing` can just be a function like 
```js
const testStyledSystemSpacing = (
  component,
  defaults,
  styleContainer,
  assertOpts
) => {
  testStyledSystemMargin(component, defaults, styleContainer, assertOpts);
  testStyledSystemPadding(component, defaults, styleContainer, assertOpts);
};
```

rather than duplicating the test logic

edleeks87: 🤦

sjohnson-sage (approved):

mkrds (approved):

PR #3635, opened 2021-02-02
fix(decimal): allow decimal to handle dead key characters - FE-3334

### LINKED ISSUE #3372: Decimal fields crashes app if "^" is used on Mac.
### Current behaviour
If you try to type ^ into a decimal carbon field on a Mac (Firefox, Chrome) it will crash the react app with an error saying that your entry is not a decimal number.

### Expected behaviour
Decimal fields should also permit typing of ^ on Mac, as it does on windows.

### Your environment
carbon-react Version 48.6.2
macOS Catalina 10.15.6
Firefox 82.0.3

DipperTheDan:
This fix allows the user to use a dead key character without causing the component to unexpectedly
crash when onBlur is triggered.

fixes #3372

### Proposed behaviour

This fix allows the user to use a dead key character without causing the component to unexpectedly
crash when onBlur is triggered.

Decimal_carrat_fix

### Current behaviour

If a dead key character is used and then the component is blurred, the component crashes.

Decimal_crash_carrat

Decimal_crash_carrat2

### Checklist

<del>[ ]  Storybook added or updated if required </del>
<del>[ ] Typescript `d.ts` file added or updated if required </del>

### Additional context
Please ensure you have your keyboard set to `German - Standard` on your Mac before testing the scenario set out in issue #3372 

### Testing instructions

- `git checkout FE-3334_decimal_fields_crashes`
- `npm i`
- `npm start` to start Storybook
- Navigate to Experimental -> Decimal -> default
- Try to replicate the scenario presented on issue #3372 *Please make sure you have your keyboard set to `German - Standard`*
- Component should not crash Storybook


src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -256,10 +256,13 @@ class Decimal extends React.Component {
    * Format a user defined value
    */
   formatValue = (value) => {
-    invariant(
-      !this.isNaN(value),
-      `The supplied decimal (${value}) is not a number`
```
jamime, line 261: Removing this means the component will error with `Cannot read property 'visibleValue' of undefined`.

Previously it would have errored with `The supplied decimal (12a.05) is not a number`.

```js
<Decimal label="Decimal" defaultValue="12a.05" />
```

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -1649,6 +1619,46 @@ describe("Decimal", () => {
...
+      "õ",
+      "ø",
+    ])("handles accented letter (%s) without error", (accentedLetter) => {
+      const onBlur = jest.fn();
+      render({ value: "20.00", onBlur });
+      setProps({ value: `${accentedLetter}` });
+      blur();
+      expect(onBlur).toHaveBeenCalled();
+      expect(value()).toBe("20.00");
+      expect(hiddenValue()).toBe("20.00");
```
jamime, line 1659: Do you need to assert that the component didn't error?

cypress/features/regression/experimental/decimalInput.feature
```diff
@@ -35,6 +35,6 @@ Feature: Decimal input component
     When I set Decimal input to <label>
     Then Decimal input is not set to <label>
     Examples:
-      | label                     |
-      | mpú¿¡üßä                  |
-      | !@#$%^*()_+=~[];:?{}&"'<> |
```
DlgSHi, line 40: I don't think the issue was about the `label`. I thought that it was about typing the `^` inside the input.

sjohnson-sage: Looks like the wording on the Cypress test is incorrect to me here, I think it should be `When I set Decimal input to <input>` etc - this test is actually testing special characters inside the input so Dan's change is correct here

DipperTheDan: Yeah. What @sjohnson-sage said. This test is testing the characters in the input field. It confused me a bit at first when I saw `label`. Maybe it's something we should look to change?

src/__experimental__/components/decimal/decimal.spec.js
DipperTheDan: My thinking behind this was that if anyone in the future makes any changes to the onBlur or the `formatValue`, we have tests to catch anything that will effect and deadkey characters. I'm happy to do this a different way if you know of a better way?

cypress/features/regression/experimental/decimalInput.feature
jamime: I think this will send the carat/backtick. It's not a problem, but we should be aware we're not testing the dead key.

src/__experimental__/components/decimal/decimal.spec.js
jamime: Thats fine, but I'm not sure if this test would fail if the component did error, please can you check?

Usually we'd use `expect().not.toThrow()`

src/__experimental__/components/decimal/decimal.component.js
DipperTheDan: Messaged you privately about this comment 😃

src/__experimental__/components/decimal/decimal.spec.js
DipperTheDan: No bother. I'll change the test to include `expect().not.toThrow()` as that is probably a better test for what I was wanting to achieve. Thanks for that 👍🏻

src/__experimental__/components/decimal/decimal.component.js
DipperTheDan: Reinstated the invariant as per the screenshot. 

Screenshot 2021-02-04 at 10 50 34

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -256,10 +256,17 @@ class Decimal extends React.Component {
    * Format a user defined value
    */
   formatValue = (value) => {
+    const isValid = this.isValidDecimal(value);
     invariant(
-      !this.isNaN(value),
+      !this.isNaN(value) || Events.isAccentKey(value),
```
jamime, line 261: You shouldn't need `isAccentKey` because all of the keys are `NaN`.

e.g. 

```js
Number.isNaN(Number("ŵ")) => true
```

src/utils/helpers/events/events.js
```diff
@@ -367,6 +367,42 @@ const Events = {
...
   },
+
+  /**
+   * Determines if the key pressed is an accent key
+   *
+   * @method isAccentKey
+   * @param {Event} ev A JavaScript event
+   * @returns {Boolean} true if end key
+   * */
+  isAccentKey: (ev) => {
```
jamime, line 378: These functions should accept an Event, this implementation is expecting a string. However, we should be able to safely delete this method.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -1555,27 +1555,6 @@ describe("Decimal", () => {
           });
         }
       );
-
```
jamime, line 1558: We should be able to re-introduce this test now that you've re-added the invariant.

jamime (changes requested):

src/__experimental__/components/decimal/decimal.spec.js
DipperTheDan: Cheers for spotting this. I thought I added all the `invariant` tests back in to the specs file. Must have missed this. My bad.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -247,7 +247,7 @@ class Decimal extends React.Component {
   removeDelimiters = (value) => {
     const format = I18nHelper.format();
     const delimiter = `\\${format.delimiter}`;
-    const delimiterMatcher = new RegExp(`[${delimiter}]*`, "g");
+    const delimiterMatcher = new RegExp(`[${delimiter}^´¨˙~]*`, "g");
```
chrisbarber86, line 250: The intention of `removeDelimiters` was purely to remove the delimiter as provided by I18n.

(in UKI that would be a `,`, so it would convert `10,000.00` down to `10000.00`)

not sure we should use this to also match other characters like `^´¨˙~`

DipperTheDan:
Closed PR due to the requirements for Decimal changing. We have decided that we are going to make user experience improvements to the component and this PR only adds to the complexity of the component.

PR #3642, opened 2021-02-05
chore: remove css style utils (FE-3085)

grabkowski:
### Proposed behaviour
Remove css and style utils

### Checklist


.storybook/main.js
```diff
@@ -1,31 +1,33 @@
...
   webpackFinal: async (config, { configType }) => {
-    config.module.rules.push(
-      {
-        test: /\.scss$/, use: ['style-loader', 'css-loader', 'sass-loader']
-      }
-    );
-  
+    config.module.rules.push({
+      test: /\.scss$/,
+      use: ["style-loader", "css-loader", "sass-loader"],
```
sjohnson-sage, line 23: I don't think we need the sass-loader anymore do we?

src/style/global-style.js
```diff
@@ -0,0 +1,26 @@
...
+    padding: 0px;
+    text-rendering: optimizeLegibility;
+    -webkit-font-smoothing: antialiased;
+  }
+
+  button, input, select, textarea {
+    font-family: inherit;
+  }
+
+  h1, .h1 { font-size: 24px; font-weight: 900; line-height: 32px; }
```
sjohnson-sage, line 18: Didn't we decide not to have the heading styles in here anymore?

docs/getting-started.stories.mdx
```diff
@@ -37,6 +37,23 @@ There is a peer dependency on `draft-js` any project intending to use either the
...
+then the `GlobalStyle` component could be placed next to the content of the app:
+```
+  <App>
+    <GlobalStyle />
+    .
+    .
+    .
+  </App>
+```
+
```
sjohnson-sage, line 56: I think we want to provide the global CSS in here as well, in case the user just wants to copy and paste it into their own global CSS rather than use our `GlobalStyle` component

.storybook/main.js
sjohnson-sage: Maybe we don't need any of these loaders even

sjohnson-sage (commented):
Looks good, just a few comments from me

src/style/global-style.js
grabkowski: we can't change it for now as our storybook stories will change, which will trigger a few chromatic failures. I'll create a task for it to be removed in the future.

.storybook/main.js
grabkowski: I've opted to leave it as it is, as it's only storybook. You're right, I could leave only one loader.

docs/getting-started.stories.mdx
grabkowski: good point

.storybook/main.js
grabkowski: actually we don't need any loader here, so I've removed them all

docs/getting-started.stories.mdx
sjohnson-sage: We have decided against this, we want to encourage the use of `GlobalStyle`

sjohnson-sage (approved):

ianforster-sage (approved):

DlgSHi (approved):

PR #3656, opened 2021-02-09
feat(heading): add pills container - FE-3376

bulaj:
### Proposed behaviour
Heading has exposed ``pills`` prop, allowing the developer to add one or more pills inline with heading.
Pills will wrap along with title text.
image

### Current behaviour
There is no effective way of using pills inline with heading.

### Checklist

- [ ] <del>All themes are supported if required
- [ ] <del>Cypress automation tests added or updated if required

### Testing instructions
Added appropriate stories to Heading section in storybook.


src/components/heading/heading.stories.mdx
```diff
@@ -1,76 +1,141 @@
...
+
 Import `Heading` into the project.
 
 ```jsx
 import Heading from "carbon-react/lib/components/heading";
 
-const MyComponent = () => (
-  <Heading title='This is a Title' />
-);
+const MyComponent = () => <Heading title="This is a Title" />;
```
edleeks87, line 29: This can be removed, we are slowly updating the component stories to just have the import path in the `Quick Start` section

src/components/heading/heading.component.js
```diff
@@ -70,6 +71,11 @@ class Heading extends React.Component {
      * Adds a separator between the title and the subheader.
      */
     separator: PropTypes.bool,
+
+    /**
+     * Pills that will be added after the title.
+     */
+    pills: PropTypes.node,
```
grabkowski, line 78: why are we allowing any node here? Shouldn't we enforce that prop to accept only Pill components?

src/components/heading/heading.component.js
```diff
@@ -29,7 +30,7 @@ class Heading extends React.Component {
     /**
      * Defines the title for the heading.
      */
-    title: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+    title: PropTypes.node,
```
grabkowski, line 33: shouldn't it be just a string?

src/components/heading/heading.spec.js
```diff
@@ -92,6 +94,39 @@ describe("Heading", () => {
...
   });
 
+  it("renders passed components inside pills container when provided", () => {
+    const testPills = [
+      <Pill key="1">test pill 1</Pill>,
+      <Pill key="2">test pill 2</Pill>,
+    ];
+    const wrapper = shallow(<Heading title="test" pills={testPills} />);
+    expect(wrapper.find(StyledHeadingPills).children().length).toEqual(2);
+    expect(wrapper.find(Pill).children().length).toEqual(2);
```
grabkowski, line 104: is there a reason that you're checking pill children length? couldn't it just be `expect(wrapper.find(Pill)).toHaveLength(2);`?

src/components/heading/heading.spec.js
```diff
@@ -120,6 +155,13 @@ describe("Heading", () => {
     });
   });
 
+  describe("no pills", () => {
+    it("returns no pills wrapper", () => {
+      const wrapper = mount(<Heading title="test" />);
+      expect(wrapper.find(StyledHeadingPills).length).toEqual(0);
```
grabkowski, line 161: you could check if the component exists:
`expect(wrapper.find(StyledHeadingPills).exists()).toBe(false)`

src/components/heading/heading.component.js
bulaj: After @edleeks87 noticed, mistake with backLink i reverted all types as they were before.

src/components/heading/heading.spec.js
bulaj: I want to make sure that additional container is rendered as has exact children as passed.

src/components/heading/heading.component.js
bulaj: There is no reason to restrict it just to Pill and actually many benefits, its common pattern used in ui libraries. eg:
- You can supply an array of elements,
- You can wrap multiple pills with fragment,
- You can use wrapper components like styled components made out of Pill ```styled(Pill)``` or other that basically return enhanced Pill,
- Leaves more flexibility for various business requirements in userland scenarios.

src/components/heading/heading.component.js
```diff
@@ -168,6 +174,20 @@ class Heading extends React.Component {
...
+   * @method pills
+   * @return {Object} JSX
+   */
+  get pills() {
+    return this.props.pills ? (
+      <StyledHeadingPills data-element="pills">
+        {this.props.pills}
+      </StyledHeadingPills>
+    ) : null;
+  }
```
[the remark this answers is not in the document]
bulaj: I think the same but figured out that actual difference would be the word get and ().
Decided that it will be slightly quicker for someone to understand whats going on before the refactor if everything is the same.
OFC i can just quickly refactor the whole file, but thought the diff would be unnecessarily big ;)

src/components/heading/heading.component.js
```diff
@@ -59,7 +60,7 @@ class Heading extends React.Component {
     /**
      * Defines the a href for the back link.
      */
-    backLink: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+    backLink: PropTypes.string,
```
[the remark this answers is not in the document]
bulaj: I will rollback changes in propTypes and remove heading.d.ts altogether as those need to be consistent.

src/components/heading/heading.spec.js
bulaj: Will change

src/components/heading/heading.spec.js
grabkowski: I was asking why are you checking the children of Pill instead of Pill directly.
As you mentioned the container - it could also be checked using `toHaveLength`:
```
expect(wrapper.find(StyledHeadingPills).children()).toHaveLength(2);
expect(wrapper.find(Pill)).toHaveLength(2);
```

grabkowski (approved):

edleeks87 (approved):

KatarzynaQA (approved):

Issue #3519, opened 2020-12-18
Card overflow isn't handled


vincentfodor:
### Current behaviour
![image](https://user-images.githubusercontent.com/30438094/102618462-25eb8500-413b-11eb-9914-a80822ef6903.png)
Currently the margin of the child components (CardRow) is not respected by the card as visible overflow.

### Expected behaviour
![image](https://user-images.githubusercontent.com/30438094/102618196-a958a680-413a-11eb-85b2-a874c7b73eac.png)

### Reproducible example
```
<Card spacing="medium">
    <CardRow>
        <Heading
            title="Title"
            subheader="Subtitle"
        />
    </CardRow>
    <CardRow>
        Lorem ipsum dolor sit, amet consectetur adipisicing
        elit. Deleniti eligendi illo ipsam, dolore autem
        eaque totam obcaecati modi possimus nihil qui facere
        neque, necessitatibus itaque illum similique beatae
        voluptates sunt?
    </CardRow>
</Card>
```

### Suggested solution(s)
Adding `overflow: auto` to the Card components styling.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 54.3.0
| React            | 17.0.1
| Browser          | Chrome

ianforster-sage:
FE-3463

PR #3682, opened 2021-02-12
feat(card): add styled-system padding props (FE-3463)

ianforster-sage:
fixes: #3519

### Proposed behaviour
Replaces margin with `styled-system/padding` in `CardRow`
Screenshot 2021-02-12 at 15 29 23

### Current behaviour
Described in the linked issue https://github.com/Sage/carbon/issues/3519

### Checklist

~- [ ] Cypress automation tests added or updated if required~
~- [ ] Storybook added or updated if required~
~- [ ] Typescript `d.ts` file added or updated if required~

### Additional context
N/a

### Testing instructions
Use the generated codesandbox
https://codesandbox.io/s/carbon-quickstart-forked-yqvpx?file=/src/index.js

### DIFF
--- a/src/components/card/card-row/card-row.component.js
+++ b/src/components/card/card-row/card-row.component.js
@@ -1,19 +1,37 @@
 import React from "react";
+import styledSystemPropTypes from "@styled-system/prop-types";
 import PropTypes from "prop-types";
 import StyledCardRow from "./card-row.style";
 import OptionsHelper from "../../../utils/helpers/options-helper/options-helper";
+import { filterStyledSystemPaddingProps } from "../../../style/utils";
 
-const CardRow = ({ children, spacing, ...props }) => {
+const paddingPropTypes = filterStyledSystemPaddingProps(
+  styledSystemPropTypes.space
+);
+
+const paddingSizes = {
+  small: 2,
+  medium: 3,
+  large: 4,
+};
+
+const CardRow = ({ children, spacing = "medium", ...props }) => {
   return (
-    <StyledCardRow data-element="card-row" spacing={spacing} {...props}>
+    <StyledCardRow
+      data-element="card-row"
+      py={paddingSizes[spacing]}
+      {...props}
+    >
       {children}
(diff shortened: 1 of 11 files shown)


DipperTheDan (dismissed):

bulaj (dismissed):

edleeks87 (changes requested):
I think we should replace the margin in `CardRow` with padding instead. We would need to add something to adjust the padding (top) when there is more than one `CardRow` to avoid the spacing in the pic below 

image

edleeks87 (approved):

bulaj (approved):

edleeks87 (dismissed):

bulaj (approved):

src/components/card/card-row/card-row.component.js
```diff
@@ -1,19 +1,37 @@
...
     </StyledCardRow>
   );
 };
 
 CardRow.propTypes = {
+  /** Styled system padding props */
+  ...paddingPropTypes,
   children: PropTypes.node.isRequired,
-  /** size of card for applying margin (small | medium | large) */
+  /** Predefined size of card for applying padding (small | medium | large). For more granular control these can be over-ridden by Padding props from styled-system (see table below). */
```
Alec-Walker-Sage, line 34: Could we make it more clear that padding is set by the spacing prop on Card and setting a spacing prop in CardRow will do nothing. 
```suggestion
  /** Spacing prop is set in Card and defines the padding for the CardRow (the first CardRow has no padding by default) For more granular control of CardRow padding these can be over-ridden by Padding props from styled-system (see table below). */
```

Alec-Walker-Sage (commented):
Could we also change the description for spacing in the CardFooter prop to be something similar to how it is in CardRow at the moment because I believe it works the same

Alec-Walker-Sage (approved):

bulaj (approved):

edleeks87 (approved):

Issue #3623, opened 2021-01-29
Focus leaves dialog after shift-tab on a radio button group 


Alec-Walker-Sage:
### Current behaviour

If you are in a dialog and you shift-tab back into a radio group and then shift-tab again the focus leaves the dialog and out of the focus trap. **This only happens if it is the last item in the radio group selected.**

![2021-02-04 11 05 24](https://user-images.githubusercontent.com/2328042/106884300-069fbb80-66d9-11eb-8c83-6f07e0392c39.gif)

### Expected behaviour

The focus should remain in the dialog (like it does when the first item in the radio group is selected).

![2021-02-04 11 04 08](https://user-images.githubusercontent.com/2328042/106884284-02739e00-66d9-11eb-8d1b-d9f690b7c5ec.gif)

### Reproducible example

https://codesandbox.io/s/compassionate-dew-17i49?file=/src/index.js
1) open the dialog 
2) select the second radio button 
3) tab onto the form inputs
4) use shift tab to go back onto the radio buttons
5) press shift tab again which take you out of the dialog

### Additional context

This was found out during FE-3586 looking into the Focus Trap.
### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 62.0.0
| Browser          | All
| Operating System | Mac OS Catalina

jamime:
FE-3690

PR #3715, opened 2021-02-16
fix(focus-trap): handle radio button focus

sjohnson-sage:
Fixes: #3623

### Proposed behaviour
Stop focus from escaping focus trap when radio buttons are the first focusable elements.

When radio buttons are present but not the first focusable elements, the tab/keyboard behaviour will be the same as default HTML radio buttons.

### Current behaviour
When radio buttons are first focusable elements in a trap, selecting the second radio, tabbing then shift-tabbing twice makes focus escape from trap.

### Checklist

### Testing instructions
Sandbox from issue, if anything else is needed let me know


edleeks87 (approved):

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -0,0 +1,42 @@
...
+    element();
+  } else {
+    const el = element.current || element;
+    el.focus();
+  }
+}
+
+const isRadio = (element) => {
+  return (
+    element.hasAttribute("type") && element.getAttribute("type") === "radio"
```
dawidzarzycki, line 15: Shouldn't we just have ` element.getAttribute("type") === "radio"`
Or we should change the function name if we want to have range out of radio

sjohnson-sage: So this is saying "if the element has a `type` attribute AND the type is `radio`" then return true - we need the first check otherwise we don't know if there's a `type` attribute to check. 

So the function name makes sense because we're only checking if the element is a radio button

dawidzarzycki (approved):

KatarzynaQA (approved):

edleeks87 (approved):

PR #3723, opened 2021-02-18
feat(radio-button): allow use of all styled system margin props (FE-3518)

grabkowski:
### Proposed behaviour
Add styled-system margin props to RadioButton and RadioButtonGroup

Filter out props so proper ones are passed to subcomponents in which they should be instead of spreading them.
Add an util to filter out unnecessary object properties.

### Checklist

### Testing instructions
1. npm start
2. open: http://localhost:9001/?path=/docs/experimental-radiobutton--with-legend-and-labels

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button-group.component.js
+++ b/src/__experimental__/components/radio-button/radio-button-group.component.js
@@ -1,14 +1,19 @@
 import React from "react";
 import PropTypes from "prop-types";
-
+import styledSystemPropTypes from "@styled-system/prop-types";
 import tagComponent from "../../../utils/helpers/tags";
 import Fieldset from "../../../__internal__/fieldset";
 import RadioButtonGroupStyle from "./radio-button-group.style";
 import RadioButtonMapper from "./radio-button-mapper.component";
 import Logger from "../../../utils/logger/logger";
 import useIsAboveBreakpoint from "../../../hooks/__internal__/useIsAboveBreakpoint";
+import filterStyledSystemMarginProps from "../../../style/utils/filter-styled-system-margin-props";
 
 let deprecatedWarnTriggered = false;
+
+const marginPropTypes = filterStyledSystemMarginProps(
+  styledSystemPropTypes.space
+);
 
 const RadioButtonGroup = (props) => {
   if (!deprecatedWarnTriggered) {
@@ -33,8 +38,6 @@ const RadioButtonGroup = (props) => {
     legendWidth,
     legendAlign,
     legendSpacing,
-    ml,
-    mb,
     labelSpacing = 1,
     adaptiveLegendBreakpoint,
     adaptiveSpacingBreakpoint,
@@ -42,9 +45,12 @@ const RadioButtonGroup = (props) => {
     styleOverride = {},
   } = props;
 
+  const marginProps = filterStyledSystemMarginProps(props);
+
   const isAboveLegendBreakpoint = useIsAboveBreakpoint(
     adaptiveLegendBreakpoint
   );
+
   const isAboveSpacingBreakpoint = useIsAboveBreakpoint(
     adaptiveSpacingBreakpoint
   );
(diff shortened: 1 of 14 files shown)


src/__experimental__/components/radio-button/radio-button-group.style.js
```diff
@@ -2,6 +2,7 @@ import styled from "styled-components";
 
 const RadioButtonGroupStyle = styled.div`
   ${({ inline }) => inline && "display: flex;"}
+  ${({ legendInline }) => legendInline && "margin-top: 4px;"}
 
   ${({ styleOverride }) => styleOverride};
```
dawidzarzycki, line 7: I suggest to list all props at the beginning 
```js
const RadioButtonGroupStyle = styled.div`
${({inline, legendInline, styleOverride}) => css``
```

grabkowski: good point, changed

src/__experimental__/components/radio-button/radio-button-group.component.js
```diff
@@ -1,15 +1,39 @@
 import React from "react";
 import PropTypes from "prop-types";
 
+import styledSystemPropTypes from "@styled-system/prop-types";
```
sjohnson-sage, line 4: Minor but this should be grouped with the external imports with the line break after it really

dawidzarzycki (approved):

src/__experimental__/components/radio-button/radio-button-group.component.js
```diff
@@ -1,15 +1,39 @@
...
+  "mt",
+  "marginBottom",
+  "mb",
+  "marginX",
+  "mx",
+  "marginY",
+  "my",
+];
+
+const marginPropTypes = filterObjectProperties(
```
sjohnson-sage, line 32: Can we move this and `marginProperties` into a reusable util somewhere? I think we'll need it in other components as well

dawidzarzycki (dismissed):

src/__experimental__/components/radio-button/radio-button.component.js
```diff
@@ -1,54 +1,119 @@
...
+  "mt",
+  "marginBottom",
+  "mb",
+  "marginX",
+  "mx",
+  "marginY",
+  "my",
+];
+
+const marginPropTypes = filterObjectProperties(
```
sjohnson-sage, line 27: Same as above, can we move these into a reusable util please?

sjohnson-sage (commented):
Looks good! Main thing from me is to move the `marginPropTypes` and `marginProperties` into a constants file or something so we don't have to redefine them in multiple places

src/__experimental__/components/radio-button/radio-button-group.component.js
grabkowski: this is already a reusable util, you'll just have to specify what you want to be filtered out. I don't think that we need an separate util for margin and padding as it could be easily done by just specifying a second argument.

sjohnson-sage: I disagree because we're going to use this `marginPropTypes` in lots of components! and this is just going to be duplicated like 50+ times. 

At the very least I think 
```
const marginProperties = [
  "margin",
  "m",
  "marginLeft",
  "ml",
  "marginRight",
  "mr",
  "marginTop",
  "mt",
  "marginBottom",
  "mb",
  "marginX",
  "mx",
  "marginY",
  "my",
];
```
should be moved into a constants file

grabkowski: ok, you've pointed the wrong line, I thought you're talking about `marginPropTypes`

grabkowski: I have moved `marginProperties` to a separate file

sjohnson-sage: I was talking about both 😉

sjohnson-sage (dismissed):

dawidzarzycki (approved):

DlgSHi (approved):

ianforster-sage (approved):

dawidzarzycki (approved):

DlgSHi (approved):

PR #3732, opened 2021-02-22
refactor(decimal): allow the user to input any value - FE-3705

### LINKED ISSUE #3372: Decimal fields crashes app if "^" is used on Mac.

DipperTheDan:
BREAKING CHANGE: 
Precision prop can no longer be changed on the fly. Your app will need to refresh in order for the user to see the new precision value.

### Proposed behaviour

- Allow user to input anything they want in to Decimal component
- Remove styled wiggle

### Current behaviour

- Special and Alphabetic characters are currently prevented from being input
- Component currently has styled wiggle to inform the user they have tried to input an illegal character

### Checklist

<del> - [ ] Screenshots are included in the PR if useful </del>

### Additional context
fixes #3372 

### Testing instructions
- Manual testing - https://codesandbox.io/s/carbon-quickstart-forked-01vep

- Locale testing - https://codesandbox.io/s/carbon-quickstart-forked-1v8qj

The below table is to give an idea of what manual tests should be used to check the Decimal component. 

| Precision | User Input | Value  |Status|
|-----------|------------|--------|-----|
|(Default)     | @£$%^         | @£$%^      |
|(Default)    | ^          | ^      |
|(Default)      | ˇ         | ˇ    |
|(Default)    | ---       | ---    |
| 0         | 1          | 1      |
| 0         | 2          | 2      |
| 0         | 1234567789 | 1,234,567,789 |
| 1         | 1          | 1.0    |
| 1         | 1.2        | 1.2    |
| 1         | 1.23       | 1.23   |
| 2         | 2          | 2.00   |
| 2         | 2.1        | 2.10   |
| 2         | 2.12       | 2.12   |
| 2         | 2.123      | 2.123  |
| 3         | 2.1        | 2.100  |
| 3         | 2.12       | 2.120  |
| 3         | 2.123      | 2.123  |
| 3         | 2.1234     | 2.1234 |
| 4         | 1.1234         | 1 .1234     |
| 4         | 2          | 2.0000      |
| 4         | 234556654          | 234,556,654.0000      |
| 5         | 1          | 1.00000    |
| 5         | 1.12345        | 1.12345    |
| 5         | 1a.23       | 1a.23   |
| 6         | 2.123456          | 2.123456   |
| 6         | 2.1        | 2.1000000   |
| 6         | 2a.12       | 2a.12   |
| 6         | 1,232.123      | 1,232.123000  |
| 7         | 1          | 1.0000000      |
| 7         | 2344.1234567  | 2,344.1234567 |
| 7         | 88652344.1234567  | 88,652,344.1234567 |
| 8         | 1          | 1.00000000    |
| 8         | 1.2        | 1.20000000    |
| 8         | 1.23       | 1.23000000   |
| 9         | 2          | 2.000000000   |
| 9         | 2.1        | 2.100000000   |
| 9         | 1222.12       | 1,222.120000000   |
| 9         | 2.123000000      | 2.123000000  |
| 10         | 2.1        | 2.1000000000  |
| 10         | 2.12       | 2.1200000000  |
| 10         | 1222.123      | 1,222.1230000000  |
| 10         | 2.12345     | 2.1234500000 |
| 11         | 1         | 1.00000000000    |
| 11         | 2345          | 2,345.00000000000      |
| 11         | 17899536472345          | 17,899,536,472,345.00000000000      |
| 12         | 1          | 1.000000000000    |
| 12         | 1.12345        | 1.123450000000    |
| 12         | 1a.23       | 1a.23   |
| 13         | 2          | 2.0000000000000   |
| 13         | 2.1        | 2.1000000000000   |
| 13         | 2a.12       | 2a.12   |
| 13         | 1,232.123      | 1,232.1230000000000  |
| 14         | 2.1        | 2.10000000000000  |
| 14         | 2.12       | 2.12000000000000  |
| 14         | 2.123      | 2.12300000000000  |
| 14         | 1222.1234     | 1,222.12340000000000 |
| 15         | 1         | 1.000000000000000  |
| 15         | 2332.78          | 2,332.780000000000000      |
| 15         | 1a3.55          | 1a3.55    |
| 15         | 1.12345        | 1.123450000000000    |
| 15         | 1a.23       | 1a.23   |
| 16         | 2.123456          | 2.1234560000000000 && Prop validation error to appear in console|


src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -230,17 +154,6 @@ class Decimal extends React.Component {
   getSafePrecisionProp = () => {
     const { precision } = this.props;
```
edleeks87, line 156: Do we want to remove these checks? It might be worth moving it into a custom prop type function

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -1649,6 +727,24 @@ describe("Decimal", () => {
...
+        onChange: (e) => {
+          setProps({ value: e.target.value.rawValue });
+        },
+        allowEmptyValue: true,
+      });
+
+      setProps({ value: "FooBar" });
+      blur();
+      expect(value()).toBe("FooBar");
+      expect(hiddenValue()).toBe("FooBar");
```
chrisbarber86, line 744: Should we update hiddenValue to be FooBar? Normally the expectation of the hiddenValue is that it is strictly a plainly formatted (i.e. only the seperator & numbers present) value. If the typed value is `FooBar` i'd expect the hiddenValue to be `null`

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -91,53 +80,6 @@ class Decimal extends React.Component {
...
-    if (Events.isMinusKey(ev) && ev.target.selectionStart === 0) {
-      // user is entering a negative symbol at the start of the number
-      return true;
-    }
-
-    this.startAnimation();
-    return false;
-  };
-
   onBlur = () => {
```
chrisbarber86, line 83: Won't the logic in here need to update as well?

chrisbarber86 (changes requested):

src/__experimental__/components/decimal/decimal.spec.js
DipperTheDan: My thoughts on this were that the consumer would be using the hiddenValue for validation purposes. I'm happy to change it to `null` to meet the expectation of it being a plainly formatted value though.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -2,10 +2,8 @@
 import React from "react";
 import PropTypes from "prop-types";
 import invariant from "invariant";
-import StyledWiggle from "./decimal.style";
 import Textbox from "../textbox";
 import I18nHelper from "../../../utils/helpers/i18n";
-import Events from "../../../utils/helpers/events";
 
 class Decimal extends React.Component {
```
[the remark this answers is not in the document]
DipperTheDan: I agree. I'll make sure we have a ticket on our backlog for it 😄

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -336,7 +223,14 @@ Decimal.propTypes = {
   /**
    * The decimal precision of the value in the input
    */
-  precision: PropTypes.number,
+  precision: (props, propName) => {
+    if (props.precision < 0 || props.precision > Decimal.maxPrecision) {
```
edleeks87, line 227: I think you need to check if it's a integer number as well here

```suggestion
    if (!Number.isInteger(precision) || props.precision < 0 || props.precision > Decimal.maxPrecision) {
```

edleeks87 (dismissed):

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -421,191 +320,13 @@ describe("Decimal", () => {
...
-      "11111a",
-      "a111111",
-      "1111a1111",
-    ])(
-      "prevents the user from pasting anything but a number, separator, delimiter or negative sign (%s)",
-      (key) => {
-        render();
-
-        const { preventDefault } = paste({ key }, "0|.00");
-        expect(preventDefault).toHaveBeenCalled();
```
jamime, line 598: Instead of deleting these tests I think we can assert on the value instead, to make sure the value changed correctly. Same for the tests below.

jamime (changes requested):
There is an inconsistency between how the user input `2` and `3.00` is handled, see the video below. I think these should be consistent.


https://user-images.githubusercontent.com/2328042/111334141-37292c80-866b-11eb-9d93-b639751fa2a8.mov

I added this to the render method to debug

```diff
  render() {
    const { name, defaultValue, ...rest } = this.props;
    return (
      <>
+      <pre>{JSON.stringify(this.state, null, 2)}</pre>
        <Textbox
```

DipperTheDan:
> There is an inconsistency between how the user input `2` and `3.00` is handled, see the video below. I think these should be consistent.
> 

This is how Decimal behaves in `master` so I left it that way as I was unsure if changing this would have an effect on existing implementations of Decimal. I've attached a video of this in `master`.

Uploading Decimal Value Recording.mov…

jamime:
> > There is an inconsistency between how the user input `2` and `3.00` is handled, see the video below. I think these should be consistent.
> 
> @jamime This is how Decimal behaves in `master` so I left it that way as I was unsure if changing this would have an effect on existing implementations of Decimal. I've attached a video of this in `master`.
> 
>  Decimal.Value.Recording.mov

Ah ok, we can address that at a later date.

jamime (changes requested):
There are some requirements that have not yet been fulfilled.

> Field should still auto format on blur adding the the internationalised separators, delimiters and trailing .00, 00, 0. This is configurable using the precision prop.

The PR still applies **rounding** to the precision limit. We don't want to round at all. We do want to append `0`'s where the value was less than the precision limit. These are the same value, but with a different representation `1.2 === 1.20 === 1.200`.

> Field should allow any character to be inputted, the consumer should then provide their own decimal validation.

When there is more decimal places than the precision limit we should not modify the value.

Here are some examples that illustrate this

| Precision | User Input | Value  |
|-----------|------------|--------|
| 0         | 1          | 1      |
| 0         | 2          | 2      |
| 1         | 1          | 1.0    |
| 1         | 1.2        | 1.2    |
| 1         | 1.23       | 1.23   |
| 2         | 2          | 2.00   |
| 2         | 2.1        | 2.10   |
| 2         | 2.12       | 2.12   |
| 2         | 2.123      | 2.123  |
| 3         | 2.1        | 2.100  |
| 3         | 2.12       | 2.120  |
| 3         | 2.123      | 2.123  |
| 3         | 2.1234     | 2.1234 |

Please let me know if you have any questions.

src/__experimental__/components/decimal/decimal.spec.js
jamime: I've just reviewed this @chrisbarber86 but in this case I think we should actually send `FooBar`. If we were to send a `POST` with nothing as the value the backend could only say `this field is required`, but not `FooBar isn't valid`.

If the field was optional, and we sent nothing the backend wouldnt be able to flag that the input value was discarded, which is the same issue we're trying to avoid in SPA validations.

src/__experimental__/components/decimal/decimal.component.js
jamime: I think we can remove all of the `onBlur` code because it is only handling the `-` key press. If a user pressed `-` accidentally instead of `9` we could pick up the user error in validations because we're reverting to the `defaultValue` currently.

src/__experimental__/components/decimal/decimal.spec.js
chrisbarber86: Ok - so the consumer on the front end would need to add validations if they don't want to send `FooBar`. otherwise if they don't it's on the server to handle that.

I'm ok with that.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -256,40 +159,23 @@ class Decimal extends React.Component {
...
-      clipboardData,
-      type,
-    } = ev;
-    let { key: change } = ev;
-
-    if (type === "paste") {
-      change = clipboardData.getData("text/plain");
+    const { separator } = I18nHelper.format();
+    const parts = value.split(separator);
+    if (this.props.precision === 0) {
```
jamime, line 170: If you enter `1234.56` it will be re-formatted as `1,234` and will loose precision. We don't want this to happen. Please add a test that covers this.

jamime (changes requested):

cypress/features/regression/experimental/decimalPrecision.feature
```diff
@@ -9,7 +9,6 @@ Feature: Decimal component - precision
     Examples:
       | labelInput         | nameOfObject | fieldHelpOutput   |
       | 150                | precision0   | 150               |
-      | 1234567890.5       | precision1   | 1234567890.5      |
```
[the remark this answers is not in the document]
DipperTheDan: Removed by mistake. Added back in 👍🏻

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -228,106 +192,36 @@ describe("Decimal", () => {
...
-      it.each(["12.85.1", "12.,85", "12a.85", "12,85"])(
-        "throws if the defaultValue is not a number (%s)",
-        (defaultValue) => {
-          expect(() => {
-            render({ defaultValue });
-          }).toThrow(`The supplied decimal (${defaultValue}) is not a number`);
-        }
-      );
-
-      it("throws if the precision is greater than 15", () => {
```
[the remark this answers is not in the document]
DipperTheDan: Test still exists but has been rewritten as this is no longer an invariant but a console error. 
https://github.com/Sage/carbon/pull/3732/files/e815557daffb13aaea8b4d92c03e5b341b0f72b3#diff-992c7dd22ced4f0d4c1a51cd13658c283bbd2e5d4e56e776892de6672be2e369R215

src/__experimental__/components/decimal/decimal.component.js
DipperTheDan: Test has been added to cover this scenario.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -228,106 +192,36 @@ describe("Decimal", () => {
...
-    });
-
     describe("precision", () => {
-      it("has a default precision of 2 (rounding up)", () => {
-        render({ defaultValue: "12345.655" });
-        expect(value()).toBe("12,345.66");
-        expect(hiddenValue()).toBe("12345.66");
-      });
-
-      it("has a default precision of 2 (rounding down)", () => {
```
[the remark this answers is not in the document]
DipperTheDan: Test has been added. Good shout thinking about adding this 👍🏻

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -228,106 +192,36 @@ describe("Decimal", () => {
...
+        jest.spyOn(global.console, "error").mockImplementation(() => {});
+        mount(<Decimal defaultValue="12345.654" precision={16} />);
+        // eslint-disable-next-line no-console
+        expect(console.error).toHaveBeenCalledWith(
+          "Warning: Failed prop type: Precision prop must be a number greater than 0 or equal to or less than 15.\n    in Decimal"
+        );
+        global.console.error.mockReset();
+      });
+
       it("allows the user to change the precision", () => {
```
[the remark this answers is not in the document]
DipperTheDan: Messaged you privately about this as it required a screen recording and a lengthy message that could trigger a discussion.

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -34,45 +31,23 @@ class Decimal extends React.Component {
       "Decide between using a controlled or uncontrolled input element for the lifetime of the component";
     const isControlled = this.isControlled();
     invariant(this.state.isControlled === isControlled, message);
-
+    if (prevProps.precision !== this.props.precision) {
+      // eslint-disable-next-line no-console
+      console.error(
+        "Decimal `precision` prop has changed value. Doing this will affect the value displayed to the user."
```
jamime, line 37: This isn't true any more. 

```suggestion
        Changing the "Decimal `precision` prop has no effect.
```

cypress/features/regression/experimental/decimalPrecision.feature
```diff
@@ -26,6 +26,7 @@ Feature: Decimal component - precision
       | 9.517567891111398  | precision15  | 9.517567891111398 |
       | 17.517567891111390 | precision15  | 17.51756789111139 |
       | 150.091            | precision0   | 150               |
+      | 1433.01            | precision1   | 1433.0            |
```
jamime, line 29: Copy manual tests into here

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -256,40 +149,43 @@ class Decimal extends React.Component {
...
-   * This allows us to prevent the user input if it is invalid
-   */
-  replace = (ev) => {
-    const {
-      target: { selectionStart, selectionEnd, value },
-      clipboardData,
-      type,
-    } = ev;
-    let { key: change } = ev;
+    const { separator } = I18nHelper.format();
```
jamime, line 160: I've sent some code on slack to update this with.

jamime (changes requested):
After these changes I'll re-review and approve and we'll then squash the commits and add a `BREAKING CHANGE` for the precision prop (and anything else).

src/__experimental__/components/decimal/decimal.component.js
DipperTheDan: 👍🏻 Applied.

src/__experimental__/components/decimal/decimal.component.js
DipperTheDan: Changed this error message.

jamime (approved):

cypress/features/regression/experimental/decimalPrecision.feature
```diff
@@ -7,37 +7,61 @@ Feature: Decimal component - precision
...
-      | 1.00000000011445   | precision10  | 1.0000000001      |
-      | 1.000000000129999  | precision11  | 1.00000000012     |
-      | 1234.5175678911199 | precision12  | 1234.517567891119 |
-      | 7.5175678911       | precision13  | 7.5175678911      |
-      | 1.517567891111399  | precision14  | 1.51756789111139  |
-      | 9.5175678911113981 | precision15  | 9.517567891111398 |
\ No newline at end of file
+      | nameOfObject         | labelInput       | fieldHelpOutput                 |
+      | precision0           | 1                | 1                               |
+      | precision0           | 2                | 2                               |
```
DlgSHi, line 12: This TC is the same as a previous one, we can change it into some character instead of checking the same number row.

DipperTheDan: That is a good idea. I'll use this replacement as a `dead key` test. Cheers for spotting that 👍🏻

edleeks87 (approved):

jamime (approved):

chrisbarber86 (approved):

edleeks87 (approved):

harpalsingh:
As per the requirements, I have checked all expected outcomes and accessibility requirements to enter any character are working as expected/

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -138,10 +138,12 @@ class Decimal extends React.Component {
...
 
   removeDelimiters = (value) => {
-    /* If a value consists of only commas or only letters and commas, we want to treat that 
+    /* If a value is passed in that is a number but has too many delimiters, we want to handle this
+       value without formatting it or removing delimiters. We also want to consider that,
+       if a value consists of only commas or only letters and commas, we want to treat that 
        value in the same way as if the value was NaN. We want to pass this value to the 
        formatValue function, before the commas can be removed. */
-    if (value.match(/^[a-zA-Z,\W_]*$/)) {
+    if (value.match(/([^A-Za-z0-9]{2,})/)) {
```
jamime, line 146: The delimiter isn't always `,`, it depends on the users locale.

jamime (changes requested):

src/__experimental__/components/decimal/decimal.component.js
edleeks87: This will match any special (non-alphanumeric) characters so I think just the comment needs updating?

jamime (commented):
@DipperTheDan is currently working on defects raised by QA. When using the FR locale the remainder is rounded if it is greater than the precision.

jamime (commented):
We reviewed the changes to date and made some minor ammendments to the I18NHelpers. You can now use `I18nHelper.formatDecimal(12345.678, 2, {round: false})` which will format the decimal as `12,345.687`. The standard implementation would return `12,345.69`. This has allowed us to remove the duplicated code in `Decimal.component.js`.

There is one outstanding issue that @DipperTheDan is going to look into.

- When the locale is `fr`
- and the user enters `123,,56` (123.56)
- value is changed to `123..56` (12,356)

These are both invalid decimals, but it is preferred if the value does not automatically change.

cypress/support/step-definitions/decimal-steps.js
```diff
@@ -1,17 +1,19 @@
...
+  commonDataElementInputPreviewNoIframe().clear().type(input).blur();
 });
 
-Then("Decimal Input is set to {word}", (labelInput) => {
+Then("Decimal Input is set to {word}", (input) => {
   commonDataElementInputPreviewNoIframe()
-    .should("have.attr", "value")
-    .should("contain", `${labelInput}`);
+    .should("have.attr", "value", input)
+    .wait(50);
```
jamime, line 10: Why do we need to wait?

jamime (commented):
👍  LGTM. I'll approve once commits have been squashed (after QA).

edleeks87:
👍 LGTM. I'll approve once commits have been squashed (after QA).

DlgSHi (approved):

jamime (approved):

edleeks87 (approved):

PR #3737, opened 2021-02-22
fix(dl): support for conditional rendering - FE- 3600

dawidzarzycki:
### Proposed behaviour
Add support for conditional rendering with using of `<React.Fragment />` component 

definition-list-conditional-rendering

### Current behaviour
`<React.Fragment />` and content of the component don't render

### Checklist

- [ ] <dl> All themes are supported if required

### Additional context
I also created MDX file as it didn't exist. The MDX is located in storybook: `Design System/Defitnion List`

### Testing instructions
MDX in storybook 
`Design System/Definition List`

### DIFF
--- a/src/components/definition-list/definition-list.stories.mdx
+++ b/src/components/definition-list/definition-list.stories.mdx
@@ -1,0 +1,64 @@
+import { Meta, Story, Preview, Props} from '@storybook/addon-docs/blocks';
+import {useState} from 'react';
+import { Dl, Dt, Dd } from "./"
+import StyledSystemProps from '../../../.storybook/utils/styled-system-props';
+
+<Meta title="Design System/Definition List" parameters={{ info: { disable: true }}} />
+
+# Definition List
+
+## Related Components
+Definition List most of the time is in use with other components. Look at the examples below
+
+- [Tile component with Definition List](/?path=/docs/design-system-tile--with-definition-list-default "Tile component with Definition List").
+- [Accordion component with Definition List](/?path=/docs/design-system-accordion--with-a-definition-list "Accordion component with Definition List").
+
+## Quick Start
+
+```js
+import { Dl, Dt, Dd } from "carbon-react/lib/components/definition-list";
+```
+
+### Default
+
+<Preview>
+  <Story name="default">
+    <Dl>
+      <Dt>First</Dt>
+      <Dd>Description</Dd>
+      <Dt>Second</Dt>
+      <Dd>Description</Dd>
+      <Dt>Third</Dt>
+      <Dd>Description</Dd>
+    </Dl>
+  </Story>
+</Preview>
+
+### Conditional rendering with `<React.Fragment />`
+*CAUTION: don't use nested `<React.Fragment />`*
+
+<Preview>
+  <Story name="with conditional rendering" parameters={{ chromatic: { disable: true }}}>
+      <Dl>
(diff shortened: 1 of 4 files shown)


edleeks87 (dismissed):

src/components/definition-list/definition-list.stories.mdx
```diff
@@ -0,0 +1,58 @@
+import { Meta, Story, Preview, Props} from '@storybook/addon-docs/blocks';
+import {useState} from 'react';
+import { Dl, Dt, Dd } from "./"
+import StyledSystemProps from '../../../.storybook/utils/styled-system-props';
+
+<Meta title="Design System/Definition List" parameters={{ info: { disable: true }}} />
+
+# Definition List
```
DipperTheDan, line 8: I can see you've created a new MDX for Definition List. We have other examples of this component in Tile and Accordion components as Definition List was meant to be just used inside of Tile and Accordion. It might be an idea to put links to Tile and Accordion examples with Definition List in this MDX example. I'm thinking it might point our consumers in the right direction.

DipperTheDan (approved):

src/components/definition-list/definition-list.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+          <Dd>Description inside of React Fragment</Dd>
+        </>)}
+      </Dl>
+  </Story>
+</Preview>
+
+## Props
+
+### Dl
+<StyledSystemProps of={Dl}  noHeader spacing/>
```
edleeks87, line 58: I don't think this captures all the props that Dl accepts, ` w`, `dtTextAlign` and `ddTextAlign` aren't documented for example. The `Dd` and `Dt` prop tables look fine to me.

```suggestion
### Dl
<Props of={Dl} />
<StyledSystemProps of={Dl}  noHeader spacing/>
```

edleeks87: My bad `StyledSystemProps` now supports the prop tables of the component as well

edleeks87 (dismissed):

src/components/definition-list/definition-list.stories.mdx
```diff
@@ -0,0 +1,64 @@
...
+      <Dd>Description</Dd>
+    </Dl>
+  </Story>
+</Preview>
+
+### Conditional rendering with `<React.Fragment />`
+*CAUTION: don't use nested `<React.Fragment />`*
+
+<Preview>
+  <Story name="with conditional rendering" parameters={{ chromatic: { disable: true }}}>
```
Alec-Walker-Sage, line 41: I think this shouldn't have its chromatic disabled as if anything with the conditional rendering broke then chromatic would be able to show us

src/components/definition-list/definition-list.stories.mdx
Alec-Walker-Sage, line 58:
```suggestion
<StyledSystemProps of={Dl} noHeader spacing/>
```

src/components/definition-list/definition-list.stories.mdx
edleeks87: The unit tests would catch it as well though imo so it depends whether we want to use a snapshot as well

edleeks87 (approved):

Alec-Walker-Sage (approved):

DipperTheDan (approved):

PR #3741, opened 2021-02-24
feat(icon): filter out unwanted props (FE-3599)

### LINKED ISSUE #3709: Icon is missing Typescript definition file
### Current behaviour

Component is currently missing a typescript definition file, requiring the use of an override ```declare module "carbon-react/XYZ"``` in order to progress without typescript failures.
### Expected behaviour

A Typescript definition file exists

### Your environment

| carbon-react     | v59.0
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

grabkowski:
### Proposed behaviour
Unnecessary props should not be passed as it could affect the display of the Component.

Typescript definition types should be documented.

### Current behaviour
All props defined on the Icon Component are being passed to the styled component.

Typescript definition types have no documentation.

### Checklist

Fixes #3709

### DIFF
--- a/src/components/icon/icon.component.js
+++ b/src/components/icon/icon.component.js
@@ -51,26 +51,29 @@ const Icon = React.forwardRef(
       }
     };
 
+    const styleProps = {
+      bg,
+      bgTheme,
+      bgSize,
+      bgShape,
+      color,
+      disabled,
+      fontSize,
+      iconColor,
+      mr,
+      ml,
+      tabIndex,
+      type: iconType(),
+    };
+
     const icon = (
       <StyledIcon
         ref={ref}
-        bgSize={bgSize}
-        bgShape={bgShape}
-        bgTheme={bgTheme}
-        fontSize={fontSize}
-        iconColor={iconColor}
-        disabled={disabled}
-        color={color}
-        bg={bg}
-        type={iconType()}
         key="icon"
         className={className || null}
-        {...tagComponent("icon", rest)}
         data-element={iconType()}
-        mr={mr}
-        ml={ml}
-        tabIndex={tabIndex}
-        {...rest}
+        {...tagComponent("icon", rest)}
+        {...styleProps}
       />
     );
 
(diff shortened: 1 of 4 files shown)


src/components/icon/icon.d.ts
```diff
@@ -1,23 +1,39 @@
 import * as React from 'react';
 
 export interface IconProps {
-  className?: string;
```
sjohnson-sage, line 4: `className` is still there in the component prop types so should be in here as well still - unless you meant to remove it from the component prop types as well?

src/components/icon/icon.component.js
```diff
@@ -51,26 +51,29 @@ const Icon = React.forwardRef(
...
-        bg={bg}
-        type={iconType()}
         key="icon"
         className={className || null}
         {...tagComponent("icon", rest)}
         data-element={iconType()}
-        mr={mr}
-        ml={ml}
-        tabIndex={tabIndex}
-        {...rest}
```
sjohnson-sage, line 73: I thought we decided we wanted our pattern to be always spread `...rest` on the root element of a component so that users can pass `data-element` etc to it?

src/components/icon/icon.d.ts
grabkowski: That prop should be no longer used, I'm leaving it in propTypes to avoid warnings, but I think we're fine to remove it in definition types.

src/components/icon/icon.component.js
grabkowski: data-element is already passed using the `tagComponent` method. As for the other props - this is the task description: "Prevent unwanted props from being passed down to Icon".

sjohnson-sage: I think you'd removed `tagComponent` when I commented this but looks like it's back now so that'll do

src/components/icon/icon.d.ts
sjohnson-sage: Let's add a "DO NOT USE" or something to the prop description then because it's documented in the `Icon` prop table, and nothing says not to use it - so a user could rightly attempt to use it which would give them a Typescript error

sjohnson-sage: Yeah sounds good to me 👍🏻

sjohnson-sage (approved):

mkrds (approved):

Alec-Walker-Sage (approved):

PR #3766, opened 2021-03-01
feat(tooltip): add support for overriding default flip placements FE-3771

edleeks87:
### Proposed behaviour

Adds `flipOverrides` prop to support passing an array of placements to override the default flip
behaviour: the order of the array sets the precedence of these overrides. See `popper` docs for more
information https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements

Adds `tooltipFlipOverrides` to `Icon` to support overriding the default flip behaviour of `Tooltip`.

Adds `tooltipFlipOverrides` to `ValidationIcon` to support overriding the default flip behaviour of
`Tooltip`, if no value passed to prop and `isPartOfInput` then the flip behaviour overrides to `["top", "bottom"]`

Adds `tooltipFlipOverrides` to `Help` to support overriding the default flip behaviour of `Tooltip`.

Sets `Tooltip` to flip "top" or " bottom" if `validationOnLabel` is false in `ButtonToggleGroup`

### Current behaviour

There is no support for overriding the default flip behaviour of the `Tooltip`, if there is no room in the original position then it flips tot he opposite (left -> right) which creates an issue if it covers an input for example.

### Checklist

### Additional context

Before changes tooltip flips to cover input:

image

After changes the tooltip flips to "top" or "bottom" if part of an input:

image

image

### Testing instructions

Check stories that render tooltips (validation stories for inputs), I have added knobs to `Help` and `Icon` `stories.js`

https://codesandbox.io/s/carbon-quickstart-forked-267kt

### DIFF
--- a/src/components/icon/icon.component.js
+++ b/src/components/icon/icon.component.js
@@ -24,6 +24,7 @@ const Icon = React.forwardRef(
       tooltipVisible,
       tooltipBgColor,
       tooltipFontColor,
+      tooltipFlipOverrides,
       tabIndex,
       isPartOfInput,
       inputSize,
@@ -90,6 +91,7 @@ const Icon = React.forwardRef(
           inputSize={inputSize}
           bgColor={tooltipBgColor}
           fontColor={tooltipFontColor}
+          flipOverrides={tooltipFlipOverrides}
         >
           {icon}
         </Tooltip>
@@ -99,6 +101,8 @@ const Icon = React.forwardRef(
     return icon;
   }
 );
+
+const placements = ["top", "bottom", "left", "right"];
 
 Icon.propTypes = {
   /**
@@ -146,13 +150,29 @@ Icon.propTypes = {
   /** The message string to be displayed in the tooltip */
   tooltipMessage: PropTypes.string,
   /** The position to display the tooltip */
-  tooltipPosition: PropTypes.oneOf(["top", "bottom", "left", "right"]),
+  tooltipPosition: PropTypes.oneOf(placements),
   /** Control whether the tooltip is visible */
   tooltipVisible: PropTypes.bool,
   /** Override background color of the Tooltip, provide any color from palette or any valid css color value. */
   tooltipBgColor: PropTypes.string,
   /** Override font color of the Tooltip, provide any color from palette or any valid css color value. */
   tooltipFontColor: PropTypes.string,
+  /** Overrides the default flip behaviour of the Tooltip, must be an array containing some or all of ["top", "bottom", "left", "right"] (see https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements) */
+  tooltipFlipOverrides: (props, propName) => {
+    const prop = props[propName];
+    const isValid =
+      prop &&
+      Array.isArray(prop) &&
+      prop.every((placement) => placements.includes(placement));
+
+    if (!prop || isValid) {
+      return null;
+    }
+    return new Error(
+      // eslint-disable-next-line max-len
+      `The \`${propName}\` prop supplied to \`Icon\` must be an array containing some or all of ["top", "bottom", "left", "right"].`
+    );
+  },
   /** @ignore @private */
   isPartOfInput: PropTypes.bool,
   /** @ignore @private */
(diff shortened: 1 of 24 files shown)


src/components/icon/icon.component.js
```diff
@@ -146,13 +150,29 @@ Icon.propTypes = {
...
-  tooltipPosition: PropTypes.oneOf(["top", "bottom", "left", "right"]),
+  tooltipPosition: PropTypes.oneOf(placements),
   /** Control whether the tooltip is visible */
   tooltipVisible: PropTypes.bool,
   /** Override background color of the Tooltip, provide any color from palette or any valid css color value. */
   tooltipBgColor: PropTypes.string,
   /** Override font color of the Tooltip, provide any color from palette or any valid css color value. */
   tooltipFontColor: PropTypes.string,
+  /** Overrides the default flip behaviour of the Tooltip, must be an array containing some or all of ["top", "bottom", "left", "right"] (see https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements) */
+  tooltipFlipOverrides: (props, propName) => {
```
sjohnson-sage, line 161: I think you should add this prop validation into `Help` as well

src/components/tooltip/tooltip.d.ts
```diff
@@ -19,6 +20,7 @@ export interface TooltipProps {
     inputSize?: 'small' | 'medium' | 'large';
     bgColor?: string;
     fontColor?: string;
```
sjohnson-sage, line 22: It's good to have the prop description comment in the ts file as well (I know some haven't been added here) but I imagine @grabkowski is going through all of the prop definitions and adding them

src/components/validations/validation-icon.component.js
```diff
@@ -81,6 +87,8 @@ ValidationIcon.propTypes = {
   iconId: PropTypes.string,
   /** Define position of the tooltip */
   tooltipPosition: PropTypes.string,
+  /** Overrides the default flip behaviour of the Tooltip, must be an array containing some or all of ["top", "bottom", "left", "right"] (see https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements) */
+  tooltipFlipOverrides: PropTypes.array,
```
sjohnson-sage, line 91: I think you should add the propType validation here as well rather than allowing any array

src/utils/helpers/options-helper/options-helper.d.ts
```diff
@@ -343,13 +345,13 @@ export interface FlexBoxProps {
   alignContent?: string;
   justifyItems?: string;
   justifyContent?: string;
-  felxWrap?: string;
+  flexWrap?: string;
```
DipperTheDan, line 348: Nice one spotting this and line 354 👍🏻

src/components/tooltip/tooltip.d.ts
edleeks87: Yeah I left them blank for that reason but happy to add them in if Kuba wants me to

src/components/validations/validation-icon.component.js
edleeks87: Fair point, I was weighing up whether we wanted three errors (ValidationIcon -> Icon -> Tooltip) to be thrown when an invalid array is passed, but if you think it's worth it I'll add it in here and in `Help`

sjohnson-sage: Yeah I see what you mean but for the sake of total clarity for the user I think it should be the component they try to use that emits an error if the prop type is incorrect, even if it's children emit errors as well

If we ever convert the project to TypeScript it would be so much better for stuff like this!

src/components/tooltip/tooltip.d.ts
sjohnson-sage: My thinking was hopefully someday we'll convert to TypeScript and only have prop definitions in one place, and if we add the comments as we're going along it'll mean someone doesn't have to do it further down the line

Not going to block the PR if you don't do it here though!

edleeks87: Yeah agree, I'll keep a n eye on them and add any missing descriptions after Kuna's finished

DipperTheDan (dismissed):

sjohnson-sage (approved):

sjohnson-sage (approved):

DipperTheDan (dismissed):

sjohnson-sage (approved):

DlgSHi (dismissed):

DipperTheDan (approved):

DlgSHi (approved):

Issue #3559, opened 2021-01-07
Decimal: formatting numeric value imprecision in some locales


arturodrigues:
### Current behaviour
 
For some given locales, such as Spanish (es-ES) or Portuguese (pt-PT), when entering a number between 1000 and 9999 in the  Decimal component, the formatting of the numeric value is not quite correct: the locale's group separator only be used when the number is greater than or equal to ten thousand. Example:

| Number | Portuguese | Spanish |
| ---------------- | ---------------- | ---------------- |
| 100     | 100 | 100 |
| 1000     | 1000 | 1000 |
| 1000.23     | 1000,23 | 1000.23 |
| 10000     | 10 000 | 10.000|
| 10000.23     | 10 000,23 | 10.000,23 |
| 100000     | 100 000 | 100.000 |
| 1000000     | 1 000 000 | 1.000.000 |

### Expected behaviour
Formatted value appearing in the Decimal component should comply with [Intl object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat).

### Reproducible example
https://codesandbox.io/s/numberformatting-kijbl

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 55.0.0 |
| React            |  17.0.0 |
| Browser          | Chrome 87 |
| npm              | 6.14.8 |
| Operating System | Windows 10 |

sjohnson-sage:
FE-3500

PR #3799, opened 2021-03-08
feat(decimal): number formatting for different locales - FE-3500

maspychaj-sage:
### Proposed behaviour
Added props to allow different formatting numbers (custom delimiter and separator).

### Current behaviour
Only delimiter with comma and separator with dot was used. By default number wasn't formatted differently on change locale.

### Checklist

- [x ] Commits follow our style guide
<del>- [ ] Screenshots are included in the PR if useful</del>
- [x ] All themes are supported if required
- [x ] Unit tests added or updated if required
<del>- [ ] Cypress automation tests added or updated if required</del>
- [x ] Storybook added or updated if required
<del>- [ ] Typescript `d.ts` file added or updated if required</del>
<del>- [ ] Carbon implementation and Design System documentation are congruent</del>


sjohnson-sage:
Please don't move this PR out of draft until it has two Carbon approvals - this is to stop Chromatic running on code that is likely to change and wasting snapshots

sjohnson-sage:
It looks like there's been a bad rebase with master from the commits on this PR, can you sort it out please @maspychaj-sage?

alzieba (approved):

PaulSumner-Sage (dismissed):

src/__experimental__/components/decimal/decimal.stories.mdx
```diff
@@ -268,3 +268,27 @@ For more information check our [Validations](?path=/docs/documentation-validatio
...
+      const [state, setState] = useState("0.01");
+      const setValue = ({ target }) => {
+        setState(target.value.rawValue);
+      };
+      const [statePt, setStatePt] = useState("0.01");
+      const setValuePt = ({ target }) => {
+        setStatePt(target.value.rawValue);
+      };
+      return <>
+        <Decimal label="Decimal - pt locale" value={statePt} onChange={setValuePt} locale="pt" delimiter=" " separator="," />
```
chrisbarber86, line 286: Why do we need to pass in `locale` AND `delimiter`/`separator`. If we are being told the delimiter and the separator then we shouldn't need the locale (or vice versa)

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -173,6 +175,18 @@ class Decimal extends React.Component {
...
+      number: {
+        format: {
+          delimiter: this.props.delimiter,
+          separator: this.props.separator,
+        },
+      },
+    };
+
+    I18n.locale = this.props.locale;
+    I18n.translations[I18n.locale] = numberFormat;
```
chrisbarber86, line 188: Why are we overwriting the translations for the locale? That's not something this component should be doing.

if the component receives `locale` then we should use I18n.t to get the format. Or we remove local and we just use the provided props.

chrisbarber86 (changes requested):

src/__experimental__/components/decimal/decimal.stories.mdx
maspychaj-sage: I used the same approach as in code example: https://codesandbox.io/s/numberformatting-kijbl?file=/src/locale.js - we can change it to use delimiter and separator only.

src/__experimental__/components/decimal/decimal.component.js
maspychaj-sage: @chrisbarber86  - removing usage of I18n.locale requires also changes for toStandardDecimal - just using delimiter and separator from props in formatDecimal functions cause change of hiddenValue. Do you really think that'll be better in that case?

src/__experimental__/components/decimal/decimal.stories.mdx
chrisbarber86: I think i would prefer that we expect delimiter and separator to be passed in. Then the consuming application can use whatever library they want for I18n.

src/__experimental__/components/decimal/decimal.component.js
chrisbarber86: my point was you are changing the value of I18n.locale AND I18n.translations[I18n.locale] - both of which are defined and owned by the application - this component has no right to override them. A requirement of using I18n here would be that the consuming application knows to setup:

```javascript
    {
      number: {
        format: {
          delimiter: ",",
          separator: ".",
        },
      }
    }
```

maspychaj-sage: In that case there is any better option to decide how number will be formatted? Right now it works according to browser settings and can't be changed, am I right?

jamime (changes requested):
I think this has missed the intended purpose of FE-3500.

There is an associated GitHub issue that has more information https://github.com/Sage/carbon/issues/3559.

The goal is to support the `es-ES` and `pt-PT` locales that only display a delimiter once there is five or more integers.

There is more detail in the linked issue. 

Once this task is complete the decimal component should not use `i18n-js` for decimal formatting. It should use [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat).

You also need access to the delimiter and separator to allow/prevent keypresses. You can get this by calling `Intl.NumberFormat().formatToParts()`. There is another PR in flight that changes this component to allow any input, at that point you won't need to do this.
I think we should wait until that PR is merged before working on this as there is too much overlap.

Feel free to DM me on slack if you have any questions.

PaulSumner-Sage (dismissed):

alzieba (dismissed):

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -138,9 +138,21 @@ class Decimal extends React.Component {
   };
 
   removeDelimiters = (value) => {
-    const format = I18nHelper.format();
-    const delimiter = `\\${format.delimiter}`;
-    const delimiterMatcher = new RegExp(`[${delimiter}]*`, "g");
+    /* If a value is passed in that is a number but has too many delimiters in succession, we want to handle this
```
DipperTheDan, line 141: From 141 to 152 isn't required as this block of code was moved to the `toStandardDecimal` function. Here is the link to where it is now.
https://github.com/Sage/carbon/pull/3799/files#diff-41d1ef079632dd8b6d4e6270f83eb09c1dcfc8017b6ee137810ea9d3d1f41dc2R200

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -723,28 +703,34 @@ describe("Decimal", () => {
...
         });
 
         it("supports a precision of 0", () => {
-          render({ defaultValue: "12345", precision: 0 });
+          render({ defaultValue: "12345", precision: 0, ...itProps });
           expect(value()).toBe("12.345");
           expect(hiddenValue()).toBe("12345");
         });
 
+        it("has a default precision of 2 (crop)", () => {
```
DipperTheDan, line 719: Test will need a different description as the value isn't cropped and the value shouldn't be cropped either.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -779,16 +765,34 @@ describe("Decimal", () => {
...
+          expect(hiddenValue()).toBe("12345.600");
+        });
+
+        it("supports having a longer number on default", () => {
+          render({ defaultValue: "123123.99999", precision: 3, ...itProps });
+          expect(value()).toBe("123.123,99999");
+          expect(hiddenValue()).toBe("123123.99999");
+        });
+
+        it("supports having a longer number on change", () => {
```
DipperTheDan, line 787: This test has a different description but looks exactly the same as the test above?

DipperTheDan (commented):
Left some comments but you should also update the `decimal.stories.js` file as that has a `locale` story. Maybe update it for the purposes of your PR or it might be easier to include a knob in the default story for the new `locale` prop.

src/__experimental__/components/decimal/decimal.spec.js
maspychaj-sage: It's not the same case - defaultValue vs value - second test is for changing after defaultValue is not passed.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -115,31 +112,6 @@ describe("Decimal", () => {
     return press({ clipboardData }, where, "paste");
   };
 
-  let translations;
-
-  testStyledSystemMargin(
```
DipperTheDan, line 120: We still need this test to cover that the styled-system margin props are applied to Decimal.

DipperTheDan (commented):
Just one small comment about a styled-system test you have deleted accidentally. 

It's just occurred to me in this review that you will need to add the new `locale` prop to the `decimal.d.ts` file too. Apologies I didn't pick up on that sooner! 😢

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -145,9 +144,8 @@ class Decimal extends React.Component {
   };
 
   removeDelimiters = (value) => {
-    const format = I18nHelper.format();
-    const delimiter = `\\${format.delimiter}`;
-    const delimiterMatcher = new RegExp(`[${delimiter}]*`, "g");
+    const delimiter = `\\${this.getSeparator("group")}`;
+    const delimiterMatcher = new RegExp(`[${delimiter} ]*`, "g");
```
jamime, line 148:
```suggestion
    const delimiterMatcher = new RegExp(`[\\${this.getSeparator("group")} ]*`, "g");
```

lets combine these lines :)

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -145,9 +144,8 @@ class Decimal extends React.Component {
...
     const noDelimiters = value.replace(delimiterMatcher, "");
```
jamime, line 149:
```suggestion
    return value.replace(delimiterMatcher, "");
```

src/__experimental__/components/decimal/decimal.component.js
```diff
@@ -189,10 +205,17 @@ class Decimal extends React.Component {
     }
 
     const withoutDelimiters = this.removeDelimiters(i18nValue);
-    const { separator } = I18nHelper.format();
+    const separator = this.getSeparator("decimal");
     return withoutDelimiters.replace(new RegExp(`\\${separator}`, "g"), ".");
   };
 
+  getSeparator(separatorType) {
+    const numberWithGroupAndDecimalSeparator = 1000.1;
```
jamime, line 213: This doesn't work for `es-ES` or `pt-PT`, these are the locales in the original issue, I think you need to change the value to `10000.1`
```suggestion
    const numberWithGroupAndDecimalSeparator = 10000.1;
```

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -125,6 +96,7 @@ export const Locale = () => {
       key={precision} // Don't do this in production, testing purposes only
       {...props}
       precision={precision}
+      locale="it"
```
jamime, line 99: Please make this configurable using a knob.

src/__experimental__/components/decimal/decimal.stories.js
```diff
@@ -164,7 +136,7 @@ Default.story = {
 };
 
 Locale.story = {
-  name: "locale",
+  name: "locale - it format",
```
jamime, line 139:
```suggestion
  name: "locale",
```

src/utils/helpers/i18n/i18n.js
```diff
@@ -58,15 +58,6 @@ const I18nHelper = {
       groupSize: 3,
     };
 
-    // valueToFormat is "en" formatted.
```
jamime, line 61: Please restore this, this could be consumed by other projects. Looks like we don't have a unit test for this functionality. Please can you add one.

jamime (changes requested):
Please can you add unit tests for the `es-es` and `pt-pt` locales. The linked issue has examples of the expected results. The current PR does not support these locales.

src/utils/helpers/i18n/i18n.js
jamime: Unfortunately it is not in a `__internal__` folder so we consider it part of the public API. This helper is available as 
```js
import {formatDecimal} from "carbon-react/lib/utils/helpers/i18n";
```

In the short term we plan to remove these helpers so I think it's OK to remove it like you have as long as it's documented in the commit as a breaking change. 

I think the commit message should be something like this


```
feat(decimal): add locale support

BREAKING CHANGE: The decimal component no longer uses `i18n-js` for decimal formatting, you should use the `locale` prop instead. 

BREAKING CHANGE: `formatDecimal` no longer supports the `round` option first introduced in v73.0.0

closes FE-3500
```

jamime (changes requested):
1. Using the `pt-pt` locale and entering `123 456 789` the value is not automatically formatted. In `en` we convert `123,456,789` to `123,456,789.00`

1. With en if the user enters `12,34` we will correct the delimiter `1,234`. This did not work for `pt-pt`.

Please ensure unit tests are added for these cases.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,70 @@ describe("Decimal", () => {
...
+      describe("pt", () => {
+        const ptProps = { locale: "pt-PT" };
+
+        it.each([
+          ["1000.23", "1000,23"],
+          ["10000.00", "10\xa0000,00"],
+          ["10000.23", "10\xa0000,23"],
+          ["100000.00", "100\xa0000,00"],
+          ["1000000.00", "1\xa0000\xa0000,00"],
+        ])(" format %s to %s", (a, b) => {
```
jamime, line 623:
```suggestion
        ])("format %s to %s", (a, b) => {
```

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,70 @@ describe("Decimal", () => {
...
+          render({ value: a, ...ptProps });
+          expect(value()).toBe(b);
+          expect(hiddenValue()).toBe(a);
+        });
+
+        it(" handles a value that has white-spaces", () => {
```
jamime, line 629:
```suggestion
        it("handles a value that has white-spaces", () => {
```

Please also add a test for 

1. the space being in wrong location.

e.g. `1 0000,00` it should correct it.

2. two spaces

e.g. `10 000` it should not format it.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -990,7 +1028,7 @@ describe("Decimal", () => {
       setProps({ value: "     " });
       blur();
       expect(value()).toBe("     ");
-      expect(hiddenValue()).toBe("     ");
+      expect(hiddenValue()).toBe("");
```
jamime, line 1031: Why did this change?

jamime (changes requested):
When using the `en` local the component should not remove any whitespace, instead it should not format the number so the consumers can display a validation message.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,86 @@ describe("Decimal", () => {
...
+
+        it(" handles a value that has white-spaces in wrong place", () => {
+          render({ ...ptProps });
+          type("1 0000,00");
+          blur();
+          expect(value()).toBe("10\xa0000,00");
+          expect(hiddenValue()).toBe("10000.00");
+        });
+
+        it(" handles a value that has multiple white-spaces", () => {
```
jamime, line 645: This value should not be formatted, please see previous review.

jamime (changes requested):
Value with multiple delimiters should not be formatted, please compare with en and see comments on previous review.

Please remove leading space from `it` blocks.

jamime (dismissed):

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,94 @@ describe("Decimal", () => {
...
+      describe("pt", () => {
+        const ptProps = { locale: "pt-PT" };
+
+        it.each([
+          ["1000.23", "1000,23"],
+          ["10000.00", "10\xa0000,00"],
+          ["10000.23", "10\xa0000,23"],
+          ["100000.00", "100\xa0000,00"],
+          ["1000000.00", "1\xa0000\xa0000,00"],
+        ])(" format %s to %s", (a, b) => {
```
jamime, line 623: Please remove leading space.

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,94 @@ describe("Decimal", () => {
...
+      describe("es", () => {
+        const esProps = { locale: "es-ES" };
+
+        it.each([
+          ["1000.23", "1000,23"],
+          ["10000.00", "10.000,00"],
+          ["10000.23", "10.000,23"],
+          ["100000.00", "100.000,00"],
+          ["1000000.00", "1.000.000,00"],
+        ])(" format %s to %s", (a, b) => {
```
jamime, line 607: Please remove leading space.

jamime (dismissed):

DipperTheDan (dismissed):

src/__experimental__/components/decimal/decimal.spec.js
```diff
@@ -614,36 +595,94 @@ describe("Decimal", () => {
...
+          render({ value: a, ...esProps });
+          expect(value()).toBe(b);
+          expect(hiddenValue()).toBe(a);
+        });
       });
 
-      afterAll(() => {
-        I18n.locale = undefined;
+      describe("pt", () => {
+        const ptProps = { locale: "pt-PT" };
```
DipperTheDan, line 615: I think it would be great if you could add a unit test that covers `1,1,1,1` not being converted to `1.1.1.1` with the `pt-PT` locale, as per the bug @DlgSHi found. It means when we come to refactor the component to be a functional component in the near future, we have a unit test that already covers this scenario.

DipperTheDan (dismissed):

DipperTheDan (approved):

DlgSHi (dismissed):

jamime (changes requested):
Lets add the same test cases that are in cypress but by testing the `rawValue` instead of `hiddenValue`.

https://github.com/Sage/carbon/blob/b9eadd94c1794e02a33ad054139fb6c30930eae1/cypress/features/regression/experimental/decimalLocale.feature#L10-L20

Please duplicate this test for each case, you can use `it.each`.

https://github.com/Sage/carbon/blob/b9eadd94c1794e02a33ad054139fb6c30930eae1/src/__experimental__/components/decimal/decimal.spec.js#L861-L874

We should also have these test cases from the original issue
https://github.com/Sage/carbon/issues/3559

| Number | Portuguese | Spanish |
| ---------------- | ---------------- | ---------------- |
| 100     | 100 | 100 |
| 1000     | 1000 | 1000 |
| 1000.23     | 1000,23 | 1000.23 |
| 10000     | 10 000 | 10.000|
| 10000.23     | 10 000,23 | 10.000,23 |
| 100000     | 100 000 | 100.000 |
| 1000000     | 1 000 000 | 1.000.000 |

Please do not rebase instead push a `test(decimal): add locale tests` commit to make reviewing easier.

DlgSHi (approved):

DipperTheDan (approved):

jamime (approved):

Issue #3748, opened 2021-02-25
Ommited property for "rel" in Link 


robertJablonskiSage:
### Current behaviour
```
No way to add "rel to link"
<Link target="_blank" rel="noopener noreferrer" ...
```

### Expected behaviour
```
 "rel" is a default and attribute is added to link properly 
```

### Reproducible example
```
call <Link target="_blank" rel="noopener noreferrer" 
```

### Suggested solution(s)
```
"rel" attribute should be set, and have default value if selected link will be open in new browser tab (target="_blank") 
```

### Your environment
 "carbon-react": "66.0.3"

sjohnson-sage:
FE-3803

PR #3815, opened 2021-03-12
feat: add `rel` prop to Link component -  FE-3803

dawidzarzycki:
fixes: #3748
It allows users to put `rel` as a property in our link component
rel is a default property supported by `a` tag

### Current behaviour

no possibility to add rel property

### Checklist

### Testing instructions
New story added into storybook
Link -> rel

### DIFF
--- a/src/components/link/link.stories.mdx
+++ b/src/components/link/link.stories.mdx
@@ -95,6 +95,17 @@ The `href` prop provides the ability for the Link to take the user to a differen
   </Story>
 </Preview>
 
+### rel
+
+The `rel` prop provides the ability for the Link to set relationship between the current document 
+and the linked document
+
+<Preview>
+  <Story name="with rel" parameters={{ chromatic: { disable: true } }}>
+    <Link href="https://carbon.sage.com/" target="_blank" rel="noreferrer noopener">This is a link</Link>
+  </Story>
+</Preview>
+
 ### Icon
 
 The `icon` prop allows you to render an icon.
--- a/cypress/features/regression/link.feature
+++ b/cypress/features/regression/link.feature
@@ -45,3 +45,8 @@ Feature: Link component
   Scenario: Check skip link is not visible without focus
     When I open "Link" component page "is skip link" in no iframe
     Then Skip link is not visible
+
+  @positive
+  Scenario: Check rel attr is able to be used in a link component
+    When I open "Link" component page "with rel" in no iframe
+    Then link has a rel attribute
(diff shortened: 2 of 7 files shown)


grabkowski:
I'm pretty sure that the commit message should be `feat(link): add rel prop`

bulaj:
Looking at the actual reason why we do this (to apply "noopener noreferrer" for security and privacy), I think it would be better if we added `rel="noopener noreferrer"` automatically if `target="_blank"` is present. Or addind a prop `<Link external />` that will automatically set target and rel. With this approach we cover the actual business case and good example to users instead of just exposing a prop they can specify anyway due to ...rest.

dawidzarzycki:
> Looking at the actual reason why we do this (to apply "noopener noreferrer" for security and privacy), I think it would be better if we added `rel="noopener noreferrer"` automatically if `target="_blank"` is present. Or addind a prop `<Link external />` that will automatically set target and rel. With this approach we cover the actual business case and good example to users instead of just exposing a prop they can specify anyway due to ...rest.

@bulaj  I think that we shouldn't make it so restrictive as `rel` can have different values. So If someone will need to use it then he will be able to use it. I think that making it more flexy is better option

bulaj:
> > Looking at the actual reason why we do this (to apply "noopener noreferrer" for security and privacy), I think it would be better if we added `rel="noopener noreferrer"` automatically if `target="_blank"` is present. Or addind a prop `<Link external />` that will automatically set target and rel. With this approach we cover the actual business case and good example to users instead of just exposing a prop they can specify anyway due to ...rest.
> 
> @bulaj I think that we shouldn't make it so restrictive as `rel` can have different values. So If someone will need to use it then he will be able to use it. I think that making it more flexy is better option

Yeah but this doesn't explicitely help with the actual use case (increase security). 
This way user has to remember to add rel at all times he uses external link. We should cover basic security pitfalls. You can always add a rel with rest (or specify it anyway with the rest of the logic).

src/components/link/link.stories.mdx
```diff
@@ -95,6 +95,17 @@ The `href` prop provides the ability for the Link to take the user to a differen
...
 </Preview>
 
+### rel
+
+The `rel` prop provides the ability for the Link to set relationship between the current document 
+and the linked document
+
+<Preview>
+  <Story name="with rel" parameters={{ chromatic: { disable: true } }}>
+    <Link rel="alternate">This is a link</Link>
```
bulaj, line 105: I'd go with `<Link target="_blank" rel="noreferrer noopener">...` example as it is most common case.

bulaj (approved):

grabkowski (dismissed):

bulaj (dismissed):

KatarzynaQA (dismissed):

KatarzynaQA (dismissed):

KatarzynaQA (dismissed):

bulaj (approved):

grabkowski (approved):

KatarzynaQA (approved):

bulaj (approved):

PR #3849, opened 2021-03-22
docs: recommend using a locale object instead of using i18next FE-3983

jamime:
Update the RFC to recommend shipping a local object instead of using `i18next`.

---

Please use 👍  if you approve.


mmaciorowski-sage (approved):

mmaciorowski-sage (dismissed):

thehiggster (dismissed):

sjohnson-sage (approved):

rfcs/text/i18n.md
```diff
@@ -6,40 +6,38 @@ Introduce first class support for internationalisation and localisation within `
 
 # Basic example
 
-The proposal is to introduce a new component `I18nProvider` that allows our consumers to provide a 
-[`react-i18next`](https://github.com/i18next/react-i18next) compatible<sup>[1] [2]</sup> `t` function.
+The proposal is to create a new `I18nProvider` component that allows carbon components to look 
+up translations by key. We will provide `en-GB` translations but we will accept pull requests to add translations for
+other common languages. We will only add complete locales to the `carbon-react` project. Consumers are able to extend
```
chrisbarber86, line 11: Potentially this could add quite a bit of overhead going forwards. For example if we add the following:

* en-GB
* en-US
* en-CA
* fr-CA
* fr-FR
* de-DE
* es-ES
* es-US

Every time a new component gets added, or an existing one is changed - we will need to enforce this line about `complete locales` - requiring that the translations are updated for all those locales (which isn't an exhaustive list). These would also require review - which the core team might not be able to do without outside help

For me it would make sense for Carbon to default always to english, and allow consumers an easy way to provide translations for their own needs in their application (e.g. using I18next or I18n.js)

rfcs/text/i18n.md
```diff
@@ -6,40 +6,38 @@ Introduce first class support for internationalisation and localisation within `
...
-We will also deprecate all [`i18n` helpers][3]. Continuing to support these is beyond the scope of a component library
+We will also deprecate all [`i18n` helpers][1]. Continuing to support these is beyond the scope of a component library
 and is better handled by a dedicated library such as [`numeral.js`](http://numeraljs.com/). Removing these will further
 reduce the complexity of this project and allow for more advanced configurations by our consumers. Where we use these
 helpers internally, we will use a dedicated library if appropriate. We will ensure that any new i18n helpers are
 considered as private methods i.e. marked as `@private` or not exported.
 
-Finally, each translation will be overridable by our consumers by providing a string prop to the component. This string 
-can also be provided by `react-i18next`.
+Finally, each translation will be overridable by our consumers by providing a prop to the component or by creating
```
chrisbarber86, line 31: I would suggest providing a locale is the main mechanism and providing props is seen more for edge cases where the content is quite different for a particular scenario. Overriding translations via props would work, but wouldn't scale particularly well.

rfcs/text/i18n.md
```diff
@@ -48,248 +46,137 @@ the bundle size and reducing performance.
...
+import { toWords } from "number-to-words";
 
-function App() {
+export default function App() {
   return (
-    <Suspense fallback="Loading...">
-      <Page />
-    </Suspense>
+    <div className="App">
+      {/* Default enGB */}
```
dawidzarzycki, line 119: is these two styles of comments are on purpose?

chrisbarber86 (changes requested):

rfcs/text/i18n.md
```diff
@@ -48,248 +46,137 @@ the bundle size and reducing performance.
...
 ```js
-// carbon-react/lib/components/Example/index.js
-import React, { Suspense, useContext } from "react";
-import Context from "../__internal__/I18nContext";
-export default ({nextLabel}) => {
-  const t = useContext(Context);
-  return <span>{nextLabel || t("example.next")}</span>;
+// carbon/locales/enGB.js
+export default {
+  locale: "enGb",
```
DlgSHi, line 62:
```suggestion
  locale: "en-Gb",
```

DlgSHi (approved):

rfcs/text/i18n.md
jamime: I cant use `//` because its in the JSX, the other comments are single line so I used `//`.

rfcs/text/i18n.md
jamime: Good point, I've added details about using `i18next` as a custom locale and removed references to locales other than `enGB`.

chrisbarber86 (approved):

jamime:
Entering the final comment period which ends on 2021-03-26 1300.

mmaciorowski-sage (dismissed):

philearley (dismissed):

DlgSHi (dismissed):

jamime:
Squashing commits before merge.

sjohnson-sage (approved):

Issue #3520, opened 2020-12-21
Tabs control with set location = false onTabChange event adds item to browser history


ravi-ankam:
### Expected behaviour
I am implementing simple routing to the Tabs control .
In my example I have 3 tabs namely Standard, custom, favourites and added a simple routing using the tabName as a variable/parameter.
OnTabchange I am updating the browser history with a tabName to the base route. 
The issue I am facing now is, after I select the tabs, 
when I click on browser back button, For the first time the tab navigates back the the correct previous selected tab using the route/browser history value but on the second browser back click it fails to navigate to the previous tab selected.
It looks like, there is a default browser history update with tabid value. the number of items added to the browser history are more than expected items.
Attached the link for code sandbox Example code I used for your reference.
Please check and advise what I am doing wrong or if there is a better way to add routes the tabs and slected tab.

### Reproducible example
https://codesandbox.io/s/shy-dust-4l2vc

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     |48.3.1
| Browser          |Chrome

jamime:
We'll need to allow consumers to pass `history` too, with a default of `window.history` so users can use a memory router, subscribe to history changes and block navigation.

ianforster-sage:
@benwilsonsage should 'back' take you back to a previous tab in this scenario? The alternative would be to take you back to your previous page.

benwilsonsage:
Thanks for the scenario.

When developing modern 'single page apps', I can see why it's helpful to allow the user to step backward through dynamic changes on the same page, not just between pages.

However, Accounting (GAC live) today doesn't behave like a single page app, and the user can't use 'back' to get to previously viewed tabs.

I guess the answer has two levels...
- If developing a new app (e.g. Client Manager, something else...) it'd be good to have the ability to back up through tabs so we can deliver modern experiences.
- But if developing something to go-live in legacy Accounting, I'd suggest not enabling this, so the user has a consistent experience and isn't surprised by different behaviours in different places. Although I don't think it'd be a catastrophe by any means!

So I guess it depends on context - if this is for the Reporting page with the list of reports, I'd suggest the standard 'back' behaviour is fine.

MatthewSmythe:
@benwilsonsage this is for the Reporting tabs, so "new app" work.

jamime:
[...]

PR #3865, opened 2021-03-30
feat(tabs): remove internal support for history and url manipulation - FE-3484

bulaj:
closes: #3520

### Proposed behaviour
Removes baked in support for manipulation of window location as implementation is extremely limited and implementing the a feature to override it will be difficult to maintain. Consumers wishing to have this functionality should utilise the `onTabChange` and `selectedTabId` props and implement their own `history` implementation.

BREAKING CHANGE: Internal support for updating window location has been removed from the `Tabs`
component, the `setLocation` prop has also been removed as it serves no purpose beyond serving as a
guard for this functionality. Consumers who wish to have this functionality should utilise the
`onTabChange` and `selectedTabId` props with whatever `history` implementation they want to.

Removes dark background from focused `TabTitle` when `isTabSelected` is false 

### Current behaviour
There is no ability to pass external history to Tabs component.
 The current implementation only replaces and only uses hash history when `setLocation` prop is true.

### Checklist

- [ ] <del>Screenshots are included in the PR if useful
- [ ] <del>All themes are supported if required
- [ ] <del>Cypress automation tests added or updated if required
- [ ] <del>Carbon implementation and Design System documentation are congruent

### Testing instructions

Needs to be tested in external project containing react-router. See #3520.

https://codesandbox.io/s/demo-external-history-and-tabs-lwk3i

### DIFF
--- a/docs/usage-with-routing.stories.mdx
+++ b/docs/usage-with-routing.stories.mdx
@@ -20,9 +20,10 @@ An example of how to use the Carbon `Button` and `Link` components as links with
 Note the use of the `useHistory` hook from `react-router-dom`.
 
 ```jsx
-import React, { useCallback } from "react";
+import React, { useCallback, useState } from "react";
 import Button from "carbon-react/lib/components/button";
 import Link from "carbon-react/lib/components/link";
+import { Tabs, Tab } from "carbon-react/lib/components/tabs";
 
 import { BrowserRouter, Switch, Route, useHistory } from "react-router-dom";
 
@@ -44,6 +45,20 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
   return <Link onClick={onClick}>{children}</Link>;
 };
 
+const CarbonTabsWithReactRouter = ({ children }) => {
+  const history = useHistory();
+  const [activeTab, setActiveTab] = useState();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      history.push(tabId);
+      setActiveTab(tabId);
+    },
+    [history]
+  );
+
+  return <Tabs onTabChange={handleTabChange} selectedTabId={activeTab}>{children}</Tabs>;
+};
+
 const Home = () => (
   <div>
     <CarbonButtonWithReactRouter to="settings">
@@ -58,11 +73,25 @@ const Settings = () => (
   </div>
 );
 
+const Overview = () => (
+  <div>
+    <CarbonTabsWithReactRouter>
+      <Tab tabId="tab-1" title="Tab 1" key="tab-1">
+        Content for tab 1
+      </Tab>
+      <Tab tabId="tab-2" title="Tab 2" key="tab-2">
+        Content for tab 2
+      </Tab>
+    </CarbonTabsWithReactRouter>
+  </div>
+);
+
 const App = () => (
   <BrowserRouter>
     <Switch>
       <Route exact path="/" element={<Home />} />
       <Route path="/settings" element={<Settings />} />
+      <Route path="/overview" element={<Overview />} />
     </Switch>
   </BrowserRouter>
 );
@@ -74,9 +103,10 @@ An example of how to use the Carbon `Button` and `Link` components as links with
 Note the use of the `useNavigate` hook from `react-router-dom`.
 
 ```jsx
-import React, { useCallback } from "react";
+import React, { useCallback, useState } from "react";
 import Button from "carbon-react/lib/components/button";
 import Link from "carbon-react/lib/components/link";
+import { Tabs, Tab } from "carbon-react/lib/components/tabs";
 
 import { BrowserRouter, Routes, Route, useNavigate } from "react-router-dom";
 
@@ -98,6 +128,20 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
   return <Link onClick={onClick}>{children}</Link>;
 };
 
+const CarbonTabsWithReactRouter = ({ children }) => {
+  const navigate = useNavigate();
+  const [activeTab, setActiveTab] = useState();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      navigate(tabId);
+      setActiveTab(tabId);
+    },
+    [navigate]
+  );
+
+  return <Tabs onTabChange={handleTabChange} selectedTabId={activeTab}>{children}</Tabs>;
+};
+
 const Home = () => (
   <div>
     <CarbonButtonWithReactRouter to="settings">
@@ -112,11 +156,25 @@ const Settings = () => (
   </div>
 );
 
+const Overview = () => (
+  <div>
+    <CarbonTabsWithReactRouter>
+      <Tab tabId="tab-1" title="Tab 1" key="tab-1">
+        Content for tab 1
+      </Tab>
+      <Tab tabId="tab-2" title="Tab 2" key="tab-2">
+        Content for tab 2
+      </Tab>
+    </CarbonTabsWithReactRouter>
+  </div>
+);
+
 const App = () => (
   <BrowserRouter>
     <Routes>
       <Route path="/" element={<Home />} />
       <Route path="settings" element={<Settings />} />
+      <Route path="overview" element={<Overview />} />
     </Routes>
   </BrowserRouter>
 );
@@ -131,3 +189,4 @@ There are many components that you can use this method with, including:
 - `Link`
 - `MenuItem`
 - `Tab`
+- `Tabs`
--- a/src/components/tabs/tabs.component.js
+++ b/src/components/tabs/tabs.component.js
@@ -11,7 +11,6 @@ import styledSystemPropTypes from "@styled-system/prop-types";
 import Tab from "./tab";
 import Event from "../../utils/helpers/events/events";
 import tagComponent from "../../utils/helpers/tags/tags";
-import Browser from "../../utils/helpers/browser/browser";
 import StyledTabs from "./tabs.style";
 import TabsHeader from "./__internal__/tabs-header";
 import TabTitle from "./__internal__/tab-title";
@@ -30,7 +29,6 @@ const Tabs = ({
   selectedTabId,
   renderHiddenTabs = true,
   position = "top",
-  setLocation = true,
   extendedLine = true,
   size,
   borders = "off",
@@ -38,7 +36,6 @@ const Tabs = ({
   validationStatusOverride,
   ...rest
 }) => {
-  const firstRender = useRef(true);
   const tabRefs = useRef([]);
   const previousSelectedTabId = useRef(selectedTabId);
   const [selectedTabIdState, setSelectedTabIdState] = useState();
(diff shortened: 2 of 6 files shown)


docs/usage-with-routing.stories.mdx
```diff
@@ -23,6 +23,7 @@ Note the use of the `useHistory` hook from `react-router-dom`.
 import React, { useCallback } from "react";
 import Button from "carbon-react/lib/components/button";
 import Link from "carbon-react/lib/components/link";
+import {Tabs, Tab} from "carbon-react/lib/components/Tabs";
```
sjohnson-sage, line 26: The import path should have `/tabs` with a lower case `t` and gaps in this bit `{ Tabs, Tab }`

sjohnson-sage: I recommend setting up prettier to run on save in vscode (or whatever editor you use)

docs/usage-with-routing.stories.mdx
```diff
@@ -75,8 +111,10 @@ Note the use of the `useNavigate` hook from `react-router-dom`.
 
 ```jsx
 import React, { useCallback } from "react";
+import { createBrowserHistory } from 'history';
 import Button from "carbon-react/lib/components/button";
 import Link from "carbon-react/lib/components/link";
+import {Tabs, Tab} from "carbon-react/lib/components/Tabs";
```
sjohnson-sage, line 117: Same as above

src/components/tabs/tabs.d.ts
```diff
@@ -7,6 +7,7 @@ export interface TabsProps {
   selectedTabId: string;
   children: React.ReactNode[] | object;
   align: 'left' | 'right';
+  history: History;
```
sjohnson-sage, line 10: This should be marked as optional

sjohnson-sage (commented):
Just some minor changes to make 👍🏻

edleeks87 (commented):
Nothing more to add that Sam hasn't already raised, happy to approve once they're addressed 👍

src/components/tabs/tabs.component.js
```diff
@@ -41,6 +42,13 @@ const Tabs = ({
   const [tabsWarnings, setTabsWarnings] = useState({});
   const [tabsInfos, setTabsInfos] = useState({});
   const _window = Browser.getWindow();
+  const isExternalHistory = !!history;
+
+  if (isExternalHistory) {
+    Object.defineProperty(_window, "history", {
```
jamime, line 48: We shouldn't be defining things on the window.

jamime (changes requested):
We should be using the `history` API instead of overriding `window.history`. I think it's acceptable to to use `history@5` as a dependency of our project so we can remove the `window.history.pushState`. However, it should still be a prop so users can pass their own `history` instance we should use the `history@5` API.

> We need to ensure that we can use react-router@5 with history@4 and react-router@6 with history@5.

This doesn't have to be done in the component as long as the user can achieve the desired result e.g. by writing a wrapper around `history@4` so it functions the same as `history@5`.

[`history@4` documentation](https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/docs/Navigation.md)

[`history@5` documentation](https://github.com/ReactTraining/history/blob/master/docs/navigation.md)

edleeks87 (approved):

nicktitchmarsh (dismissed):

edleeks87 (dismissed):

src/components/tabs/tabs.component.js
```diff
@@ -114,7 +124,12 @@ const Tabs = ({
     (tabid) => {
       if (setLocation && !sidebarContext) {
         const url = `${_window.location.origin}${_window.location.pathname}#${tabid}`;
-        _window.history.replaceState(null, "change-tab", url);
+
+        if (isExternalHistory) {
```
jamime, line 128: This doesn't look correct, the method shouldn't change based on `isExternalHistory`.

I think we need an additional prop that defaults to `replace`, the existing behaviour.
```
navigatorMethod: PropTypes.oneOf(['push', 'replace']);
```
then we would call
```js
historyAPI[navigatorMethod](url);
```

docs/usage-with-routing.stories.mdx
```diff
@@ -44,6 +45,22 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
...
+  const history = useHistory();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      history.push(tabId);
+    },
+    [history]
+  );
+
+  return (
+    <Tabs history={history} onTabChange={handleTabChange}>
```
jamime, line 58: this will cause `history.push` to be called twice based on the code in `tabs.component.js`, I think you also need to set `setLocation=false`.

docs/usage-with-routing.stories.mdx
```diff
@@ -98,6 +131,23 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
...
+  const navigate = useNavigate();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      navigate(tabId);
+    },
+    [history]
+  );
+
+  return (
+    <Tabs history={history} onTabChange={handleTabChange}>
```
jamime, line 145:
```suggestion
    <Tabs history={history} onTabChange={handleTabChange} setLocation={false}>
```

jamime (changes requested):
> There is no code in Carbon to implement a specific routing library, we aim to let the user decide which library to use.

This is no longer true in the `usage-with-routing.stories.mdx`.

jamime (changes requested):
I've left some comments, I think we need a few possible combinations

- default history push
- default history replace
- custom history push
- custom history replace
- custom history onTabChange

I've created a sandbox to show these proposed changes

https://codesandbox.io/s/3865-tabs-tip5c

edleeks87 (approved):

src/components/tabs/tabs.component.js
```diff
@@ -30,11 +29,11 @@ const Tabs = ({
   selectedTabId,
   renderHiddenTabs = true,
   position = "top",
-  setLocation = true,
   extendedLine = true,
   size,
   borders = "off",
   variant = "default",
+  history,
```
jamime, line 36:
```suggestion
```

package.json
```diff
@@ -111,6 +111,7 @@
     "i18n-js": "http://github.com/fnando/i18n-js/archive/v3.0.0.rc12.tar.gz",
     "jest": "^26.6.3",
     "jest-styled-components": "^6.3.1",
+    "history": "^5.0.0",
```
jamime, line 114:
```suggestion
```

jamime (changes requested):
Just some small leftovers from previous implementation.

src/components/tabs/tabs.component.js
edleeks87: 🤦

src/components/tabs/tabs.component.js
```diff
@@ -46,41 +44,19 @@ const Tabs = ({
   const [tabsErrors, setTabsErrors] = useState({});
   const [tabsWarnings, setTabsWarnings] = useState({});
   const [tabsInfos, setTabsInfos] = useState({});
-  const _window = Browser.getWindow();
-  const isInSidebar = !!(sidebarContext && sidebarContext.isInSidebar);
+  const isInSidebar = !!sidebarContext?.isInSidebar;
 
   useLayoutEffect(() => {
     if (firstRender.current) {
```
sjohnson-sage, line 50: This check using `firstRender` could be replaced by an empty dependency array so it only runs on the first render.

Don't have to change it here because it's not part of your changes, but if you want to tidy that up it's up to you

docs/usage-with-routing.stories.mdx
```diff
@@ -44,6 +45,18 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
...
+const CarbonTabsWithReactRouter = ({ children }) => {
+  const history = useHistory();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      history.push(tabId);
+    },
+    [history]
+  );
+
+  return <Tabs onTabChange={handleTabChange}>{children}</Tabs>;
```
sjohnson-sage, line 57: We need a `selectedTabId` on this as well don't we?

docs/usage-with-routing.stories.mdx
```diff
@@ -98,6 +127,18 @@ const CarbonLinkWithReactRouter = ({ to, children }) => {
...
+const CarbonTabsWithReactRouter = ({ children }) => {
+  const navigate = useNavigate();
+  const handleTabChange = useCallback(
+    (tabId) => {
+      navigate(tabId);
+    },
+    [navigate]
+  );
+
+  return <Tabs onTabChange={handleTabChange}>{children}</Tabs>;
```
sjohnson-sage, line 139: Same as above

src/components/tabs/tabs.stories.mdx
```diff
@@ -1588,6 +1589,70 @@ CSS string.
...
+implementation using `createHashHistory` from `history@v5.0.0`.
+
+```jsx
+  import React, { useState, useEffect, useMemo } from "react";
+  import { Tabs, Tab } from "carbon-react/lib/components/tabs";
+  import { createHashHistory } from "history";
+
+  const TabsWithExternalHashHistory = ({ history, children, ...rest }) => {
+    const { pathname } = history?.location;
+    const tabIds = useMemo(() => ["tab-1", "tab-2", "tab-3"], []);
```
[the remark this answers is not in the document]
edleeks87: Yeah agreed, it was the linter complaining about it being a dependency in the useEffect that made me useMemo, but I'll just tidy the implementation up and remove it

src/components/tabs/tabs.component.js
```diff
@@ -402,6 +364,8 @@ Tabs.propTypes = {
     "no right side",
     "no sides",
   ]),
+  /** Navigator method */
+  navigatorMethod: PropTypes.oneOf(["push", "replace"]),
```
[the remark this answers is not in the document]
edleeks87: Yeah again not sure how i missed that 🤦

src/components/tabs/tabs.component.js
edleeks87: yeah good point I'll update it now the dependencies have been reduced

docs/usage-with-routing.stories.mdx
edleeks87: Yep, I'm on fire 🔥  with this Pr

src/components/tabs/tabs.component.js
sjohnson-sage: The dependencies are irrelevant because we only want this stuff to run on first render, they're only there to trigger re-runs of the function when the dependencies change

You might have to disable an eslint rule to have the empty array but it's the right thing to do here!

edleeks87: Yeah sorry I meant they were reduced down from what they were previously so it's worth updating the layoutEffect further rather than leaving it as is for now etc

docs/usage-with-routing.stories.mdx
```diff
@@ -74,9 +103,11 @@ An example of how to use the Carbon `Button` and `Link` components as links with
 Note the use of the `useNavigate` hook from `react-router-dom`.
 
 ```jsx
-import React, { useCallback } from "react";
+import React, { useCallback, useState } from "react";
+import { createBrowserHistory } from "history";
```
sjohnson-sage, line 107: The example below doesn't use `createBrowserHistory` anywhere, I think you can get rid of this line

src/components/tabs/tabs.component.js
```diff
@@ -30,57 +29,30 @@ const Tabs = ({
...
-        } else {
-          selectedTab = children.props.tabId;
-        }
-      }
-      setSelectedTabIdState(selectedTab);
-    }
-  }, [_window.location.hash, children, firstRender, selectedTabId]);
+    const selectedTab =
+      selectedTabId ||
+      React.Children.toArray(children).filter((child) => child)[0].props.tabId;
```
sjohnson-sage, line 51: What's the point of the `.filter((child) => child)` bit here? This could just be `React.Children.toArray(children)[0].props.tabId;` I think

src/components/tabs/tabs.stories.mdx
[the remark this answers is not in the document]
sjohnson-sage: This is still here btw

src/components/tabs/tabs.stories.mdx
```diff
@@ -1588,6 +1589,72 @@ CSS string.
   </Story>
 </Preview>
 
+### Integrating with an external history
+It is possible to integrate the `Tabs` component with an external `history` to manipulte the location when a given `Tab`
```
sjohnson-sage, line 1593: `manipulate`

src/components/tabs/tabs.component.js
edleeks87: Yeah I added it to account for conditional rendering of the `Tab` `children` but I think you're right and it's overkill so I'll update it

sjohnson-sage (dismissed):

jamime (dismissed):

sjohnson-sage (dismissed):

jamime (commented):
LGTM please squash commits before merging.

sjohnson-sage (approved):

DlgSHi (approved):

jamime (approved):

Issue #3830, opened 2021-03-16
Error when typing in search box within MenuItem


jlongridge:
### Current behaviour
When typing in a search box within a menu item an error occurs:
```
TypeError
Cannot read property ‘toLowerCase’ of undefined
```
https://codesandbox.io/s/charming-matsumoto-gxyz9?file=/src/index.js

### Expected behaviour
No Errors when typing.

### Reproducible example
Use the following codesandbox and type in the search box.
https://codesandbox.io/s/charming-matsumoto-gxyz9?file=/src/index.js

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 68.3.1
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

sjohnson-sage:
Hi @jlongridge thanks for raising, we will put it on our backlog and get it fixed

sjohnson-sage:
FE-3966

PR #3868, opened 2021-03-31
fix(search): stop event propagation for characters and numbers

sjohnson-sage:
Fixes: #3830

### Proposed behaviour
Stop event propagation when typing characters or numbers into the `Search` input, these should not propagate and affect keyboard nav etc of components further up the tree

### Current behaviour
Error when typing into `Search` when it's inside of a submenu because the event bubbles up to the keyboard nav in `Submenu` when it shouldn't

### Checklist

### Testing instructions
Use the sandbox from the attached Github issue and check the console on http://localhost:9001/?path=/story/design-system-menu--submenu-with-search

### DIFF
--- a/src/__experimental__/components/search/search.spec.js
+++ b/src/__experimental__/components/search/search.spec.js
@@ -326,9 +326,13 @@ describe("Search", () => {
     });
 
     it("passes other event handlers down to the input", () => {
-      const keyDownParams = { target: { selectionStart: 1, selectionEnd: 2 } };
+      const keyDownParams = {
+        which: 65,
+        key: "a",
+        target: { selectionStart: 1, selectionEnd: 2 },
+      };
       const input = wrapper.find("input");
-      input.simulate("keydown", { keyDownParams });
+      input.simulate("keydown", keyDownParams);
       expect(onKeyDown).toHaveBeenCalled();
     });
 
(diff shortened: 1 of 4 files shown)


src/__experimental__/components/search/search.spec.js
```diff
@@ -472,4 +476,48 @@ describe("Search", () => {
...
   });
+
+  describe("when typing into the input", () => {
+    const stopPropagationFn = jest.fn();
+
+    afterEach(() => {
+      jest.clearAllMocks();
+    });
+
+    it("should stop propagation of the event for character keys", () => {
```
dawidzarzycki, line 487: I think that we can create a `it.each` for these examples

sjohnson-sage: I'm not sure there's any need for that here really, we only have 2 tests making the same assertion, the third test is asserting the opposite

dawidzarzycki: You are right! I thought 3 of them are about the same :)
Ignore my comment then

dawidzarzycki (dismissed):

edleeks87 (approved):

edleeks87 (dismissed):

dawidzarzycki (approved):

edleeks87 (approved):

KatarzynaQA (approved):

PR #3873, opened 2021-04-06
fix(filterable-select): check event has a data element attribute before checking the value FE-3802

### LINKED ISSUE #3754: [FilterableSelect] Error thrown on mouse down
### Current behaviour
When clicking on the select component it intermittently throws an error.

### Expected behaviour
It should not throw an error

### Reproducible example
https://codesandbox.io/s/cocky-napier-jmfe6?file=/src/index.js

### Suggested solution(s)
The component does not check for the existence of the `data-element` attribute on the target element before inspecting its value.
See here: https://github.com/Sage/carbon/blob/c18255b6b8c62681bbf717eab9871968bef35af6/src/components/select/filterable-select/filterable-select.component.js#L389

### Your environment
| carbon-react     | 66.11.1
| react-scripts    | 3.4.4
| React            | 16.8.3
| Browser          | Chrome
| npm              | 6.14.4
| Operating System | MacOS

edleeks87:
fix #3754

### Proposed behaviour

Adds a guard to check that `event.target.attributes["data-element"]` is not undefined before
checking if the value is `"input"`

### Current behaviour

Throws `Uncaught TypeError: Cannot read property 'value' of undefined` when input icon is clicked

### Checklist

### Testing instructions

`design-system-select-filterable--default-story` click input icon when list is open, no error should be thrown

### DIFF
--- a/src/components/select/filterable-select/filterable-select.component.js
+++ b/src/components/select/filterable-select/filterable-select.component.js
@@ -388,7 +388,10 @@ const FilterableSelect = React.forwardRef(
     function handleTextboxMouseDown(event) {
       isMouseDownReported.current = true;
 
-      if (event.target.attributes["data-element"].value === "input") {
+      if (
+        event.target.attributes["data-element"] &&
+        event.target.attributes["data-element"].value === "input"
+      ) {
         isMouseDownOnInput.current = true;
       }
     }


src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -388,7 +388,10 @@ const FilterableSelect = React.forwardRef(
     function handleTextboxMouseDown(event) {
       isMouseDownReported.current = true;
 
-      if (event.target.attributes["data-element"].value === "input") {
+      if (
+        event.target.attributes["data-element"] &&
+        event.target.attributes["data-element"].value === "input"
```
mkrds, line 393: FYI, this could be `event.target.attributes["data-element"]?.value`

mkrds (approved):

src/components/select/filterable-select/filterable-select.component.js
edleeks87: Ah good point for some reason I thought optional chaining didn't work

dawidzarzycki (approved):

KatarzynaQA (approved):

DipperTheDan (approved):

PR #3894, opened 2021-04-12
feat(split-button): surface styled system margin props FE-3559

edleeks87:
### Proposed behaviour

Surfaces `styled-system/margin` props to allow consumers to adjust spacing of `SplitButton`
Surfaces `styled-system/margin` props to allow consumers to adjust spacing of `MultiActionButton`

### Current behaviour

No margin props surfaced on either component

### Checklist

### Additional context

`d.ts` files have been left intentionally as there is an active PR to address all of them in carbon

### Testing instructions

https://codesandbox.io/s/carbon-quickstart-forked-n85nz

### DIFF
--- a/src/components/multi-action-button/multi-action-button.component.js
+++ b/src/components/multi-action-button/multi-action-button.component.js
@@ -71,6 +71,12 @@ class MultiActionButton extends SplitButton {
 }
 
 MultiActionButton.propTypes = {
+  /** The additional button to display. */
+  children: PropTypes.node.isRequired,
+
+  /** Defines an Icon position within the button: "before" | "after" */
+  iconPosition: PropTypes.oneOf(["before", "after"]),
+
   /**
    * Customizes the appearance, can be set to 'primary', 'secondary' or 'transparent'.
    */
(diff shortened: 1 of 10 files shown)


src/components/multi-action-button/multi-action-button.component.js
```diff
@@ -71,6 +71,12 @@ class MultiActionButton extends SplitButton {
 }
 
 MultiActionButton.propTypes = {
+  /** The additional button to display. */
```
sjohnson-sage, line 74: This should have the margin props in the `propTypes` here, see https://github.com/Sage/carbon/blob/master/src/components/icon-button/icon-button.component.js#L42 for an example

sjohnson-sage: and what you've done in split button

sjohnson-sage: Also the margin props aren't actually getting passed to `StyledMultiActionButton` anywhere

sjohnson-sage (commented):
Needs some changes to `MultiActionButton` mate

src/components/multi-action-button/multi-action-button.component.js
edleeks87: The prop types come from `SplitButton` which MultiActionButton extends, if they are passed to the `StyledMultiActionButton` then they are set twice. For now I think this is the right approach and when they are refactored to remove the extension we can then spread them on to the root

edleeks87:
Closed as we need to refactor `MultiActionButton` to remove the extension of `SplitButton`

PR #3902, opened 2021-04-12
refactor(flat-table): update styled system implementation

mmaciorowski-sage:
BREAKING CHANGE: FlatTable uses margin. FlatTableCell FlatTableHeader FlatTableRowHeader use padding

### Proposed behaviour
`@styled-system/space` margin only (remove padding - breaking change).
Filter styled system margin props.

### Current behaviour
Uses no filtered `propTypes.space`.

### Checklist

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-uc7mp

### DIFF
--- a/src/components/flat-table/flat-table-cell/flat-table-cell.style.js
+++ b/src/components/flat-table/flat-table-cell/flat-table-cell.style.js
@@ -1,5 +1,5 @@
 import styled, { css } from "styled-components";
-import { space } from "styled-system";
+import { padding } from "styled-system";
 
 import baseTheme from "../../../style/themes/base";
 
@@ -14,9 +14,9 @@ const StyledFlatTableCell = styled.td`
     white-space: nowrap;
     padding: 0;
 
-    > div {
+    &&& > div {
       box-sizing: border-box;
-      ${space}
+      ${padding}
     }
 
     &:first-of-type {
--- a/src/components/flat-table/flat-table-header/flat-table-header.style.js
+++ b/src/components/flat-table/flat-table-header/flat-table-header.style.js
@@ -1,5 +1,5 @@
 import styled, { css } from "styled-components";
-import { space } from "styled-system";
+import { padding } from "styled-system";
 
 import baseTheme from "../../../style/themes/base";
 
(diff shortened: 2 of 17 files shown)


edleeks87:
I think the description of the `BREAKING CHANGE` is wrong as some of the components have been updated to use the `padding` not `margin` interface etc

src/components/flat-table/flat-table-cell/flat-table-cell.style.js
```diff
@@ -16,7 +16,7 @@ const StyledFlatTableCell = styled.td`
 
     > div {
```
edleeks87, line 17: I don't think these will override the defaults applied for padding left and right (see screenshot). The unit tests pass as the styling is applied here https://github.com/Sage/carbon/blob/c547f4efdd035bb2dd0013b94c21bb38df8237c7/src/components/flat-table/flat-table.style.js#L56  I think you will need to raise the specificity like it is currently in `master` like so https://github.com/Sage/carbon/blob/6dba4e9ae48618e7d6ea9fbd6e9ffe6323aea730/src/components/flat-table/flat-table-cell/flat-table-cell.style.js#L29

image

src/components/flat-table/flat-table-header/flat-table-header.style.js
```diff
@@ -30,7 +30,7 @@ const StyledFlatTableHeader = styled.th`
 
     > div {
```
edleeks87, line 31: see comment about raising specificity

edleeks87 (changes requested):
Just the commit message and the specificity issue, everything else looks good

mmaciorowski-sage:
> I think the description of the `BREAKING CHANGE` is wrong as some of the components have been updated to use the `padding` not `margin` interface etc

Good point. Changed.

src/components/flat-table/flat-table-cell/flat-table-cell.style.js
mmaciorowski-sage: Good point, changed. Thanks!

edleeks87 (approved):

dawidzarzycki (approved):

DlgSHi (approved):

PR #3933, opened 2021-04-20
refactor(menu): deprecate submenu block

sjohnson-sage:
BREAKING CHANGE: The SubmenuBlock component has been deprecated.
It is no longer required to build a submenu.
If you need a scrollable submenu, please use ScrollableBlock

### Proposed behaviour
Remove the `SubmenuBlock` component and update any tests etc to maintain 100% coverage

### Current behaviour
`SubmenuBlock` component exists in Carbon

### Checklist

### Testing instructions
Give all the `Menu` stories a once over http://localhost:9001/?path=/docs/design-system-menu--default-story


src/components/menu/__internal__/submenu/submenu.component.js
```diff
@@ -43,7 +43,7 @@ const Submenu = React.forwardRef(
     const menuContextOpen = menuContext.openSubmenu;
 
     const formattedChildren = React.Children.map(children, (child) => {
-      if (child.type === ScrollableBlock) {
+      if (child && child.type === ScrollableBlock) {
```
dawidzarzycki, line 46: I guess this change not corresponding to the PR title and I guess it should be as a different commit at least in this case

sjohnson-sage: It can't be actually because removing the check for `SubmenuBlock` in `MenuItem` causes this to error, so the commit would be broken if I split it out

sjohnson-sage: More specifically removing the `React.Children.map` bit on line 45 means that null children can now (correctly) be passed into `Submenu` and this extra check is needed. It should have been there anyway

dawidzarzycki (approved):

mkrds (approved):

DlgSHi (approved):

PR #3944, opened 2021-04-26
fix(definition-list): fix dl focused elements outline cut by "overflow: hidden" - FE-3211

mmaciorowski-sage:
### Proposed behaviour
Remove `overflow: hidden` from `definition-list` component.
Zrzut ekranu 2021-04-26 o 10 23 03

### Current behaviour
Zrzut ekranu 2021-04-26 o 10 23 11

### Checklist

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-30grt


src/components/definition-list/dl.component.js
```diff
@@ -10,6 +10,10 @@ const Dl = ({
   w = 50,
   dtTextAlign = "right",
   ddTextAlign = "left",
+  pl = "3px",
```
edleeks87, line 13: Can you add these default values to the tests please https://github.com/Sage/carbon/blob/4e8087fc78051e8b2bd011d9283be3143adf35d4/src/components/definition-list/definition-list.spec.js#L34

mmaciorowski-sage: good point, thanks, added

src/components/definition-list/dl.component.js
```diff
@@ -57,7 +61,15 @@ const Dl = ({
   };
 
   return (
-    <StyledDl w={w} data-component="dl" {...props}>
+    <StyledDl
```
dawidzarzycki, line 64: you can set 
```js
px={'3px'}
mx={'-3px'}
```

dawidzarzycki:
@mmaciorowski-sage Hey!
Maybe better solution for this will be to remove `overflow: hidden` property from `dl` component instead of adding paddings and margins

src/components/definition-list/dl.component.js
mmaciorowski-sage: good point, thanks, changed

mmaciorowski-sage:
> @mmaciorowski-sage Hey!
> Maybe better solution for this will be to remove `overflow: hidden` property from `dl` component instead of adding paddings and margins

I talked to Daniel and he told me that it was because we wanted the list to fit in to the size of the `Tile` for example. So if someone had a definition-list with 50 lines, it wouldn’t overflow the size of the `Tile`.

dawidzarzycki (approved):

edleeks87 (approved):

mmaciorowski-sage:
> @mmaciorowski-sage Hey!
> Maybe better solution for this will be to remove `overflow: hidden` property from `dl` component instead of adding paddings and margins

FYI At first I assumed it was necessary, but it turned out that it could be removed.

ZhuoyuJin (approved):
Tested in Edge, Chrome, Firefox, and Safari. No issues found, testing passed.

PR #3952, opened 2021-04-27
fix(search): fix outline appearing onClick around the X icon - FE-3999

### LINKED ISSUE #3857: Search "X" icon outline appears onClick
### Current behaviour
When clicking the `X` icon to clear the input in the Search component, a golden outline quickly appears.

https://user-images.githubusercontent.com/56251247/112610166-4c911a00-8e14-11eb-9e76-0522b7ef5785.mov

### Expected behaviour
When clicking the "X" icon to clear the input, no outline should appear around the "X" icon.

### Suggested solution(s)
Will probably need to use a flag to control when the icon can have a focus outline.
Solution will also need to make sure that when using keyboard navigation that the "X" can still appear focused.

### Your environment

| carbon-react     | 68.5.1
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

DipperTheDan:
Fixes a styling bug where the X icon appears a golden outline onClick

fixes #3857

### Proposed behaviour
Focus outline should only appear around 'X' icon when icon is navigated to via the keyboard.

### Current behaviour
Focus outline appears onClick.

### Checklist

<del> - [ ] Screenshots are included in the PR if useful </del>
<del> - [ ] Storybook added or updated if required </del>
<del> - [ ] Typescript `d.ts` file added or updated if required </del>

### Additional context
N/A

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-dojtr?file=/src/index.js

### DIFF
--- a/src/__experimental__/components/search/search.component.js
+++ b/src/__experimental__/components/search/search.component.js
@@ -123,6 +123,10 @@ const Search = ({
     }
   };
 
+  const handleMouseDown = (ev) => {
+    ev.preventDefault();
+  };
+
   const handleBlur = () => {
     setIsFocused(false);
   };
@@ -162,6 +166,7 @@ const Search = ({
         inputIcon={iconType}
         iconTabIndex={iconTabIndex}
         iconOnClick={handleIconClick}
+        iconOnMouseDown={handleMouseDown}
         aria-label={ariaLabel}
         onFocus={onFocus}
         onClick={onClick}
@@ -202,7 +207,7 @@ Search.propTypes = {
   onKeyDown: PropTypes.func,
   /** Prop boolean to state whether the `search` icon renders */
   searchButton: PropTypes.bool,
-  /** Prop for specifing an input width length.
+  /** Prop for specifying an input width length.
    * Leaving the `searchWidth` prop with no value will default the width to '100%' */
   searchWidth: PropTypes.string,
   /** Prop for `onFocus` events */
--- a/src/__experimental__/components/search/search.d.ts
+++ b/src/__experimental__/components/search/search.d.ts
@@ -16,7 +16,7 @@ export interface SearchProps extends MarginSpacingProps {
   onKeyDown?: (ev: React.SyntheticEvent) => void;
   /** Prop boolean to state whether the `search` icon renders */
   searchButton?: boolean;
-  /** Prop for specifing an input width length.
+  /** Prop for specifying an input width length.
    * Leaving the `searchWidth` prop with no value will default the width to '100%'
    */
   searchWidth?: string;
--- a/src/__experimental__/components/search/search.spec.js
+++ b/src/__experimental__/components/search/search.spec.js
@@ -354,6 +354,21 @@ describe("Search", () => {
         });
       });
 
+      it("calls preventDefault", () => {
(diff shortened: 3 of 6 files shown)


src/__experimental__/components/search/search.component.js
```diff
@@ -14,6 +14,7 @@ const Search = ({
   defaultValue,
   onChange,
   onClick,
+  onMouseDown,
```
edleeks87, line 17: this prop is never used, I'm not sure we need it tbh as we only want the mouseDown event handler to stop the focus behaviour

src/__experimental__/components/search/search.component.js
```diff
@@ -178,13 +184,16 @@ Search.propTypes = {
   /** Prop for `onClick` events.
    * `onClick` events are triggered when the `searchButton` is clicked */
   onClick: PropTypes.func,
+  /** Prop for `onMouseDown` events.
```
edleeks87, line 187: see comment above

src/__experimental__/components/search/search.d.ts
```diff
@@ -9,13 +9,17 @@ export interface SearchProps {
    * `onClick` events are triggered when the `searchButton` is clicked
    */
   onClick?: (ev: React.SyntheticEvent) => void;
+    /** Prop for `onMouseDown` events.
```
edleeks87, line 12: This is why your linter has failed fyi, I also think we don't need to surface this prop

src/__experimental__/components/search/search.component.js
DipperTheDan: It's gone on the next commit 👍🏻

src/__experimental__/components/search/search.component.js
DipperTheDan: Consider it gone.

src/__experimental__/components/search/search.d.ts
DipperTheDan: It's gone on the next commit.

edleeks87 (approved):

sjohnson-sage (approved):

sjohnson-sage (approved):

tempertemper:
I've had a quick look at this with @harpalsingh (who'll get back to you separately with approval/etc.). One thing I noticed as I poked around the generated code was that we're using a `<span>` rather than the `<button>` element. I'll raise something separately about getting this switched as we've gone to quite a lot of effort to replicate the functionality of a `<button>` (keyboard behaviour, tab indexed, etc.) but a button would give us all of that (and more) for free (we're missing a `role="button"` that would be implicit with a `<button>` element, so that screen reader users would know what the element is and speech recognition users could interact with it as a button).

harpalsingh:
We are happy on the UX and A11y, but just a note that @tempertemper will be looking into possible updates to the element tag, whether it might need to be <button> but that will be raised separately.

ZhuoyuJin (dismissed):
Tested in Chrome, Edge, Firefox, and Safari. No issues found, testing passed

ZhuoyuJin (approved):

edleeks87 (approved):

Issue #3728, opened 2021-02-20
Flat table and stickyHeader


gusch:
### Current behaviour
If i have two rows in the FlatTableHead of a FlatTable with hasStickyHead and i scroll, only one of the row is sticky

### Expected behaviour
All the rows of the header are sticky

### Reproducible example
https://codesandbox.io/s/carbon-date-locale-forked-155rj?file=/src/App.js

### Your environment
| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 66.0.3 |

sjohnson-sage:
Hi @gusch I believe the intended use case for this component would be to only have one row in the header - can you tell me more about your use case please? Why do you need multiple header rows?

gusch:
Hi @sjohnson-sage We did a custom Table from the Flat Table where a row with filters is automatically added in the header. So we have two rows in the header. When we scroll, we would like to have the two rows always visible (sticky).

ianforster-sage:
@thehiggster is this part of the new features that you have in mind for FlatTable?

dawidzarzycki:
@thehiggster we still need clarification on this

ianforster-sage:
@thehiggster when you have a moment please could you update this

thehiggster:
I don’t get notifications from github (or I’d get 1000s) so if I’m behind on anything feel free to nudge on Slack.

Yes, we need this for Accounting Reporting, which can have many rows in the header and may be sticky.

mkrds:
FE-4023

PR #3961, opened 2021-04-29
fix(flat-table): allow multiple header rows to be sticky

sjohnson-sage:
Fixes: #3728

### Proposed behaviour
Allow any number of header rows to be sticky in `FlatTable`

### Current behaviour
Currently only one header row can be sticky.

### Checklist

### Testing instructions
Add or remove header rows on a `FlatTable` in this sandbox: https://codesandbox.io/s/eloquent-smoke-pcrd5?file=/src/index.js


src/components/flat-table/flat-table-row/flat-table-row.style.js
```diff
@@ -55,6 +55,14 @@ const StyledFlatTableRow = styled.tr`
   table-layout: fixed;
   width: auto;
 
+  ${({ stickyOffset }) =>
+    stickyOffset > 0 &&
+    css`
+      th {
```
edleeks87, line 61: I think we need to scope this to only apply within the `FlatTableHeader` somehow as when we use it with `FlatTableRowHeader`s it will mess with the positioning of the columns as they render as a `th`, or maybe use `as` to set them to be `td` when no in the header 🤷‍♂️ 

https://codesandbox.io/s/goofy-noether-j1t09

edleeks87 (commented):
Just the one comment mate, implementation looks fine to me just scoping of the style needs tweaking

src/components/flat-table/flat-table-row/flat-table-row.style.js
sjohnson-sage: Ah nice one, looks like you've found another existing bug where the sticky styles aren't applied to `FlatTableRowHeader`s within the `FlatTableHead`! 

I'll fix that in this PR as well if it isn't too much extra to add in here

edleeks87 (approved):

DipperTheDan (approved):

DipperTheDan (approved):

harpalsingh:
@sjohnson-sage looks good to me, was going to note the issue of the border color here in sticky header

Screenshot 2021-05-05 at 09 31 51

Both the issue of color and weight, however this is just to note that due to a new ticket I am currently working on, that will be resolved in that tickets request, so its a note that I am aware this would be a ux issue in visual style, but upcoming work will resolve that.

DlgSHi (approved):

PR #3977, opened 2021-04-30
refactor(fieldset): update styled system implementation - FE-3512

mmaciorowski-sage:
### Proposed behaviour
@styled-system/space margin only on root element 
In order for this to work the props must only be spread on the root element only rather than passed down to the sub components.
Remove custom margin props

### Current behaviour
Custom margin props

### Checklist

### Testing instructions

Internal `Fieldset` is a part of `RadioButton`.
https://codesandbox.io/s/carbon-quickstart-forked-kkp18

### DIFF
--- a/src/__internal__/fieldset/fieldset.style.js
+++ b/src/__internal__/fieldset/fieldset.style.js
@@ -1,19 +1,15 @@
 import styled, { css } from "styled-components";
 import PropTypes from "prop-types";
+import { margin } from "styled-system";
 import BaseTheme from "../../style/themes/base";
 
 const StyledFieldset = styled.fieldset`
-  ${({ ml, mb, theme }) => css`
-    border: none;
-    margin: 0;
-    padding: 0;
-    min-width: 0;
-    min-inline-size: 0;
-
-    ${ml && `margin-left: ${ml};`}
-    ${mb && `margin-bottom: ${mb * theme.spacing}px;`}
-    ${({ styleOverride }) => styleOverride};
-  `}
+  ${margin}
+  border: none;
+  padding: 0;
+  min-width: 0;
+  min-inline-size: 0;
+  ${({ styleOverride }) => styleOverride};
 `;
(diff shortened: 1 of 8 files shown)


src/__internal__/fieldset/fieldset.style.js
```diff
@@ -1,17 +1,15 @@
 import styled, { css } from "styled-components";
 import PropTypes from "prop-types";
+import { margin } from "styled-system";
 import BaseTheme from "../../style/themes/base";
 
 const StyledFieldset = styled.fieldset`
-  ${({ ml, mb, theme }) => css`
+  ${() => css`
```
sjohnson-sage, line 7: You can just get rid of this function completely now that we're not using any of the parameters, i.e. 
```js
const StyledFieldset = styled.fieldset`
  ${margin}
  ..etc
`;
```

sjohnson-sage: and you can remove the `StyledFieldset.defaultProps` below as well because we don't need the theme anymore

sjohnson-sage (commented):
Looks good, just one minor comment!

src/__internal__/fieldset/fieldset.style.js
mmaciorowski-sage: Hi Sam, good point thanks. Changed, but I think `defaultProps` must be defined because of `margin`.

sjohnson-sage (approved):

src/__internal__/fieldset/fieldset.style.js
sjohnson-sage: Good point you're right, I forgot about the margin bit!

DipperTheDan (dismissed):

DlgSHi (approved):

DipperTheDan (approved):

DlgSHi (approved):

PR #3978, opened 2021-04-30
chore(eslint): replace tslint with eslint

sjohnson-sage:
### Proposed behaviour
Remove tslint and replace with eslint config. We currently need different parsers for the JavaScript files and the TypeScript files.

Fix any valid eslint errors in our ts files.

Run prettier over src (everything that isn't `.js` has not been formatted in one go until now)
 

### Current behaviour
Carbon uses tslint to lint the TypeScript files. Tslint has been deprecated.

### Checklist


mkrds (approved):

.eslintrc
```diff
@@ -115,5 +116,37 @@
...
+        "plugin:react-hooks/recommended",
+        "plugin:jest/recommended",
+        "plugin:jest-dom/recommended"
+      ],
+      "settings": {
+        "react": {
+          "version": "detect"
+        }
+      },
+      "rules": {
```
jamime, line 146: Why are we turning these rules off? We don't have any TS components yet so these shouldn't be failing?

sjohnson-sage: The `import/named` one because it errors for every import we have in `.d.ts` files - I believe we have actually implemented it incorrectly but converting Carbon to TypeScript will fix it

The second one because it errors for all our class components, if they don't have state - we can turn this back on when we've done all the functional refactor tickets.

I suppose for both we could disable the rules line by line where needed but I thought it wasn't worth it, and fully intend on turning these rules back on once the TypeScript conversion and class -> functional tickets are all done

jamime (dismissed):

mkrds (approved):

jamime (approved):

PR #3979, opened 2021-04-30
test(date): remove console warn in specs FE-3989

edleeks87:
### Proposed behaviour

Remove significant amount of console output not related to deprecated warnings or triggered in deprecated components etc when running tests. Does not touch any component or prop that has been marked for deprecation
### Current behaviour

Lots of console pollution when running tests
### Checklist

### Testing instructions

There has been a small update to the `SelectList` so a quick brush test that the `Popover` functionality still works


jamime (commented):
I think we should blacklist the errors we want to suppress instead of suppressing everything.

If enzyme/react errors at the moment the error won't be shown.

```js
const { error } = global.console;
    beforeAll(() => {
      const blacklist = [
        "Invariant Violation: <Note> inlineControl must be an instance of <ActionPopover>",
      ];
      global.console.error = (...args) => {
        const msg = args.join(" ");
        if (blacklist.includes(msg)) {
          return;
        }
        error(...args);
      };
    });

    afterAll(() => {
      global.console.error = error;
    });
```

This is a bit cumbersome so I think it would be better as a utility, we would assert that the error was thrown, and swallow the output. All other unexpected errors would still be logged.

```js
it('example', () => {
const restore = expectErrors(['Invariant Violation: <Note> inlineControl must be an instance of <ActionPopover>']);
render();
restore();
})
```

What do you think?

edleeks87:
@jamime great suggestion, I'll look to get this added and update the PR to use it asap, thanks

mmaciorowski-sage (approved):

ianforster-sage (approved):

mmaciorowski-sage (approved):

sianford (approved):

sianford:
Tested in Safari, Chrome, Firefox and Edge. No console warnings, all regression testing in Storybook has passed. Testing passed ✅

Issue #3755, opened 2021-02-26
Filterable select to allow controlled filter text


stripeyjumper:
## Feature request
If external components could pass `filterText` and `onFilterChanged` props to `FilterableSelect`, it would enable server-side filtering by allowing the current value of `filterText` to be passed to an API query.  

### Current behaviour
`filterText` and `setFilterText` are managed as internal state of the `FilterableSelect` component

### Desired behaviour
`filterText` is optionally controlled by passing in `filterText` and `onFilterTextChanged` props.  If the props are not provided, it keeps the current behaviour.

### Describe alternatives you've considered
- A different dropdown control (not an option for our project).
- Tracking the filter value by inspecting the DOM input element.
- Writing a `FilterableSelect` with the required behaviour.

### Additional context
Components like [react-select](https://react-select.com/async) expose the input value through their props in a similar way.

### Your environment
| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 66.11.1
| react-scripts    | 3.4.4
| React            | 16.8.3
| Browser          | Chrome
| npm              | 6.14.4
| Operating System | MacOS

sjohnson-sage:
FE-3801

sjohnson-sage:
Hi @stripeyjumper - thanks for raising this but we've decided not to implement this as it'll add more complexity to an already overly complex component

We recommended just using a ref to get access to the filter text in the component if you wish to: https://codesandbox.io/s/sweet-fermi-ofdjm?file=/src/index.js

stripeyjumper:
Hi @sjohnson-sage, thanks for the workaround anyway.

PR #3981, opened 2021-05-04
feat(filterable-select): add filterText and onFilterChanged props

sjohnson-sage:
Fixes: #3755

### Proposed behaviour
Add `filterText` and `onFilterChanged` props to `FilterableSelect` to allow controlled usage.

### Current behaviour
These props were not available.

### Checklist


grabkowski (commented):
we don't need this functionality as all could be done using the `value` and `input` event on the `ref`

PR #4014, opened 2021-05-12
build: upgrade to node 14 lts

jamime:
### Proposed behaviour
Change GitHub CI so it uses a compatible version of node `>=14.16.0`(defined in package.json#engines).

I increased the version range to include `14.16.0`, this is latest version available for the cypress docker image. `14.16.1` was rather arbitrary, it was the latest `lts` version at the time the engine field was updated.

I've restricted the range to keep it on the current LTS (`14.x`).

We don't force our consumers to stay on `14`, but we don't officially support `15`.

We no longer need to transpile from `12` to `14` so I removed it from the `babelrc.json`.

I've also removed the other text editor recommendations. Developers are still free to choose their editor, but we can only provide support for VS Code.

### Current behaviour
GitHub CI is using `12.x`
### Checklist

### Additional context
Blocking #3799

⚠️ Developers will no longer be able to start storybook if they have the wrong node version installed.

image

To upgrade developers can use `nvm install lts/fermium`

### Testing instructions
Ensure that GitHub checks pass.

### DIFF
--- a/.github/workflows/chromatic-pr.yml
+++ b/.github/workflows/chromatic-pr.yml
@@ -15,7 +15,7 @@ jobs:
         fetch-depth: 0
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '>=14.16.0 14'
     - run: npm ci
     - uses: chromaui/action@v1
       with:
--- a/.github/workflows/cypress-push.yml
+++ b/.github/workflows/cypress-push.yml
@@ -23,7 +23,7 @@ jobs:
             11, 12, 13, 14, 15,
             16, 17, 18, 19, 20
             ]
-    container: cypress/browsers:node12.18.3-chrome89-ff86
+    container: node14.16.0-chrome89-ff86
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -87,7 +87,7 @@ jobs:
         fetch-depth: 0
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '>=14.16.0 14'
     - run: npm ci
     - uses: chromaui/action@v1
       with:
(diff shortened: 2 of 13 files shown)


sjohnson-sage (approved):

DipperTheDan (dismissed):

DipperTheDan (dismissed):

.github/workflows/cypress-push.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             11, 12, 13, 14, 15,
             16, 17, 18, 19, 20
             ]
-    container: cypress/browsers:node12.18.3-chrome89-ff86
+    container: node14.16.0-chrome89-ff86
```
DlgSHi, line 26:
```suggestion
    container: cypress/browsers:node14.16.0-chrome89-ff86
```

.github/workflows/chromatic-pr.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         fetch-depth: 0
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '>=14.16.0 14'
```
DlgSHi, line 18: why we are passing `14.16.0` and `14` as a separate two values?

sjohnson-sage: `I've restricted the range to keep it on the current LTS (14.x).`

sjohnson-sage: It's saying node version 14.16.0 and above, but you can't go to v15 or above

.github/workflows/cypress-push.yml
jamime: I've just updated, thanks.

.github/workflows/chromatic-pr.yml
jamime: This means any version of 14 greater than 14.16.0 but not 15.

DlgSHi: Understood :)

sjohnson-sage (approved):

DipperTheDan (approved):

DlgSHi (approved):

PR #4030, opened 2021-05-17
refactor: change propTypes to be more specific

grabkowski:
### Proposed behaviour
PropTypes in various components are defined as strings, these props should have more specific types.

### Checklist

### DIFF
--- a/src/__experimental__/components/switch/switch.component.js
+++ b/src/__experimental__/components/switch/switch.component.js
@@ -98,7 +98,7 @@ Switch.propTypes = {
   /** The content of the label for the input */
   label: PropTypes.string,
   /** Sets label alignment - accepted values: 'left' (default), 'right' */
-  labelAlign: PropTypes.string,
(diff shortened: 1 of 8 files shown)


sjohnson-sage (approved):

src/__experimental__/components/switch/switch.component.js
```diff
@@ -135,7 +135,7 @@ Switch.propTypes = {
    * Set the size of the Switch to 'small' (16x16 - default) or 'large' (24x24).
    * No effect when using Classic theme.
```
DipperTheDan, line 136: We don't have `classic theme` anymore. Is it worth changing this comment?

grabkowski: I think that could be addressed later in a separate commit, as the Classic Theme is referenced at least five times in our code

DipperTheDan (approved):

Issue #4001, opened 2021-05-10
Icons do not allow hover state to be disabled


nicktitchmarsh:
### Current behaviour
Icons always have a hover state except when in disabled state which adds a grey background. 
![Screenshot 2021-05-10 at 10 11 07](https://user-images.githubusercontent.com/69150954/117636334-9ccffb80-b178-11eb-9dfa-0322565ef73a.png)

### Desired behaviour
Remove hover states without affecting the styling on the icon

### Suggested solution
We should allow icons to be static without any hover or tooltip functionality via a prop flag

### Describe alternatives you've considered
Attempted styling overrides using other components. Nothing worked

PR #4031, opened 2021-05-17
feat(icon): allow hover state to be disabled - FE-4080

mmaciorowski-sage:
### Proposed behaviour
Icons allow hover state to be disabled

### Current behaviour
Icons do not allow hover state to be disabled

### Checklist

### Additional context
Resolves: #4001

### DIFF
--- a/src/components/icon/icon.component.js
+++ b/src/components/icon/icon.component.js
@@ -22,6 +22,7 @@ const Icon = React.forwardRef(
       disabled,
       fontSize,
       iconColor,
+      hover,
       type,
       tooltipMessage,
       tooltipPosition,
@@ -67,6 +68,7 @@ const Icon = React.forwardRef(
       iconColor,
       tabIndex,
       type: iconType(),
+      hover,
       ...filterStyledSystemMarginProps(rest),
     };
 
@@ -147,6 +149,8 @@ Icon.propTypes = {
   disabled: PropTypes.bool,
   /** Aria label for accessibility purposes */
   ariaLabel: PropTypes.string,
+  /** Allows the hover state */
+  hover: PropTypes.bool,
   /** The message string to be displayed in the tooltip */
   tooltipMessage: PropTypes.string,
   /** The position to display the tooltip */
@@ -185,6 +189,7 @@ Icon.defaultProps = {
   bgSize: "small",
   fontSize: "small",
   disabled: false,
+  hover: true,
 };
 
 export default Icon;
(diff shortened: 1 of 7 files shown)


src/components/icon/icon.d.ts
```diff
@@ -24,6 +24,8 @@ export interface IconProps extends MarginProps {
   bg?: string;
   /** Sets the icon in the disabled state */
   disabled?: boolean;
+  /** Allows the hover state */
+  hover: boolean;
```
sjohnson-sage, line 28:
```suggestion
  hover?: boolean;
```

src/components/icon/icon.stories.mdx
```diff
@@ -154,6 +154,32 @@ See <LinkTo kind='Documentation/Colors' name="page">Colors</LinkTo> for more inf
   </Story>
 </Preview>
 
+### Without hover state
+
+<Preview>
+  <Story
+    name="without hover state"
+    parameters={{ info: { disable: true }, chromatic: { disable: true } }}
```
sjohnson-sage, line 162: You don't need either of these parameters on this story because they're both on the `<Meta />` tag at the top - that sets the parameters for every story.

Looks like some of the other stories in this file mistakenly have these as well!

mmaciorowski-sage: Good point, changed, thanks!

src/components/icon/icon.d.ts
mmaciorowski-sage: Good point, changed, thanks!

sjohnson-sage (approved):

src/components/icon/icon.component.js
```diff
@@ -147,6 +149,8 @@ Icon.propTypes = {
   disabled: PropTypes.bool,
   /** Aria label for accessibility purposes */
   ariaLabel: PropTypes.string,
+  /** Allows the hover state */
+  hover: PropTypes.bool,
```
dawidzarzycki, line 153: I'd rather go with `disableHover` name for this blocking hover mechanism 
We would have hover normally as we have right now and after someone will put `disableHover` for the component then it will be blocked. It will also follow convention that we have when we are disabling some stuff like
`disableEscKey` in `alert/confirm/dialog`
Personally I think that `disableHover` is more understandable for the end user

dawidzarzycki (approved):
According to our private chat we had I'll approve the code

jamime:
I've contacted Chromatic, the visual changes on the `+` icon was caused by `window.chrome` being undefined in Chromatic last time.

jamime:
Chromatic are investigating, they deployed approx 8 hours ago.

jamime (changes requested):
Blocking, pending discussion.

> Should we disable hover for all icons unless they have a tooltip?

PR #4032, opened 2021-05-18
feat(drawer): add sticky header and footer options

sjohnson-sage:
### Proposed behaviour
Add the option to have sticky headers and footers in `Drawer`. As well as having a non-sticky footer, which wasn't currently possible.

Sticky header: 
image

Non-sticky footer:
image

Sticky footer:
image

### Current behaviour
`Drawer` currently has no props for a sticky header/footer, or a non-sticky footer.

### Checklist

### Testing instructions
http://localhost:9001/?path=/story/design-system-drawer--with-sticky-header
http://localhost:9001/?path=/story/design-system-drawer--with-footer
http://localhost:9001/?path=/story/design-system-drawer--with-sticky-footer


src/__internal__/sticky-footer/sticky-footer.component.js
```diff
@@ -0,0 +1,65 @@
...
+  );
+};
+
+StickyFooter.propTypes = {
+  /** child elements */
+  children: PropTypes.node.isRequired,
+  /** Ref of the container the footer should be sticky on */
+  containerRef: PropTypes.any.isRequired,
+  /** Disable the sticky behaviour manually */
+  disableSticky: PropTypes.bool,
```
mmaciorowski-sage, line 62: If it's not a required prop, do we need to have a default prop for it?

sjohnson-sage: No we don't, this prop will only affect the component if it's set to `true`

src/components/drawer/drawer.component.js
```diff
@@ -2,39 +2,46 @@ import React, { useState, useCallback, useRef, useEffect } from "react";
...
-  defaultExpanded,
+  defaultExpanded = true,
   expanded,
   onChange,
   children,
-  expandedWidth,
+  expandedWidth = "40%",
   sidebar,
-  animationDuration,
+  animationDuration = "400ms",
```
mmaciorowski-sage, line 28: I wonder why you didn't use the default props. Is this the approach I should take as well?

src/__internal__/sticky-footer/sticky-footer.component.js
mmaciorowski-sage: I am used to the fact that if a prop is not required it must have a default value (eslint rule). But that's ok for me. Thanks

src/components/drawer/drawer.component.js
sjohnson-sage: Yeah this is a cleaner way of doing it 👍🏻 I think it's better having the defaults at the top of the component like this as well, rather than having to scroll down below it to see what the defaults are

sjohnson-sage: and using `defaultProps` on functional components is going to be deprecated in React at some point as well!

mmaciorowski-sage: Great, thanks a lot!

mmaciorowski-sage (approved):

src/__internal__/sticky-footer/sticky-footer.component.js
sjohnson-sage: Yeah we have the `"react/require-default-props"` eslint rule turned off at the moment, we have `undefined` props all over the place in Carbon!

Maybe something we could think about changing in the future though

src/__internal__/sticky-footer/sticky-footer.component.js
```diff
@@ -0,0 +1,65 @@
...
+    throttle(() => {
+      let fullyScrolled;
+      const content = containerRef.current;
+      const stickyOffset = footerRef.current.clientHeight / 2;
+
+      if (
+        content.scrollHeight - content.scrollTop - stickyOffset <=
+        content.clientHeight
+      ) {
+        fullyScrolled = true;
```
edleeks87, line 23: You could update this so it doesn't need any of the `if` statements and `setIsSticky` to the result of the condition.
```suggestion
     const fullyScrolled = content.scrollHeight - content.scrollTop - stickyOffset <= content.clientHeight;
     
     setIsSticky(fullyScrolled);
```
Not a big deal so if you prefer it the way you have I'm happy to leave it and resolve this conversation

sjohnson-sage: Yeah good spot, I laid it out like that because I had some other code in originally, but didn't simplify this when I removed the other stuff - I'll change it

edleeks87 (dismissed):

mmaciorowski-sage (approved):

mmaciorowski-sage (dismissed):

edleeks87 (approved):

mmaciorowski-sage (approved):

DlgSHi (approved):

edleeks87 (approved):

PR #4046, opened 2021-05-19
fix(duelling-picklist): add error if children do not have 2 picklist FE-4059 

edleeks87:
### Proposed behaviour

Fixes bug where it is possible to have less or more than two `Picklist`s.

Adds `FocusContext` to allow the correct `PicklistItem` or `PicklistGroup` button to be focused when
a given `Picklist` child is moved. If the item or group is not the last one it focuses the next in the given `Picklist`, if it is the last it focuses the first in the other `Picklist`.

### Current behaviour

Focus is moved to the element before the first `Picklist` when an item button is clicked or receives a keydown event

### Checklist

### Additional context

item focus demo

group focus demo

### Testing instructions

`design-system-duellingpicklist--grouped`
`design-system-duellingpicklist--default`

### DIFF
--- a/src/components/duelling-picklist/picklist-group/picklist-group.component.js
+++ b/src/components/duelling-picklist/picklist-group/picklist-group.component.js
@@ -1,4 +1,10 @@
-import React, { useCallback, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import PropTypes from "prop-types";
 import { CSSTransition, TransitionGroup } from "react-transition-group";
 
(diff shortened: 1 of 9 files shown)


nicktitchmarsh (dismissed):

src/components/duelling-picklist/picklist-group/picklist-group.component.js
```diff
@@ -7,22 +13,79 @@ import {
...
+  (
+    { title, children, type, onChange, index, listIndex, isLastGroup, ...rest },
+    ref
+  ) => {
+    const { setElementToFocus, elementToFocus } = useContext(FocusContext);
     const [highlighted, setHighlighted] = useState(false);
-    const handleClick = useCallback(() => onChange(), [onChange]);
+
+    const filteredChildren = React.Children.toArray(children).filter(
+      (child) => child
```
dawidzarzycki, line 28: what this function does? Is there anything that is not a child inside?

src/components/duelling-picklist/picklist/picklist.component.js
```diff
@@ -1,30 +1,38 @@
...
       Array.from(
         {
-          length: React.Children.count(children),
+          length: filteredChildren.length,
         },
         () => React.createRef()
       ),
-    [children]
+    [filteredChildren.length]
   );
```
dawidzarzycki, line 31: both functions are the same as above, maybe it will be good to create a `picklist.utils.js` file and move it there? Just an idea

src/components/duelling-picklist/picklist-group/picklist-group.component.js
edleeks87: It's to filter out any null/ falsy children, I could have written it as `filter(Boolean)` tbh but it does the same thing under the hood

src/components/duelling-picklist/picklist/picklist.component.js
edleeks87: I originally considered it but as it's only this one function I decided it was overkill and I wouldn't be able to `useMemo` which I think in this case is worth it

src/components/duelling-picklist/picklist-group/picklist-group.component.js
edleeks87: Actually looking into it you're right, `toArray` will filter out any no truthy children so `filter` isn't needed

dawidzarzycki (dismissed):

nicktitchmarsh (dismissed):

KatarzynaQA (dismissed):

KatarzynaQA (approved):

nicktitchmarsh (approved):

dawidzarzycki (approved):

PR #4056, opened 2021-05-21
docs(search): update stories with default values to not have same text used for placeholder FE-4069

edleeks87:
### Proposed behaviour

Updates the `defaultValue` text so it isn't the same text used for `placeholder` to make it clearer
that there is what using each prop does.

The `InputIconToggle` is slightly out of alignment within the `Search` component, when it is focused
you can see it is slightly misaligned. Adding a transform to move it down vertically by 1% solves
the issue as it is trying to centre itself within the input but the input has a height of 35px

### Current behaviour

Stories that use `defaultValue` all have text `Search` which is similar to the `placeholder` text.
The `InputIconToggle` is slightly out of alignment within the `Search` component, when it is focused
you can see it is slightly misaligned.

### Checklist

### Additional context

image

### Testing instructions

`design-system-search--default-story` etc

### DIFF
--- a/src/__experimental__/components/search/search.spec.js
+++ b/src/__experimental__/components/search/search.spec.js
@@ -10,6 +10,7 @@ import {
   testStyledSystemMargin,
 } from "../../../__spec_helper__/test-utils";
 import StyledTextInput from "../input/input-presentation.style";
+import StyledInputIconToggle from "../input-icon-toggle/input-icon-toggle.style";
 import Icon from "../../../components/icon";
 import TextBox from "../textbox";
 import { rootTagTest } from "../../../utils/helpers/tags/tags-specs";
@@ -198,6 +199,17 @@ describe("Search", () => {
           borderBottom: "none",
         },
         wrapper
+      );
+    });
+
+    it("applies the expected styling to the close icon", () => {
+      wrapper = renderWrapper({ value: "FooBar" }, mount);
+      assertStyleMatch(
+        {
+          marginBottom: "-1px",
+        },
+        wrapper,
+        { modifier: `${StyledInputIconToggle}` }
       );
     });
   });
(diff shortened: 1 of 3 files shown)


dawidzarzycki (approved):

mmaciorowski-sage (approved):

DipperTheDan (approved):

sjohnson-sage (commented):
I don't think the icon is misaligned because it's positioned centre, it's the extra 2px bottom padding on the Search that's causing the issue

I think just removing that padding will solve it rather than adding the transform, or even adding a -2px bottom margin to the icon (I think this is a situation it's ok to use one)

edleeks87:
Yeah initially I though it was that but removing the padding still leaves it overlapping more on the top than the bottom

image


> I don't think the icon is misaligned because it's positioned centre, it's the extra 2px bottom padding on the Search that's causing the issue
> 
> I think just removing that padding will solve it rather than adding the transform, or even adding a -2px bottom margin to the icon (I think this is a situation it's ok to use one)

sjohnson-sage:
> Yeah initially I though it was that but removing the padding still leaves it overlapping more on the top than the bottom
> 
> image
> 
> > I don't think the icon is misaligned because it's positioned centre, it's the extra 2px bottom padding on the Search that's causing the issue
> > I think just removing that padding will solve it rather than adding the transform, or even adding a -2px bottom margin to the icon (I think this is a situation it's ok to use one)

Fair enough, I do think a negative bottom margin is a better solution for this but I'll let it go here as it's pretty minor

mmaciorowski-sage (approved):

dawidzarzycki (approved):

ZhuoyuJin (approved):

PR #4063, opened 2021-05-24
fix(icon): hover only for interactive icons - FE-4080

mmaciorowski-sage:
### Proposed behaviour
hover only for interactive icons

### Current behaviour
Icons do not allow hover state to be disabled

### Checklist

### Additional context
Resolves: #4001

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-5pc5f?file=/src/index.js


src/components/icon/icon.spec.js
```diff
@@ -148,11 +148,36 @@ describe("Icon component", () => {
...
+        const wrapper = shallow(
+          <Icon type="home" color={color} tooltipMessage="tooltip message" />
+        );
+
         assertStyleMatch(
           {
             color: shade(0.2, toColor(baseTheme, color)),
           },
-          wrapper.find(StyledIcon),
+          wrapper.find(StyledIcon).dive().dive().dive(),
```
jamime, line 180: We didn't previously need to use `dive`, we try to avoid because it makes the test more brittle.

src/components/icon/icon.spec.js
```diff
@@ -251,14 +274,6 @@ describe("Icon component", () => {
             },
             wrapper.toJSON()
           );
-
```
jamime, line 254: Please can you add positive assertions for the different variants by making the wrapper interactive instead of deleting the test.

jamime (changes requested):

src/components/icon/icon.spec.js
mmaciorowski-sage: I'm not able to call `assertStyleMatch` if there are no styles.
Will it make sense?
Zrzut ekranu 2021-05-28 o 13 11 49

jamime: > making the wrapper interactive
If we add a tooltip, these tests should still pass... I think.

src/components/icon/icon.spec.js
mmaciorowski-sage: changed, thanks.

src/components/icon/icon.spec.js
mmaciorowski-sage: ahhh that was a tricky one :D changed, thanks!

jamime:
Moving out of draft so I can check storybook.

jamime (approved):

ianforster-sage (approved):

ZhuoyuJin (approved):

PR #4096, opened 2021-05-27
fix(radio-button): allow ref forwarding to allow access to the underlying input - FE-4095

mmaciorowski-sage:
### Proposed behaviour
The component is wrapped in a forwardRef.
Type definition is updated to include refs.

### Current behaviour
RadioButton input cannot be directly accessed by Carbon developers.

### Checklist

### Testing instructions

https://codesandbox.io/s/carbon-quickstart-forked-z6lrn

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button.d.ts
+++ b/src/__experimental__/components/radio-button/radio-button.d.ts
@@ -17,7 +17,7 @@ export interface RadioButtonProps extends CommonCheckableInputProps, MarginProps
   value: string;
 }
 
-declare function RadioButton(props: RadioButtonProps): JSX.Element;
+declare function RadioButton(props: RadioButtonProps & React.RefAttributes<HTMLInputElement>): JSX.Element;
 
 export { RadioButton as PrivateRadioButton };
 export default RadioButton;
--- a/src/__experimental__/components/radio-button/radio-button.spec.js
+++ b/src/__experimental__/components/radio-button/radio-button.spec.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useRef } from "react";
 import { mount } from "enzyme";
 import { ThemeProvider } from "styled-components";
 import TestRenderer from "react-test-renderer";
@@ -22,6 +22,7 @@ function render(props = {}, theme = mintTheme, renderer = mount) {
     error,
     info,
     warning,
+    name: "radio-button-name",
   };
   return renderer(
     <ThemeProvider theme={theme}>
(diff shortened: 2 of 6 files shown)


DipperTheDan (approved):

src/__experimental__/components/radio-button/radio-button.d.ts
```diff
@@ -15,6 +15,8 @@ export interface RadioButtonProps extends CommonCheckableInputProps, MarginProps
   size?: OptionsHelper.SizesType;
   /** the value of the Radio Button, passed on form submit */
   value: string;
+  /** A callback to retrieve the input reference */
+  ref?: () => void;
```
sjohnson-sage, line 19: This isn't the correct way to add the ref in the TypeScript interface, see https://github.com/Sage/carbon/blob/5332e810e3515151efc0329f1130312610d6359c/src/components/accordion/accordion.d.ts#L54 or any other component using `forwardRef` for an example of how to do this

sjohnson-sage: It won't be a `HTMLDivElement` though, it will probably be a `HTMLInputElement` if the ref is passed down to the input itself

src/__experimental__/components/radio-button/radio-button.spec.js
```diff
@@ -41,15 +42,26 @@ describe("RadioButton", () => {
     it("does not allow a children prop", () => {
       spyOn(console, "error");
       render({ children: "someChildren" });
-      const expected =
```
sjohnson-sage, line 44: Why did you change this?

mmaciorowski-sage: Because the message was changed for `ForwardRef`. I assumed that would be sufficient. Would you like the full message there?
Zrzut ekranu 2021-06-4 o 11 22 29

src/__experimental__/components/radio-button/radio-button.d.ts
mmaciorowski-sage: Good point, changed, thanks

src/__experimental__/components/radio-button/radio-button.spec.js
edleeks87: Could we not set `RadionButton.displayName = "RadioButton"` here to get the same error message output as before?

mmaciorowski-sage: Actually you are right, good point thanks. Changed.

edleeks87 (dismissed):

sjohnson-sage (approved):

sjohnson-sage (approved):

sjohnson-sage (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #4126, opened 2021-06-04
fix(icon): implement color values from theme correctly - FE-4140

mmaciorowski-sage:
### Proposed behaviour
Icon uses default `color` function instead of `toColor` which was part of the function.

Add information that `iconColor` and `bgTheme` props are deprecated and will soon be removed.

### Current behaviour
Icon uses the toColor function, but didn’t mixin the rest of the styled-system/color which is why it doesn't work with theme.colors
https://codesandbox.io/s/dreamy-sea-ym4yb?file=/src/index.js

### Checklist

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-4jv37


src/components/icon/icon.component.js
```diff
@@ -36,6 +39,13 @@ const Icon = React.forwardRef(
     },
     ref
   ) => {
+    if (!deprecatedWarnTriggered) {
```
edleeks87, line 42: I think we should scope these to only warn if either the props has been used to avoid spamming the console with warnings unnecessarily 
```suggestion
    if (!deprecatedWarnTriggered && (iconColor || bgThme)) {
```

mmaciorowski-sage: Good point, changed, thanks

edleeks87 (approved):

DipperTheDan (approved):

edleeks87 (approved):

DipperTheDan (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #4156, opened 2021-06-14
Dsy 632 create poc of design tokens implementation

jhajduk-sage:
This is POC of design tokens implementation in carbon.

New theme 'sbiz' has been created. It includes mintTheme for backwards compatibility. Eventually old themes can be renamed end extended with design tokens.

Design tokens files has been mocked in this commit, however after implementing correct transforms, those token formats will be generated in design token library and will be imported in carbon straight from @sage/design-tokens.


ianforster-sage (changes requested):
Changes of this nature must be proceeded by an RFC that has been reviewed and approved.
https://github.com/Sage/carbon/tree/master/rfcs

DlgSHi:
Hi, your commit needs to be `verified` - use this [doc](https://github.com/Sage/carbon/blob/master/docs/dev-environment-setup.md#gpg-verification) pls

PR #4134, opened 2021-06-08
refactor(options-helper): remove util

dawidzarzycki:
OptionsHelper has been removed.
All references related to OptionsHelper in components/files changed.
Some constants that are used more than once has been moved to
theirs config.js files with following pattern:

fixes: #4188 

`<component-name>.config.js`

All affected files are listed below:
- checkbox
- date
- decimal
- form-field
- input-presentation
- input-icon
- label
- numeral-date
- radio-button
- switch
- textarea
- textbox
- accordion
- action-popover-item
- alert
- story-constants
- batch-selection
- button
- button-toggle
- card
- confirm
- detail
- dialog
- flat-table
- form
- help
- icon
- inline-inputs
- link
- loader
- message
- message-content
- multi-action-button
- menu
- pod
- tile
- popover-container
- portrait
- profile
- row
- select-textbox
- show-edit-pod
- sidebar
- split-button
- step-sequence
- table
- tabs-header
- toast
- tooltip
- validation-cion
- vertical-divider
- image
- action-popover
- accordion
- input-definition
- styled-system-interfaces
- options-helper

### Proposed behaviour
Removing OptionsHelper and update all files that uses its reference

### Current behaviour
OptionsHelper exist

### Checklist

- [x] Commits follow our style guide
- [ ] Screenshots are included in the PR if useful
- [ ] All themes are supported if required
- [x] Unit tests added or updated if required
- [ ] Cypress automation tests added or updated if required
- [x] Storybook added or updated if required
- [x] Typescript `d.ts` file added or updated if required
- [ ] Carbon implementation and Design System documentation are congruent

### Additional context
OptionsHelper never showed properly in storybook instead of values it shows OptionsHelper name

### Testing instructions
Suggestions:
- It would be cool to take a look at all affected files in storybook (listed above)
- It will be worth as well to take a look at rest of the components as I could miss something while refactoring so many files

PR #4157, opened 2021-06-14
feat(confirm): more flexible button usage - FE-4152

mmaciorowski-sage:
### Proposed behaviour
- Primary, secondary and tertiary buttons should be able to be combined with each other as desired. (eg a Primary and a tertiary button)
- Buttons can have different colors (eg Primary button - destructive, tertiary button standard <Biz accent>
- A button can have an icon (placd left or right)

### Current behaviour
- button types available only for confirm button,
- destructive color is applied for both buttons
- no icons

### Checklist

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-3hu7k


src/components/confirm/confirm.component.js
```diff
@@ -2,18 +2,29 @@ import React from "react";
 import I18n from "i18n-js";
 import PropTypes from "prop-types";
 
+import OptionsHelper from "../../utils/helpers/options-helper";
```
edleeks87, line 5: As the `OptionsHelper` is being removed, I think it would be safer to just declare them in this file or adopt the same approach as https://github.com/Sage/carbon/pull/4134

edleeks87 (commented):
Code changes look fine to me, I'm just not too sure about using the `OptionsHelper`

src/components/confirm/confirm.component.js
mmaciorowski-sage: Good point, thanks. Changed and added FIXME comments there.

edleeks87 (dismissed):

jamime (commented):
Please can you add details to the commit message for the release note on how to use the codemod to replace the destructive prop.

https://github.com/Sage/carbon-codemod/tree/master/transforms/rename-prop

src/components/confirm/confirm.d.ts
```diff
@@ -2,14 +2,30 @@ import * as React from "react";
...
+  /** Customise the confirm button label */
   confirmLabel?: string;
   /** Apply destructive style to the buttons */
   destructive?: boolean;
+  /** Apply destructive style to the cancel button */
+  cancelButtonDestructive?: boolean;
+  /** Apply destructive style to the confirm button */
+  confirmButtonDestructive?: boolean;
+  /** Defines a cancel button Icon position related to the children: "before" | "after" */
+  cancelButtonIconPosition: "before" | "after";
```
DipperTheDan, line 20: I think this prop and the three props below need a `?:` as they are not required.

src/components/confirm/confirm.spec.js
```diff
@@ -306,5 +293,174 @@ describe("Confirm", () => {
         { modifier: `${StyledIcon}` }
       );
     });
+
+    it("confrim and cancel buttons should be `destructive`", () => {
```
DipperTheDan, line 297: Spelling mistake here. Should be `confirm`.

src/components/confirm/confirm.spec.js
```diff
@@ -306,5 +293,174 @@ describe("Confirm", () => {
...
+      const cancelButton = wrapper
+        .find(Button)
+        .find('[data-element="cancel"]')
+        .at(0);
+
+      expect(confirmButton.props().destructive).toBeTruthy();
+      expect(cancelButton.props().destructive).toBeTruthy();
+    });
+
+    it("confrim button should be `destructive`", () => {
```
DipperTheDan, line 322: Same spelling mistake here.

src/components/confirm/confirm.d.ts
mmaciorowski-sage: good point, thank you, changed

src/components/confirm/confirm.spec.js
mmaciorowski-sage: good point, thank you, changed

src/components/confirm/confirm.spec.js
mmaciorowski-sage: good point, thank you, changed

DipperTheDan (approved):

edleeks87 (approved):

harpalsingh:
I've passed the UXQA on this with the testing provided, but we should probably add something in storybook to show this is possible too.

mmaciorowski-sage:
> I've passed the UXQA on this with the testing provided, but we should probably add something in storybook to show this is possible too.

I've updated storybook there:
- Destructive cancel button
- Destructive confirm button
- Change confirm button type
- Change buttons icons
Do we want to add some more stories?

DlgSHi (approved):

PR #4163, opened 2021-06-16
docs(design-tokens): rfc for implementing design tokens in carbon

jhajduk-sage:
This PR contains RFC for implementing design tokens in carbon.

---
This is a RFC and it is important to note that this PR does not follow the typical PR process. Details of the RFC process can be found [here](https://github.com/Sage/carbon/blob/2163a70006e4e7853d6b4fe05703a2348a979f3e/rfcs/README.md).

Please use 👍 if you approve.

### DIFF
--- a/rfcs/text/design-tokens.md
+++ b/rfcs/text/design-tokens.md
@@ -1,0 +1,249 @@
+- Start Date: 2021-06-14
+
+
+- [Summary](#summary)
+- [Basic example](#basic-example)
+  - [Design Tokens Library](#design-tokens-library)
+  - [Updating Design Tokens Library](#updating-design-tokens-library)
+  - [Using proper design tokens](#using-proper-design-tokens)
+- [Motivation](#motivation)
+- [Detailed design](#detailed-design)
+  - [Global Styles](#global-styles)
+  - [Compatibility](#compatibility)
+    - [Providing new properties in old themes](#providing-new-properties-in-old-themes)
+    - [Providing old properties to new themes](#providing-old-properties-to-new-themes)
+    - [Generating css variable definitions from theme definitions](#generating-css-variable-definitions-from-theme-definitions)
+  - [Theme tokens](#theme-tokens)
+    - [Old themes](#old-themes)
+    - [New themes](#new-themes)
+    - [Another themes](#another-themes)
+  - [How themes overwriting would work](#how-themes-overwriting-would-work)
+- [Drawbacks](#drawbacks)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+---
+**_NOTE:_** Token and function names used in this document are not the final ones, so they may vary in actual implementation.
+
+---
+
+---
+**_NOTE:_** Implementing Design Tokens on Carbon will be breaking change.
+
+---
+
+# Summary
+Design System is a complete and well documented set of design standards, decisions and specifications along with the toolkit (ui patterns, code snippets, component libraries, all for various platforms) that allows for achieving those standards. Following those rules helps to create consistent visual experience for the final users.
+
+Design System's role is to create unified language for developers and designers. It, helps with communication, reduces spaces for misunderstandings and simplifies things and processes.
+
+To achieve this on the lowest level, Design System introduces Design Tokens, which are system's basic, atomic part.
+
+Basically those are key-value pairs named and organized the same way regardless of the platform (e.g. web, Android, iOS) and other tools (e.g. Figma). They can define various properties, such as colors, paddings, margins, sizes, font sizes, font families, transitions, animations, and others.
+
+Design tokens purpose is to:
+- **Release developers from taking design decisions**
+  
+  Currently, often while developing a component, developer needs to take decision what tint of what color should they use. This decision should be taken by designer, and not developer.
+- **Improve handover process and communication between designers and developers.**
+  
+  Both, developers and designers are going to use the same token name for given property (color, background color, border, padding, margin, transition and so on). In the end, developers don't need to know what the final value will be.
+- **Narrow value set to only needed values.**
+  
+  Design System uses narrow set of values (spacings, colors, typography properties and others). Those are only values that are needed for visual description of the component.
+- **Keep visual consistency across all components of all component libraries.**
+
+There are some restrictions from various sides. Some of them are on designers site (processes, design software or plugin's possibilities), and some of them are on developers site (used libraries, frameworks or processes). Therefore, while introducing Design Tokens, few compromises should be made. This requires openness of developers and designers to changes and suggestions to achieve best results. After all, this is a major change.
+
+# Basic example
+
+To put it in the simplest words: in practice, the introduction of design tokens in Carbon comes down to withdrawal of using `styled-components`'s `props.theme` object in favour of using css variables
+
+```js
+const OldButton = styled.button`
+  font-size: 14px;
+  padding: ${props => props.theme.space[1]}px ${props => props.theme.space[2]}px;
+  border: none;
+  cursor: pointer;
+  
+  color: ${props => props.theme.colors.whiteMix};
+  background-color: ${props => props.theme.colors.base};
+`;
+```
+
+Design tokens are going to be applied with css variables, like this:
+
+```js
+const NewButton = styled.button`
+  font-size: 14px;
+  padding: var(--sizing100) var(--sizing200);
+  border: none;
+  cursor: pointer;
+  
+  color: var(--colorsContent600);
+  background-color: var(--colorsBase500);
+`;
+```
+
+Such approach will require additional compatibility solutions. You can read more of that in [Detailed design](#detailed-design) section.
+
+## Design Tokens Library
+Design Tokens Library will be available in `@sage/design-tokens` NPM package.
+It is built automatically based on what designers will decide in their Figma Design System Libraries.
+NPM package will contain variable sheets for different platforms and formats. Among others, it will contain several formats of javascript variables (object and consts in CommonJS and ES6 formats along with Typescript definitions).
+
+Since carbon is not the only consumer of the tokens, we decided to publish it in a separate repository and as a separate package. This allows for developers on other platforms to have the same tokens that are available anywhere else.
+
+Design Tokens Library updates shall always be backwards compatible, which means that no tokens should be removed or renamed.
+
+## Updating Design Tokens Library
+At the first stage, during the time of migration, updating Design Tokens Library in Carbon should be responsibility of Design System Frontend Team. This way, the team can control changes that are upcoming and react to them as needed. 
+
+## Using proper design tokens
+All tokens, required for given component will be listed and pointed out in upcoming Design System documentation.
+
+# Motivation
+Design tokens are important part of the Design System. They will be introduced not only in carbon, but also on other platforms. It is important to use the same scales and palettes everywhere.
+It is also important to build components the same way as they are built by designers in Figma. Design Tokens will allow to achieve this as much as possible.
+
+Components across all platforms ideally should be built the same way. They should use the same Design Tokens. It helps to keep user experience consistent across all platforms.
+
+# Detailed design
+Design Tokens Library supports tree shaking. It's dist folder size is about 600KB, but Carbon will consume only its ~20KB part.
+
+## Global Styles
+Migrating Carbon to Design Tokens implies migrating from `styled-components`'s `ThemeProvider` to css variables. They are considered more flexible, and performant way of applying tokens.
+Such approach will require using `styled-component`'s [createGlobalStyle](https://styled-components.com/docs/api#createglobalstyle).
+Later on, after migration will be completed, Carbon consumers will have possibility to choose between GlobalStyle, or plain CSS variable sheets.
+
+## Scoped Styles
+For places, where css variables needs to be scoped to certain version (as for example in MFE), wrapper will be introduced. It will set up css variable values for all children inside wrapper.
+
+## Compatibility
+Using two different theming systems will require some compatibility solutions.
+
+1. Old themes will need to be extended with new properties.
+2. New themes will need to be extended with old properties.
+
+To achieve this, utility functions will be provided. 
+
+New properties will follow design tokens naming convention.
+
+## `color`, `bg` and other styled-system props
+For some components, there is still need of setting some of their inner colors with props. Those can not be deleted. Also, styling those inner elements with css path is not good practice, since HTML markup may change, and definitions may not be up to date.
+
+However, such properties should accept token name in the first place. If token does not exist, then they should work as they do it now with preserved logic.
+
+Eventually all dynamically generated colors, spacings and values defined by tokens should be replaced with tokens.
+
+### Providing new properties in old themes
+This util maps old theme properties to new theme properties or assigns hardcoded values if equivalent does not exist in an old theme.
+
+I see no other way than hardcode those keys and assign them values. Hopefully this job will be done only once in the beginning.
+
+### Providing old properties to new themes
+This util maps values of the old theme properties to pink/0px/none, depending on the value type. It does this mapping, to make not yet migrated properties more visible.
+
+### Generating css variable definitions from theme definitions
+String generated with this util is meant to be used with `createGlobalStyle`.
+
+## Theme tokens
+Tokens in Design Tokens Library are organized into sets which can represent themes.
+Most important set is the one called **sage**, since this is the base set for all other possible themes.
+
+If there will be need of creating a theme, another set of tokens will be created. It will contain only tokens which values will need to be overwritten. Then such set will be applied on top of base set.
+
+### Old themes
+Existing `mint` and `aegean` (later referred as *old themes*) should be extended with properties with names matching design tokens convention. This will be done using [newProperties](#newproperties) utility. 
+
+Old themes should be still obligatory themes, until migration will be finished.
+
+Example of `mint` theme in Carbon in `mint-theme.config.js`:
+```js
+import newProperties from '../utils/new-properties.util';
+
+const mintTheme = {
+  name: 'mint',
+  // [...]
+};
+
+export default {
+  ...mintTheme,
+  ...newProperties(mintTheme) // extend original mint theme with new theme properties.
+};
+
+```
+
+### New themes
+New theme - `sage` will be created. It should be extended with mapped properties of old theme using [oldProperties util](#oldproperties). 
+New themes should be marked as experimental until they will be at satisfactory level.
+
+`sage` theme in Carbon in `sage-theme.config.js`:
+```js
+import sage from '@sage/design-tokens/js/cjs/sage';
+import mint from '../themes/mint';
+import oldProperties from '../utils/old-properties.util';
+
+export default {
+  ...oldProperties(mint),
+  ...sage,
+  name: 'sage'
+};
+
+```
+
+### Another themes
+Extending Carbon with other themes will be as easy, as creating css variable sheet and overwriting values of specific design tokens. Later on, such theme may be applied globally, or to a specific scope.
+
+Themes should be created by designers with minimal participation of developers. However, if there is a justified need, variable sheets can be also created by developers.
+
+## How themes overwriting would work
+So the button will consume them this way:
+
+```js
+const Button = styled.button`
+  background: var(--colorsBase);
+  color: var(--colorsText);
+  padding: var(--spacing100) var(--spacing200);
+`;
+```
+
+The base theme is:
+```css
+:root {
+  --colorsBase: #333;
+  --colorsText: #999;
+  --spacing100: 8px;
+  --spacing200: 12px;
+}
+```
+
+Then, another theme, lets say mobile theme for mobile devices would consist of:
+```css
+.small {
+  --spacing100: 16px;
+  --spacing200: 20px;
+}
+```
+
+# Drawbacks
+Main drawback of proposed idea would be using two totally different systems of theming for some time. 
+
+However, it doesn't seem to be a blocker for Carbon developers, since migrating to CSS variables is responsibility of Design System Frontend Team which will be trained how to do it.
+
+# Adoption strategy
+Adoption of the idea consists of two stages:
+1. Implementing Design Tokens in Carbon. It consists of [extending existing themes](#old-themes) and creating [new theme](#new-themes)).
+2. Migration of component's properties to use design tokens.
+
+Implementing Design Tokens support as well as migrating component properties to Design Tokens is a responsibility of Design System Frontend Team.
+
+Main rules for migrating elements:
+1. During migration, new `sage` theme shall be marked as an experimental one, since it may contain unwanted changes.
+2. Usage of proper tokens should be consulted or approved by Designer. 
+3. Migration should be done at one component at the time.
+4. Migration commits ideally should remain possibly small and should contain only migration work.
+5. Pull Request should be created along with the [Contribution rules](https://github.com/Sage/carbon/blob/master/CONTRIBUTING.md).
+
+# How we teach this
+Design Tokens as a part of Design System as well as Design Tokens Library will have their own documentation. There is no need to create additional documentation within Carbon, but it is a good idea to provide links to the aforementioned documentation.


rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+```js
+const Container = styled.div`
+  background: ${props.theme.sagedsColorsBase400};
+  color: ${props.theme.sagedsColorsTextWhite};
+`;
+```
+
+Then, themes would be handled by ThemeProvider.
+
+Themes would be created from existing ones by extending and renaming them, so that current *mint* would be named *sbiz*
```
chrisbarber86, line 33: Why?

Why is the mint theme not styled from the design tokens rather than creating another theme.

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+
+This approach uses themes feature from `styled-components` since it is already implemented and uses javascript
+variables.
+
+At the moment, we have two levels of design tokens:
+
+- **Generic tokens** - kind of palette with all the values and scales.
+- **Theme tokens** - tokens for a theme, which contain references to generic tokens.
+
+So themes can be for example: *S Biz*, *M Biz*, *High Contrast*, *Dark Mode* and so on.
```
chrisbarber86, line 74: Where are these defined? What determines the ownership of the themes and how do we handle migration of themes

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+or *Medium* or *Large*
+
+In the end it all depends on designers, how they will build components in Figma, and which tokens they will use.
+
+# Drawbacks
+
+In case of CSS approach adaptation, there would be two theming systems available for carbon for the time of
+implementation.
+
+# Alternatives
```
chrisbarber86, line 86: I would like to see 'Not using design tokens' listed as an alternative here.

i.e. what are design tokens bringing to this that we can't get from the existing styled component themes in Carbon.

chrisbarber86 (changes requested):

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+
+Design tokens can contain values like lengths in different units, colors, font properties, transition definitions and
+others.
+
+Naming of the tokens is consistent across all platforms and provides predictable structure.
+
+To put it in simple words: When developer creates design for component, everything what you, as a developer, care about
+is which token has been used.
+
+# Detailed design
```
jamime, line 64: Please can you work out the increase in bundle size and the impact to our consumers. The color codes are generated using functions at the moment. We can't remove the colors at run time because of the `color` prop, although we could remove this functionality if there is a suitable alternative.

jamime (changes requested):

rfcs/text/design-tokens.md
chrisbarber86: Agree, `mint` was supposed to be the replacement for `small`, so if `sbiz` is intended to be the same thing, why does it need to be a new theme vs just using design tokens in the `mint` theme?

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+# Summary
+
+Implementation of the design tokens with theme support and backwards compatibility.
+
+# Basic example
+
+Design tokens will be kept in `@sage/design-tokens` library. It will be updated every time tokens will change. They will
```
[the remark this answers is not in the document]
chrisbarber86: Good point - can `@sage/design-tokens` be upgraded by downstream applications themselves or should carbon lock to an explicit version of `@sage/design-tokens` to prevent that (so we can enforce compatibility with carbon)

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+design tokens.
+
+While aligning component, only design tokens should be used. This implies, that design tokens library should be updated
+with the tokens that are needed for aligning given component. If there are any tokens missing, then it needs to be
+reported to Design System Team and situation needs to be explained.
+
+Whichever approach (JS or CSS) would be chosen, the adoption will consist in rewriting current styles of components
+using newly provided tokens (JS approach), or just entering CSS variables (CSS approach).
+
+In specific, justified cases, developer can use generic tokens. However, generic tokens shouldn't be referenced
```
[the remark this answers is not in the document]
jhajduk-sage: Generic tokens would contain wider range of predefined colors, and other tokens, for example: 
```js
const tokens = {
  colorsSlate50:  "#f2f5f6",
  colorsSlate100: "#e6ebed",
  colorsSlate200: "#ccd6db",
  colorsSlate300: "#b3c2c8",
  colorsSlate500: "#8099a4",
  colorsSlate600: "#668592",
  colorsSlate700: "#4d7080",
  colorsSlate800: "#335c6d",
  colorsSlate900: "#19475b",

  colorsAegean100: "#4DA0D9",
  colorsAegean200: "#1985CE",
  colorsAegean300: "#1985CE",
  colorsAegean400: "#0073C2",
  colorsAegean500: "#006BB4",
  colorsAegean600: "#005C9A",
  colorsAegean700: "#00538C",
  colorsAegean800: "#004472",
  colorsAegean900: "#003C64",

  spacing100: "2px",
  spacing200: "8px",
  spacing300: "8px",
  spacing400: "16px",
  spacing500: "32px",
  spacing600: "64px",
  spacing700: "72px",
  spacing800: "80px",
  spacing900: "88px",

  fontFamiliesBody: "Lato",
  fontFamiliesTitle: "Roboto"

  [...]
},
```

rfcs/text/design-tokens.md
jamime: We also need to consider how we support consumers that want to use a non-sage palette. Would they still write a their own theme and pass it to the theme provider or will this change too?

rfcs/text/design-tokens.md
jhajduk-sage: It needs to be renamed for keeping naming consistent between code and design system documentation

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jhajduk-sage: Every update of `@sage/design-tokens` (https://www.npmjs.com/package/@sage/design-tokens) should be compatible backwards. Then updates would be automated so that carbon would contain the latest version of the design tokens. Of course after updating tokens new minor or patch version of carbon would be released.

However I see it this way, that updating design tokens in carbon would be responsibility of the Design Sytem Frontend Team, and they would do it manually to have better control on whats changing. At least at the beginning stage, untill we gather informations on how it would be fully automated.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jhajduk-sage: However on the last meeting were thinking of not including this one in design tokens library, since according to design sytem, only predefined colors should be used.

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Adoption strategy
+
+Design tokens will be available globally in carbon. This means, that developers will be able to choose design tokens
+over current theme variables. However, the main process of design system alignment (which includes design tokens
+migration) will be handled by Design System Development Team and this will be the team, that will be most interested in
+design tokens.
+
+While aligning component, only design tokens should be used. This implies, that design tokens library should be updated
+with the tokens that are needed for aligning given component. If there are any tokens missing, then it needs to be
```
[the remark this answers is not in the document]
jhajduk-sage: But apart of that - can you explain to me why users would need specifying custom tints and shades?

rfcs/text/design-tokens.md
jhajduk-sage: Well, using js variables approach, thats what would they do - create own theme using base theme as a template (since it will contain all possible tokens), and then provide it in the Tehem Provider.

Using CSS approach, it would be even easier, since they would create their own themes by defining values for css variables, and then scoping them with classes, or use it globally with `:root{ ... }`

rfcs/text/design-tokens.md
jamime: I'm open to suggestions for the final solution. CSS is more performant, however the we need to support having multiple versions of carbon on the page at the same time (duplicate CSS selectors?)

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jamime: >  should be compatible backwards

Please can you clarify if you mean `will be` or might there be a breaking change at some point?

With this approach it means that some consumers might have to take a breaking change in carbon to get a non breaking change in the design system. We've avoided patching old versions of carbon due to the maintenance overhead but it might be beneficial for DS alignment.

What's your thought's on this?

- Consumer is on v60
- Carbon is on v80
- DS updates pallet and releases v80.0.1

How does Consumer on v60 update, do they have to update to v80 or will you patch v60.0.1?

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,154 @@
...
+variables.
+
+At the moment, we have two levels of design tokens:
+
+- **Generic tokens** - kind of palette with all the values and scales.
+- **Theme tokens** - tokens for a theme, which contain references to generic tokens.
+
+So themes can be for example: *S Biz*, *M Biz*, *High Contrast*, *Dark Mode* and so on.
+
+Themes can overwrite not only colors, but also sizes. This way it is possible to create size themes, such as *Small*
```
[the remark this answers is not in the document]
jamime: Thanks for the explanation please can you add this to the RFC. This will help communicate when we share this with other developers.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jamime: How does this differ from the base theme, wouldn't all colors be defined there?

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jamime: This is how UX teams currently specify the color to use e.g. `slateShade20`.

rfcs/text/design-tokens.md
jamime: It relates to the comment I left on L138
https://github.com/Sage/carbon/pull/4163#discussion_r653431651

We couldn't tree shake the unused colors because the consumer can specify the color at runtime so we'd need to include 201 variants of each color (base, + 100 shade, + 100 tint).

Based on your previous reply, maybe this will no longer be supported?

What is the plan for the `color` & `bg` props, where the consumer wants to override the color. Would this accept the colors in the theme e.g. `colorsSlate100`?

rfcs/text/design-tokens.md
jamime: Is the Design System open source?

jhajduk-sage: Some parts of it - yes. I see carbon library as an open source part of the design system.
As well as component libraries and tools for other platforms (android, iOS) can be open source I think.

rfcs/text/design-tokens.md
chrisbarber86: > Aligning carbon with design system without design tokens wouldn't have much sense. One of puproses of design systems is to unify what designers do and what developers code, automate it and reduce possibilities of misunderstandings during the handover proceses. And that is what design tokens are for.

You are missing the point that Carbon is **the** implementation of the design system - therefore if it already supports theming - what is the use of design tokens bringing that gives us any additional benefit over what we have now.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
chrisbarber86: The other question here is - should the design tokens be in a separate repo - or are they just part of carbon?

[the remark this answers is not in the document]
jhajduk-sage: >Please can you clarify if you mean will be compatible backwards or might there be a breaking change at some point?

I mean, that new themes will be usable with all components inside the same version of carbon, since migrating to design tokens will take some time.

> What's your thought's on this?
>
> Consumer is on v60
> Carbon is on v80
> DS updates pallet and releases v80.0.1

Right now, any change of the color, or other values requires releasing new version of the carbon. And this workflow does not have to change.

However with design tokens it will be possible to separate this concern by providing themes to carbon from external source, whether we'll go with CSS o JS approach. Carbon would have only its default theme included.

And how are you dealing with such changes right now?

> The other question here is - should the design tokens be in a separate repo - or are they just part of carbon?

Design Tokens should be kept in separate reposiotry, since they are consumed not only by carbon, but also by libraries on iOS or Android. By definiton they should be consistent across all platforms, and keeping it in a separate repository is a good practice to achieve this.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jhajduk-sage: Using deisgn tokens will also require kind of base theme, exactly the same way as it is done now.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jhajduk-sage: Okay, so this probably will change, since designers will also use design tokens.

rfcs/text/design-tokens.md
jhajduk-sage: This would be the big change. Introduction of the Deisgn System will require consumers to adapt to the new rules. If consumer wants to be well aligned with DS, then he has to use theme tokens. 

If he needs a custom color, then he can use generic token (if eventually they will be included), or use custom color and provide it by hex, however then, the responsibility of not aligning to the Design System is on his side.

rfcs/text/design-tokens.md
jhajduk-sage: I didn't missed that. I see that carbon implements design system. However at the moment it is not implementing, which are the part of new Design System. That's why they will have to be implemented also in carbon.

> what is the use of design tokens bringing that gives us any additional benefit over what we have now.

Generally they are giving us consistency in all components across different platforms, and they are unifying language between designers, and developers in all technologies.

Design Tokens and theming are two, totally different terms and they are not connected directly, but theming functionality has impact on tokens naming.
- Theming is a feature, that allows users to change appearance of the library elements. As a functionality, it doesnt care about specific values.
- Design Tokens are sets of values that represents design decisions and are separated from the developers.

So in the end - themes can be extended with design tokens to keep this consistency.

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jamime: > And how are you dealing with such changes right now?

We follow the same approach today, consumers must be on latest carbon version to get latest theme.

I'm happy for it to stay that way, but it's important that this decision is documented. A simple change to a color, will require taking any breaking changes that have occurred in carbon since the last update.

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,352 @@
...
+      // [...]
+    },
+  
+    // [...]
+  };
+};
+```
+
+### New themes
+At the same time, new themes `sBiz` (small business), `mBiz` (medium business) and `sServ` (shared services) (later referred as *new themes*) should be created. They should be importing existing theme tokens, and should be extended with old themes for preserving backwards compatibility. Values of the 'old properties' should be overwritten with the new token values.
```
chrisbarber86, line 203: I think this needs updating as it's now one theme.

chrisbarber86 (changes requested):

rfcs/text/design-tokens.md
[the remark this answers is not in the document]
jamime: Please can you document this in the RFC as a breaking change and explain how users should upgrade.
e.g. I'm using `<Component color="slateShade20" />` how can I upgrade.

rfcs/text/design-tokens.md
jamime: Please can you document this in the RFC as a breaking change and explain how users should upgrade.
e.g. I'm using <Component color="slateShade20" /> how can I upgrade.

What was the outcome of the bundle size? I suspect if we're not supporting every shade and tint for each colour it will not be a problem (i.e. no more `slateShade20`, `slateShade21`, `slateShade22`).

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,232 @@
...
+Such approach will require additional compatibility solutions. You can read more of that in [Detailed design](#detailed-design) section.
+
+## Design Tokens Library
+Design Tokens Library will be available in `@sage/design-tokens` NPM package.
+It is built automatically based on what designers will decide in their Figma Design System Libraries.
+NPM package will contain variable sheets for different platforms and formats. Among others, it will contain several formats of javascript variables (object and consts in CommonJS and ES6 formats along with Typescript definitions).
+
+Since carbon is not the only consumer of the tokens, we decided to publish it in a separate repository and as a separate package. This allows for developers on other platforms to have the same tokens that are available anywhere else.
+
+Design Tokens Library updates shall always be backwards compatible, which means that no tokens should be removed or renamed.
```
jamime, line 91: I think we should also NOT change spacing values. This has caused problems in the past when a layout wraps incorrectly with the new size. What are your thoughts?

rfcs/text/design-tokens.md
jhajduk-sage: dist folder size of the Design Tokens Library pacakge is about 600 KB, but since It will use single theme from DTL, it will use only about 20 KB, and should tree-shae the rest of it.

I'll update RFC with and bundle size.

rfcs/text/design-tokens.md
jhajduk-sage: I see the problem you are talking about and in you may be right that it may be dangerous to change spacing values.

However I consider spacings inside components as a design decisions. And design tokens are taking them off from developers and passes them to Designers. So it is up to them what spacings are required.

And If any changes will require updating some other components, then we need to do it anyway.

rfcs/text/design-tokens.md
jamime: It's not clear how tree shaking will reduce the size of the CSS file? Isn't it true that we'll need to include all CSS properties because we don't know which ones are used at runtime?

Looks like my previous comment was truncated.

> e.g. I'm using how can I upgrade.

should have been

> e.g. I'm using `<Component color={slateShade20}  bg={slateShade21}/>`how can I upgrade.

rfcs/text/design-tokens.md
jamime: The spacing of component also affects page layout. 

We've previously discussed having two version of carbon on the page at the same time, in micro frontends, owned by different teams that release at different times.


Assuming a single global style sheet this change, will break many layouts 

```diff
.spacing100 {
-  padding: 4px;
+  padding: 8px;
}
```

I think that `spacing100` should remain `4px` and a new value should be introduced for `8px`.

Alternatively, the stylesheet needs to be scoped to a project. There would be multiple, duplicated styles on the page at any one time. This is closest to what we have today using `uuid` class names from `styled-components`.

```diff
-.spacing100 {
-  padding: 4px;
+#app1 .spacing100 {
+  padding: 8px;
}
```


Please can you document your chosen approach in the RFC.

rfcs/text/design-tokens.md
jhajduk-sage: During the time of implementation, we're going to use js variables, that will be translated into css variables and provided with `GlobalStyle`.
In package there will be available different sheets - scss, css, js (es6 and commonjs).
Also, there will be available different themes.
Overall size of package is about 600KB at the time.
Most probably, Carbon will use only one js file (es6 or commonjs) from the whole package.
So what I meant by saying "treeshaking" is that so all the others wont be bundled into final Carbon build.

> e.g. I'm using <Component color={slateShade20} bg={slateShade21}/>how can I upgrade.

You shouldn't be overwriting values since it breaks component alignment with the DS documentation.
All properties that need to be set, should be set inside component.
However if you really need to overwrite any property, you can do it by redefining css variable locally, for given component. e.g. in class definition.

rfcs/text/design-tokens.md
jhajduk-sage: we're going to use css variables instead of classes, so instead of
```
.spacing100 { [...] }
```
it will be:
```
:root {
  --spacing100: 4px;
}

#app {
  --spacing100: 8px
}

.btn.medium.localhchange {
  --spacing: 6px;;
}
```

So there always will have to be certain scope.

jhajduk-sage: So you can use variables from global scope, but if you need to, you can introduce your own scope and use different version of DTL for that

jamime: So in terms of the global scope `--spacing100`, the the global value changes, that could break a lot of layouts. In particular when we have two versions of carbon. One Micro Frontend wants to use the new spacing values, but the other micro frontend layouts would break.

How will we avoid this?

rfcs/text/design-tokens.md
jamime: To be clear, DS team will remove the existing `color` and `bg` props from all components?

jhajduk-sage: After consideration, as we agreed on, `color` and `bg` props shouldn't be removed.
Also, styling inner components cannot be done by referring to inner elements by css path.

So the way it should work is to pass color using design token name:
```
<CarbonComponent color="sageColorAction500" />
```
so the `color` prop should accept token name and use its value or preserve existing logic for retrieving value.

rfcs/text/design-tokens.md
jhajduk-sage: As we agreed in conversastion: 
There will be additional Design Tokens Provider implemented for scoped tokens.

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,249 @@
...
+export default {
+  ...mintTheme,
+  ...newProperties(mintTheme) // extend original mint theme with new theme properties.
+};
+
+```
+
+### New themes
+New theme - `sage` will be created. It should be extended with mapped properties of old theme using [oldProperties util](#oldproperties). 
+New themes should be marked as `WIP` or `alpha` until they will be at satisfactory level.
```
jamime, line 179:
```suggestion
New themes should be marked as experimental until they will be at satisfactory level.
```

Use same terminology as below.

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,249 @@
...
+2. Migration of component's properties to use design tokens.
+
+Implementing Design Tokens support as well as migrating component properties to Design Tokens is a responsibility of Design System Frontend Team.
+
+Main rules for migrating elements:
+1. During migration, new `sage` theme shall be marked as an experimental one, since it may contain unwanted changes.
+2. Usage of proper tokens should be consulted or approved by Designer. 
+3. Migration should be done at one component at the time.
+4. Migration commits ideally should remain possibly small and should contain only migration work.
+5. Pull Request should be created along with the [Contribution rules](https://github.com/Sage/carbon/blob/master/CONTRIBUTING.md)
```
jamime, line 246:
```suggestion
5. Pull Request should be created along with the [Contribution rules](https://github.com/Sage/carbon/blob/master/CONTRIBUTING.md).
```

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,249 @@
...
+
+# How we teach this
+Design Tokens as a part of Design System as well as Design Tokens Library will have their own documentations. There is no need to create separate page in Carbon Documentation, but it is a good idea to provide links to those documentations.
```
jamime, line 249:
```suggestion
Design Tokens as a part of Design System as well as Design Tokens Library will have their own documentation. There is no need to create additional documentation within Carbon, but it is a good idea to provide links to the aforementioned documentation.
```

jamime (changes requested):

jamime:
This RFC has entered the "Final Comments" stage. This will be accepted unless there are significant changes requested before Monday 13th September 2021 17:28 BST.

jamime (approved):

nicktitchmarsh (approved):

chrisbarber86 (dismissed):

DipperTheDan (approved):

DlgSHi (dismissed):

rfcs/text/design-tokens.md
```diff
@@ -0,0 +1,249 @@
...
+Design System's role is to create unified language for developers and designers. It, helps with communication, reduces spaces for misunderstandings and simplifies things and processes.
+
+To achieve this on the lowest level, Design System introduces Design Tokens, which are system's basic, atomic part.
+
+Basically those are key-value pairs named and organized the same way regardless of the platform (e.g. web, Android, iOS) and other tools (e.g. Figma). They can define various properties, such as colors, paddings, margins, sizes, font sizes, font families, transitions, animations, and others.
+
+Design tokens purpose is to:
+- **Release developers from taking design decisions**
+  
+  Currently, often while developing a component, developer needs to take decision what tint of what color should he use. This decision should be taken by designer, and not developer.
```
PaulSumner-Sage, line 47: Suggestion:
Change usage of "he" for "they":

> Currently, often while developing a component, developer needs to take decision what tint of what color should they use. This decision should be taken by designer, and not developer.

jamime (approved):

chrisbarber86 (approved):

PR #4198, opened 2021-06-24
fix(tab-title): remove call to stoppropagation onclick of tab title FE-4190 remove stoppropagaation

edleeks87:
### Proposed behaviour

Removes `ev.stopPropagation` call from click handler in `TabTitle` and removes removes passing of
`onClick` to root element as it is already passed to element a  level below

Removes the `setOpenState` call from all `Select` components `handleTextboxBlur` function as it was
only needed because we were stopping event propagation in `Tab`
### Current behaviour

stopPropagation call in `TabTitle` click handler

### Checklist

### Testing instructions

https://codesandbox.io/s/carbon-quickstart-forked-8guby?file=/src/index.js


grabkowski (dismissed):

src/components/select/simple-select/simple-select.component.js
```diff
@@ -247,12 +247,11 @@ const SimpleSelect = React.forwardRef(
     }
 
     function handleTextboxBlur(event) {
+      console.log("called");
```
mkrds, line 250: Console.log leftover

edleeks87: 🤦  updated now @mkrds

mkrds (approved):

grabkowski (approved):

ZhuoyuJin (approved):

Issue #4024, opened 2021-05-14
Tabs on left are changing width when content change


maspychaj-sage:
### Current behaviour
When text in tab is longer then space in the tab it make content narrower.
![image](https://user-images.githubusercontent.com/75613789/118251455-e5a0f080-b4a7-11eb-9222-6d92afe3c30e.png)

### Expected behaviour
Tab width should be the same width for any content.

### Reproducible example
https://codesandbox.io/s/objective-fog-0t19d?file=/src/index.js

### Suggested solution(s)
Prop for set width of tabs or max width.

### Your environment
| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     |  74.0.0
| react-scripts    |  4.0.3
| React            | 16.14.0
| Browser          | any
| Operating System | any

sjohnson-sage:
@maspychaj-sage what's the use case for having two sets of tabs on top of one another?

maspychaj-sage:
> 
> 
> @maspychaj-sage what's the use case for having two sets of tabs on top of one another?

I've got two Tabs on two different pages and it's not consistent when you change pages. It's not possible to have the same layout when tabs strings are longer.

nicktitchmarsh:
FE-4161 has been raised to address the width of the tabs

PR #4202, opened 2021-06-28
feat(tabs): add headerWidth prop (FE-4161)

dawidzarzycki:
`headerWidth` prop will set a width to `TabsHeader` component.
It will allow users with very long title to make well aligned `Tabs`
if there will be more of tabs group than one

fixes: #4024

### Proposed behaviour
image

add headerWidth prop

### Current behaviour
lack of support for custom width

### Checklist

### Testing instructions
New storybook story created `with tabsHeaderWidth`


src/components/tabs/tabs.d.ts
```diff
@@ -26,6 +26,8 @@ export interface TabsProps extends MarginProps {
   borders?: "off" | "on" | "no left side" | "no right side" | "no sides";
   /** Adds an alternate styling variant to the tab titles. */
   variant?: "default" | "alternate";
+  /** sets width to TabHeader */
+  tabsHeaderWidth?: string,
```
DipperTheDan, line 30: This will be why lint is failing. Should be a `;` not a `,`

src/components/tabs/__internal__/tabs-header/tabs-header.component.js
```diff
@@ -41,6 +46,10 @@ TabsHeader.propTypes = {
   alternateStyling: PropTypes.bool,
   noRightBorder: PropTypes.bool,
   isInSidebar: PropTypes.bool,
+  /**
+   * @ignore @private
```
sjohnson-sage, line 50: This doesn't need to have `@ignore @private` on it because the whole component is internal

src/components/tabs/tabs.component.js
```diff
@@ -356,6 +358,8 @@ Tabs.propTypes = {
     "no right side",
     "no sides",
   ]),
+  /** sets tabsHeaderWidth to TabHeader */
+  tabsHeaderWidth: PropTypes.string,
```
sjohnson-sage, line 362: I think just call this prop `headerWidth`, no need to have `tabs` before it because it's a prop of the `Tabs` component

src/components/tabs/tabs.component.js
```diff
@@ -356,6 +358,8 @@ Tabs.propTypes = {
     "no right side",
     "no sides",
   ]),
+  /** sets tabsHeaderWidth to TabHeader */
```
sjohnson-sage, line 361:
```suggestion
  /** sets width to the tab headers. Can be any valid CSS string */
```

sjohnson-sage: The user has no visibility of the `TabHeader` component because it's an internal one

src/components/tabs/tabs.d.ts
```diff
@@ -26,6 +26,8 @@ export interface TabsProps extends MarginProps {
   borders?: "off" | "on" | "no left side" | "no right side" | "no sides";
   /** Adds an alternate styling variant to the tab titles. */
   variant?: "default" | "alternate";
+  /** sets width to TabHeader */
```
sjohnson-sage, line 29: Same comments as in the js file

src/components/tabs/tabs.stories.mdx
```diff
@@ -1520,6 +1520,64 @@ By setting the `variant` prop to "alternate" it is possible to apply some additi
   </Story>
 </Preview>
 
+### With tabsHeaderWidth
```
sjohnson-sage, line 1523: Rename as per above comments

sjohnson-sage (changes requested):

DipperTheDan (commented):
I had pretty much the same comments as @sjohnson-sage so once you address Sam's comments, I'll be happy to give the PR a tick ✅

src/components/tabs/tabs.component.js
dawidzarzycki: I had in mind "we can change headers in every tab component" not the header tabs component itself :D  but you are right, it headerWidth is enough anyway

DipperTheDan (approved):

sjohnson-sage (dismissed):

harpalsingh:
Is this an issue with the width prop or just storybook?


https://user-images.githubusercontent.com/2411024/123764762-fde49780-d8bc-11eb-8c11-9da54ba4ba77.mov

dawidzarzycki:
> Is this an issue with the width prop or just storybook?
> 
>  tabs.mov

I'm pretty sure it is a bug, I will fix it

src/components/tabs/tabs.component.js
```diff
@@ -356,6 +358,10 @@ Tabs.propTypes = {
     "no right side",
     "no sides",
   ]),
+  /** sets width to the tab headers. Can be any valid CSS string.
+   * The headerWidth prop works only for `position="left"`
```
DipperTheDan, line 362: Is it worth maybe having a custom prop validation here now that this prop works only when the `position="left"`

DipperTheDan (dismissed):

nicktitchmarsh (dismissed):

DipperTheDan (approved):

nicktitchmarsh (approved):

DlgSHi (approved):

PR #4204, opened 2021-06-29
fix(radio-button, checkable-input): fix invalid values in ARIA attributes - FE-1700

DipperTheDan:
resolves the errors in axe relating to ARIA attributes must conform to valid values.

fixes FE-1700

### Proposed behaviour
No accessibility errors present for `radio-button` in Storybook.
Screenshot 2021-07-01 at 15 04 00

### Current behaviour
Axe throws errors: 
- Ensures all ARIA attributes have valid values
- Ensures elements with an ARIA role that require child roles contain them
Screenshot 2021-07-01 at 15 03 19

### Checklist

### Additional context
N/A

### Testing instructions
Check all `radio-button` stories for accessibility warnings in the Storybook `accessibility` tab.

### DIFF
--- a/src/__experimental__/components/radio-button/radio-button-group.component.js
+++ b/src/__experimental__/components/radio-button/radio-button-group.component.js
@@ -56,7 +56,6 @@ const RadioButtonGroup = (props) => {
 
   return (
     <Fieldset
-      role="radiogroup"
       legend={legend}
       error={error}
       warning={warning}
@@ -73,7 +72,7 @@ const RadioButtonGroup = (props) => {
     >
       <RadioButtonGroupStyle
         data-component="radio-button-group"
-        role="group"
+        role="radiogroup"
         inline={inline}
         legendInline={inlineLegend}
       >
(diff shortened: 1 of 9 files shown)


src/__experimental__/components/radio-button/radio-button-group.component.js
```diff
@@ -73,7 +73,7 @@ const RadioButtonGroup = (props) => {
     >
       <RadioButtonGroupStyle
         data-component="radio-button-group"
-        role="group"
+        role="radio"
```
edleeks87, line 76: Are you sure this is the right role for this? My understanding is that the `role="radio"` attribute will identify this as as an ARIA radio button. I think this needs to be `role="radiogroup"` as it's the container for the radio buttons, but I'm not sure of the impact of moving the role from the `FieldSet` in this component is tbh

DipperTheDan: That's a valid point. `radio` will give the impression of a single radio whereas this actually a radio group so should be `radiogroup`. I'll try moving the role from the `FieldSet` and if axe throws any errors.

ianforster-sage (dismissed):
Nothing to add to @edleeks87 comment. Happy to approve when you resolve that.

edleeks87 (approved):

ianforster-sage (approved):

DlgSHi (dismissed):

DlgSHi (approved):

edleeks87 (approved):

Issue #4190, opened 2021-06-23
FlatTable - text within cell with rowSpan of 2 is positioned absolutely


sjohnson-sage:
### Current behaviour
The text within a cell with a `rowSpan` of `2` or more in a `FlatTable` is positioned absolutely, which doesn't look right when you scroll.

See the below sandbox, the top left header cell.

Also when scrolling up it seems to be offset with the other header rows: 

![image](https://user-images.githubusercontent.com/14963680/123109960-94373a00-d433-11eb-91c9-3dfe8de22e3b.png)

### Expected behaviour

The text should scroll with the cell, as in the other cells, and the row should be aligned correctly when scrolled.

### Reproducible example
https://codesandbox.io/s/reverent-sea-5zmxb?file=/src/index.js:902-909

### Your environment
| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | Latest
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

grabkowski:
FE-4193

PR #4212, opened 2021-07-02
fix(flat-table): scrolling content in table header - FE-4193

dawidzarzycki:
### Proposed behaviour

fixes: #4190

### Current behaviour
There was an issue if there was not enough space for the table and scroll appears.
It causes that text inside of sticky header was scrolling when user used scroll

### Checklist

### Testing instructions

Testable in codesandbox

### DIFF
--- a/src/components/flat-table/flat-table.style.js
+++ b/src/components/flat-table/flat-table.style.js
@@ -176,6 +176,8 @@ const StyledFlatTableWrapper = styled.div`
       ${StyledFlatTableHead} ${StyledFlatTableRowHeader},
       ${StyledFlatTableHead} ${StyledFlatTableCheckbox} {
         z-index: ${({ theme }) => theme.zIndex.overlay + 2};
+        top: 0;
+        left: 0;
       }
     `}
 `;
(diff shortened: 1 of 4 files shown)


edleeks87 (changes requested):
This will break the functionality when `hasStickyHead` is set on the `FlatTable`. I think you're on the right path but might need a little more tweaking to get the behaviour right in all scenarios

https://codesandbox.io/s/silly-wind-0hvo3

ianforster-sage (dismissed):
This looks OK. All the tests are passing and the snapshots are unchanged, which is good. Though it did make me wonder why there is no assertStyleMatch on the thead.

edleeks87 (dismissed):

src/components/flat-table/flat-table.style.js
```diff
@@ -95,6 +95,15 @@ StyledFlatTableRoot.defaultProps = {
 };
 
 const StyledFlatTableWrapper = styled.div`
+  ${({ hasStickyHead }) =>
```
sjohnson-sage, line 98: There's already a function like this for `hasStickyHead` in `StyledFlatTableWrapper` - you should just put this CSS in there instead 👍🏻

dawidzarzycki: @sjohnson-sage good point! :)

sjohnson-sage (dismissed):

edleeks87 (approved):

ZhuoyuJin (approved):

ZhuoyuJin (approved):

sjohnson-sage (approved):

edleeks87 (approved):

Issue #4217, opened 2021-07-06
FlatTable height prop no longer applied to root element


sjohnson-sage:
### Current behaviour
From v76.0.0 onwards the `height` prop passed into `FlatTable` no longer works correctly as it isn't applied to the new root element.

![image](https://user-images.githubusercontent.com/14963680/124586402-4c121180-de4e-11eb-87da-e8355da7b86f.png)

### Expected behaviour
The `height` prop should be applied correctly to `FlatTable`

### Reproducible example
https://codesandbox.io/s/hopeful-waterfall-9rntr

### Suggested solution(s)
Apply the height prop to the root element. This now wraps the table footer as well so the table height will apply to the whole table.

### Your environment
| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | Latest
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

sjohnson-sage:
FE-4203

sjohnson-sage:
Closing this as we need to go back to the DS team and have a proper think about how we incorporate the table footer into the height calculations etc

PR #4220, opened 2021-07-06
fix(flat-table): apply height and maxHeight props to root element

sjohnson-sage:
Fixes: #4217

### Proposed behaviour
Apply the `height` and `maxHeight` props to the root element of the `FlatTable`. 

### Current behaviour
The root element always has `height: 100%` so the height prop no longer works correctly, as it is passed down further to a non-root element.

### Checklist

### Testing instructions
Regression test of all other `FlatTable` stories will be needed.

https://codesandbox.io/s/thirsty-sound-08j9b to test the particular issue raised which triggered this change

### DIFF
--- a/src/components/flat-table/flat-table.component.js
+++ b/src/components/flat-table/flat-table.component.js
@@ -2,7 +2,6 @@ import React from "react";
 import PropTypes from "prop-types";
 import styledSystemPropTypes from "@styled-system/prop-types";
 import {
-  StyledFlatTableRoot,
   StyledFlatTableWrapper,
   StyledFlatTable,
   StyledFlatTableFooter,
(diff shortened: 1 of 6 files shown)


src/components/flat-table/flat-table.component.js
```diff
@@ -44,12 +44,15 @@ const FlatTable = ({
...
+          {...filterStyledSystemMarginProps(rest)}
+          height={addDefaultHeight && !hasMaxHeight ? "100%" : height}
+          maxHeight={hasMaxHeight ? "100%" : undefined}
+        >
           <StyledFlatTableBox
             {...rest}
             {...((hasStickyHead || hasStickyFooter) && { overflowY: "auto" })}
-            height={addDefaultHeight && !hasMaxHeight ? "100%" : height}
-            maxHeight={hasMaxHeight ? "100%" : undefined}
+            height={footer ? "calc(100% - 40px)" : "100%"}
```
mmaciorowski-sage, line 55: Shouldn't this logic be part of the `StyledFlaTableBox`?

sjohnson-sage: I don't think so no, it has a height prop so we might as well use it

mmaciorowski-sage (approved):

ianforster-sage (approved):

sjohnson-sage:
Closing as no longer required

Issue #4211, opened 2021-07-02
Form sticky footer doesn't respect the container width when rendered within a sidebar


blaky:
### Current behaviour
Sticky footer of Form component within sidebar expands to the full width of the screen.

### Expected behaviour
The sticky footer should only expand to the width of the sidebar that it is located within.

### Reproducible example
https://codesandbox.io/s/sleepy-framework-0q4o5

### Additional context
When scrolled to the top and the footer "sticks" it expands to the full available screen width.
![image](https://user-images.githubusercontent.com/4557844/124261393-06450880-db31-11eb-8a11-0eec0707d106.png)
It works correctly when the user scrolls to the bottom
![image](https://user-images.githubusercontent.com/4557844/124261534-3096c600-db31-11eb-8005-134c87580019.png)

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | 77.3.5 (reproducible on any latest version) |
| Operating System | OSX, Windows|

ianforster-sage:
FE-4205

PR #4229, opened 2021-07-09
fix(sidebar): Support for Form with stickyFooter FE-4205

dawidzarzycki:
### Proposed behaviour
Fixed Form Footer inside of Sidebar
fixes: #4211

### Current behaviour
When Form with Form Footer in sticky mode is provided to Sidebar, the footer getting out of component

### Checklist

### Additional context
In this ticket I renamed SidebarContext that was in Drawer component because it was confusing. New name for Drawer context is DrawerSidebarContext.
I added new context SidebarContext to Sidebar component

### Testing instructions
Testable in codesandbox

### DIFF
--- a/src/components/drawer/drawer.component.js
+++ b/src/components/drawer/drawer.component.js
@@ -16,7 +16,7 @@ import {
 } from "./drawer.style";
 import StickyFooter from "../../__internal__/sticky-footer";
 
-const SidebarContext = React.createContext();
+const DrawerSidebarContext = React.createContext();
 
 const Drawer = ({
   defaultExpanded = true,
@@ -177,9 +177,9 @@ const Drawer = ({
           scrollVariant="light"
           ref={scrollableContentRef}
         >
-          <SidebarContext.Provider value={{ isInSidebar: true }}>
+          <DrawerSidebarContext.Provider value={{ isInSidebar: true }}>
             {sidebar}
-          </SidebarContext.Provider>
+          </DrawerSidebarContext.Provider>
           {footer && (
             <StickyFooter
               containerRef={scrollableContentRef}
@@ -226,6 +226,6 @@ Drawer.propTypes = {
   stickyFooter: PropTypes.bool,
 };
 
-export { SidebarContext };
+export { DrawerSidebarContext };
 
 export default Drawer;
--- a/src/components/form/form.component.js
+++ b/src/components/form/form.component.js
@@ -1,4 +1,10 @@
-import React, { useState, useRef, useEffect, useCallback } from "react";
+import React, {
+  useState,
+  useRef,
+  useEffect,
+  useCallback,
+  useContext,
+} from "react";
 import PropTypes from "prop-types";
 import throttle from "lodash/throttle";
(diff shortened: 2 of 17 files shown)


src/components/sidebar/sidebar.component.js
```diff
@@ -9,6 +9,8 @@ import FocusTrap from "../../__internal__/focus-trap";
 import SidebarHeader from "./__internal__/sidebar-header";
 import Box from "../box";
 
+export const SidebarContext = React.createContext();
```
edleeks87, line 12: This needs a default value 

`export const SidebarContext = React.createContext({});`

src/components/form/form.component.js
```diff
@@ -30,11 +37,10 @@ const Form = ({
   ...rest
 }) => {
   const [isFooterSticky, setIsFooterSticky] = useState(false);
-
+  const sidebarContext = useContext(SidebarContext);
```
edleeks87, line 40: If you provide a default value to the `SidebarContext` you can destructure `isInSidebar` instead here

src/components/drawer/drawer.component.js
```diff
@@ -16,7 +16,7 @@ import {
 } from "./drawer.style";
 import StickyFooter from "../../__internal__/sticky-footer";
 
-const SidebarContext = React.createContext();
+const DrawerSidebarContext = React.createContext();
```
edleeks87, line 19: Can you add a default value to the context here as well please. I'll create a new ticket to remove the `context && context.isInSidebar` code from the consuming components as it's not relevant for you to do that here

edleeks87 (commented):
I think the first commit message needs to have scope adding `(drawer, flat-table, flat-table-row, tabs)`

dawidzarzycki:
> I think the first commit message needs to have scope adding `(drawer, flat-table, flat-table-row, tabs)`

Hey I don't think that we need to add more components than drawer. This is a change in Drawer component to be honest, the rest components just uses the value so It's like we are changing only drawer to be honest

src/components/sidebar/sidebar.component.js
dawidzarzycki: Good spot!

DipperTheDan (commented):
Looks good. Just fix the suggestions from @edleeks87 and I'll be happy to approve 😄

DipperTheDan (dismissed):

src/components/drawer/drawer.component.js
dawidzarzycki: Update:
Adding default value causes test failures. I suggest to create a ticket for this instead of doing it in this PR.
@edleeks87 What do you think?

edleeks87: Yeah that's fine we can do it all separately no problem

edleeks87 (dismissed):

edleeks87 (approved):

DipperTheDan (approved):

edleeks87 (approved):

DlgSHi (approved):

PR #4268, opened 2021-07-22
docs(typescript): rfc for converting carbon to ts

sjohnson-sage:
### Proposed behaviour
This PR contains an RFC for converting Carbon from JavaScript to TypeScript.

### Current behaviour
Carbon is currently a JavaScript library, in which we also provide TypeScript interfaces for components

### Checklist

### DIFF
--- a/rfcs/text/typescript.md
+++ b/rfcs/text/typescript.md
@@ -1,0 +1,259 @@
+- Start Date: 22/07/2021
+
+# Table of contents
+
+- [Summary](#summary)
+- [Basic example](#basic-example)
+- [Motivation](#motivation)
+- [Detailed design](#detailed-design)
+  - [Config](#config)
+    - [TypeScript config file](#typescript-config-file)
+    - [Babel config](#babel-config)
+    - [Packages](#packages)
+    - [Scripts](#scripts)
+    - [Eslint config](#eslint-config)
+  - [Components](#components)
+  - [Styled components](#styled-components)
+  - [Unit testing](#unit-testing)
+- [Drawbacks](#drawbacks)
+  - [Conversion effort](#conversion-effort)
+  - [Developer training](#developer-training)
+  - [TypeScript lifespan](#typescript-lifespan)
+- [Alternatives](#alternatives)
+  - [Doing nothing](#doing-nothing)
+  - [Removing TypeScript support](#removing-typescript-support)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+  - [Style guides](#style-guides)
+  - [Training](#training)
+  - [Documentation](#documentation)
+
+# Summary
+
+Currently Carbon is a JavaScript component library, in which we also provide TypeScript interfaces for our components.
+
+This is a duplication of effort and a large cause of issues raised against the library as the JavaScript/TypeScript prop types are often out of step with one another.
+
+This RFC will detail why and how we should convert Carbon from a JavaScript library, to a TypeScript one.
+
+# Basic example
+
+In the current js implementation of Carbon, a simple component will look like this (using the [`prop-types`](https://www.npmjs.com/package/prop-types) library):
+
+```jsx
+// my-component.component.js
+const MyComponent = ({ propOne, propTwo, ...rest }) => {
+  /* Component logic goes here */
+
+  return <MyStyledComponent>{/* Content */}</MyStyledComponent>;
+};
+
+MyComponent.propTypes = {
+  propOne: PropTypes.string.isRequired,
+  propTwo: PropTypes.boolean,
+};
+
+export default MyComponent;
+```
+
+In order for a ts project to consume this component we also provide a separate ts interface:
+
+```ts
+// my-component.d.ts
+export interface MyComponentProps {
+  propOne: string;
+  propTwo?: boolean;
+}
+
+declare function MyComponent(props: MyComponentProps): JSX.Element;
+
+export default MyComponent;
+```
+
+Converting the project to ts would result in the same component looking like this, in a single file and the prop types defined in one place:
+
+```tsx
+// my-component.component.tsx
+type MyComponentProps = {
+  propOne: string;
+  propTwo?: boolean;
+};
+
+const MyComponent: React.FC<MyComponentProps> = ({
+  propOne,
+  propTwo,
+  ...rest
+}: MyComponentProps) => {
+  /* Component logic goes here */
+
+  return <MyStyledComponent>{/* Content */}</MyStyledComponent>;
+};
+
+export default MyComponent;
+```
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


rfcs/text/typescript.md
```diff
@@ -0,0 +1,259 @@
...
+
+We may also need to add the [`@types/enzyme`](https://www.npmjs.com/package/@types/enzyme?activeTab=versions) module to add the type definitions for Enzyme.
+
+# Drawbacks
+
+## Conversion effort
+
+The first and main drawback is: it will likely be a large job to convert Carbon to TypeScript.
+
+Converting every single component, it's related styled components and spec files will take a while. Some may be simple, but some will be more complex, and running type checking against them may/will unearth bugs currently hidden. These will need to be fixed before the conversion will be complete. The flip side to that is we will likely find and fix bugs that are currently in our source code that we're not aware of.
```
edleeks87, line 217: Can this work be done piecemeal and once we convert a component file to `tsx` do we have to do the associated `spec` files (including ones that have a given component composed within another) at the same time?

rfcs/text/typescript.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Removing TypeScript support
+
+This is out of the question really as we know for a fact that a lot of consumers of Carbon are TypeScript projects. To be a good modern open source component library we should be supporting TypeScript.
+
+# Adoption strategy
+
+If this RFC PR is accepted we will need to create a task to carry out this conversion.
+
+Ideally the work will be carried out in a single PR (although it will be a large task!), with a commit per component, and once that is merged Carbon will be a TypeScript library.
```
edleeks87, line 241: Could this work be done piecemeal and once we convert a component file to tsx do we have to do the associated spec files (including ones that have a given component composed within another) at the same time?

rfcs/text/typescript.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Eslint config
+
+Converting to TypeScript means the parser in the eslint config should change to `@typescript-eslint/parser` rather than `babel-eslint`.
+
+A [Pull Request](https://github.com/Sage/carbon/pull/3978) is already in progress to add eslint config for TypeScript in Carbon. Once we convert the library to TypeScript we can simply replace the base config with the overrides specified in that config.
+
+## Components
+
+The the [Basic Example](#basic-example) section for how components will change when converted to TypeScript. The types of each prop will remain the same but defined only in TypeScript rather than both js and ts.
```
chrisbarber86, line 195:
```suggestion
The [Basic Example](#basic-example) section for how components will change when converted to TypeScript. The types of each prop will remain the same but defined only in TypeScript rather than both js and ts.
```

rfcs/text/typescript.md
```diff
@@ -0,0 +1,259 @@
...
+
+The types of constants defined within components can be inferred by TypeScript and there is no need to type them. The [TypeScript documenation](https://www.typescriptlang.org/docs/handbook/2/basic-types.html) states "it’s best not to add annotations when the type system would end up inferring the same type anyway".
```
chrisbarber86, line 197:
```suggestion
The types of constants defined within components can be inferred by TypeScript and there is no need to type them. The [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/2/basic-types.html) states "it’s best not to add annotations when the type system would end up inferring the same type anyway".
```

rfcs/text/typescript.md
chrisbarber86: If we could do this incrementally on a per component basis it would be much more preferable. Presumably the changes should be non breaking?

grabkowski (changes requested):

rfcs/text/typescript.md
```diff
@@ -0,0 +1,259 @@
...
+type MyComponentProps = {
+  propOne: string;
+  propTwo?: boolean;
+};
+
+const MyComponent: React.FC<MyComponentProps> = ({
+  propOne,
+  propTwo,
+  ...rest
+}: MyComponentProps) => {
```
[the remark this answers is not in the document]
sjohnson-sage: Good point, I copied that from one of the other Sage repos but yeah it isn't actually needed

rfcs/text/typescript.md
sjohnson-sage: @edleeks87 we don't have to but I think we should, we can have a Jira ticket per component to convert everything in that component folder to TypeScript. Everything that needs to be converted anyway

@chrisbarber86 yep it should be non breaking. I've updated the config detailed in the RFC to show how it can be set up to have js and ts components in parallel until all of them are converted (I've tested this in my test carbon-typescript repo as well). I've rewritten this bit to no longer recommend doing it in a single PR

chrisbarber86 (approved):

edleeks87 (approved):

rfcs/text/typescript.md
```diff
@@ -0,0 +1,256 @@
...
+
+This is a common occurrence and a significant amount of the issues raised on the Carbon Github page relate to either the prop types being out of step, or the ts types being incorrect. This is causing very preventable bugs and unnecessary extra maintenance for the Carbon team.
+
+A large amount of the more modern Sage repositories consuming Carbon are ts projects and it would make sense to align the technologies used, and upskill Carbon developers in using ts so they are more versatile in being able to help out with other projects.
+
+The [React documentation](https://reactjs.org/docs/static-type-checking.html) itself recommends using either Flow or TypeScript for larger code bases, which Carbon most definitely is.
+
+# Detailed design
+
+Carbon already uses TypeScript for component interface definition files and therefore some of the setup for the project is already there, however it will need some tweaking. The plan is to use Babel for transpiling the code, and the TypeScript compiler (`tsc`) for type checking and generating definition files.
```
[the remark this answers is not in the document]
sjohnson-sage: I believe I have in the Babel Config section below

chrisbarber86 (approved):

sjohnson-sage:
This has moved to the Final Comments phase, so if there are no major changes requested within the next 3 working days we will look to merge this

rfcs/text/typescript.md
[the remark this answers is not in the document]
jamime: Ah 🤦 thanks.

jamime (dismissed):

edleeks87 (dismissed):

mkrds:
@sjohnson-sage I might be missing something but I think we still need to automatically generate propTypes because by getting rid of them we'll remove runtime prop checking. That might not be a problem for projects that are using TypeScript, but the ones that use plain javascript won't have get any incorrect prop warnings.

mkrds (approved):

edleeks87 (approved):

grabkowski (dismissed):

edleeks87 (approved):

sjohnson-sage:
Moving to Final Comments phase after changes for @mkrds comment. If there are no major changes requested within the next 3 working days we will look to merge this

Issue #4262, opened 2021-07-21
Card is setting type and tabIndex when non-interactive


edleeks87:
### Current behaviour

The `Card` component sets the root element to `type="button"` and `tabIndex="0"` when the `interactive` prop is falsy. this means you can tab and focus the `Card` (but there is no visual feedback to say it is focused). 

### Expected behaviour

These props should only be set when `interactive` is true

### CodeSandbox or Storybook URL

https://carbon.sage.com/?path=/story/design-system-card--default-story

### Suggested Solution

```
 <StyledCard
      data-component="card"
      cardWidth={cardWidth}
      interactive={interactive}
      draggable={draggable}
      spacing={spacing}
      type={interactive ? "button" : undefined}
      onClick={onClickHandler}
      tabIndex={interactive ? 0 : undefined}
      data-role={dataRole}
      {...rest}
    >
      {draggable && <Icon type="drag" />}
      {renderChildren()}
    </StyledCard>
```

### Carbon Version

77.13.2

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

MacOS

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

jamime:
FE-4243

PR #4280, opened 2021-07-29
fix(card): wrong props on not-interactive card - FE-4243

dawidzarzycki:
### Proposed behaviour
Removing prop tabIndex and type when card is not interactive
 
fixes: #4262 

### Current behaviour
Props tabIndex and type are provided even when card is non interactive. It causes problem becase it allows users to tab into the card but there is no visual representation

### Checklist

### Testing instructions
Testable in chromatic storybook => Card => default


grabkowski (commented):
I think that the message should be:
`fix(card): incorrect props on non-interactive card`

or just:
`fix(card): incorrect props on a static card`

DipperTheDan (commented):
Code looks OK. I agree with @grabkowski that the commit message should be one of the two that has been suggested in the comment. Fix the commit message and I'll gladly approve 😃

src/components/card/card.component.js
```diff
@@ -57,17 +57,21 @@ const Card = ({
 
   const onClickHandler = interactive ? handleClick : null;
 
+  const interactiveProps = interactive && {
+    tabIndex: 0,
+    type: "button",
+  };
+
```
mmaciorowski-sage, line 64: I'm not sure if it matters in this case but ternary is more clear for me
Zrzut ekranu 2021-08-2 o 10 46 49

mmaciorowski-sage (dismissed):

src/components/card/card.component.js
dawidzarzycki: Thank you for your suggestion, anyway I would leave it as it is now. We don't need the `undefined` as we don't want to do anything with it. All we need is just `tabIndex` and `type` prop if the card is interactive. I think this is clear enough. We use this approach in our project very often

dawidzarzycki: Thank you for your suggestion. If the interactive will be `false` or `undefined` it won't change anything as none of these props will be shown in DOM. Anyway I think that undefined approach is cool. I'll change it

DipperTheDan (approved):

mmaciorowski-sage (approved):

DlgSHi (approved):

PR #4294, opened 2021-08-09
docs(date): rfc proposing new date component

edleeks87:
### Proposed behaviour

RFC detailing a proposal for a new implementation for the `Date` component

### Current behaviour

Existing implementation has considerable technical debt and issues and pending features

### Checklist

### Additional context

POC (*very* rough implementation):
https://codesandbox.io/s/poc-of-rfc-proposal-ctdfw?file=/src/index.js

### DIFF
--- a/rfcs/text/refactoring-date-component.md
+++ b/rfcs/text/refactoring-date-component.md
@@ -1,0 +1,399 @@
+- Start Date: (2021-07-28)
+
+# Table of contents
+
+- [Summary](#summary)
+- [Basic example](#basic-example)
+- [Motivation](#motivation)
+- [Detailed design](#detailed-design)
+  - [Replacing Momentjs](#replacing-momentjs)
+  - [Continuing to use React DayPicker](#continuing-to-use-react-daypicker)
+  - [Locales](#locales)
+  - [Event handling](#event-handling)
+  - [Support for hidden input](#support-for-hidden-input)
+  - [DatePicker component](#datepicker-component)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+# Summary
+
+The purpose of this document is to outline a new simplified `Date` component implementation, removing any technical debt or logic that is specifically tied to any consuming project whilst supporting consumers in being able to implement their requirements in their own codebases.
+
+# Basic example
+
+The new implementation will be written as a functional component and only support being used as a controlled input: it will require the implementing developer to provide `value` and `onChange` props. When the component is blurred the input value will reformat the string based on the locale supplied via an `I18nProvider`. This will allow implementation teams to implement the component as follows:
+
+```jsx
+  // my-component.component.js
+  const MyComponent = () => {
+    const [value, setValue] = useState("28/07/1987");
+
+    const handleChange = (event) => {
+      const { formattedValue } = event.target.value;
+      setValue(formattedValue);
+    };
+
+    const handleBlur = (event) => {...}
+
+    return (
+      <I18nProvider>
+        <Date
+          value={value}
+          onChange={handleChange}
+          onBlur={handleBlur}
+        />
+      </I18nProvider>
+    );
+  };
+```
+
+The remaining props intended for the input will be opt-in or have a default value. The component will also continue to support the styling props (such as `size` and spacing props) that the existing one does. Furthermore, it will continue to support the same validation interface to ensure that it can still be integrated with `formik` or whatever form library implementation teams choose to use.
+
+```js
+  inputProps = {
+    onFocus,
+    disabled = false,
+    readOnly = false,
+    autoFocus = false,
+    size = "medium",
+    labelHelp,
+    labelAlign,
+    labelInline,
+    labelSpacing,
+    labelWidth,
+    inputWidth,
+    fieldHelp,
+    error,
+    warning,
+    info,
+    placeholder,
+    required,
+    allowEmptyValue = false,
+    ...marginSpacingProps,
+  }
+```
+
+# Motivation
+
+The current component within `carbon-react` has a significant amount of technical debt and has a lot of logic specific to a project that was previously its main consumer. For example, the component supports being used as both as a controlled and uncontrolled component which again adds to the complexity. The current implementation is also tied to using `moment` to manage the parsing and formatting of date values. However, the maintainers of the [project](https://momentjs.com/docs/#/-project-status/) have previously labelled it as a legacy project and as such will no longer be releasing new features and it will only be maintained up to the point of addressing critical security concerns. Whilst this is not necessarily something that will have a significant impact on all our consumers, we should look to implement a `Date` component that does not rely on a legacy solution as there are several lightweight alternatives. Carbon currently exports an extensive set of [date utils](https://github.com/Sage/carbon/blob/master/src/utils/helpers/date/date.js) all of which use `moment` and support a large amount of date formats for all locales which adds to the complexity of the component.  
+
+As the number of consumers of the library has grown this has led to a lot of [issues](https://github.com/Sage/carbon/issues?q=is%3Aissue+is%3Aopen+date) being raised against the component. Some of the code was written over four years ago, and the addition of some newer features during this time have led to the code for the component being significantly bloated and therefore difficult to maintain.
+
+# Detailed design
+
+The date passed to the picker component can be updated through a `useEffect` hook that checks the current value in the input is both a valid format (based on the value recieved from the locale context) and parses to a valid Date. As such the component will be able to maintain a significantly reduced amount of internal state for the `selectedDate` and the `open` state for the picker component. The `DatePicker` will be baked into the `Date` component and we will surface the interface defined in the [react-day-picker](https://react-day-picker.js.org/api/DayPicker) docs via a `pickerProps` object and as such it can be moved to an `__internal__` directory: we will also include a link to `react-day-picker` documentations in our prop tables.
+
+```jsx
+  // src/components/date/date.component.js
+  import {
+    formatToString,
+    isValidFormat,
+    parseToDate,
+    isValidDate,
+    formatISOString,
+    localeMap,
+  } from "./__internal__/date-utils"
+
+  const DateInput = ({ value, onChange, onBlur, onFocus, inputProps, pickerProps, ...rest }) => {
+    const locale = useContext(LocaleContext);
+    const { locale } = locale;
+    const { localize, options, formats } = localeMap[locale()];
+    const format = formats.javascript();
+    const formats = formats.input();
+    const [selectedDate, setSelectedDate] = useState(parseToDate(format));
+    const [open, setOpen] = useState(false));
+
+    useEffect(() => {
+      // if value string is valid format and parses to valid date setSelectedDate to parsed value
+    }, [value]);
+
+    return (
+      <div {...rest}>
+        <Textbox
+          value={value}
+          onChange={handleChange}
+          onBlur={handleBlur}
+          inputIcon="calendar"
+          iconOnClick={handleIconClick}
+          onClick={handleOnClickHandler}
+          {...inputProps}
+        />
+        <DatePicker
+          {...pickerProps}
+          value={value}
+          selectedDate={selectedDate}
+          open={open}
+          onDayClick={onDayClick}
+        />
+      </div>
+    );
+  };
+
+  DateInput.propTypes = {
+    value: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    onBlur: PropTypes.func,
+    ...inputPropsTypes,
+    ...pickerPropsTypes,
+  };
+```
+
+## Replacing Momentjs
+
+As mentioned previously the `moment` project has been marked as legacy. The maintainers have documented several alternatives of which `date-fns` seems best suited for our implementation needs. It is fast, lightweight and modular, allowing us to only import the functions we need rather than the whole package of functions. It uses the existing native JavaScript date API and the functions it provides for parsing and formatting dates are immutable and pure: returning new instances rather than modifying existing parsed dates (limiting any bugs that could arise). The adoption of `date-fns` is widespread, many component libraries either use it directly or provide an abstraction with it as the default option. We will deprecate the utils that we currently export for date formatting and parsing at the same time as rewriting the component after ensuring an appropriate warning has been added in advance. Removing the utils will help streamline the codebase and move Carbon closer to being a pure UI component library.
+
+## Continuing to use React DayPicker
+
+Research into whether we should migrate to a new solution or continue to use `react-day-picker` revealed that the library is still widely adopted and actively maintained when [comparing](https://react.libhunt.com/react-day-picker-alternatives) it to alternatives. Furthermore, `react-day-picker` has moved away from using `moment` in `v7` in favour of `date-fns` which would mean there will be consistency between the library Carbon uses and the one the calendar picker uses. Therefore, it is reasonable to continue to use this library and look to update it to at least `v7` at the same time as rewriting the date component.
+
+## Locales
+
+We will remove the `date` property from the locale object passed to the `I18nProvider` component: instead we will store the formats internally and use the `locale.locale()` string to map to the formats for that given locale. The rest of the locale object must match the interface detailed in the [i18n RFC](https://github.com/Sage/carbon/blob/master/rfcs/text/i18n.md) documenting the approach. Storing the formats internally and separating valid formats by locale will lessen the amount of maintenance the component will require.
+
+To implement translations in the new component we will leverage the locales already provided by [date-fns/locale](https://github.com/date-fns/date-fns/tree/master/src/locale), whereby the locale string can be used to map to the relevant file. The [localize](https://date-fns.org/v2.23.0/docs/I18n-Contribution-Guide#localize) properties surfaced from `date-fns` support translations for weekdays, months and so on out of the box. Whilst this would couple our implementation closely to `date-fns` if we decided later we needed to move away from it implementing a different mechanism would not be too complex.
+
+## Event handling
+
+The event emitted `onBlur` and `onChange` should meet the specification defined in the [onChange interface RFC](https://github.com/Sage/carbon/blob/master/rfcs/text/onChange.md) and that is also found in the current component. This will mean implementation teams will have access to both a `formattedValue` and `rawValue`: if the `event.type` is blur the `formattedValue` should be set to the reformatted value based on the selected date of the `DatePicker` which will be stored in state. If the `event.type` is anything else the `formattedValue` should be the `event.target.value`. The `rawValue` will be the ISO formatted date string or null if the input value does not parse to a valid date. This will allow the value to be stored in the backend database.
+
+```js
+  const buildEventObject = (event) => {
+    const { id, name, value } = event.target;
+
+    const parsedInputValue = parse(new Date(), format, value);
+
+    const formattedValueString = event.type === "blur" ? formatToString(selectedDate, format) : value;
+    const rawValueString = isDateValid(parseInputValue) ? formatISO(parsedInputValue) : null;
+
+    event.target = {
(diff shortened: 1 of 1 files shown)


rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+
+There are a number of date management solutions available as an alternative to `date-fns`: [Luxon](https://moment.github.io/luxon/); [Dayjs](https://day.js.org/); [js-Joda](https://js-joda.github.io/js-joda/); or using the Date object directly instead or a library. An additional alternative would be to integrate an abstraction for any of these potential solutions, for example [date-io](https://github.com/dmtrKovalenko/date-io) would support this. Of these options, `date-fns` is the most lightweight and will require the least integration into the new component. It is either the chosen or default option for many other UI libraries. For our requirements, it is unlikely we will need to surface an abstraction at this point as we will not be requiring implementation teams to do any configuration past what is defined in the interface and what they will need to pass in via the locale. If the need to introduce an abstraction arose it would not be too complex to refactor it to support this without introducing a breaking change.
```
sjohnson-sage, line 312:
```suggestion
There are a number of date management solutions available as an alternative to `date-fns`: [Luxon](https://moment.github.io/luxon/); [Dayjs](https://day.js.org/); [js-Joda](https://js-joda.github.io/js-joda/); or using the Date object directly instead of a library. An additional alternative would be to integrate an abstraction for any of these potential solutions, for example [date-io](https://github.com/dmtrKovalenko/date-io) would support this. Of these options, `date-fns` is the most lightweight and will require the least integration into the new component. It is either the chosen or default option for many other UI libraries. For our requirements, it is unlikely we will need to surface an abstraction at this point as we will not be requiring implementation teams to do any configuration past what is defined in the interface and what they will need to pass in via the locale. If the need to introduce an abstraction arose it would not be too complex to refactor it to support this without introducing a breaking change.
```

sjohnson-sage (commented):
The approach looks good to me in general, mostly minor comments/changes from me

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+# Basic example
+
+The new implementation will be written as a functional component and only support being used as a controlled input: it will require the implementing developer to provide `value` and `onChange` props. The `onBlur` prop will also be required so the component can handle reformatting the value string based on the locale supplied via an `I18nProvider`. This will allow implementation teams to implement the component as follows:
+
+```jsx
+  // my-component.component.js
+  const MyComponent = () => {
+    const [value, setValue] = useState("28/07/1987");
+
+    const updateDateValue = (ev) => {
```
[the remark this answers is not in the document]
edleeks87: I really should have double proof read the code snippets as well 🤦 , thanks will update when I've had a few more 👀  on this

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+    isValidDate,
+    formatISOString
+  } from "./__internal__/date-utils"
+
+  const DateInput = ({ value, onChange, onBlur, onFocus, inputProps, ...rest }) => {
+    const locale = useContext(LocaleContext);
+    const { date } = locale;
+    const format = date.formats.javascript();
+    const formats = date.formats.input();
+    const [selectedDate, setSelectedDate] = useState(parseToDate(value, formats.javascript()));
```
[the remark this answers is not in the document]
edleeks87: Yeah it can thanks

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+  const MyComponent = () => {
+    const [value, setValue] = useState("28/07/1987");
+
+    const updateDateValue = (ev) => {
+      const { formattedValue } = ev.target.value;
+      setVisibleValue(formattedValue);
+    };
+
+    return (
+      <I18nProvider>
```
mmaciorowski-sage, line 40: By default carbon uses `en-GB` translations so you don't need to wrap it in `I18nProvider` for default.
https://carbon.sage.com/?path=/story/documentation-i18n--page#default

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+    parseToDate,
+    isValidDate,
+    formatISOString
+  } from "./__internal__/date-utils"
+
+  const DateInput = ({ value, onChange, onBlur, onFocus, inputProps, ...rest }) => {
+    const locale = useContext(LocaleContext);
+    const { date } = locale;
+    const format = date.formats.javascript();
+    const formats = date.formats.input();
```
mmaciorowski-sage, line 101: Suggestion, not a blocker.
In carbon we would write:
```
const l = useContext(LocaleContext);
```
and just use like
```
l.date.formats.javascript()
```

edleeks87: Yeah I just had it structured like this for readability here

rfcs/text/refactoring-date-component.md
edleeks87: Yeah this is just for the sake of being readable and making it clear how things would work logically, I'll probably expand the example to include a custom locale object once the first round of reviews are done

jamime (commented):
- [ ] Update RFC to use composition for date picker e.g. `<Date picker={<Picker />} />`

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,330 @@
...
+        />
+      </StyledDayPicker>
+    </Popover>
+    );
+  };
+```
+
+# Drawbacks
+
+Carrying out this proposal will mean introducing several breaking changes. Writing the new component as a functional component will mean that any consuming project that has extended the previous class-based component will now need to update to use composition instead. We will also be removing the utils we currently export and any new ones we write will be in an `__internal__` directory. By moving the `DatePicker`, `Weekday` and `Navbar` components to an `__internal__` directory we will also be making a breaking change, but this will be for the better in the long run as these components should not be directly imported by consuming projects and we can then make changes to their interfaces without needing to have them be breaking. The `Date` component is used extensively by our consumers so these breaking changes are likely to have widespread impact. Although it is certainly arguable that the current implementation has enough issues with it that this too is having significant impact.
```
chrisbarber86, line 298: Should we stage the breaking changes? E.g. remove the shared utils, make the components `__internal__` then rewrite date?

edleeks87: Good idea, I think we could do this as two breaking changes, move the utils and components to `__internal__` and then carry out the rewrite of `Date`.

rfcs/text/refactoring-date-component.md
edleeks87: Decided to remove it as it's redundant in this example

rfcs/text/refactoring-date-component.md
edleeks87: @chrisbarber86 I'd missed that we're actually already moving all our utils to an `__internal__` directory as part of separate ticket so the work will be carried out as two breaking changes, I'll update the document to reflect this

rfcs/text/refactoring-date-component.md
```diff
@@ -94,12 +96,12 @@ The date passed to the picker component can be updated through a `useEffect` hoo
     formatISOString
   } from "./__internal__/date-utils"
 
-  const DateInput = ({ value, onChange, onBlur, onFocus, inputProps, ...rest }) => {
+  const DateInput = ({ value, onChange, onBlur, onFocus, inputProps, picker, ...rest }) => {
```
jamime, line 99: The convention is typically `renderPicker`

rfcs/text/refactoring-date-component.md
```diff
@@ -195,36 +190,39 @@ Whilst we will continue to support the same custom event object, we will remove
...
+We will continue to use the same sub-components such as `Navbar` and `Weekday`, although they should be moved to an `__internal__` directory of the `Picker` component as they will never need to be directly imported by consuming projects. The `containerProps` prop provides an interface for implementing keyboard accessibility within the picker (`tabIndex`, `onKeyDown` and so on). The [locale interface](https://react-day-picker.js.org/docs/localization) surfaced by `react-day-picker` will be integrated with `date-fns/locale` to override the translations strings. Below is an example of how translating weekdays can be achieved using the locales surfaced from `date-fns/locale`. It is also possible to generate an array of `months` based on the locale string.
```
jamime, line 225: It's not clear why we need a `Picker` component if we're using a render prop. Is this so we can provide our own components that `react-day-picker` then displays?

rfcs/text/refactoring-date-component.md
edleeks87: Yeah good point I'll update it to that

rfcs/text/refactoring-date-component.md
edleeks87: Yeah my intention was to surface our `Picker` which will be composed of components that are then passed into the `react-day-picker`, it will also allow us to include the translations and locale implementation as part of this component rather than having the logic in the main `Date` component and passing it down as props to the `Picker`. Of course having it as a render prop will mean someone could pass their own picker (even using an alternative to `react-day-picker` if they want) in as long as it matches the interface the input expects, this will mean they can deliver any features we don't want to add or not to have a picker at all and just have the input. I still envision that most of the time implementation teams will be using a `Picker` that is similar to the one here so think it's right that we export one to avoid forcing teams to write the same code and install the dependencies required for it etc.

mkrds:
Looks good to me, I'm just not sure about the `renderPicker` approach, if there are use cases for that than I'm ok with that

jamime (commented):
### Meeting Notes 30th Sept
Attendees: @edleeks87 @mkrds

- Change `renderPicker` to be `pickerProps` and pass props to `react-day-picker`
- Link to `react-day-picker` documentation
- Add notes on [`captionElement`](https://react-day-picker.js.org/examples/elements-year-navigation). We will add additional prop e.g. `displayInputCaption` which will add our `Select` components for switching the month/year.

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+          value={hiddenValue}
+        />
+      </>
+    );
+  };
+```
+
+## DatePicker component
+
+We will continue to use the same sub-components such as `Navbar` and `Weekday`, although they should be moved to an `__internal__` directory of the `DatePicker` component as they will never need to be directly imported by consuming projects. The `containerProps` prop provides an interface for implementing keyboard accessibility within the picker (`tabIndex`, `onKeyDown` and so on). The [locale interface](https://react-day-picker.js.org/docs/localization) surfaced by `react-day-picker` will be integrated with `date-fns/locale` to override the translations strings. Below is an example of how translating weekdays can be achieved using the locales surfaced from `date-fns/locale`. It is also possible to generate an array of `months` based on the locale string.
```
jamime, line 237:
```suggestion
We will continue to use the same sub-components such as `Navbar` and `Weekday`, although they should be moved to an `__internal__` directory of the `DatePicker` component as they will never need to be directly imported by consuming projects. The [`containerProps` prop](https://react-day-picker.js.org/api/DayPicker/#containerProps) provides an interface for implementing keyboard accessibility within the picker (`tabIndex`, `onKeyDown` and so on). The [locale interface](https://react-day-picker.js.org/docs/localization) surfaced by `react-day-picker` will be integrated with `date-fns/locale` to override the translations strings. Below is an example of how translating weekdays can be achieved using the locales surfaced from `date-fns/locale`. It is also possible to generate an array of `months` based on the locale string.
```

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+      </Weekday>
+    );
+  }}
+```
+
+```js
+  const months = Array.from({ length: 12 }).map((_, i) => localize.month(i))
+```
+
+As with the existing component we will store a `lastValidValue` in internal state and use that to control the date displayed in the `DatePicker`. This can be achieved by leveraging a `useEffect` hook to update the state value if a new `inputValue` parses to a valid Date. We will also keep the same implementation the current component has with regards to rendering the `DatePicker` in a `portal` using the existing `Popover` component. To support additional features we will expose the `react-day-picker` interface via the `pickerProps`. For example, the following [feature request](https://github.com/Sage/carbon/issues/4275) can be supported this by utilising the [captionElement](https://react-day-picker.js.org/examples/elements-year-navigation) render prop, although we would implement this via a boolean prop and handle the the composition internally ustilising Carbon's existing `Select` component: below is an example of how this can be achieved.
```
jamime, line 257: If the user enters an invalid date I think we should reset the `react-day-picker` state, it shouldn't show a selection.

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+
+The new proposed component will be tightly coupled to `date-fns` meaning if a better alternative becomes available, we will have a harder job migrating to it. However, writing our own internal utils that use the underlying `date-fns` interface should mitigate some of this work. Furthermore, if any issues arise with it we are reliant on support from the maintainers of `date-fns`, although it is widely adopted and maintained at present. The library uses different [Unicode patterns](https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md) for date formats so we will need to ensure these differences are communicated to implementation teams.
```
jamime, line 381:
```suggestion
The new proposed component will be tightly coupled to `date-fns` meaning if a better alternative becomes available, we will have a harder job migrating to it. However, writing our own internal utils that use the underlying `date-fns` interface should mitigate some of this work. Furthermore, if any issues arise with it we are reliant on support from the maintainers of `date-fns`, although it is widely adopted and maintained at present.
```

We don't need to communicate this as we're only using the `I18nProvider#locale()`

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+
+Currently `react-day-picker` does not support styled-components or any other CSS-in-JS so we will have to continue to apply the styles as is currently implemented in the existing component. This has the potential to lock us to a specific version or at the very least create potential pain points when we do upgrade. However, the CSS classes we would be hooking into are part of the defined interface the package exposes to support customising styles and has not changed from the version we are currently using `v6.1.1` to the latest available one `v7.10.0` so we will likely have advanced warning of any changes to them.
```
jamime, line 383: >  This has the potential to lock us to a specific version 

I don't think this would lock us to a version. We'd update the styles each time we update the react-day-picker.

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+
+Similarly there are a number of other options for integrating within the `DatePicker`, for example [react-datepicker](https://reactdatepicker.com/) and [react-flatpickr](https://github.com/haoxins/react-flatpickr). However, they all come with limitations: a lack of documentation, no support for CSS-in-JS and more configuration requirements with respect to locale support.
+
+# Adoption strategy
+
+As mentioned, this proposal will be implemented as a breaking change for the significant technical debt to be removed and to fix the existing bugs we have raised against the current implementation. In terms of how consuming projects will migrate to it, the majority of the interface should remain the same so the bulk of the work will be on adding the correct locale properties, so the desired formats are provided to the component.
```
jamime, line 399: >  adding the correct locale properties, so the desired formats are provided to the component

This is no longer true.

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,407 @@
...
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+# Summary
+
+The purpose of this document is to outline a new simplified `Date` component implementation, removing any technical debt or logic that is specifically tied to any consuming project whilst supporting consumers in being able to implement their requirements in their own codebases.
+
+# Basic example
+
+The new implementation will be written as a functional component and only support being used as a controlled input: it will require the implementing developer to provide `value` and `onChange` props. The `onBlur` prop will also be required so the component can handle reformatting the value string based on the locale supplied via an `I18nProvider`. This will allow implementation teams to implement the component as follows:
```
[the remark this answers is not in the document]
edleeks87: It doesn't have to be, I think I had this in before I moved to trigger the onChange when the blur event occurred, I'll remove it now

rfcs/text/refactoring-date-component.md
edleeks87: Yeah that's fair enough, I didn't really have a strong opinion on how this should behave so went with how our current component behaves. It's easy enough to add so I'm happy to implement it as it could help avoid some weird behaviour

rfcs/text/refactoring-date-component.md
edleeks87: Yeah I should have said it makes the upgrade more complicated, but isn't a blocker

rfcs/text/refactoring-date-component.md
edleeks87: Again should have removed this 🤦

edleeks87: Without the guard it throws just after mount as it's setting the `selectedValue` to undefined on the next render

jamime (approved):

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,398 @@
...
+      </div>
+    );
+  };
+
+  DateInput.propTypes = {
+    value: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    onBlur: PropTypes.func.isRequired,
+    ...inputProps
+  };
```
samtjo, line 139: `...pickerProps` should be here if I've read the paragraph above right

edleeks87: 🤦  yep I should have added them to the prop types

edleeks87: `onBlur` should not be required as well and it should be `inputPropTypes` really so I've included `pickerPropsTypes`

jamime (approved):

samtjo (approved):

chrisbarber86 (dismissed):

edleeks87:
PR has moved into final comments stage

rfcs/text/refactoring-date-component.md
```diff
@@ -0,0 +1,399 @@
...
+    localeMap,
+    isValidDate,
+    parseToDate
+  } from "./__internal__/date-utils";
+
+  const DatePicker = ({
+    inputValue,
+    selectedDate,
+    setSelectedDate,
+    onDayClick,
```
damianOfKyan, line 320: So I understand that you can either:
1. Select Date using Picker and your mouse
2. Type in the date and ignore the picker

Shouldn't DatePicker have `onChange` as `onDayClick` is not sufficient?

edleeks87: Yes so the typical pattern would be what you've described, we also intend to make the picker keyboard accessible as well so a user will be able to select a date using tabbing, arrow keys and then enter to select the date. There's no need to pass an onChange to the picker for this to work but the `containerProps` interface supports passing the event callbacks if we want. I've updated the POC to include an example of the keyboard support working

https://codesandbox.io/s/poc-of-rfc-proposal-ctdfw?file=/src/index.js:3195-3197

damianOfKyan: sweet, sounds great. Great work guys, I know first hand what this Draft means as I was "suffering" while implementing an old Carbon DateInput component. This proposed changes are very well thought through!

edleeks87: Thanks for the feedback and taking the time to read it @damianOfKyan 😄

samtjo (approved):

jamime (approved):

quinnturner:
Considering alternatives to `moment`, I'd like to add another point to the discussion; migration workload for downstream projects.

Since `carbon-react` is the last project that uses `moment` within our bundle, we will probably end up going with whatever `carbon-react` chooses to minimize our bundle size. `dayjs` generally has a smoother transition from `moment` than `date-fns` because `dayjs` is closer to being API-compatible than `date-fns` (after using a few dayjs plugins). Therefore, it may be less work from each dependent project on `carbon-react`, that wishes to use the same date lib, to upgrade as well.

PR #4313, opened 2021-08-17
test: add react test

DlgSHi:
### Proposed behaviour

We need to try and implement tests that won't depend on storybook, and use `components` as they are used by `end User`

Implement new `cypress` approach for `react` testing by `mount`'ing `component` and passing `props` into 

```Javascript
<Component `props`> 
```

like: 

```Javascript
mount(<Component readOnly/>);
```

Add possibility to check all events (with `Object` properties check)

```Javascript
// use cy.stub for mock the onChange callback
const callback = cy.stub();
// and mount the story using @cypress/react library
mount(<Decimal onChange={callback} />);
```

and then check the `onChange` event `Object` like:
```Javascript
.then(() => {
    // eslint-disable-next-line no-unused-expressions
    expect(callback).to.have.been.calledThrice;
     iterable.map((item, index) => {
         expect(callback.getCalls()[index].args[0].target.value.rawValue).to.equals(item[0]);
         expect(callback.getCalls()[index].args[0].target.value.formattedValue).to.equals(item[1]);
         return this;
         });
      });
```

Check: 
- `src/components/app-wrapper/app-wrapper.test.js`
- `src/components/decimal/decimal.test.js`

the time for running `app-wrapper.test.js` and `decimal.test.js` after refactor:
Screenshot 2021-10-21 at 12 24 11

### Current behaviour

We need to start storybook to test components

Currently the time for running `appWrapper.feature`, `decimalInput.feature`, `decimalPrecision.feature`,
`decimalInput.feature` is around 1:40 min ~ 2:00 mins + time from starting `storybook`.

Screenshot 2021-08-25 at 08 20 54

### Checklist

### Testing instructions

`run:`
- npm install
- npm run cypress:react


package.json
```diff
@@ -133,16 +137,19 @@
...
   "dependencies": {
     "@popperjs/core": "^2.9.0",
+    "@storybook/testing-react": "0.0.16",
     "@styled-system/prop-types": "^5.1.5",
     "@tippyjs/react": "^4.2.5",
     "axios": "^0.21.1",
     "bignumber.js": "^9.0.1",
     "bowser": "~1.5.0",
     "classnames": "~2.2.6",
+    "create-cypress-tests": "^1.1.1",
```
jamime, line 152:
```suggestion
```

src/components/decimal/decimal.test.js
```diff
@@ -0,0 +1,302 @@
...
+        [14, "2.1", "2.10000000000000"],
+        [14, "2.12", "2.12000000000000"],
+        [14, "2.123", "2.12300000000000"],
+        [14, "1222.1234", "1,222.12340000000000"],
+
+        [15, "1", "1.000000000000000"],
+        [15, "2332.78", "2,332.780000000000000"],
+        [15, "1a3.55", "1a3.55"],
+        [15, "1.12345", "1.123450000000000"],
+        [15, "1a.23", "1a.23"],
```
jamime, line 94: IMO this should use `it.each`, it will provider better test results if one failed.

src/components/decimal/decimal.test.js
```diff
@@ -0,0 +1,302 @@
...
+      cy.get(decimalComponent)
+        .find("input")
+        .should("have.attr", "value", "   ");
+    });
+  });
+
+  describe("check events for Decimal component", () => {
+    it("should return onChange event", () => {
+      // use cy.stub for mock the onChange callback
+      const callback = cy.stub().as("eventCallback");
```
jamime, line 244: Why does this have a `eventCallback`, the other stubs do not.

cypress/support/helper.js
```diff
@@ -22,7 +22,10 @@ export function visitComponentUrl(
   prefix = "",
   env = "story"
 ) {
-  cy.visit(prepareUrl(component, suffix, iFrameOnly, prefix, env));
+  const spy = cy.spy();
+  cy.visit(prepareUrl(component, suffix, iFrameOnly, prefix, env), {
+    onBeforeLoad: spy,
```
jamime, line 27: Why?

cypress/support/index.js
```diff
@@ -6,32 +6,13 @@ require("cypress-plugin-tab");
...
-// 'supportFile' configuration option.
-//
-// You can read more here:
-// https://on.cypress.io/configuration
-// ***********************************************************
-
-// Import commands.js using ES2015 syntax:
-// import './commands';
+export default DEBUG_FLAG;
+Cypress.Screenshot.defaults({ screenshotOnRunFailure: DEBUG_FLAG });
```
jamime, line 10: We could always make this `true` when running in CI so they show in the cypress dashboard e.g.

```js
const DEBUG_FLAG = process.env.CI || false;
```

package.json
```diff
@@ -59,6 +61,8 @@
     "@babel/preset-typescript": "7.12.7",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
+    "@cypress/react": "^5.9.2",
+    "@cypress/webpack-dev-server": "^1.4.0",
```
jamime, line 65: Is this used?

DlgSHi: yes, it is used :) 
1. import { mount } from "@cypress/react"; for `mount()`'ing;
2. Webpack is used for proper component testing

cypress/support/helper.js
DlgSHi: because it wasn't working for both (`BDD`(actual approach) and `component testing`(future approach))

src/components/decimal/decimal.test.js
DlgSHi: just a test to make possible to use
`cy.get(@ eventCallback)` - will delete this :)

sianford (dismissed):

PR #4315, opened 2021-08-18
refactor(link): convert a class component to a function component - FE-3654

mmaciorowski-sage:
### Proposed behaviour
- refactor component to be functional
- breaking change to remove get functions
- remove InternalLink export
- remove innerRef prop as useRef can be used instead
- remove className, change to styled component and target that within pod.style

### Current behaviour
class component

### Checklist

### Testing instructions

https://codesandbox.io/s/carbon-quickstart-forked-nnbxe?file=/src/index.js


nicktitchmarsh (dismissed):

src/components/link/link.style.js
```diff
@@ -78,15 +78,21 @@ const LinkStyle = styled.span`
...
 };
 
-LinkStyle.propTypes = {
+StyledLink.propTypes = {
   disabled: PropTypes.bool,
   hasContent: PropTypes.bool,
 };
 
-export default LinkStyle;
+const StyledContent = styled.span``;
```
sjohnson-sage, line 92: This can be removed, there's no need for an empty styled component, can just use a `span` instead

src/components/pod/pod.style.js
```diff
@@ -3,6 +3,7 @@ import { margin } from "styled-system";
 
 import { baseTheme } from "../../style/themes";
 import Link from "../link";
+import { StyledContent as StyledLinContent } from "../link/link.style";
```
sjohnson-sage, line 6:
```suggestion
import { StyledContent as StyledLinkContent } from "../link/link.style";
```

src/components/pod/pod.style.js
```diff
@@ -238,7 +239,7 @@ const StyledEditAction = styled(Link)`
       })}
   }
 
-  .carbon-link__content {
+  ${StyledLinContent} {
```
sjohnson-sage, line 242:
```suggestion
  ${StyledLinkContent} {
```

src/components/link/link.component.js
```diff
@@ -2,131 +2,126 @@ import React from "react";
...
+          iconColor="business-color"
+          disabled={disabled}
+          ariaLabel={ariaLabel}
+          tooltipMessage={tooltipMessage}
+          tooltipPosition={tooltipPosition}
+        />
+      ) : null;
+    };
+    const handleClick = (ev) => {
+      /* istanbul ignore if */
```
sjohnson-sage, line 63: I think we need to be testing this rather than skipping coverage here

sjohnson-sage (changes requested):
Looks good! Just a few things to change from me 👍🏻

src/components/link/link.component.js
mmaciorowski-sage: I have no idea how I can test it. I tried to simulate click and mocked event, but there is no way to simulate click if button is disabled. I've searched for force click but cannot find any way to accomplish this. Do you have any suggestions?

sjohnson-sage: So in the browser if it's disabled, does it not trigger this handleClick at all? If so we can just remove the check here

mmaciorowski-sage: probably we can remove it, I'll confirm that, thanks

src/components/pod/pod.style.js
mmaciorowski-sage: thanks

src/components/link/link.style.js
mmaciorowski-sage: In general yes, but because we remove className this is the clearest way to have a link content selector that we use in pod.style.js

sjohnson-sage: Ok fair enough, I see why you've done that now 👍🏻

src/components/link/link.style.js
```diff
@@ -78,15 +78,21 @@ const LinkStyle = styled.span`
...
+
+StyledContent.defaultProps = {
```
sjohnson-sage, line 94: This doesn't need the theme prop as it isn't used anywhere at the moment, you can just remove this `defaultProps` altogether

mmaciorowski-sage: good point, changed, thanks

sjohnson-sage (approved):

nicktitchmarsh (approved):

DlgSHi (approved):

sjohnson-sage (approved):

PR #4323, opened 2021-08-23
fix: added missing null check to element resize

blaky:
The form component uses a throttle function that triggers a call to ElementResize's `removeListener` function. By the time that the function is executed, the element might be removed from the DOM which results an unhandled error.

### Proposed behaviour
ElementResize's `removeListener` function should not throw an error if it is called with a `null` element. This can happen when the `Form` component is placed in a dialog, which gets removed from the DOM.

### Current behaviour
The empty element parameter scenario is not handled in the `removeListener` function (although it is in the `addListener` function) and a "null-pointer" exception is thrown when it is called with a `null` element argument.

### Checklist

- N/A Screenshots are included in the PR if useful
- N/A All themes are supported if required
- N/A Cypress automation tests added or updated if required
- N/A Storybook added or updated if required
- N/A Typescript `d.ts` file added or updated if required
- N/A Carbon implementation and Design System documentation are congruent

### Testing instructions
It is difficult to manually test, I couldn't isolate the issue into a codesandbox repository.

### DIFF
--- a/src/utils/helpers/element-resize/__spec__.js
+++ b/src/utils/helpers/element-resize/__spec__.js
@@ -16,6 +16,11 @@ describe("ElementResize", () => {
         ElementResize.addListener(null, callback);
         expect(element.__resizeListenerCallbacks__).toEqual(undefined);
       });
+      it("does nothing on removal", () => {
+        expect(() =>
+          ElementResize.removeListener(null, callback)
+        ).not.toThrow();
+      });
     });
     describe("no listener yet set", () => {
       it("creates an object to watch", () => {
--- a/src/utils/helpers/element-resize/element-resize.js
+++ b/src/utils/helpers/element-resize/element-resize.js
@@ -82,6 +82,10 @@ const ElementResize = {
    * @param {Function} fn - the callback
    */
   removeListener: (element, callback) => {
+    if (!element) {
+      return;
+    }
+
     if (element.__resizeListenerCallbacks__) {
       // remove the event listener from the array
       element.__resizeListenerCallbacks__.splice(


jamime (changes requested):
Hi @blaky,

We have a detailed [pull request template](https://raw.githubusercontent.com/Sage/carbon/f23993ab5c939abcecbe4065409217f50a4eead5/.github/PULL_REQUEST_TEMPLATE.md) that would help us get this change approved faster.
In particular for QA we need a reproducible example in form of a codesandbox that shows the broken behaviour.

Our CI process then builds a new version and QA can then confirm that the bug is indeed fixed.

Please use the template to ensure the CI process picks up your codesandbox.

It would be good to update this test and find why it's not failing.
https://github.com/Sage/carbon/blob/0f4583451da684c575bc468b63db32730982faab/src/components/form/form.spec.js#L46-L62

We also require signed commits, there are instructions on how to do that [here](https://github.com/Sage/carbon/blob/ff9ce77d40bca4339f8b5a4bbd1eef65ef6c1d92/docs/dev-environment-setup.md#gpg-verification).

Thanks,
Jamie

blaky:
Hi @jamime ,

I tried to reproduce the issue in code sandbox but I couldn't isolate the problem from our project to here. It is very obvious that the null check is missing from this function, it is there for the addListener. Logically, this change cannot have any side-effects: so far if the value was not provided an exception was thrown which then was not handled, and forwarded to the consumer.

Kind regards,
Bence

blaky:
> Hi @blaky,
> 
> We have a detailed [pull request template](https://raw.githubusercontent.com/Sage/carbon/f23993ab5c939abcecbe4065409217f50a4eead5/.github/PULL_REQUEST_TEMPLATE.md) that would help us get this change approved faster.
> In particular for QA we need a reproducible example in form of a codesandbox that shows the broken behaviour.
> 
> Our CI process then builds a new version and QA can then confirm that the bug is indeed fixed.
> 
> Please use the template to ensure the CI process picks up your codesandbox.
> 
> It would be good to update this test and find why it's not failing.
> https://github.com/Sage/carbon/blob/0f4583451da684c575bc468b63db32730982faab/src/components/form/form.spec.js#L46-L62
> 
> We also require signed commits, there are instructions on how to do that [here](https://github.com/Sage/carbon/blob/ff9ce77d40bca4339f8b5a4bbd1eef65ef6c1d92/docs/dev-environment-setup.md#gpg-verification).
> 
> Thanks,
> Jamie

The test in question is not failing because it is doing one operation at the time. The `checkStickyFooter` function is throttled, which means that the first call is still fine, it can remove the listener form the DOM element while it is mounted, but subsequent calls to the function might get executed when the element is already destroyed. 
I am more than happy to show you our setup, we are also experiencing issues in the Form component's `getBoundingClientRect()` calls, but those don't end up blocking the user.

Thanks in advance,
Bence

mkrds (approved):

jamime:
@blaky thanks for updating the PR. We did some manual testing and found some scenarios that were not covered by this PR. We'd like to spend a bit more time and deliver a more robust solution for you. 

We have the ticket FE-4290 which is in progress now.

In particular, we're going to remove the `ResizeObserver` and replace it with our hook [`useResizeObserver`](https://github.com/Sage/carbon/blob/master/src/hooks/__internal__/useResizeObserver/index.js). In form we also found that it's possible to remove the event listener from the wrong element because it's not memoized correctly.
https://github.com/Sage/carbon/blob/0f4583451da684c575bc468b63db32730982faab/src/components/form/form.component.js#L73

We'll update this ticket once we've produced a fix.

PR #4331, opened 2021-08-24
refactor(icon): use design system icon names

jamime:
### Proposed behaviour
Refactor the icon unicodes to follow the convention on the design system so it is easier to maintain.

### Current behaviour
The file is hard to maintain with a lot of legacy/duplicates.

### Checklist

### Additional context
I first added a test with a snapshot so we can be sure the final list of unicodes did not change.

Fixes FE-4293

### Testing instructions
Jest tests should be passing and there should be no difference to any component.


src/components/icon/__snapshots__/icon-unicodes.spec.js.snap
```diff
@@ -0,0 +1,206 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`the unicodes have not changed 1`] = `
+Object {
+  "add": "\\\\e940",
```
dawidzarzycki, line 5: Is it correct that we have 4 `\` instead of 2?

edleeks87: It's adding escape characters which it doesn't seem to do in any of the snapshots for components using `Icon` 🤔

jamime: So we actually need `"\e940"` `\` by itself is an escape char, so we need a second one to cancel the escape and treat it as a normal backslash.

I'm not sure why jest adds additional slashes, two should be sufficient. My guess is they replace every `\` with two.
I logged out the object to check and the object does only contain two backslashes as expected.

image

edleeks87 (approved):

dawidzarzycki (approved):

DlgSHi (approved):
no `chromatic` changes noticed

PR #4336, opened 2021-08-26
feat(duelling-picklist): add support for reordering selected items by drag&drop - FE-4120

rdzielakpgs:
### Proposed behaviour
**Use case**

As a user
I want to be able to reorder the items I have selected using the duelling pick list.

Context (original request from Attie Snyders):

I would like to propose an update to the Duelling Pick-list Pattern for some of our use cases. It is to allow you to reorder assigned items. Some context are sequence or order sensitive. So when adding items in the correct order using the duelling pick-list, it works, but when you need to change the order, the only current way is to remove all items and re-add them in the correct order. This isn't a great experience.

I am proposing that we allow a reorder icon (similar to table rows that you can reorder) in the Duelling Pick-list to make this experience better for users.

**Acceptance Criteria**

I can click and drag a selected item to reorder.
I cannot reorder the unselected items on the left.
The cursor should act in the same way we use for other drag & drop UI elements.
This functionality should be optional (for implementors).
No version available in Design System yet.
Storybook MDX updated with a compatibility table showing which version of the Design System is implemented

### Current behaviour
It's not implemented

### Checklist

### Additional context
image

### DIFF
--- a/src/components/draggable/draggable-container.component.js
+++ b/src/components/draggable/draggable-container.component.js
@@ -101,7 +101,7 @@ DraggableContainer.propTypes = {
   /** Callback fired when order is changed */
   getOrder: PropTypes.func,
   /** Overrides the default rendered HTML tag of the DraggableContainer component */
-  as: PropTypes.string,
+  as: PropTypes.oneOf(["div", "ul"]),
   /**
    * The content of the component
    *
--- a/src/components/draggable/draggable-container.component.js
+++ b/src/components/draggable/draggable-container.component.js
@@ -25,7 +25,7 @@ const DropTarget = ({ children, getOrder, ...rest }) => {
   );
 };
 
-const DraggableContainer = ({ children, getOrder, ...rest }) => {
+const DraggableContainer = ({ children, getOrder, as, ...rest }) => {
   const [draggableItems, setDraggableItems] = useState(
     React.Children.toArray(children)
   );
@@ -75,12 +75,13 @@ const DraggableContainer = ({ children, getOrder, ...rest }) => {
 
   return (
     <DndProvider backend={Backend}>
-      <DropTarget getOrder={getItemsId} {...marginProps}>
+      <DropTarget as={as} getOrder={getItemsId} {...marginProps}>
         {draggableItems.map((item) =>
           React.cloneElement(
             item,
             {
               id: `${item.props.id}`,
+              as: as === "ul" ? "li" : undefined,
               findItem,
               moveItem,
             },
@@ -99,6 +100,8 @@ DraggableContainer.propTypes = {
   ...marginPropTypes,
   /** Callback fired when order is changed */
   getOrder: PropTypes.func,
+  /** Overrides the default rendered HTML tag of the DraggableContainer component */
+  as: PropTypes.string,
   /**
    * The content of the component
    *
@@ -129,5 +132,7 @@ DropTarget.propTypes = {
   children: PropTypes.node.isRequired,
   getOrder: PropTypes.func,
 };
+
+DraggableContainer.displayName = "DraggableContainer";
 
 export default DraggableContainer;
--- a/src/components/draggable/draggable-item.style.js
+++ b/src/components/draggable/draggable-item.style.js
@@ -1,14 +1,69 @@
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import { padding, margin } from "styled-system";
-
 import { baseTheme } from "../../style/themes";
 import Icon from "../icon";
 
+const StyledIcon = styled(Icon)`
+  margin-left: auto;
+`;
+
 const StyledDraggableContainer = styled.div`
+  ${({ as }) =>
+    as === "ul" &&
+    css`
+      padding-inline-start: 0;
+      width: 100%;
+      height: 100%;
+      overflow-x: visible;
+      margin: 0;
+      padding-right: 0;
+      ${({ theme }) =>
+        css`
+          scrollbar-color: ${theme.scrollbar.light.thumb}
+            ${theme.scrollbar.light.track};
+
+          &::-webkit-scrollbar {
+            width: 8px;
+          }
+
+          &::-webkit-scrollbar-thumb {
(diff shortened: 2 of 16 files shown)


src/components/draggable/draggable-container.component.js
```diff
@@ -6,26 +6,34 @@ import styledSystemPropTypes from "@styled-system/prop-types";
 
 import { filterStyledSystemMarginProps } from "../../style/utils";
 import DraggableItem from "./draggable-item.component";
-import { StyledIcon, StyledDraggableContainer } from "./draggable-item.style";
+import {
```
mkrds, line 9: Instead of adding `renderAsUl` and `renderAsLi` props couldn't you just pass `as` props like that? :
```javascript
<DraggableContainer as="ul">
  <DraggableItem key="1" id={1} as="li">
    simple content
  </DraggableItem>
  <DraggableItem key="2" id={2} as="li">
    simple content
  </DraggableItem>
</DraggableContainer>
```
All you'd have to do is just to ensure that `as` is passed straight to the wanted StyledComponent

src/components/duelling-picklist/picklist-item/picklist-item.component.js
```diff
@@ -4,6 +4,7 @@ import { CSSTransition } from "react-transition-group";
 
 import {
   StyledPicklistItem,
+  StyledPicklistItemAsDiv,
```
mkrds, line 7: Same as above, you could simply pass `as="div"` to the PicklistItem and ensure that it get's passed down to the  `StyledPicklistItem`

rdzielakpgs: fixed, thanks!

mkrds:
In provided example it seems that the styles are not ok as the PicklistItems in the left Picklist are wider than the ones on the right side.

src/components/duelling-picklist/duelling-picklist.style.js
```diff
@@ -30,6 +40,54 @@ const StyledDuellingPicklist = styled.div`
   ${StyledPicklist}:last-of-type {
     padding-left: 36px;
   }
+  ${StyledDraggableContainer} {
```
mkrds, line 43: Are these styles really needed? They're already defined in `Picklist` component so it's kinda redundant

rdzielakpgs: These are needed to match style for picklist-only.

src/components/draggable/draggable-item.style.js
```diff
@@ -1,14 +1,28 @@
...
 const StyledDraggableItem = styled.div`
+  ${({ as }) =>
+    as === "li" &&
+    css`
+      border-bottom: none !important;
+      padding: 0 !important;
+      &:not(:first-child) {
+        margin-top: 8px !important;
+      }
+    `};
```
edleeks87, line 25: `;` isn't needed here

src/components/draggable/draggable-item.style.js
```diff
@@ -1,14 +1,28 @@
...
+      padding-right: 4px;
+    `};
   ${margin}
 `;
 
 const StyledDraggableItem = styled.div`
+  ${({ as }) =>
+    as === "li" &&
+    css`
+      border-bottom: none !important;
```
edleeks87, line 20: We want to avoid using `!important`, you should use the mechanism built in to styled-components to achieve this if possible

``` 
&& {
      border-bottom: none;
      &:not(:first-child) {
        margin-top: 8px;
      }
    }
```

src/components/draggable/draggable-item.style.js
```diff
@@ -1,14 +1,28 @@
...
+      padding: 0 !important;
```
edleeks87, line 21: Do you need this as seems to be overriding this override?
https://github.com/Sage/carbon/blob/e7bcb2c352768bd58127dbc9ec440c13452584d5/src/components/duelling-picklist/duelling-picklist.style.js#L66

edleeks87 (commented):
The markup for the `draggable` stories is still invalid

image

edleeks87 (dismissed):

src/components/duelling-picklist/duelling-picklist.style.js
```diff
@@ -30,6 +35,53 @@ const StyledDuellingPicklist = styled.div`
...
+
+        &::-webkit-scrollbar-track {
+          background-color: ${theme.scrollbar.light.track};
+        }
+      `}
+
+    ${StyledIcon}:last-of-type {
+      display: none;
+    }
+    ${StyledPicklistItem} {
```
mkrds, line 68: One more thing, in scenario of a DuellingPicklist which already has some items selected, this will result in the animation being run on all the items as shown below:

https://user-images.githubusercontent.com/20651022/133985647-8b25e92a-82e2-4a0b-903e-c27e3d1407eb.mov

src/components/duelling-picklist/duelling-picklist.style.js
mkrds: I understand that, but we'll end up with two sets of styles that do exactly the same. It is very likely that someone doing changes to this component in the future will not be aware that the styling changes related to the animation or the scrollbar have to be applied in two different places. Please try to play around with the component structure so that it's not necessary to add these styles at all.

mkrds (commented):
Normal PicklistItem and draggable PicklistItem still have different widths

src/components/duelling-picklist/duelling-picklist.stories.mdx
```diff
@@ -64,6 +64,14 @@ Example of composed Duelling Picklist components with order preservation and sea
   <Story id="design-system-duellingpicklist--default" name="Default" />
 </Preview>
 
+### Draggable
+
+Example of composed Duelling Picklist components with draggable selected items
+
+<Preview>
+  <Story id="design-system-duellingpicklist--draggable" name="Draggable" />
```
edleeks87, line 72: This id needs updating to `design-system-duellingpicklist-test--draggable`

image

PR #4337, opened 2021-08-26
feat(definition-list): add support for rendering component as single column

edleeks87:
### Proposed behaviour

Adds `asSingleColumn` prop and `DlContext` to support rendering the `Dl` as single column and
overriding default spacing applied to `Dt` component. Tidied up some of the prop tables and added a
more complex example using multiple single column `Dl` with segment headers and horizontal dividers

### Current behaviour

No support for rendering component as a single column

### Checklist

### Additional context

image

image

### Testing instructions

`design-system-definition-list--as-a-single-column`
`design-system-definition-list--multiple-single-columns-with-segments`

https://codesandbox.io/s/carbon-quickstart-forked-p7cz9?file=/src/index.js


nicktitchmarsh (approved):

src/components/definition-list/dd.component.js
```diff
@@ -3,13 +3,11 @@ import propTypes from "@styled-system/prop-types";
...
 
-const Dd = ({ mb = 2, children, ...props }) => {
-  return (
-    <StyledDd mb={mb} {...props}>
-      {children}
-    </StyledDd>
-  );
-};
+const Dd = ({ children, ...rest }) => (
+  <StyledDd mb={rest.mb || 2} {...rest}>
```
mmaciorowski-sage, line 7: That line looks a little strange. It should be defined that default `mb` is 2.

src/components/definition-list/dt.component.js
```diff
@@ -1,11 +1,19 @@
...
+const Dt = ({ children, ...rest }) => {
+  const { asSingleColumn } = useContext(DlContext);
 
-const Dt = ({ pr = 3, children, mb = 2, ...rest }) => {
   return (
-    <StyledDt data-element="dt" mb={mb} pr={pr} {...rest}>
+    <StyledDt
+      data-element="dt"
+      mb={rest.mb || asSingleColumn ? undefined : 2}
+      pr={rest.pr || asSingleColumn ? undefined : 3}
```
mmaciorowski-sage, line 14: I'm not sure, but I think it would be nice to separate out the `mb` and `pr` props there it will be more readable and props won't be overwritten by `{...rest}`.

src/components/definition-list/dd.component.js
edleeks87: The issue here is that if we define/ destructure the prop in the args we end up with some weird prop tables. Adding them as a default prop would work in the short term but eventually we will have to move all default props into the args again as that is the pattern react are moving towards. I'd prefer to destructure in the component and instead to avoid that, I don't think it's an issue as the default value is defined in the prop tables now I've updated it to use the `StyledSystemProps` table

image

src/components/definition-list/dt.component.js
edleeks87: I could destructure them within the component itself but having it done in the args causes the same issues with the prop tables I mentioned above

image

nicktitchmarsh (approved):

mmaciorowski-sage (approved):

mmaciorowski-sage (approved):

DlgSHi (approved):

PR #4342, opened 2021-08-27
 feat(inline-inputs): add labelWidth and inputWidth props (FE-4279)

### LINKED ISSUE #4300: Inline Input labels do not align with other inputs on a form
### Current behaviour

InlineInputs labels are always inline and the same width of the text within them, this does not match with the other fields you can put on a form which have properties to configure where to display them, how much width to give them, or when to break between displaying them inline or above the field.
The result is a strange looking form where the InlineInputs are not aligned with anything else, including each other.
image

### Expected behaviour

When I do not provide the labelInline property to an InlineInput component, the inline inputs span the full width of the form with the label displayed above them. If I do provide the labelInline property, then I can use the same labelWidth and input width properties so that it aligns with the other fields on the form.

https://codesandbox.io/s/wonderful-moore-f1nc4?file=/src/index.js

### Suggested Solution

_No response_

### Carbon Version

80.1.0

Firefox

Windows

_No response_

### Confidentiality

grabkowski:
### Proposed behaviour
Inline inputs should have the ability to set label and content width the same way as other form elements 
Screenshot 2021-08-27 at 17 53 43

### Current behaviour
Inline inputs could not be aligned with other form elements when their labels are in-line
Screenshot 2021-08-27 at 17 53 30

### Checklist

### Additional context
Resolves #4300 

### Testing instructions
1. run npm start
2. open http://localhost:9001/?path=/story/design-system-form--default-with-sticky-footer
3. open http://localhost:9001/?path=/story/design-system-inline-inputs--default-story

### DIFF
--- a/src/components/form/form.stories.mdx
+++ b/src/components/form/form.stories.mdx
@@ -10,6 +10,7 @@ import { RadioButton, RadioButtonGroup } from "../radio-button";
 import DateInput from "../date";
 import { Select, Option } from "../select";
 import { Checkbox } from "../checkbox";
+import InlineInputs from "../inline-inputs";
 import Hr from "../../components/hr";
 import Switch from "../switch";
 import Button from "../button";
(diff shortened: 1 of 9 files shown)


src/components/form/form.stories.mdx
```diff
@@ -50,25 +51,31 @@ When `stickyFooter` prop is set as true, footer becomes stickied to the bottom o
...
-      stickyFooter
-    >
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
-      <Textbox label="Textbox" />
+    <Form>
```
mkrds, line 54: Why did you remove the footer buttons from `Form` ?

src/components/form/form.style.js
```diff
@@ -53,6 +54,15 @@ export const StyledForm = styled.form`
       theme.spacing * fieldSpacing}px;
   }
 
+  ${StyledInlineInputs} {
```
mkrds, line 57: How about adding another `&` to raise specificity of the rule on line 51 instead of adding this whole block?

DipperTheDan (commented):
Nothing further to add than what @mkrds has already commented. Address the comments and I'll happily approve 😄

src/components/form/form.style.js
grabkowski: there's no need to do this, as it will specifically target inline inputs, I don't understand why should I raise specificity of it.

DipperTheDan (approved):

mkrds (dismissed):

DlgSHi (dismissed):

DipperTheDan (dismissed):

DlgSHi (approved):

nicktitchmarsh (approved):

PR #4381, opened 2021-09-09
feat(textbox): add a character counter - FE-4265

mmaciorowski-sage:
### Proposed behaviour
Textbox should implement the character limit functionality the same way and with the same format Textarea does.

### Current behaviour
Textbox no have implemented the character limit functionality.

### Checklist

### Testing instructions

https://codesandbox.io/s/carbon-quickstart-forked-ct15l?file=/src/index.js


src/components/textbox/textbox.spec.js
```diff
@@ -36,6 +38,37 @@ describe("Textbox", () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it("renders a counter", () => {
+    const wrapper = mount(<Textbox value="test string" characterLimit="100" />);
```
dawidzarzycki, line 42: I wonder, maybe it would be worth to do
```js
const characterLimit = "100"
const value = "test string"
```

```js
expect(wrapper.find(CharacterCount).text()).toBe(value.length '/' characterLimit);
``` 
or at least
```js
expect(wrapper.find(CharacterCount).text()).toBe(`${value.length}/100`);
```

So we can see that it should be `11` or other number (if different value used) without reading the value and counting characters to be sure we didn't make any mistake

This is just an idea, it's up to u

src/components/textbox/textbox.spec.js
```diff
@@ -36,6 +38,37 @@ describe("Textbox", () => {
...
+        value="test string"
+        characterLimit="10"
+        enforceCharacterLimit={false}
+        warnOverLimit
+      />
+    );
+
+    assertStyleMatch(
+      {
+        color: "#C7384F",
```
dawidzarzycki, line 59: Is this value comes from our theme object? If so it would be good to use baseTheme here as well

src/hooks/__internal__/useCharacterCount/index.spec.js
```diff
@@ -0,0 +1,111 @@
...
+    <>
+      {characterCount}
+      <span data-element="max-length">{maxLength}</span>
+    </>
+  );
+};
+
+const render = (props, params) => mount(<TestComponent {...props} />, params);
+
+TestComponent.propTypes = {
```
dawidzarzycki, line 33: I'm not sure we need to create propTypes for Mock component

src/hooks/__internal__/useCharacterCount/index.spec.js
```diff
@@ -0,0 +1,111 @@
...
+  value: PropTypes.string,
+  characterLimit: PropTypes.string,
+  warnOverLimit: PropTypes.bool,
+  enforceCharacterLimit: PropTypes.bool,
+};
+
+describe("useCharacterCount", () => {
+  it("returns a counter and no maxLength", () => {
+    const wrapper = render({
+      value: "test string",
```
dawidzarzycki, line 43: I think that we can assign it to the default prop in Mock Component as we are using it always the same

mmaciorowski-sage: I think now it's more readable. It's just a unit test, not a component.

dawidzarzycki (approved):

edleeks87 (approved):

edleeks87 (approved):

divyajindel (approved):

PR #4385, opened 2021-09-10
refactor(pod): convert class component to functional (FE-3660)

grabkowski:
Convert Pod class component to functional.
Remove deprecated PodManager component.

### Checklist


src/components/pod/pod.component.js
```diff
@@ -26,328 +25,220 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
+    const [isEditHovered, setEditHovered] = useState(false);
+    const [isDeleteFocused, setDeleteFocused] = useState(false);
+    const [isDeleteHovered, setDeleteHovered] = useState(false);
+    const [isUndoFocused, setUndoFocused] = useState(false);
+    const [isUndoHovered, setUndoHovered] = useState(false);
 
-  handleEditBlur = () => {
-    this.setState({ isEditFocused: false });
-  };
+    let podHeight;
```
dawidzarzycki, line 60: Maybe instead of doing `let` we could create simple `useRef` for this variable?

src/components/pod/pod.component.js
```diff
@@ -26,328 +25,220 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
 
-  handleDeleteMouseEnter = () => {
-    this.setState({ isDeleteHovered: true });
-  };
+    if (height && typeof height === "number") {
+      podHeight = `${height}px`;
+    } else if (height) {
+      podHeight = height;
+    }
```
dawidzarzycki, line 66: I guess we can do something like 
```js
if(height){
  podHeight = `${String(height)px}`
}
```
so I think that we won't need conditionals here and then maybe that useRef won't be necessary

src/components/pod/pod.component.js
```diff
@@ -26,328 +25,220 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
 
     return (
       <StyledPod
         {...rest}
-        className={this.props.className}
+        className={className}
         internalEditButton={internalEditButton}
-        {...tagComponent("pod", this.props)}
+        {...tagComponent("pod", rest)}
         height={podHeight}
```
dawidzarzycki, line 237: Maybe we can do just 
```js
`${String(height)px}`
```
So the conditional that is above in this file won't be needed

edleeks87 (approved):

src/components/pod/pod.component.js
grabkowski: I don't think that we need another ref here

src/components/pod/pod.component.js
grabkowski: as @edleeks87 already said, we could end up with `pxpx`

nicktitchmarsh (approved):

ZhuoyuJin (approved):

PR #4389, opened 2021-09-13
feat(simple-select): add prop to specify the open direction of a select list - FE-4316

mmaciorowski-sage:
### Proposed behaviour
- Add `listPlacement` prop passed to Popover component `placement` prop.
- Add `flipEnabled` prop to enforce list placement if the set placement doesn't fit.
- Change the list opening animation from accordion to fading.

### Current behaviour
No way to set the select options list position.

### Checklist

### Testing instructions
https://codesandbox.io/s/carbon-quickstart-forked-ngpvp

### DIFF
--- a/src/components/select/select-list/select-list.component.js
+++ b/src/components/select/select-list/select-list.component.js
@@ -25,7 +25,7 @@ import StyledSelectListContainer from "./select-list-container.style";
 import Loader from "../../loader";
 import Option from "../option/option.component";
 
-const popoverModifiers = [
+const fixedPopoverModifiers = [
   {
     name: "offset",
     options: {
@@ -59,6 +59,8 @@ const SelectList = React.forwardRef(
       multiColumn,
       tableHeader,
       loaderDataRole,
+      listPlacement = "bottom-start",
+      flipEnabled = true,
       ...listProps
     },
     listContainerRef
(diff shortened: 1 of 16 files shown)


src/components/select/select-list/select-list.component.js
```diff
@@ -59,6 +59,8 @@ const SelectList = React.forwardRef(
       multiColumn,
       tableHeader,
       loaderDataRole,
+      listPlacement = "bottom-start",
+      flipEnabled = true,
```
edleeks87, line 63: I think it would be better to have the default value `flipEnabled = true` in each of the `Select` components rather than defaulting it in the `SelectList` so that it shows up in the prop tables. Alternatively we could change invert it and surface a `flipDisabled` which would also convey that the flip functionality is "on" by default

image

image

src/components/select/select-list/select-list.component.js
```diff
@@ -59,6 +59,8 @@ const SelectList = React.forwardRef(
       multiColumn,
       tableHeader,
       loaderDataRole,
+      listPlacement = "bottom-start",
```
edleeks87, line 62: The defaults might be better being in the `Select` components so the values show up in the prop tables

src/components/select/select-list/select-list.component.js
mmaciorowski-sage: Good point, thanks.

Zrzut ekranu 2021-09-15 o 16 16 57

edleeks87 (dismissed):

DipperTheDan (approved):

DipperTheDan (approved):

edleeks87 (approved):

divyajindel (approved):

PR #4402, opened 2021-09-16
refactor(split-button): rewrite class component to functional

grabkowski:
Rewrite SplitButton class component to functional

BREAKING CHANGE: SplitButton is now a functional component and cannot be extended

### Checklist

#### QA

- [ ] Tested in sandbox/storybook

### DIFF
--- a/src/components/split-button/split-button.component.js
+++ b/src/components/split-button/split-button.component.js
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { useRef, useState, useCallback, useEffect } from "react";
 import PropTypes from "prop-types";
 import styledSystemPropTypes from "@styled-system/prop-types";
 import Icon from "../icon";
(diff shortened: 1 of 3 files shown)


src/components/split-button/split-button.component.js
```diff
@@ -17,108 +17,99 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
-      document.addEventListener("keydown", this.handleKeyDown);
-      this.listening = true;
-    }
-  };
-
-  hideButtons = () => {
-    if (this.isToggleButtonFocused) return;
+const SplitButton = ({
+  align = "left",
+  as = "secondary",
```
mkrds, line 22: I'd assign the `secondary` default value to the `buttonType` instead of `as` since the `as` prop is meant to be deprecated here

src/components/split-button/split-button.component.js
```diff
@@ -17,108 +17,99 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
+      ) {
+        hideButtons();
+      }
+    },
+    [hideButtons]
+  );
+
+  const handleKeyDown = useCallback(
+    (ev) => {
+      if (!listening.current) {
```
mkrds, line 68: Kinda confusing because to be precise the listeners are still attached. Maybe it would be better to use the `showAdditionalButtons` here?

src/components/split-button/split-button.component.js
```diff
@@ -17,108 +17,99 @@ const marginPropTypes = filterStyledSystemMarginProps(
...
-  };
+      if (Events.isUpKey(ev)) {
+        nextIndex = currentIndex > 0 ? currentIndex - 1 : numOfChildren;
+        ev.preventDefault();
+      }
+      if (Events.isDownKey(ev)) {
+        nextIndex = currentIndex < numOfChildren ? currentIndex + 1 : 0;
+        ev.preventDefault();
+      } else if (Events.isTabKey(ev)) {
+        // timeout enforces thet the "hideButtons" method will be run after browser focuses on the next element
```
DipperTheDan, line 84:
```suggestion
        // timeout enforces that the "hideButtons" method will be run after browser focuses on the next element
```

src/components/split-button/split-button.component.js
grabkowski: if I'll do it then it will be another breaking change. The `as` prop should be deprecated  in a separate task.

src/components/split-button/split-button.component.js
grabkowski: this is a leftover variable from the class based version. Looks like it's only used there, so you're right, it could be replaced with `showAdditionalButtons `.

src/components/split-button/split-button.component.js
grabkowski: wow - this typo was there for more than two years now! Fixed.

mkrds (dismissed):

DipperTheDan (dismissed):

DlgSHi (approved):

mkrds (approved):

DipperTheDan (approved):

Issue #4321, opened 2021-08-23
locales/en-GB is missing a typescript file


philearley:
### Current behaviour

When you import this file into a project using Typescript an error is generated 
`"Could not find a declaration file for module 'carbon-react/lib/locales/en-gb'."`

Manual work around at the moment is to declare the module in a top level file
`declare module "carbon-react/lib/locales/en-gb";`

### Expected behaviour

To not need to manually declare this in our own app.

### CodeSandbox or Storybook URL

N/A

### Carbon Version

81.2.1

### What browsers are you seeing the problem on?

Other

### What Operating System are you seeing the problem on?

MacOS

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

jamime:
FE-4291

PR #4415, opened 2021-09-21
feat(i18n): add types for the base locale file - FE-4291

mmaciorowski-sage:
### Proposed behaviour
Added types for the base locale file (en-gb.js)

### Current behaviour
When you import this file into a project using Typescript an error is generated
"Could not find a declaration file for module 'carbon-react/lib/locales/en-gb'."

https://user-images.githubusercontent.com/77274590/134500108-ef1e9e51-3bd3-46cd-9451-00d97a63fe1d.mov

### Checklist

#### QA

- [x] Tested in sandbox/storybook

### Additional context
Resolves: #4321

### Testing instructions
Please download and test it locally:
https://codesandbox.io/s/carbon-quickstart-forked-1y1ui


src/locales/en-gb.d.ts
```diff
@@ -0,0 +1,96 @@
...
+  date: {
+    formats: {
+      inputs: () => string[];
+      javascript: () => string;
+    };
+  };
+  errors: {
+    messages: {
+      formSummary: (
+        errors: number | string,
```
sjohnson-sage, line 30: `errors` and `warnings` should only be strings because they're passed into `isSingular` which passes them into `parseInt`

image

src/locales/en-gb.d.ts
```diff
@@ -0,0 +1,96 @@
...
+  numeralDate: {
+    validation: {
+      day: () => string;
+      month: () => string;
+      year: () => string;
+    };
+  };
+  pager: {
+    show: () => string;
+    records: (count: number | string, showNumber: boolean) => string;
```
sjohnson-sage, line 48: `count` can only be a string for the reason in my other comment

mmaciorowski-sage: https://github.com/Sage/carbon/blob/master/src/components/pager/pager.component.js#L203
Zrzut ekranu 2021-09-22 o 10 19 05

I based on the prop types here. I think if we want to have only string type, we will have add toString():
https://github.com/Sage/carbon/blob/master/src/components/pager/pager.component.js#L167
and
https://github.com/Sage/carbon/blob/master/src/components/pager/pager.component.js#L159

src/locales/en-gb.d.ts
mmaciorowski-sage: Here based on the prop types should be a number type.
Zrzut ekranu 2021-09-22 o 10 24 35

sjohnson-sage: Ok in that case, the `isSingular` function needs to change if the `count ` parameter is a number - because it won't be valid in TypeScript if we try to pass a number into `parseInt` as the first parameter.

So the simple solution would be to check the type of `count` in `isSingular`, and if it's not a string don't pass it into `parseInt`. I.e something like this: 

```js
export const isSingular = (count) =>
  typeof count === "string" ? parseInt(count, 10) === 1 : count === 1;
```

sjohnson-sage (approved):

edleeks87 (approved):

edleeks87 (dismissed):

sjohnson-sage (approved):

edleeks87 (approved):

PR #4429, opened 2021-09-26
fix(i18n-provider): fix type definition

blaky:
The `[string]` type definition describes an array with a single member that is a string, but based on the examples in the en-gb.js locale file it looks like a string array of any size can be used.


sjohnson-sage (approved):

mkrds (approved):

sjohnson-sage:
Good spot @blaky and thanks for taking the time to fix it! I'll get this merged for you

PR #4442, opened 2021-09-29
fix(icon): add support for keyboard focus, gold outline, role and ariaLabel when has tooltip FE-4364

### LINKED ISSUE #4404: Tooltip component create Axe fails - ARIA tooltip nodes must have an accessible name
### Current behaviour

Using `Tooltip` in `Select` or `Textbox` for exemple create Axe fails.
https://carbon.sage.com/?path=/story/design-system-textbox--with-label-help
image

### Expected behaviour

Add an aria-label maybe to avoid this Axe fails.

https://carbon.sage.com/?path=/story/design-system-textbox--with-label-help

### JIRA Ticket (Sage Only)

SBS-20270

### Suggested Solution

_No response_

### Carbon Version

87.0.0

Chrome

Windows

_No response_

### Confidentiality

### LINKED ISSUE #4428: Icon component - Keyboard accessibility, focus and props are incorrect
### Current behaviour

In discussion with @edleeks87 issue raised for the Icon component.
"`aria-label` and `role` props won't be passed down to Icon"
and
"the keyboard accessibility isn’t quite right (no gold border)"

### Expected behaviour

Created on request, Icon should support accessible props and show a gold border

https://carbon.sage.com/?path=/docs/design-system-accordion--validation-icon

### JIRA Ticket (Sage Only)

SBS-19183

### Suggested Solution

_No response_

### Carbon Version

^84.1.0

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4428
fix #4404

### Proposed behaviour

Ensures `tabIndex` is set to 0 if has `tooltipMessage` and no prop passed. Sets `role` to "tooltip"
if none passed and `Icon` is not composed as part of `ValidationIcon` or `Help`. Sets `ariaLabel` to
`type` if none passed and not composed as part of `ValidationIcon` or `Help`.

Moves the `role` prop and surfaces `ariaLabel` so they can be passed to `Icon` to fix axe warning.

### Current behaviour

No `ariaLabel`, `tabIndex` or `role` set on `Icon` when it has `tooltipMessage`, no gold outline when focused.

Axe warning on all `labelHelp` and `Help` stories.

### Checklist

#### QA

### Testing instructions

Check `labelHelp` stories and `Help`
Check `icon--with-tooltip` ensure added props present and can tab to `Icon` etc
Check `ValidationIcon` html is unchanged

will add sandbox if needed

### DIFF
--- a/src/__internal__/validations/validation-icon.component.js
+++ b/src/__internal__/validations/validation-icon.component.js
@@ -87,6 +87,7 @@ const ValidationIcon = ({
         }
         isPartOfInput={isPartOfInput}
         inputSize={size}
+        blockFocusBehaviour
       />
     </ValidationIconStyle>
   );
--- a/src/components/help/help.component.js
+++ b/src/components/help/help.component.js
@@ -26,6 +26,7 @@ const Help = ({
   tooltipBgColor,
   tooltipFontColor,
   tooltipFlipOverrides,
+  ariaLabel,
   ...rest
 }) => {
   const helpElement = useRef(null);
@@ -56,7 +57,6 @@ const Help = ({
 
   return (
     <StyledHelp
-      role="tooltip"
       className={className}
       as={tagType}
       href={href}
@@ -84,6 +84,8 @@ const Help = ({
         tooltipBgColor={tooltipBgColor}
         tooltipFontColor={tooltipFontColor}
         tooltipFlipOverrides={tooltipFlipOverrides}
+        ariaLabel={ariaLabel || type}
+        role="tooltip"
         blockFocusBehaviour
       />
     </StyledHelp>
@@ -134,6 +136,8 @@ Help.propTypes = {
       `The \`${propName}\` prop supplied to \`${componentName}\` must be an array containing some or all of ["top", "bottom", "left", "right"].`
     );
   },
+  /** Aria label for accessibility purposes */
+  ariaLabel: PropTypes.string,
 };
 
 Help.defaultProps = {
--- a/src/components/help/help.component.js
+++ b/src/components/help/help.component.js
@@ -84,6 +84,7 @@ const Help = ({
         tooltipBgColor={tooltipBgColor}
         tooltipFontColor={tooltipFontColor}
         tooltipFlipOverrides={tooltipFlipOverrides}
+        blockFocusBehaviour
       />
     </StyledHelp>
   );
--- a/src/components/icon/icon.component.js
+++ b/src/components/icon/icon.component.js
@@ -37,6 +37,9 @@ const Icon = React.forwardRef(
       tabIndex,
       isPartOfInput,
       inputSize,
+      role,
+      ariaLabel,
+      blockFocusBehaviour,
       ...rest
     },
     ref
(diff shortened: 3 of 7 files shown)


DipperTheDan (approved):

src/components/help/help.component.js
```diff
@@ -84,6 +84,9 @@ const Help = ({
         tooltipBgColor={tooltipBgColor}
         tooltipFontColor={tooltipFontColor}
         tooltipFlipOverrides={tooltipFlipOverrides}
+        ariaLabel={ariaLabel || type}
```
jamime, line 87: Same here regarding the label being `type`.

src/__internal__/validations/validation-icon.component.js
```diff
@@ -87,6 +87,7 @@ const ValidationIcon = ({
         }
         isPartOfInput={isPartOfInput}
         inputSize={size}
+        blockFocusBehaviour
```
jamime, line 90: This props is not clear, 

`blockFocusBehaviour=true` allows the icon to be focused.

I think we should run this by @tempertemper - is it necessary for the icon to be focusable, couldn't we use [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) and setting [`aria-invalid=false`](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21.html) if the `error` prop is passed (warnings shouldn't fail validation).

**New Behaviour**

https://user-images.githubusercontent.com/2328042/135587809-e2b957a1-b335-4f13-be14-b11cfb4f9e40.mov

**Existing Behaviour**

https://user-images.githubusercontent.com/2328042/135587828-96a1e0d5-2515-44b6-affb-3b48065f1095.mov

src/components/help/help.component.js
```diff
@@ -84,6 +84,9 @@ const Help = ({
...
+        role="tooltip"
+        blockFocusBehaviour
```
jamime, line 89: `blockFocusBehaviour` is making this focusable. I think it should be opt-out, so focusable by default, and opt-out on form validations, what do you think?

src/components/icon/icon.component.js
```diff
@@ -79,11 +82,12 @@ const Icon = React.forwardRef(
       fontSize,
       isInteractive,
       iconColor,
-      tabIndex,
       type: iconType(),
       ...filterStyledSystemMarginProps(rest),
     };
 
+    const hasTooltip = tooltipMessage && !blockFocusBehaviour;
```
jamime, line 89: Why does `blockFocusBehaviour ` change the `role`? 

Please can you update the PR description
> Sets ariaLabel to type if none passed and not composed as part of ValidationIcon or Help.

We should document the prop as well as the components it affects.

jamime (changes requested):

src/components/icon/icon.component.js
edleeks87: Yeah good point I think it shouldn't affect role really, I had it this way as originally Help set the role, but then I spotted that this caused axe issues so moved the role to the Icon but missed reverting this

src/components/help/help.component.js
edleeks87: Yeah I'll change it to be `focusable` or something similar

src/__internal__/validations/validation-icon.component.js
edleeks87: Ah I missed that, yeah happy to defer to the expert and implement whatever is deemed the right approach

edleeks87: I'll add the `DO NOT MERGE` label for now

PR #4460, opened 2021-10-05
fix(flat-table): move footer to be included in the component height to prevent it hiding table rows

### LINKED ISSUE #4273: Minimum height prop for FlatTable content
### Desired behaviour

We need a solution to set a minimum height for content of the table to avoid hiding content behind the footer and header.

### Current behaviour

FlatTables content can be hidden when using `hasStickyFooter` and `hasStickyHead`

https://user-images.githubusercontent.com/78448948/126976300-624e8176-a411-4633-b543-ebb0849c9b40.mp4

### Suggested Solution

_No response_

https://codesandbox.io/s/interesting-architecture-lz9jv?file=/src/index.js

_No response_

### Confidentiality

edleeks87:
fix #4273

### Proposed behaviour

Removes two unnecessary wrappers `StyledFlatTableRoot` and `StyledFlatTableBox` from `FlatTable`.
Moves `footer` node to be a direct sibling of the table so it is included as part of the component
height. Adds `Pager` to `custom sidebar` `Drawer` story

### Current behaviour

Footer is outside element that has adjustable height, there are unnecessary wrapper elements in the html structure, minHeight prop is not documented, no example of row based min-height

### Checklist

#### QA

### Additional context

image

### Testing instructions

see linked issue for codesandbox and add `minHeight` to `FlatTable`
`docs/flat-table--with-row-based-min-height` added
Worth checking all stories using flat-table for any regressions (including Drawer `custom sidebar`)


src/components/flat-table/flat-table-row/flat-table-row.style.js
```diff
@@ -114,9 +114,13 @@ const StyledFlatTableRow = styled.tr`
 
       ${StyledFlatTableHeader} {
         border-bottom: 1px solid ${borderColor(colorTheme, theme)};
-        :first-child {
-          border-left: 1px solid ${borderColor(colorTheme, theme)};
-        }
+
+        ${!isInSidebar &&
+        `
```
sjohnson-sage, line 119:
```suggestion
        css`
```

src/components/flat-table/flat-table.spec.js
```diff
@@ -507,11 +506,21 @@ describe("FlatTable", () => {
...
         {
           position: "sticky",
-          bottom: "-40px",
+          bottom: "0px",
         },
         wrapper.find(StyledFlatTableFooter)
       );
     });
+
+    it("sets the wrapper flex css props expected", () => {
```
sjohnson-sage, line 515:
```suggestion
    it("sets the wrapper flex css props as expected", () => {
```

src/components/flat-table/flat-table.component.js
```diff
@@ -102,6 +101,8 @@ FlatTable.propTypes = {
   hasStickyFooter: PropTypes.bool,
   /** Set the height of the table */
   height: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Set the min-height of the table */
```
sjohnson-sage, line 104: Worth noting in the comment that the string can be any valid CSS string, and a number will be converted to a pixel value - it'll be less confusing for users in the prop table

src/components/flat-table/flat-table.stories.mdx
```diff
@@ -1732,6 +1731,222 @@ of the button control.
...
+      useEffect(() => {
+        setMinHeight(getMinHeight(rowCount));
+      }, [rowCount])
+      return (
+        <div style={{ height: "220px", marginBottom: "100px" }}>
+          <Button mb={1} mr={1} disabled={rowCount <= 1} onClick={() => setRowCount((prev) => prev - 1)}>Decrease min rows</Button>
+          <Button mb={1} disabled={rowCount >= rows.length - 1} onClick={() => setRowCount((prev) => prev + 1)}>Increase min rows</Button>
+          <FlatTable
+            minHeight={minHeight}
+            height="100px"
```
sjohnson-sage, line 1917: I removed the `height="100px"` because we shouldn't need it as well here, but it broke the whole example so I think something is going wrong somewhere

Before you look into it, do we even need to provide this example at all? All we've done is add a `minHeight` prop, the implementation is then down to the user. IMO this is overkill and will just muddy up the docs a bit

sjohnson-sage:
Also the commit should probably be a `feat` saying as we're adding a new prop to `FlatTable`

src/components/flat-table/flat-table.stories.mdx
edleeks87: Yeah for the `minHeight` to be demoed you either need to set the `height` on the table or on the wrapper. I take your point about it being overkill for storybook, I'll just use this in the sandbox to show the linked issue has been fixed

src/components/flat-table/flat-table.component.js
edleeks87: Yeah good point will add that in thanks

src/components/flat-table/flat-table.stories.mdx
edleeks87: Actually looking at again I think it's related to it defaulting the height to `100%` if the sticky props are used and no height is passed

src/components/flat-table/flat-table-row/flat-table-row.style.js
edleeks87: `css` is used at the `return` so I don't think it's needed here as well?

DipperTheDan (approved):

PR #4463, opened 2021-10-07
test: change the approach for accessibility tests

DlgSHi:
### Proposed behaviour

- implement new approach for `accessibility` tests
- made new `cypress-axe-accessibility.yml` file where declared what we need to run
- made new `a11y-util.js` file to get all `components`/`stories` from `stories.js` file
- made separate files for 3 parts of tests to make possible to run each spec on different CI machine
- `TODO` uncomment `WGAC 2.0` `axe` rules

### Current behaviour

Currently we are passing only few stories per component
- we have currently `axe` tests for `WGAC 2.1`

### Checklist

#### QA

### Testing instructions

All tests should `pass`
New workflow should also `pass` after it will be merged with `master`


sianford (approved):
The changes look great and will give us rapid and extensive coverage.

cypress/features/regression/accessibility/a11y-third-part.test.js
```diff
@@ -0,0 +1,3 @@
+import tests from "../../../support/step-definitions/accessibility/a11y-utils";
+
+tests(601, 1000);
```
jamime, line 3: This should be left open-ended to capture any new tests that go beyond the 1000 range.
[Slice does not include start](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) so it should start at `600`.
```suggestion
tests(600);
```

cypress/features/regression/accessibility/a11y-second-part.test.js
```diff
@@ -0,0 +1,3 @@
+import tests from "../../../support/step-definitions/accessibility/a11y-utils";
+
+tests(301, 600);
```
jamime, line 3: [Slice does not include start](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) so it should start at `300`
```suggestion
tests(300, 600);
```

cypress.json
```diff
@@ -12,7 +12,7 @@
   "failOnStatusCode": false,
   "ignoreTestFiles": [
     "**/examples/*",
-    "**/!(*.spec).js"
+    "**/*.spec.js"
```
jamime, line 15: Why did this change?

cypress/support/step-definitions/accessibility/a11y-utils.js
```diff
@@ -0,0 +1,112 @@
...
+} from "../../../locators/action-popover";
+import { popoverSettingsIcon } from "../../../locators/popover-container";
+import { visitComponentUrl } from "../../helper";
+import storiesJSON from "../../../../storybook-static/stories.json";
+
+const A11YOptions = {
+  runOnly: {
+    type: "tag",
+    values: [
+      // "wcag2a", // WCAG 2.0 & WCAG 2.1 Level A
```
jamime, line 18: These should all be enabled.

jamime (changes requested):

cypress.json
DlgSHi: because it was ignoring all `.js` files

cypress/support/step-definitions/accessibility/a11y-utils.js
DlgSHi: nope, we will enable it after merge - because will have a lot of failures

cypress.json
jamime: Previously it would ignore any JS file that is not a spec (so spec was included). Now spec is being excluded. Why?

DlgSHi: because we will use `***.test.js` as file name to run and just ignore `unit` tests (`***.spec.js`)

.github/workflows/cypress-axe-accessibility.yml
```diff
@@ -0,0 +1,91 @@
...
+          install: false
+          start: npm start
+          wait-on: "http://localhost:9001"
+          wait-on-timeout: 250
+          record: true
+          parallel: true
+          group: ubuntu-accessibility
+          quiet: true
+          browser: chrome
+          spec: cypress/features/regression/accessibility/*test.js
```
samtjo, line 86:
```suggestion
          spec: cypress/features/regression/accessibility/*.test.js
```

cypress/support/step-definitions/accessibility/a11y-utils.js
```diff
@@ -0,0 +1,112 @@
...
+  getDataElementByValue,
+} from "../../../locators";
+import { accordionDefaultTitle } from "../../../locators/accordion";
+import {
+  actionPopoverButton,
+  actionPopoverWrapper,
+} from "../../../locators/action-popover";
+import { popoverSettingsIcon } from "../../../locators/popover-container";
+import { visitComponentUrl } from "../../helper";
+import storiesJSON from "../../../../storybook-static/stories.json";
```
samtjo, line 12: This `stories.json` file is only generated when running `npx sb extract` so if you try to run any of the Cypress tests locally after just running `npm start` or `npm run build-storybook` they will fail with this error: 

image

That's for any spec, not just the accessibility ones.

So I think we just need to add some docs somewhere (maybe [here](https://github.com/Sage/carbon/blob/master/docs/testing-styleguide.md#automated-testing) ) which tell the user they need to run `npx sb extract` before `npm run test-cypress`

cypress/support/step-definitions/accessibility/a11y-utils.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+
+  // add start and end points for indexes of components
+  urlList = urlList.slice(from, end);
+
+  context("Accessibility tests", () => {
+    describe("Should render storybook component", () => {
+      it.each(urlList)(
+        "should render %s component with %s story and has no accessibility violations",
```
samtjo, line 66:
```suggestion
        "should render %s component with %s story and have no accessibility violations",
```

samtjo (changes requested):
Looks good in general, my main concern is running the tests locally, but some docs explaining how to should be enough IMO

cypress/support/step-definitions/accessibility/a11y-utils.js
samtjo: Actually now I've thought about this a bit more, maybe we should just add `npx sb extract` to the `test-cypress` script as it won't work without running it first anyway.

I also think the other Cypress CI in `cypress.yml` might fail without this, unless we added `npx sb extract` in there as well

cypress/support/step-definitions/accessibility/a11y-utils.js
DlgSHi: updated

.github/workflows/cypress-axe-accessibility.yml
DlgSHi: updated

cypress/support/step-definitions/accessibility/a11y-utils.js
DlgSHi: added an info to the `readme.md` file, also added the `config` prop for running `a11y` tests. This issue no more occurs for all regression tests

cypress/README.md
```diff
@@ -22,16 +22,25 @@ We use the [Cypress.io](https://www.cypress.io) testing framework for functional
 
 Storybook must be running before Cypress tests can be run:
 
-1. Run Storybook `npm start`.
-2. Open a new terminal in the root path of the project.
-3. Run Cypress using the runner with `npx cypress open` or `npm run test-cypress`, then select the required feature file. Test results can be seen directly in the Cypress Test Runner UI.
-4. To run specific Cypress tests at the command line (headless browser for continuous integration) use: `npx cypress run --spec 'cypress/features/[tests-type]/[featureFileName].feature'`. Test results can be seen in the console run summary.
-5. To run in the Chrome/Firefox browser add `--browser chrome` or `--browser firefox` to the above command.
+1. Run `npm run build-storybook` to build the static-storybook folder.
```
samtjo, line 25: We don't need these two steps in this section anymore, now that they're in the accessibility section below

samtjo (approved):

sianford (approved):

cypress.json
jamime: But why are we ignoring `*.spec.js` previously it was explicitly included.

jamime (approved):

PR #4544, opened 2021-11-09
feat(focus-trap): add flag to trigger refocus of either last focused element first element FE-4422

### LINKED ISSUE #4502: Tab focus is set outside parent Dialog component when child Dialog/Confirm component closes
### Current behaviour

When using a Dialog/Confirm component inside a parent Dialog, tab behaviour is as follows:

1. User opens Parent Dialog
2. User tabs through elements inside Parent Dialog
3. User opens either Child/Confirm Dialog
4. User tabs through elements inside Child/Confirm Dialog
5. User closes Child/Confirm Dialog
6. Tab focus is set outside Parent Dialog

https://user-images.githubusercontent.com/77396222/138100819-ebda18a5-af46-48df-812a-70e836a1fecb.mov

### Expected behaviour

Tab focus is expected to be set back inside parent Dialog when child Dialog/Confirm component closes

https://codesandbox.io/s/great-fast-e7of6?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-20824

### Suggested Solution

_No response_

### Carbon Version

91.0.2

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4502

### Proposed behaviour

Add `triggerRefocusFlag` to trigger the refocusing of an element in the trap. Also adds a listener to support
maintaining a record of last focused element which will be focused when the flag is set.

When there are at least two `Modal`s stacked and one closes the `ModalManager` will trigger a
callback to set a flag which is passed to the `FocusTrap` via `ModalContext` and triggers
refocussing of an element within the top most `Modal`

### Current behaviour

Stacked Modals will result in focus not being trapped when the top most modal closes

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/__internal__/focus-trap/focus-trap.component.js
+++ b/src/__internal__/focus-trap/focus-trap.component.js
@@ -28,7 +28,9 @@ const FocusTrap = ({
   const [focusableElements, setFocusableElements] = useState();
   const [firstElement, setFirstElement] = useState();
   const [lastElement, setLastElement] = useState();
-  const { isAnimationComplete } = useContext(ModalContext);
+  const [currentFocusedElement, setCurrentFocusedElement] = useState();
+  const { isAnimationComplete, triggerRefocusFlag } = useContext(ModalContext);
+
   const hasNewInputs = useCallback(
     (candidate) => {
       if (!focusableElements || candidate.length !== focusableElements.length) {
@@ -128,6 +130,43 @@ const FocusTrap = ({
       document.removeEventListener("keydown", trapFn);
     };
   }, [firstElement, lastElement, focusableElements, bespokeTrap]);
+
+  const updateCurrentFocusedElement = useCallback(() => {
+    const element = focusableElements?.find(
+      (el) => el === document.activeElement
+    );
+
+    if (element) {
+      setCurrentFocusedElement(element);
+    }
+  }, [focusableElements]);
+
+  useEffect(() => {
+    document.addEventListener("focusin", updateCurrentFocusedElement);
+
+    return () => {
+      document.removeEventListener("focusin", updateCurrentFocusedElement);
+    };
+  }, [updateCurrentFocusedElement]);
+
+  const refocusTrap = useCallback(() => {
+    /* istanbul ignore else */
+    if (
+      currentFocusedElement &&
+      !currentFocusedElement.hasAttribute("disabled")
+    ) {
+      // the trap breaks if it tries to refocus a disabled element
+      setElementFocus(currentFocusedElement);
+    } else if (firstElement) {
+      setElementFocus(firstElement);
+    }
+  }, [currentFocusedElement, firstElement]);
+
+  useEffect(() => {
+    if (triggerRefocusFlag) {
+      refocusTrap();
+    }
+  }, [triggerRefocusFlag, refocusTrap]);
 
   return <div ref={trapRef}>{children}</div>;
 };
--- a/src/__internal__/focus-trap/focus-trap.spec.js
+++ b/src/__internal__/focus-trap/focus-trap.spec.js
@@ -9,14 +9,25 @@ import { ModalContext } from "../../components/modal/modal.component";
 jest.useFakeTimers();
 
 // eslint-disable-next-line
-const MockComponent = ({ children, ...rest }) => {
+const MockComponent = ({ children, triggerRefocusFlag, ...rest }) => {
   const ref = useRef();
-
+  const [isDisabled, setIsDisabled] = useState(false);
   return (
-    <ModalContext.Provider value={{ isAnimationComplete: true }}>
+    <ModalContext.Provider
+      value={{ isAnimationComplete: true, triggerRefocusFlag }}
+    >
       <FocusTrap wrapperRef={ref} {...rest}>
         <div ref={ref} id="myComponent">
-          {children}
+          {React.Children.map(children, (child) => {
+            if (child?.props?.id === "disable-on-focus") {
+              return React.cloneElement(child, {
+                onFocus: () => setIsDisabled(true),
+                disabled: isDisabled,
+              });
+            }
+
+            return child;
+          })}
         </div>
       </FocusTrap>
     </ModalContext.Provider>
@@ -24,6 +35,7 @@ const MockComponent = ({ children, ...rest }) => {
 };
 
 describe("FocusTrap", () => {
+  let wrapper;
   const element = document.createElement("div");
   const htmlElement = document.body.appendChild(element);
   const tabKey = new KeyboardEvent("keydown", { key: "Tab" });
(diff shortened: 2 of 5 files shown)


DipperTheDan (approved):

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -101,6 +103,7 @@ const FocusTrap = ({
           /* shift + tab */
           if (activeElement === firstElement) {
             lastElement.focus();
+            setElementFocus(lastElement);
```
edleeks87, line 106: I don't think this line is needed I'll remove when  PR has had two reviews

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -129,6 +132,47 @@ const FocusTrap = ({
...
+    /* istanbul ignore else */
+    if (
+      currentFocusedElement &&
+      !currentFocusedElement.hasAttribute("disabled")
+    ) {
+      // last focused element was set and is not disabled
+      // the trap breaks if it tries to refocus a disabled element
+      setElementFocus(currentFocusedElement);
+    } else if (firstElement) {
+      // focus the first focusable element
```
samtjo, line 164: Can remove this comment

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -129,6 +132,47 @@ const FocusTrap = ({
     };
   }, [firstElement, lastElement, focusableElements, bespokeTrap]);
 
+  const updateCurrentFocusedElement = useCallback(() => {
+    const element = focusableElements?.find(
+      (el) => el === document.activeElement
+    );
+
+    // if element matches activeElement update
```
samtjo, line 140: Can remove this comment, the code speaks for itself

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -129,6 +132,47 @@ const FocusTrap = ({
...
+    };
+  }, [updateCurrentFocusedElement]);
+
+  const refocusTrap = useCallback(() => {
+    /* istanbul ignore else */
+    if (
+      currentFocusedElement &&
+      !currentFocusedElement.hasAttribute("disabled")
+    ) {
+      // last focused element was set and is not disabled
```
samtjo, line 160: Can remove the first line of comment here, the second line is useful though

src/__internal__/focus-trap/focus-trap.spec.js
```diff
@@ -34,32 +37,78 @@ describe("FocusTrap", () => {
...
+        document.querySelectorAll("input")[0].focus();
+      });
+      act(() => {
+        document.querySelectorAll("input")[0].blur();
+      });
+      act(() => {
+        wrapper.setProps({ triggerRefocusFlag: true });
+      });
+      wrapper.update();
+      expect(focusMock).toHaveBeenCalledTimes(2);
```
samtjo, line 65: We should be testing that the same input that was focused re-received focus here

src/__internal__/focus-trap/focus-trap.spec.js
```diff
@@ -34,32 +37,78 @@ describe("FocusTrap", () => {
...
     });
 
-    afterEach(() => {
-      wrapper.unmount();
+    it("refocuses the first element within the trap when flag is set", () => {
```
samtjo, line 68: This is just the same test as the first one, with focus on a different element - I think we just need to set the flag here without manually triggering any focus and checking that the first focusable element received focus

src/components/modal/__internal__/modal-manager.js
```diff
@@ -1,26 +1,54 @@
...
+    if (topModal && setTrapFlag) {
+      setTrapFlag(false);
+    }
+
+    this.#modalList.push({ modal, setTriggerRefocusFlag });
   };
 
   isTopmost(modal) {
-    if (!modal || !this.#modalList.length) {
+    if (!modal || !this.#getTopModal().modal) {
```
samtjo, line 25: This is just the same check essentially, why did you change it?

src/components/modal/__internal__/modal-manager.js
```diff
@@ -1,26 +1,54 @@
...
     }
+    const { modal: topModal } = this.#getTopModal();
 
-    return this.#modalList.indexOf(modal) === this.#modalList.length - 1;
+    return modal === topModal;
   }
 
   removeModal(modal) {
-    const modalIndex = this.#modalList.indexOf(modal);
+    const modalIndex = this.#modalList.findIndex(({ modal: m }) => m === modal);
```
samtjo, line 34: Why did you change this?

src/components/modal/__internal__/modal-manager.js
```diff
@@ -1,26 +1,54 @@
...
     this.#modalList.splice(modalIndex, 1);
+
+    if (!this.#modalList.length) {
+      return;
+    }
+
+    const { modal: topModal, setTriggerRefocusFlag } = this.#getTopModal();
+
+    /* istanbul ignore else */
+    if (topModal && setTriggerRefocusFlag) {
```
samtjo, line 49: Saying as we've already returned out the function earlier if `this.#modalList.length` - it shouldn't be possible to get to this line and `topModal` to not be set so we can remove that from the if statement.

A test should be added for when `setTriggerRefocusFlag` is falsey rather than ignoring the coverage here

src/components/modal/__internal__/modal-manager.js
edleeks87: The list has been changed to contain object elements rather than just the refs so the old implementation won't work

src/__internal__/focus-trap/focus-trap.spec.js
edleeks87: Yeah good point there's no need to focus the element like in the test above, I'll remove it and update it to say the callback is called once

src/__internal__/focus-trap/focus-trap.spec.js
edleeks87: That is what's being tested here, the same input is focused as the same callback is called twice

src/components/modal/__internal__/modal-manager.js
edleeks87: Tbh I want to refactor this a little bit to the following, the old implementation won't work as I've updated the list composition so we need to check the modal param against the ref not just the indexed. I could leave it as checking the length of the list but the return at the bottom would potentially evaluate against a topModal that is undefined

```
const { modal: topModal } = this.#getTopModal();
if (!modal || !topModal) {
return false;
}
return modal === topModal;
```

src/components/modal/__internal__/modal-manager.js
edleeks87: Yeah good point, I'll update the if statement and write and else test 👍

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -129,6 +132,47 @@ const FocusTrap = ({
...
+      // focus the first focusable element
+      setElementFocus(firstElement);
+    }
+  }, [currentFocusedElement, firstElement]);
+
+  useEffect(() => {
+    if (triggerRefocusFlag) {
+      refocusTrap();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
edleeks87: I'll give it a try, I decided against it as the refocusTrap will likely update a lot as `currentFocusedElement` updates on each new element focusing but the if guard probably negates that cost

DipperTheDan (dismissed):

samtjo (approved):

samtjo (dismissed):

edleeks87:
Codesandbox https://codesandbox.io/s/withered-flower-19iph?file=/src/index.js for QA to test

ZhuoyuJin (dismissed):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #4545, opened 2021-11-09
refactor(decimal): convert class to functional component - FE-3640

DipperTheDan:
### Proposed behaviour

Refactor component to be a functional component.

### Current behaviour

Component is currently a class based component.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

Codesandboxes for testing
- Manual Testing: https://codesandbox.io/s/decimal-manual-testing-9zf24
- Locale Testing: https://codesandbox.io/s/decimal-locale-testing-5wz6w

### DIFF
--- a/src/components/decimal/decimal.component.js
+++ b/src/components/decimal/decimal.component.js
@@ -1,5 +1,10 @@
-/* eslint-disable react/no-did-update-set-state */
-import React from "react";
+import React, {
+  useState,
+  useEffect,
+  useCallback,
+  useContext,
+  useRef,
+} from "react";
 import PropTypes from "prop-types";
 import styledSystemPropTypes from "@styled-system/prop-types";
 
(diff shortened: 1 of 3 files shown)


src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+      };
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
+
+  /**
+   * Determine if the component is controlled at the time of call
+   */
+  function isControlled(arg) {
```
edleeks87, line 207: I think this should just return `value !== undefined` rather than taking an arg and evaluating it

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
-    return this.props.value !== undefined;
-  }
-
-  isNaN = (value) => {
-    return Number.isNaN(Number(value));
-  };
-
-  getSafeValueProp = (isInitialValue) => {
-    const { value, allowEmptyValue } = this.props;
+  const getSafeValueProp = (isInitialValue) => {
```
edleeks87, line 40: To me `isInitialValue` suggests a boolean, maybe calling it `initialValue` would be a bit more intuitive?

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+  );
+
+  const decimalValue = getSafeValueProp(defaultValue || value || emptyValue);
+  const [stateValue, setStateValue] = useState(
+    isNaN(toStandardDecimal(decimalValue))
+      ? decimalValue
+      : formatValue(decimalValue)
+  );
+
+  const createEvent = useCallback(
```
edleeks87, line 187: I don't think this needs to `useCallback`

src/components/decimal/decimal.component.js
edleeks87: Also if we are going to remove the invariant around switching between controlled/ uncontrolled do we rally need this function at all, could we not just check `value !== undefined` on line 242 instead of ` const isComponentControlled = isControlled(value);`

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+        name={name}
+        value={toStandardDecimal(stateValue)}
+        type="hidden"
+        data-component="hidden-input"
+      />
+    </>
+  );
+};
 
 Decimal.contextType = LocaleContext;
```
edleeks87, line 275: Don't think we need this anymore

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
 
-    const formattedInteger = Intl.NumberFormat(
-      this.props.locale || this.context.locale(),
-      {
+      const formattedInteger = Intl.NumberFormat(locale || l.locale(), {
         maximumFractionDigits: 0,
+      }).format(integer);
+
+      let formattedNumber = formattedInteger;
+      if (remainder && remainder.length > precision) {
```
edleeks87, line 93:
```suggestion
      if (remainder?.length > precision) {
```

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+        formattedNumber += `${separator + remainder}`;
+      } else if (remainder && remainder.length <= precision) {
```
edleeks87, line 95:
```suggestion
      } else if (remainder?.length <= precision) {
```

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+
+  /**
+   * Determine if the component is controlled at the time of call
+   */
+  function isControlled(arg) {
+    return arg !== undefined;
+  }
+
+  const handleOnChange = (ev) => {
+    const val = ev.target.value;
```
edleeks87, line 212: It would be good to use a consistent code styling approach here, I think you should destructure like you have in the `handleOnBlur`

src/components/decimal/decimal.component.js
DipperTheDan: I think you're right. It isn't used in any dependancy arrays. I had used it in the `useEffect` dependancy array at one point in the refactor but removed it in the end so probably just left over from that.

src/components/decimal/decimal.component.js
DipperTheDan: Could be a good shout doing it that way. I'll find out if we need the invariant for controlled/uncontrolled switching and make that changes based on the outcome.

src/components/decimal/decimal.component.js
DipperTheDan: You're right. Functional components can't use `contextType` according to the console warning message I hadn't seen previously.

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+    if (onBlur) onBlur(event);
   };
 
-  getSeparator(separatorType) {
-    const numberWithGroupAndDecimalSeparator = 10000.1;
-    return Intl.NumberFormat(this.props.locale || this.context.locale())
-      .formatToParts(numberWithGroupAndDecimalSeparator)
-      .find((part) => part.type === separatorType).value;
-  }
+  const isComponentControlled = isControlled(value);
```
samtjo, line 237: This const is only used in the `useEffect` below so let's just move it into there

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+      }
+
+      const withoutDelimiters = removeDelimiters(valueWithoutNBS);
+      return withoutDelimiters.replace(new RegExp(`\\${separator}`, "g"), ".");
+    },
+    [getSeparator, removeDelimiters]
+  );
+
+  const decimalValue = getSafeValueProp(defaultValue || value || emptyValue);
+  const [stateValue, setStateValue] = useState(
```
samtjo, line 181: Let's move the `useState` to the top of the component, that's where it would normally go for clarity

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
-    }
+  const formatValue = useCallback(
+    (valueToFormat) => {
+      if (isNaN(valueToFormat)) {
+        return valueToFormat;
+      }
 
-    /* Guards against any white-space only strings like "   " being 
+      /* Guards against any white-space only strings like "   " being 
        mishandled and returned as `NaN` for visibleValue */
```
samtjo, line 80: `visibleValue` no longer exists so this comment needs to change

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+          name,
+          id,
+          value: {
+            formattedValue: formatValue(toStandardDecimal(formatted)),
+            rawValue: raw || toStandardDecimal(formatted),
+          },
+        },
+      };
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
samtjo, line 200: You should remove this disable comment and add all the dependencies eslint tells you to here. Or do the above and remove the `useCallback` completely

src/components/decimal/decimal.component.js
```diff
@@ -11,247 +16,261 @@ import LocaleContext from "../../__internal__/i18n-context";
...
+  useEffect(() => {
+    if (prevPrecisionValue && prevPrecisionValue !== precision) {
+      // eslint-disable-next-line no-console
+      console.error(
+        "Decimal `precision` prop has changed value. Changing the Decimal `precision` prop has no effect."
+      );
     }
-    return formattedNumber;
-  };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
samtjo, line 130: Lets remove this comment and add the deps it tells you to

src/components/decimal/decimal.component.js
DipperTheDan: It can't be moved any higher up the code as it uses the `toStandardDecimal` function. When I tried this originally I got a warning to say that functions cannot be used before they are declared.

src/components/decimal/decimal.component.js
DipperTheDan: I've chosen to remove the `useCallback`. `createEvent` is in any dependancy arrays so there is no real need for it to be a `useCallback`.

src/components/decimal/decimal.component.js
DipperTheDan: Good spot. I'll get this amended.

src/components/decimal/decimal.component.js
DipperTheDan: I'll get this comment removed. Good spot.

src/components/decimal/decimal.component.js
DipperTheDan: I had originally used this const elsewhere hence why it is outside of the `useEffect`. I can move it in there easy enough 👍🏻

edleeks87 (approved):

samtjo (approved):

divyajindel (approved):

samtjo (approved):

divyajindel (approved):

PR #4574, opened 2021-11-18
feat(message): describe Message component using design tokens

mjaniczko-sage:
### Proposed behaviour
Describes the message using design tokens. Removes all possible 'theme' in css files 

### Current behaviour
The Message use the theme styled-component property.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/message/message.style.js
+++ b/src/components/message/message.style.js
@@ -4,13 +4,20 @@ import { margin } from "styled-system";
 import BaseTheme from "../../style/themes/base";
 import StyledIconButton from "../icon-button/icon-button.style";
 
+const messageVariants = {
+  error: "var(--colorsSemanticNegative500)",
+  info: "var(--colorsSemanticNeutral500)",
+  success: "var(--colorsSemanticPositive500)",
+  warning: "var(--colorsSemanticCaution500)",
+};
+
 const MessageStyle = styled.div`
   position: relative;
   display: flex;
   justify-content: flex-start;
   align-content: center;
-  border: 1px solid ${({ theme, variant }) => theme.colors[variant]};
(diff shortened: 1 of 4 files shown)


src/components/message/message.style.js
```diff
@@ -4,13 +4,20 @@ import { margin } from "styled-system";
 import BaseTheme from "../../style/themes/base";
```
samtjo, line 4: You need to remove this import and the `theme` default prop below

samtjo (changes requested):

src/components/message/message.style.js
mjaniczko-sage: When I'm removing theme from there all tests from message.spec.js from function testStyledSystemMargin are failing. I got info that we cannot delete that if styles are overwritten in another component which use theme and when we use margin/space/padding from styled-system since they are using theme.space

samtjo: Ah yeah good point, I didn't notice the styled-system margin in here before, that's fine then 👍🏻

samtjo (approved):

jhajduk-sage (dismissed):

mkrds (dismissed):

samtjo (approved):

samtjo (dismissed):

mkrds (dismissed):

harpalsingh:
the border and the left background should be the same value, but as shown in this visual, you can see the success and default messages borders are darker.

Screenshot 2021-12-10 at 08 20 34

samtjo (approved):

mkrds (approved):

DlgSHi (approved):

Issue #4453, opened 2021-10-04
Typing in DDMMYYYY format in default DateInput is not interpreted as en-GB


liam-cochrane:
### Current behaviour

When I type 01022021 into the default DateInput component, onBlur it is interpreted as 2nd January 2021, rather than 1st February 2021. This can be seen on the storybook default DateInput example.

### Expected behaviour

As I understand, on a default DateInput, with no i18n provider configured, and no i18n properties set on the component, it should be behaving in an en-GB manner, so typing 01022021 should be interpreted as 2nd February 2021. 

### CodeSandbox or Storybook URL

https://carbon.sage.com/?path=/docs/date-input--default-story#default

### Carbon Version

91.2.0

### What browsers are you seeing the problem on?

Firefox

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

jamime:
Hi @liam-cochrane we're actually in the process of requirement gathering for the Date component. There's an open request for comments [here](https://github.com/Sage/carbon/pull/4294).

I'll ensure we add this as a test.

In the interim, I've created a ticket and we will create a fix or provide an updated component.

 FE-4384

Issue #3945, opened 2021-04-26
Date Formatting 


Mahergall:
### Current behaviour
adding 01082020 to the DateInput , in the onChange event this date will be formatted to 01.08.2020 , in the onBlur Event , the date will be formatted to 08.01.2020  
### Expected behaviour
customize in which format the date will be formatted
### Reproducible example
any dateinput Component in the storybook

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| carbon-react     | latest
| carbon-factory   |
| react-scripts    | 
| React            |
| Browser          |
| npm              |
| Operating System |

ianforster-sage:
FE-4056

PR #4576, opened 2021-11-18
fix(date): date formatting when locale overriden (FE-4391)

grabkowski:
### Proposed behaviour
Include moment locales to allow user to use their own locale.
Remove date "sanitization" in `_parseDate` method to enforce ISO date in the `value` prop, which should also fix all day/month swap issues.
Modify custom event generation so the event passed to the `onBlur` contains correct values when the date is correct.

### Current behaviour
Localization for the date component is not working as an error is triggered when a user is trying to implement it.
Date is being incorrectly validated, which leads to errors, when the ISO date is not passed as a valid date in the localization config.

### Checklist

#### QA

### Additional context

additional issues fixed by this commit:
https://github.com/Sage/carbon/issues/3945
https://github.com/Sage/carbon/issues/4453

### Testing instructions

1. run `npm start`
2. open `http://localhost:9001/?path=/story/date-input--default-story`
3. open `http://localhost:9001/?path=/story/date-input--with-custom-date-format`

https://codesandbox.io/s/friendly-burnell-muhwp?file=/src/index.js

Invariant Violation when a date with slashes is not added in valid date list:

https://codesandbox.io/s/hungry-goldstine-f7pip


samtjo (approved):

src/components/date/date.component.js
```diff
@@ -141,13 +141,10 @@ class BaseDateInput extends React.Component {
...
 
       if (this.props.onBlur && !this.state.isDatePickerOpen) {
-        const dateWithSlashes = DateHelper.sanitizeDateInput(
-          this.state.visibleValue
-        );
         const event = this.buildCustomEvent(
           { target: this.input, type: "blur" },
           DateHelper.formatValue({
-            value: dateWithSlashes,
+            value: this.state.visibleValue,
```
edleeks87, line 147: I think this will prevent the date from auto formatting on blur, it will just leave the value in the input as typed which is not the correct behaviour

src/components/date/date.component.js
```diff
@@ -315,9 +313,8 @@ class BaseDateInput extends React.Component {
   handleVisibleInputChange = (ev) => {
     const { disabled, readOnly } = this.props;
     const value = ev.target.value.formattedValue || ev.target.value;
-    const dateWithSlashes = DateHelper.sanitizeDateInput(value);
     const isValidDate = DateHelper.isValidDate({
-      value: dateWithSlashes,
+      value,
```
edleeks87, line 317: I think this change will prevent some valid input formats from setting the selectedDate in the picker

src/components/date/date.component.js
grabkowski: we do not need to change the value's delimiters to slashes and then change them back again to the proper value, as the format method does tat anyway. By changing them to slashes we were introducing an issue, in which when an user did not specified a date containing slashes in the list of valid formats, an error would be thrown, as moment would compare the incorrect slashes date instead of what user typed in.

src/components/date/date.component.js
grabkowski: it would prevent setting a date in a format not specified by an user, if they decide to add their own valid formats

grabkowski: another issue with that is when a user is setting their own list of valid formats that is not containing the one with slashes, this method would throw an error for correct dates!

edleeks87: Yeah but it prevents previously reformatted values from being formatted as well, I think it would be better to add a guard against attempting to pass an `invalid date` to the picker if the format is custom etc

src/components/date/date.component.js
edleeks87: This is not how the date should behave, IMHO it needs to reformat to the locale based format when the input blurs, currently it doesn't set the picker or reformat and then on refocusing the picker it jumps back to different date

How it behaves in master:
https://user-images.githubusercontent.com/44157880/142845573-6970d2ec-88c2-4c2b-9af2-65c216cff38c.mov

After your changes:
https://user-images.githubusercontent.com/44157880/142845628-c8275b3c-bdc3-4483-92c6-0d557ece38dd.mov

edleeks87: This is not how the date should behave, IMHO it needs to reformat to the locale based format when the input blurs, currently it doesn't set the picker or reformat and then on refocusing the picker it jumps back to different date

How it behaves in master:


After your changes:

src/components/date/date.component.js
grabkowski: moment already does it for us, there is no need to convert dots, hyphens and spaces into slashes

src/components/date/date.component.js
grabkowski: looks like it was already broken, but we were enforcing it to have slashes, which shouldn't be done. Good spot!

grabkowski: I'm taking that back, the date with dot delimiter is not a valid date, as it's not declared in the valid date list

edleeks87 (approved):

samtjo (approved):

edleeks87 (approved):

DlgSHi (approved):

PR #4592, opened 2021-11-25
test: add react tests

DlgSHi:
### Proposed behaviour

We need to try and implement tests that won't depend on storybook, and use `components` as they are used by `end User`

Implement new `cypress` approach for `react` testing by `mount`'ing `component` and passing `props` into 

```Javascript
<Component `props`> 
```

like: 

```Javascript
mount(<Component readOnly/>);
```

Add possibility to check all events (with `Object` properties check)

```Javascript
// use cy.stub for mock the onChange callback
const callback = cy.stub();
// and mount the story using @cypress/react library
mount(<Decimal onChange={callback} />);
```

and then check the `onChange` event `Object` like:
```Javascript
.then(() => {
    // eslint-disable-next-line no-unused-expressions
    expect(callback).to.have.been.calledThrice;
     iterable.map((item, index) => {
         expect(callback.getCalls()[index].args[0].target.value.rawValue).to.equals(item[0]);
         expect(callback.getCalls()[index].args[0].target.value.formattedValue).to.equals(item[1]);
         return this;
         });
      });
```

Check: 
- `src/components/app-wrapper/app-wrapper.test.js`
- `src/components/decimal/decimal.test.js`

the time for running `app-wrapper.test.js` and `decimal.test.js` after refactor:
Screenshot 2021-10-21 at 12 24 11

### Current behaviour

We need to start storybook to test components

Currently the time for running `appWrapper.feature`, `decimalInput.feature`, `decimalPrecision.feature`,
`decimalInput.feature` is around 1:40 min ~ 2:00 mins + time from starting `storybook`.

Screenshot 2021-08-25 at 08 20 54

### Checklist

### Testing instructions

`run:`
- npm install
- npm run cypress:react


cypress.json
```diff
@@ -23,5 +23,10 @@
   "retries": {
     "runMode": 1,
     "openMode": 1
-  }
+  },
+   "component": {
+     "componentFolder": "src",
+     "testFiles": "**/*.test.js",
+     "watchForFileChanges": true
+   }
```
sianford, line 31: Space before the `}` to remove.

DlgSHi: I think it should be fixed separately :) lets get merge this one :)

package.json
```diff
@@ -18,6 +18,8 @@
     "test": "jest --config=./jest.conf.json",
     "test-update": "jest --config=./jest.conf.json --updateSnapshot",
     "test-cypress": "cypress open",
+    "cypress:react": "npx cypress open-ct",
```
jamime, line 21: I prefer the `:` syntax that we have here, but we should update the other commands to match in a different PR.

package.json
```diff
@@ -147,6 +152,7 @@
     "classnames": "~2.2.6",
     "crypto-js": "~3.3.0",
     "dotenv": "^10.0.0",
+    "har-validator": "^5.1.5",
```
jamime, line 155: We shouldn't be adding this as a dependency. We should wait for the cypress fix.

cypress/plugins/index.js
```diff
@@ -17,32 +17,78 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
+                      "@babel/preset-env",
+                      "@babel/preset-react",
+                      "@babel/preset-typescript",
+                    ],
+                  },
+                },
+              ],
+            },
+            {
+              test: /\.css$/i,
```
jamime, line 79: We shouldn't need this

cypress/plugins/index.js
```diff
@@ -17,32 +17,78 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
+              use: ["style-loader", "css-loader"],
+            },
+            {
+              test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
jamime, line 83: We shouldn't need this

cypress/support/helper.js
```diff
@@ -11,7 +11,10 @@ function prepareUrl(component, suffix = "default story") {
 }
 
 export function visitComponentUrl(component, suffix) {
-  cy.visit(prepareUrl(component, suffix));
+  const spy = cy.spy();
```
jamime, line 14:
```suggestion
  const onBeforeLoad = cy.spy();
```

cypress/support/helper.js
```diff
@@ -11,7 +11,10 @@ function prepareUrl(component, suffix = "default story") {
...
+  cy.visit(prepareUrl(component, suffix), {
+    onBeforeLoad: spy,
```
jamime, line 16:
```suggestion
    onBeforeLoad
```

cypress/support/helper.js
```diff
@@ -22,6 +25,7 @@ export function visitComponentUrlWithParameters(
   path = "",
   nameOfObject = ""
 ) {
+  const spy = cy.spy();
```
jamime, line 28:
```suggestion
  const onBeforeLoad = cy.spy();
```

cypress/support/helper.js
```diff
@@ -32,7 +36,9 @@ export function visitComponentUrlWithParameters(
         url += `${prop}:${encodeURIComponent(el[prop])};`;
       }
     }
-    cy.visit(`${prepareUrl(component, story)}${url}`);
+    cy.visit(`${prepareUrl(component, story)}${url}`, {
+      onBeforeLoad: spy,
```
jamime, line 40:
```suggestion
      onBeforeLoad
```

src/components/decimal/decimal.test.js
```diff
@@ -0,0 +1,290 @@
...
+
+    it.each(input)(
+      "should use %s as precision and %s as input value to produce %s output value",
+      (precision, inputValue, outputValue) => {
+        mount(<Decimal precision={precision} />);
+
+        // then run our tests
+        commonDataElementInputPreview().type(inputValue).blur({ force: true });
+        commonDataElementInputPreview().should("have.value", outputValue);
+        return this;
```
edleeks87, line 102: I don't think we need to return anything in these it blocks

cypress.json
DlgSHi: done

package.json
DlgSHi: done

package.json
DlgSHi: done

cypress/plugins/index.js
DlgSHi: done

cypress/plugins/index.js
DlgSHi: done, but I think we might need this in future

cypress/support/helper.js
DlgSHi: done

cypress/support/helper.js
DlgSHi: done

cypress/support/helper.js
DlgSHi: done

cypress/support/helper.js
DlgSHi: done

src/components/decimal/decimal.test.js
DlgSHi: done

sianford (approved):

edleeks87 (approved):

cypress/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "types": [
       "cypress",
       "cypress-real-events",
-      "cypress-axe"
+      "cypress-axe",
```
samtjo, line 9: Why do we have a `tsconfig.json` file in the Cypress folder? We don't have any tests written in TypeScript - This whole file can be removed

.github/workflows/cypress.yml
```diff
@@ -107,6 +107,74 @@ jobs:
...
+  cypress-react:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    strategy:
+      # when one test fails, DO NOT cancel the other
+      # containers, because this will kill Cypress processes
+      # leaving the Dashboard hanging ...
+      # https://github.com/cypress-io/github-action/issues/48
+      fail-fast: false
+      matrix:
```
samtjo, line 119: I don't think we need a matrix at all for just running 1 copy of the job, this whole `strategy` bit can be removed

cypress/plugins/index.js
```diff
@@ -17,32 +17,70 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
     }
+
     return launchOptions;
   });
+
+  if (config.testingType === "component") {
+    // eslint-disable-next-line global-require
+    require("@cypress/react/plugins/babel")(
+      on,
+      { ...config, addTranspiledFolders: [".storybook"] },
```
samtjo, line 58: Why do we need the `.storybook` folder for component testing? I've tried removing this locally and running the tests and they work fine without it

samtjo (changes requested):
Good job mate this component testing stuff looks great 👍🏻  Just a few things to remove then I'll give it a ✅ 

We're also inadvertently adding the `.test.js` files into the build package that will be deployed to npm - we need to ignore `.test.js` files in the babel script in the package.json here: https://github.com/Sage/carbon/blob/64610d757e00e356a6fe31f92a0286aab9fce4c1/package.json#L30

cypress/tsconfig.json
DlgSHi: it is just a preparation for migration :)

.github/workflows/cypress.yml
DlgSHi: nope, because after refactor we will still use those 20 machines

cypress/plugins/index.js
DlgSHi: there was some issue, but I will recheck it

cypress/tsconfig.json
samtjo: Even if we convert Cypress tests to TypeScript (which I don't think we should unless we teach all QAs TypeScript on top of JavaScript), it will not be for a long time so we don't need this file yet. 

YAGNI principle 😉  https://betterprogramming.pub/yagni-you-aint-gonna-need-it-f9a178cd8e1

samtjo: Having said that we might actually be forced to write the tests in TypeScript with this approach which isn't ideal 🤔

.github/workflows/cypress.yml
samtjo: Yeah fair enough actually

cypress/plugins/index.js
```diff
@@ -17,32 +17,70 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
+  if (config.testingType === "component") {
+    // eslint-disable-next-line global-require
+    require("@cypress/react/plugins/babel")(
+      on,
+      { ...config },
+      {
+        setWebpackConfig: (webpackConfig) => {
+          webpackConfig.module.rules = [
+            {
+              test: /\.(tsx?)|(js)$|(jsx)$/,
```
grabkowski, line 63: what about `ts` files?

grabkowski: I think that it should be `\.(tsx?|jsx?)$`

grabkowski: shouldn't we also narrow it down to a specific folder?

cypress/plugins/index.js
```diff
@@ -17,32 +17,70 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
     }
+
     return launchOptions;
   });
+
+  if (config.testingType === "component") {
+    // eslint-disable-next-line global-require
+    require("@cypress/react/plugins/babel")(
+      on,
+      { ...config },
```
samtjo, line 58: This can just be:

```suggestion
      config,
```

cypress/plugins/index.js
```diff
@@ -17,32 +17,70 @@ const cucumber = require("cypress-cucumber-preprocessor").default;
...
+      {
+        setWebpackConfig: (webpackConfig) => {
+          webpackConfig.module.rules = [
+            {
+              test: /\.(tsx?)|(jsx?)$/,
```
samtjo, line 63: What about ts and js files?

package.json
```diff
@@ -27,7 +29,7 @@
     "watch": "npm run clean-lib && npm run copy-files -- --watch & npm run babel -- --watch",
     "build-storybook": "build-storybook -c .storybook -s .assets",
     "start:static": "npx http-server -p 9001 ./storybook-static",
-    "babel": "cross-env NODE_ENV=production babel ./src --config-file ./babel.config.js --out-dir ./lib --ignore '**/*/__spec__.js','**/*.spec.js','**/__definition__.js' --quiet",
+    "babel": "cross-env NODE_ENV=production babel ./src --config-file ./babel.config.js --out-dir ./lib --ignore '**/*/__spec__.js','**/*.spec.js','**/__definition__.js', '**/*.test.js' --quiet",
```
samtjo, line 32:
```suggestion
    "babel": "cross-env NODE_ENV=production babel ./src --config-file ./babel.config.js --out-dir ./lib --ignore '**/*/__spec__.js','**/*.spec.js','**/__definition__.js','**/*.test.js' --quiet",
```

samtjo: This doesn't actually work with a space in it 😅

samtjo (changes requested):

cypress/plugins/index.js
DlgSHi: the `x?` is an optional, and will cover the `js/ts` files

samtjo: Ah fair enough, regex'a aren't my strong point!

samtjo (approved):

grabkowski (approved):

sianford (approved):

PR #4617, opened 2021-12-02
fix(flat-table-cell): ensure that pl prop is correctly applied to a sub row - FE-4306

### LINKED ISSUE #4348: Padding Left pl property  is not working for FlatTableCell in SubRows
### Desired behaviour

Padding Left (pl) property is not working for FlatTableCell  when used in SubRows of the FlatTable component.
But the pl (padding-left) property works fine for FlatTableCell when used in normal Rows in a FlatTable component.
Here is the CodeSandbox example to replicate the issue.
https://codesandbox.io/s/flattable-subrows-padding-example-4i6ni

### Current behaviour

Add padding-left with pl property  to the FlatTableCell when used in SubRows of FlatTable Component
Padding Left (pl) property is not working for FlatTableCell  when used in SubRows of the FlatTable component.
But the pl (padding-left) property works fine for FlatTableCell when used in normal Rows in a FlatTable component.
Here is the CodeSandbox example to replicate the issue.
https://codesandbox.io/s/flattable-subrows-padding-example-4i6ni

### Suggested Solution

_No response_

https://codesandbox.io/s/flattable-subrows-padding-example-4i6ni

_No response_

### Confidentiality

DipperTheDan:
Increase specificity of padding props so that the pl prop is correctly applied to a flat table cell
when it is a sub row.

fixes #4348

### Proposed behaviour

pl prop should apply the desired padding-left to a flat table cell when it is nested in a sub row.

### Current behaviour

pl prop is not being applied to a flat table cell when it is nested in a sub row.

### Checklist

#### QA

### Testing instructions

Please use the codesandboxes provided [here](https://github.com/Sage/carbon/pull/4617#issuecomment-984547871)

### DIFF
--- a/src/components/flat-table/flat-table-test.stories.mdx
+++ b/src/components/flat-table/flat-table-test.stories.mdx
@@ -15,7 +15,7 @@ import {
 } from ".";
 import guid from "../../__internal__/utils/helpers/guid";
 import { FLAT_TABLE_SIZES, FLAT_TABLE_THEMES } from "./flat-table.config";
-import { FlatTableSortableStory } from "./flat-table.stories.mdx";
+import { FlatTableSortableStory } from "./flat-table.stories.mdx"
 
 <Meta
   title="Flat Table/Test"
--- a/cypress/fixtures/commonComponents/flatTable.json
+++ b/cypress/fixtures/commonComponents/flatTable.json
@@ -9,5 +9,17 @@
     "hasHeaderRow": true,
     "hasStickyHead": true,
     "hasClickableRows": true
+  },
+  "pl1": {
+    "pl": 1
+  },
+  "pl4": {
+    "pl": 4
+  },
+  "pl25px": {
+    "pl": "25px"
+  },
+  "pl6em": {
+    "pl": "6em"
   }
 }
--- a/cypress/integration/common/flatTable.feature
+++ b/cypress/integration/common/flatTable.feature
@@ -291,3 +291,15 @@ Feature: FlatTable component
       And I wait 100
     When I click "Expand All" button on preview
     Then The subrows are visible
+  
+  @positive
+  Scenario Outline: Correct padding left value of <nameOfObject> is applied to a sub row
+    Given I open default "Flat Table Expandable" component with "flatTable" json from "commonComponents" using "<nameOfObject>" object name
+    When I click on the first cell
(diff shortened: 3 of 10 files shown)


edleeks87 (approved):
I don't love the approach of adding an extra `&` but I can't think of a different way that wouldn't require a lot of refactoring (which feels like way more work than the ticket outlined). The cypress tests you've added will let us know if it breaks again and if so we should explore refactoring to fix at that point to avoid hundreds of `&`

DipperTheDan:
> I don't love the approach of adding an extra `&` but I can't think of a different way that wouldn't require a lot of refactoring (which feels like way more work than the ticket outlined). The cypress tests you've added will let us know if it breaks again and if so we should explore refactoring to fix at that point to avoid hundreds of `&`

Yeah I agree. It was the best option I could think of that didn't involved an excessive amount of code to maintain really. I also agree that if we need to add anymore `&` to increase the specificity, we should really look further in to why this keeps happening with a flat table cell when it is nested in a sub row. With the Cypress test we should know if a PR will introduce this bug again also.

src/components/flat-table/flat-table-test.stories.mdx
```diff
@@ -15,7 +15,8 @@ import {
 } from ".";
 import guid from "../../__internal__/utils/helpers/guid";
 import { FLAT_TABLE_SIZES, FLAT_TABLE_THEMES } from "./flat-table.config";
-import { FlatTableSortableStory } from "./flat-table.stories.mdx";
+import { FlatTableSortableStory } from "./flat-table.stories.mdx"
```
mkrds, line 18: I think we have decided that having `test` stories in MDX is a bad idea and we're slowly rewriting them to be `.js` again, if it's not a problem could you please do it in this PR?

mkrds:
Looks good to me, just the MDX -> JS rewrite part, but it's up to you if you'd like to do this in this PR

DipperTheDan:
> Looks good to me, just the MDX -> JS rewrite part, but it's up to you if you'd like to do this in this PR

Just because it's you asking, I'll do it 😉 
Shouldn't be a difficult job adding it in as a new commit 👍🏻

mkrds (dismissed):

divyajindel (approved):

edleeks87 (dismissed):

divyajindel (approved):

nicktitchmarsh (approved):

PR #4629, opened 2021-12-08
fix(button): pass aria-hidden='true' to icon (FE-4477)

### LINKED ISSUE #4579: Button with Icon: Add aria-hidden="true" to span wrapping icon
### Current behaviour

Presently the span element that wraps the icon within a button is considered as the first character of the button text preventing the Voice Control tool on Mac to recognise the button text when scanning the page for the phrase spoken by the user.

### Expected behaviour

When the user speaks the button text phrase, it should be recognised on the page so the user can use voice commands to click the button.

N/A

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

91.0.2

Safari

MacOS

Adding `aria-hidden="true"` to the `<span>` element that wraps the icon within a button will ensure the Voice Control tool or any other assistive technology does not consider the icon to be part of the button text while scanning the page for the phrase spoken by the user and so the user is able to click button using the Voice Control tool.

### Confidentiality

Parsium:
closes #4579

### Proposed behaviour

Have `Button` control the visibility of its icon to assistive technologies by passing down an `aria-hidden` prop - setting it to `true` if the _button has an icon_ and is _not icon-only_.

This is to ensure voice control tools don't deem the icon as a part of the button text, so the user can use voice commands on `Button` like normal, such as clicking on the button when its text is spoken.

### Current behaviour

The icon within `Button` is recognised as part of the button text by Voice Control on Mac in Safari. Therefore when the user speaks the button text as the element to target via a voice command, Voice Control does not recognise it.

### Checklist

#### QA

### Additional context

Voice Control on Mac and Safari were used in the original raised issue.

### Testing instructions

Open Storybook and test using Voice Control on `Button` with icon. Check that voice commands where the button text is spoken (e.g. "click submit" for clicking on a button with text "Submit") are recognised.


mkrds (dismissed):

edleeks87 (dismissed):

src/components/button/button.component.js
```diff
@@ -29,6 +29,7 @@ function renderChildren({
   };
 
   const iconProps = {
+    "aria-hidden": !!children,
```
grabkowski, line 32: I think that it could be always hidden, as the icon in button should always be ignored by readers

Parsium: On further discussion, `Icon` should always be hidden and `aria-label` should be set to `Button` instead. This is to avoid being able to focus on the inner icon.

mkrds (approved):

grabkowski (approved):

divyajindel (approved):

Issue #4631, opened 2021-12-09
Select - The focus cannot be removed once it's disabled


MarcServat:
### Current behaviour

Given the `Select` component
When it is focused and disabled 
Then the focus cannot be removed

For the codesandbox:

1. Select two options in the MultiSelect
2. Press Tab to move the focus to the Reset button

As a result, both elements are focused.

![image](https://user-images.githubusercontent.com/11317965/145381031-f9c8c3e8-470f-4530-afae-c161d1615650.png)

### Expected behaviour

Once the `Select` component is disabled the focus should be removed.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/wandering-feather-vtqi0?file=/src/index.js:875-886

### Carbon Version

102.0.1

### What browsers are you seeing the problem on?

Firefox, Chrome

### What Operating System are you seeing the problem on?

MacOS

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

samtjo:
This should just be a case of removing the focus CSS when the disabled prop is set to true. If the Select is currently focused when it gets disabled I think focus should be moved to the next focusable element on the page (although let's check that bit with UX).

samtjo:
FE-4631

PR #4643, opened 2021-12-14
fix: remove border style when the input is disabled

MarcServat:
### Proposed behaviour

This PR tries to solve the issue [4631](https://github.com/Sage/carbon/issues/4631). The behaviour proposed when an `Input` component gets disabled then the outline is removed and the focus is passed to the following element in the DOM. 
 

### Current behaviour

When an `Input` component (or components that contains an `Input`) are disabled then the focus remains in the disabled element, which is the `Input` and once `tab` is pressed the focus is removed however the outline sticks to the element. That create two elements with an outline.

image

### Checklist

#### QA

### Additional context

None

### Testing instructions

1. Make sure the outline is removed from the `Input` when the component is disabled.
2. Add different elements surrounding an `Input` and once it's disabled, the next focusable element should be focused.

https://codesandbox.io/s/affectionate-pine-vcox0?file=/src/index.js


src/__internal__/utils/helpers/browser/browser.js
```diff
@@ -188,6 +188,31 @@ const Browser = {
...
+   * @param {Node} refNode
+   */
+  focusNextElement: (refNode) => {
+    const treeWalker = document.createTreeWalker(
+      Browser.getDocument().body,
+      NodeFilter.SHOW_ELEMENT,
+      {
+        acceptNode(node) {
+          return node.tabIndex >= 0 &&
+            // eslint-disable-next-line no-bitwise
```
MarcServat, line 204: please, give a hint which option you'd like to use to avoid comment.
https://eslint.org/docs/rules/no-bitwise#options

src/__internal__/input/input.component.js
```diff
@@ -43,6 +44,12 @@ const Input = React.forwardRef(
       if (context.inputRef) context.inputRef(input);
     }, [context, input]);
 
+    useEffect(() => {
+      if (disabled) {
+        Browser.focusNextElement(input.current);
```
samtjo, line 49: I think we need to check whether the input is actually focused when it's disabled before focusing the next element, as we don't want to do this if it isn't

samtjo: As an example see this sandbox: https://codesandbox.io/s/carbon-quickstart-forked-pj15g?file=/src/index.js 

It's a simple Form where the third input is disabled. As you can see when the page loads, the focus immediately jumps to the fourth input because the third on is disabled

samtjo (changes requested):
I think we need to check whether the input is actually focused when it's disabled before focusing the next element, as we don't want to do this if it isn't.

As an example see this sandbox: https://codesandbox.io/s/carbon-quickstart-forked-pj15g?file=/src/index.js (which is using your changes).

It's a simple Form where the third input is disabled. As you can see when the page loads, the focus immediately jumps to the fourth input because the third one is disabled.

samtjo:
Hi @MarcServat - have you checked with UX that this is the correct behaviour here? Default HTML behaviour would not focus the next element when a currently focused element is disabled.

Also looking at your sandbox on this PR with the current changes in, the functionality doesn't appear to work as you intended it to - disabling the select when it is focused does not focus the next element.

If/when you check with UX, you might just be able to remove all the code here focusing the next element anyway if that isn't the correct behaviour. I recommend checking with @harpalsingh and @tempertemper

src/__internal__/utils/helpers/browser/browser.js
samtjo: This approach is fine 👍🏻

MarcServat:
> Hi @MarcServat - have you checked with UX that this is the correct behaviour here? Default HTML behaviour would not focus the next element when a currently focused element is disabled.

@samtjo I spoke to @tempertemper and makes sense to keep the current HTML behavior. So I guess, if it's fine for you guys, I'll remove the code which focuses the next element when the `Input` gets disabled and I'll leave the code that removes the orange border from the disabled `Input`.
 
> Also looking at your sandbox on this PR with the current changes in, the functionality doesn't appear to work as you intended it to - disabling the select when it is focused does not focus the next element.

I think this is because codesandbox wraps the code inside of an iframe, and the `document` is not the expected one.

samtjo:
> @samtjo I spoke to @tempertemper and makes sense to keep the current HTML behavior. So I guess, if it's fine for you guys, I'll remove the code which focuses the next element when the `Input` gets disabled and I'll leave the code that removes the orange border from the disabled `Input`. Let me know if you agree on this, please.
> 
> > Also looking at your sandbox on this PR with the current changes in, the functionality doesn't appear to work as you intended it to - disabling the select when it is focused does not focus the next element.
> 
> I think this is because codesandbox wraps the code inside of an iframe, and the `document` is not the expected one.

Hi @MarcServat - that approach sounds good to me 👍🏻  Thanks for checking with UX

samtjo (approved):

mkrds (approved):

samtjo (approved):

PR #4646, opened 2021-12-15
fix(alert): fix incorrect prop type for title prop - FE-4490

### LINKED ISSUE #4600: Alert - Typescript - children is not declared as a valid prop
### Current behaviour

When using the Alert component where the title prop is a ReactElement, we receive this typescript error:
```typescript
Type '{ children: string; "data-role": string; onCancel: () => void; open: boolean; size: string; title: Element; }' is not assignable to type 'IntrinsicAttributes & DialogProps'.
  Property 'children' does not exist on type 'IntrinsicAttributes & DialogProps'.
```

### Expected behaviour

children to be a valid prop and to match: https://github.com/Sage/carbon/blob/master/src/components/alert/alert.component.js#L14

n/a

### JIRA Ticket (Sage Only)

SBS-24897

### Suggested Solution

_No response_

### Carbon Version

95.1.1

Other

MacOS

_No response_

### Confidentiality

DipperTheDan:
The prop type for title within dialog is node, where as in Alert we have specified a string or object must be used.

fixes #4600

### Proposed behaviour

Alert, Dialog and Modal have correct prop types declared and their respective d.ts files should also reflect this.

### Current behaviour

Alert, Dialog and Modal have prop types declared that do not exist in their respective d.ts files.

### Checklist

#### QA

### Additional context

N\A

### Testing instructions


edleeks87 (dismissed):

src/components/modal/modal.d.ts
```diff
@@ -1,16 +1,22 @@
...
   enableBackgroundUI?: boolean;
   /** A custom close event handler */
   onCancel?: (ev: React.KeyboardEvent<HTMLElement>) => void;
   /** Sets the open state of the modal */
   open: boolean;
+  /** Transition time */
+  timeout?: number;
 }
 
 declare class Modal<T extends ModalProps> extends React.Component<
```
samtjo, line 22: This needs to change as `Modal` is no longer a class anymore - I think it may be the root cause of the `children` problem. 

The `React.Component` type we're extending here has `children` already defined on it I believe so we shouldn't need to specifically add it

src/components/dialog/dialog.d.ts
```diff
@@ -2,6 +2,8 @@ import * as React from "react";
 import Modal, { ModalProps } from "../modal/modal";
 
 export interface DialogProps extends ModalProps {
+  /** Dialog content */
+  children?: React.ReactNode;
```
samtjo, line 6: If we're extending `ModalProps` and that has `children` on it, we don't need this

samtjo (changes requested):

src/components/dialog/dialog.d.ts
DipperTheDan: Good spot. I worked backwards from `Alert` through `Dialog` to `Modal` so it didn't occur to me 😶‍🌫️

edleeks87 (dismissed):

samtjo (approved):

samtjo (dismissed):

divyajindel (dismissed):

divyajindel (approved):

nicktitchmarsh (approved):

PR #4678, opened 2022-01-05
fix(split-button): apply type of button to split button toggle - FE-4969

### LINKED ISSUE #4651: SplitButton dropdown toggle button inside a form triggers submit event when clicked
### Current behaviour

When a `<SplitButton>` is used within a `<form>` (either post or get), clicking on, rather than just hovering over the dropdown toggle button causes a submit event.

This is because there is no `type` attribute set on the button, so the browser default whilst within a form is to treat the button as a submit button.

Our `<SplitButton>`s outside of forms are not impacted as the default here for browsers is not to treat them as a submit.

### Expected behaviour

On clicking the dropdown toggle button, the dropdown should be displayed and the dropdown toggle button should not fire a submit event.

https://codesandbox.io/s/blissful-hermann-xbwb4

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

Set `type="button"` on the dropdown toggle button.

### Carbon Version

95.1.0

Windows, iOS

On Sage 200 we are upgrading to Carbon 95.1.0 and assume this is a regression introduced when `<SplitButton>` became a functional component. We're using a wrapper around this component as a temp fix.

### Confidentiality

DipperTheDan:
Ensures that a type of button is specified on the split button toggle. This is to make sure that
when split button is used inside of a HTML form, the split toggle button does not automatically
submit the form when clicked.

fixes #4651

### Proposed behaviour

- `Split Button Toggle` has a type of `button`.

### Current behaviour

- `Split Button Toggle` does not a type specified so when `Split Button` is used in a Form, the toggle button acts as a submit button by default.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- Codesandbox built using the example in Github issue can be found [here](https://github.com/Sage/carbon/pull/4678#issuecomment-1005928848).

### DIFF
--- a/src/components/split-button/split-button-test.stories.mdx
+++ b/src/components/split-button/split-button-test.stories.mdx
@@ -96,7 +96,7 @@ export const SplitButtonStory = ({
   );
 };
 
-# Alert
+# Split Button
 
 ### Default
 
--- a/src/components/split-button/__snapshots__/split-button.spec.js.snap
+++ b/src/components/split-button/__snapshots__/split-button.spec.js.snap
@@ -427,6 +427,7 @@ exports[`SplitButton for business themes renders styles correctly 1`] = `
     onMouseEnter={[Function]}
     onTouchStart={[Function]}
     size="medium"
+    type="button"
   >
     <span
       className="c4 c5"
--- a/src/components/split-button/split-button.component.js
+++ b/src/components/split-button/split-button.component.js
@@ -185,6 +185,7 @@ const SplitButton = ({
         aria-label="Show more"
         data-element="toggle-button"
         key="toggle-button"
+        type="button"
         {...toggleButtonProps()}
       >
         <Icon
--- a/src/components/split-button/split-button.spec.js
+++ b/src/components/split-button/split-button.spec.js
@@ -106,7 +106,7 @@ const renderWithNoChildren = (mainProps = {}, renderer = shallow) => {
 
 const buildSizeConfig = (name, size) => {
   const sizeObj = {};
-  sizeObj.fontSizze = size === "large" ? "16px" : "14px";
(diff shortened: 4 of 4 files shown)


src/components/split-button/split-button-test.stories.mdx
```diff
@@ -121,3 +121,22 @@ export const SplitButtonStory = ({
     {SplitButtonStory.bind({})}
   </Story>
 </Canvas>
+
+### In a Form
```
samtjo, line 125: I appreciate what you're trying to do here but I don't think we really need a story to prove that there's no bug tbh, it doesn't really provide any value to our docs for consumers. I think we can just remove this 👍🏻 

A codesandbox will be enough for QA to test this on

edleeks87: Yeah agree with Sam the snapshot will guard against this being removed so the story is overkill

DipperTheDan: That's fine I can get it removed. I figured having something like this in the `test` directory would give QA something to regression test in case this bug were to reappear again. At least if it's in the `test` directory it wouldn't be published to `carbon.sage.com`.

samtjo: A unit test checking that the button always has a type of `button` would be enough for regression here if you wanted to add something

edleeks87 (approved):

samtjo (dismissed):

divyajindel (approved):

edleeks87 (approved):

samtjo (dismissed):

divyajindel (approved):

divyajindel (approved):

samtjo (approved):

divyajindel (approved):

PR #4756, opened 2022-01-27
feat(dialog): surface contentPadding prop to allow overriding default padding values dialog content FE-4758

edleeks87:
### Proposed behaviour

Surfaces `contentPadding` prop which has an interface that partially supports the styled-system
padding `p`, `py` and `px` props

### Current behaviour

No support for adjusting padding of `Dialog` content.

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/modest-forest-fzd81?file=/src/index.js

### DIFF
--- a/src/components/dialog/dialog.component.js
+++ b/src/components/dialog/dialog.component.js
@@ -36,6 +36,7 @@ const Dialog = ({
   disableClose,
   help,
   role = "dialog",
+  contentPadding = {},
   ...rest
 }) => {
   const locale = useLocale();
@@ -194,10 +195,11 @@ const Dialog = ({
           data-element="dialog"
           data-role={rest["data-role"]}
           role={role}
+          {...contentPadding}
         >
           {dialogTitle()}
-          <DialogContentStyle>
-            <DialogInnerContentStyle ref={innerContentRef}>
+          <DialogContentStyle {...contentPadding}>
+            <DialogInnerContentStyle ref={innerContentRef} {...contentPadding}>
               {React.Children.map(children, (child) => {
                 if (child?.type === Form) {
                   return React.cloneElement(child, {
@@ -224,7 +226,7 @@ Dialog.propTypes = {
    */
   "aria-label": PropTypes.string,
   /**
-   * Prop to specify the aria-labeledby property of the Dialog component
+   * Prop to specify the aria-labelledby property of the Dialog component
    * To be used when the title prop is a custom React Node,
    * or the component is labelled by an internal element other than the title.
    */
@@ -274,6 +276,12 @@ Dialog.propTypes = {
   bespokeFocusTrap: PropTypes.func,
   /** The ARIA role to be applied to the Dialog container */
   role: PropTypes.string,
+  /** Padding to be set on the Dialog content */
+  contentPadding: PropTypes.shape({
+    p: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8]),
+    px: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8]),
+    py: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8]),
+  }),
 };
 
 Dialog.defaultProps = {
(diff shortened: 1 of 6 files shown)


src/components/dialog/dialog.component.js
```diff
@@ -34,12 +36,20 @@ const Dialog = ({
...
   const dialogRef = useRef();
   const innerContentRef = useRef();
   const titleRef = useRef();
   const listenersAdded = useRef(false);
+  const { p, py, px } = contentPadding;
+
+  const paddingTopInnerContent = [py, p].some(
+    (padding) => padding !== undefined
+  )
+    ? `${ZERO_PADDING}px`
```
grabkowski, line 51: I think that should be just 0 (without px), and we don't need to set it to a const as it's pretty self explanatory

src/components/dialog/dialog.component.js
```diff
@@ -34,12 +36,20 @@ const Dialog = ({
...
+  contentPadding = {},
   ...rest
 }) => {
   const dialogRef = useRef();
   const innerContentRef = useRef();
   const titleRef = useRef();
   const listenersAdded = useRef(false);
+  const { p, py, px } = contentPadding;
+
+  const paddingTopInnerContent = [py, p].some(
```
grabkowski, line 48: could that be moved into the styled component?

src/components/dialog/dialog.component.js
```diff
@@ -273,6 +289,12 @@ Dialog.propTypes = {
   bespokeFocusTrap: PropTypes.func,
   /** The ARIA role to be applied to the Dialog container */
   role: PropTypes.string,
+  /** Padding to be set on the Dialog content */
+  contentPadding: PropTypes.shape({
+    p: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6, 7, 8]),
```
samtjo, line 294: Did it come from UX to lock these props down to 8px multipliers only? As a user I would expect these to work in the same way as all our other padding props, where we can pass in any valid CSS string

src/components/dialog/dialog.config.js
```diff
@@ -7,5 +7,7 @@ export const DIALOG_SIZES = [
   "large",
   "extra-large",
 ];
-
 export const TOP_MARGIN = 20;
+export const CONTENT_TOP_PADDING = 24;
```
samtjo, line 11: This was previously 20px if you look at `DialogInnerContentStyle`

edleeks87: It's a requirement of this ticket to update that value to 24

src/components/dialog/dialog.component.js
edleeks87: Yeah I spoke with @harpalsingh about this and we agreed for now this basic interface is all that we'll need, if consumers request a more flexible one we can open it up at that point. The issue is the need to calculate the values for the form margins, if we allow all the variations (strings, px/ %/ em and so on) it becomes quite complex and messy to pull the values out.

src/components/dialog/dialog.config.js
samtjo: Would be good to put all the intended changes in the PR description 👍🏻

src/components/dialog/dialog.style.js
```diff
@@ -124,12 +164,13 @@ const DialogContentStyle = styled.div`
...
   flex: 1;
   padding: 0px ${HORIZONTAL_PADDING}px ${CONTENT_BOTTOM_PADDING}px;
+  ${paddingFn}
 `;
 
 const DialogInnerContentStyle = styled.div`
-  padding-top: 20px;
   position: relative;
   flex: 1;
+  ${paddingFn}
```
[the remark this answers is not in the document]
edleeks87: Yeah missed that one 🤦

grabkowski (dismissed):

src/components/dialog/dialog.component.js
samtjo: Ok makes sense 👍🏻

samtjo (approved):

divyajindel (approved):

grabkowski (dismissed):

samtjo (dismissed):

grabkowski (approved):

divyajindel (approved):

PR #4765, opened 2022-01-28
docs(unit-tests): rfc for converting tests to rtl

samtjo:
### Proposed behaviour

This PR contains an RFC detailing why and how I think we should convert our unit tests from using Enzyme to React Testing Library.

### Current behaviour

Our unit tests currently use Enzyme as a testing utility for React components. This does not support React 17+ and it doesn't look like it will anytime soon.

---

Please use 👍  if you approve.

### DIFF
--- a/rfcs/text/enzyme-to-rtl.md
+++ b/rfcs/text/enzyme-to-rtl.md
@@ -1,0 +1,168 @@
+- Start Date: 28/01/2022
+
+# Table of contents
+
+- [Summary](#summary)
+- [Basic example](#basic-example)
+- [Motivation](#motivation)
+- [Detailed design](#detailed-design)
+  - [Config setup](#config-setup)
+  - [Component testing](#component-testing)
+  - [User events](#user-events)
+- [Drawbacks](#drawbacks)
+  - [Conversion effort](#conversion-effort)
+  - [Developer training](#developer-training)
+  - [Supporting both utilities concurrently](#supporting-both-utilities-concurrently)
+- [Alternatives](#alternatives)
+  - [Sticking with Enzyme](#sticking-with-enzyme)
+  - [Other React testing utilities](#other-react-testing-utilities)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+
+# Summary
+
+In Carbon we currently use [Jest](https://jestjs.io/) as our unit testing framework, and [Enzyme](https://enzymejs.github.io/enzyme/) as a testing utility for our React components.
+
+This works well and Enzyme has been a very popular utility for testing React applications for a long time. Recently however, support for it seems to have stalled, particularly for the most recent version of React (at the time of writing), React 17.
+
+An [issue was raised](https://github.com/enzymejs/enzyme/issues/2429) in August 2020 to add support for React 17 to Enzyme, but as of January 2022 this has still not been added and it doesn't look like it will be for the foreseeable future.
+
+This is preventing us from upgrading Carbon to use React 17 and supporting it for Carbon's consumers. React 17 was released in October 2020, and [React 18](https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html) is now in beta so we are at risk of falling behind, which is not acceptable for a component library supporting so many other projects. We have already had a number of issues raised about the lack of React 17 support in our components.
+
+The logical solution for us is to move towards the officially recommended tesing utility for React, [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/). This has much better support, and is actually already used by most of the modern Sage projects we support.
+
+# Basic example
+
+The React Testing Library docs have a great [migration from Enzyme](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme) guide. This has a section on [basic Enzyme to React Testing Library migration examples](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme#basic-enzyme-to-react-testing-library-migration-examples) which show exactly how the two libraries compare. I will add one of the basic examples here, but for a more in depth look, read the above docs.
+
+A simple Enzyme test for one of our components might look something like this:
+
+```jsx
+describe("Example Component", () => {
+  it("renders the title in a h1 element", () => {
+    const wrapper = shallow(<ExampleComponent title="Enzyme test" />);
+    expect(wrapper.find("h1").text()).toEqual("Enzyme test");
+  });
+});
+```
+
+The same test converted to React Testing Library will look like this:
+
+```jsx
+describe("Example Component", () => {
+  test("renders the title in a h1 element", () => {
+    render(<ExampleComponent title="Enzyme test" />);
+    expect(screen.getByRole("heading", { level: 1 })).toHaveTextContent(
+      "Enzyme test"
+    );
+  });
+});
+```
+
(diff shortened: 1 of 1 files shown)


rfcs/text/enzyme-to-rtl.md
```diff
@@ -0,0 +1,162 @@
...
+    );
+  });
+});
+```
+
+The two libraries have slightly different APIs, but the tests are fairly similar and the effort required to convert them should be minimal.
+
+# Motivation
+
+The main motivation for us making this conversion is that, as previously mentioned, Enzyme does not yet support React 17. It is well over a year after it was released, and it does not look like support will be added in the near future. React 18 is in beta and we we be blocked from upgrading to that as well when it is released. It could be years before Enzyme add support for that, if ever.
```
Parsium, line 66: Looks great @samtjo! 🚀  This looks very comprehensive and the migration docs look useful. Just found this one spelling mistake - sorry to be that guy! Should be `we will be blocked`.

Parsium (changes requested):

Parsium (dismissed):

DlgSHi:
What about considering using `cypress` for a unit testing?

rfcs/text/enzyme-to-rtl.md
```diff
@@ -0,0 +1,162 @@
...
+No modifications would be needed to our Jest config either, until we finish the conversion and remove anything related to Enzyme.
+
+We currently wrap components in theme or i18n providers on a test-by-test basis, but if we decide we want to wrap all components in these by default, we can create a custom render method as desribed [here](https://testing-library.com/docs/react-testing-library/setup/#custom-render) in the React Testing Library docs.
+
+If we were to do so, we would likely want to wrap our components in the `CarbonProvider` component (to provide the theme and the Design Token CSS variables) and the `I18nProvider` component for translations, like this:
+
+```jsx
+const CustomProviders = ({ children }) => {
+  return (
+    <CarbonProvider theme="mint">
```
edleeks87, line 97: I know this is just a demo of a potential implementation but I think it would be good to have `theme` and `locale` as props of `CustomProvider`. `mint` and `enGB` are already the default values for the `CarbonProvider` and `I18nProvider` but it might be more explicit also have default values of `"mint"` and `enGB` if you do choose to add the props to the top level destructure

edleeks87 (commented):
Looks good to me @samtjo, one small suggestion from me but happy to approve if you don't think it's worthwhile. It might also be worth adding something around the question @DlgSHi has asked as well

rfcs/text/enzyme-to-rtl.md
samtjo: Thanks @edleeks87 - it's a good idea in theory but unfortunately using the `wrapper` option with the RTL `render` function, it only passes `children` as a prop, you can't pass any other props in. You can see that [here](https://github.com/testing-library/react-testing-library/blob/8f17a2bc4fa9c2c17a623b38ded9ab18b086e457/src/pure.js#L56).

There's really no need to change the theme in the unit tests anyway, they should be for testing the functionality of the code. Chromatic will take care of testing the differences between themes for us. 

Translations wise, if we need to use a different locale for a certain test, we can simply just not use this custom render function, just use the normal `render` from RTL and an `I18nProvider` with whatever locale we need. 

Good point about the defaults but I just added them specifically in for the sake of the example here, in reality we don't have to have theme if we do just want to go with the default values 👍🏻

edleeks87: Ah fair point I hadn't realised RTL's render constrained it like that, I'm happy to leave as is then 👍

mkrds (dismissed):

DipperTheDan (approved):

Parsium:
> What about considering using `cypress` for a unit testing?

I was wondering this as they provide similar functionality - however after looking online RTL is quoted as being generally faster on a unit/integration testing level - as it doesn't do as the same amount of heavy-lifting as Cypress does.

DlgSHi:
> 

yes, but RTL is rendering tests NOT in the really browser but `cypress` does it where we can see what exactly is performing by User

samtjo:
Section added in Alternatives about Cypress unit testing @DlgSHi

Parsium (approved):

rfcs/text/enzyme-to-rtl.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Cypress Unit Testing
+
+The tool we use for e2e testing, Cypress, has utilities for [unit testing](https://go.cypress.io/get-started?item=unit%20tests%20made%20easy&utm_adgroup=130177744717&utm_keyword=unit%20testing&utm_source=google&utm_medium=cpc&utm_campaign=15376239586&utm_term=unit%20testing&hsa_acc=8898574980&hsa_cam=15376239586&hsa_grp=130177744717&hsa_ad=564551372576&hsa_src=g&hsa_tgt=kwd-304023822183&hsa_kw=unit%20testing&hsa_mt=p&hsa_net=adwords&hsa_ver=3&gclid=Cj0KCQiAmKiQBhClARIsAKtSj-lZ4Hcgjnxu0eAYlCevyDw0UbXlJJFd6PdYAFB_8yfkU38y0NWoWt0aAmvOEALw_wcB). This could be a reasonable alternative to using React Testing Library, however it runs tests in an actual browser, rather than using jsdom like Jest does. This is potentially overkill for unit tests, and Cypress is typically used for e2e testing. Cypress is not widely adopted in the React community for unit testing (it may get more popular) and given that we already use it for higher level testing of our components in the browser, it makes more sense to use the widely adopted React Testing Library for unit testing and have the extra layer of confidence in our automated testing by using both.
```
chrisbarber86, line 146: Probably also worth noting that the other significant reason not to use Cypress is that it's a lot slower - it currently takes requires us to run the tests across 20 parallel pipelines just to keep the build under 10 minutes. Adding more tests to this would require us to scale even more horizontally to maintain performance, which has several implications not least increased GH Actions usage.

chrisbarber86 (dismissed):

rfcs/text/enzyme-to-rtl.md
samtjo: Thanks @chrisbarber86 - I've added that to the Cypress Unit Testing section 👍🏻

DlgSHi: I actually do not agree with that. After converting tests to `cypress-react` we won't need to start storybook. So it will decrease the amount of used time.

edleeks87 (approved):

rfcs/text/enzyme-to-rtl.md
StylianosKoidakisSage: I do not see a reason of cypress and for an end to end test on a component. The need on a component of a UI library should be to test its UI on all the different states. e2e frameworks are mainly testing happy paths and critical flows. Here i would say we need to cover all the corner cases to ensure the consumer will never have issues with the components.

StylianosKoidakisSage:
Do we have any document with out testing pyramid/strategy for carbon? 

I would say RTL is a need for this project. Cypress should not be a need as it gives value for e2e test but in our case they are some small cases that can be used as we have components are doing too many things. 

From my point of view we should have:
- unit test(jest): for complex units
- RTL: for integration test of component state
- visual regression test

Thank you :)

StylianosKoidakisSage (commented):
As some components do too many things would be nice to define a way to write test to check the responsiveness

PR #4818, opened 2022-02-10
test(button): refactor button component

ZhuoyuJin:
### Proposed behaviour

Refactor for Button component to use new cypress-component-react framework for testing.

### Current behaviour

### Checklist

#### QA

### Testing instructions

 - Check if the button.test.js file passed


src/components/button/button.test.js
```diff
@@ -0,0 +1,184 @@
...
+      }
+    );
+  });
+
+  describe("Check events for Button component", () => {
+    it("should call click function in Actions Tab", () => {
+      const callback = cy.stub();
+
+      CypressMountWithProviders(<Button onClick={callback} />);
+      // then run our tests
```
DlgSHi, line 154: delete this please

src/components/button/button.test.js
```diff
@@ -0,0 +1,184 @@
...
+            .children()
+            .children()
+            .should("not.have.value", cssValue, "wrap");
+        }
+      }
+    );
+  });
+
+  describe("Check events for Button component", () => {
+    it("should call click function in Actions Tab", () => {
```
DlgSHi, line 150:
```suggestion
    it("should call onClick event", () => {
```

src/components/button/button.test.js
```diff
@@ -0,0 +1,184 @@
...
+      const callback = cy.stub();
+
+      CypressMountWithProviders(<Button onClick={callback} />);
+      // then run our tests
+      buttonDataComponent().click({ force: true });
```
DlgSHi, line 155: this is only the action  and no assertion is made for `onClick`

src/components/button/button.test.js
```diff
@@ -0,0 +1,184 @@
...
+      buttonDataComponent()
+        .focus()
+        .blur({ force: true })
+        .then(() => {
+          // eslint-disable-next-line no-unused-expressions
+          expect(callback).to.have.been.calledOnce;
+        });
+    });
+
+    it("should be triggered on keyDown", () => {
```
DlgSHi, line 172:
```suggestion
    it("should call the onKeyDown", () => {
```

DlgSHi (changes requested):
1. You have deleted some step that were used in other `feature` files. Please check with steps and move them into `common-steps.js` file.
2. You need to remove `buttonAsSibling.feature` file as well because you covered those tests here.
3. Update the description for the PR :) it is not a `Detail`.

src/components/button/button.test.js
```diff
@@ -0,0 +1,184 @@
...
+    );
+
+    it.each([
+      ["after", "10px"],
+      ["before", "10px"],
+    ])(
+      "should set postion to %s for button as a button",
+      (iconPosition, margin) => {
+        CypressMountWithProviders(
+          <Button iconType="add" iconPosition={iconPosition} />
```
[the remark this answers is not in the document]
ZhuoyuJin: Why do we layout like this here?

src/components/button/button.test.js
```diff
@@ -0,0 +1,192 @@
...
+
+    it.each([
+      ["after", "left"],
+      ["before", "right"],
+    ])(
+      "should set position to %s for icon in a button",
+      (iconPosition, margin) => {
+        CypressMountWithProviders(
+          <Button iconType="add" iconPosition={iconPosition}>
+            {" "}
```
DlgSHi, line 78: why do we need this here ? and the line below ?

ZhuoyuJin: Added by mistake, will delete it.

DlgSHi (approved):

cypress/support/step-definitions/common-steps.js
```diff
@@ -22,6 +22,10 @@ import {
 import { dialogTitle } from "../../locators/dialog";
 import DEBUG_FLAG from "..";
 import { pagerSummary } from "../../locators/pager";
+import {
```
sianford, line 25: Why were Button-specific steps added to this common file? Were they required for other components that had Button embedded?

src/components/button/button.test.js
```diff
@@ -0,0 +1,191 @@
...
+        );
+
+        icon().should("have.css", `margin-${margin}`, "8px");
+      }
+    );
+
+    it("should render Button with full width", () => {
+      CypressMountWithProviders(<Button fullWidth />);
+
+      buttonDataComponent().should("have.css", "width").and("eq", "1350px");
```
sianford, line 89: Why does the width need to be specified in number of pixels for a full width button? Shouldn't it just be a percentage? I'd have thought that the number of pixels width will differ depending on viewport size.

src/components/button/button.test.js
```diff
@@ -0,0 +1,191 @@
...
+
+      buttonDataComponent()
+        .trigger("keydown", { force: true }, keyCode("rightarrow"))
+        .then(() => {
+          // eslint-disable-next-line no-unused-expressions
+          expect(callback).to.have.been.calledOnce;
+        });
+    });
+  });
+});
```
sianford, line 191: Why are the onChange and onFocus actions not tested here if onClick, onBlur and onKeyDown are?

cypress/support/step-definitions/common-steps.js
ZhuoyuJin: They are testing for the enable and disable to different types of button, I don't think it's needed for other components.

src/components/button/button.test.js
```diff
@@ -0,0 +1,191 @@
...
+      (subtext) => {
+        CypressMountWithProviders(<Button size="large" subtext={subtext} />);
+
+        buttonSubtextPreview().should("have.text", subtext);
+      }
+    );
+
+    it.each(testData)(
+      "should render tooltip message with %s special characters",
+      (tooltipmessage) => {
```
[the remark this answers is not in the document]
ZhuoyuJin: I think tooltipmessage in this case is a passing in parameter, so I assume if we keep the same name in this function should be ok. But you are right, to make it look consistent, we should change it to 'tooltipMessage'.

src/components/button/button.test.js
ZhuoyuJin: We could do a test for onFocus, but not onChange. onChange is something different with more complicated scenarios. It is hard to test onChange event in button, we discussed it and I decided to leave it for now.

cypress/support/step-definitions/common-steps.js
DlgSHi: I think those steps are also used in `splitButton` or `multiActionButton`

src/components/button/button.test.js
DlgSHi: I think it should be similar to `onClick` test :)

ZhuoyuJin: onChange test will fail if I apply the same tests

sianford: After discussion we've found that onChange does not apply to Button so it's correct that you did not include it.

sianford (approved):

Parsium (approved):

PR #4861, opened 2022-02-18
fix(date): ensure day value is validated against format with the month value included FE-4963

### LINKED ISSUE #4853: Date input component rawValue is undefined when is greater than 28
### Current behaviour

When we choose a date that is greater than 28 raw value is null

### Expected behaviour

The raw value should be the correct one

https://codesandbox.io/s/floral-sun-5y3tq2?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

I suppose the limit of a month is connected with the current month that is 28.  We should try to find a way to define which is the limit of each month.

### Carbon Version

104.18.1

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4853

### Proposed behaviour

Ensure that any update to the day value also includes the current month value that is in the input
to prevent it using the month from today's date. Without this change valid day values will be
rejected when they are outside of today's month's valid days, eg. 29 when the today's month is
February.

### Current behaviour

Day values over 28 is invalid as it uses the current month (February) when comparing if the value is valid

### Checklist

#### QA

### Testing instructions


src/components/date/date.spec.js
```diff
@@ -473,6 +473,7 @@ describe("Date", () => {
       "12/42/3213",
       "foo",
       "1/01.11",
+      "29/02/2019",
```
samtjo, line 476: I think worth adding in a `31/04/2019` for a month with 30 days as well

edleeks87: Ah yeah great shout

edleeks87: also a great shout 👍

samtjo (approved):

Parsium (approved):

divyajindel (approved):

Parsium (approved):

Issue #4822, opened 2022-02-11
Pager does not respect width of FlatTable


hagad:
### Current behaviour

When adding a width to `FlatTable` - the width of the Pager can overflow which causes scrolling issues.

### Expected behaviour

The pager should be 100% of the width of the FlatTable and not able to be scrolled.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/bitter-microservice-28l1w?file=/src/index.js

### Carbon Version

103.2.1

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

Issue #4821, opened 2022-02-11
FlatTable with colspan : missing props


hagad:
### Current behaviour

The props :  "width"  "overflowX" "overflowY" don't exist in flat-table.d.ts

### Expected behaviour

Looking at https://carbon.sage.com/?path=/docs/flat-table--with-colspan   :
The props :  "width"  "overflowX" "overflowY" should exist in  flat-table.d.ts

### CodeSandbox or Storybook URL

https://carbon.sage.com/?path=/docs/flat-table--with-colspan 

### Carbon Version

104.12.1 

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

nicktitchmarsh:
FE-4950

samtjo:
Hi @hagad - thanks for raising, this will be fixed in https://github.com/Sage/carbon/pull/4868.

Just note we will add `width` and `overflowX`, but not `overflowY` - this is not a prop we want to expose to users as we currently set that based on the values of other props.

PR #4868, opened 2022-02-18
feat(flat-table): add width prop

samtjo:
Fixes: #4822
Fixes: #4821

Note we will not be adding the `overflowY` prop for issue #4821 

### Proposed behaviour

Add a `width` prop to `FlatTable`.

Add a wrapper around the table element in `FlatTable` to handle the horizontal scrolling so it does not horizontally scroll the table footer. This means only the table content (including headings) can be scrolled horizontally, and not a `Pager` when it is in the footer. The `Pager` should be 100% of the width of the table. It is down to the user to use the responsive props on `Pager` to ensure that it fits within the table width.

The `width` and `overflowX` props cannot just be added to the `StyledFlatTable` as they don't work as expected on a `table` element.

### Current behaviour

When using the `width` and `overflowX` props on `FlatTable` with a `Pager` in the footer, scrolling horizontally incorrectly scrolls the `Pager` as well as the table content.

### Checklist

#### QA

### Testing instructions

Use the following sandbox: 
(https://codesandbox.io/s/condescending-sky-i5pnku?file=/src/index.js)

You can modify the width of the table using the textbox, and it shows an example of how the Pager content should be modified by the user at different widths. We do not support widths of less than 300px.

The Pager can no longer be horizontally scrolled with this fix but the table content can be.

**DO NOT USE THE SANDBOX FROM THE ISSUE.** This does not have the responsive props on Pager set correctly.

### DIFF
--- a/cypress/locators/flat-table/index.js
+++ b/cypress/locators/flat-table/index.js
@@ -5,10 +5,12 @@ import {
   FLAT_TABLE_PAGE_SIZE_SELECT,
   FLAT_TABLE_PAGE_SELECT_LIST,
   PAGE_SELECT_INPUT,
+  FLAT_TABLE_WRAPPER,
 } from "./locators";
 
 // component preview locators
 export const flatTable = () => cy.get(FLAT_TABLE_COMPONENT);
+export const flatTableWrapper = () => cy.get(FLAT_TABLE_WRAPPER);
 export const flatTableHeader = () => flatTable().find("thead tr");
 export const flatTableHeaderCells = () => flatTableHeader().find("th");
 export const flatTableBodyRowByPosition = (index) =>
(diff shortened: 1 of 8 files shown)


Parsium (approved):

edleeks87 (commented):
All the changes you've made look good to me @samtjo  but I've noticed an issue when using the `FlatTableRowHeader` whereby the `Pager` scrolls with the table. I think we can prevent this by setting `position: sticky` and `left: 0;` on the Pager container but you may find a better solution

https://codesandbox.io/s/mutable-cloud-o23vgf?file=/src/index.js

edleeks87 (dismissed):

edleeks87 (approved):

Parsium (approved):

PR #4890, opened 2022-02-24
test(badge): cypress refactor

stephenogorman:
### Proposed behaviour

Refactor for Badge component to use new cypress-component-react framework for testing.

### Current behaviour

### Checklist

#### QA

### Testing instructions

 - Check if the badge.test.js file passed


src/components/badge/badge.test.js
```diff
@@ -0,0 +1,83 @@
...
+    it.each([[99]])(
+      "Badge should display cross icon when hovered over",
+      (value) => {
+        CypressMountWithProviders(<BadgeComponent counter={value} />);
+
+        badge().realHover();
+        badge()
+          .should("have.css", "background")
+          .then(($el) => {
+            expect($el).contains("rgb(0, 126, 69)");
```
DlgSHi, line 67: we also need to check if shows us the `cross-icon`

src/components/badge/badge.test.js
```diff
@@ -0,0 +1,83 @@
...
+    );
+
+    it.each([[99]])("Badge should call click action when clicked", (value) => {
+      CypressMountWithProviders(<BadgeComponent counter={value} />);
+
+      badge().click();
+      badge()
+        .should("have.css", "background")
+        .then(($el) => {
+          expect($el).contains("rgb(0, 126, 69)");
```
DlgSHi, line 79: this is not checking the `event`

DlgSHi: we need to use the similar code from other TC with using `onClick`

DlgSHi (commented):
also we need to cover the `children` prop

src/components/badge/badge.test.js
stephenogorman: Test now checks the cross-icon.

src/components/badge/badge.test.js
stephenogorman: Test now checks the event.

DlgSHi (dismissed):

src/components/badge/badge.test.js
```diff
@@ -0,0 +1,87 @@
...
+        buttonType="tertiary"
+      >
+        Filter
+      </Button>
+    </Badge>
+  );
+};
+
+context("Testing Badge component", () => {
+  describe("Should render Badge component", () => {
```
divyajindel, line 27: describe("should render Badge component", () => {

stephenogorman: Changed 'Should' to 'should'

src/components/badge/badge.test.js
```diff
@@ -0,0 +1,87 @@
...
+        CypressMountWithProviders(<BadgeComponent counter={value} />);
+
+        badgeCounter()
+          .invoke("show")
+          .should("be.visible")
+          .invoke("text")
+          .and("contain", 99);
+      }
+    );
+
```
DianeGit22, line 50: Question on this please Stephen. The purpose of this test is to mount the badge with the counter value of 999 and to verify the counter is actually set to 99 as that's the  max value allowed. If there was a bug and the counter was actually set to 999, this test would still pass, wouldn't it? Because 999 contains 99. Could you mount it with a value of 980 so the verification would fail when it should? Or if there is a reason to use 999, could we use "equals" instead of "contain" for the verification? Thanks.

DianeGit22 (changes requested):

src/components/badge/badge.test.js
stephenogorman: Okay, I've made that change.

DianeGit22: Thanks Stephen

DlgSHi: or it could be also changed to `equals` and the TC using `999` could stay unchanged.

DlgSHi (approved):

DlgSHi:
squash the commits please

DlgSHi (approved):

ZhuoyuJin (approved):

DianeGit22 (approved):

DlgSHi (approved):

edleeks87 (approved):

PR #4908, opened 2022-02-28
feat(base-theme): swap space values for values described by tokens - FE-4973

JoannaMikul:
### Proposed behaviour
Swaps space values in array for values described by tokens. 
Updates tests related to changes.

### Current behaviour
Space array in `base-theme.config.js` is described by numeric values.

### Checklist

### Testing instructions
Components that have props : margin/spacing/padding, should get the correct values.


src/components/button/button.spec.js
```diff
@@ -737,24 +737,4 @@ describe("Button", () => {
       });
     });
   });
-
-  describe("overriding size based padding", () => {
```
mkrds, line 741: Why is this test removed?

src/components/dialog/dialog.spec.js
```diff
@@ -585,7 +588,9 @@ describe("Dialog", () => {
         (["top", "bottom"].includes(position) && !["p", "py"].includes(prop)) ||
         (["left", "right"].includes(position) && !["p", "px"].includes(prop))
       ) {
-        return `${isMargin ? "-" : ""}${defaultPaddingValues[position]}px`;
+        return isMargin
+          ? `calc(-1 * ${defaultPaddingValues[position]})`
```
mkrds, line 592: Why not use the `setNegativeValue` here as well?

src/__spec_helper__/test-utils.js
```diff
@@ -192,8 +194,16 @@ const backgroundProps = [
   ["backgroundRepeat", "background-repeat", "no-repeat"],
 ];
 
-const getDefaultValue = (value) => {
-  if (typeof value === "number") {
+export const getDefaultValue = (value) => {
+  const spaceArrayLength = space.length - 1;
+  const parsedValue = +value;
+  if (typeof value === "string" && value > spaceArrayLength) {
```
samtjo, line 200: This only makes sense if it is a string containing a number, if it's something like "auto" this doesn't work.

samtjo: This function should be something like: 

```js
const getDefaultValue = (value) => {
  if (typeof value === "number" && value < space.length) {
    return space[value];
  }
  return value;
};
```

We don't number numerical values above the length of the space array, and any string should just be set as a string

src/style/themes/base/base-theme.config.js
```diff
@@ -430,24 +444,9 @@ export default (palette) => {
...
-        spacing400: `${this.space[4]}px`, // 32px
-        spacing500: `${this.space[5]}px`, // 40px
-        spacing600: `${this.space[6]}px`, // 48px
-        spacing700: `${this.space[7]}px`, // 56px
-        spacing800: `${this.space[8]}px`, // 64px
-        spacing900: `${this.space[9]}px`, // 72px
-
         fontSizes100: this.text.size,
+
+        spacing1000: "80px",
```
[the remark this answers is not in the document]
JoannaMikul: We don't currently have that token. It is in the process of being created and will be removed. This is a temporary solution to present the main concept.

src/__spec_helper__/test-utils.js
mkrds: @samtjo I think it actually does work correctly, because if value is `auto` then it will return false

JoannaMikul: @samtjo The style-system works in such a way that if I give for example m={1} it will take element 2 from the array. But if you put m=“1” it will also take the 2nd element from the space array.
At least that’s how it behaves when I checked in browser.

src/components/button/__snapshots__/button.spec.js.snap
```diff
@@ -102,11 +102,9 @@ exports[`Button when only the "iconPosition" and "iconType" props are passed int
 
 .c6 .c1 {
   margin-left: 0px;
-  margin-right: 0px;
-  margin-bottom: 1px;
+  margin-right: var(--spacing100);
```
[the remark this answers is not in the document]
JoannaMikul: swapped .c6.c1 and .c7.c1 on the screenshots

src/__spec_helper__/test-utils.js
samtjo: Yeah you're right, I didn't realise styled-system worked with numerical string like that, in that case I think it's fine 🙂

samtjo (dismissed):

mkrds (approved):

samtjo (approved):

mkrds (dismissed):

DlgSHi (approved):

mkrds (approved):

samtjo (approved):

PR #4946, opened 2022-03-07
fix(select): ensure textbox placeholder can be set when passed as prop

### LINKED ISSUE #4894: The placeholder property of the Select component cannot be updated
### Current behaviour

We can no longer update the placeholder property of the Select component.

Example: https://codesandbox.io/s/select-placeholder-0vus7w?file=/src/index.js

### Expected behaviour

To be able to modify the placeholder property of the Select component.

https://codesandbox.io/s/select-placeholder-0vus7w

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

104.26.1

Chrome

Windows

_No response_

### Confidentiality

Parsium:
fixes #4894 

### Proposed behaviour

- Remove placeholder logic in `SelectText` - let `SelectTextbox` handle this instead and just use the text that is  passed from it.
- Adjust conditionals in `SelectTextbox` that determines which placeholder text to use.
- Add and amend unit tests

https://user-images.githubusercontent.com/18368713/157647151-dd5f8ca3-2edc-4d4a-82a2-bf9976203bac.mov

### Current behaviour

- Passing a custom placeholder as a prop of `Select` does not override the default locale placeholder.

### Checklist

#### QA

### Additional context

Added and amended unit tests have been written in a RTL style - referring to DOM elements only and not components - given future plans to migrate to RTL.

### Testing instructions

The following CodeSandbox is an example of the broken behaviour:
https://codesandbox.io/s/select-placeholder-0vus7w

And the following CodeSandbox shows the fixed behaviour:
https://codesandbox.io/s/select-placeholder-kub9e8


jamime (dismissed):

src/components/select/__internal__/select-text/select-text.component.js
```diff
@@ -57,7 +52,7 @@ const SelectText = ({
       tabIndex="-1"
       transparent={transparent}
     >
-      {formattedValue || placeholderText}
+      {formattedValue || (hasPlaceholder ? placeholder : "")}
```
edleeks87, line 55: Could we not add a default like so `formattedValue = ""` and then change this to be like the below, that way we only check its value the one time when setting `hasPlaceholder`?

```
hasPlaceholder ? placeholder : formattedValue
```

Parsium: Yeah this looks a lot cleaner 👍🏼

edleeks87 (approved):

divyajindel (approved):

PR #4970, opened 2022-03-16
test(loader): refactor for component

DlgSHi:
### Proposed behaviour

Refactor for <Loader> component to use new cypress-component-react framework for testing.

### Current behaviour

### Checklist

#### QA

### Testing instructions

 - Check if the loader.test.js file passed


src/components/loader/loader.test.js
```diff
@@ -0,0 +1,105 @@
...
+        "have.css",
+        "background-color",
+        color
+      );
+    });
+
+    it("should render Loader inside the Button component and be able to focus it", () => {
+      CypressMountWithProviders(<LoaderInsideButton />);
+
+      buttonDataComponent()
```
ZhuoyuJin, line 100: Shouldn't it be loaderInsideButton()?

DlgSHi: yes, you are right, fixed

ZhuoyuJin (dismissed):

grabkowski (dismissed):

src/components/loader/loader.test.js
```diff
@@ -0,0 +1,103 @@
...
+      [LOADER_SIZES[2], 128],
+    ])("should render Loader using %s as size in Button", (size, width) => {
+      CypressMountWithProviders(<LoaderInsideButton size={size} />);
+
+      loaderInsideButton()
+        .should("have.css", "height", "40px")
+        .and("have.css", "width", `${width}px`);
+    });
+
+    it("should render Loader inside the Button component", () => {
```
sianford, line 57: What is the purpose of this test case when there are already three preceding this one to test the Loader inside a button? The only difference I can see is that the background colour is checked. Could this assertion be added to the previous Loader in button test cases and remove this one?

DlgSHi: as per testing theory we suppose to check one assertion in one TS (at least the same parameters) It the test above we are checking the `size` but here only the colour :)

sianford: Ok, could you amend the test to specify "should render Loader inside the Button component with correct background colour" then please so it's clear what its purpose is as when you see it running it looks like it's a duplicate test case from the test name.

DlgSHi: fixed

sianford (approved):

grabkowski (approved):

PR #4974, opened 2022-03-18
refactor(app-wrapper): pass props and propTypes rather than instance to validProps FE-4353

edleeks87:
### Proposed behaviour

Converts files in `utils` directory to `TypeScript`
Removes any utils that are no longer used/ needed
Replaces `WithUniqueIdProps` HOC with a new `useUniqueId` hook and replaced usage in `Textbox`

### Current behaviour

Files in `utils` directory are JS, HOC wraps `Textbox`, unused utils still exist

### Checklist

#### QA

### Testing instructions

Not really much to test here but needs QA approval as spec files have been updated


samtjo (commented):
LGTM, I think we can just replace usages of the ether util and delete it and then happy to ✅

src/__internal__/utils/ether/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Optionally includes the safeProps which can be defined as a static property on a Class or passed as the second
+ * argument.
+ *
+ */
+function validProps(
+  component: ComponentInterface,
+  safeProps?: string[]
+): Record<string, unknown> {
+  const unsafeProps: string[] = difference(
+    Object.keys(component.propTypes),
```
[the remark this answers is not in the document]
edleeks87: Yeah sounds like the right approach cheers mate

samtjo (approved):

src/__internal__/utils/helpers/events/composedPath.ts
```diff
@@ -1,8 +1,9 @@
 // A basic cross-browser implementation of Event.protoType.composedPath
 // JSDOM does not suppor event bubbling
 // IE does not support composedPath
+
 // https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath
-export default (ev) => {
+export default (ev: CustomEvent): EventTarget[] | undefined[] => {
```
grabkowski, line 6: this `undefined[]` looks a bit weird, shouldn't there be just `EventTarget[]`?

src/__internal__/utils/helpers/events/events.ts
```diff
@@ -359,9 +361,8 @@ const Events = {
    * @param {Event} ev A Javascript event
    * @returns {EventTarget[]} objects representing the objects on which an event listener will be invoked
    */
-  composedPath: (ev) => {
+  composedPath: (ev: CustomEvent): EventTarget[] | undefined[] => {
```
grabkowski, line 364: is there a case that an array of undefined will be returned here?

src/__internal__/utils/helpers/tags/tags.ts
```diff
@@ -1,12 +1,21 @@
+interface Props {
```
grabkowski, line 1: I think that name could be more meaningful, e.g. `DataProps`

src/__internal__/utils/helpers/tags/tags.ts
```diff
@@ -1,12 +1,21 @@
...
+  "data-element"?: string;
+  "data-role"?: string;
+}
+
+interface TagProps extends Props {
+  "data-component": string;
+}
```
grabkowski, line 8: shouldn't these interfaces be exported?

src/__internal__/utils/helpers/tags/tags.ts
```diff
@@ -1,12 +1,21 @@
...
+
 /**
  * Builds props object containing top level data tags
  *
  * @param {String} name of component
  * @param {Object} component props
- * @return {Object} dataTagProps
+ * @return {Object} dataProps
  */
```
grabkowski, line 16: I think that could be removed, as we already have the types known

src/__internal__/utils/helpers/tags/tags.ts
```diff
@@ -1,12 +1,21 @@
...
-function tagComponent(component, props) {
-  const tagProps = {
+function tagComponent(component: string, props: Props): TagProps {
```
grabkowski, line 17:
```suggestion
function tagComponent(componentName: string, dataProps: DataProps): TagProps {
```

src/__internal__/utils/helpers/tags/tags.ts
```diff
@@ -1,12 +1,21 @@
...
+  const tagProps: TagProps = {
```
grabkowski, line 18:
```suggestion
  const tagProps: TagProps = {
    "data-component": component,
    ...dataProps,
  }
```

edleeks87: If I'm understanding the suggestion correctly, I don't think we should make this change as it will spread the rest on to any component that uses it rather than filtering and spreading just the tag props

```
function tagComponent(componentName: string, props: DataProps): TagProps {
  return {
    "data-component": componentName,
    ...props
  };
 ```

File: src/__internal__/utils/helpers/tags/tags.ts (hunk 4 above)
Comment (edleeks87) [reply src/__internal__/utils/helpers/tags/tags.ts:1]: Fair point, I realised I'll need to extend the interface to support other unknown props as the function takes the `rest` props as the second argument

File: src/__internal__/utils/helpers/tags/tags.ts (hunk 7 above)
Comment (edleeks87) [reply src/__internal__/utils/helpers/tags/tags.ts:17]: I don't think we should rename the second arg to `dataProps` as it will contain a mix of all props passed as `rest` to a component

**Review by samtjo — APPROVED** [review]

File: src/__internal__/utils/helpers/tags/tags.ts (hunk 8 above)
Comment (grabkowski) [reply src/__internal__/utils/helpers/tags/tags.ts:14]: if indeed all properties are passed in this object, then I agree that this code does not need to be changed.
At the time of writing the comment above, the object interface looked like this:
```
interface Props {
  "data-element"?: string;
  "data-role"?: string;
}
```
so then it would made perfect sense to do that change.

**Review by grabkowski — APPROVED** [review]

**Review by grabkowski — APPROVED** [review]

**Review by samtjo — APPROVED** [review]

File: src/__internal__/utils/helpers/events/events.ts [hunk 9]
```diff
@@ -0,0 +1,256 @@
...
+   * A method to determine if an event is of a particular type
+   * */
+  isEventType: (ev: Event, type: string): boolean => {
+    return ev.type === type;
+  },
+
+  /**
+   * A method to determine whether a key down event was an arrow key
+   * */
+  isNavigationKeyup: (ev: KeyboardEvent): boolean => {
```
samtjo, line 24:
```suggestion
  isNavigationKeyup: (ev: KeyboardEvent | React.KeyboardEvent): boolean => {
```

I've realised from using these that they need to be able to handle React's synthetic events as well as native browser ones!

samtjo (approved):

grabkowski (dismissed):

divyajindel (dismissed):

grabkowski (approved):

divyajindel (approved):

PR #4983, opened 2022-03-21
refactor(button): convert button to typescript

samtjo:
### Proposed behaviour

- Convert Button to TypeScript.
- Convert TooltipProvider to TypeScript as the TypeScript definition was incorrect and Button needed it to be right.
- Fix the logic in Button for sharing a ref between forwardRef and internal use by using the `useImperativeHandle` hook from React.

### Current behaviour

Button and TooltipProvider are JavaScript components. 

### Checklist

#### QA

### Testing instructions


src/components/button/button.component.tsx
```diff
@@ -101,21 +193,29 @@ const Button = ({
...
+  const internalRef = useRef<HTMLButtonElement>(null);
+  useImperativeHandle<HTMLButtonElement | null, HTMLButtonElement | null>(
+    forwardRef,
+    () => internalRef.current
+  );
 
   const buttonType = as || buttonTypeProp;
 
-  if (subtext.length > 0 && size !== "large") {
+  if (subtext && subtext.length > 0 && size !== "large") {
```
edleeks87, line 208: This could just be the following as `""` is falsy

```suggestion
  if (subtext && size !== "large") {
```

edleeks87 (commented):
Just one small comment @samtjo, if another dev approves I'm happy to let it slide as it's not a big deal

src/components/button/button.component.tsx
samtjo: Yeah good spot, and this whole block could probably just be another invariant as well actually

src/components/button/button.component.tsx
```diff
@@ -101,21 +193,29 @@ const Button = ({
     );
   }
 
-  const [internalRef, setInternalRef] = useState(null);
```
mkrds, line 104: As far as I remember internalRef has to be set by using useState instead of useRef in this case because Tippy that is used in Tooltip is initially rendered with an empty ref. And to reinitialize itself when a ref is changed it has to receive a new `target` prop. When you use `useRef` instead of `useState` here it receives the same object but with a mutated `current` prop meaning it won't rerender and might cause accessibility issues - because that was the reason why I added this change 2 months ago

samtjo: Hmm interesting. I'll talk to you on Slack about this and see if we can replicate the issue you're talking about. Unfortunately bits of the change you added added invalid code in TypeScript - you can't set `forwardRef.current = reference` because `current` is a readonly property of a React reference. 

If we can replicate the issue you've mentioned then let's find another way to solve the problem.

edleeks87 (dismissed):

edleeks87 (approved):

src/components/button/index.d.ts
```diff
@@ -1,2 +0,0 @@
-export { default } from "./button";
-export * from "./button";
```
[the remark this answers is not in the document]
samtjo: I genuinely thought I'd deleted it, my mistake!

mkrds (dismissed):

edleeks87 (approved):

mkrds (approved):

DlgSHi (approved):

PR #4985, opened 2022-03-22
fix(pager): ensure id passed to page size select is unique

### LINKED ISSUE #4947: Accessibility issues when using Pager component
### Current behaviour

1) When multiple Pager components are used in a single webpage, axe reports as follows:
axe issue: IDs of active elements must be unique. Ensures every id attribute value of active elements is unique.
axe link: https://dequeuniversity.com/rules/axe/4.4/duplicate-id-active?application=AxeChrome
probable cause is pager component has id="page-select"
pager code link: https://github.com/Sage/carbon/blob/d191ea10a3fe7ab8370f1e6e784cdc5e23f0e2ab/src/components/pager/pager.component.js#L141

2) When page size options are displayed within the Pager component, axe reports as follows:
axe-issue: ARIA attributes must conform to valid values
axe-link: https://dequeuniversity.com/rules/axe/4.4/aria-valid-attr-value?application=AxeChrome
probable cause is select-list component has aria-labelledby attribute to which a guid is passed that has no links to any other element on the page.
pager code link: https://github.com/Sage/carbon/blob/d191ea10a3fe7ab8370f1e6e784cdc5e23f0e2ab/src/components/pager/pager.component.js#L135
select code link: https://github.com/Sage/carbon/blob/d191ea10a3fe7ab8370f1e6e784cdc5e23f0e2ab/src/components/select/simple-select/simple-select.component.js#L59

### Expected behaviour

Pager component should not report accessibility issues described in current behaviour.

N/A

### JIRA Ticket (Sage Only)

SBS-30534

### Suggested Solution

N/A

### Carbon Version

104.7.0

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4947

### Proposed behaviour

Adds guid to `id` passed to page size `Select`

### Current behaviour

`id` for page size `Select` is not unique

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/pager-size-options-mwpmf3?file=/src/index.js -- test the build version of this


samtjo (dismissed):

jamime (commented):
Can we remove the `id` instead? We don't use it for anything (pager has no label) and consumers wouldn't be able to find it because of the guid.

samtjo (approved):

divyajindel (approved):

Issue #4945, opened 2022-03-07
Button with Icon in FlatTable component: the icon of the buttons have the wrong color


hegou33:
### Current behaviour

With `sageTheme` theme, if we create a `FlatTable` and having a `Button with an Icon` in it, the `icon` don't have the same color than the text.

### Expected behaviour

With `sageTheme` theme, if we create a `FlatTable` and having a `Button with an Icon` in it, we expect that the `icon` have the same color than the text.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/gallant-albattani-40ulwx?file=/src/index.js

### Carbon Version

104.50.0

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
FE-4997

Issue #4904, opened 2022-02-25
Icon color in Table


gusch:
### Current behaviour

If I give a color to an Icon in a FlatTable, the color is overwritten by 
.kubjnu [data-component="icon"] {
    color: var(--colorsActionMinor500);
}

### Expected behaviour

The Icon has the wanted color

### CodeSandbox or Storybook URL

https://codesandbox.io/s/angry-dust-tr3ci9?file=/src/index.js

### Carbon Version

104.31.1

### What browsers are you seeing the problem on?

Firefox, Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

gusch:
Same problem for a Link with an Icon in a FlatTable

edleeks87:
FE-4997

PR #4996, opened 2022-03-25
fix(flat-table/sort): fix color of sort arrows

robinzigmond:
### Proposed behaviour

Fix regression in color of sort arrows in Flat Table

### Current behaviour
The arrow icons are the same color as the background and therefore not visible

### Checklist

#### QA

### Additional context

Update the relevant story to make the table already sorted - in order to make a chromatic snapshot that will catch any future regressions of this.

### Testing instructions

On the following story: https://carbon.sage.com/?path=/story/flat-table--with-sorting-headers - the sort arrows in the table header should be visible.

Also worth regression testing the following issues as the regression was introduced in a fix for them:  https://github.com/Sage/carbon/issues/4904, https://github.com/Sage/carbon/issues/4945


samtjo (changes requested):
The change looks good to me, let's add a unit test for the fix as well to prevent any regressions.

samtjo (dismissed):

edleeks87 (approved):

src/components/flat-table/sort/sort.component.js
```diff
@@ -33,7 +33,10 @@ const Sort = ({ children, onClick, sortType }) => {
       >
         {children}
         {sortType && (
-          <Icon type={sortType === "ascending" ? "sort_up" : "sort_down"} />
+          <Icon
+            type={sortType === "ascending" ? "sort_up" : "sort_down"}
+            color="--colorsUtilityMajor200"
```
jamime, line 38: Given the value is always the same, I think this should be defined in `flat-table-row-header.style.js`. This would allow any icon to be used, not just the sort component.

edleeks87 (approved):

PR #5013, opened 2022-04-01
refactor: convert theme selector plugin - FE-5114

jamime:
### Proposed behaviour

- The theme selector now uses the toolbar api.
- You can debug the chromatic builds using `&debug_chromatic`.
- Transitions are disabled during chromatic tests.
- Render the story after fonts have finished loading.

### Current behaviour

- We have a custom add-on that creates the toolbar button.
- Debugging chromatic doesn't work consistently (uses `theme` query param but the switcher uses `globals:theme`).
- Transitions are enabled which _may_ be causing some of our intermittent test issues.
- The stories are rendered immediately, this sometimes causes a race condition where the wrong font size is used for layout calculations.

### Checklist

#### QA

No consumer facing changes. Approval required.

### Additional context

N/A

### Testing instructions

Ensure all tests are passing.


cypress/support/helper.js
```diff
@@ -28,17 +29,29 @@ export function visitComponentUrlWithParameters(
...
   cy.fixture(`${path}/${json}`).then(($json) => {
     const el = $json[nameOfObject];
-    let url = "&args=";
+    let url = "";
+    let args = [];
+    let globals = [];
     for (const prop in el) {
       if (prop === "theme") {
-        url += `&theme=${encodeURIComponent(el[prop])}`;
+        globals.push(`${prop}:${encodeURIComponent(el[prop])}`);
```
edleeks87, line 37: As you're pushing the same string onto different arrays here, couldn't you just concat them onto the url string in the for loop instead of having the extra conditionals below?

eg:

```
url += `&${prop === "theme" ? "globals" : "args"}=${prop}:${encodeURIComponent(el[prop])}`
```

edleeks87, line 37: As you're pushing the same string onto different arrays here, couldn't you just concat them onto the url string in the for loop instead of having the extra conditionals below?

eg:

```url += `&${prop === "theme" ? "globals" : "args"}=${prop}:${encodeURIComponent(el[prop])}`
```

edleeks87: Discussed on slack and I'm happy to leave this as is

edleeks87 (approved):

cypress/support/helper.js
```diff
@@ -15,6 +15,7 @@ export function visitComponentUrl(component, suffix) {
   cy.visit(prepareUrl(component, suffix), {
     onBeforeLoad,
   });
+  cy.get("#root :first-child", { timeout: 10000 }).should("exist");
```
DlgSHi, line 18: I don't think we need one more `timeout` here. Because we already have it in line 54.

cypress/support/helper.js
```diff
@@ -28,17 +29,29 @@ export function visitComponentUrlWithParameters(
...
+    }
+
+    if (globals.length) {
+      url += `&globals=${globals.join(";")}`;
+    }
+
     cy.visit(`${prepareUrl(component, story)}${url}`, {
       onBeforeLoad,
     });
+    cy.get("#root :first-child", { timeout: 10000 }).should("exist");
```
DlgSHi, line 54: I think it is enough to have such timeout
```suggestion
    cy.get("#root :first-child", { timeout: 5000 }).should("exist");
```

cypress/support/helper.js
DlgSHi, line 37:
```suggestion
        globals.push(`${prop}=${encodeURIComponent(el[prop])}`);
```

DlgSHi (approved):

PR #5022, opened 2022-04-06
fix(simple-select): stop key filtering when command/control pressed

robinzigmond:
disable the searching for an option beginning with the pressed key if control or command is held
down at the same time, indicating the user is using a browser shortcut. This avoids unexpected
issues when using Voiceover and is consistent with how native HTML inputs work.

### Proposed behaviour

When using Simple Select and trying to use a keyboard shortcut involving the Control or Command key - eg Command/Control-F to find text on the page, the non-control key used should not be treated as a keypress to autoselect an option.

Also when using Voice Control. giving a command such as Find Next while focused on the textbox should not cause any option to be selected.

### Current behaviour

When pressing a key combination of Command/Control plus another key, the key pressed auto selects the first option beginning with that character. This also affects Voice Control, because that simulates keyboard events - so eg saying "Find Next" autoselects the first option beginning with G, because the keyboard shortcut for that is Command-G.

### Checklist

#### QA

### Testing instructions

Test in storybook: https://carbon.sage.com/?path=/story/select--default-story
Test both Voice Control and keyboard shortcuts involving Command or Control


samtjo (dismissed):

DipperTheDan (dismissed):

edleeks87 (approved):

harpalsingh:
chekced ux of both keyboard and voice control, all good.

edleeks87 (approved):

divyajindel (approved):

PR #5025, opened 2022-04-07
test(message): add refactored tests for message component FE-4551

divyajindel:
### Proposed behaviour

Refactor for `message` component to use new cypress-component-react framework for testing.

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [x] Run npx cypress open-ct to check if there is newly added test.file
- [x] Check the message.test.js file passed
- [x] Check all tests have been moved/refactored from .feature file -> .test.js file


src/components/message/message.test.js
```diff
@@ -0,0 +1,103 @@
...
+context("Tests for Message component", () => {
+  describe("should check Message component properties", () => {
+    it.each(["error", "info", "success", "warning"])(
+      "should check %s variant for Message component",
+      (variant) => {
+        CypressMountWithProviders(<MessageComponent variant={variant} />);
+        messagePreview().should(
+          "have.css",
+          "background-color",
+          "rgb(255, 255, 255)"
```
DlgSHi, line 39: there is no point fot this test, we are checking the same background for each variant

src/components/message/message.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    it("should check when open is false for Message component", () => {
+      CypressMountWithProviders(<MessageComponent open={false} />);
+      messagePreview().should("not.exist");
+    });
+
+    it("should check onDismiss for Message component", () => {
+      CypressMountWithProviders(<MessageComponent onDismiss />);
+      messagePreview().should("be.visible");
```
DlgSHi, line 71: we should check if the `onDismiss` is passed then the function runs when user click dismiss button and could be checked with `events`

src/components/message/message.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    it("should check showCloseIcon for Message component", () => {
+      CypressMountWithProviders(<MessageComponent showCloseIcon={false} />);
+      messageDismissIcon().should("not.exist");
+    });
+
+    it("should check close button aria-label for Message component", () => {
+      CypressMountWithProviders(
+        <MessageComponent closeButtonAriaLabel="shut" />
```
DlgSHi, line 98:
```suggestion
        <MessageComponent closeButtonAriaLabel="closed" />
```

src/components/message/message.test.js
```diff
@@ -0,0 +1,103 @@
...
+      );
+      messageDismissIcon().should("be.visible");
```
DlgSHi, line 100: we should the `aria-label` for `close icon` instead of visibility of `icon`

DlgSHi (changes requested):

src/components/message/message.test.js
```diff
@@ -0,0 +1,115 @@
...
+    it.each(testData)(
+      "should check %s title for Message component",
+      (title) => {
+        CypressMountWithProviders(<MessageComponent title={title} />);
+        messageTitle().should("have.text", title);
+      }
+    );
+
+    it("should check transparent background for Message component", () => {
+      CypressMountWithProviders(<MessageComponent transparent />);
```
stephenogorman, line 83: This is also boolean, so you could check both true and false, and pass in the background colour as a second parameter e.g.
      ["true", "rgb(204, 214, 219)"],
      ["false", "rgba(0, 0, 0, 0.9)"],
then verify:
      messagePreview().should(
        "have.css",
        "background-color",
        colour

src/components/message/message.test.js
```diff
@@ -0,0 +1,115 @@
...
+      messagePreview().should(
+        "have.css",
+        "background-color",
+        "rgba(0, 0, 0, 0)"
+      );
+    });
+
+    it("should check showCloseIcon for Message component", () => {
+      CypressMountWithProviders(<MessageComponent showCloseIcon={false} />);
```
stephenogorman, line 92: This is also boolean, so you could check true and false.

stephenogorman (changes requested):

stephenogorman (dismissed):

src/components/message/message.test.js
```diff
@@ -0,0 +1,135 @@
...
+        CypressMountWithProviders(<MessageComponent showCloseIcon={boolVal} />);
+        if (boolVal === true) {
+          messageDismissIcon().should("be.visible");
+        } else {
+          messageDismissIcon().should("not.exist");
+        }
+      }
+    );
+
+    it("should check close button aria-label for Message component", () => {
```
sianford, line 116: This test should use the string data type const to test with, similar to the children and title tests, as this is a custom aria label that can be set with any string.

src/components/message/message.test.js
```diff
@@ -0,0 +1,135 @@
...
+      CypressMountWithProviders(
+        <MessageComponent closeButtonAriaLabel="closed" />
+      );
+      closeButton().should("have.attr", "aria-label", "closed");
+    });
+
+    it("should check onDismiss for Message component", () => {
+      const callback = cy.stub();
+
+      CypressMountWithProviders(<MessageComponent onDismiss={callback} />);
```
sianford, line 126: Also test the `keydown` and `click` action events that fire when either the `Enter` key or mouseclick is used to close the message.

src/components/message/message.test.js
```diff
@@ -0,0 +1,135 @@
...
+    );
+
+    it("should set className for Message component", () => {
+      CypressMountWithProviders(
+        <MessageComponent className="message-cypress-classname" />
+      );
+      messagePreview().should("have.class", "message-cypress-classname");
+    });
+
+    it("should check custom id for Message component", () => {
```
sianford, line 64: This test should use the string data type const to test with, similar to the children and title tests, as this can be set with any string.

cypress/locators/message/locators.js
```diff
@@ -4,4 +4,8 @@ export const CLICK_ACTION = " > div:nth-child(2)";
 // component preview locators
 export const MESSAGE_PREVIEW = 'div[data-component="Message"]';
 export const MESSAGE_CHILDREN = 'div[data-element="content-body"]';
+export const MESSAGE_TITLE = '[data-element="content-title"]';
 export const MESSAGE_DISMISS_ICON = 'span[data-element="close"]';
+export const CLOSE_BUTTON = 'button[data-element="close"]';
+export const VARIANT_PREVIEW =
```
DlgSHi, line 10: I think we already have it defined above :)

cypress/locators/message/locators.js
```diff
@@ -4,4 +4,8 @@ export const CLICK_ACTION = " > div:nth-child(2)";
 // component preview locators
 export const MESSAGE_PREVIEW = 'div[data-component="Message"]';
 export const MESSAGE_CHILDREN = 'div[data-element="content-body"]';
+export const MESSAGE_TITLE = '[data-element="content-title"]';
 export const MESSAGE_DISMISS_ICON = 'span[data-element="close"]';
+export const CLOSE_BUTTON = 'button[data-element="close"]';
```
DlgSHi, line 9: this is defined in `common` locators

src/components/message/message.test.js
DlgSHi: where do you see the Message supports those events in documentation?

sianford: @DlgSHi I saw these event types in the Actions tab in Storybook. When you use the mouse/trackpad to close the Message it shows as a type `click` and when you use the Enter key it shows as a type `keydown`. I figured those actions should be tested too, but I'm possibly being overly thorough here. If you think that calling the onDismiss using click alone is sufficient then that's fine.

sianford (approved):

DlgSHi (approved):

PR #5026, opened 2022-04-07
feat(select): make select list respond to pageup/pagedown keys

robinzigmond:
### Proposed behaviour

When focused on a Select textbox and pressing the Page Up or Page Down keys (on keyboards lacking these specific keys, the same functionality can be triggered by pressing the up/down arrow while holding the Function key), the list of options should scroll up/down. Also when using Voice Control, the "scroll up" and "scroll down" commands should have the same effect.

### Current behaviour

Pressing Page Up/Down, or giving the Voice Control command scroll up/down, does nothing.

### Checklist

#### QA

### Testing instructions

Test each type/variation of Select and ensure that sensible behaviour is present for each of the following:
- pressing the pageup/pagedown key manually (Function+Up/Down)
- using Voice Control on Mac and giving the voice command "scroll up" or "scroll down"


DipperTheDan (commented):
Everything looks good to me. Only thing I would suggest is maybe adding the `PageUp` and `PageDown` keys to the `Scenario Outline: Open select list using arrow key` Cypress test we have in the `simpleSelect.feature` file. If you want a hand with it, just drop me a message but the Cypress docs are good and we should already have examples in our code 😄

DipperTheDan (approved):

src/components/select/utils/is-navigation-key.spec.js
```diff
@@ -0,0 +1,30 @@
...
+  describe("non-navigation keys", () => {
+    it.each(["a", "q", "z", "1", "0", " ", "Enter", "Tab", "Delete"])(
+      "%s is not counted as a navigation key",
+      (key) => {
+        expect(isNavigationKey(key)).toBe(false);
+      }
+    );
+  });
+
+  describe("left/right arrow keys are not counted as navigation keys as far as Select is concerned", () => {
```
samtjo, line 22:
```suggestion
  describe("left/right arrow keys", () => {
```
No need for the extra description here, now the full test output will read:

> isNavigationKey > left/right arrow keys > ArrowLeft/ArrowRight is not counted as a navigation key

src/components/select/utils/get-next-index-by-key.spec.js
```diff
@@ -5,56 +5,103 @@ describe("getNextIndexByKey", () => {
...
   const upKey = "ArrowUp";
   const downKey = "ArrowDown";
+  const pageUpKey = "PageUp";
+  const pageDownKey = "PageDown";
   const aKey = "a";
-  const list = ["a", "b", "c", "d", "e", "f"];
+  const list = ["a", "b", "c", "d", "e", "f", "g", "h"];
+  const lastIndex = list.length - 1;
+
+  const PAGE_SIZE = 4;
```
samtjo, line 14: Let's import this from the util file rather than declaring it twice

cypress/integration/common/simpleSelect.feature
```diff
@@ -48,6 +48,8 @@ Feature: Select component
       | key       | position | value  |
       | downarrow | first    | Amber  |
       | uparrow   | eleventh | Yellow |
+      | pagedown  | fourth   | Brown  |
```
samtjo, line 51: Where did we get to with updating this Cypress test? It needs to assert that the correct option is focused, not just that the select list is open.

src/components/select/utils/get-next-index-by-key.spec.js
robinzigmond: I guess you're referring to the PAGE_SIZE const?

I did consider doing that but decided that it's really a UX decision how big the "page" should be, and didn't want to end up in a position where if some developer decided to change the size to something else, the test would automatically pass without at least some acknowledgement of the fact that the behaviour has changed so the test should be updated too. I realise this goes against DRY but I'd rather the test was explicit about what it wants to test, instead of just saying whatever const is chosen in the util file is automatically fine.

Very happy to change this if you disagree, I just thought I should explain what I was thinking.

samtjo: Yeah I disagree, this is a constant so just need to be defined in the one place, if we don't have confidence that the test is correctly testing the behaviour regardless of what value it is, then the test should be rewritten to gain that confidence. 

We don't need to preempt and code around what future devs may or may not be thinking, let's change it 👍🏻

samtjo (approved):

DipperTheDan (approved):

src/components/select/utils/get-next-index-by-key.spec.js
```diff
@@ -1,60 +1,105 @@
-import getNextIndexByKey from "./get-next-index-by-key";
+import getNextIndexByKey, { PAGE_SIZE } from "./get-next-index-by-key";
```
divyajindel, line 1:
```suggestion
import { getNextIndexByKey, PAGE_SIZE } from "./get-next-index-by-key";
```

divyajindel (approved):

PR #5029, opened 2022-04-08
docs: rfc to remove support for uncontrolled inputs - FE-4341

DipperTheDan:
Proposal to remove uncontrolled support for our Carbon input components.

---

Please use 👍  if you approve.

### DIFF
--- a/rfcs/text/remove-support-for-uncontrolled-inputs.md
+++ b/rfcs/text/remove-support-for-uncontrolled-inputs.md
@@ -1,0 +1,235 @@
+- Start Date: 31/03/2022
+
+# Table of contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Basic examples](#basic-examples)
+  - [Component Example - Controlled and Uncontrolled](#component-example---controlled-and-uncontrolled)
+  - [Component Example - Controlled](#component-example---controlled)
+- [Detailed design](#detailed-design)
+  - [Components](#components)
+- [Unit Testing](#unit-testing)  
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Compatibility](#compatibility) 
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+- [Documentation](#documentation)
+- [Further Considerations](#further-considerations)
+
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon input components in an uncontrolled manner.
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we have to account for an input being either uncontrolled or controlled and therefore have to test both of these use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.  
+- Uncontrolled behaviour is not the promoted approach to handling inputs in React. React state in their docs that ["In most cases, we recommend using controlled components to implement forms."](https://reactjs.org/docs/uncontrolled-components.html)
+- React form libraries, such as Formik, strongly encourage using controlled components. As can be seen in their docs regarding [using handleChange](https://formik.org/docs/api/formik#handlechange-e-reactchangeeventany--void) and [initialValues](https://formik.org/docs/api/formik#initialvalues-values).
+- To bring the rest of our components in line with the recently refactored Date component which only supports being controlled.
+
+# Basic Examples
+
+## Component Example - Controlled and Uncontrolled
+
+In the current implementation of supporting controlled and uncontrolled, a component will look like this.
+
+```jsx 
+// my-component.component.js
+const MyComponent = ({ value, onChange, defaultValue, ...rest }) => {
+  /* Component logic goes here */
+  const isControlled = useRef(value !== undefined);
+  const inputRef = useRef(null);
+
+  const handleOnChange = (event) => {
+    if (isControlled.current && onChange) {
+      onChange(event);
+    }
+  };
+
+  const prevControlledRef = useRef();
+
+  useEffect(() => {
+    const message =
+      "Input elements should not switch from uncontrolled to controlled (or vice versa)."
+
+    invariant(prevControlledRef.current !== isControlled, message);
+
+    prevControlledRef.current = isControlled
+
+  }, [isControlled]);
+
+  return (
+    <input
+      value={value}
+      onChange={handleOnChange}
+      defaultValue={defaultValue}
+      ref={inputRef}
+    />
+  );
+};
+
+MyComponent.propTypes = {
+  value: PropTypes.string,
+  onChange: PropTypes.function,
+  defaultValue: PropTypes.string,
+};
+
+export default MyComponent;
+```
+Here is an example of how a basic component will look in TS with uncontrolled support.
+
+```tsx
+// my-component.component.tsx
+interface MyComponentProps = {
+  value?: string;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  defaultValue?: string;
+};
+
+const MyComponent = ({ value, onChange, defaultValue, ...rest }: MyComponentProps) => {
+  const isControlled = useRef(value !== undefined);
+  const inputRef = useRef(null);
+
+   const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (isControlled && onChange ) {
+      onChange(event);
+    } 
+  };
+
+  useEffect(() => {
+    // invariant goes in here.
+  }, [];
+
+  /* Component logic goes here */
+
+return (
+    <input 
+      value={value} 
+      onChange={handleOnChange} 
+      defaultValue={defaultValue} 
+      ref={inputRef}
+    />
+  )
+};
+
+export default MyComponent;
+```
+## Component Example - Controlled
+
+Here is an example of how a basic component will look in JS without uncontrolled support.
+
+```jsx 
+// my-component.component.js
+const MyComponent = ({ value, onChange, ...rest }) => {
+   const handleOnChange = (event) => {
+    if (onChange) {
+      onChange(event);
+    } 
+  };
+
+  /* Component logic goes here */
+
+  return <input value={value} onChange={handleOnChange} />;
+};
+
+MyComponent.propTypes = {
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.function.isRequired,
+};
+
+export default MyComponent;
+```
+Here is an example of how a basic component will look in TS without uncontrolled support.
+
+```tsx
+// my-component.component.tsx
+interface MyComponentProps = {
+  value: string;
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+const MyComponent = ({ value, onChange, ...rest }: MyComponentProps) => {
+   const handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (onChange) {
+      onChange(event.target.value);
+    } 
+  };
+
+  /* Component logic goes here */
+
+  return <input value={value} onChange={handleOnChange} />;
+};
+
+export default MyComponent;
+```
+# Detailed design
+
+- We will add deprecation warnings to our input components so that our consumers are aware of the pending changes to remove uncontrolled support. This will appear as a console warning if our consumer is using the uncontrolled mode and a warning on our Storybook documentation.
+
+- We will remove uncontrolled behaviour from our input components and associated props. This will be done in one Pull Request with each component having it's own commit message with a `BREAKING CHANGE`. The benefit of this is that we will only release one major version of Carbon. 
+
+## Components
+
+The input components that currently support being uncontrolled are: 
+1. Textbox
+1. Decimal
+1. Number Input
+1. Simple Select
+1. Filterable Select
+1. Multi Select
+1. Grouped Character
+1. Numeral Date
+1. Search
+1. Checkbox 
+1. RadioButton 
+1. Switch 
+1. SimpleColorPicker 
+1. AdvancedColorPicker 
+1. ButtonToggle
+
+## Unit Testing
+
+Currently we have unit tests for both controlled and uncontrolled behaviours. As part of this work we could reduce the amount of code in the spec files quite considerably. If we take `Decimal` as an example of this, `Decimal` currently has 1430 lines of code in the spec file. This could be reduced to down to around 480 lines of code if we were to remove the uncontrolled tests. This would make the tests easier to read and maintain, while also reducing the runtime of the tests.
+# Drawbacks
+
+- This work will be a `BREAKING CHANGE`, which we will communicate in the appropriate Carbon channels. Consumers of Carbon will also have the opportunity to comment on this RFC.
+
+- Any consumers of Carbon that currently use uncontrolled components will have to rewrite their form implementations.  
+
+# Alternatives
+
+The alternative is to do nothing: we leave the input components as they are and continue to support both controlled and uncontrolled behaviours. This course of action would mean no impact to our consumers. 
+
+By following this course of action though we would not get any of the benefits detailed above.
+
+# Compatibility 
+
+Compatibility was checked with both `react-hook-form` and `Formik`. `react-hook-form` was more performant and you can see the code used and the results [here](https://github.com/Sage/sbc.research.react-hook-form/blob/master/README.md).
+
+# Adoption strategy
+
+This proposed work will be implemented as a `BREAKING CHANGE`. In terms of how consuming projects will upgrade Carbon, we will need to document the upgrade process and link to it in the release notes. If a project has used an input as uncontrolled, they will need to update their codebase to make this controlled instead. This will involve making sure that the value is stored using React's [useState](https://reactjs.org/docs/hooks-state.html) hook to retain the value and making sure this is managed using an `onChange` function. We may have to provide additional support in this scenario to provide examples of how they can achieve the same functionality using this approach.
+
+We also already provide various codemods in our [Carbon Codemod](https://github.com/Sage/carbon-codemod) repo that could be used by our consumers to assist with the move to a controlled component. Examples of this are:
+- A [rename-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/rename-prop) codemod that consumers could use to change a `defaultValue` value prop to `value`.
+- A [add-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/add-prop) codemod that could be used to add to a `value` prop.
+- A [remove-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/remove-prop) codemod that could be used to remove a `defaultValue` prop.
+- A [replace-prop-value](https://github.com/Sage/carbon-codemod/tree/master/transforms/replace-prop-value) codemod that changes the value of a prop.
+
+There is also support for writing controlled components in React's documentation. This can be found [here](https://reactjs.org/docs/forms.html#controlled-components).
+
+# How we teach this
+
+This RFC will be communicated in the Sage `#carbon` and `#carbon-announcement` slack channels, where we will ask our consumers for feedback on the RFC. Beyond that, it will be visible in the release notes when consumers upgrade. This RFC document will reside in the repository for future reference.
+
+We will be able to support consumers of Carbon in the `#carbon` slack channel too. 
+
+# Documentation
+
+As part of this work we will make sure that the documentation reflects what we support. We will update the component description to include that we specifically only support controlled usage of these components. We will also ensure Storybook examples are updated to remove or update any that use uncontrolled components.
+
+# Further Considerations
+
+If/when we introduce a `FileInput` component into Carbon, this will have to be an uncontrolled component. This is because `<input type="file" />` can only be uncontrolled as [the value can only be specified by the user and not be done programmatically](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag). As file inputs cannot be controlled, the Carbon `FileInput` component would have to be uncontrolled.


rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+- A [remove-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/remove-prop) codemod that could be used to remove a `defaultValue` prop.
+- A [replace-prop-value](https://github.com/Sage/carbon-codemod/tree/master/transforms/replace-prop-value) codemod that changes the value of a prop.
+
+There is also a lot of support for writing controlled components.
+
+# How we teach this
+
+This RFC will be communicated in the Sage `#carbon` and `#carbon-announcement` slack channels, where we will ask our consumers to feedback on the RFC. Beyond that, it will be visible in the release notes when consumers upgrade. This RFC document will reside in the repository for future reference.
+
+We will be able to support
```
edleeks87, line 216: is this missing the rest of a sentence?

DipperTheDan: Yes 🤣 Thanks for spotting that.
It should be something like "We will be able to support consumers of Carbon in the `#carbon` slack channel too."

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+
+We also already provide various codemods in our [Carbon Codemod](https://github.com/Sage/carbon-codemod) repo that could be used by our consumers to assist with the move to a controlled component. Examples of this are:
+- A [rename-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/rename-prop) codemod that consumers could use to change a `defaultValue` value prop to `value`.
+- A [add-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/add-prop) codemod that could be used to add to a `value` prop.
+- A [remove-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/remove-prop) codemod that could be used to remove a `defaultValue` prop.
+- A [replace-prop-value](https://github.com/Sage/carbon-codemod/tree/master/transforms/replace-prop-value) codemod that changes the value of a prop.
+
+There is also a lot of support for writing controlled components.
```
jamime, line 210: Where?

DipperTheDan: That is a valid question. I shall make this more specific 👍🏻

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+1. Simple Select
+1. Filterable Select
+1. Multi Select
+1. Grouped Character
+1. Numeral Date
+1. Search
+
+# Unit Testing
+
+Currently uncontrolled and controlled behaviours are unit tested. As part of this work we could reduce the amount of code in the spec files quite considerably. If we take `Decimal` as an example of this, `Decimal` currently has 1430 lines of code in the spec file. This could be reduced to down to around 480 lines of code if we were to remove the uncontrolled tests. This would make the tests easier to read and maintain.
```
samtjo, line 190: and reduce the runtime of the tests

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+
+# How we teach this
+
+This RFC will be communicated in the Sage `#carbon` and `#carbon-announcement` slack channels, where we will ask our consumers to feedback on the RFC. Beyond that, it will be visible in the release notes when consumers upgrade. This RFC document will reside in the repository for future reference.
+
+We will be able to support consumers of Carbon in the #carbon slack channel too. 
+
+# Documentation
+
+As part of this work we will ensure that Storybook examples are updated where any uncontrolled behaviours are demonstrated.
```
samtjo, line 220: Updated in what way? A bit more info needed here I think.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+
+# Further Considerations
+
+If/when we introduce a `File Input` component in to Carbon, this will have to be an uncontrolled component. This is because `<input type="file" />` can only be uncontrolled as [the value can only be specified by the user and not be done programmatically](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag). As a file input only works uncontrolled, we would not need to maintain any controlled behaviour.
```
samtjo, line 224:
```suggestion
If/when we introduce a `FileInput` component into Carbon, this will have to be an uncontrolled component. This is because `<input type="file" />` can only be uncontrolled as [the value can only be specified by the user and not be done programmatically](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag). As a file input only works uncontrolled, we would not need to maintain any controlled behaviour.
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+- [Documentation](#documentation)
+- [Further Considerations](#further-considerations)
+
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon inputs in an uncontrolled manner.
```
samtjo, line 21:
```suggestion
The purpose of this document is to outline our intention to remove support for using Carbon input components in an uncontrolled manner.
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we are having to account for an input being uncontrolled or controlled and therefore having to test these functionalities. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.
```
samtjo, line 25:
```suggestion
- To reduce the code complexity in our input components. Currently we have to account for an input being uncontrolled or controlled and therefore have to test these both use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.  
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon inputs in an uncontrolled manner.
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we are having to account for an input being uncontrolled or controlled and therefore having to test these functionalities. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.  
+- Uncontrolled behaviour is not the promoted approach to handling inputs in React. React state in their docs that ["In most cases, we recommend using controlled components to implement forms."](https://reactjs.org/docs/uncontrolled-components.html)
+- Libraries, such as Formik, strongly encourage using controlled components. As can be seen in their docs regarding [using handleChange](https://formik.org/docs/api/formik#handlechange-e-reactchangeeventany--void) and [initialValues](https://formik.org/docs/api/formik#initialvalues-values).
```
samtjo, line 27:
```suggestion
- React form libraries, such as Formik, strongly encourage using controlled components. As can be seen in their docs regarding [using handleChange](https://formik.org/docs/api/formik#handlechange-e-reactchangeeventany--void) and [initialValues](https://formik.org/docs/api/formik#initialvalues-values).
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+- We will add deprecation warnings to our input components so that our consumers are aware of the pending changes to remove uncontrolled support. This will appear as a console warning and a warning on our Storybook documentation.
+
+- We will remove uncontrolled behaviour from our input components and associated props piecemeal. Each of these will be a `BREAKING CHANGE`.
+
+This approach will provide immediate benefit to developers of Carbon as it means we will have less code and tests to maintain. It will also mean that consumers of Carbon can steadily upgrade their version of Carbon.
+
+# Components
+
+The input components that currently support being uncontrolled are: 
+1. Textbox
```
samtjo, line 178: `Switch`?

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+1. Numeral Date
+1. Search
+
+# Unit Testing
+
+Currently uncontrolled and controlled behaviours are unit tested. As part of this work we could reduce the amount of code in the spec files quite considerably. If we take `Decimal` as an example of this, `Decimal` currently has 1430 lines of code in the spec file. This could be reduced to down to around 480 lines of code if we were to remove the uncontrolled tests. This would make the tests easier to read and maintain.
+
+# Drawbacks
+
+This work will be a `BREAKING CHANGE` however we will communicate this in the appropriate Carbon channels. Consumers of Carbon will also have the opportunity to comment on this RFC.
```
samtjo, line 194: Maybe worth mentioning the effort involved for users who use uncontrolled components currently to rewrite their form implementations in this drawbacks section

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+      ref={inputRef}
+    />
+  )
+};
+
+export default MyComponent;
+```
+## Component Example - Controlled
+
+Here is a basic example of how a basic component will look in JS without uncontrolled.
```
mkrds, line 121:
```suggestion
Here is a basic example of how a basic component will look in JS without uncontrolled support.
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+};
+
+MyComponent.propTypes = {
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.function.isRequired,
+};
+
+export default MyComponent;
+```
+Here is a basic example of how a basic component will look in TS without uncontrolled.
```
mkrds, line 144:
```suggestion
Here is a basic example of how a basic component will look in TS without uncontrolled support.
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+The input components that currently support being uncontrolled are: 
+1. Textbox
+1. Decimal
+1. Number Input
+1. Simple Select
+1. Filterable Select
+1. Multi Select
+1. Grouped Character
+1. Numeral Date
+1. Search
```
mkrds, line 186: Also Checkbox, RadioButton, Switch, SimpleColorPicker, ButtonToggle

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: @mkrds has added on the ones I've left out here https://github.com/Sage/carbon/pull/5029#discussion_r860627422

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,224 @@
...
+  return <input value={value} onChange={handleOnChange} />;
+};
+
+export default MyComponent;
+```
+# Detailed design
+
+- We will add deprecation warnings to our input components so that our consumers are aware of the pending changes to remove uncontrolled support. This will appear as a console warning and a warning on our Storybook documentation.
+
+- We will remove uncontrolled behaviour from our input components and associated props piecemeal. Each of these will be a `BREAKING CHANGE`.
```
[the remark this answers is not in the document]
nicktitchmarsh: One PR is fine with separate commits, let's keep this down to one breaking change.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,232 @@
...
+
+# How we teach this
+
+This RFC will be communicated in the Sage `#carbon` and `#carbon-announcement` slack channels, where we will ask our consumers to feedback on the RFC. Beyond that, it will be visible in the release notes when consumers upgrade. This RFC document will reside in the repository for future reference.
+
+We will be able to support consumers of Carbon in the `#carbon` slack channel too. 
+
+# Documentation
+
+As part of this work we will ensure that Storybook examples are updated where any uncontrolled behaviours are demonstrated. This will involve removing any uncontrolled examples and ensuring we have the controlled examples.
```
samtjo, line 228: We should also be documenting that we don't support uncontrolled usage in these components otherwise we'll get issues raised about it.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,232 @@
...
+1. Numeral Date
+1. Search
+1. Checkbox 
+1. RadioButton 
+1. Switch 
+1. SimpleColorPicker 
+1. AdvancedColorPicker 
+1. ButtonToggle
+
+## Unit Testing
```
samtjo, line 194: Needs to be added to the contents

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: Good plan. I'll add something in here that we'll specifically state in the component description that we only support controlled usage.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+      ref={inputRef}
+    />
+  )
+};
+
+export default MyComponent;
+```
+## Component Example - Controlled
+
+Here is a basic example of how a basic component will look in JS without uncontrolled support.
```
edleeks87, line 122: Can we say basic less in this sentence? 🤣

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+MyComponent.propTypes = {
+  value: PropTypes.string,
+  onChange: PropTypes.function,
+  defaultValue: PropTypes.string,
+};
+
+export default MyComponent;
+```
+Here is a basic example of how a basic component will look in TS with uncontrolled support.
```
edleeks87, line 82: Can we say basic less in this sentence? 🤣

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+};
+
+MyComponent.propTypes = {
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.function.isRequired,
+};
+
+export default MyComponent;
+```
+Here is a basic example of how a basic component will look in TS without uncontrolled support.
```
edleeks87, line 145: Can we say basic less in this sentence? 🤣

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+- [Documentation](#documentation)
+- [Further Considerations](#further-considerations)
+
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon input components in an uncontrolled manner.
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we have to account for an input being uncontrolled or controlled and therefore have to test these both use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.
```
robinzigmond, line 26: I would say `Currently we have to account for an input being either uncontrolled or controlled` (ie add the word "either") to make it clearer that currently there are 2 cases we have to consider. Might just be me but it reads a bit strange to me without this.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon input components in an uncontrolled manner.
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we have to account for an input being uncontrolled or controlled and therefore have to test these both use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.  
+- Uncontrolled behaviour is not the promoted approach to handling inputs in React. React state in their docs that ["In most cases, we recommend using controlled components to implement forms."](https://reactjs.org/docs/uncontrolled-components.html)
+- React form libraries, such as Formik, strongly encourage using controlled components. As can be seen in their docs regarding [using handleChange](https://formik.org/docs/api/formik#handlechange-e-reactchangeeventany--void) and [initialValues](https://formik.org/docs/api/formik#initialvalues-values).
+- To bring the rest of our components inline with the recently refactored Date component which only supports being controlled.
```
robinzigmond, line 29: Should be "in line" rather than "inline". See http://www.kirkmahoney.com/blog/2007/11/in-line-vs-inline/

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+      onChange(event);
+    }
+  };
+
+  useEffect(() => {
+    const message =
+      "Input elements should not switch from uncontrolled to controlled (or vice versa)."
+
+    /* invariant takes two parameters.
+        First parameter is the condition in which the invariant should be triggered. In this example this is invariantCondition just for illustration purposes. Second parameter is the message that is displayed on the users screen.
```
robinzigmond, line 55: I think this is potentially a little confusing as written, and we should make it clear that in this context `invariantCondition` isn't some random unrelated thing but relates to the controlled-ness of the component having been changed.

In fact, since it's barely any code, I don't see why we couldn't effectively copy in here the actual implementation from eg Decimal, ie with `const prevControlledRef = useRef();`, use `invariant(prevControlledRef.current !== isControlled, message);` and have `prevControlledRef.current = isControlled` just below.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+1. Checkbox 
+1. RadioButton 
+1. Switch 
+1. SimpleColorPicker 
+1. AdvancedColorPicker 
+1. ButtonToggle
+
+## Unit Testing
+
+Currently uncontrolled and controlled behaviours are unit tested. As part of this work we could reduce the amount of code in the spec files quite considerably. If we take `Decimal` as an example of this, `Decimal` currently has 1430 lines of code in the spec file. This could be reduced to down to around 480 lines of code if we were to remove the uncontrolled tests. This would make the tests easier to read, maintain and reduce the runtime of the tests too.
```
robinzigmond, line 197: similar to my first comment above, I would add the word "both" - `... are both unit tested` - as I think it's clearer. Or perhaps better still, rewrite the sentence as `Currently we have unit tests for both controlled and uncontrolled behaviours.`

Also the last sentence is not quite grammatical - we're not making it "easier to reduce the runtime", we're actually reducing the runtime. I think the phrasing should be `This would make the tests easier to read and maintain, while also reducing the runtime of the tests.` ("while also reducing" could be "and also reduce", or if you prefer to keep the "too" you can have that at the end and remove "also". Many options!)

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+# Drawbacks
+
+- This work will be a `BREAKING CHANGE` however we will communicate this in the appropriate Carbon channels. Consumers of Carbon will also have the opportunity to comment on this RFC.
```
robinzigmond, line 201: I prefer `This work will be a BREAKING CHANGE, which we will communicate in the appropriate Carbon channels.` I'm not sure "however" fits as posting in Carbon channels isn't anything out of the ordinary for a breaking change. (Or is it?)

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Any consumers of Carbon that currently use uncontrolled components will have to rewrite their form implementations.
```
robinzigmond, line 203: I can't really explain why but `rewrite their form implementations` sounds a bit clumsy to me. I would just say `will have to rewrite their application` (and perhaps add `to use controlled components instead` to the end but that's not really necessary).

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+# Drawbacks
+
+- This work will be a `BREAKING CHANGE` however we will communicate this in the appropriate Carbon channels. Consumers of Carbon will also have the opportunity to comment on this RFC.
+
+- Any consumers of Carbon that currently use uncontrolled components will have to rewrite their form implementations.  
+
+# Alternatives
+
+Do nothing; we leave the input components as they are and continue to support controlled and uncontrolled behaviours. This course of action would mean no impact to our consumers. Uncontrolled input components can be used with Formik, even though this is not encouraged, through implementing a store type (React's [useReducer hook](https://reactjs.org/docs/hooks-reference.html#usereducer) or [Redux](https://redux.js.org/) for example) mechanism in order to retrieve a value. However, we would always still recommend using controlled input components with Formik.
```
robinzigmond, line 207: Doesn't the heading "alternatives" indicate some sort of bullet list? I realise that wouldn't be appropriate here when there's only one, but perhaps emphasise this by changing the first sentence to `The alternative is to do nothing: we leave...`.

As above I recommend using the word "both" before "controlled and uncontrolled"

I also don't think the semicolon is correct, it should be a colon. or perhaps a dash. Not too bothered about this though.

Also this whole paragraph feels a little aimless and disjointed. I think it should certainly add the downside of doing nothing - ie we wouldn't get the benefits detailed above. Also perhaps put a paragraph break where the subject changes a bit. In short I'd rewrite this paragraph as follows - but feel free to improve it or even not take on all my suggestions!

```
The alternative is to do nothing: we leave the input components as they are and continue to support both controlled and uncontrolled behaviours. This course of action would mean no impact to our consumers. 

Uncontrolled input components can be used with Formik, even though this is not encouraged, through implementing a store type (React's [useReducer hook](https://reactjs.org/docs/hooks-reference.html#usereducer) or [Redux](https://redux.js.org/) for example) mechanism in order to retrieve a value. However, we would always still recommend using controlled input components with Formik.

By following this course of action though we would not get any of the benefits detailed above.
```

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+- A [rename-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/rename-prop) codemod that consumers could use to change a `defaultValue` value prop to `value`.
+- A [add-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/add-prop) codemod that could be used to add to a `value` prop.
+- A [remove-prop](https://github.com/Sage/carbon-codemod/tree/master/transforms/remove-prop) codemod that could be used to remove a `defaultValue` prop.
+- A [replace-prop-value](https://github.com/Sage/carbon-codemod/tree/master/transforms/replace-prop-value) codemod that changes the value of a prop.
+
+There is also support for writing controlled components in React's documentation. This can be found [here](https://reactjs.org/docs/forms.html#controlled-components).
+
+# How we teach this
+
+This RFC will be communicated in the Sage `#carbon` and `#carbon-announcement` slack channels, where we will ask our consumers to feedback on the RFC. Beyond that, it will be visible in the release notes when consumers upgrade. This RFC document will reside in the repository for future reference.
```
robinzigmond, line 223: "ask our consumers ~~to~~ for feedback on the RFC"

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+We will be able to support consumers of Carbon in the `#carbon` slack channel too. 
+
+# Documentation
+
+As part of this work we will make sure that the documentation reflects what we will support. We will update the component description to include that we specifically only support controlled usage of these components. We will also ensure Storybook examples are updated where any uncontrolled behaviours are demonstrated. This will involve removing any uncontrolled examples and ensuring we have the controlled examples.
```
robinzigmond, line 229: `reflects what we will support` - I don't think the "will" is necessary, as this is talking about a future time after the support for uncontrolled components is removed.

Also the 2 sentences about updating Storybook documentation would be clearer as just 1 imo: `We will also ensure Storybook examples are updated to remove or update any that use uncontrolled components.`

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,233 @@
...
+
+# Further Considerations
+
+If/when we introduce a `FileInput` component into Carbon, this will have to be an uncontrolled component. This is because `<input type="file" />` can only be uncontrolled as [the value can only be specified by the user and not be done programmatically](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag). As a file input only works uncontrolled, we would not need to maintain any controlled behaviour.
```
robinzigmond, line 233: The last sentence here is a little confusing, and I'm not sure it adds any value anyway - I think the previous 2 sentences and accompanying link say it all. If we do keep it I prefer a phrasing such as `As file inputs cannot be controlled, the Carbon FileInput component would have to be uncontrolled.`

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: It's only something we've started doing recently. Not something we would have done historically. I took inspiration from previous RFC's here so I guess your point is valid as it makes more sense in the here and now.

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: I'll get this added 👍🏻

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: Valid point. I'll get this changed 😄

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: To me `rewrite their application` implies that the whole app will need rewriting because of this change. If I were a consumer of Carbon reading that, I'd be quite alarmed.

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: Good point.

rfcs/text/remove-support-for-uncontrolled-inputs.md
robinzigmond: OK fair enough. Perhaps then, instead of "rewrite...", say something like "update their uncontrolled components to be controlled".

But the original isn't bad, I think it's just a weird unexplainable personal thing that it reads a bit weird to me. So feel free to leave this as is.

DipperTheDan: I'll leave as is if that is OK? Sam suggest this in this comment.
https://github.com/Sage/carbon/pull/5029#discussion_r859749087

robinzigmond: Fair enough. I don't think my suggested change contradicts Sam's, I agree with him that we should have a note about consumers having to rewrite, I just think this can be phrased a bit better. But honestly, this was the one of my comments I felt least strongly about so happy to let it go :)

Parsium (dismissed):

grabkowski (approved):

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,234 @@
...
+
+export default MyComponent;
+```
+# Detailed design
+
+- We will add deprecation warnings to our input components so that our consumers are aware of the pending changes to remove uncontrolled support. This will appear as a console warning if our consumer is using the uncontrolled mode and a warning on our Storybook documentation.
+
+- We will remove uncontrolled behaviour from our input components and associated props. This will be done in one Pull Request with each component having it's own commit message with a `BREAKING CHANGE`. The benefit of this is that we will only release one major version of Carbon. 
+
+This approach will provide immediate benefit to developers of Carbon as it means we will have less code and tests to maintain. It will also mean that consumers of Carbon can steadily upgrade their version of Carbon.
```
jamime, line 172:
```suggestion
```

We're doing one release so consumers will have to take a big change.

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,234 @@
...
+
+- This work will be a `BREAKING CHANGE`, which we will communicate in the appropriate Carbon channels. Consumers of Carbon will also have the opportunity to comment on this RFC.
+
+- Any consumers of Carbon that currently use uncontrolled components will have to rewrite their form implementations.  
+
+# Alternatives
+
+The alternative is to do nothing: we leave the input components as they are and continue to support both controlled and uncontrolled behaviours. This course of action would mean no impact to our consumers. 
+
+Uncontrolled input components can be used with Formik, even though this is not encouraged, through implementing a store type (React's [useReducer hook](https://reactjs.org/docs/hooks-reference.html#usereducer) or [Redux](https://redux.js.org/) for example) mechanism in order to retrieve a value. However, we would always still recommend using controlled input components with Formik.
```
jamime, line 206: I don't understand this paragraph. To use a reducer you'd need to use a controlled input, unless you're meaning to use `onChange` to maintain the reducer state?

jamime (changes requested):
Please can you check if this would work with https://react-hook-form.com and mention it in the docs in it's own section. There has been increasing interest in this as we've seen degraded performance in certain situations using formik. In particular a single keypress causing a re-render of all inputs on the page.

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: Removed this point.

rfcs/text/remove-support-for-uncontrolled-inputs.md
DipperTheDan: Removed this paragraph as discussed.

jamime (approved):

edleeks87 (dismissed):

rfcs/text/remove-support-for-uncontrolled-inputs.md
```diff
@@ -0,0 +1,235 @@
...
+- [Documentation](#documentation)
+- [Further Considerations](#further-considerations)
+
+# Summary
+
+The purpose of this document is to outline our intention to remove support for using Carbon input components in an uncontrolled manner.
+
+# Motivation
+
+- To reduce the code complexity in our input components. Currently we have to account for an input being either uncontrolled or controlled and therefore have to test these both use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.
```
Parsium, line 27:
```suggestion
- To reduce the code complexity in our input components. Currently we have to account for an input being either uncontrolled or controlled and therefore have to test both of these use cases. We would be able to remove any code relating to invariant messages regarding switching from controlled to uncontrolled or vice versa too.  
```

Sorry to be that guy 😅

edleeks87 (approved):

nicktitchmarsh (approved):

PR #5066, opened 2022-04-19
refactor(global-style): convert file to typescript FE-4352 

edleeks87:
### Proposed behaviour

Refactors `style` directory files to TypeScript

### Current behaviour

The `style` directory files are JavaScript

### Checklist

#### QA

### Testing instructions

Files changed relate to theme and some style utils so should just be covered by chromatic and only require manual testing as an additional check for things like the spacing props (the filter utils have also been refactored)


src/global.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@styled-system/prop-types";
```
samtjo, line 1: We shouldn't need this, what was your reason for adding it?

src/style/design-tokens/carbon-scoped-tokens-provider/carbon-scoped-tokens-provider.component.ts
```diff
@@ -9,7 +10,9 @@ import generateCssVariables from "../generate-css-variables.util";
  *
  */
 
-const CarbonScopedTokensProvider = styled.div`
+const CarbonScopedTokensProvider = styled.div<
+  ThemeObject | Record<string, string>
```
samtjo, line 14: Why are we giving the option to pass anything other than a `ThemeObject` into this?

src/style/global-style.ts
```diff
@@ -1,6 +1,9 @@
-import { createGlobalStyle } from "styled-components";
+import { createGlobalStyle, GlobalStyleComponent } from "styled-components";
+import { ThemeObject } from "./themes/base/base-theme.config";
 
-const GlobalStyle = createGlobalStyle`
+const GlobalStyle = createGlobalStyle<
```
samtjo, line 4: This can just be as it was before, we're not passing any props into `createGlobalStyle` so this is unnecessary.

Also `GlobalStyleComponent` is what's returned by `createGlobalStyle` so this would be wrong anyway!

src/style/themes/base/index.ts
```diff
@@ -10,22 +10,26 @@ const colorsWithHex = addHexSymbols(baseColors);
...
 };
 
 const baseTheme = configureBase(palette);
 
 export default baseTheme;
 
-export const mergeWithBase = (configureTheme) => {
+export const mergeWithBase = (
+  configureTheme: (p: BasePalette) => Partial<ThemeObject>
+): ThemeObject => {
```
samtjo, line 26:
```suggestion
): Partial<ThemeObject> => {
```

src/style/themes/base/index.ts
```diff
@@ -10,22 +10,26 @@ const colorsWithHex = addHexSymbols(baseColors);
...
   const themeToMergeWithBase = configureTheme(palette);
 
   return {
     ...mergeDeep(baseTheme, themeToMergeWithBase),
     palette,
-  };
+  } as ThemeObject;
```
samtjo, line 32:
```suggestion
  };
```

src/style/utils/filter-styled-system-margin-props.ts
```diff
@@ -1,6 +1,6 @@
 import filterObjectProperties from "../../__internal__/filter-object-properties";
```
samtjo, line 2: `import { MarginProps } from "styled-system";`

src/style/utils/filter-styled-system-margin-props.ts
```diff
@@ -17,6 +17,8 @@ const marginPropertyNames = [
   "my",
 ];
 
-export default function filterStyledSystemMarginProps(originalObject) {
+export default function filterStyledSystemMarginProps(
+  originalObject: Record<string, unknown>
+): Record<string, unknown> {
```
samtjo, line 22:
```suggestion
): MarginProps {
```

src/style/utils/filter-styled-system-padding-props.ts
```diff
@@ -17,6 +17,8 @@ const paddingPropertyNames = [
   "py",
 ];
 
-export default function filterStyledSystemPaddingProps(originalObject) {
+export default function filterStyledSystemPaddingProps(
+  originalObject: Record<string, unknown>
+): Record<string, unknown> {
```
samtjo, line 22:
```suggestion
): PaddingProps {
```

samtjo (changes requested):

src/global.d.ts
edleeks87: There was a lint error, I'll see what happens when it's removed

edleeks87: Yeah we don't need it not sure why I added it tbh

src/style/design-tokens/carbon-scoped-tokens-provider/carbon-scoped-tokens-provider.component.ts
edleeks87: I think I must have had an incorrect type somewhere else, I'll get it removed 👍

src/global.d.ts
edleeks87: It doesn't look like there's any type defined in that package and I can't find any in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

src/style/themes/base/base-theme.config.ts
```diff
@@ -2,15 +2,11 @@ import tokens from "@sage/design-tokens/js/base/common";
...
-
-export type BasePalette = Record<string, string | PaletteFunction> &
-  ColorsWithHex & {
-    blackOpacity: (opacity: number | string) => string;
-    whiteOpacity: (opacity: number | string) => string;
-  };
+export type BasePalette = Record<string, string | PaletteFunction> & {
+  blackOpacity: PaletteFunction;
+  whiteOpacity: PaletteFunction;
+  [key: string]: PaletteFunction;
```
edleeks87, line 8: @samtjo @grabkowski I ended up having to do something like this as it was complaining in the Theme when I was calling the palette functions that it might be a string and therefore not callable. I tried adding `as const` etc but had no luck and some of it was bit above my understanding

src/global.d.ts
samtjo: But that package is just for js prop types isn't it?

samtjo: Ah I see, we're actually using this in the `filter-out-styled-system-spacing-props` util. 

My thought was this package only contains js prop-types so it won't be used in any TypeScript files, but I see that we are. This is fine for now imo, but by the end of the ts conversion that will be the only usage of `@styled-system/prop-types` so I wonder if in the future you'll want to find another way of writing that util without it so the dependency can be removed.

samtjo (approved):

src/global.d.ts
edleeks87: Yeah I agree we should look to replace it with something else down the line

grabkowski (approved):

samtjo (dismissed):

grabkowski (dismissed):

src/style/style.spec.ts
```diff
@@ -1,20 +1,25 @@
 import config from "./color-config";
 import mix from "./utils/mix";
-import generatePalette from "./palette";
+import generatePalette, { PaletteFunction } from "./palette";
```
ZhuoyuJin, line 3: Should it be "import { generatePalette, PaletteFunction } from "./palette"; "?

edleeks87: The `generatePalette` function is a default export and the `PaletteFunction` is a named so the above syntax is correct

ZhuoyuJin (approved):

grabkowski (approved):

samtjo (dismissed):

grabkowski (approved):

samtjo (approved):

DlgSHi (approved):

PR #5067, opened 2022-04-19
feat: consume Sage UI font from @sage/design-tokens package

Parsium:
### Proposed behaviour

- Consume `Sage UI` font files directly from `@sage/design-tokens` library. This should hopefully fix the flakey loading behaviour we are experiencing with Chromatic snapshots.
- Amend Getting Started Guide to reflect this change.

### Current behaviour

- `Sage UI` font files are loaded from an external CDN. Consuming apps also have to preload the font files in their project's HTML. 

### Checklist

#### QA

### Testing instructions

Double check the fonts are loading correctly in Storybook and Chromatic.


src/style/fonts.css
```diff
@@ -1,30 +1,36 @@
...
 }
 
-@font-face{
+@font-face {
   font-family: "Sage UI";
-  src: url('https://fonts.sage.com/Sage_UI-Medium.woff2') format("woff2"), url('https://fonts.sage.com/Sage_UI-Medium.woff') format("woff");
-  font-weight: 700;
   font-style: normal;
-  font-display: swap;
+  font-weight: 400;
```
samtjo, line 11: The css file in the design-tokens package [here](https://github.com/Sage/design-tokens/blob/master/assets/fonts/sageui.css) uses font weights of 400, 500 and 700.

Why are we using different font weights to that package?

Parsium: Thanks for taking a look @samtjo. This is a misalignment between `design-tokens` and Carbon/legacy SOP apps, which @harpalsingh has flagged with DS and other product teams. Unfortunately if we were to use the 400/500/700 weights instead then our components won't look quite right. Therefore the plan is include this weight override for now in Carbon, then once we've updated the whole codebase to use typography tokens, we should be good to remove this override

jamime (commented):
This should be a `feat` commit because we're changing `src` files.

jamime (approved):

samtjo (approved):

DlgSHi (approved):

PR #5111, opened 2022-05-04
feat(textarea): modify padding to use design tokens

Parsium:
### Proposed behaviour

- Modify inner `InputPresentation` container to use design token for controlling its vertical padding.
- Add label to `Textarea` test story to fix accessibility violation 

Screenshot 2022-05-04 at 09 41 34

### Current behaviour

- The vertical padding is not correct and doesn't use design tokens.

Screenshot 2022-05-04 at 09 42 03

### Checklist

#### QA

### Additional context

(For devs) Both `InputPresentation` and `Input` have css rules for their `min-height`. `Textarea` programmatically controls its height with its internal logic (using `MIN_HEIGHT` constant in `textarea.style.js` on line 10). Since `InputPresentation` controls its own min-height using design tokens, the logic in `Textarea` becomes partially redundant, however we cannot currently modify the `MIN_HEIGHT` constant to use the design token value since we can importing its value as a css custom property and not a js variable. This does not cause an issue but does introduce some complexity to the code related to inputs, so it may be worth having a quick discussion over whether we should refactor this or not. 

### Testing instructions

- Double check violation no longer exists for `Textarea` test story.
- Double check no regressions have been introduced.

### DIFF
--- a/cypress/integration/common/textarea.feature
+++ b/cypress/integration/common/textarea.feature
@@ -40,7 +40,7 @@ Feature: Textarea component
       | 300  | rows300      |
 
   @positive
-  Scenario Outline: Set expandable and rows to <rows> 
+  Scenario Outline: Set expandable and rows to <rows>
     When I open default "Textarea Test" component with "textarea" json from "commonComponents" using "<nameOfObject>" object name
     Then rows is set to "<rows>"
       And Textarea height is "119px"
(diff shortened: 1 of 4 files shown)


samtjo (changes requested):
The code changes are fine but the first commit should be a `docs` commit, and it should reference the relevant Github issue rather than the Jira ticket. This means it will be linked to the PR and automatically closed when the PR is merged 👍🏻

samtjo (dismissed):

mkrds (dismissed):

ZhuoyuJin (dismissed):

samtjo (approved):

mkrds (approved):

ZhuoyuJin (approved):

mkrds (approved):

samtjo (approved):

PR #5117, opened 2022-05-05
feat(dialog, sidebar, menu-full-screen): set focus to dialog container on open

### LINKED ISSUE #4975: Dialog: Ensure focus position starts from the Dialog content
### Desired behaviour

When the dialog component is opened, the keyboard focus position begins from the cross close button despite the body of the dialog containing focusable elements.

### Current behaviour

When the dialog component is opened and it contains focusable elements within the body, the keyboard focus position bypasses the cross close button and begins from the first focusable element within the body.

### Suggested Solution

Adding a `tabindex="-1"` attribute to the dialog element should ensure the focus cannot return to it which would mean screen reader users start reading from the start of the content, and keyboard users would get to the first interactive control (cross close button) by pressing the tab key.

https://codesandbox.io/s/objective-butterfly-qwk6b8?file=/src/index.js

_No response_

### Confidentiality

robinzigmond:
### Proposed behaviour

When Dialog opens, place the focus by default on the dialog container, rather than on the first focusable element inside the dialog. This makes screen readers read out the full dialog content when it is first opened.

Also ensure that when tabbing through, the close button is focused before the content.

Also applies to DialogFullScreen, SideBar, and MenuFullScreen

closes #4975

### Current behaviour

Focus is placed on first focusable element within the dialog content. If here is a close button, even though that appears at the top of the dialog, it cannot be accessed in the tab order without going backwards (Shift+Tab) or tabbing through the entire content and cycling back to the start.

### Checklist

#### QA

### Testing instructions
Ensure the above-described behaviour is implemented on the following components:
- Dialog
- DialogFullScreen
- Sidebar
- MenuFullScreen
- AdvancedColorPicker
- Confirm

On all of these:
- there should be no immediately visible focus indicator on opening the dialog (or whichever component)
- when testing with a screenreader, this should make the full content of the dialog (or sidebar etc.) be announced
- pressing tab from then on goes to the first focusable element on the page (usually a close icon if this is present at the top-right corner)
- continuing to tab from there (and shift-tab likewise) should cycle through the focusable elements within the container, but never again focus the container, and also never breaking the focus trap

- Test that the above behaviour continues to work even after closing and reopening the component with no page reload, in particular make sure focus never remains outside of the component

Also test with nested dialogs to ensure the behaviour is as expected. You can test this using the following codesandbox: https://codesandbox.io/s/heuristic-jennings-26n7cn?file=/src/index.js


src/components/menu/menu-full-screen/menu-full-screen.component.js
```diff
@@ -33,13 +44,30 @@ const MenuFullscreen = ({
     }
   };
 
-  useLayoutEffect(() => {
-    const checkTransitionEnd = () => {
-      menuContentRef.current.focus();
+  useEffect(() => {
```
edleeks87, line 47: I'm a little confused by this implementation. The `useLayoutEffect` will run before the `useEffect`. So as far as I can tell this will run `setAnimationComplete(true)`, then it will `setAnimationComplete(false)` then `setAnimationComplete(true)`


I also noticed it doesn't seem to reset the state as expected so the trap only works when the menu is opened the first time etc
https://user-images.githubusercontent.com/44157880/167400283-dc0fd652-bf48-47e2-8d6a-fed061fd3507.mov

robinzigmond: Interesting, thanks - I'll look into this. The useLayoutEffect is just because this was using useLayoutEffect before, even though I've changed the internals a little - I never dived into why it needed to wait for the actual DOM to be updated (I think that's the only difference between useLayoutEffect and useEffect?).

Either way your demo clearly shows a serious bug that I'd completely missed so I will look into this!

src/hooks/__internal__/useModalFocus/useModalFocus.ts
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from "react";
```
mkrds, line 1: Is it necessary that this logic is a separate hook? It seems like it's not really useful outside of FocusTrap context? Would it be possible to contain it inside of the `focus-trap.component.js` and apply the return values of this hook to the `div ref={trapRef}>{children}</div>` element? If not, maybe you could name it in a way that makes it more descriptive? Like `useModalTabIndex` or `useFocusableModalTabIndex`?

src/components/menu/menu-full-screen/menu-full-screen.component.js
```diff
@@ -25,6 +33,9 @@ const MenuFullscreen = ({
   const menuWrapperRef = useRef();
   const menuContentRef = useRef();
   const { menuType } = useContext(MenuContext);
+  const [isAnimationComplete, setAnimationComplete] = useState(false);
```
mkrds, line 36: Do we actually need this? Is there any issue with the MenuFullScreen being already focused during the 300ms that the transition takes? Looks like a lot of unnecessary code that does not solve any real problem here

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -79,12 +79,18 @@ const FocusTrap = ({
...
+      (focusFirstElement || wrapperRef?.current)
     ) {
-      setElementFocus(focusFirstElement || firstElement);
+      setElementFocus(focusFirstElement || wrapperRef?.current);
       firstOpen.current = false;
     }
-  }, [autoFocus, firstElement, focusFirstElement, isAnimationComplete]);
+  }, [
+    autoFocus,
+    firstElement,
```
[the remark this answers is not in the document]
robinzigmond: good point, thanks - surprised ESLint didn't pick this up

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -157,10 +166,15 @@ const FocusTrap = ({
     ) {
       // the trap breaks if it tries to refocus a disabled element
       setElementFocus(currentFocusedElement);
+    } else if (
+      wrapperRef?.current &&
+      wrapperRef.current.hasAttribute("tabindex")
```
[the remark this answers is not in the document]
robinzigmond: Agreed. Actually I thought that line already existed in the code before I changed anything here but it seems that's not the case - regardless I agree it might as well be written the more modern and readable way.

src/hooks/__internal__/useModalFocus/useModalFocus.ts
robinzigmond: The separate hook was simply to avoid repetition, especially as the logic used here is far from obvious. The props returned by the hook have to go in the Dialog (or whatever particular component it is) contained in the Focus Trap rather than FocusTrap itself - but I agree it's worth looking at putting this in FocusTrap and adding the props automatically to the children, thanks for the idea. And thanks for the alternative name suggestions, I like useModalTabIndex.

src/components/menu/menu-full-screen/menu-full-screen.component.js
robinzigmond: I had a lot of problems with getting this to work on MenuFullScreen as that's implemented quite differently from the other components here. The problem is that it uses FocusTrap but not Modal, but FocusTrap sort of assumes it's inside a Modal and in particular uses ModalContext, including isAnimationComplete from that to decide whether to focus on opening or not, which is why I ended up with this sort of messy solution of just putting the ModalContext.Provider into MenuFullScreen (as I understand why it can't use Modal directly). Really what this code is about is ensuring the menu gets the automatic focus at the write time - previously this was done in a rather simple way by putting tabindex of -1 on the container of the actual menu (not including the header) and directly focusing that on open, but now we want to focus the container first and reuse the logic that's currently (in this PR) in useModalFocus it becomes much more tricky.

Ideally these two components (Modal and FocusTrap) should be properly decoupled but it's not clear if that's even possible.

I'm currently looking at all this again anyway to address the issue that Ed raised in his comments so it will likely change a lot - certainly the code is getting way more complex than I would like so I'll bear your comments in mind and see if anything can be removed! But I'm not particularly optimistic.

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -1,12 +1,40 @@
 const defaultFocusableSelectors =
   'button:not([disabled]), [href], input:not([type="hidden"]):not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]';
 
+// callback-based rather than Promise-based, as Jest fake timers unfortunately don't work nicely with Promises
```
edleeks87, line 4: I don't think we need this comment

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -1,12 +1,40 @@
...
+};
+
 function setElementFocus(element) {
   if (typeof element === "function") {
     element();
   } else {
     const el = element.current || element;
-    el.focus();
+    // sometimes (eg on MenuFullScreen) the .focus call happens before the wrapper element has CSS visibility set
+    // to visible - so we need to wait for the element to be visible otherwise the focusing fails silently
```
edleeks87, line 36: I think this second comment can be removed as there's a comment in the above about it failing silently

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -1,12 +1,40 @@
 const defaultFocusableSelectors =
   'button:not([disabled]), [href], input:not([type="hidden"]):not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]';
 
+// callback-based rather than Promise-based, as Jest fake timers unfortunately don't work nicely with Promises
+const waitForCss = (element, expectedStyles, maxTime, callback) => {
```
edleeks87, line 5: This function isn't called anywhere else so do we need to pass in these as parameters (apart from `element`) rather than setting them in the function body?

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -157,18 +165,53 @@ const FocusTrap = ({
...
     if (triggerRefocusFlag) {
       refocusTrap();
     }
   }, [triggerRefocusFlag, refocusTrap]);
 
-  return <div ref={trapRef}>{children}</div>;
+  const [tabIndex, setTabIndex] = useState(0);
+
+  useEffect(() => {
+    // ideally would set to -1, but that fails Cypress tests due to Cypress bug.
```
edleeks87, line 184: I think this could just a one line comment something like below

```
Issue in cypress prevents setting tabIndex to -1 (link to issue if there is one), instead tabIndex is removed onBlur and reset on close

File: src/__internal__/focus-trap/focus-trap.component.js [hunk 10]
```diff
@@ -157,18 +165,53 @@ const FocusTrap = ({
...
+    // It's OK to set to 0 instead as the tabindex is removed on blur anyway.
+    // We set this on close rather than open so that the element is already focusable
+    // when reopened, to avoid timing issues. This doesn't cause problems with keyboard tabbing
+    // as elements are not keyboard focusable when hidden by CSS.
+    if (!isOpen) {
+      setTabIndex(0);
+    }
+  }, [isOpen]);
+
+  // need to remove the container's tabindex after it has been used for initial focus,
```
edleeks87, line 194: I don't think these comments are needed IMHO

src/__internal__/focus-trap/focus-trap.component.js
```diff
@@ -157,18 +165,53 @@ const FocusTrap = ({
...
+  const onBlur = () => {
+    /* istanbul ignore else */
+    if (isOpen) {
+      setTabIndex(undefined);
+    }
+  };
+
+  const focusProps = { tabIndex, onBlur };
+
+  // children should always consist of only a single element, a wrapper to
```
edleeks87, line 206:
```
Passes focusProps if no tabIndex has been explicitly set on the wrapper
```

src/__internal__/focus-trap/focus-trap.spec.js
```diff
@@ -9,15 +9,15 @@ import { ModalContext } from "../../components/modal/modal.component";
 jest.useFakeTimers();
 
 // eslint-disable-next-line
-const MockComponent = ({ children, triggerRefocusFlag, ...rest }) => {
+const MockComponent = ({ children, triggerRefocusFlag, tabIndex, ...rest }) => {
   const ref = useRef();
   const [isDisabled, setIsDisabled] = useState(false);
   return (
     <ModalContext.Provider
       value={{ isAnimationComplete: true, triggerRefocusFlag }}
```
edleeks87, line 17:
```suggestion
      const MockComponent = ({ children, triggerRefocusFlag, isAnimationComplete, tabIndex, ...rest }) => {
    ...
      return (
         <ModalContext.Provider
             value={{ isAnimationComplete, triggerRefocusFlag }}
   ...
```
`isAnimationComplete` is now true by default, I think we should test it by using the default value and maybe add a test where we set it to false and assert no focus etc (it seems we've missed writing a test for that previously 🤦 )

edleeks87 (approved):

mkrds (approved):

divyajindel (approved):

PR #5127, opened 2022-05-09
fix(date): prevent calling onChange when input is blurred and value has not changed FE-5161

### LINKED ISSUE #5120: [Date] the onChange event is triggered after blur calling on component
### Current behaviour

1. `git checkout master` & `npm ci` & `npm start`
2. Open [storybook](https://carbon.sage.com/?path=/docs/date-input-test--default-story) 
3. Press `Tab` twice to focus and then blur the component
4. onChange event is called

### Expected behaviour

the onChange event should be called only after real change of the value of the input

https://carbon.sbc-ui.com/?path=/story/date-input-test--default-story

### JIRA Ticket (Sage Only)

FE-5161

### Suggested Solution

_No response_

### Carbon Version

106.1.5

Firefox, Chrome, Safari

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5120

### Proposed behaviour

Prevents `onChange` from being called when input is blurred and the value has not updated

### Current behaviour

When the initial value is different from the given locale's format it will call onChange when the input is blurred

### Checklist

#### QA

### Testing instructions

Compare the the story in linked issue with local version

### DIFF
--- a/src/components/date/__internal__/utils.js
+++ b/src/components/date/__internal__/utils.js
@@ -103,7 +103,7 @@ function findMatchWithNoSeparators(valueString, formatString) {
   const valueArray = makeSeparatedValues(indexArray, valueString);
 
   if (checkForCompleteMatch(formatArray, valueArray)) {
-    return [formatArray.join("."), valueArray.join(".")];
+    return [formatArray.join(" "), valueArray.join(" ")];
   }
 
   return null;
--- a/src/components/date/__internal__/utils.spec.js
+++ b/src/components/date/__internal__/utils.spec.js
@@ -355,18 +355,18 @@ describe("utils", () => {
   });
 
   describe("findMatchedFormatAndValue", () => {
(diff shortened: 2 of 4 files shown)


src/components/date/__internal__/utils.js
```diff
@@ -103,7 +103,7 @@ function findMatchWithNoSeparators(valueString, formatString) {
   const valueArray = makeSeparatedValues(indexArray, valueString);
 
   if (checkForCompleteMatch(formatArray, valueArray)) {
-    return [formatArray.join("."), valueArray.join(".")];
+    return [formatArray.join(" "), valueArray.join(" ")];
```
edleeks87, line 106: this was changed to prevent clashes with locales that use `.` as a separator, as the guard added to `onBlur` would not work for them

DipperTheDan (dismissed):

mkrds (dismissed):

DipperTheDan (approved):

PR #5131, opened 2022-05-09
fix(toast): fix escape key behaviour when toast launched from dialog 

robinzigmond:
### Proposed behaviour

When a Toast is launched from a Dialog (or similar component using a Modal), after pressing Escape to close the Toast, a subsequent press of the Escape key should close the Dialog.

### Current behaviour

In the situation described above, pressing Escape does nothing once the Toast has been dismissed

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/awesome-banach-tip6mx?file=/src/index.js


src/hooks/__internal__/useModalEscape/useModalEscape.ts
```diff
@@ -0,0 +1,76 @@
+import React, { useEffect, useRef, useCallback } from "react";
+import ModalManager from "../../../components/modal/__internal__/modal-manager";
+
+const useModalEscape = (
+  open: boolean,
+  closeModal: (e: Event) => void,
+  modalRef: React.RefObject<HTMLElement>,
+  setTriggerRefocusFlag: (flag: boolean) => void | undefined
```
mkrds, line 8:
```suggestion
  setTriggerRefocusFlag?: (flag: boolean) => void | undefined
```

src/hooks/__internal__/useModalEscape/useModalEscape.ts
```diff
@@ -0,0 +1,76 @@
+import React, { useEffect, useRef, useCallback } from "react";
+import ModalManager from "../../../components/modal/__internal__/modal-manager";
+
+const useModalEscape = (
+  open: boolean,
+  closeModal: (e: Event) => void,
```
mkrds, line 6:
```suggestion
  closeModal: (e: KeyboardEvent) => void,
```

mkrds:
I think that `useModalManager` would be a better name for the hook you've implemented. 
Also this bit of code `const isTopmost = ModalManager.isTopmost(ref.current);` which is used in both `toast.component` and `modal.component` could be abstracted to the `useModalManager` hook too. This way all the modal manager logic would be contained in `useModalManager`

mkrds (dismissed):

src/components/confirm/confirm.spec.js
```diff
@@ -111,17 +114,14 @@ describe("Confirm", () => {
...
-        domNode = wrapper.getDOMNode();
-        document.body.appendChild(domNode);
       });
 
       afterEach(() => {
-        document.body.removeChild(domNode);
         onCancelFn.mockReset();
       });
 
       it("should not close the modal if ESC key is pressed", () => {
```
grabkowski, line 123: I think that we could change it to be what we are actually testing: `"onCancel" callback should not be called`

grabkowski (approved):

divyajindel (approved):

PR #5163, opened 2022-05-18
fix(decimal): ensure that value prop is reliably displayed - FE-5153

### LINKED ISSUE #5103: Decimal: doesn't reliably display its props.value
### Current behaviour

If onchange handler changes the value passed back to the control, it seems to ignore it.

Typically we will truncate the number of characters after the decimal point, any subsequent characters being suppressed.
This used to work on Carbon 95.1, but not on 106.0
In Sandbox: I've mimicked behaviour: 
Type 22.22222 into the decimal. We are overriding value to remain at 22.22, but control displays the longer string, and stays that way when tabbing out.
On Carbon 95.1, it seems to toggle between 22.22 and 22.222, but would always revert to 22.22 on tabbing out.

But not an issue when setting value programmatically. 
Clicking button on sandbox does work.

### Expected behaviour

The decimal control would display the value being passed into its props.value

https://codesandbox.io/s/pensive-easley-tz9lsf

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

106.0

Chrome

Windows

_No response_

### Confidentiality

DipperTheDan:
The decimal component should reliably show the correct value when the component is controlled. Due
to how the controlled functionality is managed internally, ensuring that `stateValue` is included
the `useEffect` that manages the controlled value is enough to ensure this happens.

fixes #5103

### Proposed behaviour

- Decimal component should be reliably displaying its `value` prop.

decimal_value_reliable

### Current behaviour
 
- In certain conditions, Decimal component is not reliably displaying its `value` prop.

decimal_value_not_reliable

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

Codesandbox to test this PR is this one: [modest-dust-omhv0j](https://codesandbox.io/s/modest-dust-omhv0j)

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -20,7 +20,7 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        # run 20 copies of the current job in parallel
+        # run 15 copies of the current job in parallel
         containers:
           [
             1,
@@ -38,11 +38,6 @@ jobs:
             13,
             14,
             15,
-            16,
-            17,
-            18,
-            19,
-            20,
           ]
     container: cypress/browsers:node14.17.6-slim-chrome100-ff99-edge
     steps:
@@ -117,10 +112,11 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        # run 1 copy of the current job in parallel
+        # run 2 copy of the current job in parallel
         containers:
           [
-            1
+            1,
+            2,
           ]
     container: cypress/browsers:node14.17.6-slim-chrome100-ff99-edge
     steps:
@@ -188,7 +184,7 @@ jobs:
       # https://github.com/cypress-io/github-action/issues/48
       fail-fast: false
       matrix:
-        # run 6 copies of the current job in parallel
+        # run 10 copies of the current job in parallel
         containers:
           [
             1,
@@ -196,7 +192,11 @@ jobs:
             3,
             4,
             5,
-            6
+            6,
+            7,
+            8,
+            9,
+            10,
           ]
     container: cypress/browsers:node14.17.6-slim-chrome100-ff99-edge
     steps:
--- a/cypress/integration/accessibility/a11y-eighth-part.test.js
+++ b/cypress/integration/accessibility/a11y-eighth-part.test.js
@@ -1,3 +1,3 @@
 import tests from "../../support/accessibility/a11y-utils";
 
-tests(376, 450);
+tests(526, 600);
--- a/cypress/integration/accessibility/a11y-eleventh-part.test.js
+++ b/cypress/integration/accessibility/a11y-eleventh-part.test.js
@@ -1,3 +1,3 @@
 import tests from "../../support/accessibility/a11y-utils";
 
-tests(376, 450);
+tests(751, 825);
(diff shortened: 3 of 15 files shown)


src/components/decimal/decimal.component.js
```diff
@@ -227,17 +227,24 @@ const Decimal = ({
...
+        setStateValue(formatValue(valueProp));
+
+        if (valueProp === "" && prevValue === "") {
+          setStateValue(formatValue(emptyValue));
+        }
       }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [value]);
+  }, [value, componentShouldUpdate]);
```
edleeks87, line 247: I think we should remove the disable lint and add all the dependencies here, or at least replace `componentShouldUpdate` with `stateValue`

```
  }, [value, stateValue, emptyValue, formatValue, getSafeValueProp, isControlled, prevValue, toStandardDecimal]);
```

src/components/decimal/decimal.component.js
```diff
@@ -227,17 +227,24 @@ const Decimal = ({
...
+  const componentShouldUpdate = stateValue === value;
+
   useEffect(() => {
     const unformattedValue = toStandardDecimal(stateValue);
 
     if (isControlled) {
       const valueProp = getSafeValueProp(value);
       if (unformattedValue !== valueProp) {
-        setStateValue(formatValue(value));
+        setStateValue(formatValue(valueProp));
```
edleeks87, line 239:
```suggestion
      if (valueProp === "" && prevValue === "") {
          setStateValue(formatValue(emptyValue));
        } else {
          setStateValue(formatValue(valueProp));
        }
```

So we're not potentially setting state twice 👍

src/components/decimal/decimal.component.js
```diff
@@ -227,17 +227,24 @@ const Decimal = ({
     prevControlledRef.current = isControlled;
   }, [isControlled]);
 
+  const prevValue = usePrevious(value);
+  const componentShouldUpdate = stateValue === value;
+
   useEffect(() => {
     const unformattedValue = toStandardDecimal(stateValue);
```
edleeks87, line 234: IMHO the naming here could be a little clearer, as we discussed on slack something like `standardDecimalValue` might be more intuitive but I'm also okay with you keeping it as is if you disagree

src/components/decimal/decimal.component.js
edleeks87: It will likely lead to the test breaking that we discussed on slack, lets catch up tomorrow and we can look at updating it as I'm not convinced it's testing controlled behaviour like we want

edleeks87 (approved):

src/components/decimal/decimal-test.stories.mdx
```diff
@@ -119,6 +119,29 @@ export const NewValidationStory = (args) => {
...
+  const [state, setState] = useState("0.01");
+  const handleChange = (e) => {
+    let newValue = e.target.value.rawValue;
+    if (newValue.startsWith("22.22")) newValue = "22.22";
+    action("onChange")(e, newValue);
+    setState(newValue);
+  };
+  return (
+    <div>
+      If you try and type "22.222", the onChange should block the last "2" from being entered and you should see "22.22" in the textbox.
```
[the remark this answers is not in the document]
DipperTheDan: Good spot. I'll get this amended 👍🏻

robinzigmond (approved):

edleeks87 (approved):

PR #5180, opened 2022-05-27
test(date): refactor for cypress tests

DlgSHi:
### Proposed behaviour

Refactor for `Date` component to use new cypress-component-react framework for testing.

### Current behaviour

### Checklist

#### QA

### Testing instructions

 - Check if the `date.test.js` file passed


src/components/date/date.test.js
```diff
@@ -0,0 +1,545 @@
...
+
+    it("should open dayPicker after click on input", () => {
+      CypressMountWithProviders(<DateInputCustom />);
+
+      dateInputParent().click();
+
+      dayPickerWrapper().should("be.visible");
+    });
+
+    it("should not close dayPicker after doble click on input", () => {
```
edleeks87, line 173:
```suggestion
    it("should not close dayPicker after double click on input", () => {
```

src/components/date/date.test.js
```diff
@@ -0,0 +1,545 @@
...
+
+    it("should open dayPicker after click on icon", () => {
+      CypressMountWithProviders(<DateInputCustom />);
+
+      dateIcon().click();
+
+      dayPickerWrapper().should("be.visible");
+    });
+
+    it("should close dayPicker after click on icon", () => {
```
edleeks87, line 189:
```suggestion
    it("should close dayPicker after click on icon and dayPicker is open", () => {
```

src/components/date/date.test.js
```diff
@@ -0,0 +1,545 @@
...
+        .invoke("height")
+        .should("be.greaterThan", minValue)
+        .and("be.lessThan", maxValue);
+    });
+  });
+
+  it("should check the allowEmptyValue prop", () => {
+    CypressMountWithProviders(<DateInputCustom allowEmptyValue />);
+
+    dateInput().clear().blur().should("have.attr", "value").and("be.empty");
```
edleeks87, line 275: we should check that the `rawValue` emitted onBlur is `""` and not `null` here

src/components/date/date.test.js
```diff
@@ -0,0 +1,545 @@
...
+      <Confirm open height="60px">
+        <DateInputCustom disablePortal={state} />
+      </Confirm>
+    );
+
+    dateInput().click();
+    dayPickerWrapper().should(visibility);
+  });
+
+  describe("check events for TextEditor component", () => {
```
edleeks87, line 319:
```suggestion
  describe("check events for Date component", () => {
```

src/components/date/date.test.js
```diff
@@ -0,0 +1,545 @@
...
+
+        dateInput().clear().type(getDatetoEnter(), { delay: 0 });
+        dateInput().click();
+        dayPickerByText("12").click();
+
+        dateInput()
+          .wait(250)
+          .should("have.attr", "value", formattedValueParam)
+          .then(() => {
+            expect(callback.getCalls()[12].args[0].rawValue).to.equals(
```
edleeks87, line 403: Instead of using a magic number to get the last call on the stack, we could use a util function like so

```suggestion
           const getLastCall = (calls) => calls.pop();
...
            expect(getLastCall(callback.getCalls()).args[0].rawValue).to.equals(
```

DlgSHi: it isn't a last call always :)

stephenogorman (approved):

src/components/date/date.test.js
edleeks87: They are as it's the last call you're testing, even `callback.getCalls()[0]` is the last call on the stack is you try it out

edleeks87: Happy to approve this now and we can add the above as a util so it can be shared in other test files at a later date

stephenogorman (approved):

PR #5202, opened 2022-06-06
refactor(label): convert files to typescript FE-4861

edleeks87:
### Proposed behaviour

Convert `Label` component files to TS

### Current behaviour

`Label` component is JS

### Checklist

#### QA

### Testing instructions

Test for regressions in inputs that use label

https://codesandbox.io/s/carbon-quickstart-typescript-forked-7xpo14?file=/src/index.tsx
Will likely need to export the project locally and confirm no types are broken


src/__internal__/label/label.component.tsx
```diff
@@ -1,25 +1,73 @@
...
 
-const tooltipPosition = ({ error, warning, info, inline }) => {
-  return (error || warning || info) && inline === true ? "top" : "right";
+const tooltipPosition = ({
+  error,
+  warning,
+  info,
+  inline,
+}: ValidationProps & { inline?: boolean }) => {
+  return (error || warning || info) && inline ? "top" : "right";
```
grabkowski, line 63: I think that the return type should be set as `TooltipPositions` type imported from `tooltip.config.ts`

src/__internal__/label/label.style.ts
```diff
@@ -1,7 +1,7 @@
 import styled, { css } from "styled-components";
-import PropTypes from "prop-types";
+import { LabelProps } from "./label.component";
 
-const LabelStyle = styled.label`
+const LabelStyle = styled.label<Pick<LabelProps, "isRequired" | "disabled">>`
```
grabkowski, line 4: I think that now will be the best time change `LabelStyle` to `StyledLabel` to match our naming convention, what do you think about that?

grabkowski, line 4: what do you think about defining `StyledLabelProps` here and then extending the component interface with it?

src/__internal__/label/label.style.ts
```diff
@@ -24,12 +24,11 @@ const LabelStyle = styled.label`
...
-LabelStyle.propTypes = {
-  disabled: PropTypes.bool,
-  isRequired: PropTypes.bool,
-};
+const DEFAULT_CONTAINER_WIDTH = 30;
 
-export const StyledLabelContainer = styled.div`
+export const StyledLabelContainer = styled.div<
+  Pick<LabelProps, "align" | "inline" | "pr" | "pl" | "width" | "optional">
+>`
```
grabkowski, line 31: what do you think about defining `StyledLabelContainerProps` here and then extending the component interface with it?

src/__internal__/label/label.style.ts
edleeks87: I prefer the approach you're suggest but I'd like this approach to be adopted in the other components we're refactoring as well just so the codebase is consistent etc. I'll raise this with the rest of the team in slack and make the changes if we get agreement

src/__internal__/label/label.style.ts
edleeks87: See above

DipperTheDan (commented):
Nothing more to add from me. Just for you to address the comments from @grabkowski

DipperTheDan (dismissed):

grabkowski (approved):

divyajindel (dismissed):

DipperTheDan (approved):

divyajindel (approved):

PR #5205, opened 2022-06-07
refactor(field-help): convert files to typescript

edleeks87:
### Proposed behaviour

Converts files for `FieldHelp` to TypeScript
### Current behaviour

Files for `FieldHelp` are JavaScript

### Checklist

#### QA

### Testing instructions

Linked sandbox and regression test of inputs

https://codesandbox.io/s/carbon-quickstart-typescript-forked-kitc96?file=/src/index.tsx -- some inputs that use FieldHelp, you may need to export and run locally


src/__internal__/field-help/field-help.spec.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
 import { mount } from "enzyme";
-import FieldHelp from "./field-help.component";
+import FieldHelp, { FieldHelpProps } from "./field-help.component";
 import { assertStyleMatch } from "../../__spec_helper__/test-utils";
 
-function render(props) {
+function render(props?: FieldHelpProps) {
```
Parsium, line 6: I appreciate this was not part of this ticket, but I noticed on line 29 we have a `inputWidth` prop which doesn't exist in `FieldHelpProps`. Looking at the prop definitions I think `inputWidth` needs to be `labelWidth`. Would you be able to add this? This may allow you to workaround the dynamic prop definition in `FieldHelpProps` if this was the reason behind it :)

src/__internal__/field-help/field-help.component.tsx
```diff
@@ -0,0 +1,34 @@
...
+  children?: React.ReactNode;
+  /** When true, label is placed in line an input */
+  labelInline?: boolean;
+  /** Width of a label in percentage. Works only when labelInline is true */
+  labelWidth?: number;
+  /** The unique id of the FieldHelp component */
+  id?: string;
+  /** use of any to support rest props */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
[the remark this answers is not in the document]
edleeks87: The comment above explains why any was used

src/__internal__/field-help/field-help.spec.tsx
edleeks87: I was a little confused by it as it's asserting the default value for the width and if I update it the asserted styles will be incorrect. I suspect it's a test that was left after some refactor work but I'll leave it in as it asserts the default and confirms the rest props are working as expected

edleeks87: I've removed the rest interface so have updated the test, one for default width the other for a custom value

Parsium (approved):

divyajindel (approved):

PR #5222, opened 2022-06-13
refactor(help): refactor files to typescript FE-4796

edleeks87:
### Proposed behaviour

Refactors `Help` to be TypeScript

### Current behaviour

`Help` is JavaScript

### Checklist

#### QA

### Testing instructions

Regression test in storybook and use linked sandbox

https://codesandbox.io/s/jovial-cherry-z0uek6?file=/src/index.tsx -- export locally


src/components/help/help.spec.tsx
```diff
@@ -1,17 +1,22 @@
...
-import Help from "./help.component";
+import Help, { HelpProps } from "./help.component";
 import { rootTagTest } from "../../__internal__/utils/helpers/tags/tags-specs";
 import StyledHelp from "./help.style";
 import Tooltip from "../tooltip";
 import { testStyledSystemMargin } from "../../__spec_helper__/test-utils";
 
 jest.mock("@tippyjs/react/headless");
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Parsium, line 13: What do you think about using `mount` only here instead of allowing both it and `shallow`? This would eliminate the need to suppress the `any` warning

src/components/help/help.component.tsx
```diff
@@ -9,54 +10,102 @@ import Events from "../../__internal__/utils/helpers/events";
...
+  type = "help",
   ...rest
-}) => {
+}: HelpProps): JSX.Element => {
   const defaultTooltipId = useRef(guid());
-  const helpElement = useRef(null);
+  const helpElement = useRef<HTMLDivElement>(null);
   const [isTooltipVisible, updateTooltipVisible] = useState(false);
   const { helpAriaLabel } = useContext(TooltipContext);
+  const flipBehaviourCheck = useMemo(
```
Parsium, line 74: Would the typescript compiler be able to handle this check for us? Or is it possible for the values in `tooltipFlipOverrides` to change at runtime?

edleeks87: If the consuming project is TypeScript based then we wouldn't need the check but if they are using JS we will need to have a way of constraining the values passed in to `tooltipFlipOverrides`

src/components/help/help.spec.tsx
edleeks87: My personal preference is to leave it as is for now as there's about half the cases that are using shallow which offers a degree of optimisation when running the test file and we'll also be migrating to RTL anyway so the less changes the better IMHO. However, I'm happy to make the change if the second reviewer also agrees with you 👍

grabkowski: alternatively you could just return the JSX element and use `shallow` and `mount` directly in tests. In my opinion we shouldn't set the renderer in that method in the first place:
`shallow(renderHelp({ type: "home", tooltipFlipOverrides: ["top", "bottom"] }));

src/components/help/help.component.tsx
grabkowski: I think that we could remove this check, as this will be already checked in the Icon component to which this prop is passed

src/components/help/help.spec.tsx
```diff
@@ -223,35 +228,36 @@ describe("Help", () => {
...
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore:next-line
           { type: "home", tooltipFlipOverrides: ["foo", "bar"] },
           mount
         );
-      }).toThrow();
+      }).toThrow(errorMessage);
 
       mockGlobal.mockReset();
     });
```
grabkowski, line 261: I think that this will be already tested in the Icon component

src/components/help/help.component.tsx
edleeks87: Valid point, I'll block this ticket until your refactor is merged 👍

grabkowski (approved):

grabkowski (dismissed):

Parsium (approved):

grabkowski (approved):

PR #5235, opened 2022-06-16
fix(multi-action-button): add new keyboard functionality to component FE-5188

### LINKED ISSUE #4522: Multi Action Button can't close with Esc key
### Current behaviour

The `MultiActionButton` cannot close when the Esc key is used.

### Expected behaviour

The `MultiActionButton` should be able to close when the Esc key is used.

https://codesandbox.io/s/autumn-wave-wg9gc?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

95.1.0

Chrome, Firefox, Edge, Safari

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4522

### Proposed behaviour

Updates the keyboard navigation behaviour for `MultiActionButton` and `SplitButton`. Up and down arrow keys can still be used for navigating children buttons but there is no longer any looping back around. When the children container is open tabbing can now be used to navigate the children buttons, the tab key takes you down the list and back tab goes up. When the last child button is focused and tab is pressed it will close the container and move focus to the next focusable element on the page. When the shift and tab keys are pressed and the first child button is focused it will close the container and return focus back to the main button control. Pressing the escape key when the children container is open will now close the list and return focus back to the main button control. All other keyboard functionality remains the same

### Current behaviour

Up and down arrow keys navigate children buttons and loop back around. Tabbing moves focus away from the children buttons to the next element on the page, back tabbing moves focus back to main button control

### Checklist

#### QA

### Testing instructions

Can be tested in storybook


src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -104,38 +104,59 @@ export const MultiActionButton = ({
...
+            document.querySelectorAll(
+              defaultFocusableSelectors
+            ) as NodeListOf<HTMLElement>
+          ).filter((el) => Number(el.tabIndex) !== -1);
+
+          const indexOf = elements.indexOf(
+            buttonRef.current as HTMLButtonElement
+          );
+
+          elements[indexOf]?.focus();
```
robinzigmond, line 143: I realise this wasn't your code and existed before, but is all this really necessary? We're finding the index of buttonRef.current in the elements array, then grabbing the element at that index, and focusing it. In other words, we're focusing the main button which should be accessible at buttonRef.current.

At the risk of sounding stupid, why can't we do this with just buttonRef.current?.focus ?? (I'm not even sure the ?. is needed there but it does no harm and prevents crashes in potential weird situations.)

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -198,29 +217,29 @@ export const MultiActionButton = ({
...
+      ev.preventDefault();
+      setShowAdditionalButtons(false);
     }
   };
 
-  const focusMainButton = () => {
+  const handleFocusMainButton = () => {
     isMainButtonFocused.current = true;
     if (isFocusedAfterClosing.current) {
       isFocusedAfterClosing.current = false;
```
robinzigmond, line 229: as far as I can tell, the isFocusedAfterClosing ref is now no longer needed at all - its only purpose seemed to be here, to avoid showing the buttons if it was true. And we're now not showing the buttons in any case. I think this code can be cleaned up by removing the ref entirely.

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -198,29 +217,29 @@ export const MultiActionButton = ({
...
         additionalButtons.current[0]?.current?.focus();
       }, 0);
+    } else if (Events.isEscKey(ev)) {
+      ev.preventDefault();
+      setShowAdditionalButtons(false);
     }
   };
 
-  const focusMainButton = () => {
+  const handleFocusMainButton = () => {
```
robinzigmond, line 226: Again not something you've added, and this isn't as clear-cut as the isMainButtonFocused ref (see next comment) - but it makes me uneasy that we are using a ref to tell us if the main button is focused or not. We can tell this whenever we need it by just grabbing document.activeElement and comparing it to the ref we've got for the main button. Might still be better to have the ref if we needed this check in a bunch of places - but it looks to me like it's only once, on line 59 inside hideButtons (and I'm not 100% sure we even need that any more as we no longer want the menu to always open on focus). And it would avoid lots of boilerplate like this handleFocus/handleBlur functions.

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -76,91 +78,140 @@ describe("MultiActionButton", () => {
...
-        simulateBlur(mainButton);
-        wrapper.find(StyledMultiActionButton).simulate("mouseleave");
-        wrapper.update();
-        mainButton.simulate("keydown", { key });
-        jest.runAllTimers();
+    describe("using keyboard to open children container", () => {
+      it.each([
+        ["Enter", "Enter"],
+        ["Space", " "],
+        ["ArrowDown", "ArrowDown"],
```
robinzigmond, line 85: we should also test ArrowUp here, as per the requirements (see my comment on the whole PR)

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -76,91 +78,140 @@ describe("MultiActionButton", () => {
...
+        }
+      );
 
       it("does not open additional buttons if opened already - coverage", () => {
-        mainButton.simulate("keydown", { key });
+        mainButton.simulate("keydown", { key: "Enter" });
         jest.runAllTimers();
       });
+
+      it("does not call setTimeout if key is not Enter, Space or ArrowDown", () => {
```
robinzigmond, line 109: are we testing setTimeout or preventDefault here? The title says setTimeout but the test is actually testing preventDefault.

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -76,91 +78,140 @@ describe("MultiActionButton", () => {
...
+        const preventDefault = jest.fn();
+        mainButton.simulate("keydown", { key: "1", preventDefault });
+
+        expect(preventDefault).not.toHaveBeenCalled();
+      });
     });
 
     describe('when "down" key is pressed', () => {
-      it("the additional buttons should be stepped through in sequence", () => {
+      it("the additional buttons should be stepped through in sequence but should not loop", () => {
```
robinzigmond, line 118: personal preference I guess but I'd make the "should not loop" test into a separate test (ie separate it(...) block)

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -76,91 +78,140 @@ describe("MultiActionButton", () => {
...
+        // focus should not loop back to start of button list
+        expect(additionalButtons.last().getDOMNode()).toStrictEqual(
           document.activeElement
         );
       });
     });
 
     describe('when "up" key is pressed', () => {
-      it("the additional buttons should be stepped through in sequence", () => {
+      it("the additional buttons should be stepped through in sequence but should not loop", () => {
```
robinzigmond, line 143: as above

robinzigmond (commented):
A couple of general points:

Having done a quick test, I think there's a few requirements given on FE-5188 that this PR doesn't meet:

- when focusing on the parent button and pressing the up arrow, the menu doesn't open
- holding command while pressing up or down arrow only moves up/down once, when according to the requirements it should move focus all the way to the first/last menu item. (Control+up/down should do the same on Windows, as should the Home/End buttons - might need some investigation into how presses on those buttons are sent to the browser.)

Also, as a general point: there's so much code here replicated between Multi Action Button and Split Button, can't we DRY it up a bit, at least with some utils perhaps to handle the keyboard functionality? I recognise it was like this before, but I'd have been very tempted to do this if I was rewriting it anyway as you are. Will leave it to your better judgement!

Needless to say, most/all of my many comments on the MultiActionButton changes also apply to SplitButton

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -104,38 +104,59 @@ export const MultiActionButton = ({
...
         ev.preventDefault();
-      }
+      };
 
-      if (Events.isDownKey(ev)) {
-        nextIndex =
-          currentIndex < buttonChildren.length - 1 ? currentIndex + 1 : 0;
-        ev.preventDefault();
+      if (Events.isUpKey(ev) && currentIndex > 0) {
+        nextIndex = currentIndex - 1;
```
mkrds, line 114:
```suggestion
        nextIndex = currentIndex - 1;
        ev.preventDefault();
```

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -104,38 +104,59 @@ export const MultiActionButton = ({
...
-          ) as NodeListOf<HTMLElement>
-        ).filter((el) => Number(el.tabIndex) !== -1);
-
-        const indexOf = elements.indexOf(
-          buttonRef.current as HTMLButtonElement
-        );
-
-        elements[indexOf]?.focus();
+      if (Events.isDownKey(ev) && currentIndex < buttonChildren.length - 1) {
+        nextIndex = currentIndex + 1;
```
mkrds, line 118:
```suggestion
        nextIndex = currentIndex + 1;
        ev.preventDefault();
```

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -104,38 +104,59 @@ export const MultiActionButton = ({
       );
       let nextIndex = -1;
```
mkrds, line 106: It looks like you can remove all `ev.preventDefault()` from `handleKeyDown` and add just one at the top of this method because it seems that we always want to prevent default browser behavior when using keyboard on these buttons.
Also this will fix the issue with the screen scrolling when using the Up and Down arrows.

src/components/multi-action-button/multi-action-button.component.tsx
mkrds: It seems that according to the new requirements it should be `elements[indexOf + 1]?.focus();`

robinzigmond (approved):

src/components/multi-action-button/multi-action-button.test.js
```diff
@@ -174,8 +174,90 @@ context("Tests for MultiActionButton component", () => {
...
+            .eq(0)
+            .trigger("keydown", keyCode("downarrow"));
+          multiActionButtonList().eq(2).should("be.focused");
+        });
+    });
+  });
+
   describe("pressing tab while MultiActionButton is open", () => {
-    it("should focus second MultiActionButton", () => {
+    it("should move focus to previous child button and to main button when start of list reached", () => {
```
edleeks87, line 260: This should be `it("should focus second MultiActionButton", () => {`

mkrds (dismissed):

robinzigmond (approved):

robinzigmond (approved):

DlgSHi (approved):

DlgSHi (approved):

mkrds (approved):

PR #5239, opened 2022-06-17
Allow focus to reach a Toast (or other element) from within a dialog, and fix various problems with radio buttons in a FocusTrap

### LINKED ISSUE #5092: [Dialog | Toast] - possibility of focus management
### Desired behaviour

Hey! I'd love to have possibility to manage focus for the Dialog and Toast. When Toast appears to the screen then focus should target the Toast while Dialog is open.

### Current behaviour

For now when we use Dialog component and then Toast will appear with an error message, then we are able to close the Toast with ESC key but there is no focus on the Toast

### Suggested Solution

_No response_

https://codesandbox.io/s/zealous-flower-m797mw?file=/src/index.js

_No response_

### Confidentiality

### LINKED ISSUE #5224: Various problems handling keyboard focus of Radio Buttons inside a Focus Trap (Dialog etc.)
### Current behaviour

Open the dialog, and tab through. Notice that after reaching the radio buttons at the end, focus leaves the dialog. (This only happens when the first radio button is selected, not the second.)

### Expected behaviour

The focus should remain within the dialog and go back to the close button at the start, no matter which radio button is selected.

https://codesandbox.io/s/agitated-sara-b086su?file=/src/index.js

### JIRA Ticket (Sage Only)

FE-5156

[not directly related but will be fixed as part of this ticket as this ticket will require refactoring this functionality anyway]

### Suggested Solution

Note: this is a repeat of [a similar issue](https://github.com/Sage/carbon/issues/3623) that was raised regarding a Radio Button group as the first element in a Dialog. Although this was [fixed](https://github.com/Sage/carbon/pull/3715) this was only done for using shift-tab for moving back from a radio group that was first in the Focus Trap, not for the similar situation of using Tab without Shift from a radio group that's the last focusable element in a Trap.

This wasn't likely to come up previously as the Dialog's close button was always last in the focus order, now that has been moved to earlier in the DOM as a result of [these changes](https://github.com/Sage/carbon/pull/5117) then it is possible to produce a bug. In my opinion this should have been fixed more robustly by checking for radio buttons with "plain tab" as well as "shift tab".

### Carbon Version

106.7.0

### Design Tokens Version

2.21.0

Chrome

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #5295: Back tabbing breaks focus trap
### Current behaviour

When back tabbing the FocusTrap does not catch when the first focusable element is focused and allows natural tabbing order rather than looping it to the end of the elements

### Expected behaviour

Focus should loop to last focusable element when back tabbing from first element

https://carbon.sage.com/?path=/story/dialog--default-story

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

latest

### Design Tokens Version

_No response_

MacOS

_No response_

### Confidentiality

robinzigmond:
fixes #5224
fixes #5295
fixes #5092 

### Proposed behaviour

- fixes to various issues with Radio buttons inside a component using Focus Trap
- allow passing a prop `focusableContainers` to Dialog/DialogFullScreen/Sidebar to enable other elements to be included in the the main element's trapped focus order
- make Toast autofocus the close button, if one exists, on open, and add a disableAutoFocus prop to disable this behaviour
- when Toast autofocuses in this way, make focus return to the previously-focused element on close

### Current behaviour

- there are a number of bugs with the behaviour of radio buttons inside a dialog (or other focus-trap component) in certain situations: see https://github.com/Sage/carbon/issues/5224
- when a Toast is opened from within a dialog, it is not possible to reach the dialog's close button with the keyboard

### Checklist

#### QA

### Testing instructions

For the radio button issues, these are demonstrated in these 3 sandboxes, and all should be fixed by this PR:
https://codesandbox.io/s/agitated-sara-b086su?file=/src/index.js (focus escaping dialog when radio buttons are the last element)
https://codesandbox.io/s/fervent-morning-8voljy?file=/src/index.js (tabbing skipping radio button groups when there are 2 next to each other)
https://codesandbox.io/s/wizardly-leftpad-9uxsw2?file=/src/index.js (focus going to the wrong button in a group when one has been previously selected)

For the toast focus, the original issue is demoed here https://codesandbox.io/s/zealous-flower-m797mw?file=/src/index.js - this is fixed in the PR but only if some additional work is done by the consumer, this is demoed in this sandbox which should work as intended after the PR https://codesandbox.io/s/hopeful-tu-730tmr


src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -40,27 +40,67 @@ const isRadio = (element) => {
   );
 };
 
-const nextNonRadioElementIndex = (element, focusableElements) => {
+const getRadioElementToFocus = (groupName, shifKey) => {
```
DipperTheDan, line 43:
```suggestion
const getRadioElementToFocus = (groupName, shiftKey) => {
```

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -40,27 +40,67 @@ const isRadio = (element) => {
...
+  const buttonsInGroup = document.querySelectorAll(
+    `input[type="radio"][name="${groupName}"]`
+  );
+  const selectedButton = [...buttonsInGroup].find((button) => button.checked);
+
+  if (selectedButton) {
+    return selectedButton;
+  }
+  return buttonsInGroup[shifKey ? buttonsInGroup.length - 1 : 0];
```
DipperTheDan, line 52:
```suggestion
  return buttonsInGroup[shiftKey ? buttonsInGroup.length - 1 : 0];
```

DipperTheDan (commented):
It might also be a good idea to speak with @DlgSHi to see if any new Cypress tests need adding to test any new functionality.

src/components/toast/toast.stories.mdx
```diff
@@ -440,6 +440,60 @@ Toast variant is `success` by default
...
+
+<Canvas>
+  <Story name="dismissible without autoFocus" parameters={{ info: { disable: true } }}>
+    {() => {
+      const onDismissClick = (evt) => {
+        if (!isOpen) {
+          window.scrollTo(0, 0);
+        }
+        setIsOpen(!isOpen);
+        action("click")(evt);
```
edleeks87, line 453: I don't think we need the `action` calls in the mdx stories

src/components/toast/toast.stories.mdx
```diff
@@ -440,6 +440,60 @@ Toast variant is `success` by default
...
+      };
+      const [isOpen, setIsOpen] = useState(false);
```
edleeks87, line 455: I think this would be a good story to let chromatic capture so we can confirm the close icon isn't autoFocused, I'd update this to be

```
import isChromatic from "relative path/isChromatic";
export const isOpenForChromatic = isChromatic();
...
      const [isOpen, setIsOpen] = useState(isOpenForChromatic);
```

src/__internal__/focus-trap/focus-trap-utils.js
```diff
@@ -40,27 +40,67 @@ const isRadio = (element) => {
...
+    if (nextIndex < 0) {
+      nextIndex += focusableElements.length;
+    }
+    if (nextIndex >= focusableElements.length) {
+      nextIndex -= focusableElements.length;
+    }
+
+    const nextElement = focusableElements[nextIndex];
+
+    if (nextElement === element) {
```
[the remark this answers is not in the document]
robinzigmond: hi Ed, having looked at this I don't think there's any real benefit to adding this.

If focusableElements.length is 1, then the simple nextIndex calculations above will ensure that nextElement is the same as element, so it'll full into the guard in the while loop. And while that's is doing a little bit more, it's only adding and subtracting one, and indexing into a 1-element array, and that inside a loop that will only run once in this case, so I honestly don't see any performance impact. This strikes me as a micro-optimisation with no real benefit and which will make the code more confusing. We can always revisit in the unlikely event that focus traps with a single focusable element (which seems unlikely in practice anyway) are found to cause a performance issue.

robinzigmond:
> It might also be a good idea to speak with @DlgSHi to see if any new Cypress tests need adding to test any new functionality.

Good point - unfortunately Igor is away for a couple of weeks but I think it would be a good idea to add some new Cypress tests, it can't do any harm.

DipperTheDan (approved):

src/__internal__/focus-trap/focus-trap-utils.js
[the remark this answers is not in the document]
edleeks87: I'm happy to leave it as is as it's likely to be a rare use case and there is never a case that I can think of when we would have multiple traps running at once (even if there is there would need to be a lot) so the cost isn't too high when the above scenario happens. I commented the above as we actually provide Modals that would match 1 focusable element, for example https://carbon.sage.com/?path=/story/confirm--single-action and with current implementation it would be doing checks and assignments it doesn't need to do.

edleeks87 (dismissed):

edleeks87:
This PR also fixes #5295

DlgSHi (dismissed):

DlgSHi (approved):

edleeks87 (approved):

PR #5241, opened 2022-06-20
fix(heading): fix axe violations (FE-4979)

grabkowski:
### Proposed behaviour

Add default `aria-label` to the "back button" in Heading component.
Add `aria-balel`when Help is a link.

### Current behaviour

"Ensures links have discernible text" Axe violation in "with help link" story

### Checklist

#### QA

### Testing instructions


src/components/heading/heading.spec.js
```diff
@@ -22,63 +22,105 @@ import DefaultPages from "../pages/pages.component";
...
+
+  describe("when the subheader prop is provided", () => {
+    const subheader = "subheader text";
+    let wrapper;
+
+    beforeEach(() => {
+      wrapper = mount(<Heading title="foo" subheader={subheader} />);
+    });
+
+    it("renders a subheader with text the same as prop content", () => {
```
robinzigmond, line 111: I don't think this test name reads very well. The prop is called `subheader`, not `content`. Maybe something like `"renders a subheader with text given by the subheader prop"`

src/components/heading/heading.spec.js
```diff
@@ -202,11 +231,11 @@ describe("Heading", () => {
   describe("when the backLink is a function", () => {
     it("sets it as the link onClick prop", () => {
       const backLinkSpy = jasmine.createSpy(),
-        wrapper = shallow(<Heading title="Test" backLink={backLinkSpy} />),
-        link = wrapper.find('[data-element="back"]');
+        wrapper = mount(<Heading title="Test" backLink={backLinkSpy} />),
+        link = wrapper.find('[data-element="back"]').at(0);
 
       expect(link.props().onClick.toBeDefined);
```
robinzigmond, line 237: I realise you didn't add this line, but since you seem to be fixing up tests anyway...

looks like `toBeDefined` is a function, so we need to call it: `.toBeDefined()`.

src/components/heading/heading.spec.js
```diff
@@ -252,7 +281,7 @@ describe("Heading", () => {
         />
       );
 
-      elementsTagTest(wrapper, ["back", "help", "subtitle", "title", "pills"]);
+      elementsTagTest(wrapper, ["help", "subtitle", "title", "pills"]);
```
robinzigmond, line 284: I'm probably missing something but I don't see why you've removed "back" from the `data-element` properties we're testing. It looks to me like that should still be there after your changes? If the test fails for some reason with that in I'd rather understand why and fix it up, than just remove it.

src/components/heading/heading.component.js
```diff
@@ -55,14 +56,19 @@ const Heading = ({
...
     return (
-      <StyledHeadingBackButton
-        // this event allows an element to be focusable on click event on IE
-        onMouseDown={(e) => e.currentTarget.focus()}
-        data-element="back"
-        {...backButtonProps}
-      >
-        <StyledHeadingIcon type="chevron_left" divider={divider} />
-      </StyledHeadingBackButton>
+      <LocaleContext.Consumer>
```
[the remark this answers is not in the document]
grabkowski: I did this one before that component was refactored to functional, so there was no way to use a hook. Thank you for spotting that, I'll change it to hook.

src/components/heading/heading.spec.js
grabkowski: I'm already using that tag on line 235 to find the button, so looking for it again seems redundant

DipperTheDan (commented):
Nothing more to add from me. @robinzigmond seems to have done a good review 😄

DipperTheDan (approved):

divyajindel (approved):

PR #5255, opened 2022-06-22
fix(use-scroll-block): remove right padding added to the body

### LINKED ISSUE #5220: [Modal] - Right padding is added for no reason
### Current behaviour

Given a page with scrollbar
When navigating to a page without scrollbar
And open a `Sidebar`
Then right padding is added to the `<body>` HTML tag

### Expected behaviour

Given a page with scrollbar
When navigating to a page without scrollbar
And open a `Sidebar`
Then no right padding is added to the `<body>` HTML tag

https://codesandbox.io/s/epic-elbakyan-06zsuw?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

106.7.0

### Design Tokens Version

2.30.0

Chrome

MacOS

In General settings, scrollbar should be always enabled.

### Confidentiality

MarcServat:
### Summary

Once the `useScrollBlock` is called, the `clientWidth` of the document is retrieved to calculate the scroll width in a memoized value. If the `clientWidth` changes then the value should be re-calculated.

fixes #5220

### Proposed behaviour

the `clientWidth` is added as a dependency of the `useMemo` hook, so every time the `useScrollBlock` is called, the value is re-calculated if changed.

### Current behaviour

Currently, as the `useMemo` has an empty array of dependencies, it's only called once. If after that, the `clientWidth` changes, for instance, due to router navigation, the value will be out of date.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

For more info about the broken behaviour, see #5220  


mkrds:
Hey @MarcServat Thanks for the PR, your fix is ok but with that approach we'll get a forced reflow each time any of the components that use the `useScrollBlock` rerenders. ( https://gist.github.com/paulirish/5d52fb081b3570c81e3a )

The way I'd fix this issue is to change the `rules` to be a `getRules` method that returns the rules array instead of being a memoized rules array. This way we'll get the reflow only when the `clientWidth` value really has to be checked.

DipperTheDan (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #5260, opened 2022-06-27
refactor(link): remove tabbable prop (FE-5089)

### LINKED ISSUE #5270: Item with submenu doesn't fire onClick
### Current behaviour

OnClick on menu item is not fired when submenu prop exists. Try to click transactions item and check console.

### Expected behaviour

OnClick on menu item is should be fired always.

https://codesandbox.io/s/condescending-pond-ebt2gb

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

108.0.0

### Design Tokens Version

_No response_

Chrome, Safari

MacOS

_No response_

### Confidentiality

grabkowski:
### Proposed behaviour

Remove redundant `tabbable` prop, update the documentation.

Render SubMenu parent item having `onClick` and no `href` as a button, by passing `onClick` to the Link component.

Remove redundant `href`, `target` and `rel` stories, as props described by them are just passed down and rendered as HTML attributes.

Group iconAlign and tooltip stories under the icon story, as these props are dependent on the icon being set.

Change `disabled` story description to match current behaviour.

### Current behaviour

The `tabbable` attribute is redundant, as the component renders either anchor or a button, in which tabbable property could be toggled by disabling a button or not specifying a href in an anchor.

When the Menu is displayed in fullscreen mode, SubMenu parent item with `onClick` prop passed and no `href` is rendered as link instead of a button.

### Checklist

#### QA

### Additional context

fixes #5270

### Testing instructions

Components affected by the change: Link, Menu, TextEditor (editor link).


src/components/menu/__internal__/submenu/submenu.component.js
```diff
@@ -273,6 +273,7 @@ const Submenu = React.forwardRef(
             tabIndex={asPassiveItem ? -1 : 0}
             variant={variant}
             inFullscreenView={inFullscreenView}
+            onClick={!asPassiveItem && !href ? () => {} : undefined}
```
edleeks87, line 276: https://github.com/Sage/carbon/issues/5270 We've had this issue raised the other day, I wonder if we shouldn't add a default value of `() => {}` to the onClick prop and then change this to be like below as it would mean we can close the issue at the same time?

```
  onClick={!asPassiveItem && !href ? onClick : undefined}
```

edleeks87 (commented):
I think you'll also need to update, it shouldn't be "tabbable" when in the editor IIRC
 https://github.com/Sage/carbon/blob/a436ea5b2ee05e4d54afa2a930815aedf4673e84/src/components/text-editor/__internal__/editor-link/editor-link.component.js#L37

src/components/menu/__internal__/submenu/submenu.component.js
grabkowski: we cannot pass down the onClick method as it will be triggered twice - it is already handled in the component. The problem here is how the Link component is rendering a HTML element depending on the passed props.

grabkowski: I take that back, I see that this is handled only in non fullscreen case. That could be simplified even more, as onClick even could be passed down without any conditionals.

edleeks87 (approved):

DipperTheDan (dismissed):

edleeks87 (approved):

DipperTheDan (approved):

src/components/text-editor/__internal__/editor-link/editor-link.component.js
```diff
@@ -29,12 +29,11 @@ const EditorLink = ({ children, contentState, entityKey, ...rest }) => {
 
   return (
     <StyledLink
-      href={validUrl}
+      href={!editMode ? validUrl : undefined}
       title={validUrl}
-      aria-label={validUrl}
+      aria-label={!editMode ? validUrl : undefined}
```
DipperTheDan, line 34: Seeing as `href` and `aria-label` have the same ternary statement, could this not be assigned to `const` instead.

grabkowski: I think that it's easier to read when it's presented like that as you instantly know why it's not rendered here

DipperTheDan: That's fine. Personal preference at the end of the day. Happy to approve the PR still 😄

DipperTheDan (approved):

edleeks87 (approved):

divyajindel (approved):

PR #4981, opened 2022-03-21
fix(menu): change HTML to be more semantic and fix for screenreaders - FE-5009

robinzigmond:
### Proposed behaviour

Fixes for Menu behaviour with screenreaders:
- add role="list" to main menu and submenu ul's, so that Voiceover on Safari announces as a list (see https://www.tempertemper.net/blog/accessibility-issues-when-removing-list-markers)
- change menu items which open submenus from <a> tags with no href (which semantically are just spans) to <button>s
- remove role="link" from Link component when it renders as a button due to onClick prop
- fix Scrollable Block for screen readers so that in a scrollable submenu the correct number of menu items is announced
- update documentation to onChange story on Link component to explain why using this pattern is not accessible and suggest alternatives
- change link text in that story to make clear it's actually a button not a link

### Current behaviour
- Voiceover on Safari does not announce the menus as a list
- screenreaders do not indicate to an unsighted user that any action can be taken on the items which trigger submenus. Voiceover on Chrome even confusingly announces "empty group"
- Link component, when onClick is provided but no href, is announced as a link by screenreaders, which is confusing as it's actually an HTML button so doesn't have the expected behaviour of a link
- when using a scrollable submenu, screenreaders announce each submenu item incorrectly as "list 1 item"
- documentation of onClick on Link component does not mention the problems with this usage
- onClick story on Link has text "this is a link" although it's not correct rendered as a button, which is confusing with screenreaders

### Checklist

- [X] Commits follow our style guide
- [X] Related issues linked in commit messages if required
- [ ] Screenshots are included in the PR if useful
- [X] All themes are supported if required
- [X] Unit tests added or updated if required
- [ ] Cypress automation tests added or updated if required
- [X] Storybook added or updated if required
- [ ] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [ ] Typescript `d.ts` file added or updated if required

#### QA

- [x] Tested in CodeSandbox/storybook
- [ ] Add new Cypress test coverage if required
- [x] Carbon implementation matches Design System/designs
- [x] UI Tests GitHub check reviewed if required

### Testing instructions

Test with a various screenreader/browser combinations (particularly Voiceover on Safari) and ensure users are able to navigate to all menu items.

Also test for regressions on both Menu and Link components.

PR #5268, opened 2022-06-29
Menu: changes to ScrollableBlock component

robinzigmond:
### Proposed behaviour

Change the markup rendered by ScrollableBlock so that:
- if all children of a submenu are wrapped in a ScrollableBlock, then the styles that make the scrollblock work are attached to the submenu's ul element, with the child li elements directly inside.
- wrapping some, but not all, submenu children in a ScrollableBlock is no longer directly supported. The same behaviour is now enabled by having the scrollable list as a nested list, as a child of a parent item in the submenu - and the JSX markup now reflects this.

It is now possible to have multiple ScrollableBlock components within the same submenu (see Menu Item Four in updated "scrollable submenu" story).

### Current behaviour

ScrollableBlock component renders as a couple of divs wrapping their children - which typically are Menuitems, rendered as li tags. As a result we have invalid HTML inside a ScrollableBlock - with divs as direct children of a ul. Although this was fixed for most screenreaders in in [this PR](https://github.com/Sage/carbon/pull/4981), it still doesn't work in Firefox (on either Mac/Voiceover or Windows/NVDA) - and even if it did, invalid HTML is itself a WCAG failure (see https://www.digitala11y.com/understanding-sc-4-1-1-parsing/).

It was not previously possible to have multiple ScrollableBlock components within the same submenu.

It was possible previously to have a ScrollableBlock wrapping some, but not all, items of a submenu, and including the first item. This is no longer supported after this PR, as this use-case (which was not shown in Storybook and is not used by anyone as far as I know) does not make sense semantically and cannot be rendered with valid HTML.

### Checklist

#### QA

### Testing instructions

Check specifically that keyboard navigation within submenus with one or more ScrollableBlocks works as intended.

Test with all applicable browser/screenreader combinations. (Note that the semantics have changed and therefore so has the expected screenreader behaviour - the new Menu Item Four in the storybook example would have previously been announced as having 14 items in the submenu list, now it will announce as 3 items but where the second and third each have sublists of 6 items each.)

Check the following CodeSandboxes showing a demo of a searchbox and scrollable list to search in. This is how it would work under the old Carbon:
https://codesandbox.io/s/submenu-scrollable-search-working-example-i0tjcr
This is updated to work with this PR:
https://codesandbox.io/s/submenu-scrollable-search-rewritten-requires-updated-carbon-jxjxte


src/components/menu/menu.stories.mdx
```diff
@@ -457,15 +457,15 @@ In order to align text and icons within a submenu a `Box` component will need to
 
 ### Scrollable submenu
 
-A scrollable submenu can be added using the `ScrollableBlock` component. This can be used for all of the submenu items, or just a selection, as shown in `Menu Itme Four` below.
-Note that only one `ScrollableBlock` can be used within a single submenu.
+A scrollable submenu can be added using the `ScrollableBlock` component. This can be used for all of the submenu items, or for a child list of one menu item.
```
Parsium, line 460: I think it would be a good idea to expand on your points with some visual examples of correct and incorrect uses of `ScrollableBlock` in the docs, similar to the examples you used in your tests. Perhaps something like this?:

```
Correct ✅ 
<Menu>
  <MenuItem submenu="Shopping items">
     <MenuItem>Apple</MenuItem>
     <MenuItem>Carrot</MenuItem>
     <MenuItem>
         Meat
         <ScrollableBlock>
             <MenuItem>Beef</MenuItem>
             <MenuItem>Chicken</MenuItem>
         </ScrollableBlock>
     </MenuItem>
  </MenuItem>
</Menu>

Incorrect ❌ 
<Menu>
  <MenuItem submenu="Shopping Items">
     <MenuItem>Apple</MenuItem>
     <MenuItem>Carrot</MenuItem>
     <MenuItem>Meat</MenuItem>
     <ScrollableBlock>
         <MenuItem>Beef</MenuItem>
         <MenuItem>Chicken</MenuItem>
     </ScrollableBlock>
  </MenuItem>
</Menu>
```

Parsium (commented):
I learnt several new things from this PR @robinzigmond! It's great the html involved with `ScrollableBlock` looks a lot more semantically sound now 👏🏼  

What do you think about adding some `invariant` statements to `getBlocks()` in `submenu.component.js`?  It might be an idea to do some error handling to alert consumers if their use of `ScrollableBlock` is not semantically correct as outlined by your design.

grabkowski (commented):
Wouldn't it be easier to just make the ScrollableBlock an internal component and introduce two props to the MenuItem in it's place?
Instead of a lot of complex code that checks component types, we could just wrap the submenu children in a ScrollableBlock on our side.
```<MenuItem submenu="Submenu" scrollable scrollHeight="200px">...</MenuItem>```

robinzigmond:
closing as I have just raised https://github.com/Sage/carbon/pull/5305 from a separate branch which fulfils the same functionality but changes the interface and (slightly) simplifies the code as suggested by @grabkowski

PR #5285, opened 2022-07-05
test(button): modify tests to check only one icon type (FE-5229)

### LINKED ISSUE #5273: [Button] Page is getting stuck for button test component and observing lots of icons
### Current behaviour

Steps to reproduce:

1) go to VS code
2) checkout master
3) run npm start
4) it will navigated to the http://localhost:9001
5) go to button component and select test 
6) try to click on default story after that click on next one
7) the page will get stuck

Screenshot 2022-06-30 at 11 04 33

### Expected behaviour

Page should work accordingly.

http://localhost:9001/?path=/docs/button-test--primary-buttons-icons-before

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

latest

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

grabkowski:
### Proposed behaviour

Test only one icon in various button types.

### Current behaviour

Every icon is being rendered for every type of button, which results in enormous tests that are freezing browsers and causing Chromatic to not render the capture image on Chrome.

### Checklist

#### QA

### Additional context

closes #5273

### Testing instructions


edleeks87 (approved):

src/components/button/button-test.stories.mdx
```diff
@@ -113,7 +113,7 @@ export const ButtonAsASiblingStory = ({
 export const generateButtons = (buttonType, iconPosition) => {
   return (
     <>
-      {["", ...ICONS].map((iconType) => {
+      {["", "bin"].map((iconType) => {
```
DipperTheDan, line 116: What would you say to having the `services` icon in this snapshot too so we can have a visual test to cover this code block?
https://github.com/Sage/carbon/blob/master/src/components/button/button.style.ts#L10-L13

```suggestion
      {["", "bin", "services"].map((iconType) => {
```

grabkowski: I think that line is added because the `services` icon is broken and it needs to be changed. This case is already tested in a dedicated unit test, both for the `Icon` and the `Button` component.
In my opinion, we shouldn't render that icon here, as it will result in rendering more than twenty buttons more.

PR #5287, opened 2022-07-06
fix(pager): change styles so all items are exposed to accessibility tree

### LINKED ISSUE #4637: Screen Reader reads two separate elements as one within Flat Table with StickyFooter
### Current behaviour

The screen readers, NVDA (Windows) and VoiceOver (MacOS) combines two elements on the StickyFooter in the Flat Table component and reads the elements on the StickyFooter - Pager component as one. 

### Expected behaviour

The screen readers are able to read the elements separately instead of reading it at once.

https://carbon.sbc-ui.com/?path=/docs/flat-table--with-sticky-footer

### JIRA Ticket (Sage Only)

SBS-25588

### Suggested Solution

_No response_

### Carbon Version

101.1.0

Chrome, Safari, Microsoft Edge

MacOS, Windows

If the StickyFooter - Pager component has a `Next` and `Previous` button, the screen reader reads the elements correctly.

### Confidentiality

robinzigmond:
When using Chrome, when the Previous and Next buttons are missing due to only 1 page being present,
screenreaders such as NVDA and VoiceOver were treating "items" and "page", and the total page number
and "X items" messages, as the same element when moving through one item at a time. This appears to
be due to a bug or feature in Chrome whereby generic elements (div or span) with display property
inline or flex were not being exposed to the accessibility tree at all. So "X items" has been
changed from flex to block, and "page" from inline to block (div in place of span), to fix this.

fix #4637

### Proposed behaviour

When using the NVDA screenreader on Windows, and navigating one page at a time, when a Pager is displayed but missing the "previous" and "next" buttons, the word "items" (next to the select) should be read separately from "page", and the numbers in "page 1 and 1" and "2 items" should be read separately from each other.

### Current behaviour

When using Chrome, both NVDA on Windows and Voiceover on Chrome read the above two items as one when stepping through one element at a time - so you hear "items-page" and "1-2 items" at the same time, which is illogical and confusing. It's particularly bad on NVDA as there the "1-2items" is actually read as "12 items".

### Checklist

#### QA

### Testing instructions

Can be tested on this Storybook story: https://carbon.sage.com/?path=/story/flat-table--with-sticky-footer-inside-of-larger-div

Test in all supported browser/screenreader combinations on both Windows and Mac, stepping through one item at a time (shortcut-keys: Control+Option+RightArrow on VO, just right arrow on NVDA), to ensure the behaviour is as expected.


src/components/pager/pager.style.js
```diff
@@ -109,9 +109,8 @@ const StyledPagerNoSelect = styled.span`
 `;
 
 const StyledPagerSummary = styled.div`
-  display: flex;
   flex: 1 1 30%;
-  justify-content: flex-end;
+  text-align: right;
```
edleeks87, line 113: This will slightly alter how it looks when the text wraps at smaller screens so might need UX to approve just to be sure (I'm happy to approve the change though). 

Updated behaviour:
image

Current behaviour :
image



Looking at the html for this it can do with a real tidy up (I suspect Git BLAME will say it was me 🤦 so don't look), we have a div rendered empty when the page size selector isn't rendered and multiple containers for the elements are flexed when we can probably just rely on the Pager container handling the flex and layout and would likely negate the needing a `StyledPagerSummary` (a standard div would work). However, all this can be tidied up in a separate ticket as we have one on the backlog to review and fix the html for Pager.

robinzigmond: good point, I didn't think about what would happen when the content has to wrap. I'll take a look - there's probably a way to keep the flex while fixing what screenreaders do, it was just laziness from me that I thought "that flex is only there to basically right justify it, so text-align: right will work instead" without thinking of wrapping.

edleeks87 (approved):

sianford (approved):

PR #5305, opened 2022-07-13
fix incorrect HTML output by ScrollableBlock, add scrollable prop and move ScrollableBlock to internal component

robinzigmond:
Ensure that the HTML output by a ScrollableBlock within a submenu does not incorrectly put divs
between `<ul>` and `<li>` elements - this invalid markup is a WCAG failure in itself and also causes the
menu to be announced incorrectly by some browser/screenreader combinations. Scrollable sections are
now rendered as child lists, and enabled by using the new `scrollable` prop on a MenuItem (either
the top-level item that opens the submenu, or an item within a submenu), with the ScrollableBlock
component now being a purely internal component.

BREAKING CHANGE: ScrollableBlock is now no longer usable. To get the same functionality, add the
`scrollable` prop to the MenuItem that contains it - along with the `scrollHeight` and
`scrollVariant` props that replace the previous `height` and `variant` props of ScrollableBlock. For
cases where ScrollableBlock would have wrapped a subset of the menu items in a submenu, the previous
menu item is now treated as a parent and should wrap the following MenuItems, as well as its text,
and be given the new props. Note that cases where a ScrollableBlock wrapped a subset of menu items
including the first one are no longer supported.

### Proposed behaviour

Change the markup rendered by ScrollableBlock so that:

- if all children of a submenu are wrapped in a ScrollableBlock, then the styles that make the scrollblock work are attached to the submenu's ul element, with the child li elements directly inside.
- wrapping some, but not all, submenu children in a ScrollableBlock is no longer directly supported. The same behaviour is now enabled by having the scrollable list as a nested list, as a child of a parent item in the submenu

Also simplify consumer code by moving ScrollableBlock to an internal component, and enabling this behaviour by adding a new `scrollable` prop to the parent MenuItem component

### Current behaviour

ScrollableBlock component renders as a couple of divs wrapping their children - which typically are Menuitems, rendered as li tags. As a result we have invalid HTML inside a ScrollableBlock - with divs as direct children of a ul. Although this was fixed for most screenreaders in in https://github.com/Sage/carbon/pull/4981, it still doesn't work in Firefox (on either Mac/Voiceover or Windows/NVDA) - and even if it did, invalid HTML is itself a WCAG failure (see https://www.digitala11y.com/understanding-sc-4-1-1-parsing/).

It was not previously possible to have multiple ScrollableBlock components within the same submenu.

It was possible previously to have a ScrollableBlock wrapping some, but not all, items of a submenu, and including the first item. This is no longer supported after this PR, as this use-case (which was not shown in Storybook and is not used by anyone as far as I know) does not make sense semantically and cannot be rendered with valid HTML.

### Checklist

#### QA

### Testing instructions

Check specifically that keyboard navigation within submenus with one or more ScrollableBlocks works as intended.

Test with all applicable browser/screenreader combinations. (Note that the semantics have changed and therefore so has the expected screenreader behaviour - the new Menu Item Four in the storybook example would have previously been announced as having 14 items in the submenu list, now it will announce as 3 items but where the second and third each have sublists of 6 items each.)

Check the following CodeSandboxes showing a demo of a searchbox and scrollable list to search in. This is how it would work under the old Carbon:
https://codesandbox.io/s/submenu-scrollable-search-working-example-i0tjcr
This is updated to work with this PR:
https://codesandbox.io/s/submenu-scrollable-search-rewritten-requires-updated-carbon-jxjxte


robinzigmond:
closing this one in favour of the new approach in https://github.com/Sage/carbon/pull/5329

PR #5326, opened 2022-07-22
Refactor FocusTrap to TypeScript FE-4853

### LINKED ISSUE #5295: Back tabbing breaks focus trap

### LINKED ISSUE #5342: Focus does not remain within open dialog when Shift + tab is used
### Current behaviour

Focus does not remain within open dialog when Shift + tab on keyboard is used

### Expected behaviour

Focus remains within open dialog when Shift + tab on keyboard is used

https://carbon.sage.com/v/109.1.2/index.html?path=/story/dialog--default-story

### JIRA Ticket (Sage Only)

SBS-39486

### Suggested Solution

_No response_

### Carbon Version

109.1.2

### Design Tokens Version

2.40.0

Firefox, Chrome, Safari

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5295
fix #5342

### Proposed behaviour

Ensures the keydown listener is registered in the event capture phase
Refactors files into TypeScript
Refactors unit tests into RTL

### Current behaviour

The keydown listener is registered in the event propagation phase
Component files are JavaScript
Unit tests use Enzyme

### Checklist

#### QA

### Testing instructions

Test bug described in issue in storybook and in linked sandbox

https://codesandbox.io/s/beautiful-gagarin-cqorrv?file=/src/index.js


src/__internal__/focus-trap/focus-trap-utils.ts
```diff
@@ -25,34 +26,31 @@ const waitForVisibleAndFocus = (element) => {
...
   }
   return buttonsInGroup[shiftKey ? buttonsInGroup.length - 1 : 0];
 };
 
-const getNextElement = (element, focusableElements, shiftKey) => {
+const getNextElement = (
+  element: HTMLElement,
+  focusableElements: HTMLElement[],
+  shiftKey: boolean
+) => {
```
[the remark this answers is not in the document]
edleeks87: Yeah I think you are right 👍

src/__internal__/focus-trap/focus-trap.component.tsx
```diff
@@ -109,43 +146,43 @@ const FocusTrap = ({
       const { activeElement } = document;
```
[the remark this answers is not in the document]
edleeks87: I'll have another look and go with the suggestion if I get nowhere 👍

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+
+const shiftTabPress = () =>
+  fireEvent.keyDown(screen.getByTestId(WRAPPER_ID), {
+    key: "Tab",
+    shiftKey: true,
+  });
+
+const focusElement = (target: HTMLElement) => {
+  target.focus();
+  expect(target).toBeFocused();
```
Parsium, line 42: We may be able to replace several of our custom matchers/test helpers with identical ones from the jest-dom library, such as the `toHaveFocus` here. That way we could reduce the complexity of our spec helpers code
```suggestion
  expect(target).toHaveFocus();
```

edleeks87: Good point, I'll update it as we no longer need to handle for scenarios where the passed argument isn't a DOM element when using RTL 👍

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+      expect(screen.getByText(BUTTON_IN_ADDITIONAL_WRAPPER_ONE)).toBeFocused();
+      tabPress();
+      expect(screen.getByText(BUTTON_IN_ADDITIONAL_WRAPPER_TWO)).toBeFocused();
+      tabPress();
+      expect(screen.getByText(BUTTON_IN_WRAPPER)).toBeFocused();
+    });
+  });
+
+  describe("when content in the children tree changes", () => {
+    it("should trigger the MutationObserver", () => {
```
Parsium, line 718: We might be able to workaround testing the implementation details of `MutationObserver` here by using some of the [async wait utilities part of RTL](https://testing-library.com/docs/guide-disappearance#waiting-for-appearance) to wait until `children` changes and verify if the button is then focused. That way we could mimic what the user would see. Though we may still need to write a manual jest mock for `MutationObserver` like you've done here, since I can't seem to find any information that `js-dom` simulates it for us. What do you think?

If this isn't feasible, perhaps it would be worth us in a separate ticket creating a manual jest mock for `MutationObserver` inside `src/__spec_helper__` anyway, like we have done for `matchMedia` and `ResizeObserver`, to help keep things consistent?

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+      </FocusTrap>
+    </ModalContext.Provider>
+  );
+};
+
+describe("FocusTrap", () => {
+  describe("triggerRefocusFlag", () => {
+    it("refocuses the last element that had focus within the trap when flag is set", () => {
+      const { rerender } = render(
+        <MockComponent autoFocus={false} triggerRefocusFlag={false}>
```
Parsium, line 93: Could we separate out the UI code here as another mock component, such as `<TriggerRefocusExample/>`, similarly to how the QAs have done with the refactored cypress tests? That way we could avoid duplicating the larger UI code examples.

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+          <input type="text" />
+        </MockComponent>
+      );
+      rerender(
+        <MockComponent autoFocus={false} triggerRefocusFlag tabIndex={-1}>
+          <button type="button">{BUTTON_ONE}</button>
+          <input type="text" />
+        </MockComponent>
+      );
+      expect(screen.getByTestId(WRAPPER_ID)).toBeFocused();
```
Parsium, line 139:
```suggestion
      const { rerender } = render(
        <MockComponent
          autoFocus={false}
          triggerRefocusFlag={false}
          tabIndex={-1}
        >
          <button type="button">{BUTTON_ONE}</button>
          <input type="text" />
        </MockComponent>
      );
      expect(screen.getByTestId(WRAPPER_ID)).toBeFocused();
      rerender(
        <MockComponent autoFocus={false} triggerRefocusFlag tabIndex={-1}>
          <button type="button">{BUTTON_ONE}</button>
          <input type="text" />
        </MockComponent>
      );
      expect(screen.getByTestId(WRAPPER_ID)).toBeFocused();
```
Would we need to verify the wrapper had focus before the flag was set as well so we are testing it stays in focus?

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+
+  describe("when autoFocus is false", () => {
+    it("should not focus the first focusable element by default", () => {
+      render(
+        <MockComponent autoFocus={false}>
+          <button type="button">{BUTTON_ONE}</button>
+          <input type="text" />
+        </MockComponent>
+      );
+      expect(document.querySelectorAll("body")[0]).toBeFocused();
```
Parsium, line 184:
```suggestion
      const button = screen.getByRole("button", { name: BUTTON_ONE });
      expect(button).not.toHaveFocus();
```

What do you think of this?

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+
+  describe("when isAnimationComplete is false", () => {
+    it("should not focus the first focusable element by default", () => {
+      render(
+        <MockComponent isAnimationComplete={false}>
+          <button type="button">{BUTTON_ONE}</button>
+          <input type="text" />
+        </MockComponent>
+      );
+      expect(document.querySelectorAll("body")[0]).toBeFocused();
```
Parsium, line 196: Similarly to the previous test :)

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+
+    it("should focus the element that ref passed to focusFirstElement and loop round when back tabbing", () => {
+      expect(screen.getByText(FIRST_ELEMENT)).toBeFocused();
+      shiftTabPress();
+      expect(screen.getByText(BUTTON_TWO)).toBeFocused();
+      shiftTabPress();
+      expect(screen.getByText(BUTTON_ONE)).toBeFocused();
+      shiftTabPress();
+      expect(screen.getByText(FIRST_ELEMENT)).toBeFocused();
+    });
```
Parsium, line 218: Could we split this into two tests here? One being `should focus one the first focusable child element` and `should loop back to first focusable element through back tabbing`

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+
+    it("refocuses the container within the trap when flag is set, if the wrapper has no tabindex", () => {
+      const { rerender } = render(
+        <MockComponent autoFocus={false} triggerRefocusFlag={false}>
+          <button type="button">{BUTTON_ONE}</button>
+          <input type="text" />
+        </MockComponent>
+      );
+      rerender(
```
Parsium, line 149:
```suggestion
      const { rerender } = render(
        <MockComponent autoFocus={false} triggerRefocusFlag={false}>
          <button type="button">{BUTTON_ONE}</button>
          <input type="text" />
        </MockComponent>
      );
      expect(screen.getByTestId(WRAPPER_ID)).toHaveFocus();
      rerender(
```

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -0,0 +1,779 @@
...
+        <MockComponent
+          autoFocus={false}
+          triggerRefocusFlag={false}
+          tabIndex={-1}
+        >
+          <button type="button">{BUTTON_ONE}</button>
+          <input type="text" />
+        </MockComponent>
+      );
+      rerender(
```
Parsium, line 133:
```suggestion
      const { rerender } = render(
        <MockComponent
          autoFocus={false}
          triggerRefocusFlag={false}
          tabIndex={-1}
        >
          <button type="button">{BUTTON_ONE}</button>
          <input type="text" />
        </MockComponent>
      );
      expect(screen.getByTestId(WRAPPER_ID)).toHaveFocus();
      rerender(
```
Would we need to verify the wrapper had focus before the flag was set as well so we are testing it stays in focus?

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: I'm happy to add this in, I'll probably leave the other one in to assert that focus wasn't moved etc

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: I did consider this but ended up deciding that it was okay as one test as if I split I'd still be asserting that the first focusable element is focused in the looping one

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: There are other test blocks that assert that the wrapper is focused, but I can delete them and add it here I guess as there's no point in duplicating

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: I'll see if I can just move everything into a single wrapper and avoid the need to have any but the one

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: I don't think there's a need to write a mock for the MutationObserver as this is the only file that uses it (if that changes it will be worthwhile) and I think we should really be testing this behaviour in a browser via cypress but we have coverage constraints. The coverage is actually met elsewhere for this but I decided to keep this test as it will let anyone working on it know that it has failed straightaway rather than catching it via missing coverage etc. I do agree your suggestion is a bit tidier than what we originally had so I'll update it a remove the mocking of the observer.

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: see above reply

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -86,37 +85,35 @@ const MockComponent = ({
...
+const defaultChildren = (
+  <>
+    <button type="button">{BUTTON_ONE}</button>
+    <button type="button">{BUTTON_TWO}</button>
+  </>
+);
+
+const renderMockComponent = (props: Partial<MockComponentProps> = {}) => {
+  const children = props.children || defaultChildren;
+  return render(<MockComponent {...props}>{children}</MockComponent>);
```
robinzigmond, line 97: While I assume it works, this whole setup reads confusingly to me. `children` is of course a prop too, so `{...props}` will pass down the children - but then were also implicitly passing `{children}` as the children. Certainly makes me scratch my head, even though the 2 different children props should always be the same if both present.

Better in my opinion would be to write it like this:

```
const renderMockComponent = ({ children = defaultChildren, ...rest }: Partial<MockComponentProps> = {}) => {
  return render(<MockComponent {...rest}>{children}</MockComponent>);
 };
```

(which I just tried and seems to work OK)

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -86,37 +85,35 @@ const MockComponent = ({
...
         <MockComponent autoFocus={false} triggerRefocusFlag={false}>
-          <button type="button">{BUTTON_ONE}</button>
-          <Select
-            id="c0499f86-d5a7-4a72-a0b7-753a2d218c54"
-            label="the dropdown"
-            aria-label="input"
-          >
-            <Option value="1" text="Option 1" />
-          </Select>
+          {defaultChildren}
```
robinzigmond, line 105: surely you don't need any children here if you're just passing in `defaultChildren`? You can just call `renderMockComponent` with the non-children props? Either that or not bother with the default inside the `renderMockComponent` definition.

(Same comment applies to other places where you've used `{defaultChildren"` as the children, of course. I won't comment on all of them separately!)

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -258,190 +237,187 @@ describe("FocusTrap", () => {
...
     });
 
     describe("and element does not have focusable items", () => {
       beforeEach(() => {
-        render(
-          <MockComponent>
-            <p>Test content</p>
-          </MockComponent>
-        );
+        renderMockComponent({ children: <p>Test content</p> });
```
robinzigmond, line 289: actually when there are bespoke children I think `render(...)` reads better than `renderMockComponent({children: ...})`, particularly in the later examples with complex children. But I don't mind if it's kept as is.

src/__internal__/focus-trap/focus-trap.spec.tsx
```diff
@@ -706,34 +684,16 @@ describe("FocusTrap", () => {
...
-        this.disconnect = jest.fn();
-        this.trigger = () => {
-          callback();
-        };
-      });
-      // casted to any here as we don't really care about the interface we are mocking
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      (global as any).MutationObserver = mutationObserverMock;
-
+    it("should trigger the MutationObserver", async () => {
```
robinzigmond, line 696: I'm not entirely sure what this is testing now, but it certainly doesn't have anything to do with MutationObserver any more! I'm OK with a decision to change what this is testing (MutationObserver should be just an implementation detail anyway and we should test that the trap still works even as the focusable elements change, which I think is what we're doing now and seens perfectly correct) - but the name of the test should change to reflect this.

edleeks87: Good point the description no longer matches what is being tested

src/__internal__/focus-trap/focus-trap.spec.tsx
edleeks87: I'm not sure this would work here as I need to call the `rerender` which is desctructured from the render function, I can do it like below but I'd still have to pass the same composed component back in so decided to leave it as is for these test cases.

```
      const { rerender } = renderMockComponent({autoFocus: false, triggerRefocusFlag: false});
  
      rerender(
        <MockComponent autoFocus={false} triggerRefocusFlag>
          {defaultChildren}
        </MockComponent>
      );
```

I've decided to do a slight refactor so that I can use re-use the function by just having it return the component and passing that into the `render`, so I can do something like below

```

 const { rerender } = render(renderMockComponent({autoFocus: false, triggerRefocusFlag: false}));
 
 rerender(renderMockComponent({autoFocus: false, triggerRefocusFlag: true}))
 ```

File: src/__internal__/focus-trap/focus-trap.spec.tsx (hunk 14 above)
Comment (edleeks87) [reply src/__internal__/focus-trap/focus-trap.spec.tsx:289]: Yeah inclined to agree, especially based on my above comment about refactoring to support the rerender test cases

**Review by robinzigmond — DISMISSED** [review]

**Review by Parsium — APPROVED** [review]

**Review by robinzigmond — CHANGES_REQUESTED** [review]
the "fix issue with additional wrappers" commit should really be a "fix" rather than a "chore". I thought for a while this wouldn't matter as there are other fix commits here, but then I remember that this means the list of bugfixes in the new version won't include this one which might be confusing for anyone who might be looking out for this particular bugfix now or in future.

**Comment by edleeks87** [comment]
> the "fix issue with additional wrappers" commit should really be a "fix" rather than a "chore". I thought for a while this wouldn't matter as there are other fix commits here, but then I remember that this means the list of bugfixes in the new version won't include this one which might be confusing for anyone who might be looking out for this particular bugfix now or in future.

The `chore` commits will be squashed when I rebase and merge, they've just been done separately to make reviewing easier

**Review by DipperTheDan — DISMISSED** [review]

**Review by robinzigmond — DISMISSED** [review]

File: src/components/dialog-full-screen/dialog-full-screen.stories.mdx [hunk 16]
```diff
@@ -949,11 +949,11 @@ This may occasionally be useful with things like Toasts where they persist on th
               stickyFooter={true}
               height="500px"
               leftSideButtons={
-                <Button onClick={() => setIsToast1Open(true)}>Show first toast</Button>
+                <Button onClick={() => setIsOpen(false)}>Cancel</Button>
```
Parsium, line 952:
```suggestion
                <Button onClick={() => setIsDialogOpen(false)}>Cancel</Button>
```
Currently the cancel button does not work in the story due to this typo

src/components/dialog/dialog.stories.mdx
```diff
@@ -428,11 +428,11 @@ This may occasionally be useful with things like Toasts where they persist on th
               stickyFooter={true}
               height="500px"
               leftSideButtons={
-                <Button onClick={() => setIsToast1Open(true)}>Show first toast</Button>
+                <Button onClick={() => setIsOpen(false)}>Cancel</Button>
```
Parsium, line 431:
```suggestion
                <Button onClick={() => setIsDialogOpen(false)}>Cancel</Button>
```
Same here :)

src/components/sidebar/sidebar.stories.mdx
```diff
@@ -269,11 +270,11 @@ This may occasionally be useful with things like Toasts where they persist on th
               stickyFooter={true}
               height="500px"
               leftSideButtons={
-                <Button onClick={() => setIsToast1Open(true)}>Show first toast</Button>
+                <Button onClick={() => setIsOpen(false)}>Cancel</Button>
```
Parsium, line 273:
```suggestion
                <Button onClick={() => setIsDialogOpen(false)}>Cancel</Button>
```

DipperTheDan (approved):

Parsium (approved):

robinzigmond (approved):

DlgSHi (approved):

PR #5346, opened 2022-08-02
feat(rebrand): updated storybook to new brand and added styling overrides

harpalsingh:
### Proposed behaviour

This PR updates storybook to ensure new brand is applied visually.

**The following elements have been updated:**

- Updated "Carbon by Sage" logo
- Updated footer Sage logo
- Update storybook theme base values for new brand colours
- Welcome page updated background colors
- Custom overrides on storybook CSS for better brand alignment and style
- Colour of search term
- Focus state colors
- Skip to button style corrected

### Screenshots of update

Screenshot 2022-08-14 at 20 32 36

Screenshot 2022-08-14 at 20 32 51

Screenshot 2022-08-14 at 20 33 10

Screenshot 2022-08-14 at 20 33 15

Screenshot 2022-08-14 at 20 33 25

### Current behaviour

Screenshot 2022-08-02 at 14 45 39

Screenshot 2022-08-02 at 14 46 09

### Checklist

#### QA

### Testing instructions


.storybook/manager-head.html
```diff
@@ -0,0 +1,34 @@
...
+  .sidebar-item[data-nodetype="component"]:hover {
+    background: #007E45 !important;
+    color:  white !important;
+  }
+
+  .search-result-item--label mark {
+    color:  #0060a7 !important;
+  }
+
+  .css-1va4tas {
```
jamime, line 31: This class will be re-generated when we upgrade storybook. Is there another class you can use?

harpalsingh: I wasn't sure if that was the case as I rebuilt and it seemed to stay the same, I will double check, otherwise I will remove this line.

jamime (commented):
I spotted some blue on the focus state.
image

Skip to canvas is hard to read
image


Shall we use brand icons for the home page too instead of these?
image


Product image is out of date, shall we update/remove?
image


I think this background should be darker to match the new logo, what do you think?
image

.storybook/manager-head.html
jamime: I think it's a guid based on the styles of the component. So It'll only change if we upgrade storybook, and, the styles change in storybook itself.

harpalsingh:
@jamime Yea I think we will do another pass at the design but I wanted to get another look at all those and changes/updates we might want to do as well as content edit, this pass is mostly a rebrand basic phase 1 type. I will look at the canvas skip button and the blue was hard to find why its doing it via focus/active, will check that out.

.storybook/manager-head.html
```diff
@@ -0,0 +1,34 @@
+<style>
+  .sidebar-item {
```
edleeks87, line 2: Instead of `!important`

```suggestion
  .sidebar-item.sidebar-item {
```

.storybook/manager-head.html
```diff
@@ -0,0 +1,34 @@
...
+    padding-bottom: 6px !important;
+    padding-top: 6px !important;
+  }
+
+  .sidebar-item svg {
```
edleeks87, line 7:
```suggestion
  .sidebar-item.sidebar-item svg {
```

edleeks87: You can likely remove all the uses of `!important` if you double up on the class name to raise the specificity.

Eg: `.sidebar-item.sidebar-item[data-selected="true"]`

.storybook/manager-head.html
edleeks87: Think Jamie is correct about this, I haven't seen anything in the docs that offers an alternate way of achieving this. Maybe a little bit less brittle (but still less than ideal) approach would be to target `[role="main"]` or just not have the box shadow at all for now 🤷 

PS you can also remove `!important` here if you do

.storybook/manager-head.html
harpalsingh: I was trying a few things and it wouldn't take, but I can check that again.

edleeks87 (approved):

edleeks87 (dismissed):

edleeks87 (approved):

jamime (dismissed):

edleeks87 (dismissed):

jamime (dismissed):

edleeks87 (approved):

harpalsingh:
As discussed with @DlgSHi there are improvements to be made to focus states and further investigation and knock on effects of these and we will handle this in the next PR of work to further update storybook.

DlgSHi:
The focus outline for the main header and for the `actions/controls/accessibility` is incorrect.

https://user-images.githubusercontent.com/34001198/186857829-8a841060-3de5-46ed-9041-9ef4c2fd1b12.mov


Lack of focus for the `go to fullscreen` button
Screenshot 2022-08-26 at 10 24 56

The `skipLink` blinks when is focused for Safari only: 

https://user-images.githubusercontent.com/34001198/186858332-2be8b11e-2a16-4891-8ac8-a9647f521224.mov

DlgSHi (approved):

DipperTheDan (approved):

jamime (approved):

PR #5347, opened 2022-08-02
fix(popover-container): ensure clicks inside popover do not close it

### LINKED ISSUE #5322: PopoverContainer closes when clicking inside of it
### Current behaviour

The PopoverContainer closes wen a click event occurs inside the popover, the functionality regressed when this PR was merged https://github.com/Sage/carbon/pull/5262

### Expected behaviour

It should remain open and only close when the click occurs outside of the container

https://codesandbox.io/s/keen-sammet-zuw4wk?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

It looks like we just need to add the `popoverContentNodeRef` to the `useClickAwayListener` refs array argument
  
```
  useClickAwayListener([ref, popoverContentNodeRef], handleClickAway, "mousedown");
```

### Carbon Version

latest

### Design Tokens Version

_No response_

MacOS

There also seems to be an issue with using Selects in `PopoverContainer` where it closes the container when an Option is clicked. The linked gif has the above suggested fix applied

select in popover

### Confidentiality

robinzigmond:
Fix bug where portals inside a component did not automatically work with useClickAwayListener.

Refactor useClickAwayListener to no longer take an array of refs and return a stopPropagation handler which can be attached to the parent component - this automatically handles any React children no matter whether they are rendered as DOM children or not. This ensure that eg a Select inside a Popover Container will not cause the popover to close when a select option is clicked.

Remove any refs in parent components which are now unneeded.

fix #5322

### Proposed behaviour
Clicking inside a Popover Container should not close it

### Current behaviour
Clicking inside a Popover Container causes it to close (since version 109.1.3)

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/keen-sammet-zuw4wk?file=/src/App.js

Also test the case with a Select inside a Popover Container (mentioned towards the bottom of the original bug report on Github) - it should be possible to choose an option from the Select without the popover closing. This can be checked with the new Test story "with select" in Popover Container (and a Cypress test has been added to guard against regressions).


src/hooks/__internal__/useClickAwayListener/useClickAwayListener.ts
```diff
@@ -1,29 +1,20 @@
-import React, { useEffect, useRef } from "react";
+import React, { useEffect } from "react";
+
+const stopPropagation = (ev: React.MouseEvent) => {
+  ev.stopPropagation();
```
edleeks87, line 4: Whilst this on the whole will work I don't think we should be stopping event propagation in Carbon (or at least we should be extremely careful when we do) as we don't know what the UI composition for consumers will be and therefore what we could be affecting. I also believe this will require the user to double click to outside the the container to close it if they have an additional popover (Select for example) open when I would expect both to close with single click outside of the PopoverContainer boundary.

https://stackoverflow.com/a/54817255

robinzigmond: Yes, thanks - I guess I agree having thought about it and read that SO comment that it's ideally something to avoid. I don't see how else to do this in a way that works with Portals but will look at how MUI do it. If there isn't another good way I'm still inclined to use stopPropagation because I honestly don't see what event handlers for click or mousedown a consumer is going to want on a parent component. But point taken, I'm going to do some research and if there's a way to avoid stopping propagation - without writing hideously overcomplicated code which likely still won't cover all possible cases - I will.

edleeks87 (approved):
Happy to approve this as is, might be worth checking whether the popover container should close when the click is outside its boundary and a select list is also open or if the current behaviour is okay (2 clicks to close). I've also noticed that we've (it was me really 🤦 ) missed adding the `useClickAwayListener` to the Selects but that can be done in a separate ticket once this has been merged 👍

edleeks87 (approved):

mkrds (approved):

harpalsingh:
@edleeks87 For me 2 clicks to close is ok, in particular if we think about the possible use case, so happy to approve otherwise closing 2 things might be troublesome based on context and 2 clicks quickly would close it anyway. I will however investigate this further and dont want that to hold up this ticket so if design requires that change, we will create an additional ticket.

ZhuoyuJin (approved):

PR #5348, opened 2022-08-02
feat(portrait): convert component to typescript - FE-4821

mkrds:
### Proposed behaviour

This PR contains Portrait component TS conversion

### Checklist

#### QA

### Testing instructions

Portrait component regression


src/components/portrait/portrait-gravatar.component.tsx
```diff
@@ -4,15 +4,28 @@ import MD5 from "crypto-js/md5";
...
 
 const PortraitGravatar = ({
   gravatarEmail,
   size,
   alt,
   shape,
   errorCallback,
-  ...rest
-}) => {
+}: PortraitGravatarProps) => {
```
edleeks87, line 28: You've left the prop types in this file

src/components/portrait/portrait.component.tsx
```diff
@@ -129,60 +177,4 @@ const Portrait = ({
   return renderComponent();
 };
 
-Portrait.propTypes = {
-  ...marginPropTypes,
-  /** The size of the Portrait. */
-  size: PropTypes.oneOf(["XS", "S", "M", "ML", "L", "XL", "XXL"]),
-  /** A custom image URL. */
-  src: (props) => {
```
edleeks87, line 137: Should we have an invariant for the `src` prop? As I don't think any consuming JS projects won't get any feedback that they shouldn't set both `src` and `gravatar` props at the same time

mkrds: I thought about it and decided that it's not worth it in this case. It's not breaking anything crucial and even if it does it is an immediately visible visual change.

edleeks87: That's a good point I hadn't checked if it breaks when both are applied, the fact it doesn't means we shouldn't add an invariant 👍

edleeks87 (approved):

grabkowski:
is there a new feature that is implemented with the refactor? If not, then the commit should be a refactor, as it would not affect developers that are using Carbon.

package-lock.json
```diff
@@ -8661,7 +8667,7 @@
     "ansicolors": {
       "version": "0.3.2",
       "resolved": "https://registry.npmjs.org/ansicolors/-/ansicolors-0.3.2.tgz",
-      "integrity": "sha512-QXu7BPrP29VllRxH8GwB7x5iX5qWKAAMLqKQGWTeLWVlNHNOpVMJ91dsxQAIWXpjuW5wqvxu3Jd/nRjrJ+0pqg==",
+      "integrity": "sha1-ZlWX3oap/+Oqm/vmyuXG6kJrSXk=",
```
grabkowski, line 8670: is that change expected here? You are only updating the TypeScript types in this PR.

package.json
```diff
@@ -89,6 +89,7 @@
     "@storybook/theming": "^6.4.18",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^12.1.3",
+    "@types/crypto-js": "^4.1.1",
```
grabkowski, line 92: have you thought of adding that in a separate commit?

src/components/portrait/portrait-initials.component.tsx
```diff
@@ -1,23 +1,37 @@
...
+  shape = "square",
   darkBackground,
   alt,
-  ...rest
-}) => {
-  const [cachedImageDataUrl, setCachedImageDataUrl] = useState();
+}: PortraitInitialsProps) => {
+  const [cachedImageDataUrl, setCachedImageDataUrl] = useState<string | null>(
+    null
+  );
```
grabkowski, line 34: I think that you could also refactor `setCachedImageDataUrl(null);` to `setCachedImageDataUrl("");`, then you could drop the null type here. IMO as we are refactoring to TypeScript, we should avoid changing types in our variables.

```suggestion
  const [cachedImageDataUrl, setCachedImageDataUrl] = useState<string>();
```

package-lock.json
mkrds: Well, that comes from installation of `@types/crypto-js` package

edleeks87 (dismissed):

package-lock.json
grabkowski: but that should be done by `crypto-js` package, and this package contains just TypeScript types for it

edleeks87 (approved):

grabkowski (approved):

ZhuoyuJin (approved):

PR #5350, opened 2022-08-03
feat(sidebar): add padding props for customising content container (FE-4458)

### LINKED ISSUE #4494: Add padding props to Sidebar
### Desired behaviour

I would like to customise the padding within a sidebar component so that the content within can span across the entire width of the sidebar.

### Current behaviour

It is not possible to change the padding within the sidebar without writing custom CSS.

### Suggested Solution

_No response_

https://codesandbox.io/s/floral-pine-hd3zt?file=/src/index.js

_No response_

### Confidentiality

### LINKED ISSUE #5327: Sidebar colour
### Desired behaviour

Ability change background colour of sidebar component to a chosen colour so it matches webpage colour scheme
Screenshot 2022-07-22 at 13 26 45

### Current behaviour

Currently the Sidebar component can only be the default white. The space cannot be completely filled with a coloured box because the padding always leaves a white space.

### Suggested Solution

Pass in a desired colour the samemway you can for other componets eg. `backgroundColor="#000000"`.
Or make it possible to set padding to 0 so the space can be filled with. coloured box eg. `p={0}`.

_No response_

_No response_

### Confidentiality

Parsium:
resolves #4494

### Proposed behaviour

- Added styled system padding props to `Sidebar`, which are applied to the sidebar content container. Updated PropTypes, types in `d.ts` file and storybook prop table
- Define default content padding using design tokens:
- Added new story to demonstrate that custom padding can be added
- Minor refactors:
   - remove `defaultProps`
   - rename inner style component to fit convention
   - modify prop types of `size` and `position` props - so all of their possible values are displayed clearly in storybook, as opposed to `SIDEBAR_SIZES` and `SIDEBAR_ALIGNMENTS` 

### Current behaviour

- Cannot pass custom padding props to `Sidebar`
- Padding of content is not defined using design tokens

### Checklist

#### QA

### Additional context

The content padding styles have been updated with design tokens to match the DS design: 

Screenshot 2022-08-03 at 14 15 36

### Testing instructions

- Check new story `custom-padding-around-content` to see new behaviour.
- Ensure no regressions have been introduced as a result of this change.
- Double check for any ambiguity in storybook docs for `Sidebar` as a result of this change.


mkrds (dismissed):

mkrds (approved):

src/components/sidebar/sidebar.stories.mdx
```diff
@@ -87,6 +88,30 @@ with all the UI pass `enableBackgroundUI={ true }` to the component
...
+### Custom padding around content
+
+<Canvas>
+  <Story name="custom padding around content">
+    {() => {
+      const [isOpen, setIsOpen] = useState(isOpenForChromatic);
+      return (
+        <>
+          <Button onClick={() => setIsOpen(true)}>Open sidebar</Button>
+          <Sidebar open={isOpen} onCancel={() => setIsOpen(false)} padding={0}>
```
robinzigmond, line 100: It might be better to use the `p` prop rather than `padding` here - they have the same effect (I believe, from the Styled System docs, that they are aliases of each other), but only `p` is included in our props tables, and it might be confusing to use a prop which is absent from the props table on the same page!

(I'd be happy with an alternative fix to `styled-system-props.js` in `/.storybok/utils` to include the `padding` prop as well though, I think similar changes could be made for margin etc. I assume it might have been a deliberate decision to leave the long versions of the names out though?)

src/components/sidebar/sidebar.stories.mdx
```diff
@@ -42,7 +43,7 @@ Useful to perform an action in context without navigating the user to a separate
 
 Useful if the user might need to refer back to the page behind the sidebar. For example, the user could select from a list of clients on the page, with detailed information on the client loading in the sidebar.
 
-Several pre-set widths are available, and the height is always 100%. It’s best to avoid sidebars that are taller than the user’s viewport height, as a scroll would hide some information. Typical user viewport heights can be as little as 650 pixels.
+The height of the sidebar is always 100%. It’s best to avoid sidebars that are taller than the user’s viewport height, as a scroll would hide some information. Typical user viewport heights can be as little as 650 pixels.
```
robinzigmond, line 46: I'm not sure why you've removed the part of the docs about several pre-set widths being available? As far as I can tell from the code, there is still a `size` prop which takes one of the predefined size strings in SIDEBAR_SIZES. I don't see why adding padding props should change this, and no notes in your PR about anything to do with this?

src/components/sidebar/sidebar.stories.mdx
Parsium: Good point, perhaps the util fix is something for us to look into in the future, but will make this change now 👍🏼

Parsium: Partially related to the prop table - I will update the proptypes for the `size` and `position` props - since I've noticed the permitted values for both props aren't being displayed for the types.

robinzigmond: great stuff - I noticed that too but decided it wasn't relevant to mention in this PR!

robinzigmond (dismissed):

mkrds (approved):

edleeks87 (approved):

mkrds (approved):

ZhuoyuJin (approved):

PR #5354, opened 2022-08-04
fix(select): ensure aria-activedescendant is always valid

### LINKED ISSUE #5289: Select - Axe Violation - ARIA attributes must conform to valid values
### Current behaviour

Select component has an axe violation when a value is selected, ensures all ARIA attributes have valid values.
image

### Expected behaviour

No accessibility violations should occur.

https://carbon.sage.com/?path=/story/select--default-story

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

Latest

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
make the Select List always part of the DOM, and hide it with CSS instead. This allows the component
to ensure that the guids provided as IDs to the Option children are always stable, so that the
aria-activedescendant attribute of the Select Textbox always refers to the ID of an existing
element. Also make the aria-controls attribute of the textbox always present so that screenreaders
can know the list of options available when focused on the textbox.

fix #5289

### Proposed behaviour

The Select textbox should always have an aria-controls attribute that refers to an option list that is always in the DOM, and is hidden and shown with CSS - similar to the ARIA Working Practices example here https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html

Further, the IDs of each Option element should be stable (assuming no changes to the available options), so that the aria-activedescendant attribute, which changes as focus is moved among options with the keyboard as well as when selection is made with the mouse, always holds the actual ID of an element that is in the DOM (even if hidden).

There has been a further small change to the FilterableSelect component - this was purely to fix a failing cypress test which seemingly only happened with this PR and proved to be due to a race condition when handleGlobalClick runs at the same time as an update to the `value` prop in the controlled case.

### Current behaviour

The Select option list is removed from the DOM when closed, so that when selecting an option and closing the list, the aria-activedescendant is invalid, causing Axe to flag an error.

Further, even when reopening the select so that the options are in the DOM again, the IDs are newly randomly-generated guids so the existing aria-activedescendant is still not referring to an element that's present in the DOM.

### Checklist

#### QA

### Testing instructions

Test Stories for each type of Select (Simple select, FilterableSelect, MultiSelect). Select an option, close the select, and run Axe devtools. Reopen the select (without choosing a new option) and run again. Ensure that no errors are flagged (or at least not one saying ARIA attribute values are invalid!)

Also test with screenreaders across all supported screenreader/browser combinations, to ensure the Selects are usable - or at least there are no issues introduced here which didn't exist before.

Regression testing should also be done with FilterableSelect to ensure that the change I've made to the controlled case (see under "proposed behaviour" above, final paragraph) does not cause any unforeseen issues.


src/__internal__/popover/popover.component.tsx
```diff
@@ -45,6 +45,8 @@ export interface PopoverProps {
   disablePortal?: boolean;
   // Reference element, children will be positioned in relation to this element - should be a ref shaped object
   reference: React.RefObject<HTMLElement>;
+  // Determines if the popover is currently open/visible or not. Defaults to true.
+  isOpen?: boolean;
 }
 // TODO: Remove TempModalContext after modal has been converted to TS
 type TempModalContext = {
```
edleeks87, line 52: While you're updating this file can you update this to use the `ModalContext` as the refactor of `Modal` is now in master?

src/components/select/filterable-select/filterable-select.spec.js
```diff
@@ -112,22 +112,22 @@ describe("FilterableSelect", () => {
 
     it("the SelectList should not be rendered", () => {
       wrapper.find("input").simulate("focus");
-      expect(wrapper.find(SelectList).exists()).toBe(false);
+      expect(wrapper.find(SelectList).prop("isOpen")).toBe(false);
```
edleeks87, line 115: For these tests let's assert the styling is or isn't applied (using `toHaveStyleRule`) here rather than the prop value

src/components/select/multi-select/multi-select.spec.js
```diff
@@ -39,25 +39,25 @@ describe("MultiSelect", () => {
     describe("and that element is part of the Select", () => {
       it("then the SelectList should be open", () => {
         wrapper.find("input").simulate("focus");
-        expect(wrapper.find(SelectList).exists()).toBe(true);
+        expect(wrapper.find(SelectList).prop("isOpen")).toBe(true);
```
edleeks87, line 42: Same comment about using `toHaveStyleRule` instead of the prop value

src/components/select/select-list/select-list.component.js
```diff
@@ -319,29 +327,35 @@ const SelectList = React.forwardRef(
...
+      }
 
       return function cleanup() {
-        window.removeEventListener(keyboardEvent, handleGlobalKeydown);
-        listElement.removeEventListener("scroll", handleListScroll);
+        if (isOpen) {
+          window.removeEventListener(keyboardEvent, handleGlobalKeydown);
+          listElement.removeEventListener("scroll", handleListScroll);
+        }
       };
```
grabkowski, line 357: the same issue as with scroll blocking above, the event listeners should be removed regardless of the `isOpen` state.

src/components/select/select-list/select-list.component.js
```diff
@@ -319,29 +327,35 @@ const SelectList = React.forwardRef(
...
     useEffect(() => {
       const keyboardEvent = "keydown";
       const listElement = listRef.current;
 
-      window.addEventListener(keyboardEvent, handleGlobalKeydown);
-      listElement.addEventListener("scroll", handleListScroll);
+      if (isOpen) {
+        window.addEventListener(keyboardEvent, handleGlobalKeydown);
+        listElement.addEventListener("scroll", handleListScroll);
+      }
```
grabkowski, line 350: is there a reason to not handle it in the `handleListScroll` method?

src/components/select/select-list/select-list.component.js
```diff
@@ -82,12 +83,16 @@ const SelectList = React.forwardRef(
...
-      blockScroll();
+      if (isOpen) {
+        blockScroll();
+      }
 
       return () => {
-        allowScroll();
+        if (isOpen) {
+          allowScroll();
+        }
```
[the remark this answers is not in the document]
robinzigmond: maybe? As I recall, I had to stop `blockScroll` being called if `isOpen` is false, otherwise scroll gets blocked when it shouldn't be (took me quite a while to figure out why the storybook page seemed to have broken completely :D). This should mean it's safe to not call `allowScroll` when tidying up in this case, but if there's no harm in calling it when scroll isn't blocked already then I can add it - I'll check.

src/components/select/select-list/select-list.component.js
```diff
@@ -258,30 +258,32 @@ const SelectList = React.forwardRef(
...
-          document.activeElement === listActionButtonRef.current;
-
-        if (key === "Tab" && listActionButton) {
-          handleActionButtonTab(event, isActionButtonFocused);
-        } else if (key === "Tab" || key === "Escape") {
-          onSelectListClose();
-        } else if (key === "Enter" && !isActionButtonFocused) {
-          event.preventDefault();
-          const currentOption = childrenList[currentOptionsListIndex];
+        if (isOpen) {
```
grabkowski, line 261: instead of wrapping the whole block you could just not handle that when the element is not open:
```
if (!isOpen) {
  return;
}
```

edleeks87 (dismissed):

grabkowski (dismissed):

divyajindel (dismissed):

edleeks87 (approved):

grabkowski (dismissed):

divyajindel (dismissed):

grabkowski (approved):

edleeks87 (dismissed):

divyajindel (approved):

grabkowski (approved):

edleeks87 (approved):

divyajindel (approved):

PR #5356, opened 2022-08-04
fix(toast): ensure toast component has first focus when opened - FE-5281

### LINKED ISSUE #5320: Toast - should focus container rather than close button when opened
### Desired behaviour

When opening a Toast, the focus should be placed on the toast container so that screenreader users hear the toast message. The container should not be focusable again with the keyboard and pressing tab should take the focus to the close button.

Note: even where the Toast does not have a close button, the focus should go to the toast container so that screenreader users have the message read out.

### Current behaviour

Focus goes to the close button where present, since Carbon version 109.3.0. In versions prior to that, and in all versions if the toast has no close button, the toast is not focused at all. When focus goes to the close button, screenreader users just hear "close button" but no information on what they are closing.

### Suggested Solution

As under "desired behaviour". Note: similar behaviour is already implemented in Dialog, DialogFullScreen, Sidebar and MenuFullScreen.

https://carbon.sage.com/?path=/docs/toast--dismissible

cc @tempertemper who requested it this morning

### Confidentiality

DipperTheDan:
fixes #5320 

### Proposed behaviour

Toast component should be first component to be focused

### Current behaviour

Currently Toast is not focusable

toast_original

toast-dissmissable-original

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

Test that the Toast stories in Storybook work as expected with the new functionality and check for any regressions

https://codesandbox.io/s/carbon-quickstart-forked-j2p39d?file=/src/index.js


src/components/toast/toast.style.js
```diff
@@ -38,7 +38,10 @@ const ToastStyle = styled(MessageStyle)`
     position: relative;
     margin-right: ${isCenter ? "auto" : "30px"};
   `}
- 
+
```
edleeks87, line 41: I'm not sure we want a outline here, can we check with UX to confirm?

src/components/toast/toast.component.js
```diff
@@ -75,12 +82,22 @@ const Toast = React.forwardRef(
       if (onDismiss && !disableAutoFocus) {
```
edleeks87, line 82: I think this can be condensed to 

```
     if (!disableAutoFocus) {
        if (open) {
          focusedElementBeforeOpening.current = document.activeElement;
          toastContentNodeRef.current?.focus();
          setHasBeenFocused(true);
        } else if (onDismiss && focusedElementBeforeOpening.current) {
          focusedElementBeforeOpening.current.focus();
          focusedElementBeforeOpening.current = undefined;
        }
     }
 ```

Comment (edleeks87) [reply src/components/toast/toast.component.js:82]: But maybe we don't actually want it to focus without the `onDismiss` at all 🤔

File: src/components/toast/toast.style.js (hunk 1 above)
Comment (DipperTheDan) [reply src/components/toast/toast.style.js:41]: Will do.

Comment (robinzigmond) [reply src/components/toast/toast.style.js:41]: for what it's worth we don't currently have an outline on Dialog etc. when their containers are autofocused after opening - we should try for consistency unless UX say otherwise

File: src/components/toast/toast.component.js (hunk 2 above)
Comment (robinzigmond) [reply src/components/toast/toast.component.js:82]: no, I think we do want it to focus even if there's no `onDismiss` unless of course `disableAutoFocus` is set. I put much of this original code in in order to focus the close button, and the only reason to check for `onDismiss` there was because that prop is (somewhat confusingly) what controls the presence or absence of the close button. I'm inclined to think, now we're going to focus the container, it should focus regardless of dismiss button or not.

Mind you if it times out then I suspect focus will be lost entirely when it disappears, which isn't great. Might be another question for UX... (I discussed some of this with Harpal before my changes, albeit mostly around how Toast should interact with a Dialog if launched from there)

File: src/components/toast/toast.component.js [hunk 3]
```diff
@@ -128,6 +145,7 @@ const Toast = React.forwardRef(
             {...tagComponent(restProps["data-component"] || "toast", restProps)}
             {...toastProps}
             ref={toastContentNodeRef}
+            tabIndex={hasBeenFocused ? -1 : 0}
```
robinzigmond, line 148: strictly speaking this ternary isn't needed - tabIndex -1 still allows focus to be given programatically, while not including it in the natural keyboard tab order. However I recall there were issues with cypress tests working with tabIndex -1 - if you haven't already, it might be worth testing with just `tabIndex="-1"` here and leaving it with that if there are no issues, as that's much simpler/better code.

A further benefit would be that it we don't use `hasBeenFocused` to control a rendered prop then it can become a ref rather than state, and cause fewer rerenders. I suspect Cypress will say "no" but worth a try if you haven't already done so (which you may well have, in which case ignore this!)

src/components/toast/toast.component.js
DipperTheDan: Cheers for the bits of context @robinzigmond 
I'll ask UX about the timeout thing too and we'll see what they come back regarding how focus should behave.

src/components/toast/toast.component.js
DipperTheDan: I on purposely avoided doing any tabIndex tests in Cypress as there is a known bug with the `cypress-plugin-tab` package with trying to test `tabIndex={-1}`. Hence, I used the CSS value for testing focus. Not ideal but best I could think of without testing the tabIndex directly.

edleeks87 (dismissed):

robinzigmond (dismissed):

harpalsingh:
I know @tempertemper is away this week, so I have done the accessibility review by testing the voiceover feature too.

tempertemper:
I think the focus was moved from the close button to the toast itself, wasn't it? That way screen reader users get read the message rather than "Close" when the toast pops-up. @robinzigmond was looking at this a short while ago.

edleeks87 (dismissed):

src/components/toast/toast.component.js
```diff
@@ -128,6 +134,10 @@ const Toast = React.forwardRef(
             {...tagComponent(restProps["data-component"] || "toast", restProps)}
             {...toastProps}
             ref={toastContentNodeRef}
+            tabIndex={-1}
+            role="status"
```
robinzigmond, line 138: I might be missing something here but why does this need to be a live region? The content should already be announced by a screenreader when it appears, due to the auto focus.

src/components/toast/toast.component.js
```diff
@@ -128,6 +134,10 @@ const Toast = React.forwardRef(
...
+            aria-atomic="true"
+            onKeyDown={handleKeyDown}
```
robinzigmond, line 140: I'm curious why we need a custom `onKeyDown` here to focus the close icon when tabbing from the container. For one thing, this should work automatically - I assume it doesn't since you've added this, but I'd rather figure out what the underlying problem is with that and solve it rather than just hardcoding the specific behaviour we need.

Especially because this onKeyDown is now on the whole toast container so will it not mean tabbing from anywhere with in will focus the close icon even when it isn't supposed to? Eg if you're already focused on the close icon won't this then refocus it, therefore stopping keyboard users navigating away...

src/components/dialog/dialog.stories.mdx
```diff
@@ -406,6 +406,7 @@ set the padding to 0. Please see the [table below](#props) for more iformation a
 Using the `focusableContainers` prop allows elements that are not DOM children of the Dialog to be reachable with the keyboard while the Dialog is open.
 This may occasionally be useful with things like Toasts where they persist on the page until dismissed by clicking a close icon.
 
+**Please note that due to a known bug with `VoiceOver` and `Safari`, the `Toast` messages will not be read out in this example.**
```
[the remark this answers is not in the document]
DipperTheDan: I figured I'd put this message in there to prevent us getting Github issues logged for stuff we already know about. It's easily removed if you think it isn't needed.

src/components/toast/toast.component.js
DipperTheDan: I added this after reading this article on Toast best practices https://adrianroselli.com/2020/01/defining-toast-messages.html#NeededFeatures
`alert` and `log` don't fit the uses of toast as `alert` will interrupt anything the user is doing and read out in an assertive style. I can probably remove the `aria-atomic` as that is implicit with `status`.

src/components/toast/toast.component.js
DipperTheDan: This was added because when `Toast` is re-opened from a `Dialog` [like in this example](https://carbon.sage.com/?path=/docs/dialog--other-focusable-containers), the close icon would be the first item to get focus. I couldn't get it to work any other way than this but I'm open to other potential fixes if you have any?

src/components/dialog/dialog.stories.mdx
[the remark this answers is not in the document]
robinzigmond: I still say leave it out - especially because it'll need removing once the fix is in (which I've currently got a PR in for!). It's unlikely a bug gets logged in the short time between the two PR's getting merged, and if it does, well there's at least 2 of us who will know and can prevent duplicate tickets getting logged!

src/components/toast/toast.component.js
robinzigmond: I still say remove the `role="status"` as well as the `aria-atomic` (which only applies to live regions anyway) - I think you've misunderstood that part of the article. It's talking about the *lots* of different ways of implementing what are vaguely known as "toast" notifications. Some simply pop some text up at the top/bottom of the screen, and in that case a live region is probably the best way. We're using what is effectively a dialog, which is given focus when it appears, so screenreaders already pick it up. In addition, the article underneath the part you linked to says `Generally live regions must exist on the page at render`, which means they won't even work with a dialog that only pops into existence at the point the toast message appears.

Unless I'm missing something, and you experienced specific issues when testing with a screenreader which this solved, I say get rid of it.

src/components/toast/toast.component.js
```diff
@@ -130,6 +138,8 @@ const Toast = React.forwardRef(
             {...tagComponent(restProps["data-component"] || "toast", restProps)}
             {...toastProps}
             ref={toastContentNodeRef}
+            tabIndex={tabIndex}
+            onBlur={() => setTabIndex(undefined)}
```
robinzigmond, line 142: really the tabIndex should be -1. That can cause issues with Cypress tests though, which was why in Dialog etc. I changed it to open with a tabIndex of 0 that is removed on blur (to prevent the container being refocusable with the keyboard).

Does setting tabIndex to -1 definitely cause issues here with Toast? If not then I'd prefer using that. If it does then we need to ensure that the `tabIndex` is only set to 0 when `disableAutoFocus` is `false`, to ensure that the user can't tab into the toast container manually. (If autofocus is disabled there is no reason to want to programatically focus the container either.)

edleeks87: I'll investigate and update if I don't find any issues. Good point about the outo focus behaviour, I'll make sure the guard is added if needed

src/components/toast/toast.spec.js
```diff
@@ -249,9 +263,9 @@ describe("Toast", () => {
           wrapper = mount(
             <Toast open onDismiss={onDismiss} disableAutoFocus />
           );
-          const icon = wrapper.find("[data-element='close']").first();
-
-          expect(icon).not.toBeFocused();
+          const toast = wrapper.find(ToastStyle);
```
nineteen88, line 266: Is this actually testing what the test says it is testing? The ToastStyle seems to be the styled component for the Toast itself but the Test wants to make sure close icon isn't focussed. Or are we able to safely assume that if there's no tabindex set on Toast Message then the close button/icon wouldn't be focussed?

src/components/toast/toast.spec.js
```diff
@@ -248,9 +263,9 @@ describe("Toast", () => {
...
+          expect(toast.getDOMNode().hasAttribute("tabIndex")).toBe(false);
```
nineteen88, line 267: Is this actually testing what the test says it is testing? The ToastStyle seems to be the styled component for the Toast itself but the Test wants to make sure close icon isn't focussed. Or are we able to safely assume that if there's no tabindex set on Toast Message then the close button/icon wouldn't be focussed?

edleeks87: Ah good spot, I've forgotten to update the test description, the new behaviour is that it focuses the Toast container on open rather than the close Icon. I'll get this updated asap thanks

src/components/toast/toast.component.js
```diff
@@ -44,6 +50,8 @@ const Toast = React.forwardRef(
 
     const focusedElementBeforeOpening = useRef();
 
+    const [tabIndex, setTabIndex] = useState(!disableAutoFocus && 0);
+
```
robinzigmond, line 54: Thanks for making the change but I don't like this line I'm afraid. This will set the initial `tabIndex` to `false` when `disableAutoFocus` is true, which is meaningless in HTML. (And likely gets ignored by the browser, so this will still work, but it's not something we should do.)

I would do `const [tabIndex, setTabIndex] = useState(disableAutoFocus ? 0 : undefined)` but happy to hear alternatives - but I do think we should ensure the value is `undefined` when `disableAutoFocus` isn't set.

src/components/toast/toast.component.js
robinzigmond, line 142: not a big deal but I'd probably prefer to check `disableAutoFocus` and only set the tab index when that's `false`. I don't mind leaving it unconditional though given that when it's true the tabindex should already be undefined.

edleeks87: Yeah as far as I'm aware it will not set it if it is already undefined but I'll update anyway as there's no need for either prop being passed if `disableAutoFocus` is true

src/components/toast/toast.component.js
edleeks87: Valid point, I'd misunderstood how non-integer values would be treated

nineteen88 (approved):

robinzigmond (approved):

nineteen88 (approved):

DlgSHi (approved):

PR #5364, opened 2022-08-08
build(node, npm): remove constraint on node and npm versions, update gh actions to use v16.16 npm FE-5297

edleeks87:
### Proposed behaviour

Removes `engines` and `engineStrict` properties from `package.json` file. Update GH actions to use
npm `v16.16.0`. Update `dev-environment-setup` docs and `check_node_version` has been removed as
`node` version should not matter

### Current behaviour

Constrains `node` and `npm` versions to `fermium/lts`, actions also use that version, node version is checked when `npm start` is run

### Checklist

#### QA

### Testing instructions


.github/workflows/chromatic-push.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-node@v1
         with:
-          node-version: ">=14.16.0 14"
+          node-version: ">=16.16.0 16"
```
mkrds, line 22: Why not target all versions of v16 node?
```suggestion
          node-version: ">=16"
```

edleeks87: I decided target the LTS of v16 as wasn't sure there was much point using a version below it. I also locked to v16 as it was previously for v14 so that it doesn't use a v17/v18 node version for now

scripts/check_node_version.js
```diff
@@ -1,13 +0,0 @@
-const semver = require("semver");
```
jamime, line 1: Keep this file, its to prevent carbon contributors using an old version of node. We still want to support node v14 while it's in maintenance lts.

.github/workflows/ci.yml
```diff
@@ -20,6 +20,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: ">=14.16.0 14"
+           node-version: ">=16.16.0 16"
```
jamime, line 23: Let's change the `npm test` command to run on both node 14 and node 16 (The current active and maintenance lts versions).

image

jamime (changes requested):

.github/workflows/chromatic-push.yml
jamime: Im fine for us to use `">=16.16.0 16"`, the current active lts.

scripts/check_node_version.js
edleeks87: Ah my bad I thought we were saying any version is now ok, I'll add it back in 👍

jamime: any version >= 14 should work.

.github/workflows/ci.yml
```diff
@@ -20,6 +20,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: ">=14.16.0 14"
+           node-version: ">=14.16.0 14 || >=16.16.0 16"
```
jamime, line 23: This needs to run as a matrix i.e we should run the tests on v14 and on v16
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

jamime (changes requested):

.github/workflows/ci.yml
edleeks87: @jamime I've added the matrix in but it's added an additional `test` check and I can't work out why, I don't see any syntax issues and haven't been able to find anything that explains why online.

image

jamime (approved):

mkrds (approved):

PR #5365, opened 2022-08-10
refactor(button-toggle): convert to typescript (FE-4771)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/button-toggle/button-toggle-input.component.tsx
```diff
@@ -33,23 +58,6 @@ const ButtonToggleInput = React.forwardRef((props, forwardRef) => {
...
-  /** Value for the input */
-  value: PropTypes.string,
-  /** Callback triggered by change event on the input. */
-  onChange: PropTypes.func,
-  /** Callback triggered by blur event on the input. */
-  onBlur: PropTypes.func,
-  /** Callback triggered by focus event on the input. */
-  onFocus: PropTypes.func,
-};
+ButtonToggleInput.displayName = "ttonToggleInput";
```
mkrds, line 61:
```suggestion
ButtonToggleInput.displayName = "ButtonToggleInput";
```

src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -28,29 +49,27 @@ const ButtonToggle = (props) => {
...
+  const inputRef = useRef<HTMLInputElement>(null);
   let icon;
 
   if (buttonIcon) {
     icon = (
       <ButtonToggleIcon
-        aria-hidden
         buttonIcon={buttonIcon}
         buttonIconSize={buttonIconSize}
-        hasNoContent={!children}
```
mkrds, line 46: Why are both `aria-hidden` and `hasNoContent` props removed?

src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -1,22 +1,43 @@
...
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-element"?: string;
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-role"?: string;
+  /** Set the default value of the Group if component is meant to be used as uncontrolled. */
+  defaultChecked?: boolean;
+  /** Remove spacing from between buttons. */
+  grouped?: boolean;
+  /** Callback triggered by click event on the input. */
+  onClick?: (ev: React.MouseEvent<HTMLInputElement>) => void;
```
Parsium, line 32:
```suggestion
```
I don't think this prop is used anywhere in the component?

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -40,7 +41,18 @@ const StyledButtonToggleContentWrapper = styled.div`
...
-const StyledButtonToggleLabel = styled.label`
+export interface StyledButtonToggleLabelProps {
+  /** buttonIcon to render. */
+  buttonIcon?: IconType;
+  /** Sets the size of the buttonIcon (eg. large) */
+  buttonIconSize?: ButtonToggleIconSizes;
+  /** Disable all user interaction. */
+  disabled?: boolean;
+  /** ButtonToggle size */
+  size: "small" | "medium" | "large";
```
Parsium, line 52:
```suggestion
  /** ButtonToggle size */
  size?: "small" | "medium" | "large";
```

Since consumers may not be specifying the `size` prop (even though we set its default value ourselves in `button-toggle.component.tsx`) the prop's type should be optional to prevent consumers receiving an error.

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -40,7 +41,18 @@ const StyledButtonToggleContentWrapper = styled.div`
   height: 100%;
 `;
 
-const StyledButtonToggleLabel = styled.label`
+export interface StyledButtonToggleLabelProps {
+  /** buttonIcon to render. */
+  buttonIcon?: IconType;
+  /** Sets the size of the buttonIcon (eg. large) */
+  buttonIconSize?: ButtonToggleIconSizes;
```
Parsium, line 48:
```suggestion
  buttonIconSize?: `small` | `large`;
```
Since `button-toggle-types.d.ts` is only one line, in my opinion I think we could remove the file and include the type values here unless you think otherwise

src/components/button-toggle/button-toggle.component.tsx
grabkowski: `aria-hidden` could be applied directly to the `Icon` instead of passing it down from here (please check the `button-toggle-icon.component.tsx` file).  As for `hasNoContent` - this prop was redundant, all it was doing was that:
https://github.com/Sage/carbon/pull/5365/files#diff-9c787793b3c94e782a732088653da4f22cc3491425db3d5a70eb601ff1ce7d85L114-L120

src/components/button-toggle/button-toggle.style.ts
grabkowski: this is an internal component, the default size is `medium` and it's being set here:
https://github.com/Sage/carbon/pull/5365/files#diff-3f6aa630f9b196e239402e9dd6a2a350de858f8f653eb9e092430a971d60c695R52

mkrds (dismissed):

src/components/button-toggle/button-toggle-test.stories.tsx
```diff
@@ -29,42 +32,34 @@ export const ButtonToggleStory = ({
...
-    {ButtonToggleStory.bind({})}
-  </Story>
-</Canvas>
+Default.story = {
+  name: "default",
+  args: {
+    children: "Options",
+    childrenSpecialCharacters: undefined,
+  },
+};
```
Parsium, line 65:
```suggestion
Default.storyName = "default";
Default.args = {
    children: "Options",
    childrenSpecialCharacters: undefined,
};
```
Apparently the `storyFunction.story` syntax has been deprecated in Storybook v6: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#hoisted-csf-annotations

Parsium (approved):

mkrds (dismissed):

Parsium (dismissed):

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -111,13 +124,18 @@ const iconFontSizes = {
...
+  hasContent?: boolean;
+}
+
+const StyledButtonToggleIcon = styled.div<StyledButtonToggleIconProps>`
+  ${({ hasContent }) => hasContent && `margin-right: 8px;`}
+  ${({ hasContent }) => {
+    console.log(hasContent);
+    return "";
+  }}
```
Parsium, line 138:
```suggestion
  ${({ hasContent }) => hasContent && `margin-right: 8px;`}
```

src/components/button-toggle/button-toggle-icon.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+   */
+  buttonIcon: IconType;
+  /** Sets the icon in the disabled state */
+  disabled?: boolean;
+}
+
+const ButtonToggleIcon = ({
+  buttonIcon,
+  buttonIconSize,
+  disabled,
```
Parsium, line 22:
```suggestion
const ButtonToggleIcon = ({
  buttonIcon,
  buttonIconSize,
  disabled,
  hasContent,
```
I think we need to surface `hasContent` as a prop, otherwise we could make the `margin-right` css rule for `StyledBUttonToggleIcon` default behaviour?

src/components/button-toggle/button-toggle-icon.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: ButtonToggleIconProps) => (
+  <StyledButtonToggleIcon buttonIconSize={buttonIconSize} hasContent>
```
Parsium, line 24:
```suggestion
const ButtonToggleIcon = ({
  buttonIcon,
  buttonIconSize,
  disabled,
  hasContent = true,
}: ButtonToggleIconProps) => (
  <StyledButtonToggleIcon buttonIconSize={buttonIconSize} hasContent={hasContent}>
```
I think we need to surface hasContent as a prop, otherwise we could make the margin-right css rule for StyledButtonToggleIcon default behaviour?

grabkowski: content means the children of button toggle component, so the prop is already there. The perfect solution would be to remove the ButtonTogleIcon and let the users set is as content themselves.

src/components/button-toggle/button-toggle.style.ts
grabkowski: oops :)

Parsium (approved):

DlgSHi (approved):

PR #4949, opened 2022-03-09
refactor: convert Heading to functional component and remove deprecated components

Parsium:
**This PR introduces multiple breaking changes in one to limit number of releases.**

### Proposed behaviour

- Rewrite `Heading` to a functional component.
- Move `RadioButtonMapper` to internal components directory
- Remove following deprecated components:
    - `AppWrapper`
    - `MountInApp`
    - `MultiStepWizard`
    - `ScrollableList`
    - `SubmenuBlock`
- Remove following deprecated props:
    - `as` of `Button`
    - `destructive` of `Confirm`
    - `as` of `Tile`
    - `as` of `Toast`
    - `as` of `SplitButton`
    - `as` of `MultiActionButton`
    - `stickyPosition` and `stickyOffset` of `NavigationBar`

### Current behaviour

- `Heading` is currently class-based.
- `RadioButtonMapper` can be imported by consuming apps - even though it's only intended to be used internally.
- Components and props that have been deprecated for some time are still part of Carbon.

### Checklist

- [x] Commits follow our style guide
- [x] Related issues linked in commit messages if required
- [x] Screenshots are included in the PR if useful
- [x] All themes are supported if required
- [x] Unit tests added or updated if required
- [x] Cypress automation tests added or updated if required
- [x] Storybook added or updated if required
- [x] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [x] Typescript `d.ts` file added or updated if required

#### QA

- [ ] Tested in CodeSandbox/storybook
- [ ] Add new Cypress test coverage if required
- [ ] Carbon implementation matches Design System/designs
- [ ] UI Tests GitHub check reviewed if required

### Testing instructions

- Double-check stories for any major regressions.
- Check for any remaining imported/exports that need removing or replacing

PR #4573, opened 2021-11-18
fix(select): render wrapped text correctly in SelectList - FE-4044

DipperTheDan:
Wrapped text should render at the correct height in the SelectList. Adding listRef.current to the
height calculation enables this to happen as expected.

fixes #4489

### Proposed behaviour
- Text in the SelectList should wrap correctly and apply the correct height the select option.

Screenshot 2021-11-18 at 11 34 30

### Current behaviour
- Select option height is not being calculated correctly.

Screenshot 2021-11-18 at 11 31 00

### Checklist

- [x] Commits follow our style guide
- [x] Related issues linked in commit messages if required
- [x] Screenshots are included in the PR if useful
- [ ] All themes are supported if required
- [x] Unit tests added or updated if required
- [x] Cypress automation tests added or updated if required
- [x] Storybook added or updated if required
- [ ] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [ ] Typescript `d.ts` file added or updated if required

#### QA

- [ ] Tested in CodeSandbox/storybook
- [x] Add new Cypress test coverage if required
- [ ] Carbon implementation matches Design System/designs
- [ ] UI Tests GitHub check reviewed if required

### Additional context
N/A

### Testing instructions

The following CodeSandbox is an example of the broken behaviour.
You can see the new behaviour by looking at the version in the comment by `codesandbox[bot]`.

Use the codesandbox build from the Source `Issue #4489`

Issue #4489, opened 2021-10-14
Select - When an option with text that wraps onto a new line part of the text is cut off


philearley:
### Current behaviour

When you open a select component where one of the options wraps onto a new line, the select component cuts off some of this text.

### Expected behaviour

The options section should expand to contain the content.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/festive-snow-v97zf?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-22181

### Suggested Solution

Modifying the Carbon code locally, we resolved the UI by adding listRef.current into the dependencies array here: https://github.com/Sage/carbon/blob/master/src/components/select/select-list/select-list.component.js#L312

### Carbon Version

91.0.2

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

MacOS

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

PR #5367, opened 2022-08-10
fix(select): ensure select list always has correct height to fit content

### LINKED ISSUE #5341: Select component cuts off option text when it wraps onto a new line
### Current behaviour

Select component cuts off option text when it wraps onto a new line

image

### Expected behaviour

Select component should display option text when it wraps onto a new line without cutting it off.

https://codesandbox.io/s/select-options-height-resize-uy2shp?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-38045

### Suggested Solution

_No response_

### Carbon Version

109.1.2

### Design Tokens Version

2.40.0

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
Use a callback ref to ensure that the the code that calculates the list height is ran whenever the
DOM node is attached.

fix #5341

### Proposed behaviour

Select list should always be the correct height to show all its options, with the correct padding/margin, no matter how many times the text wraps.

### Current behaviour

In some circumstances (needs few enough options for there not to be a scrollbar), options with long text that wraps cause the list height to be too small and some parts of some options therefore are not visible.

### Checklist

#### QA

### Additional context

There's a bit of history behind this bug/change that reviewers may find useful.

This was first reported as issue #4489.
Fixed in [this PR](https://github.com/Sage/carbon/pull/4573) by adding `listRef.current` to the `useLayoutEffect`'s dependency array - this works because the height needs to be recalculated when the select list container appears in the DOM, which wasn't happening previously.
This fix was removed in [this PR](https://github.com/Sage/carbon/pull/4949), as part of a much wider set of changes, it seems without any comment at all!

While restoring the dependency fixes the problem, I assume the reason for the removal was that it causes lint warnings saying that a `ref.current` shouldn't be used as an effect dependency - effects only run after a render but an update to a ref does not cause a rerender to happen, so this only works when we are fortunate enough for a prop or state change to happen at about the same time, and I didn't want to rely on this.

It seems the way to guarantee code to run when a ref is updated is to use a callback ref, so that is what I have implemented here.

### Testing instructions

Codesandbox from original issue: https://codesandbox.io/s/select-options-height-resize-uy2shp?file=/src/index.js

I have changed the "Select/Test" story to a configuration that is more likely to trigger the bug, so this can also be tested (and I have added a new Cypress test to hopefully catch any regressions in this behaviour).


src/components/select/select-list/select-list.component.js
```diff
@@ -318,17 +335,7 @@ const SelectList = React.forwardRef(
...
-
-      newHeight = listRef.current.clientHeight;
-
-      if (listActionButtonRef.current) {
-        newHeight += listActionButtonRef.current.parentElement.clientHeight;
-      }
-
-      setListHeight(`${newHeight}px`);
-    }, [children]);
+    useLayoutEffect(updateListHeight, [children]);
```
mkrds, line 338: I might be wrong but I think we don't need `useLayoutEffect` anymore here.

robinzigmond: OK, I'll give it a go. I never gave it much thought, I just assumed as it was already using `useLayoutEffect` there would be a good reason (particularly as it's setting the size of DOM elements).

DipperTheDan (commented):
I'm happy with the fix. Just address the comment left by @mkrds and I'll happily approve the PR.

DipperTheDan (dismissed):

mkrds (dismissed):

src/components/select/simple-select/simple-select.test.js
```diff
@@ -996,4 +1011,14 @@ context("Tests for Simple Select component", () => {
...
     );
   });
+
+  describe("check height of Select list when opened", () => {
+    it("with long option text, should not cut off any text", () => {
+      CypressMountWithProviders(<SimpleSelectWithLongWrappingTextComponent />);
+
+      selectText().click();
+      selectListWrapper().should("be.visible");
+      selectListWrapper().should("have.css", "height", "152px");
```
divyajindel, line 1021: describe("check height of Select list when opened", () => {
    it("should not cut off any text with long option text", () => {
      CypressMountWithProviders(<SimpleSelectWithLongWrappingTextComponent />);
      selectText().click();
      selectListWrapper().should("have.css", "height", "152px").and("be.visible");
    });
  });

divyajindel (changes requested):

divyajindel (approved):

PR #5370, opened 2022-08-10
test(fieldset): refactor for cypress tests

stephenogorman:
### Proposed behaviour

Refactor Cypress tests for Fieldset component to use new cypress-component-react framework for testing.
Add fieldset.test.js.

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [x] Run npx cypress open --component to check if there is newly added test.file
- [x] Check if the `fieldset.test.js` file passed
- [x] Run npx cypress open --component to check none of the other *.test.js files have regressed
- [x] Run npx cypress run --e2e to check none of the feature files have regressed.


src/components/fieldset/fieldset.test.js
```diff
@@ -0,0 +1,212 @@
...
+      CypressMountWithProviders(<FieldsetComponent legend="" />);
+
+      legendPreview().should("not.exist");
+    });
+
+    it.each([
+      [true, 36, 39.984375, 78.1875],
+      [false, 18, 76.859375, 919],
+    ])(
+      "should verify Fieldset preview is not displayed",
```
sianford, line 85: The test name appears to be the same as the test above but shouldn't it specify that it relates to verifying that the fields are inline or not?

stephenogorman: Yes it should. The perils of copying and pasting!

ZhuoyuJin (dismissed):

src/components/fieldset/fieldset.test.js
```diff
@@ -0,0 +1,229 @@
...
+    ])(
+      "should verify Fieldset is displayed %s if inline prop is %s",
+      (state, bool, labelHeight, labelWidth, inputWidth) => {
+        CypressMountWithProviders(<FieldsetComponent inline={bool} />);
+
+        const upperLabelHeight = labelHeight + 1;
+        const upperLabelWidth = labelWidth + 1;
+        const upperInputWidth = inputWidth + 1;
+
+        getDataElementByValue("label")
```
DlgSHi, line 93: thanks for fixing this, but this could be more simplified. Instead of using the same code 3 times, we can create a `const` like:

`
const verifyCssProps = (element, cssProp, lessValue, greaterValue) => {
getDataElementByValue(element)
           .invoke("css", cssProp)
           .then(parseFloat)
           .then(($el) => {
             expect($el).to.be.gt(lessValue);
             expect($el).to.be.lt(greaterValue);
           })
}
`

and then use it 3 times below.

DlgSHi (dismissed):

ZhuoyuJin (approved):

PR #5399, opened 2022-08-18
fix(form-field): modify typing of data props (FE-5318)

Parsium:
### Proposed behaviour

- Modify `tagComponent` and `TagProps` type interface so `data-component` is `string | undefined` 
- Refactor tests so `id` prop must be explicitly passed in each test

### Current behaviour

- `data-component` prop is marked as required in `FormFieldProps` - which is incorrect and results in an error message when consumers do not pass the `data-component` prop to input-based components like `Textbox` or `Select`.

### Checklist

#### QA

### Additional context

Currently, `FormFieldProps` inherits `TagProps` - an interface defined for our internal `tagComponent` util. However, this interface should be reserved for the `tagComponent` only, since its use case of the data tags is slightly different.  

### Testing instructions

The following CodeSandbox is an example of the broken behaviour (note console error):

https://codesandbox.io/s/nervous-ptolemy-ed2jn6

You can see the new behaviour by looking at the version in the comment by `codesandbox[bot]` (which should not have the same console error)


src/__internal__/form-field/form-field.component.tsx
```diff
@@ -43,7 +43,13 @@ interface CommonFormFieldProps extends MarginProps, ValidationProps {
   validationIconId?: string;
 }
 
-export interface FormFieldProps extends CommonFormFieldProps, TagProps {
+export interface FormFieldProps extends CommonFormFieldProps {
```
edleeks87, line 46: It might be easier to just update TagProps to make the `data-component` type optional, like Maciek has done [here](https://github.com/Sage/carbon/blob/47fa084e2eb12ca80c2f17fd93bd15c59aa727cf/src/__internal__/utils/helpers/tags/tags.ts), then you can keep using the util

src/__internal__/form-field/form-field.component.tsx
```diff
@@ -90,6 +99,8 @@ const FormField = ({
   validationIconId,
   ...rest
 }: FormFieldProps) => {
+  const { validationRedesignOptIn } = useContext(NewValidationContext);
```
edleeks87, line 102: Shall we move this context to the `Label` component instead? I don't think `FormField` needs to care about the `validationRedesignOptIn` and we can add it as a guard to https://github.com/Sage/carbon/blob/19da80816cc29da9a6c4b3b47412f967aeff99b5/src/__internal__/label/label.component.tsx#L100

Parsium: Good idea @edleeks87 👍🏼

Parsium: @edleeks87 I noticed if we do move the context into `Label` and bypass `FormField`, we would need to move the `invalidValidationProp` check into each input component as well. I have experimented creating a new helper for this, but to avoid delaying this fix, I will leave it as it is and come back to it in the future.

mkrds (approved):

src/__internal__/form-field/form-field.component.tsx
```diff
@@ -41,6 +41,8 @@ interface CommonFormFieldProps extends MarginProps, ValidationProps {
   reverse?: boolean;
   /** Id of the validation icon */
   validationIconId?: string;
+  /** Flag dedicating if latest validation design should be used */
```
edleeks87, line 44: If we're going to keep this as prop for now can we use @ignore @private here so it's not in the prop table

src/__internal__/form-field/form-field.component.tsx
edleeks87: I'm not sure that's the case tbh but I'm happy to leave it as it's only one level we're trying to bypass and we'll eventually be removing it all anyway 👍

edleeks87 (approved):

mkrds (approved):

ZhuoyuJin (approved):

ZhuoyuJin (approved):

PR #5421, opened 2022-08-25
test(multi select): add refactored tests for cypress

sianford:
### Proposed behaviour

Refactor Cypress tests for MultiSelect component to use new cypress-component-react framework for testing.  Add multi-select.test.js file. Remove unnecessary prop from number.test.js.

Also, update filterable-select.test.js and simple-select.test.js files with additional test cases and enhancements.

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [x] Run npx cypress open --component to check there is a newly added test.file
- [x] Run the multi-select.test.js file and check it passes 
- [x] Check all unnecessary files were deleted, including .json file if applicable
- [x] Run npx cypress run --component to check none of the other *.test.js files have regressed
- [x] Run npx cypress run --e2e to check none of the feature files have regressed
- [x] Run npx cypress run --e2e to check none of the accessibility files have regressed


src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+      CypressMountWithProviders(<MultiSelectComponent />);
+
+      dropdownButton().click();
+      selectListText(option1).click();
+      selectInput().should("have.attr", "aria-expanded", "true");
+      selectList().should("be.visible");
+      multiSelectPill().should("have.attr", "title", option1);
+      selectListText(option2).click();
+
+      for (let i = 0; i < 2; i++) {
```
DlgSHi, line 806: why do we need to use here a `for` loop and an `if` statement? we can just delete `for` and `if` and use

```suggestion
      multiSelectPillByPosition(0).should("have.attr", "title", option1);
      multiSelectPillByPosition(1).should("have.attr", "title", option2);
```

and that's it :)

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+        let option;
+        if (i === 0) {
+          option = option1;
+        }
+        if (i === 1) {
+          option = option2;
+        }
+
+        multiSelectPillByPosition(i).should("have.attr", "title", option);
```
DlgSHi, line 842: same here as above

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
```
DlgSHi, line 955: same here to avoid `for` loop and `if`

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+        if (i === 0) {
+          borderColor = "rgb(255, 191, 0)";
+          pillColor = borderColor;
+        }
+        if (i === 1) {
+          borderColor = "rgb(0, 119, 200)";
+          pillColor = "rgba(0, 0, 0, 0)";
+        }
+
+        multiSelectPillByPosition(i)
```
DlgSHi, line 1145: it is much easier to put those 2 values here - instead of wrapping it in a `for` loop and `if` statement

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+
+      const position = "first";
+      const option = ["1"];
+
+      dropdownButton().click();
+      selectOption(positionOfElement(position))
+        .click()
+        .then(() => {
+          // eslint-disable-next-line no-unused-expressions
+          expect(callback).to.have.been.calledWith({
```
DlgSHi, line 1278: good solution to use `calledWith` method 👍🏻

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+            target: { value: option },
+          });
+        });
+    });
+
+    it.each([["downarrow"], ["uparrow"]])(
+      "should call onKeyDown event when %s key is pressed",
```
DlgSHi, line 1285: we should also check if won't pass `onKeyDownEnabled` prop but pass `onKeyDown={callback}` it shouldn't trigger this event

DlgSHi (changes requested):

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+        cy.viewport(400, 300);
+
+        CypressMountWithProviders(
+          <MultiSelectComponent
+            labelInline
+            adaptiveLabelBreakpoint={breakpoint}
+          />
+        );
+
+        if (alignment === "inline") {
```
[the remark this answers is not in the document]
ZhuoyuJin: It's more like a question than a review message, wondering is there any reason that we need to get rid of the if statement here? I think Sian has this in her filterable-select test and it makes sense of having the if here, so why do we try to use another parameter to avoid if?

ZhuoyuJin (changes requested):
Can we add another test for defaultValue prop?

src/components/select/multi-select/multi-select.test.js
sianford: I did it that way to avoid repeating code for each option.

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1318 @@
...
+          <MultiSelectComponent
+            labelHelp="Help"
+            tooltipPosition={tooltipPositionValue}
+          />
+        );
+
+        helpIcon().trigger("mouseover");
+        tooltipPreview()
+          .should("be.visible")
+          .and("have.css", tooltipPositionValue);
```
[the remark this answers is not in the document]
sianford: @DlgSHi this now fails for the "top" position test case when I change the code to this. The tooltip is now rendered in the "bottom" position :(

src/components/select/multi-select/multi-select.test.js
sianford: I just love a good `for` loop and `if` statement :D

src/components/select/multi-select/multi-select.test.js
sianford: Adding `onKeyDownEnabled` of any of the Select components is an error on my part. It's not a valid prop. They only have the `onKeyDown` prop. Removed `onKeyDownEnabled` from Multi, Filterable and Simple Select.

src/components/select/multi-select/multi-select.test.js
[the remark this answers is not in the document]
sianford: @ZhuoyuJin on reflection, it most likely uses more resources to cycle around the `if` statement in this case than it would to check one more variable in the array so I've amended both Multi and Filterable Select with @DlgSHi 's suggested change.

sianford:
> Can we add another test for defaultValue prop?

@ZhuoyuJin I already had a test case for that prop at line 1115 :)

src/components/select/multi-select/multi-select.test.js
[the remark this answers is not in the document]
sianford: I've corrected this in Multi, Filterable and Simple Select. Good spot.

ZhuoyuJin (approved):

src/components/select/multi-select/multi-select.test.js
```diff
@@ -0,0 +1,1281 @@
...
+      <Option text="Pink" value="7" borderColor="pink" />
+      <Option text="Purple" value="8" borderColor="purple" />
+      <Option text="Red" value="9" borderColor="red" fill />
+      <Option text="White" value="10" borderColor="white" />
+      <Option text="Yellow" value="11" borderColor="yellow" fill />
+    </MultiSelect>
+  );
+};
+
+const testData = ["mp150ú¿¡üßä", "!@#$%^*()_+-=~[];:.,?{}&\"'<>"];
```
DlgSHi, line 466: we will use the  import of those values from our `constants` (are added be @DianeGit22 )

DlgSHi (approved):

DlgSHi (approved):

ZhuoyuJin (approved):

PR #5427, opened 2022-08-29
refactor(message): convert to typescript - FE-4807

mkrds:
### Proposed behaviour

This PR contains Message TypeScript conversion 

### Checklist

#### QA

### Testing instructions


DipperTheDan (approved):

src/components/content/content.component.js
```diff
@@ -21,11 +21,13 @@ const Content = ({
   align = "left",
   titleWidth,
   bodyFullWidth = false,
+  className,
   ...rest
 }) => (
   <StyledContent
     align={align}
     bodyFullWidth={bodyFullWidth}
+    className={className}
```
grabkowski, line 30: what is the reason to add that prop? I thought that we are deprecating `className` and we should not encourage users to provide their own classes.

mkrds: It's because of this line of code - https://github.com/Sage/carbon/pull/5427/files#diff-f1d316a518fbce5a7175b4c521660a2f015b9bb78ed4f0f36d61b34447f9f765R16

Anyways I'm not sure that we're deprecating className to be honest

grabkowski: ok, but why do we need that class for? I think that might just be a leftover from the times when we have not used styled-components.

As for the className - when we have implemented styled-components and styled-system we no longer have to pass any classes to the html.

DipperTheDan (approved):

grabkowski (approved):

divyajindel (approved):

PR #5428, opened 2022-08-30
feat(search): add validation props (FE-5292)

grabkowski:
### Proposed behaviour

Add validation props to the Search component and pass them down to the Textbox

### Current behaviour

Validations could not be applied to the Search component

### Checklist

#### QA

### Testing instructions


src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+<Canvas>
+  <Story name="validations - string - component">
+    {() => {
+      const [state, setState] = useState({
+        error: "<>/",
+        warning: "<>/",
+        info: "<>/",
+      });
+      const handleChange = (validation) => (e) => {
+        setState({ ...state, [validation]: e.target.value.rawValue });
```
robinzigmond, line 201:
```suggestion
        setState({ ...state, [validation]: e.target.value });
```

Not sure what the `rawValue` is doing here - the component is crashing for me with this in, and `e.target.value` is just the string value. (And this matches the "controlled" story above.)

src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+
+For more information check our [Validations](?path=/docs/documentation-validations--page "Validations") documentation page
+
+#### As a string
+
+<Canvas>
+  <Story name="validations - string - component">
+    {() => {
+      const [state, setState] = useState({
+        error: "<>/",
```
robinzigmond, line 196: I know you want to show something with "invalid characters" to make it look realistic, but this looks weird to me with those values. I'd put some plain text in there, maybe with one invalid character like `foo/`? Up to you, I don't mind leaving this as it is if you prefer it.

src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+It is possible to use the `tooltipPosition` to override the default placement of tooltips rendered as part of this component.
+
+<Canvas>
+  <Story
+    name="validations - string - with tooltipPosition overriden - component"
+    parameters={{ chromatic: { disable: true } }}
+  >
+    {() => {
+      const [state, setState] = useState({
+        error: "<>/",
```
robinzigmond, line 226: same comment as above re these initial values

src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+        warning: "<>/",
+        info: "<>/",
+      });
+      const handleChange = (validation) => (e) => {
+        setState({ ...state, [validation]: e.target.value.rawValue });
```
robinzigmond, line 231:
```suggestion
        setState({ ...state, [validation]: e.target.value });
```

src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+  </Story>
+</Canvas>
+
+#### As a boolean
+
+<Canvas>
+  <Story name="validations - boolean">
+    {() => {
+      const [state, setState] = useState({
+        error: "<>/",
```
robinzigmond, line 254: same comment as above re these initial values

src/components/search/search.stories.mdx
```diff
@@ -177,6 +177,101 @@ In this example the `variant` prop is "dark" and no search button is present.
...
+        warning: "<>/",
+        info: "<>/",
+      });
+      const handleChange = (validation) => (e) => {
+        setState({ ...state, [validation]: e.target.value.rawValue });
```
robinzigmond, line 259:
```suggestion
        setState({ ...state, [validation]: e.target.value });
```

robinzigmond (changes requested):
I think we should add some unit tests for these validation props on Search. I know it will already be covered by the tests on Textbox where we're passing them through to but it would be good to make sure they're getting passed through to make sure we don't accidentally remove these from Search.

grabkowski:
> I think we should add some unit tests for these validation props on Search. I know it will already be covered by the tests on Textbox where we're passing them through to but it would be good to make sure they're getting passed through to make sure we don't accidentally remove these from Search.

Could you tell me why do we need unit tests for that, when we have everything already covered by a Chromatic snapshot?

src/components/search/search.stories.mdx
grabkowski: my mistake, I'll remove that

robinzigmond:
> > I think we should add some unit tests for these validation props on Search. I know it will already be covered by the tests on Textbox where we're passing them through to but it would be good to make sure they're getting passed through to make sure we don't accidentally remove these from Search.
> 
> Could you tell me why do we need unit tests for that, when we have everything already covered by a Chromatic snapshot?

Good point, I'd forgotten about Chromatic. I'll let this one go :)

robinzigmond (approved):

mkrds (approved):

robinzigmond (approved):

mkrds (approved):

PR #5432, opened 2022-08-31
feat: add optional list height prop - FE-5313

### LINKED ISSUE #5257: Impossible to change height of Filterable Select
### Desired behaviour

https://carbon.sage.com/?path=/docs/select-filterable--default-story#new-designs-validation
Zrzut ekranu 2022-06-23 o 13 52 55

### Current behaviour

height is hardcoded (dropdown)

### Suggested Solution

add prop to change height

_No response_

_No response_

### Confidentiality

mkrds:
### Proposed behaviour

Closes #5257
This PR adds support for `listMaxHeight` prop which allows to override the `max-height` css rule on the SelectList

### Checklist

#### QA

### Testing instructions


src/components/select/select-list/select-list.component.js
```diff
@@ -461,7 +462,6 @@ const SelectList = React.forwardRef(
           >
             <StyledSelectListContainer
               data-element="select-list-wrapper"
-              height={listHeight}
```
robinzigmond, line 464: since we're removing this (which I agree is not needed any more), we should also remove all references to `listHeight` and `setListHeight` elsewhere in the component, as they're not used anymore. That means the whole `useLayoutEffect `that sets the height can also be removed!

src/components/select/simple-select/simple-select.test.js
```diff
@@ -847,6 +847,17 @@ context("Tests for Simple Select component", () => {
       selectListOptionGroup("1").should("have.text", "Group one");
     });
 
+    it("should render option list with proper maxHeight value", () => {
```
robinzigmond, line 850: I'm not sure if we need a Cypress test for this new prop when there's already a unit test for it. But the tests are simple enough and I'd rather write too many than too few, so no harm done.

robinzigmond (commented):
Aside from the comments below, it would be good I think to add tests for FilterableSelect and MultiSelect similar to the ones you've got for SimpleSelect, just so we're sure this prop works across all 3 selects.

src/components/select/select-list/select-list.component.js
mkrds: It's actually still used to set height on `StyledPopoverContainer`

grabkowski (approved):

src/components/select/select-list/select-list.component.js
robinzigmond: oh yes, sorry - not sure how I missed that

grabkowski (approved):

robinzigmond (approved):

PR #5449, opened 2022-09-06
refactor(carousel): convert to typescript

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/carousel/carousel.component.tsx
```diff
@@ -38,16 +54,18 @@ const BaseCarousel = ({
...
   ...props
-}) => {
+}: CarouselProps) => {
   const [selectedSlideIndex, setSelectedSlideIndex] = useState(
     Number(slideIndex) || Number(initialSlideIndex)
   );
   const transitionDirection = useRef(NEXT);
-  const lastSlideIndexProp = useRef(props.slideIndex);
+  const lastSlideIndexProp = useRef(slideIndex);
   const id = guid();
```
Parsium, line 63:
```suggestion
  const id = useMemo(() => guid());
```
It might be worth memorising this value so the id isn't recalculated on the next render

src/components/carousel/carousel.component.tsx
```diff
@@ -188,28 +206,8 @@ const BaseCarousel = ({
...
-};
-
-const Carousel = withTheme(BaseCarousel);
 Carousel.displayName = "Carousel";
 
-export default BaseCarousel;
+export default Carousel;
 
-export { Carousel, Slide };
+export { Slide };
```
Parsium, line 213: Instead of importing and exporting the `Slide` component here, could we do this in the `index` file for `Carousel` instead to reduce the number of import/exports?

src/components/carousel/slide/slide.component.tsx
```diff
@@ -2,14 +2,14 @@
 import React from "react";
 import tagComponent from "../../../__internal__/utils/helpers/tags/tags";
 import { SlideStyle } from "./slide.style";
-import baseTheme from "../../../style/themes/base";
 
-const Slide = (props) => (
+export interface SlideProps {
```
Parsium, line 6:
```suggestion
export interface SlideProps extends TagProps {
```
Since we have the `tagComponent` util for this component, I think we should extend the interface so the `data-*` attributes are recognised

Parsium, line 6:
```suggestion
import { TagProps } from ""../../../__internal__/utils/helpers/tags/tags";
export interface SlideProps extends TagProps {
```

Since we have the tagComponent util for this component, I think we should extend the interface so the data-* attributes are recognised

src/components/carousel/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Carousel, Slide } from "./carousel.component";
+export type { CarouselProps } from "./carousel.component";
```
Parsium, line 2:
```suggestion
export { Carousel } from "./carousel.component";
export { Slide } from "./slide"; 
export type { CarouselProps } from "./carousel.component";
export type { SlideProps } from "./slide";
```
Instead of importing and exporting the Slide component twice both here and in `carousel.component`, could we do this once here to reduce the number of import/exports?

grabkowski: good point

Parsium (approved):

DipperTheDan (commented):
Happy to approve once the comment regarding extending the `TagProps` by @Parsium has been done.

DipperTheDan (approved):

divyajindel (approved):

PR #5457, opened 2022-09-08
refactor(image): convert to typescript

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/image/image.component.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import invariant from "invariant";
+import { StyledImage, StyledImageProps } from "./image.style";
+
+export const Image = ({ alt, src, children, ...rest }: StyledImageProps) => {
+  invariant(
+    !src || (src && !children),
```
robinzigmond, line 7:
```suggestion
    !src || !children,
```

Comes to the same thing but imo it's easier to understand this way.

src/components/image/image.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+    "The 'Image' component renders as an 'img' element when the 'src' prop is used and therefore does not accept children."
+  );
+
+  invariant(
+    !src || (src && alt && typeof alt === "string"),
```
robinzigmond, line 12:
```suggestion
    !src || (alt && typeof alt === "string"),
```

same as above

src/components/image/image.style.ts
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+StyledImage.defaultProps = {
+  backgroundPosition: "center",
+  backgroundRepeat: "no-repeat",
+  backgroundSize: "cover",
+  theme: baseTheme,
+};
+
+// eslint-disable-next-line import/prefer-default-export
```
robinzigmond, line 46: I don't think we need to disable eslint here as we're also exporting the `StyledImageProps` (I'm not seeing any errors/warnings when I remove the comment).

src/components/image/image.spec.tsx
```diff
@@ -29,36 +29,33 @@ describe("Image", () => {
...
+        );
+      }).toThrow(errorMessage);
     });
 
     it("throws an error when `src` is passed and no `alt` provided", () => {
-      mount(<Image src="foo.jpg" />);
-      expect(console.error).toHaveBeenCalled();
-    });
-
-    it("does not throw an error when `src` is passed and `alt` provided", () => {
```
robinzigmond, line 54: is there a good reason for removing this test? Why not something like `expect(() => { mount(<Image src="foo.jpg" alt="foo"/>); }).not.toThrow();`?

edleeks87 (approved):

src/components/image/image.component.tsx
grabkowski: right, I got carried away when reversing that conditional for invariant

src/components/image/image.spec.tsx
robinzigmond: Oh yes, of course. I withdraw the comment then :)

robinzigmond (approved):

edleeks87 (approved):

divyajindel (approved):

PR #5461, opened 2022-09-08
test(pod): add cypress tests

divyajindel:
### Proposed behaviour

Refactor Cypress tests for the `Pod` component to use new cypress-component-react framework for testing.
Add `pod.test.js`. 

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [x] Check if the `pod.test.js` file passed


src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
 );
 
 context("Testing Pod component", () => {
+  describe("should render Pod component", () => {
+    it.each([
+      [true, "0px none rgb(0, 0, 0)"],
+      [false, "0px none rgb(0, 0, 0)"],
+    ])(
+      "should check when border is %s for Pod component",
```
ZhuoyuJin, line 63: Think it would be better to say "should check when border is %s the border value is %s for Pod component "

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+        "rgb(255, 255, 255) none repeat scroll 0% 0% / auto padding-box border-box",
+        "rgba(2, 18, 36, 0.2) 0px 2px 3px 0px",
+      ],
+      [
+        "transparent",
+        "rgba(0, 0, 0, 0) none repeat scroll 0% 0% / auto padding-box border-box",
+        "none",
+      ],
+    ])(
+      "should check %s variant for Pod component",
```
ZhuoyuJin, line 118: similar as above, something like "should check for %s variant pod, the color is %s and boxShadow is %s"

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+          podDelete().should("be.visible");
+        }
+      }
+    );
+
+    it.each([
+      [true, "1350px"],
+      [false, "77.5625px"],
+    ])(
+      "should check when internalEditButton is %s for Pod component",
```
ZhuoyuJin, line 212: same as above "should check the width value when internalEditButton is %s for Pod component" something like this

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+      }
+    );
+
+    it.each(specialCharacters)(
+      "should check title as %s for Pod component",
+      (title) => {
+        CypressMountWithProviders(<PodComponent title={title} />);
+        podTitle().should("have.text", title);
+      }
+    );
```
DlgSHi, line 133:
```suggestion
    );
    
```

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+      (boolVal, state) => {
+        CypressMountWithProviders(
+          <SoftDeletePodComponent softDelete={boolVal} />
+        );
+        podSoftDelete().should(state);
+      }
+    );
+
+    it.each([
+      [false, "77.5625px"],
```
DlgSHi, line 180: same is here with `"77.5625px"`. I think we need to use `integers` instead of `decimals`

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+          podDelete().should("not.be.visible");
+        } else {
+          podDelete().should("be.visible");
+        }
+      }
+    );
+
+    it.each([
+      [true, "1350px"],
+      [false, "77.5625px"],
```
DlgSHi, line 210: same here

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+    });
+
+    it.each([false, true])(
+      "should check when triggerEditOnContent is %s for Pod component",
+      (boolVal) => {
+        CypressMountWithProviders(
+          <PodComponent triggerEditOnContent={boolVal} />
+        );
+        if (boolVal === false) {
+          podContent()
```
DlgSHi, line 258: what is the difference between those 2 assertions? and `IF` statement has no sense here

src/components/pod/pod.test.js
```diff
@@ -13,6 +55,234 @@ const PodWithSoftDelete = () => (
...
+            .then(() => {
+              // eslint-disable-next-line no-unused-expressions
+              expect(callback).not.to.have.been.calledOnce;
+            });
+        }
+      }
+    );
+
+    it.each(specialCharacters)(
+      "should check children as %s for Pod component",
```
DlgSHi, line 276: why this TC is in `events` describe ?

DlgSHi (changes requested):

DlgSHi (dismissed):

ZhuoyuJin (approved):

src/components/pod/pod.test.js
```diff
@@ -1,9 +1,55 @@
-import React from "react";
+import * as React from "react";
```
DlgSHi, line 1: do we need this import?

src/components/pod/pod.test.js
```diff
@@ -13,22 +59,236 @@ const PodWithSoftDelete = () => (
...
-      const blockBackgroundColor = "rgb(230, 235, 237)";
+  describe("should render Pod component", () => {
+    it.each([
+      [true, "0px none rgb(0, 0, 0)"],
+      [false, "0px none rgb(0, 0, 0)"],
+    ])(
+      "should check when border is %s the border value is %s for Pod component ",
+      (boolVal, border) => {
+        CypressMountWithProviders(<PodComponent border={boolVal} />);
+        podComponent().should("have.css", "border", border);
```
Parsium, line 70:
```suggestion
  describe("should render Pod component", () => {
    it.each([
      [true, "1px solid rgb(204, 214, 219)"],
      [false, "0px none rgb(0, 0, 0)"],
    ])(
      "should check when border is %s the border value is %s for Pod component ",
      (boolVal, border) => {
        CypressMountWithProviders(<PodComponent border={boolVal} />);
        podBlock().should("have.css", "border", border);
```
I think you are testing the wrong HTML element here. When `border` is `true` we should not be having these styles.

src/components/pod/pod.test.js
divyajindel: I'll correct this one.

src/components/pod/pod.test.js
divyajindel: Thanks @Parsium . I'll fix it now.

src/components/pod/pod.test.js
```diff
@@ -13,22 +59,236 @@ const PodWithSoftDelete = () => (
...
+        CypressMountWithProviders(<PodComponent footer={footerText} />);
+        podFooter().should("have.text", footerText).and("be.visible");
+      }
+    );
+
+    it.each([
+      [true, "be.visible"],
+      [false, "not.exist"],
+    ])(
+      "should check when softDelete is %s for Pod component",
```
Parsium, line 163:
```suggestion
      "should check visibility of undo button when softDelete is %s",
```
Personally I think we should make it clear what this test is actually verifying here. This makes it easier to understand when it is being run.

src/components/pod/pod.test.js
```diff
@@ -13,22 +59,236 @@ const PodWithSoftDelete = () => (
...
+        );
+        if (boolVal === true) {
+          podContent().click();
+          podEditContainer().should("be.visible");
+          podDelete().should("not.be.visible");
+        } else {
+          podDelete().should("be.visible");
+        }
+      }
+    );
```
Parsium, line 201:
```suggestion
    it("edit button is only visible when user clicks on Pod component and displayEditButtonOnHover is true", () => {
        CypressMountWithProviders(
          <PodComponent displayEditButtonOnHover />
        );
        podEditContainer().should("not.be.visible");
        
        podContent().click();
        
        podEditContainer().should("be.visible");
        podDelete().should("not.be.visible");
      }
    );
    
    it("edit and delete buttons are visible when displayEditButtonOnHover is false", () => {
    CypressMountWithProviders(
          <PodComponent displayEditButtonOnHover={false} />
          podEditContainer().should("be.visible");
          podDelete().should("be.visible");
        );
    });
```

In my opinion I would split this test into two to avoid the conditional statement. This would ensure one piece of behaviour is tested per test which would make debugging easier if one of tests begin failing.

src/components/pod/pod.test.js
```diff
@@ -13,22 +59,236 @@ const PodWithSoftDelete = () => (
...
+        divPodComponent().should(
+          "have.css",
+          "background-color",
+          "rgb(217, 224, 228)"
+        );
+      }
+    );
+
+    it.each([100, 200, 300])(
+      "should check when internalEditButton is %s for Pod component",
```
Parsium, line 239:
```suggestion
      "renders Pod component with correct height when height prop is %s",
```

src/components/pod/pod.test.js
```diff
@@ -13,22 +59,236 @@ const PodWithSoftDelete = () => (
...
+        const childrenColor = "rgba(0, 0, 0, 0.65)";
 
-      CypressMountWithProviders(<PodWithSoftDelete />);
+        CypressMountWithProviders(<PodWithSoftDelete />);
 
-      cy.contains("Content").should("have.css", "color", childrenColor);
-      cy.contains("More content").should("have.css", "color", childrenColor);
+        cy.contains("Content").should("have.css", "color", childrenColor);
+        cy.contains("More content").should("have.css", "color", childrenColor);
+      });
```
Parsium, line 291: You could probably remove the `PodWithSoftDelete` component here as you could achieve the same result using `SoftDeletePodComponent` component instead.

src/components/pod/pod.test.js
```diff
@@ -3,7 +3,53 @@ import Pod from "./pod.component";
...
+      footer="Footer"
+      onEdit={() => {}}
+      onDelete={() => {}}
+      mb={3}
+      {...props}
+    />
+  );
+};
+
+const SoftDeletePodComponent = ({ ...props }) => {
```
Parsium, line 39: Personally I would either add the `softDelete` prop to the internal pod by default and amend the affected tests, or rename this mock component, since this is a bit misleading currently as this is not a render soft delete pod by default.

cypress/locators/pod/index.js
```diff
@@ -14,4 +17,11 @@ export const podPreview = () => cy.get(POD_DATA_COMPONENT).children();
 export const podFooter = () => podComponent().find(POD_FOOTER);
 export const podContent = () => podComponent().find(POD_CONTENT);
 export const podSubTitle = () => cy.get(POD_SUBTITLE);
-export const podEdit = () => podComponent().find(POD_EDIT_ICON);
+export const ppodEditIcon = () => podComponent().find(POD_EDIT_ICON);
```
Parsium, line 20:
```suggestion
export const podEditIcon = () => podComponent().find(POD_EDIT_ICON);
```

cypress/locators/pod/index.js
```diff
@@ -14,4 +17,11 @@ export const podPreview = () => cy.get(POD_DATA_COMPONENT).children();
...
+export const divPodComponent = () => cy.get(POD_DATA_COMPONENT).find("div");
```
Parsium, line 21:
```suggestion
```
This `div` is the same as the `podBlock`, so we don't need this locator.

DlgSHi (commented):
Nothing to add apart from comments from @Parsium

src/components/pod/pod.test.js
divyajindel: Both the custom components are different as we are passing props in SoftDeletePodComponent.

src/components/pod/pod.test.js
```diff
@@ -3,7 +3,52 @@ import Pod from "./pod.component";
...
+      footer="Footer"
+      onUndo={() => {}}
+      {...props}
+    >
+      Soft delete state
+    </Pod>
+  );
+};
 
 const PodWithSoftDelete = () => (
```
Parsium, line 53:
```suggestion
const SoftDeletePodWithChildren = () => (
```
Appreciate this is not part of your changes, but personally I think now we have `SoftDeletePodComponent` we should rename this slightly to make it clear why this mock component should exist independently to `SoftDeletePodComponent`

src/components/pod/pod.test.js
```diff
@@ -3,7 +3,52 @@ import Pod from "./pod.component";
...
+
+const SoftDeletePodComponent = ({ ...props }) => {
+  return (
+    <Pod
+      title="Title"
+      subtitle="Subtitle"
+      description="Description"
+      footer="Footer"
+      onUndo={() => {}}
+      {...props}
```
Parsium, line 46:
```suggestion
      softDelete
      {...props}
```
Personally I think we should add this prop by default, since the `Pod` will not render with the soft delete styles unless both `softDelete` and `onUndo` defined. This should not affect the test on line 165, as we are spreading the props on line 46 which will overwrite any prop values.

Parsium (dismissed):

DlgSHi (approved):

Parsium (approved):

PR #5471, opened 2022-09-13
 refactor(preview): refactor code and convert files to typescript (FE-4822)

grabkowski:
### Proposed behaviour

Things refactored:
- set width and height directly in a styled component instead of passing them as html attributes
- rewrite unit tests to replace snapshots
- extract preview-placeholder to an internal component

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/preview/preview.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const Preview = ({
+  children,
+  loading,
+  lines = 1,
+  ...props
+}: PreviewProps) => {
+  const marginProps = filterStyledSystemMarginProps(props);
+  const hasPlaceholder = typeof loading !== "undefined" ? loading : !children;
```
robinzigmond, line 26:
```suggestion
  const hasPlaceholder = loading === undefined ? !children : loading;
```

Not very significant but I prefer to write it this way, for 2 reasons:
1) `type of loading !== "undefined"` is exactly the same as `loading !== undefined` (the only difference is if `loading` isn't even declared, in which case my version would crash whereas yours would give `false` - but the `loading` is defined in the props argument here so that's never going to be an issue)
2) I much prefer ternaries and if/else to be the "right" way round - check if two things are equal, not if they're not equal. Mostly because I'm stupid :-)

This is just a suggestion, if you prefer it your way then you don't need to justify it to me, just leave it!

src/components/preview/__internal__/preview-placeholder.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+  index,
+  lines,
+  width,
+  ...props
+}: PreviewPlaceholderProps) => {
+  const isLastLine = lines > 1 && lines === index;
+
+  return (
+    <StyledPreviewPlaceholder
+      key={index}
```
robinzigmond, line 27: I know this was here before but I don't think we need a `key` here at all, as we're not directly rendering `StyledPreviewPlaceholder` in an array of components. The key is added later, inside the actual `Preview` component.

src/components/preview/preview.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  if (hasPlaceholder) {
+    const placeholders = [];
+
+    for (let i = 1; i <= lines; i++) {
+      placeholders.push(
+        <PreviewPlaceholder key={i} index={i} lines={lines} {...props} />
+      );
+    }
```
Parsium, line 35: What do you think about creating a function out of this or declaring this declaratively so we can memorise the value to save rerenders?

src/components/preview/__internal__/preview-placeholder.component.tsx
grabkowski: right, I have moved the key to the new component but I forgot to remove it here

robinzigmond (approved):

divyajindel (approved):

PR #5474, opened 2022-09-14
fix(dialog, sidebar): ensure aria-modal is only added to top modal

### LINKED ISSUE #5303: Voiceover not reading elements for nested DialogFullScreen component in Safari browser
### Current behaviour

When a child `DialogFullScreen` component is nested within a parent `DialogFullScreen` component, Voiceover does not read out elements in the child `DialogFullScreen` component using Safari browser.

### Expected behaviour

When a child `DialogFullScreen` component is nested within a parent `DialogFullScreen` component, expect Voiceover to read out elements in the child `DialogFullScreen` component using Safari browser.

https://codesandbox.io/s/voiceover-dialog-i64x57?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-31787

### Suggested Solution

_No response_

### Carbon Version

109.0.0

### Design Tokens Version

_No response_

Safari

MacOS

_No response_

### Confidentiality

robinzigmond:
Currently we add aria-modal=true to the container div of each modal component (dialog, dialog-full-screen, sidebar). This ensures screenreaders do not announce content outside the modal. However when opening one modal from another, VoiceOver on Safari refuses to read the content on the new modal while the old one still has aria-modal=true. The workaround is to ensure aria-modal=true only appears on whatever the current top modal is. This requires keeping track of the current top modal in a context which has been added to CarbonProvider, as well as keeping a global list of "top modal setters" in the window object to ensure that multiple Carbon Providers in the same application (potentially even from different versions of Carbon) still update each other's context correctly.

fix #5303

### Proposed behaviour

aria-modal=true should only be present on at most one modal (dialog/dialogfullscreen/sidebar) at a time - whichever one is currently on top of a stack of modals

VoiceOver on Safari should correctly announce a new modal when opened,  whether from another modal or not.

### Current behaviour

aria-modal=true is placed on all modals. VoiceOver on Safari will not inform the user about a modal opening when opened from another modal.

### Checklist

#### QA

### Testing instructions

Test with all supported browser/screenreader combinations, as well as without a screenreader, to ensure that situations with multiple dialog/sidebar components work as expected.

If possible, test with multiple CarbonProviders wrapping different parts of an application and ensure that everything still works correctly, with aria-modal only ever applied to one modal (but is always applied to any just-opened modal).

### DIFF
--- a/src/components/carbon-provider/carbon-provider.component.tsx
+++ b/src/components/carbon-provider/carbon-provider.component.tsx
@@ -6,6 +6,8 @@ import CarbonScopedTokensProvider from "../../style/design-tokens/carbon-scoped-
 
 import { ThemeObject } from "../../style/themes/base";
 
+import { TopModalContextProvider } from "./top-modal-context";
+
 export interface CarbonProviderProps {
   theme?: Partial<ThemeObject>;
   children: React.ReactNode;
@@ -24,7 +26,7 @@ export const CarbonProvider = ({
   <ThemeProvider theme={theme}>
     <CarbonScopedTokensProvider>
       <NewValidationContext.Provider value={{ validationRedesignOptIn }}>
-        {children}
+        <TopModalContextProvider>{children}</TopModalContextProvider>
       </NewValidationContext.Provider>
     </CarbonScopedTokensProvider>
   </ThemeProvider>
--- a/src/components/carbon-provider/top-modal-context.tsx
+++ b/src/components/carbon-provider/top-modal-context.tsx
@@ -1,0 +1,53 @@
+import React, {
+  createContext,
+  useState,
+  useEffect,
+  useRef,
+  ReactNode,
+} from "react";
+
+type TopModalContextProps = {
+  topModal: HTMLElement | null;
+};
+
+const TopModalContext = createContext<TopModalContextProps>({
+  topModal: null,
+});
+
+export const TopModalContextProvider = ({
+  children,
+}: {
+  children: ReactNode;
+}) => {
+  const [topModal, setTopModal] = useState<HTMLElement | null>(null);
+
+  // can't add the setter to the global list inside useEffect because that doesn't run until
+  // after the render. We use a ref to ensure it only runs once.
+  const isFirstRender = useRef(true);
+
+  if (isFirstRender.current) {
+    if (!window.__CARBON_INTERNALS_MODAL_SETTER_LIST) {
+      window.__CARBON_INTERNALS_MODAL_SETTER_LIST = [];
+    }
+
+    window.__CARBON_INTERNALS_MODAL_SETTER_LIST.push(setTopModal);
+
+    isFirstRender.current = false;
+  }
+
+  useEffect(() => {
+    return () => {
+      window.__CARBON_INTERNALS_MODAL_SETTER_LIST = window.__CARBON_INTERNALS_MODAL_SETTER_LIST?.filter(
+        (setter) => setter !== setTopModal
+      );
+    };
+  }, []);
+
+  return (
+    <TopModalContext.Provider value={{ topModal }}>
+      {children}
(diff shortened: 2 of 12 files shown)


src/components/carbon-provider/top-modal-context.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const TopModalContextProvider = ({
+  children,
+}: {
+  children: ReactNode;
+}) => {
+  const [topModal, setTopModal] = useState<HTMLElement | null>(null);
+
+  // can't add the setter to the global list inside useEffect because that doesn't run until
+  // after the render. Use useMemo instead so that it only runs once.
+  useMemo(() => {
```
mkrds, line 26: https://reactjs.org/docs/hooks-reference.html#usememo

`You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.`

It looks like use of `useMemo` does not guarantee that it's going to run only once.

mkrds: Also I think it would actually be better to contain that logic inside of `modal-manager` or `useModalManager` files, just pass the the `setTopModal` method through the context

robinzigmond: thanks. I'd forgotten useMemo isn't a guarantee. I guess I'll once again have to rethink how to do this.

As for your other comment - just passing setTopModal through the context doesn't work, we need to keep a global record of all such setters in order to ensure that when a modal is opened or closed in one MFE, it updates the topModal context in all others. I'm open to moving the logic to ModalManager though if it makes sense - I'll look at that.

edleeks87: You might be able to do something with `useRef` to create the equivalent of a `componentWillMount` hook.

```
useBeforeMount(setter) => {
  const mounted = useRef(false)

  if (!mounted.current) {
    if (!window.__CARBON_INTERNALS_MODAL_SETTER_LIST) {
      window.__CARBON_INTERNALS_MODAL_SETTER_LIST = [];
    }

    window.__CARBON_INTERNALS_MODAL_SETTER_LIST.push(setter);
  }
  mounted.current = true
}
```

src/components/carbon-provider/carbon-provider.component.tsx
```diff
@@ -20,14 +22,16 @@ export const CarbonProvider = ({
...
-}: CarbonProviderProps) => (
-  <ThemeProvider theme={theme}>
-    <CarbonScopedTokensProvider>
-      <NewValidationContext.Provider value={{ validationRedesignOptIn }}>
-        {children}
-      </NewValidationContext.Provider>
-    </CarbonScopedTokensProvider>
-  </ThemeProvider>
-);
+}: CarbonProviderProps) => {
```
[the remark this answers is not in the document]
robinzigmond: nope, definitely not. Thanks for catching this, it's only because I initially put the TopModalContext stuff in there before pulling it out to its own component - at that point it obviously needed the longhand function body but I forgot to change this back when I refactored.

edleeks87 (dismissed):

src/components/carbon-provider/top-modal-context.tsx
mkrds: @robinzigmond I don't really understand what is the issue with passing down the `setTopModal` through the context to the `useModalManager` for the sake of having that logic contained in one place.

mkrds (approved):

ZhuoyuJin (dismissed):

edleeks87 (approved):

mkrds (approved):

ZhuoyuJin (approved):

PR #5484, opened 2022-09-16
refactor(pod): convert component code to typescript

Parsium:
### Proposed behaviour

- Refactor component, style and index files of `Pod` to typescript
- Reorganise `Pod` tests
    - Rewrite tests in terms of props as opposed to internal components to improve reliability
    - Replace all snapshot tests
- Refactor stories to typescript
- Supplementary changes
    - Ensure footer is not also aligned when `alignTitle` prop is set
    - Add aria labels to delete and undo buttons
    - Modify `with different sizes` story to include soft delete variants - so chromatic can check for changes.
    - Deprecate support for passing strings and objects to the `Pod`'s `onEdit` prop (see additional context for details)

### Current behaviour

- `Pod` is written in javascript
- `Pod` tests are written in terms of internal styled components

### Additional context

Currently, one accessibility violation with the `Pod` edit button: 

> Elements must only use allowed ARIA attributes

has been left intact. To address this the edit button must be rendered as a html `button` rather than an `a` link - but this first requires removed support for passing external links (as strings or objects) to the `onEdit` prop. Therefore, support for these two prop types has been deprecated in this PR with the intention of removal at a later date.

### Checklist

#### QA

### Testing instructions

- Verify no regressions were introduced from this refactor.
- Check all accessibility violations have been addressed, excluding one violation with edit button (see additional context).
- Verify that `Pod` footer no longer has same alignment as the title when the `alignTitle` prop is passed


src/components/pod/pod.style.ts
```diff
@@ -1,36 +1,52 @@
...
+  variant: PodVariant;
+  noBorder: boolean;
+  softDelete?: boolean;
+  hasButtons: boolean;
+  contentTriggersEdit?: boolean;
+  fullWidth?: boolean;
+  internalEditButton?: boolean;
+  isHovered?: boolean;
+  isFocused?: boolean;
+}
```
grabkowski, line 46: I think that we could pick most of this props out of the main interface.
What do you think of changing "fullWidth" to "editContentFullWidth" to be consistent?

src/components/pod/__snapshots__/pod.spec.js.snap
```diff
@@ -909,25 +794,17 @@ exports[`ActionButtons StyledUndoButton when isHovered prop is set should match
...
-  margin-left: 0px;
-  margin-right: var(--spacing100);
-  margin-bottom: 0px;
-  height: 16px;
-}
-
-.c35 .c2 svg {
-  margin-top: 0;
-}
-
```
grabkowski, line 930: Maybe it's a good time to remove this snapshot? What do you think about that?

src/components/pod/pod.style.ts
Parsium: Thanks for spotting this @grabkowski, seems I forgot to go back and amend these after we decided on this. Will amend now 👍🏼

src/components/pod/__snapshots__/pod.spec.js.snap
Parsium: I would be happy to look into this - I agree the tests we have for `Pod` would be better as explicit tests rather than snapshots

Parsium: I already removed the snapshot in a later commit part of this PR 😄

robinzigmond (approved):

grabkowski (approved):

robinzigmond (dismissed):

grabkowski (dismissed):

robinzigmond (approved):

grabkowski (approved):

divyajindel (dismissed):

grabkowski (approved):

robinzigmond (dismissed):

robinzigmond (dismissed):

grabkowski (approved):

grabkowski (approved):

robinzigmond (approved):

divyajindel (approved):

PR #5486, opened 2022-09-16
fix(multi-action-button): ensure that enter and space keys trigger onClick callback FE-5372

### LINKED ISSUE #5472: MultiActionButton: we cannot use the keyboard to launch a button action
### Current behaviour

Open the MultiActionButton component with arrow down.
Select one of the item with arrow down or arrow up
<Enter> on the keyboard or <Space> on the keyboard
result:
The action link to the button is not launch
The MultiActionButton component is not closed (we see the list of buttons)

### Expected behaviour

Open the MultiActionButton component with arrow down.
Select one of the item with arrow down or arrow up
<Enter> on the keyboard or <Space> on the keyboard

Expected:
The action kinked to the button is launched
The list of buttons is closed.

https://carbon.sage.com/?path=/docs/multi-action-button--default-story

### JIRA Ticket (Sage Only)

SBS-41940

### Suggested Solution

_No response_

### Carbon Version

110.7.0

### Design Tokens Version

2.21.0

Chrome

MacOS, Windows

_No response_

### Confidentiality

edleeks87:
fix #5472

### Proposed behaviour

Ensures that `preventDefault` is only called if `Enter` or `Space` keys are not pressed.

### Current behaviour

`Enter` and `Space` keys tigger `preventDefault`

### Checklist

#### QA

### Testing instructions


edleeks87:
For QA to test https://codesandbox.io/s/carbon-quickstart-forked-4y69qj

nineteen88 (dismissed):

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -170,6 +173,22 @@ describe("MultiActionButton", () => {
...
+    describe.each(["Enter", "Space"])(
+      "when the %s is pressed on a child button",
+      (key) => {
+        it("should call the onClick handler", () => {
+          mainButton.simulate("keydown", { key });
+
+          const button = wrapper
+            .find('[data-element="additional-buttons"]')
+            .find(Button);
+          button.at(0).simulate("click");
```
robinzigmond, line 185: Not sure if this is a typo- but why are we simulating `click`?. Isn't the whole point of this test to ensure that when Enter or Space is pressed, the onClick callback happens? By simulating click, we ensure that this test always passes, even when it shouldn't (I just tried removing the `if` statement you added in order to break the functionality - and sure enough, this test still passes.)

I understand that due to the rather limited way Enzyme's `.simulate` works, just replacing `"click"` with `"keydown"` here isn't going to trigger the click action - but triggering click manually afterwards to force it to pass isn't the right way of fixing this test. I expect we have to not use `.simulate` at all but use `document.dispatchEvent` or similar (as we do in some tests elsewhere).

Or maybe just refactor the tests to RTL :D

src/hooks/__internal__/useMenuKeyboardNavigation/useMenuKeyboardNavigation.ts
```diff
@@ -11,7 +11,10 @@ export default (
 
   const handleKeyDown = useCallback(
     (ev) => {
-      ev.preventDefault();
+      if (!(Events.isEnterKey(ev) || Events.isSpaceKey(ev))) {
```
robinzigmond, line 14: While this is probably OK because I can't think of a specific case it would break anything, I think it's probably safer to only explicitly call `ev.preventDefault()` in the cases where we know we specifically want it (the arrow keys), rather than blanket call it whenever the key isn't Enter or Space. But as I said this is probably OK, and can easily be updated if we discover edge cases I guess.

edleeks87: Yeah I had the same thought, I decided to go with the above as we also need to prevent the default for some of the tabbing behaviour (when leaving the child button container) so went with this to avoid adding it in every if/ else block etc. For now I'd prefer to leave it as is and we can update if we find any edge cases.

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -57,13 +59,14 @@ describe("MultiActionButton", () => {
...
+      childOnClickCB = jest.fn();
       container = document.createElement("div");
       container.id = "enzymeContainer";
       document.body.appendChild(container);
       wrapper = mount(
         <MultiActionButton text="Main Button">
-          <Button>First</Button>
-          <Button>Second</Button>
+          <Button onClick={childOnClickCB}>First</Button>
+          <Button onClick={childOnClickCB}>Second</Button>
```
robinzigmond, line 69: I'm not sure the value of adding these `onClick` handlers no we're not actually asserting anything to do with them. I think this spec file can probably be left alone as we're no longer adding any tests to it (nor do we need to adjust anything to account for the fixed behaviour).

nineteen88 (approved):

ZhuoyuJin (approved):

PR #5489, opened 2022-09-20
refactor(typography): convert typography components to typescript

nineteen88:
### Proposed behaviour

Convert Typography components to Typescript

### Checklist

#### QA

### Testing instructions


src/components/typography/list.component.tsx
```diff
@@ -1,15 +1,21 @@
 import React from "react";
 import propTypes from "prop-types";
-import Typography from "./typography.component";
+import { SpaceProps } from "styled-system";
+import Typography, { TypographyProps } from "./typography.component";
 
-const getListStyleType = (as) => {
+export interface ListProps extends SpaceProps, TypographyProps {
+  as?: keyof JSX.IntrinsicElements;
```
mkrds, line 7:
```suggestion
  as?: "ul" | "ol";
```

src/components/typography/list.component.tsx
```diff
@@ -1,15 +1,21 @@
 import React from "react";
 import propTypes from "prop-types";
```
mkrds, line 2: PropTypes should be removed

src/components/typography/list.component.tsx
```diff
@@ -25,7 +31,7 @@ List.propTypes = {
   as: propTypes.oneOf(["ul", "ol"]),
 };
 
-const ListItem = ({ children, ...props }) => (
+const ListItem = ({ children, ...props }: ListProps) => (
```
mkrds, line 34: Valid `as` values will be different for `ListItem` and `List` so they should have slightly different prop interfaces

src/components/typography/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default } from "./typography.component";
+export { List, ListItem } from "./list.component";
+export type { TypographyProps } from "./typography.component";
+export type { ListProps } from "./list.component";
```
mkrds, line 4:
```suggestion
export type { ListProps, ListItemProps } from "./list.component";
```

See my comment about ListItem having slightly different prop interface than List

src/components/typography/typography.component.tsx
```diff
@@ -170,17 +217,17 @@ const Typography = styled.span.attrs(
     line-height: ${lineHeight};
     margin: ${defaultMargin};
     padding: 0;
-    white-space: ${truncate ? "nowrap" : whiteSpace};
-    word-wrap: ${wordWrap};
-    text-overflow: ${truncate ? "ellipsis" : textOverflow};
-    ${truncate &&
+    white-space: ${truncate ? "nowrap" : (whiteSpace as string)};
```
mkrds, line 220: Instead of assertions you should assign an interface for arguments of this function in line 211

src/components/typography/typography.spec.tsx
```diff
@@ -588,7 +597,7 @@ describe("Typography", () => {
 });
 
 describe("List", () => {
-  it.each([
+  it.each<[keyof JSX.IntrinsicElements, string]>([
```
mkrds, line 600: I think you could also do something like this:

```
  ["ul", "square"] as const,
  ["ol", "decimal"] as const
```

src/components/typography/list.component.tsx
robinzigmond: agreed - except I don't think anything other than `li` is valid for list items (if the parent component is `ul` or `ol`) so maybe it's best just not to expose `as` in the props interface for `ListItem`

src/components/typography/list.component.tsx
robinzigmond: It's also worth noting that `as?: keyof JSX.IntrinsicElements` is already in `TypographyProps`, so if we really did want to allow any such element name to be supplied then there would be no point adding this to `ListProps`

src/components/typography/typography.component.tsx
robinzigmond: I don't think we should assert these (and similar props below) `as string` anyway, because the props. interface allows them to be undefined, which is a valid value for the CSS properties as well (ie they can just not be set). I'm not sure how best to fix the TS errors here but it looks like for some reason TS thinks all these props are `unknown` so I'd rather figure out how to fix this issue than assert props are `strings` which might not be.

src/components/typography/typography.spec.tsx
robinzigmond: I think the `as const` assertion on the whole array of the two 2-tuples is probably the best way to go here?

src/components/typography/list.component.tsx
nineteen88: @mkrds Yeah this is fair - I was battling Typescript giving non-descript error messages realised a semi solution but yeah it would make sense to just be more specific for sure.

@robinzigmond As I recall I couldn't remove the linting typescript errors unless I defined the "as" again for ListProps. So yeah I thought that would be the case too. Perhaps I'm wrong though

src/components/typography/typography.component.tsx
nineteen88: Yeah, will take a look Robin - the linting tool was being here too so just kept it to simple assertions

src/components/typography/list.component.tsx
nineteen88: Setting as to be specific elements like this causes conflicts with the as prop in general. Also, following feedback from Jamie we shouldn't be so restrictive to limit it to specific strings as a user should be able to pass in their own web component. So with that said the type for this is now 'React.ElementType'.

src/components/typography/typography.stories.mdx
```diff
@@ -34,97 +35,7 @@ It's possible to change the text position, using the spacing props. See the prop
...
-        <ListItem>
-          Bread <Typography variant="b">500g</Typography>
-        </ListItem>
-        <ListItem>
-          Sugar <Typography variant="b">1Kg</Typography>
-        </ListItem>
-      </List>
-    </>
-  </Story>
+  <Story name="variants" story={stories.VariantsStory} />
```
robinzigmond, line 38: I'm not sure if it was deliberate or not to remove the `parameters={{ info: { disable: true }}`? I'm not sure what it does exactly but it seems to hide the props tables inside the "controls" tab in the addons. Not sure if that's good or not but we should probably be consistent and not change it here - unless there's a good reason for it that I'm missing!

(Note also, after recent experience on something similar, it might not work to leave it here in JSX and you might need to add it in the `.tsx` file as `VariantStory.parameters = { info: { disable: true }}` and the same with the other story.)

robinzigmond (approved):

mkrds (approved):

DlgSHi (approved):

PR #5508, opened 2022-09-28
refactor(card): convert to typescript (FE-4774)

Parsium:
### Proposed behaviour

- Refactor `Card` and its subcomponents `CardRow`, `CardFooter` and `CardColumn` to typescript
- Update `Card` documentation to reflect the current version of the component
- Move stories to dedicated stories file written in typescript CSF - to allow for type checking - and import them into docs written in MDX. 
- Refactor draggable Card story so it can be rewritten in typescript. Replace custom colours with design tokens.
- Deprecate `dataRole` prop and replace with `data-role` for consistency with other Carbon components.

https://user-images.githubusercontent.com/18368713/192796546-0f0ce403-9f31-47ee-900a-23122bcf1f52.mov

- Convert test stories file from MDX to typescript CSF

### Current behaviour

- `Card` and its subcomponents are all written in javascript
- All stories are written in MDX - making it harder to flag prop type errors.

### Checklist

#### QA

### UX QA Testing instructions
 
- In storybook, check updated `with draggable` story visuals and the Carbon components it uses (under Show Code option) to verify if story complies with Design System standards.

### QA Testing instructions

- Verify no visual or functional regressions have been introduced from this refactor compared to `master`.
- Verify storybook starts correctly with no console errors.


src/components/card/card.style.ts
```diff
@@ -1,16 +1,23 @@
...
 };
 
-const StyledCard = styled.div`
+export interface StyledCardProps extends MarginProps {
+  cardWidth: string;
+  interactive: boolean;
+  draggable: boolean;
+  spacing: CardSpacing;
+  /** @private @ignore */
+  theme?: ThemeObject;
```
robinzigmond, line 18: same comment about `theme` prop here that I left on another component.

src/components/card/index.ts
```diff
@@ -1,4 +1,5 @@
 export { default, default as Card } from "./card.component";
+export type { CardProps } from "./card.component";
 export { default as CardColumn } from "./card-column";
```
robinzigmond, line 3: should we not also export the props interfaces rom all the subsidiary components as well? (We used to provide these in `index.d.ts`.)

src/components/typography/typography.component.js
```diff
@@ -246,3 +246,4 @@ Typography.propTypes = {
 
 Typography.displayName = "Typography";
 export default Typography;
+Typography.displayName = "Typography";
```
robinzigmond, line 249: I'm sure you didn't mean to include the `displayName` twice! I assume a merge conflict incorrectly handled?

src/components/typography/typography.d.ts
```diff
@@ -50,7 +50,7 @@ export interface TypographyProps extends SpaceProps, ColorProps {
 
 declare function Typography(
   attrs: StyledComponentProps<
-    "div",
+    "span",
```
robinzigmond, line 53: while "span" looks correct to me based on the Typography component code, I'm not quite sure what's motivating this change as part of this PR on Card?

src/components/card/card-row/card-row.style.ts
```diff
@@ -0,0 +1,35 @@
...
+import { CardSpacing } from "../card.config";
+
+export interface StyledCardRowProps extends PaddingProps {
+  /**
+   * Spacing prop is set in Card and defines the padding for the CardRow (the first CardRow has no padding by default).
+   * For more granular control of CardRow padding these can be over-ridden by Padding props from styled-system.
+   */
+  spacing: CardSpacing;
+  /** @ignore @private */
+  theme?: ThemeObject;
```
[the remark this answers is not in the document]
Parsium: The `theme` prop does not necessarily have to be there, since the object is not used in the component's code. However, we need to set the theme in `defaultProps` otherwise styled system will use its default spacing scale when calculating custom padding, whereas we want to use our own using design tokens.

src/components/card/index.ts
Parsium: Good point 👍🏼

robinzigmond (dismissed):

src/components/card/card-footer/card-footer.component.tsx
```diff
@@ -0,0 +1,34 @@
...
+  children,
+  transparent,
+  ...rest
+}: CardFooterProps) => {
+  return (
+    <StyledCardFooter
+      key="card-footer"
+      data-element="card-footer"
+      spacing={spacing}
+      transparent={!!transparent}
```
edleeks87, line 24: `transparent` is already a boolean so doesn't need `!!`

edleeks87: Actually looking into this a bit more I don't think we can go with this approach. We've missed documenting the `variant` prop in the tables but we do have an [example](https://carbon.sage.com/v/109.2.2/iframe.html?id=card--more-examples-of-card-footer&viewMode=docs#more-cardfooter--examples) of it being used so I think we will need to keep using `variant="transparent"` for now and document it in storybook

src/components/card/card-footer/card-footer.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <div id="non-interactive">View Stripe Dashboard</div>
+      </CardFooter>
+    ).toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("renders background transparent when transparent prop is passed", () => {
+    const cardFooter = mount(
+      <CardFooter transparent>
```
edleeks87, line 34: see comment above

src/components/card/card-footer/card-footer.style.ts
```diff
@@ -17,11 +15,18 @@ const paddingSizes = {
   large: "20px 48px",
 };
 
-const StyledCardFooter = styled.div`
+export interface StyledCardFooterProps extends SpaceProps {
+  /** Predefined size of CardFooter for applying padding. For more granular control, this prop can be over-ridden by the spacing props from styled-system */
+  spacing: CardSpacing;
+  /** Render with transparent background */
+  transparent: boolean;
```
edleeks87, line 22: see comment above

src/components/box/box.component.tsx
```diff
@@ -80,3 +80,4 @@ Box.defaultProps = {
 };
 
 export default Box;
+Box.displayName = "Box";
```
edleeks87, line 83: For sake of consistency can you move the `displayName` above the export please?

edleeks87 (changes requested):
IMHO we should avoid making a breaking change (currently the commit doesn't include that)  and use the `variant="transparent"` approach for now. The type could be `"default" | "transparent"` for example

robinzigmond (approved):

edleeks87 (dismissed):

robinzigmond (dismissed):

edleeks87 (dismissed):

edleeks87 (approved):

PR #5509, opened 2022-09-28
refactor(checkbox):  convert files to typescript (FE-4776)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


mkrds:
I think it would be best if you'd also move all the stories from `.stories.mdx` to `.stories.tsx`

Check my decimal PR

mkrds (approved):

src/components/checkbox/checkbox-group.spec.tsx
```diff
@@ -54,26 +51,31 @@ describe("CheckboxGroup", () => {
...
-  ])("when the %s prop is set", (propName, propValue, expectedPropName) => {
-    it("then it should be passed to the underlying Fieldset component", () => {
-      expect(
-        render({ [propName]: propValue })
-          .find("Fieldset")
-          .prop(expectedPropName || propName)
-      ).toBe(propValue);
-    });
-  });
+  ] as TestConfig)(
```
Parsium, line 62: What do you think about breaking this test fixture down into one or several smaller tests to verify whether the group legend element is rendered, with the correct styles, etc?
Given this approach is testing implementation details, I feel this would give us better confidence in these tests in the long run.

src/components/checkbox/checkbox-test.stories.tsx
```diff
@@ -1,19 +1,24 @@
...
+import { Checkbox, CheckboxProps } from ".";
 
-<Meta
-  title="Checkbox/Test"
-  parameters={{
+interface StoryProps {
+  labelSpecialCharacters: string;
+  fieldHelpSpecialCharacters: string;
+  labelHelpSpecialCharacters: string;
+}
```
Parsium, line 11:
```suggestion
interface StoryProps {
  labelSpecialCharacters?: string;
  fieldHelpSpecialCharacters?: string;
  labelHelpSpecialCharacters?: string;
}
```
As you are setting these to `undefined` in `Default.args` I think we should make these props optional

src/components/checkbox/checkbox.style.ts
```diff
@@ -10,11 +9,20 @@ import StyledLabel, {
...
 } from "../../__internal__/label/label.style";
 import StyledValidationIcon from "../../__internal__/validations/validation-icon.style";
-import StyledFormField from "../../__internal__/form-field/form-field.style";
-import StyledIcon from "../icon/icon.style";
 import baseTheme from "../../style/themes/base";
-
-const CheckboxStyle = styled.div`
+import { ValidationProps } from "../../__internal__/validations";
+
+interface CheckboxStyleProps extends ValidationProps, MarginProps {
```
Parsium, line 15:
```suggestion
interface StyledCheckboxProps extends ValidationProps, MarginProps {
```
In my opinion I think this would be good for consistency with other components, though this is minor and happy to accept otherwise

Parsium:
I agree with @mkrds's point on refactoring the `.mdx` stories to typescript if possible - I think we would benefit from being able to catch potential prop bugs in our stories.

src/components/checkbox/checkbox.style.ts
grabkowski: I agree, I'll change that

src/components/checkbox/checkbox-group.spec.tsx
grabkowski: I agree with you that there tests do look a bit confusing, but if you look at the resulting tests, everything looks clear what is being tested. As for the rendering of the legend and it's styles, we already have it covered by Chromatic snapshots.

src/components/checkbox/checkbox.component.tsx
```diff
@@ -0,0 +1,139 @@
+import React, { useContext } from "react";
+import { MarginProps } from "styled-system";
+import CheckboxStyle from "./checkbox.style";
```
Parsium, line 3:
```suggestion
import StyledCheckbox from "./checkbox.style";
```

Parsium (commented):
Hi @grabkowski, recently @robinzigmond and @DipperTheDan noticed chromatic was flagging some new stories in their builds, which eventually lead us to finding out that story parameters included in the MDX file were not being recognised by storybook. I noticed you have also defined the story parameters using the same pattern so I just wanted to make you aware of this.

To fix this, we need to move the the `parameters` for each exported story into the TSX story file instead (as shown in this [storybook recipe](https://github.com/storybookjs/storybook/blob/master/addons/docs/docs/recipes.md#csf-stories-with-mdx-docs)). 

Would you be able to add these changes too please? I would suggest defining the `storyName` in the TSX file as well just for consistency.

Parsium (approved):

src/components/checkbox/checkbox-validations.stories.tsx
```diff
@@ -0,0 +1,311 @@
...
+      id="checkbox_info"
+      info="Message"
+      key="checkbox_info"
+      label="Example checkbox (info)"
+      name="checkbox_info"
+    />
+  </>
+);
+
+export const StringValidationWithTooltipPosition: ComponentStory<
```
mkrds, line 31: Instead of having multiple stories that are nearly identical, you could always define 1 template story and then generate new ones based on that template. 

Take a look at `DefaultStory` and for example `Disabled`, `Empty` and `Required` stories to see how it can be done
https://github.com/Sage/carbon/pull/5517/files#diff-145ae5850fb206056a2ce09868044bd4f1bd5e96a4b55b35982d95ddef9efe95R7

src/components/checkbox/checkbox.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <Checkbox
+      label="Large"
+      key="checkbox-large"
+      name="checkbox-large"
+      size="large"
+    />
+  </>
+);
+
+export const Disabled: ComponentStory<typeof Checkbox> = () => (
```
mkrds, line 35: Same as in `checkbox-validations.stories.tsx` file you could use a template for that

src/components/checkbox/checkbox-validations.stories.tsx
grabkowski: unfortunately we cannot do that in this case, as every checkbox has different argument (error, warning, info).

mkrds (dismissed):

Parsium (dismissed):

divyajindel (changes requested):

mkrds (approved):

Parsium (approved):

divyajindel (approved):

PR #5512, opened 2022-09-29
feat(box): add support for the gap css property when component is displayed flex or inline-flex FE-5357

### LINKED ISSUE #5434: Update Box component to have a gap prop
### Desired behaviour

Box component to have a prop which enables customisation of the gap property in CSS. See info: https://css-tricks.com/almanac/properties/g/gap/

### Current behaviour

Box has no prop which allows customisation of the gap property in CSS.

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

edleeks87:
fix #5434

### Proposed behaviour

Adds `gap` prop to the `Box` component. The prop accepts numbers from 0-8 or valid CSS strings. If a number is passed it will be multiplied by 8 to generate a px value. The prop only works when the component has `display` `flex` or `inline-flex`.

image

### Current behaviour

No support for `gap` CSS property when `Box` is flex container

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/romantic-jackson-wrt9no?file=/src/index.js


src/components/box/box.stories.mdx
```diff
@@ -279,5 +365,17 @@ If no `scrollVariant` prop is set then the scroll bar will resort to the browser
...
       required: false,
     },
+    {
+      name: "boxSizing",
+      type: { summary: "content-box | border-box" },
+      description: "Set the box-sizing attribute of the Box component",
+      required: false,
+    },
+    {
+      name: "tabIndex",
```
DipperTheDan, line 375: Good spot. I totally forgot to add this when I added tabIndex 🙄

DipperTheDan (approved):

src/components/box/box.stories.mdx
edleeks87: all good there was a few missing

src/components/box/box.spec.tsx
```diff
@@ -8,9 +8,19 @@ import {
...
+  ScrollVariant,
+  BoxSizing,
+  AllowedNumericalValues,
+  Gap,
+} from "./box.component";
 import boxConfig from "./box.config";
 
+const numericalGapValues = Array.from({ length: 9 }).map(
+  (_, i) => i as AllowedNumericalValues
+);
```
robinzigmond, line 22: I'm not going to suggest a change as I'm not sure how else to do this (convert a union type to an array of all its values - note than the other way round is easier so that might be something to consider), but this is potentially dangerous. If we ever add or remove allowed numeric values from the type (not likely I know but who's to know what design changes might happen in future?), this won't have any effect on `numericalGapValues` which will always be the array of integers from 0 to 8.

I think I prefer the alternative of defining an array explicitly and generating the type from that.

src/components/box/box.spec.tsx
```diff
@@ -71,4 +81,13 @@ describe("Box", () => {
       assertStyleMatch({ boxSizing }, wrapper);
     }
   );
+
+  it.each<Gap>([...numericalGapValues, "20px", "20%"])(
+    "has styles applied when gap is set to %s",
+    (gapValue) => {
+      const wrapper = mount(<Box display="flex" gap={gapValue} />);
+      const gap = boxConfig.gap(gapValue);
+      assertStyleMatch({ gap }, wrapper);
```
robinzigmond, line 90: I'm not sure whether this is a useful test. It's basically testing "are we calling `boxConfig.gap(gapValue)` and applying that to the CSS `gap` property?".

I guess that's still a useful thing to check but I'd want to also include tests for `boxConfig.gap` to make sure that's doing what it should.

src/components/box/box.stories.mdx
```diff
@@ -119,6 +119,94 @@ Please note that in order to see the `position: fixed` Box, you will need to vie
   </Story>
 </Canvas>
 
+### Gap
+
+The `Box` component also supports the `gap` property when its `display` property is set to either 
+`flex` or `inline-flex`. See the [prop table](#props) below for more information on the types it accepts
+
+<Canvas>
```
robinzigmond, line 127: in might be worth demonstrating the 2-value syntax (eg `gap="10px 20px"`) as part of this story?

.storybook/utils/styled-system-props.js
```diff
@@ -751,7 +766,9 @@ const StyledSystemProps = ({
     rows.push(...generateStyledSystemWidthProps(defaults));
   }
   if (flexBox) {
-    rows.push(...generateStyledSystemFlexBoxProps(defaults));
+    rows.push(
+      ...generateStyledSystemFlexBoxProps(defaults, flexBox === "with gap")
```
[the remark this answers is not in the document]
robinzigmond: Another thing to think about here, which might make the above redundant: since this is our own prop and not actually part of styled system, maybe it should be in the "main" props table rather than one of the styled system ones? (Since most of those are "CSS-y" things like `overflowWrap` anyway.)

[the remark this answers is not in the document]
edleeks87: I did consider using a `hasGap` prop but decided it wasn't as needed we just checking for the "truthyness" of it, personally think either approach is fine but chose the current as this is likely to be the only time we need to update the flexbox prop table. Wrt to where it's listed in the prop table, I prefer it in the `flexbox` subsection as it will only work when the Box is flexed etc; the other props, on the other hand` are independent.

src/components/box/box.component.tsx
```diff
@@ -35,13 +37,16 @@ export interface BoxProps
   boxSizing?: BoxSizing;
   /** Allows a tabindex to be specified */
   tabIndex?: number | string;
+  /** Sets the gap between children of a flex box container. Accepts any valid CSS string or any number 0 - 8 which will be multiplied by 8 and result in a px value */
+  gap?: Gap;
```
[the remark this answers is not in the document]
edleeks87: You are completely right, I completely forgot to add the props for setting the individual gaps, will get this updated.

src/components/box/box.spec.tsx
edleeks87: I'll change it so `const gap = typeof gapValue === number ? `${gapValue * 8}px` : gapValue` but don't feel there's any more value in writing tests for the config function past that

src/components/box/box.spec.tsx
edleeks87: If we remove/ or add to the type we will need to update the map so the tests would either fail or we would not have full coverage

src/components/box/box.stories.mdx
edleeks87: I'll add a demo of the row and column gap  props being used as you rightly pointed out they need to be included in the interface.

.storybook/utils/styled-system-props.js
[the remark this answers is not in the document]
edleeks87: I've ended up amending as it was trivial to do

src/components/box/box.spec.tsx
edleeks87: End up updating it after thinking about it a bit more, still think any changes would have been caught but changing the approach to this will allow it to be caught faster

robinzigmond (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #5529, opened 2022-10-05
Sage/DSY 1671 progress tracker update

DobroTora:
### Proposed behaviour

Screenshot 2022-11-23 at 11 05 19

Screenshot 2022-11-23 at 11 05 58
Screenshot 2022-11-23 at 11 07 01

The scope of this task is to update Progress Tracker to be aligned with the latest version of the design from Design System by applying the design tokens for colours. Making sure that component matches the latest design requirements from Design System.

Please Note that the Vertical orientation of the Progress Tracker is going to be deprecated.

### Current behaviour
Screenshot 2022-11-23 at 10 39 41
Screenshot 2022-11-23 at 10 40 01
Screenshot 2022-11-23 at 10 40 22
Screenshot 2022-11-23 at 10 48 58
Screenshot 2022-11-23 at 10 49 22

### Checklist

#### QA

### Testing instructions


src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -49,6 +54,13 @@ const ProgressTracker = ({
     }
   }, [barRef, orientation]);
 
+  if (!isDeprecationWarningTriggered && isVertical) {
+    isDeprecationWarningTriggered = true;
+    Logger.deprecate(
+      "The `isVertical` prop of `ProgressTracker` is now deprecated."
+    );
+  }
+
```
Parsium, line 63:
```suggestion
  if (!isDeprecationWarningTriggered && orientation) {
    isDeprecationWarningTriggered = true;
    Logger.deprecate(
      "The `orientation` prop of `ProgressTracker` is now deprecated."
    );
  }
```
It's the `orientation` prop were deprecating here. This will also need a unit test to check the warning is displayed in the console when the prop is provided. See this [similar example ](https://github.com/Sage/carbon/blob/9b29fd1c9245932d0a07b2378b3e774b7c1634eb/src/components/card/card.spec.tsx#L210) with `Card` for how to achieve this.

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -86,8 +98,14 @@ const ProgressTracker = ({
...
             </StyledValue>
+            {/* <StyledValuesLabel>of</StyledValuesLabel> */}
+            {/* {maxProgressLabel === "100%" ? (
+              <StyledValuesLabel>of</StyledValuesLabel>
+            ) : (
+              " "
+            )} */}
             <StyledValue isMaxValue>
-              {label(maxProgressLabel, "100%")}
+              {label(maxProgressLabel, "100%" && description)}
```
Parsium, line 108: I appreciate you are still working on the labelling currently so this is subject to change, but in this example I would personally refactor the function so all the logic is handled inside it instead of passing a boolean expression like this. This helps makes the function easier to read and understand.

Parsium (commented):
This can be done in retrospect once this work is complete, but we should amend the commit messages later on since the commits marked as a `style` change are really be `feat` or `fix` changes.  A `style` commit would be more concerned with code formatting rather than css changes.

DobroTora:
> This can be done in retrospect once this work is complete, but we should amend the commit messages later on since the commits marked as a `style` changes should really be `feat` or `fix` changes. A `style` commit would be more concerned with code formatting rather than css styling.

Noted, thank you for that James :)

harpalsingh:
@DobroTora I've just changed this back to Pending UXQA, so you don't self add the label , but could you also please list in the PR the changes done?

DobroTora:
> @DobroTora I've just changed this back to Pending UXQA, so you don't self add the label , but could you also please list in the PR the changes done?

Done.

src/components/textbox/textbox.stories.mdx
```diff
@@ -25,14 +24,6 @@ Captures a single line of text.
...
 ```
 
-## Designer Notes
-
-- Use placeholder text to give the user examples of data formats (e.g. AB123456C for a UK National Insurance number).
-- Use prefixes if your data always begins with a certain sequence (e.g. a UK VAT number usually starts with ‘GB’).
-- If content in a textbox is never editable, think about removing the field border so it appears as static text.
-- You can disable a textbox, but try to avoid this. If you need to, make it clear what the user needs to do in order to activate the textbox.
-- Use wider fields for longer data (e.g. an address line), and narrower fields for shorter data (e.g. a postcode), to give the user a clue about the data expected.
-
```
Parsium, line 35: I'm not sure why the designer notes have been removed here, is there some context behind this or was this a mistake?

Parsium, line 35: May I ask what the reason was for removing these designer notes? Was this intentional or a mistake?

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -25,7 +28,9 @@ const ProgressTracker = ({
   "aria-valuetext": ariaValueText,
   size = "medium",
   length = "256px",
+  error = false,
```
Parsium, line 31: Personally I think ultimately `error` should be of type `string` only, as we would want to display an error message where "error" is in the example below, so users can clearly see what caused the error.

Screenshot 2022-11-30 at 15 34 02

This would be similar to how validation styling used by components such as [Textbox](https://carbon.sage.com/?path=/docs/textbox--validations-string-new-design#new-designs-validation) and [Textarea](https://carbon.sage.com/?path=/docs/textarea--validations-string-new-design#new-designs-validation):

Screenshot 2022-11-30 at 15 36 24

Screenshot 2022-11-30 at 15 36 01

I realise you have implemented the design you have been given here, which does not clearly show how error messages should be handled. Could you feedback this to the designer please? As I am wary this would be inconsistent with Design System's latest approach to validation messages.

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -157,256 +168,54 @@ This is an example of a `ProgressTracker` component with `currentProgressLabel`
...
 
 <Canvas>
   <Story name="default and custom label values">
     <div
       style={{ display: "flex", flexDirection: "column", alignItems: "center" }}
     >
-      <ProgressTracker progress={15} currentProgressLabel="$50" />
-      <ProgressTracker mt={2} progress={50} maxProgressLabel="$200" />
+      <ProgressTracker progress={50} currentProgressLabel="50%" />
+      <ProgressTracker mt={2} progress={50} currentProgressLabel="Step 3 of 5" />
```
Parsium, line 205: While this visually achieves the original design, I don't think this is the right approach here, since `currentProgressLabel` was designed to represent the current level of progress, not both the current and completed levels - in other words "Step 3" should really be passed to `currentProgressLabel` and "5" to `maxProgressLabel`.

Personally I think we would need to update the existing labelling props and release this update as a breaking change, since the current props don't seem to work well with the redesign you have been given.

One potential solution would be to split `showDefaultLabels` into two separate props - one for the current level of progress, and the other for the completed level of progress. That way we would be able to handle the case where both labels need to be rendered with the "of" keyword between them.

@nicktitchmarsh @edleeks87 what do you think about this?

src/components/textbox/textbox.stories.mdx
DobroTora: it was definitely a mistake. Thank you @Parsium

src/components/progress-tracker/progress-tracker.component.js
DobroTora: So this a part of the larger discussions we had before -  how the error is going to be indicated by the progress tracker itself and how do we define error. I am not sure if tracking progress and error handling on input fields have the same idea ...

src/components/progress-tracker/progress-tracker.stories.mdx
DobroTora: I think that a potential solution looks like a great idea.

src/components/progress-tracker/progress-tracker.component.js
edleeks87: I think it would be fine to leave it as a boolean for now but expand it to support string as well if the need arises. As there's no designs for the latter it doesn't make sense to handle for it yet imo

src/components/button/button.style.ts
```diff
@@ -87,6 +86,7 @@ const StyledButton = styled.button<StyledButtonProps>`
     justify-content: center;
     vertical-align: middle;
     outline-offset: 0;
+    ${space}
```
edleeks87, line 89: did this need moving?

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -66,15 +78,15 @@ const ProgressTracker = ({
...
+      return showDefaultLabels ? defaultValue : null;
     };
 
     return (
       <StyledValuesLabel position={labelsPosition} isVertical={isVertical}>
         {isVertical && direction === "up" && (
           <>
             <StyledValue isMaxValue>
-              {label(maxProgressLabel, "100%")}
+              <>{label(maxProgressLabel, "100%")}</>
```
edleeks87, line 89: do you need a fragment here?

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -66,15 +78,15 @@ const ProgressTracker = ({
       if (value) {
         return value;
       }
-      return showDefaultLabels ? defaultValue : undefined;
+      return showDefaultLabels ? defaultValue : null;
```
edleeks87, line 81: Does it need to be null? It doesn't matter just wondering why the change was made

src/components/progress-tracker/progress-tracker.component.js
edleeks87: This comment doesn't look like it's been addressed but the conversation has been resolved? I'm inclined to agree with James

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -85,19 +84,23 @@ This is an example of a `ProgressTracker` component with different colors depend
...
-        size="large"
         mt={2}
         progress={100}
         currentProgressLabel="100%"
       />
+      <ProgressTracker
+        mt={2}
+        progress={100}
+        error={true}
+        currentProgressLabel={'error'}
```
edleeks87, line 102:
```suggestion
        currentProgressLabel="error"
```

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -143,6 +147,13 @@ It is possible to override the default `labelsPosition` to `"bottom"` when in a
         progress={100}
         showDefaultLabels
       />
+      <ProgressTracker
+        mt={2}
+        labelsPosition="bottom"
+        progress={100}
+        error={true}
+        currentProgressLabel={'error'}
```
edleeks87, line 155:
```suggestion
        currentProgressLabel="error"
```

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -112,9 +115,10 @@ percentages.
     <div
       style={{ display: "flex", flexDirection: "column", alignItems: "center" }}
     >
-      <ProgressTracker progress={15} showDefaultLabels />
+      <ProgressTracker progress={15} showDefaultLabels descriprion="Descriptive text(optional)" />
       <ProgressTracker mt={2} progress={50} showDefaultLabels />
       <ProgressTracker mt={2} progress={100} showDefaultLabels />
+      <ProgressTracker mt={2} progress={100} error={true} currentProgressLabel={'error'} />
```
edleeks87, line 121:
```suggestion
      <ProgressTracker mt={2} progress={100} error={true} currentProgressLabel="error" />
```

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -157,256 +168,54 @@ This is an example of a `ProgressTracker` component with `currentProgressLabel`
...
       <ProgressTracker
         mt={2}
         progress={100}
         currentProgressLabel="$200"
-        maxProgressLabel="$200"
+      />
+      <ProgressTracker
+        mt={2}
+        progress={100}
+        error={true}
```
edleeks87, line 189:
```suggestion
        error
```

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -143,6 +147,13 @@ It is possible to override the default `labelsPosition` to `"bottom"` when in a
         progress={100}
         showDefaultLabels
       />
+      <ProgressTracker
+        mt={2}
+        labelsPosition="bottom"
+        progress={100}
+        error={true}
```
edleeks87, line 154:
```suggestion
        error
```

src/components/progress-tracker/progress-tracker.stories.mdx
edleeks87: If we were building from scratch James approach sounds reasonable, but as it's an existing component I'd like to avoid a breaking change. I don't really mind Debra's approach of using the `currentProgressLabel` but we could update it to be a `node` so that they compose it with `Typography` components to achieve the above

src/components/progress-tracker/progress-tracker.style.js
```diff
@@ -49,17 +51,21 @@ const StyledProgressBar = styled.span`
 `;
 
 const StyledValue = styled.span`
-  color: var(--colorsUtilityYin090);
+  flex-basis: 30px;
+  margin-right: 8px;
   ${({ isMaxValue }) => css`
-    ${isMaxValue && `color: var(--colorsUtilityYin055);`}
-    ${!isMaxValue && `font-weight: bold;`}
+    ${!isMaxValue &&
```
edleeks87, line 57: Lets do this as a ternary instead so we're checking `isMaxValue` once

```
${({ isMaxValue}) => isMaxValue ? "color: var(--colorsUtilityYin055);"  : "font-weight: bold;"}
```

src/components/progress-tracker/progress-tracker.style.js
```diff
@@ -127,11 +134,18 @@ function getHeight(size) {
   }
 }
 
-function getInnerBarColour(progress) {
+function getBackgroundColour(progress, error) {
+  if (error === true) return "var(--colorsSemanticNegative500)";
```
edleeks87, line 138:
```suggestion
  if (error) return "var(--colorsSemanticNegative500)";
```

src/components/progress-tracker/progress-tracker.style.js
```diff
@@ -127,11 +134,18 @@ function getHeight(size) {
...
   if (progress >= 100) return "var(--colorsSemanticPositive500)";
   return "var(--colorsSemanticNeutral500)";
 }
 
+function getBorderColour(progress, error) {
+  if (error === true) return "var(--colorsSemanticNegative500)";
```
edleeks87, line 144:
```suggestion
  if (error) return "var(--colorsSemanticNegative500)";
```

src/components/progress-tracker/progress-tracker.component.js
DobroTora: It hasn't been address as I am not sure what would be the best approach to this one....

src/components/progress-tracker/progress-tracker.stories.mdx
DobroTora: Can you elaborate a bit more on your approach @edleeks87 ?

src/components/button/button.style.ts
```diff
@@ -77,6 +77,7 @@ type StyledButtonProps = SpaceProps &
   };
 
 const StyledButton = styled.button<StyledButtonProps>`
+  ${space}
```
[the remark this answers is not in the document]
DobroTora: Done.

mkrds:
You should also modify `ProgressTrackerProps` interface `progress-tracker.d.ts` file considering that the prop interface has changed

src/components/progress-tracker/progress-tracker.spec.js
```diff
@@ -275,237 +315,271 @@ describe("ProgressBar", () => {
...
-            assertStyleMatch(
-              {
-                textAlign: "right",
-              },
-              wrapper.find(StyledValuesLabel),
-              { modifier: `${StyledValue}` }
-            );
-          }
-        });
+describe("when isMaxValue and description are set for labels", () => {
```
mkrds, line 461:
```suggestion
describe("when showDefaultLabels and description props are set", () => {
```

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -49,6 +54,13 @@ const ProgressTracker = ({
     }
   }, [barRef, orientation]);
 
+  if (!isDeprecationWarningTriggered && orientation) {
+    isDeprecationWarningTriggered = true;
+    Logger.deprecate(
+      "The `orientation` prop of `ProgressTracker` is now deprecated."
```
mkrds, line 60:
```suggestion
      "The `orientation` and `direction` props of `ProgressTracker` are now deprecated."
```

src/components/progress-tracker/progress-tracker.stories.mdx
```diff
@@ -460,3 +265,19 @@ This example shows the automatic calculation of `aria-valuenow` when not provide
...
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object
+to the [i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
+
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "progressTracker.ofLabel",
```
edleeks87, line 277: `progressTracker.ofLabel` -> `progressTracker.of`

edleeks87 (changes requested):
Changes on the whole look good to me but there's a few things that need updating.:

- I'm a bit confused as to why `date/__internal__/utils.spec.` was updated as part of this, was it needed?

- You'll also need to update `progress-tracker.test` as there are now failing cypress tests following these changes.

- The commit message needs to have include the scope eg `feat(progress-tracker): implement component redesign`. The text for the `BREAKING CHANGE` is fine so can remain

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -156,21 +153,21 @@ ProgressTracker.propTypes = {
   length: PropTypes.string,
   /** Current progress (percentage). */
   progress: PropTypes.number,
+  /** If error occurs. */
+  error: PropTypes.bool,
   /** Flag to control whether the default value labels (as percentages) should be rendered. */
   showDefaultLabels: PropTypes.bool,
   /** Value to display as current progress. */
-  currentProgressLabel: PropTypes.string,
+  currentProgressLabel: PropTypes.node,
```
mkrds, line 161:
```suggestion
  currentProgressLabel: PropTypes.string,
```

src/components/progress-tracker/progress-tracker.stories.mdx
DobroTora: Hi @edleeks87  - first change was a result of failing tests - it has been resolved and branch has been updated. I reached out to Igor Dlugosh regarding failing Cypress tests. Once all tests are passing I will amend the commit message accordingly. Thank you for your feedback :)

edleeks87 (dismissed):

mkrds (dismissed):

src/components/progress-tracker/progress-tracker.test.js
```diff
@@ -78,6 +81,24 @@ context("Tests for ProgressTracker component", () => {
       progressTrackerLine().should("have.css", "background-color", color);
     });
 
+    it("render component with error pros", () => {
```
sianford, line 84: Should read `error prop` not `error pros`

LaiaSage:
Hi @DobroTora 
We are aware of the tiny hole but we consider it a hard-to-be-noticed issue so we can move forward for now with it 👌 Thanks a lot!

edleeks87 (approved):

DlgSHi (approved):

edleeks87 (approved):

mkrds (approved):

PR #5531, opened 2022-10-06
refactor(search): convert files to typescript (FE-4828)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/search/search.component.tsx
```diff
@@ -92,31 +148,32 @@ const Search = ({
...
-      onFocus(e);
+      onFocus(event);
     }
   };
 
-  let buttonProps = {};
   if (searchButton && onClick) {
     buttonProps = {
-      onClick: (ev) => {
+      onClick: () => {
```
robinzigmond, line 172: This seems to be a change in behaviour, at least in theory. The old version called the `onClick` prop with the `name` and `id` from the actual event target, whereas this new version will always provide the `name` and `id` from the props provided to `Search`.  I don't think these will be the same because the event target will be the search button itself (or possibly a child of it), not the actual Search component.

So I don't think this works, and even if it does, we shouldn't be changing behaviour in a TS refactor.

Please let me know if I'm misunderstanding something! :)

src/components/search/search.spec.tsx
```diff
@@ -545,66 +523,26 @@ describe("Search", () => {
...
+        defaultValue: "",
+        id: "Search",
+        name: "Search",
+      });
       const input = wrapper.find(TextBox);
       expect(input.prop("iconTabIndex")).toEqual(-1);
     });
   });
 
-  describe("tab index should be set on search button", () => {
```
robinzigmond, line 574: Is there a good reason to simply remove these tests? (Happy to accept it if there is!)

src/components/search/search.stories.mdx
```diff
@@ -27,154 +28,36 @@ import Search from "carbon-react/lib/components/search";
...
 <Canvas>
-  <Story
-    name="with SearchButton and Custom Width using px"
-    parameters={{ info: { disable: true } }}
-  >
-    <Search defaultValue="Here is some text" searchButton searchWidth="375px" />
-  </Story>
-</Canvas>
-
-### Search without Search Button and custom width
```
robinzigmond, line 121: Again, is there a good reason to remove all these stories you've removed?

src/components/search/search.spec.tsx
grabkowski: to further explain: tabIndex on the button is changed to `-1`, when the search input has no text. The search button is not rendered at all when the input has no text and it's not focused, so there is no need to pass that prop down, as the button will be removed as soon as user move focus out of the input.

src/components/search/search.stories.mdx
grabkowski: I think that we don't have to show every prop combination to our users, that makes the documentation over bloated. I think that was done for testing purposes. All of those combinations could be tested in the test story.

src/components/search/search.component.tsx
grabkowski: This refactor is not changing anything, as the event could only originate in the search button, or button icon and none of them have either `id` or `name` attribute, so value of these were always defaulting to those of the component.

robinzigmond: I see now - thanks for clarifying this.

robinzigmond (dismissed):

robinzigmond (dismissed):

DipperTheDan (dismissed):

robinzigmond (approved):

DipperTheDan (dismissed):

robinzigmond (approved):

DlgSHi (approved):

edleeks87 (approved):

PR #5535, opened 2022-10-07
test(navigation-bar): refactor for cypress tests

DlgSHi:
### Proposed behaviour

Refactor Cypress tests for the `Navigation-Bar` component to use new cypress-component-react framework for testing.
Add `navigation-bar.test.js`. 

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [x] Check if the `navigation-bar.test.js` file passed


src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,169 @@
...
+        navigationBar().should("have.css", "position", position);
+      }
+    );
+
+    it.each([[25], [100], [-100]])(
+      "should render NavigationBar component with offset prop set to %s",
+      (offset) => {
+        CypressMountWithProviders(<NavigationBarComponent offset={offset} />);
+
+        navigationBar().should("have.attr", "offset", offset);
```
robinzigmond, line 98: I don't think this is the right way to test the `offset` prop. This only works because the rendered output does include `offset="25px"` as an attribute - I'm not actually sure what this does but I think it's a unintended consequence of spreading `...rest` onto the underlying element.

What this prop is actually doing is providing the value of the CSS `top` or `bottom` property - which one depends on the `orientation`. (See line 61 of `navigation-bar.style.ts`.) So that CSS is what we need to test.

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,169 @@
...
+    );
+
+    it.each([["top"], ["bottom"]])(
+      "should render NavigationBar component with orientation prop set to %s",
+      (orientation) => {
+        CypressMountWithProviders(
+          <NavigationBarComponent orientation={orientation} />
+        );
+
+        navigationBar().should("have.attr", "orientation", orientation);
```
robinzigmond, line 109: as in the above comment, I think it's just an accident that the rendered HTML element has an `orientation` attribute with this value. Orientation is actually just used to decide whether to use CSS `top` or `bottom` for the `offset` value.

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,169 @@
...
+      }
+    );
+
+    it.each([
+      ["fixed", "be.visible"],
+      ["sticky", "not.be.visible"],
+    ])(
+      "should render NavigationBar component with position prop set to %s and check it works as expected",
```
robinzigmond, line 117:
```suggestion
      "should render NavigationBar component with position prop set to %s and work as expected",
```

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,169 @@
...
+          <NavigationBarComponent navigationType={navigationType} />
+        );
+
+        navigationBar().should("have.css", "background-color", color);
+      }
+    );
+
+    it.each([
+      [true, "not.exist"],
+      [false, "be.visible"],
```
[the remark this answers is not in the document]
DlgSHi: Hi @robinzigmond, you are right. But, cypress cares if the element is visible or not (but it definitely should be rendered in DOM).

src/components/navigation-bar/navigation-bar.test.js
DlgSHi: why? we are providing the `offset` to some values and then just asserting that those values were rendered properly. Like in screenshot below. 
Screenshot 2022-10-10 at 07 50 34

src/components/navigation-bar/navigation-bar.test.js
DlgSHi: updated

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,169 @@
...
+          <NavigationBarComponent isLoading={boolean}>
+            {testData}
+          </NavigationBarComponent>
+        );
+
+        cy.contains(testData).should(assertion);
+      }
+    );
+
+    it.each([["fixed"], ["sticky"]])(
```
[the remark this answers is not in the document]
DlgSHi: applied

src/components/navigation-bar/navigation-bar.test.js
robinzigmond: hi Igor, I must have explained it badly :-)

What I meant was that, while the `offset` prop is indeed setting an `offset` attribute on the HTML element, which is what you're testing, as far as I can tell this attribute is an accident and doesn't actually do anything. The important thing about what this prop does is that it sets the CSS `top` or `bottom` value (which one depends on the `orientation` prop) to the value of this prop. So I think that's what we should be testing.

By all means ask James if you want a second opinion, as he's the one who implemented this component so will probably have more understanding than me. I could be wrong but I expect he'll be surprised like I was to find `offset` and `orientation` as actual attributes in the DOM. (They are passed through by styled components automatically, presumably because they're technically valid attributes - although I couldn't find them documented anywhere...)

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+    it.each(["top", "bottom"])(
+      "should render NavigationBar component with orientation prop set to %s",
+      (orientation) => {
+        CypressMountWithProviders(
+          <NavigationBarComponent orientation={orientation} />
+        );
+
+        navigationBar().should("have.attr", "orientation", orientation);
```
Parsium, line 120:
```suggestion
          <NavigationBarComponent position="fixed" orientation={orientation} />
        );

        navigationBar().should("have.attr", "position", "fixed");
        navigationBar().should("have.css", orientation, "0px");
```
I would amend this test to do the above since we want to check if the navigation bar is rendered at the top or the bottom of the viewport when a particular value is passed to `orientation`.

Like @robinzigmond mentions in his comment the prop itself is a custom React prop, so there must be a separate bug occurring `orientation` is being rendered on the `<nav>` DOM element.

src/components/navigation-bar/navigation-bar.test.js
```diff
@@ -0,0 +1,180 @@
...
+            <NavigationBarComponent
+              offset={`${offset}px`}
+              orientation={orientation}
+              position="fixed"
+            />
+          );
+          navigationBar().then(($el) => {
+            expect($el).to.have.attr("offset").to.equals(`${offset}px`);
+            expect($el).to.have.css(orientation).to.equals(`${offset}px`);
+          });
```
Parsium, line 108:
```suggestion
          navigationBar().then(($el) => {
            expect($el).to.have.css(orientation).to.equals(`${offset}px`);
          });
```
I don't think this line is needed since `offset` is a custom React prop here. It seems like @robinzigmond mentions that we may have a separate bug where an `offset` attribute is being added onto the `<nav>` element. This shouldn't affect the visual functionality but it would be good to investigate why this is being added to the DOM element.

robinzigmond (dismissed):

robinzigmond (approved):

Parsium (approved):

ZhuoyuJin (approved):

src/components/navigation-bar/navigation-bar.test.js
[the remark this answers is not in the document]
DianeGit22: I think it would be good to verify the same property for both scenarios.  I would think visibility would be the one to check. But you could also check both visibility and existence for both scenarios?

PR #5548, opened 2022-10-12
fix(tags): update TagProps interface so only the data tag types are exported

### LINKED ISSUE #5534: Sidebar props not picked up correctly by TypeScript
### Current behaviour

For the Sidebar component, the prop types are not enforced correctly by the TypeScript compiler. This can be seen on both VS Code and on Codesandbox in the codesandbox demo below. Note how:
- the incorrect type given to `disableEscKey` is not flagged as an error at all by TS in the editor (although it does generate an error from PropTypes at runtime)
- the type of all props, seen if you hover over them, is simply inferred from what is entered, rather than picked up from the actual type declaration. Eg that `disableEscKey` prop is thought to be a `string` (when it should be `boolean`), while `open` is inferred as just the literal `true` when it again should be `boolean`.
- there is no autocompletion if you want to add another prop

Assuming this is consistent behaviour and not just a quirk of the editor (it seems to be consistent as I'm seeing the same on both VS Code and CodeSandbox - and even if it's specific to those VSC is a very common editor!), this needs to be fixed as it defeats the point of using TypeScript for this specific component.

### Expected behaviour

TypeScript should check the types and flag an error for props of incorrect types, as well as show the correct declared type on hover for each prop.

https://codesandbox.io/s/cool-satoshi-pdp5rp?file=/src/index.tsx

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

111.1.0

### Design Tokens Version

_No response_

Chrome

MacOS

on further investigation, I think this is due to the following in the `DataProps` interface, which are extended by `TagProps` which is in turn extended by `SidebarProps` (as well as other component props interfaces):

`[restKeys: string]: any;`

In particular, if I comment that out locally then the types on Sidebar seem to work correctly.

While other components use `TagProps`, Sidebar seems to be the only one that also uses `forwardRef`, and this does indeed seem to be the problem (due to how React itself types forwardRef). See [this CSB demo](https://codesandbox.io/s/agitated-williamson-z1cy52?file=/src/App.tsx) for how adding `[rest: string]: any` to a props type causes type checking to fail when used with forwardRef, but not without.

I think the only solution is to remove the `[restKeys: string]: any;` and find some other way to achieve what we intended by that.

### Confidentiality

edleeks87:
fix #5534

### Proposed behaviour

Updates the `TagProps` interface to ensure only the `data-` props are exported and extended in the public interfaces. A new private `RestProps` interface has been added to support passing the rest props object into the `tags` util. The previous implementation was preventing `intellisense` from displaying the correct types for the `Sidebar` component; instead it was being assigned `any`.

### Current behaviour

The `TagProps` interface supports any type via `[restKeys: string] any` which prevents intellisense from computing the types of components and instead displaying `any`

### Checklist

#### QA

### Testing instructions


src/components/sidebar/sidebar.spec.tsx
```diff
@@ -25,14 +25,7 @@ describe("Sidebar", () => {
   beforeEach(() => {
     spy = jest.fn();
     wrapper = mount(
-      <Sidebar
-        open
-        title="Test"
```
robinzigmond, line 30: not really related to this file but I notice that although there is no `title` prop on `Sidebar` (we call it `header` instead - I think in other components `title` is a simple string which gets included in a styled header but Sidebar doesn't have it for some reason), it is referred to in the doc comments for some other props like the `aria-label`. It would be good to update that either now or later.

robinzigmond: Also this isn't currently visible as the prop table doesn't show up (it doesn't on master either). Not sure why as it's definitely not any of the usual reasons like not exporting the component as a named export.

src/components/textbox/textbox.component.tsx
```diff
@@ -237,8 +237,6 @@ export const Textbox = ({
             validationRedesignOptIn ? undefined : validationIconId
           }
           validationRedesignOptIn={validationRedesignOptIn}
-          size={size}
-          readOnly={readOnly}
```
robinzigmond, line 241: looking at the internals of `FormField`, while it doesn't actually use either of these props, there is an invariant message telling people to use `readOnly` if trying to use `disabled` with validation props. We should update that (although I'm not sure what to!) if `readOnly` isn't actually accepted. (We should probably just add `readOnly` as a new prop that gets passed to the HTML element.)

src/components/sidebar/sidebar.spec.tsx
edleeks87: Good spot, I'll address these in this PR 👍

src/components/textbox/textbox.component.tsx
edleeks87: I think leaving the invariant in `FormField` is correct as it will mean all inputs will have this guard and the prop isn't used by `FormField` so it not being passed isn't and issue. I've also confirmed `readOnly` is passed to the input element.

robinzigmond (dismissed):

Parsium (approved):

robinzigmond (dismissed):

Parsium (approved):

src/components/sidebar/sidebar.component.tsx
```diff
@@ -62,7 +67,7 @@ export interface SidebarProps extends PaddingProps, TagProps {
     | "large"
     | "extra-large";
   /** an optional array of refs to containers whose content should also be reachable by tabbing from the sidebar */
-  focusableContainers?: React.MutableRefObject<HTMLElement>[];
+  focusableContainers?: CustomRefObject<HTMLElement | null>[];
```
robinzigmond, line 70:
```suggestion
  focusableContainers?: CustomRefObject<HTMLElement>[];
```

the `null` option shouldn't be needed here as that's taken care of in `CustomRefObject`

src/components/dialog/dialog.component.tsx
```diff
@@ -82,7 +86,7 @@ export interface DialogProps extends ModalProps, TagProps {
   /** Padding to be set on the Dialog content */
   contentPadding?: ContentPaddingInterface;
   /** an optional array of refs to containers whose content should also be reachable by tabbing from the dialog */
-  focusableContainers?: React.MutableRefObject<HTMLElement | null>[];
+  focusableContainers?: CustomRefObject<HTMLElement | null>[];
```
robinzigmond, line 89:
```suggestion
  focusableContainers?: CustomRefObject<HTMLElement>[];
```

the `null` option shouldn't be needed here as that's taken care of in `CustomRefObject`

Parsium (dismissed):

robinzigmond (approved):

DlgSHi (dismissed):

robinzigmond (approved):

DlgSHi (approved):

PR #5553, opened 2022-10-13
Add new component to Carbon: Button Minor

DobroTora:
### Proposed behaviour
Screenshot 2022-11-23 at 14 12 13
Screenshot 2022-11-23 at 14 13 39
Screenshot 2022-11-23 at 14 14 02
Screenshot 2022-11-23 at 14 14 27
Screenshot 2022-11-23 at 14 14 45
Screenshot 2022-11-23 at 14 15 01
Screenshot 2022-11-23 at 14 15 16
Screenshot 2022-11-23 at 14 15 32

### Current behaviour
The Button Minor is a new component. The boolean prop isMinor is added to render a Button Minor variant. Storybook stories has been updated to reflect the differences between Minor and Major varinats.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/button-minor/button-minor-test.stories.tsx
+++ b/src/components/button-minor/button-minor-test.stories.tsx
@@ -1,0 +1,90 @@
+import React from "react";
+import { action } from "@storybook/addon-actions";
+import { ButtonProps } from "components/button/button.component";
+import ButtonMinor from "./button-minor.component";
+import { ICONS } from "../icon/icon-config";
+import {
+  BUTTON_SIZES,
+  BUTTON_VARIANTS,
+  BUTTON_ICON_POSITIONS,
+} from "../button/button.config";
+
+export default {
+  title: "Button Minor/Test",
+  includeStories: "DefaultStory",
+  parameters: {
+    info: { disable: true },
+    chromatic: { disable: true },
+  },
+};
+const commonArgTypesButtonMinor = {
+  size: {
+    options: BUTTON_SIZES,
+    control: { type: "select" },
+  },
+  buttonType: {
+    options: BUTTON_VARIANTS,
+    control: {
+      type: "select",
+    },
+  },
+  iconType: {
+    options: [...ICONS, ""],
+    control: {
+      type: "select",
+    },
+  },
+  iconPosition: {
+    options: BUTTON_ICON_POSITIONS,
+    control: { type: "select" },
+  },
+  href: {
+    control: {
+      type: "text",
+    },
+  },
+};
+const commonArgsButtonMinor = {
+  size: "medium",
+  subtext: "",
+  buttonType: "secondary",
+  fullWidth: false,
+  disabled: false,
+  destructive: false,
+  noWrap: false,
+  href: undefined,
+  iconPosition: "before",
+};
+export const DefaultStory = (args: ButtonProps) => (
+  <ButtonMinor onClick={action("click")} {...args}>
+    Example Button
+  </ButtonMinor>
+);
+export const Default = (args: ButtonProps) => <ButtonMinor {...args} />;
+export const ButtonMinorCustom = (props: ButtonProps) => (
+  <ButtonMinor {...props}>Example Button</ButtonMinor>
+);
+export const ButtonMinorDifferentTypes = (props: ButtonProps) => {
+  return (
+    <div>
+      <ButtonMinor buttonType="primary" {...props}>
+        Primary
+      </ButtonMinor>
+      <ButtonMinor buttonType="secondary" {...props}>
+        Secondary
+      </ButtonMinor>
+      <ButtonMinor buttonType="tertiary" {...props}>
+        Tertiary
+      </ButtonMinor>
+    </div>
+  );
+};
+DefaultStory.story = {
+  name: "default",
+  args: {
+    ...commonArgsButtonMinor,
+  },
+  argTypes: {
+    ...commonArgTypesButtonMinor,
+  },
+};
--- a/src/components/button-minor/button-minor.component.tsx
+++ b/src/components/button-minor/button-minor.component.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+import StyledButtonMinor from "./button-minor.style";
+import { ButtonProps } from "../button";
+
+export const ButtonMinor = ({
+  buttonType = "secondary",
+  size = "medium",
+  ...rest
+}: ButtonProps) => (
+  <StyledButtonMinor size={size} buttonType={buttonType} {...rest} />
+);
+
+ButtonMinor.displayName = "ButtonMinor";
+
+export default ButtonMinor;
--- a/src/components/button-minor/button-minor.spec.tsx
+++ b/src/components/button-minor/button-minor.spec.tsx
@@ -1,0 +1,111 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
+import Icon from "../icon";
+import {
+  ButtonProps,
+  ButtonTypes,
+  SizeOptions,
+} from "../button/button.component";
+import ButtonMinor from "./button-minor.component";
+import { assertStyleMatch } from "../../__spec_helper__/test-utils";
+
+const render = (props: ButtonProps, renderer: typeof shallow = shallow) => {
+  return renderer(<ButtonMinor {...props} />);
+};
+
+describe("when no props other than children are passed into the component", () => {
+  it("renders the default props and children", () => {
+    const wrapper = render({ children: "foo" });
+    expect(wrapper.contains(<Icon type="filter" />)).toBeFalsy();
+    expect(wrapper.props().buttonType).toEqual("secondary");
+  });
+});
+
+const minorSizesPadding: [SizeOptions, string][] = [
+  [
+    "small",
+    "var(--spacing000) var(--spacing100) var(--spacing000) var(--spacing100)",
+  ],
+  ["medium", "var(--spacing100)"],
+  ["large", "var(--spacing100)"],
+];
+
+interface VariantColorProperties {
+  background?: string;
+  borderColor?: string;
+  color?: string;
+}
+
+const minorColors: [ButtonTypes, VariantColorProperties][] = [
+  [
+    "primary",
+    {
+      background: "var(--colorsActionMinor500)",
+      borderColor: "var(--colorsActionMinorTransparent)",
+      color: "var(--colorsActionMinorYang100)",
+    },
+  ],
+  [
+    "secondary",
+    {
+      background: "transparent",
+      borderColor: "var(--colorsActionMinor500)",
+      color: "var(--colorsActionMinor500)",
+    },
+  ],
+  [
+    "tertiary",
+    {
+      background: "transparent",
+      borderColor: "transparent",
+      color: "var(--colorsActionMinor500)",
+    },
+  ],
+  [
+    "darkBackground",
+    {
+      borderColor: "transparent",
+    },
+  ],
+];
+
+describe("Button Minor", () => {
+  it.each(minorSizesPadding)(
+    "when size is %s the padding is %s",
+    (size, padding) => {
+      const wrapper = mount(<ButtonMinor size={size}>Foo</ButtonMinor>);
+      assertStyleMatch({ padding }, wrapper);
+    }
+  );
+
+  it.each(minorColors)(
+    "when buttonType is %s, renders with correct styling",
+    (buttonType, styles) => {
+      const wrapper = mount(
+        <ButtonMinor buttonType={buttonType}>Foo</ButtonMinor>
+      );
+      assertStyleMatch({ ...styles }, wrapper);
+    }
+  );
+
+  it("when destructive prop is passed, renders with destructive styling", () => {
+    const wrapper = mount(<ButtonMinor destructive>foo</ButtonMinor>);
+    assertStyleMatch(
+      {
+        background: "transparent",
+        color: "var(--colorsSemanticNegative500)",
+      },
+      wrapper
+    );
+  });
+  it("when disabled prop is passed, renders with disabled styling", () => {
+    const wrapper = mount(<ButtonMinor disabled>foo</ButtonMinor>);
+    assertStyleMatch(
+      {
+        background: "transparent",
+        color: "var(--colorsActionMajorYin030)",
+      },
+      wrapper
+    );
+  });
+});
--- a/src/components/button-minor/button-minor.stories.mdx
+++ b/src/components/button-minor/button-minor.stories.mdx
@@ -1,0 +1,241 @@
+import { Meta, Story, Canvas } from "@storybook/addon-docs";
+
+import ButtonMinor from ".";
+import Button from "../button";
+import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
+import * as stories from "./button-minor.stories";
+
+<Meta title="Button Minor" parameters={{ info: { disable: true } }} />
+
+# Button Minor
+
+<a
+  target="_blank"
+  href="https://zeroheight.com/2ccf2b601/p/4121af-button-minor/b/03680d"
+  style={{ color: '#007E45', fontWeight: 'bold', textDecoration: 'underline' }}
+>
+  Product Design System component - Button Minor Component
+</a>
+
+- A minor button is a user interface button rather than a primary action button.
+- It is less prominent than a major button.
+
+## Contents
+
+- [Quick Start](#quick-start)
+- [Examples](#examples)
+- [Props](#props)
+
+## Quickstart
+
+```javascript
+import ButtonMinor from "carbon-react/lib/components/button-minor";
+```
+
+## Examples
+
+### Primary Buttons
+
+For the single most prominent call to action on the page (e.g. Save, Submit, Continue).
+
+<Canvas>
+  <Story 
+    name="primary" 
+    story={stories.PrimaryButton}
+  />
+</Canvas>
+
+Primary buttons can be destructive.
+
+<Canvas>
+  <Story
+    name="primary/destructive"
+    story={stories.PrimaryDestructiveButton}
+  />
+</Canvas>
+
+Primary buttons can be disabled.
+
+<Canvas>
+  <Story
+    name="primary/disabled"
+    story={stories.PrimaryDisabledButton}
+  />
+</Canvas>
+
+Primary buttons can have an icon positioned before or after the text.
+
+<Canvas>
+  <Story
+    name="primary/icon"
+    story={stories.PrimaryIconButton}
+  />
+</Canvas>
+
+Primary buttons can be `fullWidth`.
+
+<Canvas>
+  <Story
+    name="primary/full-width"
+    story={stories.PrimaryFullWidthButton}
+  />
+</Canvas>
+
+Primary buttons can be set into `noWrap` mode to prevent text wrapping.
+
+<Canvas>
+  <Story
+    name="primary/noWrap"
+    story={stories.PrimaryNoWrapButton}
+  />
+</Canvas>
+
+### Secondary Buttons
+
+Less prominent, there can be multiple secondary buttons on a page.
+Secondary buttons are transparent, rather than white.
+
+<Canvas>
+  <Story
+    name="secondary"
+    story={stories.SecondaryButton}
+  />
+</Canvas>
+
+Secondary buttons can be destructive.
+
+<Canvas>
+  <Story
+    name="secondary/destructive"
+    story={stories.SecondaryDestructiveButton}
+  />
+</Canvas>
+
+Secondary buttons can be disabled.
+
+<Canvas>
+  <Story
+    name="secondary/disabled"
+    story={stories.SecondaryDisabledButton}
+  />
+</Canvas>
+
+Secondary buttons can have an icon positioned before or after the text.
+
+<Canvas>
+  <Story
+    name="secondary/icon"
+    story={stories.SecondaryIconButton}
+  />
+</Canvas>
+
+Secondary buttons can be `fullWidth`.
+
+<Canvas>
+  <Story
+    name="secondary/full-width"
+    story={stories.SecondaryFullWidthButton}
+  />
+</Canvas>
+
+Secondary buttons can be set into `noWrap` mode to prevent text wrapping.
+
+<Canvas>
+  <Story
+    name="secondary/noWrap"
+    story={stories.SecondaryNoWrapButton}
+    parameters={{ chromatic: { disable: true } }}
+  />
+</Canvas>
+
+### Tertiary Buttons
+
+Tertiary Buttons or 'ghost' buttons are used for reversing actions, like 'Cancel' or 'Back'.
+
+<Canvas>
+  <Story
+    name="tertiary"
+    story={stories.TertiaryButton}
+  />
+</Canvas>
+
+Tertiary buttons can be destructive.
+
+<Canvas>
+  <Story
+    name="tertiary/destructive"
+    story={stories.TertiaryDestructiveButton}
+  />
+</Canvas>
+
+Tertiary buttons can be disabled.
+
+<Canvas>
+  <Story
+    name="tertiary/disabled"
+    story={stories.TertiaryDisabledButton}
+  />
+</Canvas>
+
+Tertiary buttons can have an icon positioned before or after the text.
+
+<Canvas>
+  <Story
+    name="tertiary/icon" 
+    story={stories.TertiaryIconButton} />
+</Canvas>
+
+Tertiary buttons can be `fullWidth`.
+
+<Canvas>
+  <Story
+    name="tertiary/full-width"
+    story={stories.TertiaryFullWidthButton}
+  />
+</Canvas>
+
+Tertiary buttons can be set to `noWrap` mode to prevent text wrapping.
+
+<Canvas>
+  <Story
+    name="tertiary/noWrap"
+    story={stories.TertiaryNoWrapButton}
+    parameters={{ chromatic: { disable: true } }}
+  />
+</Canvas>
+
+### Icon Only Button
+
+Buttons can be rendered with just an icon.
+
+<Canvas>
+  <Story
+    name="icon only button"
+    story={stories.IconOnlyButton}
+  />
+</Canvas>
+
+Icon only buttons can also display a tooltip message when the icon is hovered over. This can be acheived by passing a string to the `iconTooltipMessage` prop.
+
+<Canvas>
+  <Story
+    name="icon only button with tooltip"
+    story={stories.IconOnlyWithTooltipButton}
+    parameters={{ chromatic: { disable: true } }}
+  />
+</Canvas>
+
+
+## Props
+
+Options shared between all of the above types of buttons. When setting padding we recommend using either the `p`, `py`
+or `px` props to ensure the spacing within the button is applied evenly.
+
+### Button Minor
+
+<StyledSystemProps
+  of={Button}
+  spacing
+  defaults={{ pt: "1px", pb: "1px", px: "24px" }}
+  noHeader
+/>
--- a/src/components/button-minor/button-minor.test.js
+++ b/src/components/button-minor/button-minor.test.js
@@ -1,0 +1,416 @@
+import React from "react";
+
+import {
+  Default as ButtonMinor,
+  ButtonMinorCustom,
+  ButtonMinorDifferentTypes,
+} from "./button-minor-test.stories";
+
+import * as stories from "./button-minor.stories.tsx";
+
+import { BUTTON_SIZES, BUTTON_ICON_POSITIONS } from "../button/button.config";
+
+import {
+  buttonSubtextPreview,
+  buttonDataComponent,
+} from "../../../cypress/locators/button";
+
+import { cyRoot, icon, tooltipPreview } from "../../../cypress/locators";
+import { positionOfElement } from "../../../cypress/support/helper";
+import { CHARACTERS } from "../../../cypress/support/component-helper/constants";
+import CypressMountWithProviders from "../../../cypress/support/component-helper/cypress-mount";
+import { useJQueryCssValueAndAssert } from "../../../cypress/support/component-helper/common-steps";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+
+const destructive = "rgb(203, 55, 74)";
+const transparent = "rgba(0, 0, 0, 0)";
+
+context("Test for Button Minor component", () => {
+  describe("Check props for Button Minor component", () => {
+    it("should render Button Minor with aria-label prop", () => {
+      CypressMountWithProviders(
+        <ButtonMinorCustom aria-label="cypress-aria" />
+      );
+
+      buttonDataComponent().should("have.attr", "aria-label", "cypress-aria");
+    });
+
+    it.each(testData)(
+      "should render Button Minor label using %s special characters",
+      (label) => {
+        CypressMountWithProviders(<ButtonMinor>{label}</ButtonMinor>);
+
+        buttonDataComponent().should("have.text", label);
+      }
+    );
+
+    it.each(testData)(
+      "should render Button Minor subtext with %s special characters",
+      (subtext) => {
+        CypressMountWithProviders(
+          <ButtonMinorCustom size="large" subtext={subtext} />
+        );
+
+        buttonSubtextPreview().should("have.text", subtext);
+      }
+    );
+
+    it.each(testData)(
+      "should render Button Minor name using %s special characters",
+      (name) => {
+        CypressMountWithProviders(<ButtonMinorCustom name={name} />);
+
+        buttonDataComponent().should("have.attr", "name", name);
+      }
+    );
+
+    it.each(testData)(
+      "should render Button Minor id using %s special characters",
+      (id) => {
+        CypressMountWithProviders(<ButtonMinorCustom id={id} />);
+
+        buttonDataComponent().should("have.attr", "id", id);
+      }
+    );
+
+    it.each(testData)(
+      "should render tooltip message with %s special characters",
+      (tooltipMessage) => {
+        CypressMountWithProviders(
+          <ButtonMinor
+            iconType="bin"
+            iconTooltipMessage={tooltipMessage}
+            m="100px"
+          />
+        );
+
+        buttonDataComponent().children().last().realHover();
+        tooltipPreview().should("have.text", tooltipMessage);
+        cyRoot().realHover({ position: "topLeft" });
+      }
+    );
+
+    it.each([
+      [BUTTON_SIZES[0], 32],
+      [BUTTON_SIZES[1], 40],
+      [BUTTON_SIZES[2], 48],
+    ])("should render Button Minor in %s size", (size, minHeight) => {
+      CypressMountWithProviders(<ButtonMinor size={size}>{size}</ButtonMinor>);
+
+      buttonDataComponent().should("have.css", "min-height", `${minHeight}px`);
+    });
+
+    it.each([
+      [BUTTON_ICON_POSITIONS[0], "right"],
+      [BUTTON_ICON_POSITIONS[1], "left"],
+    ])(
+      "should set position to %s for icon in a button",
+      (iconPosition, margin) => {
+        CypressMountWithProviders(
+          <ButtonMinorCustom iconType="add" iconPosition={iconPosition} />
+        );
+
+        icon().should("have.css", `margin-${margin}`, "8px");
+      }
+    );
+
+    it("should render Button Minor with full width", () => {
+      CypressMountWithProviders(<ButtonMinorCustom fullWidth />);
+
+      buttonDataComponent().then(($el) => {
+        useJQueryCssValueAndAssert($el, "width", 1365);
+      });
+    });
+
+    it("should render Button Minor with href", () => {
+      CypressMountWithProviders(
+        <ButtonMinorCustom href="https://carbon.sage.com/" />
+      );
+
+      buttonDataComponent().should(
+        "have.attr",
+        "href",
+        "https://carbon.sage.com/"
+      );
+    });
+
+    it.each([
+      [true, "white-space"],
+      [false, "flex-wrap"],
+    ])(
+      "should render the Button Minor text with noWrap prop set to %s",
+      (booleanState, cssValue) => {
+        const assertion = booleanState ? "nowrap" : "wrap";
+
+        CypressMountWithProviders(
+          <ButtonMinor noWrap={booleanState}>
+            {" "}
+            Long long long long long text{" "}
+          </ButtonMinor>
+        );
+
+        buttonDataComponent().should("have.css", cssValue, assertion);
+      }
+    );
+
+    it("should check Button Minor is disabled", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes disabled />);
+
+      for (let i = 0; i < 3; i++) {
+        buttonDataComponent()
+          .eq(i)
+          .should("be.disabled")
+          .and("have.attr", "disabled");
+      }
+    });
+
+    it("should check Button Minor is enabled", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes />);
+
+      for (let i = 0; i < 3; i++) {
+        buttonDataComponent().eq(i).should("be.enabled");
+      }
+    });
+
+    it("should check Button Minor is destructive", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes destructive />);
+
+      cyRoot().realHover({ position: "topRight" });
+
+      buttonDataComponent()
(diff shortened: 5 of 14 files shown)


LaiaSage:
Hi @DobroTora , gonna try my first PR review. Don't hesitate to complain if I do something wrong 😅

1. **Minor buttons** are a different component that **Major buttons**. They look different and we recommend them for different purposes. So please, if possible, follow the same categorisation (and sub-sections inside) we have on Zeroheight: https://zeroheight.com/2ccf2b601/p/10358e-button-major/b/66a1c8 
image
2. Copy also the content of the components (we always use 'Button').
3. The height token outcome for Minor S size should be 32px (now 37px).
4. The padding for L size Minor button should correspond to 16px.
5. The padding for M size Minor button should correspond to 12px.
6. Please, fix all those sizes in the other sections.
7. There is a misalignment on the destructive section.
8. On the secondary and tertiary sections, the M and L padding are different but also wrong.
9. Wrong colour token on the secondary and tertiary destructive.
You can inspect the designs here: https://www.figma.com/file/Tau6qv5iyr8wthWNqYktej/DS-Official-Components?node-id=20917%3A368071
image
and call me every time you need.
Thanks!

src/components/button/button.spec.tsx
```diff
@@ -102,6 +133,57 @@ describe("Button", () => {
     });
   });
 
+  describe( "Button Minor Padding",() => {
```
edleeks87, line 136:
```suggestion
  describe("Button Minor Padding",() => {
```

src/components/button/button.spec.tsx
```diff
@@ -102,6 +133,57 @@ describe("Button", () => {
...
+      it.each(isMinorSizesPadding)("when size is %s the padding is %s", (size, padding) => {
+          const wrapper = mount(
+            <Button isMinor size={size}>Foo</Button>
+          )
+          assertStyleMatch({padding}, wrapper)    
+      });
+  });
+
+  describe( "Button Minor Colors ",() => {
```
edleeks87, line 145:
```suggestion
  describe("Button Minor Colors ",() => {
```

src/components/button/button.spec.tsx
```diff
@@ -102,6 +133,57 @@ describe("Button", () => {
...
+    it.each(isMinorColors)("when buttonType is %s", (buttonType, styles) => {
+      const wrapper = mount(
+        <Button isMinor buttonType={buttonType} >Foo</Button>
+      )
+      assertStyleMatch({...styles}, wrapper)    
+  });
+
+  describe( "Button Minor Colors for destructive and disabled",() => {
+    it('matches the style for the Button Minor destructive', () => {
```
edleeks87, line 154: This isn't testing is `destructive` is true (see above for proposed changes above)

src/components/button/button.spec.tsx
```diff
@@ -102,6 +133,57 @@ describe("Button", () => {
...
+      assertStyleMatch(
+        {
+          background: "transparent",
+          color: "var(--colorsActionMinor500)",
+        },
+        wrapper
+      );
+    });
+
+    it('matches the style for the Button Minor disabled', () => {
```
edleeks87, line 168: This isn't testing if `disabled` is true (see above for proposed changes above)

src/components/button/button.spec.tsx
edleeks87: It might be a bit better to scope this into a describe block for all `isMinor` tests

```
describe("isMinor", () => {
  it.each(isMinorSizesPadding)("when size is %s the padding is %s", (size, padding) => {
        const wrapper = mount(
          <Button isMinor size={size}>Foo</Button>
        );
        assertStyleMatch({padding}, wrapper);
    });
    
    it.each(isMinorColors)("when buttonType is %s", (buttonType, styles) => {
        const wrapper = mount(
          <Button isMinor buttonType={buttonType}>Foo</Button>
        );
        assertStyleMatch({...styles}, wrapper);
     });
     
     it('matches the style for the Button Minor destructive', () => {
        const wrapper = mount(
          <Button isMinor buttonType={buttonType} destructive>Foo</Button>
        )
        assertStyleMatch(
          {
           // assert the relevant destructive styles here
          },
          wrapper
        );
    });

    it('matches the style when disabled', () => {
       const wrapper = mount(
          <Button isMinor buttonType={buttonType} disabled>Foo</Button>
        );
        assertStyleMatch(
          {
            // assert relevant disabled styles here
          },
          wrapper
        );
    });
});
```

src/components/button/button.spec.tsx
```diff
@@ -222,6 +304,8 @@ describe("Button", () => {
     );
   });
 
+  describe("When icon type is specified and button has no children", () => {})
```
edleeks87, line 307: empty block can be removed

src/components/button/button.style.ts
```diff
@@ -67,6 +71,30 @@ function stylingForType({
       font-size: 16px;
       min-height: 48px;
     `}
+
+    /* BUTTON MINOR */
+    ${isMinor &&
+    size === "small" &&
+    css`
+      min-height: 32px;
+      padding: var(--spacing000) var(--spacing100) var(--spacing000)
```
edleeks87, line 80: These values will present consumers being able to set their own custom padding values as it is applied after the `space` [styling](https://github.com/Sage/carbon/blob/b9ff6740d3e1987d6361c3f2a446d28c1afa9fad/src/components/button/button.style.ts#L108). They will need to be the other way around imo so they act as defaults that can be overridden

src/components/button/button.style.ts
```diff
@@ -67,6 +71,30 @@ function stylingForType({
       font-size: 16px;
       min-height: 48px;
     `}
+
+    /* BUTTON MINOR */
+    ${isMinor &&
```
edleeks87, line 76: You can just check `isMinor` once here:


```
 ${isMinor && css`
   ${ size === "small" &&
    css`
      min-height: 32px;
      padding: var(--spacing000) var(--spacing100) var(--spacing000) var(--spacing100);
   `}
    ${size === "medium" &&
    css`
      padding-left: var(--spacing150);
      padding-right: var(--spacing150);
   `}
    $size === "large" &&
    css`
      padding-left: var(--spacing200);
      padding-right: var(--spacing200);
    `}
  `}
```

src/components/button/button.test.js
```diff
@@ -46,9 +46,11 @@ context("Test for Button component", () => {
       "should render Button subtext with %s special characters",
       (subtext) => {
         CypressMountWithProviders(
-          <Button size="large" subtext={subtext}>
-            Foo
-          </Button>
+          <>
```
edleeks87, line 49: This change wasn't needed

edleeks87 (changes requested):
FYI the linter has failed as prettier isn't able to format the code you've added, you should be able to identify where the issues are by running `npm run format`/ `npm run lint` locally and seeing which files are updated/flagged.

Can you also squash the commits into a single commit with a message something like:

```feat(button): add support for minor button styling"```

File: src/components/button/button.spec.tsx [hunk 9]
```diff
@@ -102,6 +133,57 @@ describe("Button", () => {
...
+
+  describe( "Button Minor Colors ",() => {
+    it.each(isMinorColors)("when buttonType is %s", (buttonType, styles) => {
+      const wrapper = mount(
+        <Button isMinor buttonType={buttonType} >Foo</Button>
+      )
+      assertStyleMatch({...styles}, wrapper)    
+  });
+
+  describe( "Button Minor Colors for destructive and disabled",() => {
```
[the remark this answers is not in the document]
edleeks87: If you don't make the change describe here https://github.com/Sage/carbon/pull/5553#discussion_r1015293812

src/components/button/button.spec.tsx
edleeks87: If you don't make the change describe here https://github.com/Sage/carbon/pull/5553#discussion_r1015293812

LaiaSage:
Thanks for the fixes @DobroTora 👌

src/components/button/button.test.js
edleeks87: This hasn't been addressed

src/components/minor-button/minor-button.stories.mdx
```diff
@@ -0,0 +1,417 @@
...
+
+## Quickstart
+
+```javascript
+import MinorButton from "carbon-react/lib/components/minor-button";
+```
+
+## Examples
+
+### Minor Primary Buttons
```
edleeks87, line 27:
```suggestion
### Primary
```

src/components/minor-button/minor-button.stories.mdx
```diff
@@ -0,0 +1,417 @@
...
+
+For the single most prominent call to action on the page (e.g. Save, Submit, Continue).
+
+<Canvas>
+  <Story name="minor/primary">
```
edleeks87, line 32:
```suggestion
  <Story name="primary">
```

src/components/minor-button/minor-button.stories.mdx
```diff
@@ -0,0 +1,417 @@
...
+        Large
+      </MinorButton>
+    </>
+  </Story>
+</Canvas>
+
+Minor Primary buttons can be destructive.
+
+<Canvas>
+  <Story name="minor/primary/destructive">
```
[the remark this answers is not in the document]
edleeks87: Same approach for all stories

src/components/button/button.major.stories.mdx
```diff
@@ -23,14 +23,14 @@ import Button from "carbon-react/lib/components/button";
 
 ## Examples
 
-### Primary Buttons
+### Major Primary Buttons
```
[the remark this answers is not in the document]
DobroTora: Yes it was requested to separate button into two stories Major and Minor as per screenshot attached.

src/components/button/button.major.stories.mdx
```diff
@@ -23,14 +23,14 @@ import Button from "carbon-react/lib/components/button";
...
 
 For the single most prominent call to action on the page (e.g. Save, Submit, Continue).
 
 <Canvas>
-  <Story name="primary">
+  <Story name="major/primary">
```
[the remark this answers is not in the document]
DobroTora: Yes, that was the request as per first screenshot from storybook for this PR.

src/components/button/button.major.stories.mdx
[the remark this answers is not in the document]
Parsium: Hi @DobroTora,

I am inclined to agree with @edleeks87 on this, as since we have a new separate `MinorButton` component, adding the "major" reference and scope in the stories for `Button` doesn't really convey very much and could be slightly confusing for new Carbon consumers. 

I do agree in the future we should rename `Button` to `MajorButton` so we are aligned with DS's naming convention as @LaiaSage mentioned, but since this would be a breaking change it would be better to release this with other breaking changes we need to make, so consumers won't have to do multiple Carbon upgrades for trivial changes.

src/components/button/button.major.stories.mdx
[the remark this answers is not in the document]
Parsium: See comment above 😄

src/components/button/button.test.js
```diff
@@ -46,9 +46,11 @@ context("Test for Button component", () => {
...
+            <Button size="large" subtext={subtext}>
+              Foo
+            </Button>
+          </>
```
Parsium, line 53:
```suggestion
            <Button size="large" subtext={subtext}>
              Foo
            </Button>
```
This should work without the fragment

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ).toJSON();
+        assertStyleMatch(
+          {
+            background: "transparent",
+            color: "var(--colorsActionMinor500)",
+          },
+          wrapper
+        );
+      });
+    });
```
Parsium, line 98:
```suggestion
it("when destructive prop is passed, renders with destructive styling", () => {
    const wrapper = mount(<MinorButton destructive>foo</MinorButton>);
    assertStyleMatch(
      {
        background: "transparent",
        color: "var(--colorsSemanticNegative500)",
      },
      wrapper
    );
  });

  it("when disabled prop is passed, renders with disabled styling", () => {
    const wrapper = mount(<MinorButton disabled>foo</MinorButton>);
    assertStyleMatch(
      {
        background: "transparent",
        color: "var(--colorsActionMajorYin030)",
      },
      wrapper
    );
  });
```

I would use `mount` here rather than `TestRenderer`. That way you could also remove the custom `render` function as well. 

I think writing a `render` function would be more of an over-abstraction here with little benefits, since we have a small number of tests.

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+      "when size is %s the padding is %s",
+      (size, padding) => {
+        const wrapper = mount(<MinorButton size={size}>Foo</MinorButton>);
+        assertStyleMatch({ padding }, wrapper);
+      }
+    );
+  });
+
+  describe("Colors", () => {
+    it.each(isMinorColors)("when buttonType is %s", (buttonType, styles) => {
```
Parsium, line 77:
```suggestion
    it.each(isMinorColors)("when buttonType is %s, renders with correct styling", (buttonType, styles) => {
```
I would add this so other maintainers are aware at a glance what the test is specifically asserting

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+describe("Button Minor", () => {
+  describe("Padding", () => {
+    it.each(isMinorSizesPadding)(
+      "when size is %s the padding is %s",
+      (size, padding) => {
+        const wrapper = mount(<MinorButton size={size}>Foo</MinorButton>);
+        assertStyleMatch({ padding }, wrapper);
+      }
+    );
+  });
```
Parsium, line 74:
```suggestion
    it.each(isMinorSizesPadding)(
      "when size is %s the padding is %s",
      (size, padding) => {
        const wrapper = mount(<MinorButton size={size}>Foo</MinorButton>);
        assertStyleMatch({ padding }, wrapper);
      }
    );
```
Personally I would remove the `describe` block here, since we have a small number of tests in this file and the padding tests themselves are quite self-explanatory.

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  describe("Colors", () => {
```
Parsium, line 76:
```suggestion
```
Similarly to a previous comment, I feel the `describe` block doesn't contain lots of tests and the test descriptions themselves are already self-explanatory enough, so we could remove it.

src/components/minor-button/minor-button.stories.mdx
```diff
@@ -0,0 +1,417 @@
...
+import Button from "../button";
+import MinorButton from ".";
+import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
+
+<Meta title="Minor Button" parameters={{ info: { disable: true } }} />
+
+# Minor Button
+
+A Button triggers a single action or event.
+Use it to submit a form (Save), to advance to the next step in a process (Next), or to create a new item (New).
```
Parsium, line 11: What do you think about amending the descriptions like this one to explain why `MinorButton` is different to `Button` (aka button major), similarly to how Design System have done in their documentation? 
I think this would help consumers understand why we have two separate button components and where each type should be used.

Perhaps @LaiaSage may be able to help you with this? 😄

src/components/minor-button/minor-button.spec.tsx
Parsium, line 74:
```suggestion
    it.each(isMinorSizesPadding)(
      "when size is %s the padding is %s",
      (size, padding) => {
        const wrapper = mount(<MinorButton size={size}>Foo</MinorButton>);
        assertStyleMatch({ padding }, wrapper);
      }
    );
```
Since the `describe` block doesn't contain a lot of tests, and the test themselves are fairly self-explanatory, I think we could remove the `describe` block.

src/components/minor-button/minor-button.stories.mdx
DobroTora: Please keep in mind that  PR which is about separating Designer Notes from Carbon is in the pipeline too. I think we would require more Developer's comments for that one? What do you think @Parsium ?

src/components/button/button.major.stories.mdx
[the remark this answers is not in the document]
ljemmo: So to give context to @edleeks87 and @Parsium. Originally when Laia did the UXQA we had minor and major as one component and on one storybook page. We therefore suggested adding major and minor labels to each section would help differentiate  the two buttons.

If minor button is now a separate component, we can remove the reference to major to avoid dev confusion as @Parsium  suggests. We have an carbon/DS alignment stream in the new year where we can update naming then if required. Perhaps we fix all the names in one go as you suggest (all breaking changes).

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -14,6 +14,14 @@ const render = (props: ButtonProps, renderer: any = shallow) => {
   return renderer(<MinorButton {...props} />);
 };
 
+describe("when no props other than children are passed into the component", () => {
+  it("renders the default props and children", () => {
+    const wrapper = render({ children: "foo" });
```
Parsium, line 19:
```suggestion
    const wrapper = mount(<MinorButton>foo</MinorButton>);
```
If you are just rendering the button with a child here, you can do this more simply using `mount`. This would also allow you to remove the `render` function entirely.

src/components/minor-button/minor-button.stories.mdx
```diff
@@ -5,10 +5,12 @@ import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
...
 
-# Minor Button
+# Button
 
-A Button triggers a single action or event.
-Use it to submit a form (Save), to advance to the next step in a process (Next), or to create a new item (New).
+- A minor button is a user interface button rather than a primary action button.
+- It is less prominent than a major button.
+- Use minor buttons when displaying many buttons on a page.
+- Group minor buttons together to form a button bar.
```
Parsium, line 13: Thanks for adding this @DobroTora 👍🏼 

I feel we can tweak the wording a little bit so its clearer, but perhaps it would be better if this is something @harpalsingh or Design System would be able to assist you with 😄

src/components/minor-button/minor-button.spec.tsx
edleeks87: I think it's okay to use the render function for now, it will likely mean the refactor to RTL will be slightly more straightforward as we will be using render functions etc

src/components/minor-button/minor-button.stories.mdx
DobroTora: @Parsium I requested our content writer opinion and check from @emmacassidy

ljemmo: @Parsium just to let you know, we've been deliberately deleting out all design guidance from every carbon page on a separate PR. Part 2 of this work is to cross-link to the design system website where we can maintain designer notes like this in one place. 

With that in mind, i don't think we should be adding this in as it undos the work we are doing elsewhere.

DobroTora: So the content writer - @emmacassidy  suggested to keep 2 first bullet points as we are migrating Designer's Notes into ZH and the other 2 are part of it.

src/components/minor-button/minor-button.spec.tsx
Parsium: Good point 👍🏼 In that case @DobroTora, could you please amend the `render` function's `renderer` parameter to remove the `any` type and replace it with `typeof shallow`? As we are trying to avoid the use of `any` in TypeScript where possible as it means type checks are bypassed.

Parsium (changes requested):
Sorry I just noticed this @DobroTora, please could you also rename `button.major.stories.mdx` back to `button.stories.mdx`?

src/components/minor-button/minor-button.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+import Icon from "components/icon";
+import {
+  ButtonProps,
+  ButtonTypes,
+  SizeOptions,
+} from "../button/button.component";
+import MinorButton from "./minor-button.component";
+import { assertStyleMatch } from "../../__spec_helper__/test-utils";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Parsium, line 12:
```suggestion
```
You can remove this line now we've changed the type of `renderer` :)

src/components/button/button.stories.mdx
```diff
@@ -478,7 +479,7 @@ Buttons can be rendered with just an icon.
       <Button mt={2} destructive ml={2} iconType="bin" aria-label="Delete" />
       <Button
         mt={2}
-        buttonType="dashed"
+        disabled
```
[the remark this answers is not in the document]
DobroTora: Hi @Parsium, since dashed buttons are going to be deleted anyway -  this change was to replace them with another type of button. In that case disabled. Also there are duplicates of props for button types.

src/components/button/button.stories.mdx
```diff
@@ -507,15 +509,16 @@ Icon only buttons can also display a tooltip message when the icon is hovered ov
...
         iconType="bin"
         iconTooltipMessage="This is a tooltip"
         aria-label="Delete"
       />
       <Button
         mt={2}
-        buttonType="dashed"
-        size="large"
         ml={2}
+        disabled
```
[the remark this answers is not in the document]
DobroTora: In that case that was the order requested.

Parsium (dismissed):

edleeks87 (dismissed):
LGTM, you can squash the extra commit later once UX and QA have reviewed/ approved

ljemmo:
@DobroTora **some feedback below:**

- The medium minor button looks like it has the wrong left and right padding applied. It should use the **{spacing.150}** token.
Screenshot 2022-12-12 at 16 26 01

- Minor buttons with icons should respect the same left and right paddings as those without icons. (small should use {spacing.100}, medium should use {spacing.150} and large should use {spacing.200}.
Screenshot 2022-12-12 at 16 30 11

- Default secondary buttons look fine but destructive secondary buttons have the same padding issues.
- Default tertiary buttons look fine but destructive secondary buttons have the same padding issues.
- The minor button appears far away from the major button component in carbon. Please can we name it 'Button: Minor' or something similar so both button types appear next to each other.

src/components/minor-button/minor-button.stories.mdx
Parsium: Ah awesome - thanks for clarifying this @ljemmo @DobroTora 👍🏼

src/components/button-minor/button-minor.stories.mdx
```diff
@@ -0,0 +1,417 @@
+import { Meta, Story, Canvas } from "@storybook/addon-docs";
```
mkrds, line 1: Recently we've switched to having the user facing stories in a separate TS files (to have an additional layer of testing the prop interfaces we provide) which are imported and embedded in `mdx` files like this. We're doing this because typechecking, eslint and intellisense unfortunately does not work in `mdx` files. It would be great if you could refactor this file and move the stories to a new`button-minor.stories.tsx` file.

You can check a `Card` component and `card.stories.mdx` and `card.stories.tsx` files as an example of how we're doing it right now.

mkrds (dismissed):

edleeks87 (approved):

src/components/button-minor/button-minor.stories.mdx
DobroTora: I think it would be great to raise a separate PR for this task.

edleeks87 (commented):
[src/components/button-minor/a11y-utils.js](https://github.com/Sage/carbon/pull/5553/files#diff-55c3e5c0b58a0d187020c8d59a21698c8244b5c7dfefae4ceff6521542cf6c71)

looks like you've pushed an empty file as well

src/components/button-minor/button-minor-test.stories.tsx
```diff
@@ -1,331 +1,70 @@
 import React from "react";
-
 import ButtonMinor from "./button-minor.component";
-
+import {
+  BUTTON_SIZES,
+  BUTTON_VARIANTS,
+  BUTTON_ICON_POSITIONS,
+  // eslint-disable-next-line import/no-useless-path-segments
```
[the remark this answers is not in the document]
sianford: @edleeks87 that was me. I did have it that way but for some reason it the file couldn't be found there. I used the 'Quick fix' option in VSCode and it put that line in.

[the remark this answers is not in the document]
DobroTora: It could have been my VS Code when I have been using prettier.

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+            .should("not.have.value", cssValue, "wrap");
+        }
+      }
+    );
+
+    it("should check Button Minor is disabled", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes disabled />);
+
+      buttonDataComponent()
+        .eq(positionOfElement("first"))
```
DlgSHi, line 193: we can put this in a `loop` and just avoid `first`, `second` and `third` and as well a code duplications

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+      buttonDataComponent()
+        .eq(positionOfElement("third"))
+        .should("be.disabled")
+        .and("have.attr", "disabled");
+    });
+
+    it("should check Button Minor is enabled", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes />);
+
+      buttonDataComponent().eq(positionOfElement("first")).should("be.enabled");
```
DlgSHi, line 209: we can put this in a `loop` and just avoid `first`, `second` and `third` and as well a code duplications

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+      buttonDataComponent()
+        .eq(positionOfElement("second"))
+        .should("be.enabled");
+      buttonDataComponent().eq(positionOfElement("third")).should("be.enabled");
+    });
+
+    it("should check Button Minor is destructive", () => {
+      CypressMountWithProviders(<ButtonMinorDifferentTypes destructive />);
+
+      buttonDataComponent()
```
DlgSHi, line 219: we can put this in a `loop` and just avoid `first`, `second` and `third` and as well a code duplications

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+        .should(
+          "have.css",
+          "background",
+          "rgba(0, 0, 0, 0) none repeat scroll 0% 0% / auto padding-box border-box"
+        )
+        .and("have.css", "border-color", transparent)
+        .and("have.css", "color", destructive);
+    });
+
+    it.each([["noopener"], ["noreferrer"], ["opener"]])(
```
DlgSHi, line 248:
```suggestion
    it.each(["noopener", "noreferrer", "opener"])(
```

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+      "should render Button Minor with rel prop set to %s",
+      (rel) => {
+        CypressMountWithProviders(<ButtonMinor rel={rel}>Foo</ButtonMinor>);
+
+        buttonDataComponent().should("have.attr", "rel", rel);
+      }
+    );
+
+    it.each([["_blank"], ["_self"], ["_parent"], ["_top"]])(
```
DlgSHi, line 257:
```suggestion
    it.each(["_blank", "_self", "_parent", "_top"])(
```

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+      (target) => {
+        CypressMountWithProviders(
+          <ButtonMinor target={target}>Foo</ButtonMinor>
+        );
+
+        buttonDataComponent().should("have.attr", "target", target);
+      }
+    );
+
+    it.each([["add"], ["share"], ["tick"]])(
```
DlgSHi, line 268:
```suggestion
    it.each(["add", "share", "tick"])(
```

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+        });
+    });
+
+    it("should call onKeyDown callback when a keydown event is triggered", () => {
+      CypressMountWithProviders(
+        <ButtonMinor onKeyDown={callback}>Foo</ButtonMinor>
+      );
+
+      buttonDataComponent()
+        .trigger("keydown", { force: true }, keyCode("rightarrow"))
```
DlgSHi, line 318: you can try to use `realPress()` method

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+    });
+
+    it("should call onFocus callback when a focus event is triggered", () => {
+      CypressMountWithProviders(
+        <ButtonMinor onFocus={callback}>Foo</ButtonMinor>
+      );
+
+      buttonDataComponent()
+        .focus()
+        .blur({ force: true })
```
DlgSHi, line 332: focus or blur ?

src/components/button-minor/button-minor.component.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const ButtonMinor = ({
+  buttonType = "secondary",
+  size = "medium",
+  ...rest
+}: ButtonProps) => (
+  <StyledButtonMinor size={size} buttonType={buttonType} {...rest} />
+);
+
+ButtonMinor.displayName = "ButtonMinor";
```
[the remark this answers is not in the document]
edleeks87: ButtonMinor is correct imo as it's how we've named other components

image

src/components/button-minor/button-minor.stories.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+import ButtonMinor from ".";
+import Button from "../button";
+import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
+import * as stories from "./button-minor.stories";
+
+<Meta
+  title="Button Minor"
+  parameters={{ info: { disable: true }, chromatic: { disable: true } }}
```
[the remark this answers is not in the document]
edleeks87: agree this should be applied on a per component basis as most will need to be captured by chromatic

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+      "should render the Button Minor text without wrapping when noWrap is called",
+      (booleanState, cssValue) => {
+        CypressMountWithProviders(
+          <ButtonMinor noWrap={booleanState}>
+            {" "}
+            Long long long long long text{" "}
+          </ButtonMinor>
+        );
+
+        if (booleanState === true) {
```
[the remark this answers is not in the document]
sianford: Is this correct? This is not testing that if `noWrap` is true it will have just one ccs value or not. It is testing that if `noWrap` is true then `white-space` will be a value or if `noWrap` is false then `flex-wrap` will be a value.

src/components/button-minor/button-minor-test.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  buttonType: "secondary",
+  fullWidth: false,
+  disabled: false,
+  destructive: false,
+  noWrap: false,
+  href: undefined,
+  iconPosition: "before",
+};
+
+export const DefaultStory = ({ ...args }) => (
```
mkrds, line 60:
```suggestion
export const DefaultStory = (args) => (
```

mkrds: The `args` argument should be typed correctly.

src/components/button-minor/button-minor-test.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  <ButtonMinor onClick={action("click")} {...args}>
+    Example Button
+  </ButtonMinor>
+);
+
+export const Default = ({ ...args }) => <ButtonMinor {...args} />;
```
mkrds, line 66:
```suggestion
export const Default = (args) => <ButtonMinor {...args} />;
```

The `args` argument should be typed correctly.

src/components/button-minor/button-minor-test.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const ButtonMinorDifferentTypes = ({ ...props }) => {
```
mkrds, line 68:
```suggestion
export const ButtonMinorDifferentTypes = (props) => {
```

The `props` argument should be typed correctly.

src/components/button-minor/button-minor.test.js
sianford: In this test, first, second and third have different expected values. Some are transparent, some are destructive around the outline or background depending on the button type. Looping around this would be complicated.

src/components/button-minor/button-minor-test.stories.tsx
sianford: When I change that I get this error "Parameter 'args' implicitly has an 'any' type." Is that correct?

src/components/button-minor/button-minor-test.stories.tsx
sianford: When I change that I get this error "Parameter 'args' implicitly has an 'any' type." Is that correct?

src/components/button-minor/button-minor-test.stories.tsx
sianford: When I change that I get this error "Parameter 'props' implicitly has an 'any' type." Is that correct?

src/components/button-minor/index.tsx
```diff
@@ -0,0 +1 @@
+export { default } from "./button-minor.component";
```
mkrds, line 1:
```suggestion
export { default } from "./button-minor.component";
export type { ButtonProps as ButtonMinorProps } from "../button"
```

src/components/button-minor/button-minor-test.stories.tsx
edleeks87: You'll need to pass the types to the args

`export const DefaultStory = (args: ButtonProps))`

edleeks87:
Nothing else from me that @mkrds and @DlgSHi haven't already raised

src/components/button-minor/button-minor-test.stories.tsx
DobroTora: Is this change acceptable @mkrds ?

`export const ButtonMinorDifferentTypes = ((props: JSX.IntrinsicAttributes & ButtonMinorProps) => {`

blaky:
Hi, do you guys expect this to be closed in the next week or so? I am just asking so whether we should wait a bit before upgrading carbon in our projects.

src/components/button-minor/button-minor-test.stories.tsx
mkrds: I've edited my comment and suggestion :)

mkrds (approved):

src/components/button-minor/button-minor.spec.tsx
```diff
@@ -0,0 +1,111 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
+import Icon from "components/icon";
```
edleeks87, line 3: Incorrect import path?

edleeks87 (commented):
I think the commits should be squashed as imo they are not atomic, if the first one is reverted the `test` commit will have tests that won't work etc

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,413 @@
...
+
+    beforeEach(() => {
+      callback = cy.stub();
+    });
+
+    it("should call onClick callback when a click event is triggered", () => {
+      CypressMountWithProviders(<ButtonMinorCustom onClick={callback} />);
+
+      buttonDataComponent()
+        .click({ force: true })
```
DlgSHi, line 249: why do we need the force click here ?

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,413 @@
...
+          expect(callback).to.have.been.calledOnce;
+        });
+    });
+
+    it("should call onBlur callback when a blur event is triggered", () => {
+      CypressMountWithProviders(<ButtonMinorCustom onBlur={callback} />);
+
+      buttonDataComponent()
+        .focus()
+        .blur({ force: true })
```
DlgSHi, line 261: the same question here?

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,413 @@
...
+          expect(callback).to.have.been.calledOnce;
+        });
+    });
+
+    it("should call onKeyDown callback when a keydown event is triggered", () => {
+      CypressMountWithProviders(<ButtonMinorCustom onKeyDown={callback} />);
+
+      buttonDataComponent()
+        .focus()
+        .realPress("ArrowRight")
```
DlgSHi, line 273: can we pass few keys to check instead of 1?

src/components/button-minor/button-minor.stories.mdx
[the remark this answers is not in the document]
sianford: Corrected.

src/components/button-minor/index.tsx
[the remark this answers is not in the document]
sianford: Corrected.

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
+import React from "react";
+
+import {
+  Default as ButtonMinor,
+  ButtonMinorDifferentTypes,
+} from "./button-minor-test.stories";
+
+import {
+  PrimaryButton,
```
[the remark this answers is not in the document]
sianford: Amended.

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+
+const destructive = "rgb(203, 55, 74)";
+const transparent = "rgba(0, 0, 0, 0)";
+
+context("Test for Button Minor component", () => {
+  describe("Check props for Button Minor component", () => {
+    it("should render Button Minor with aria-label prop", () => {
+      CypressMountWithProviders(
+        <ButtonMinor aria-label="cypress-aria">Foo</ButtonMinor>
```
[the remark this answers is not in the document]
sianford: Amended.

src/components/button-minor/button-minor.test.js
```diff
@@ -0,0 +1,461 @@
...
+
+      buttonDataComponent().should(
+        "have.attr",
+        "href",
+        "https://carbon.sage.com/"
+      );
+    });
+
+    it.each([
+      ["true", "white-space"],
```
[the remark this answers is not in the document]
sianford: Corrected.

src/components/button-minor/button-minor.test.js
sianford: Amended.

src/components/button-minor/button-minor.test.js
sianford: Corrected.

src/components/button-minor/button-minor-test.stories.tsx
sianford: Updated.

src/components/button-minor/button-minor-test.stories.tsx
sianford: Updated.

src/components/button-minor/button-minor-test.stories.tsx
sianford: Updated.

src/components/button-minor/index.tsx
sianford: Added.

src/components/button-minor/button-minor.spec.tsx
sianford: Corrected to `import Icon from "../icon";`. Other components using the same incorrect path will be corrected in another ticket.

src/components/button-minor/button-minor.test.js
sianford: Removed the force.

src/components/button-minor/button-minor.test.js
sianford: Removed the force.

src/components/button-minor/button-minor.test.js
sianford: Added ArrowLeft, ArrowUp, ArrowDown and Enter to test.

src/components/button-minor/button-minor.test.js
edleeks87: You were the chosen one Sianakin!

mkrds (approved):

PR #5562, opened 2022-10-19
refactor(form): convert to typescript

robinzigmond:
Convert Form component to typescript. Also extract stories to new `.stories.tsx` file and import into MDX docs.

### Proposed behaviour

Form should be a TypeScript component

### Current behaviour

Form is currently a Javascript component.

### Checklist

#### QA

### Testing instructions

As well as regression testing Form, please check the storybook docs, including all examples, and make sure that nothing (other than code reformatting) has changed in how the examples are presented.


src/components/form/form.stories.mdx
```diff
@@ -1,25 +1,8 @@
...
-import Switch from "../switch";
-import Button from "../button";
-import DialogFullScreen from "../dialog-full-screen";
-import Dialog from "../dialog";
-import Box from "../box";
-import CarbonProvider from "../carbon-provider/carbon-provider.component";
-
+import * as stories from "./form.stories";
 
 <Meta title="Form" parameters={{ info: { disable: true } }} />
```
Parsium, line 7:
```suggestion
<Meta title="Form" parameters={{ info: { disable: true }, controls: { disable: true } }} />
```
We will want to disable storybook controls for these stories since QAs will not be needing them here.

src/components/form/form.component.tsx
```diff
@@ -29,11 +57,11 @@ const Form = ({
...
   fullWidthButtons = false,
   ...rest
-}) => {
+}: FormProps) => {
   const { isInSidebar } = useContext(SidebarContext);
   const { isInModal } = useContext(ModalContext);
-  const formRef = useRef();
-  const formFooterRef = useRef();
+  const formRef = useRef() as React.MutableRefObject<HTMLFormElement>;
+  const formFooterRef = useRef() as React.MutableRefObject<HTMLDivElement>;
```
Parsium, line 64:
```suggestion
  const formRef = useRef<HTMLFormElement>(null);
  const formFooterRef = useRef<HTMLDivElement>(null);
```
Is there a reason why these refs need this mutable type? I think ideally we should avoid forcing type assertions where possible, since it could lead us to overlook potential issues.

robinzigmond: no, there's no good reason, I just needed a stricter type on these refs and this was the first thing I came up with, without thinking too much. Perhaps because I saw this in other components? In any case I agree with you that type assertions aren't ideal and much prefer your suggestion so I will change this :)

src/components/form/form.spec.tsx
```diff
@@ -183,7 +184,7 @@ describe("Form", () => {
 
     it("when rendered in a container, render footer with sticky styles", () => {
       const Component = () => {
-        const ref = useRef();
+        const ref = useRef() as React.MutableRefObject<HTMLDivElement>;
```
Parsium, line 187:
```suggestion
        const ref = useRef<HTMLDivElement>(null);
```

Parsium:
> > Since chromatic seems to flag any changes in story names, I think we should refactor the stories to typescript as part of a separate PR and merge that before this one - just in case we do introduce any name changes that could be mixed up with any potential regressions.
> > Please could you raise a new PR for this @robinzigmond? 😄
> 
> Thanks, yes I'll do this. (I don't need to raise a new PR though do I? I'll just split up the commits and force-push the change to this same branch.)

@robinzigmond Unfortunately because chromatic only runs for PRs and not commits, we will need to create a separate PR for the refactored stories, else any name changes will be flagged and mask any other regressions we may have introduced.

src/components/form/form.component.tsx
```diff
@@ -14,7 +13,36 @@ import {
   StyledFullWidthButtons,
 } from "./form.style";
 
-const Form = ({
+export interface FormProps extends SpaceProps {
+  /** Alignment of buttons */
+  buttonAlignment?: "left" | "right";
```
Parsium, line 18: What do you think about importing the `FORM_BUTTON_ALIGNMENTS` from `form.config.ts` to create a type here?

src/components/form/form.component.tsx
```diff
@@ -14,7 +13,36 @@ import {
...
+  /** Child elements */
+  children?: React.ReactNode;
+  /** The total number of errors present in the form */
+  errorCount?: number;
+  /** Spacing between form fields, given number will be multiplied by base spacing unit (8) */
+  fieldSpacing?: 0 | 1 | 2 | 3 | 4 | 5 | 7;
```
Parsium, line 24: Looking at the type defined for `fieldSpacing` in `StyledFormProps`, I think we might have made a typo in the original `form.d.ts` file by missing out the `6`. Could you add this in please since the `StyledForm` technically supports this?

src/components/form/form.style.ts
```diff
@@ -76,7 +85,13 @@ const formBottomMargins = (fieldSpacing) =>
 // Accounts for height of the header of Modal parent, the height form footer and some additional spacing
 const HEIGHT_SPACING = 216;
 
-export const StyledForm = styled.form`
+interface StyledFormProps extends StyledFormContentProps {
+  height?: string;
+  fieldSpacing: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7;
```
Parsium, line 90: Technically the `StyledForm` doesn't support the value `6` for `fieldSpacing` here. What do you think about adding support for this in? I would assume this would make the bottom margin `var(--spacing600)`?

src/components/form/__internal__/form-summary.component.tsx
```diff
@@ -1,15 +1,31 @@
...
+  children?: React.ReactNode;
+  /** The total number of errors present in the form */
+  errors?: number;
+  /** The total number of warnings present in the form */
+  warnings?: number;
+  /** Applies full width styling */
+  fullWidth?: boolean;
+}
+
+export const Summary = ({ type, errors = 0, warnings = 0 }: SummaryProps) => {
```
Parsium, line 28: What do you think about refactoring the `type` parameter to have type: `"error" | "warning"` instead of the plural forms? That way we can avoid using the type assertion below.

grabkowski, line 28: I would also change `errors` to `errorCount` and `warnings` to `warningCount`. What do you think about it?

src/components/form/form.spec.tsx
```diff
@@ -28,12 +28,13 @@ import StyledSearch from "../search/search.style";
...
 import StyledTextarea from "../textarea/textarea.style";
 import Dialog from "../dialog";
+import { SidebarContext } from "../sidebar/sidebar.component";
 
 jest.mock("lodash/debounce", () => jest.fn((fn) => fn));
 jest.mock("../../hooks/__internal__/useResizeObserver");
 
 describe("Form", () => {
-  let wrapper;
+  let wrapper: ReactWrapper | ShallowWrapper;
```
grabkowski, line 37: do we need to have two types here? We could just create another variable, e.g. `shallowWrapper` to avoid that.

robinzigmond:
> > > Since chromatic seems to flag any changes in story names, I think we should refactor the stories to typescript as part of a separate PR and merge that before this one - just in case we do introduce any name changes that could be mixed up with any potential regressions.
> > > Please could you raise a new PR for this @robinzigmond? 😄
> > 
> > 
> > Thanks, yes I'll do this. (I don't need to raise a new PR though do I? I'll just split up the commits and force-push the change to this same branch.)
> 
> @robinzigmond Unfortunately because chromatic only runs for PRs and not commits, we will need to create a separate PR for the refactored stories, so we can run chromatic separately on the refactored stories and the refactored component.

So we agreed this morning as a team to use a separate commit for moving stories from mdx to tsx, but to leave it in the same PR as the TS conversion of the component code. So I'm going to implement it this way for this PR.

src/components/form/form.style.ts
robinzigmond: Good spot! So it's actually the `Form` component itself that doesn't allow 6, and it looks like a mistake of mine here to include the `6` option in the `StyledFormProps` when there isn't a `formBottomMargin` case for it and it technically shouldn't ever be passed from the parent anyway.

However I don't see any downside to adding 6 as an option with `var(--spacing600)` as the margin value as you say so I will get this added 👍

src/components/form/form.spec.tsx
robinzigmond: hi Kuba, sorry I'm late coming to this but I don't think I understand this comment.

As in several other of our Enzyme-based test suites, `wrapper` can end up holding the rest of either a call to `mount` or to `shallow`. `mount` produces a `ReactWrapper` type while `shallow` produces a `ShallowWrapper`. So without refactoring the test suite to only use one of these, in order to allow both patterns to be used in `wrapper` we need to accept both types.

We have exactly the same in some other test suites we've converted to TS (pages and pill).

Please let me know if I've missed something and there's some easy way to avoid having both types as options for `wrapper`.

src/components/form/__internal__/form-summary.component.tsx
robinzigmond: yes I'm happy doing that as FormSummary is an internal component

src/components/form/form.spec.tsx
grabkowski: my bad, I encountered an issue during one of my refactors an issue where these types were incompatible, looks like in this case everything is ok

grabkowski (approved):

Parsium (approved):

divyajindel (approved):

PR #5571, opened 2022-10-20
refactor(switch): convert to typescript - FE-4836

mkrds:
### Proposed behaviour

This PR contains a Switch TypeScript conversion

### Checklist

#### QA

### Testing instructions


src/components/switch/switch.component.tsx
```diff
@@ -0,0 +1,159 @@
...
+import SwitchStyle from "./switch.style";
+import CheckableInput, {
+  CommonCheckableInputProps,
+} from "../../__internal__/checkable-input";
+import SwitchSlider from "./__internal__/switch-slider.component";
+import useIsAboveBreakpoint from "../../hooks/__internal__/useIsAboveBreakpoint";
+import { TooltipProvider } from "../../__internal__/tooltip-provider";
+import { filterStyledSystemMarginProps } from "../../style/utils";
+
+export type LabelAlign = "left" | "right";
```
robinzigmond, line 13: I'm not sure if there's a good reason to export the `LabelAlign` type or not (although I personally don't see one). But if there is, should it not also be exported from `index.ts`?

src/components/switch/switch.style.ts
```diff
@@ -9,9 +9,15 @@ import { StyledCheckableInput } from "../../__internal__/checkable-input/checkab
 import StyledSwitchSlider from "./__internal__/switch-slider.style";
 import StyledValidationIcon from "../../__internal__/validations/validation-icon.style";
 import { FieldLineStyle } from "../../__internal__/form-field/form-field.style";
+import { SwitchProps } from "./switch.component";
```
robinzigmond, line 12: same comment here as in `switch-slider` regarding importing props from the style file to the component file rather than the other way round

robinzigmond (approved):

Parsium (commented):
I noticed for `labelWidth` given though the type in the original `switch.d.ts` has not changed:
```ts
labelWidth?: number;
```
the prop's JS prop type used to allow for strings as well:
```js
labelWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
So even though this is small, this would count as a breaking change for our JavaScript consumers.

src/components/switch/switch.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+  return (
+    <Switch
+      label="Label"
+      name="switch-name"
+      checked={isChecked}
+      onChange={(e) => setIsChecked(e.target.checked)}
+      {...args}
+    />
+  );
+};
```
Parsium, line 17: Looking at this, in my opinion I think we should render the full source of the story rather than just the output to render.
I understand your reason against this since its expected our consumers would already be fully familiar with React state, but I feel including the state in the snippet is good for completeness, and would be consistent with how our other components, that are controlled by default, are documented.

src/components/switch/switch.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <Switch
+        info={args.validation}
+        label="Example switch (info)"
+        name="switch-info"
+        validationOnLabel={args.validationOnLabel}
+        tooltipPosition={args.tooltipPosition}
+      />
+    </>
+  );
+};
```
Parsium, line 148: Sort of related to this previous comment, I think the use of `Template.bind({})` is fine here, since we don't have any internal state

Parsium (approved):

Parsium:
> I noticed for `labelWidth` even though the type in the original `switch.d.ts` has not changed:
> 
> ```ts
> labelWidth?: number;
> ```
> 
> the prop's JS prop type used to allow for strings as well:
> 
> ```js
> labelWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
> ```
> 
> So even though this is small, this could count as a breaking change for our JavaScript consumers since we are now generating the prop types from our TypeScript interfaces.

Resolved offline as not as issue 👍🏼

robinzigmond (dismissed):

ZhuoyuJin (dismissed):

ZhuoyuJin (approved):

PR #5574, opened 2022-10-24
refactor(numeral-date): convert files to typescript (FE-4814)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/numeral-date/numeral-date.spec.tsx
```diff
@@ -276,47 +309,71 @@ describe("NumeralDate", () => {
...
-            "Year should be a number within a 1800-2200 range.\n"
-        );
-      });
+      describe.each<NumeralDateProps["dateFormat"][]>([
+        [["dd", "mm"]],
+        [["dd", "mm", "yyyy"]],
+      ])(
+        "when the format is set to %o and values are changed to be invalid",
+        (dateFormat) => {
+          it("multiline internal validation matching the format should be rendered", () => {
```
Parsium, line 339: What do you think about splitting this test in two - one for `["dd","mm"]` format and one for the `["dd", "mm", "yyyy"]` format? That way we won't need to add the conditional statements within the test.

Parsium (commented):
I noticed we are using type assertions quite a bit in `numeral-date.component.tsx`. Personally I think we should be avoiding using them where possible since it could mask potential future bugs. What do you think about this here?

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -83,7 +186,23 @@ const NumeralDate = ({
 
   const refs = useRef(dateFormat.map(() => React.createRef()));
 
-  const [internalMessages, setInternalMessages] = useState({});
+  const [internalMessages, setInternalMessages] = useState<NumeralDateObject>({
+    ...((Object.fromEntries(
+      dateFormat.map((datePart) => [datePart, ""])
+    ) as Partial<FullDate>) as NumeralDateObject),
```
Parsium, line 192: I've tried to have a look to see if there is a way we could avoid using the type assertion for `Object.fromEntries`. However, given the [in-built ES2019 type definition for `Object.fromEntries`](https://github.com/microsoft/TypeScript/blob/c242d4a46b1085718d5607591b013288c9c88f9c/src/lib/es2019.object.d.ts) is the following:

```ts
fromEntries<T = any>(entries: Iterable<readonly [PropertyKey, T]>): { [k: string]: T };
```

This means the type of the object returned by `fromEntries` can only be an **index signature** `[x: string]: string` if the assertion is not used, and not of type:

```ts
type NumeralDateObject = 
| {
   dd: string;
   mm: string;
   yyyy: string;
} | {
   dd: string;
   mm: string;
} | {
   mm: string;
   yyyy: string;
}
```
as we want.

The only way round this would be to declare our own type definition for `Object.fromEntries` in `src/global.d.ts`, similar to [this example](https://dev.to/svehla/typescript-object-fromentries-389c). I would say the benefit to Carbon doing this would be better type safety wherever we use `Object.fromEntries` so we can catch further type bugs, but I'll leave this up to you if you think making this change would be beneficial or just add unnecessary complexity.

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -103,38 +222,45 @@ const NumeralDate = ({
...
   const [dateValue, setDateValue] = useState({
-    ...(initialValue ||
-      dateFormat.reduce((dateObject, key) => {
-        dateObject[key] = "";
-        return dateObject;
-      }, {})),
+    ...((initialValue ||
+      (Object.fromEntries(
+        dateFormat.map((datePart) => [datePart, ""])
+      ) as Partial<FullDate>)) as NumeralDateObject),
```
Parsium, line 228: Similar to the above two type assertions comments

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -144,12 +270,14 @@ const NumeralDate = ({
...
 
-  const handleBlur = (datePart) => {
+  const handleBlur = (datePart: DatePart) => {
     const internalValidationEnabled =
       enableInternalError || enableInternalWarning;
     /* istanbul ignore else */
     if (validations[datePart] && internalValidationEnabled) {
-      const errorMessage = validations[datePart](dateValue[datePart])
+      const newDatePart: string =
+        dateValue[datePart as keyof NumeralDateObject];
```
Parsium, line 279: Same as the type assertion comment on `Object.fromEntries`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -170,8 +298,13 @@ const NumeralDate = ({
   };
 
   const internalMessage = Object.keys(internalMessages).reduce(
-    (string, key) =>
-      internalMessages[key] ? `${string + internalMessages[key]}\n` : string,
+    (combinedMessage, datePart) =>
+      internalMessages[datePart as keyof NumeralDateObject]
```
Parsium, line 302: Same as the type assertion comment on `Object.fromEntries`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -170,8 +298,13 @@ const NumeralDate = ({
...
+        ? `${
+            combinedMessage +
+            internalMessages[datePart as keyof NumeralDateObject]
```
Parsium, line 305: Same as the type assertion comment on `Object.fromEntries`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -244,7 +378,7 @@ const NumeralDate = ({
                       disabled={disabled}
                       readOnly={readOnly}
                       placeholder={datePart}
-                      value={dateValue[datePart]}
+                      value={dateValue[datePart as keyof NumeralDateObject]}
```
Parsium, line 381: Same as the type assertion comment on `Object.fromEntries`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -83,7 +186,23 @@ const NumeralDate = ({
 
   const refs = useRef(dateFormat.map(() => React.createRef()));
 
-  const [internalMessages, setInternalMessages] = useState({});
+  const [internalMessages, setInternalMessages] = useState<NumeralDateObject>({
+    ...((Object.fromEntries(
+      dateFormat.map((datePart) => [datePart, ""])
+    ) as Partial<FullDate>) as NumeralDateObject),
+  });
```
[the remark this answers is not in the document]
grabkowski: that cannot be done without type assertion, as you'll end up with a type error, you still have to tell the compiler that the resulting object will have the same props as NumeralDateObject

src/components/numeral-date/numeral-date.component.tsx
grabkowski: The assertion is to satisfy the compiler, as the NumeralDateObject has three types of property combinations.

The date part will always be a key of the NumeralDateObject here, as there is no possibility to have an input that has no corresponding property in that object (please look at invariant errors above).

src/components/numeral-date/numeral-date.component.tsx
grabkowski: date part will always be a key of the `NumeralDateObject`, this assertion is here to satisfy the compiler.

src/components/numeral-date/numeral-date.component.tsx
grabkowski: response in the comment above

src/components/numeral-date/numeral-date.spec.tsx
grabkowski: if I do that we'll end up with duplication of more than 30 lines of code instead of one conditional statement. 
Instead of that conditional I could move the expected message into a variable that is passed in the `.each` block, what do you think about that?

Parsium: Sounds good with me 👍🏼

src/components/numeral-date/numeral-date.component.tsx
Parsium: That's a good point given the existing invariant we have 👍🏼

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -17,23 +17,126 @@ import { TooltipProvider } from "../../__internal__/tooltip-provider";
...
+const incorrectDateFormatMessage =
+  "Forbidden prop dateFormat supplied to NumeralDate. " +
+  "Only one of these date formats is allowed: " +
+  "['dd', 'mm', 'yyyy'], " +
+  "['mm', 'dd', 'yyyy'], " +
+  "['dd', 'mm'], " +
+  "['mm', 'dd'], " +
+  "['mm', 'yyyy']";
+
+const isDayValid = (day: string) => (day ? +day > 0 && +day < 32 : true);
```
edleeks87, line 133: Not a problem for this PR but if we're baking in validations here I wonder why we aren't using `Date` here to check the days are valid for a given month and year 🤷‍♂️ 

```
const isDayValid = (year, month, day) => {
  const computedDay = new Date(`${+year}/${+month}/${+day}`).getDate();
  return computedDay === day;
}
```

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -17,23 +17,126 @@ import { TooltipProvider } from "../../__internal__/tooltip-provider";
...
+  helpAriaLabel?: string;
+}
+
+const ALLOWED_DATE_FORMATS: ValidDateFormat[] = [
   ["dd", "mm", "yyyy"],
   ["mm", "dd", "yyyy"],
   ["dd", "mm"],
   ["mm", "dd"],
   ["mm", "yyyy"],
 ];
```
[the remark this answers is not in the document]
grabkowski: that would not work here, as the resulting type will be an array of strings, and we need specific values in those arrays

src/components/numeral-date/numeral-date.component.tsx
edleeks87: Yeah was just a comment rather than something I expected you to change here, I'm not convinced it's a breaking change as to me that is a bug in our internal validations but again not a problem for this PR

src/components/numeral-date/numeral-date.component.tsx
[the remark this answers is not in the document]
grabkowski: nevermind, I missed the `as const` in the `ALLOWED_DATE_FORMATS`

Parsium (approved):

edleeks87 (approved):

edleeks87 (approved):

Parsium (approved):

ZhuoyuJin (approved):

src/components/numeral-date/numeral-date.stories.mdx
```diff
@@ -30,164 +31,22 @@ import NumeralDate from "carbon-react/lib/components/numeral-date";
...
-      return (
-        <NumeralDate
-          value={value}
-          onChange={(e) => setValue(e.target.value)}
-          label="Default"
-        />
-      );
-    }}
-  </Story>
+  <Story name="default" story={stories.Default} />
```
Parsium, line 37: I'm happy to approve this change since I agree it makes sense, but wanted to highlight that chromatic will treat this as a new story due to the change in story name.

Parsium (dismissed):

src/components/numeral-date/numeral-date.stories.mdx
grabkowski: I think that we have agreed that we'll be changing the stories during TS refactor

src/components/numeral-date/numeral-date.stories.tsx
```diff
@@ -0,0 +1,143 @@
+import React, { useState } from "react";
+import { ComponentStory } from "@storybook/react";
+import NumeralDate from ".";
+
+export const Default: ComponentStory<typeof NumeralDate> = () => {
```
edleeks87, line 5: If we're going to rename this to `Default` it should not be the controlled implementation, it should be the minimum props required for the component to work. When we remove the support for uncontrolled this will become the default

grabkowski: ok, story name changed

edleeks87 (approved):

edleeks87 (dismissed):

src/components/numeral-date/numeral-date.stories.mdx
```diff
@@ -30,164 +31,22 @@ import NumeralDate from "carbon-react/lib/components/numeral-date";
...
 </Canvas>
 
 ### Uncontrolled
 
 <Canvas>
```
Parsium, line 42:
```suggestion
### Controlled

<Canvas>
```

src/components/numeral-date/numeral-date.stories.mdx
```diff
@@ -30,164 +31,22 @@ import NumeralDate from "carbon-react/lib/components/numeral-date";
...
-          onChange={(e) => setValue(e.target.value)}
-          label="Default"
-        />
-      );
-    }}
-  </Story>
+  <Story name="default" story={stories.Default} />
 </Canvas>
 
 ### Uncontrolled
```
Parsium, line 40:
```suggestion
### Controlled
```

Parsium (dismissed):

edleeks87 (dismissed):

ZhuoyuJin (approved):

Parsium (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #5577, opened 2022-10-25
fix(flat-table): ensure th elements within sticky thead overlap tbody elements when scrolling FE-5095

### LINKED ISSUE #5459: The content is overlapped by the header
### Current behaviour

The report header is overlapped by the drill down.

### Expected behaviour

The report header must remain in front.

https://codesandbox.io/s/flat-table-header-objfyh

### JIRA Ticket (Sage Only)

SBS-36717

### Suggested Solution

_No response_

### Carbon Version

110.3.0

### Design Tokens Version

_No response_

Firefox, Chrome, Safari

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #5576: Multiple Sticky table header rows are not sticky
### Current behaviour

Given a FlatTable with sticky headers
When the headers are updated to nested headers
Then the header rows below the first are not sticky

### Expected behaviour

The header rows below the first remain sticky when updated.

https://codesandbox.io/s/table-headers-oh5trr?file=/src/Table.tsx

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

111.8.0

### Design Tokens Version

2.48.0

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5459
fix #5576

### Proposed behaviour

Increments the `z-index` of `th` elements within `thead` when `hasStickyHead` prop is set. This is
because when `tr` elements are interactive or have `subRows` they set the `z-index` of the child
`td` elements to the same `z-index` and are defined later in the DOM so overlap.

Adds additional support for rendering both left and right hand side sticky columns using the
`FlatTableRowHeader` component. A new `stickyAlignment` prop has been added which defaults to
`left`. When `right` is passed it will lock the given column and any columns following it whilst the
remaining columns can be scrolled.

image

image

expandable sticky cols

sticky cols

### Current behaviour

When `hasStickyHead` prop is true and rows are interactive and focused the header is overlapped when scrolling the table body.

No support for rendering right hand side sticky columns

### Checklist

#### QA

### Additional context

I have not added a cypress test as the work to refactor the tests for `FlatTable` is currently in flight, I will add one once that has been merged

### Testing instructions

Linked sandbox for issue, storybook stories. I will be happy to build sandboxes if QA request certain examples


src/components/flat-table/flat-table.stories.mdx
```diff
@@ -198,6 +199,115 @@ Any cells preceding the `FlatTableRowHeader`s will automatically also be made st
...
+        </FlatTableRow>
+      </FlatTableHead>
+      <FlatTableBody>
+        <FlatTableRow>
+          <FlatTableCell>Sticky Content</FlatTableCell>
+          <FlatTableRowHeader>Sticky Content</FlatTableRowHeader>
+          <FlatTableCell>Scrollable Content</FlatTableCell>
+          <FlatTableCell>Scrollable Content</FlatTableCell>
+          <FlatTableCell>Scrollable Content</FlatTableCell>
+          <FlatTableRowHeader stickyAlignment="right">Sticky Content</FlatTableRowHeader>
```
Parsium, line 231: Do we want to allow consumers to make certain rows in a column sticky as opposed to the whole column? If not, what do you think about only passing `stickyAlignment="right"` in to `FlatTableRowHeader` in the head of the table so we don't have to repeat using the `stickyAlignment` prop for every row in the column?

src/components/flat-table/flat-table.stories.mdx
```diff
@@ -198,6 +199,115 @@ Any cells preceding the `FlatTableRowHeader`s will automatically also be made st
   </Story>
 </Canvas>
 
+### Using row headers to create left and right hand side sticky columns
+
+In order to have right hand side sticky columns you should set the `FlatTableRowHeader` `stickyAlignment` prop to `right`.
+Any columns to following it will also be sticky.
+
+By default the left hand side columns have the prop set to `left`.
+It is important to make sure that the left hand side columns are not rendered before the right hand side ones.
```
Parsium, line 208: I can't tell if there is some ambiguity here, are you referring to the case where columns that should precede the left-aligned sticky column being rendered after the sticky column by mistake, or that for cases with complex tables the render times for the columns should be accounted for so the styles are applied correctly?

src/components/flat-table/flat-table.stories.mdx
edleeks87: you would need to maintain that index and share it between the other rows, it becomes extremely messy to achieve. This component will be rewritten once the pending RFC has been accepted and this interface will change as a result to something like what you suggest where a single sticky index is set but that isn't easily achievable with current implementation

src/components/flat-table/flat-table.stories.mdx
edleeks87: It means do not render cells that you want to be sticky on the lhs of the table after the cells you want to be sticky on the rhs, I'll try and update this but I was struggling find a succinct way of describing it tbh

src/components/flat-table/flat-table-row/flat-table-row.spec.js
```diff
@@ -675,39 +779,12 @@ describe("FlatTableRow", () => {
 
       assertStyleMatch(
         {
-          borderLeft: "2px solid var(--colorsUtilityMajor100)",
+          borderRight: "2px solid var(--colorsUtilityMajor100)",
         },
         wrapper,
-        { modifier: `:focus ${StyledFlatTableRowHeader}` }
+        { modifier: `${StyledFlatTableRowHeader}:nth-child(4)` }
```
robinzigmond, line 785: I'm sure there's nothing wrong with this but I'm a bit puzzled why a test for "applies the correct focus styling" has had the `:focus` modifier removed in the assertion. Or why we need to specify which child it is when there's only one `FlatTableRowHeader` child in this row.

src/components/flat-table/flat-table-row/flat-table-row.spec.js
```diff
@@ -675,39 +779,12 @@ describe("FlatTableRow", () => {
...
       );
     });
-
-    it.each(["dark", "light", "transparent-white", "transparent-base"])(
```
robinzigmond, line 685: I'm a bit puzzled why these have been removed as the changes in this PR don't remove any of this styling do they?

src/components/flat-table/flat-table.stories.mdx
robinzigmond, line 208:
```suggestion
It is important to make sure that the left hand side columns are rendered before the right hand side ones.
```

Error while editing I assume? As the requirement is that the left columns *do* come before the right ones.

src/components/flat-table/flat-table-head/flat-table-head.style.js
```diff
@@ -4,11 +4,18 @@ import { StyledFlatTableRowHeader } from "../flat-table-row-header/flat-table-ro
...
+      border-left: none;
+      border-right: none;
+    }
+    ${StyledFlatTableRowHeader}, ${StyledFlatTableCheckbox} {
+      font-weight: 700;
+    }
+  }
+
+  ${StyledFlatTableRowHeader} {
+    z-index: ${({ theme }) => theme.zIndex.overlay + 5};
```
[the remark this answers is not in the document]
edleeks87: Yeah I was going to do something like this but ended up not as I'm hoping we can removed some of these when we rewrite. I'm happy to add it in as you make a good point and who knows when it will be rewritten 👍

src/components/flat-table/flat-table-head/flat-table-head.style.js
```diff
@@ -4,11 +4,18 @@ import { StyledFlatTableRowHeader } from "../flat-table-row-header/flat-table-ro
 import StyledFlatTableCheckbox from "../flat-table-checkbox/flat-table-checkbox.style";
```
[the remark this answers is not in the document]
edleeks87: Yes these were due to a steer form @harpalsingh relating to some styling that was incorrect

src/components/flat-table/flat-table-row/flat-table-row.spec.js
edleeks87: Good point I'll update the test description

src/components/flat-table/flat-table.stories.mdx
edleeks87: good spot thanks 👍

src/components/flat-table/flat-table-header/flat-table-header.d.ts
```diff
@@ -15,6 +15,8 @@ export interface FlatTableHeaderProps extends PaddingProps {
   rowspan?: number | string;
   /** Sets a custom vertical right border */
   verticalBorder?: TableBorderSize;
+  /** Sets the color of the right border */
+  verticalBorderColor?: string;
```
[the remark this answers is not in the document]
edleeks87: Fair point, I had originally had an issue with making the commits atomic but I ended up removing the code that relied on this as I found a "better" way but forgot to separate them. I'm not sure if it should be a `feat` or `fix` tbh as Harpal thought it should already exist, it's irrelevant in the scheme of things though as the other commit is a `feat` etc

src/components/flat-table/flat-table-row-header/flat-table-row-header.spec.js
```diff
@@ -202,6 +202,100 @@ describe("FlatTableRowHeader", () => {
...
+          },
+          wrapper,
+          { modifier: "&&&&" }
+        );
+      });
+    }
+  );
+
+  describe.each([["red"], ["#ffffff"], ["--colorsUtilityMajor550"]])(
+    "when the verticalBorderColor prop is set to %s and stickyAlignment='right'",
```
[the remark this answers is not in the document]
edleeks87: One asserts right border colour the other right, I can probably get it working via an `each` block though

src/components/flat-table/flat-table-row/flat-table-row.component.js
```diff
@@ -47,37 +48,74 @@ const FlatTableRow = React.forwardRef(
...
       },
-      [stickyCellWidths]
+      [lhsRowHeaderIndex, leftStickyCellWidths, rightStickyCellWidths]
     );
 
     let interactiveRowProps = {};
 
-    const firstCellIndex = () =>
-      childrenArray[0].type === FlatTableCheckbox ? 1 : 0;
+    const firstCellIndex = useMemo(
```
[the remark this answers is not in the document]
edleeks87: My thinking here is it's worth memoising wherever possible given that some consumers will have large tables that can have bulk selections/ updates, under these scenarios this index will be recalculated unnecessarily in many cases. However, your'e correct it is an inexpensive calculation and the fact we've not seen any impact with the previous approach I'm inclined to agree the cost of `useMemo` is probably more or at least equal so I'll remove

src/components/flat-table/flat-table-row/flat-table-row.style.js
```diff
@@ -119,6 +145,47 @@ const StyledFlatTableRow = styled.tr`
...
+      css`
+        ${StyledFlatTableRowHeader}:nth-child(${rhsRowHeaderIndex + 1}) {
+          border-left: 2px solid ${verticalBorderColor(colorTheme)};
+        }
+
+        ${StyledFlatTableHeader}:nth-child(${rhsRowHeaderIndex}) {
+          border-right: none;
+        }
+      `}
+
```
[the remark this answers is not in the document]
edleeks87: yeah good point, I think I was using `nth-of-type` originally

src/components/flat-table/flat-table-row/flat-table-row.spec.js
edleeks87: The styles have changed so the assertion was no longer relevant, there's little value in adding a test back in as these styles are already caught by visual regression and there's no coverage missing either

src/components/flat-table/flat-table-head/flat-table-head.style.js
```diff
@@ -3,12 +3,21 @@ import baseTheme from "../../../style/themes/base";
 import { StyledFlatTableRowHeader } from "../flat-table-row-header/flat-table-row-header.style";
 import StyledFlatTableCheckbox from "../flat-table-checkbox/flat-table-checkbox.style";
 
+const ROW_HEADER_OVERLAY_Z_INDEX = baseTheme.zIndex.overlay + 5;
```
edleeks87, line 6: These constants will be added into the fix commit when I squash the commits

src/components/flat-table/flat-table-row/flat-table-row.spec.js
```diff
@@ -106,29 +106,43 @@ describe("FlatTableRow", () => {
...
         wrapper,
-        { modifier: `:focus ${StyledFlatTableRowHeader}` }
+        {
+          modifier: `:focus ${StyledFlatTableRowHeader}:first-of-type:not(:last-of-type)::before`,
+        }
+      );
+
+      assertStyleMatch(
+        {
+          borderRight: "2px solid var(--colorsSemanticFocus500)",
```
Parsium, line 129: I noticed in this file we are verifying the golden focus border as 2px wide instead of 3px which is used elsewhere in Carbon. I appreciate one pixel difference is incredibly minor - so if this change is not complex, could you update the styling for `StyledFlatTableRowHeader` so its consistent with Design System?

Parsium (approved):

robinzigmond (dismissed):

robinzigmond (dismissed):

robinzigmond (approved):

Parsium (approved):

robinzigmond (approved):

Parsium (approved):

ZhuoyuJin (approved):

PR #5578, opened 2022-10-26
test(textbox): refactor for cypress tests

DlgSHi:
### Proposed behaviour

Refactor Cypress tests for the `Textbox` component to use new cypress-component-react framework for testing.
Add `textbox.test.js`. 

- removed few unnecessary `json` files

### Current behaviour

### Checklist

#### QA

### Testing instructions

- [ ] Run `npx cypress open --component` to check if there is newly added test.file
- [ ] Check if the `textbox.test.js` file passed
- [ ] Check all tests were moved/refactored from .feature file -> .test.js file
- [ ] Check all unnecessary files were deleted, including .json file
- [ ] Run `npx cypress run --component` to check none of the other *.test.js files have regressed
- [ ] Run `npx cypress run --e2e` to check none of the feature files have regressed


src/components/textbox/textbox.test.js
```diff
@@ -0,0 +1,712 @@
...
+        .then(($el) => {
+          checkGoldenOutline($el.parent());
+        });
+    });
+
+    it("should render Textbox with readOnly prop", () => {
+      CypressMountWithProviders(<TextboxComponent readOnly />);
+
+      textboxInput().and("have.attr", "readOnly");
+      textbox().should("have.attr", "readOnly");
```
robinzigmond, line 295: I would remove this line. The important thing is the `readonly` attribute on the input - the fact that this also appears on the containing `div` is just an accident and doesn't actually do anything.

src/components/textbox/textbox.test.js
```diff
@@ -0,0 +1,712 @@
...
+          .and("have.css", "font-size", "16px")
+          .and("have.css", "font-weight", "900")
+          .and("have.css", "margin-right", "8px");
+      }
+    );
+
+    it.each([
+      [true, "input", "label"],
+      [false, "label", "input"],
+    ])(
```
robinzigmond, line 389: unfortunately, these tests don't work as intended - if you switch the order of "input" and "label" in one of them, which should make it fail, it still passes. I've not yet been able to figure out how to fix it - happy to leave it to you as you know Cypress much better than me! - but these needs fixing otherwise this test has no value.

src/components/textbox/textbox.test.js
```diff
@@ -0,0 +1,712 @@
...
+
+    it("should render Textbox with formattedValue prop", () => {
+      CypressMountWithProviders(
+        <TextboxComponent formattedValue={CHARACTERS.STANDARD} />
+      );
+
+      textboxInput().should("have.attr", "value", CHARACTERS.STANDARD);
+    });
+
+    it.each(["add", "filter", "play"])(
```
ZhuoyuJin, line 484: Can we list a few more elements here like "arrow_left_boxed" , "bullet_list_dotted", etc...?

src/components/textbox/textbox.test.js
```diff
@@ -0,0 +1,712 @@
...
+            // eslint-disable-next-line no-unused-expressions
+            expect(callback).to.have.been.calledOnce;
+          });
+      }
+    );
+
+    it.each([
+      [1000, "1"],
+      [5000, "5"],
+      [10000, "10"],
```
ZhuoyuJin, line 677: In this test case, do we need to include a bit longer time, for example [50000, "50"]?

stephenogorman (commented):
I'm not sure if this scenario from textbox.feature is covered:

  Scenario Outline: Verify input of Textbox component
    Given I open "Textbox Test" component page "default"
    When I type <input> into Textbox
    Then Textbox input on preview is set to <input>

src/components/textbox/textbox.test.js
DlgSHi: what it is the point of that ?

src/components/textbox/textbox.test.js
DlgSHi: why? how 10s is different from 50s?

src/components/textbox/textbox.test.js
DlgSHi: thank you for this comment, fixed this as well

stephenogorman (approved):

robinzigmond (approved):

ZhuoyuJin (approved):

PR #5587, opened 2022-10-31
chore: add lint rule to prevent prop-type imports in typescript files

edleeks87:
### Proposed behaviour

Adds eslint plugin `no-restricted-imports` and configures to prevent `prop-types` being imported in `ts/x` files

### Current behaviour

No linting exists to prevent `prop-types` from being imported in `ts/x` files

### Checklist

#### QA

### Testing instructions

Add `import PropTypes from "prop-types"` to a `ts/x` file and run the linter

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -151,7 +151,11 @@
         "no-use-before-define": "off",
         "@typescript-eslint/no-use-before-define": ["error"],
         "@typescript-eslint/no-empty-function": ["error", {"allow": ["arrowFunctions"]}],
-        "jest-dom/prefer-to-have-attribute": "off"
+        "jest-dom/prefer-to-have-attribute": "off",
+        "no-restricted-imports": ["error", {
+          "name": "prop-types",
+          "message": "Please do not import `prop-types` in TypeScript files."
+      }]
       }
     }
   ]


DlgSHi (dismissed):

.eslintrc
```diff
@@ -151,7 +151,11 @@
         "no-use-before-define": "off",
         "@typescript-eslint/no-use-before-define": ["error"],
         "@typescript-eslint/no-empty-function": ["error", {"allow": ["arrowFunctions"]}],
-        "jest-dom/prefer-to-have-attribute": "off"
+        "jest-dom/prefer-to-have-attribute": "off",
+        "no-restricted-imports": ["error", {
+          "name": "prop-types",
+          "message": "Please do not import `prop-types` in this file."
```
mkrds, line 157:
```suggestion
          "message": "Please do not import `prop-types` in TypeScript files."
```

.eslintrc
```diff
@@ -127,7 +127,7 @@
         },
         "project": ["./tsconfig.json"]
       },
-      "plugins": ["@typescript-eslint", "jest"],
+      "plugins": ["@typescript-eslint", "jest", "no-restricted-imports"],
```
mkrds, line 130: I think there's no need for the `eslint-plugin-no-restricted-imports` considering that `no-restricted-imports` is a rule built in eslint
https://eslint.org/docs/latest/rules/no-restricted-imports#version

I've tested it and it seems to work without adding that plugin

edleeks87: Yeah I wrongly thought it was only added in v8, I think I had the syntax incorrect when I tested if it was built in as it does work, I'll  update it soon 🤦

mkrds (approved):

DlgSHi (approved):

PR #5599, opened 2022-11-07
feat: improve intellisense and autocompletion for props using typecript

robinzigmond:
### Proposed behaviour

When using Carbon components in a TypeScript project, the auto-completion and intellisense on props should be helpful. In particular, when hovering over a prop name:
- if the prop is a union of strings or numbers, the options should be displayed, not just a type alias which is meaningful to the consumer
- if the prop is a relatively complex object or function, the type shown on hover should describe what properties it has and their types, rather than be something somewhat cryptic.

Note: this PR does not give a "ideal" representation of all props - firstly because what is best is often subjective, and secondly because particularly with some styled-system props it wasn't possible to do this completely to my satisfaction. But all props should at least display better than before, and any cases we are unhappy with can be left for future investigation.

### Current behaviour

Many types shown of props on hover do not live up to the above description and are confusing or even meaningless to consumers - see the linked codesandbox for examples.

### Checklist

#### QA

### Testing instructions

Note that the following Codesandbox example does not even run, for reasons I do not fully understand but which I'm sure are unrelated to the TS issues fixed by this PR. Even if it did run, the output in the browser would be nonsensical, and it wouldn't matter anyway. The point of the sandbox instead is to highlight what consumers of Carbon see when using our components - in particular, what type is displayed when hovering over a prop. Only those props which have a problematic display are used (as well as some that are required by the component - these are highlighted in the comments). Each prop is commented with a brief description of what is wrong with it.

To test this PR, the best way - although unfortunately very time-consuming - is to compare every prop given in the sandbox both before and after the PR, and confirm that the version after the PR is more helpful to a developer consuming Carbon.

https://codesandbox.io/s/bad-type-display-jvkjr2?file=/src/index.tsx:29249-29257

### DIFF
--- a/src/__internal__/utils/helpers/types/expand.ts
+++ b/src/__internal__/utils/helpers/types/expand.ts
@@ -1,0 +1,23 @@
+// taken from answers to https://stackoverflow.com/questions/57683303/how-can-i-see-the-full-expanded-contract-of-a-typescript-type
+
+export type ExpandOnce<T> = T extends (...args: infer A) => infer R
+  ? (...args: ExpandOnce<A>) => ExpandOnce<R>
+  : T extends infer O
+  ? { [K in keyof O]: O[K] }
+  : never;
+
+// note: does not appear to fully recursively expand all styled-system types. But the alternatives
+// have all had worse problems.
+export type Expand<T> = T extends (...args: infer A) => infer R
+  ? (...args: Expand<A>) => Expand<R>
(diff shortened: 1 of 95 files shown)


grabkowski:
I wonder why do we need this, as the type could be easily checked by just ctrl+click on Windows and cmd+click on Mac on the prop and going to the definition. I also cannot see problems with Intellisense as it is already suggesting and autofilling these properties:
image

In my opinion this will only add more code to our project with a little gain of showing the types in a preview instead of checking them just by one click more.

PR #5612, opened 2022-11-15
refactor(step-sequence): move demo stories to tsx file FE-5501 

edleeks87:
### Proposed behaviour

Refactor stories, moved to tsx file
Removes the hardcoded padding styling and surfaces the padding interface with a default value of 0
Ensures height flexes to container when in `vertical` `orientation`
Update styling for `incomplete` steps to include dashed line
### Current behaviour

Hardcoded padding styling
Height does not flex to container when in `vertical` `orientation`
Styling for `incomplete` steps to has solid line

### Checklist

#### QA

### Additional context

image

image

### Testing instructions

https://codesandbox.io/s/sad-field-nu72jj?file=/src/index.js

### DIFF
--- a/src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
+++ b/src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
@@ -10,9 +10,15 @@ import {
   StyledStepSequenceItemHiddenLabel,
   StyledStepSequenceItemIndicator,
 } from "./step-sequence-item.style";
+import { StepSequenceProps } from "../step-sequence.component";
+
+const orientations = ["horizontal", "vertical"] as const;
 
 describe("StepSequenceItem", () => {
-  const render = (props: Partial<StepSequenceItemProps>, renderer = mount) =>
+  const render = (
+    props: Partial<StepSequenceItemProps & StepSequenceProps>,
+    renderer = mount
+  ) =>
     renderer(
       <StepSequenceItem indicator="1" {...props}>
         Item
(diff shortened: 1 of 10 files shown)


robinzigmond (dismissed):

robinzigmond (dismissed):

src/components/step-sequence/step-sequence.component.tsx
```diff
@@ -20,15 +20,16 @@ export const StepSequence = ({
   ...props
 }: StepSequenceProps) => {
   return (
-    <StepSequenceStyle
+    <StyledStepSequence
       data-component="step-sequence"
       orientation={orientation}
+      p={0}
```
Parsium, line 26: I am assuming we want to allow consumers to control padding now as well as margins - so should we update `StepSequenceProps` to extend `SpaceProps` instead of `MarginProps`?

robinzigmond (dismissed):

src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
```diff
@@ -26,7 +32,61 @@ describe("StepSequenceItem", () => {
...
-  describe("when complete", () => {
+  describe("when status is incomplete", () => {
+    it.each(orientations)(
+      "renders the correct styling when orientation is %s",
+      (orientation) => {
+        const wrapper = render({
+          ...defaultProps,
+          status: "incomplete",
+          orientation,
+        });
```
Parsium, line 43:
```suggestion
      const wrapper = mount(
          <StepSequence orientation={orientation}>
              <StepSequenceItem status="incomplete" indicator="1">Name</StepSequenceItem>
          </StepSequence>
      );
```
I'm not sure if this test would work this way since `StepSequenceItem` does not accept an `orientation` prop and instead relies on context passed down from the parent `StepSequence`. Would it be worth doing the suggested, mocking the context or moving this test to `step-sequence.spec.tsx`?

src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
```diff
@@ -26,7 +32,61 @@ describe("StepSequenceItem", () => {
...
+
+  describe("when status is complete", () => {
+    it.each(orientations)(
+      "renders the correct styling when orientation is %s",
+      (orientation) => {
+        const wrapper = render({
+          ...defaultProps,
+          status: "complete",
+          orientation,
+        });
```
Parsium, line 66: Likewise with the previous comment

src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
```diff
@@ -40,38 +100,55 @@ describe("StepSequenceItem", () => {
...
-          color: "var(--colorsUtilityYin090)",
-        },
-        wrapper
-      );
-    });
+  describe("when status is current", () => {
+    it.each(orientations)(
+      "renders the correct styling when orientation is %s",
+      (orientation) => {
+        const wrapper = render({
```
Parsium, line 116: Likewise with previous comments

src/components/step-sequence/step-sequence.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+      indicator="5"
+      status="incomplete"
+    >
+      Confirm
+    </StepSequenceItem>
+  </StepSequence>
+);
+
+export const Vertical: ComponentStory<typeof StepSequence> = () => (
+  <Box height="600px">
```
Parsium, line 59: What do you think of moving the `Box` to a decorator for this story so it doesn't shown in the code snippet? Happy to accept otherwise as this is minor

edleeks87: I don't think it's that big of an issue they will need to render it in a Box or container and the point is to show that is flexes to the height of the container etc

src/components/step-sequence/step-sequence-item/step-sequence-item.spec.tsx
edleeks87: This test works as the `StepSequenceItem` component spreads `rest`; we actually only care about rendering the underlying `StyledStepSequenceItem` here but I decided to use existing render function to achieve it rather than writing a long form mount just for this. There's already snapshot tests in `step-sequence.spec.tsx` that capture the behaviour of the whole component in the different `orientation`s so this one is more focused on the styling of the underlying sc.

Parsium (dismissed):

src/components/step-sequence/step-sequence.test.js
```diff
@@ -80,19 +80,16 @@ const StepSequenceItemCustom = ({ ...props }) => {
...
-      (orientation, padding) => {
+      (orientation) => {
         CypressMountWithProviders(
           <StepSequenceComponent orientation={orientation} />
         );
 
         stepSequenceDataComponent()
           .should("have.attr", "orientation", orientation)
-          .and("have.css", "padding-top", `${padding}px`);
+          .and("have.css", "padding-top", "0px");
```
divyajindel, line 92:
```suggestion

```

robinzigmond (approved):

PR #5623, opened 2022-11-21
refactor(textarea): convert textarea to typescript

nineteen88:
### Proposed behaviour

This PR contains the Textarea to TypeScript conversion

### Checklist

#### QA


src/components/textarea/textarea.component.tsx
```diff
@@ -109,7 +198,9 @@ const Textarea = ({
 
   useEffect(() => {
     if (rows) {
-      minHeight.current = inputRef.current.scrollHeight;
+      /* istanbul ignore next */
+      minHeight.current =
+        inputRef && inputRef.current ? inputRef.current.scrollHeight : 0;
```
grabkowski, line 203:
```suggestion
       inputRef?.current?.scrollHeight || 0;
```

src/components/textarea/textarea.component.tsx
```diff
@@ -122,7 +213,9 @@ const Textarea = ({
   useEffect(() => {
     if (expandable) {
       window.addEventListener("resize", expandTextarea);
-      minHeight.current = inputRef.current.clientHeight;
+      /* istanbul ignore next */
+      minHeight.current =
+        inputRef && inputRef.current ? inputRef.current.clientHeight : 0;
```
grabkowski, line 218:
```suggestion
       inputRef?.current?. clientHeight || 0
```

src/components/textarea/textarea.spec.tsx
```diff
@@ -23,18 +23,24 @@ import StyledTextarea from "./textarea.style";
...
   describe("when textarea is rendered with default props", () => {
-    let textarea;
+    let textarea: HTMLInputElement;
 
     beforeAll(() => {
       wrapper = renderTextarea();
-      textarea = wrapper.find("textarea").instance();
+      textarea = (wrapper
+        .find("textarea")
+        .instance() as unknown) as HTMLInputElement;
```
grabkowski, line 43: we should not test on instances, I think that it is a good moment to rewrite this test with that refactor.

src/components/textarea/textarea.spec.tsx
```diff
@@ -260,12 +268,12 @@ describe("Textarea", () => {
...
 
   describe('when the "expandable" prop is set to "true"', () => {
-    let textarea, textareaInstance;
+    let textarea, textareaInstance: HTMLInputElement;
 
     beforeEach(() => {
       wrapper = renderTextarea({ expandable: true });
       textarea = wrapper.find("textarea");
-      textareaInstance = textarea.instance();
+      textareaInstance = (textarea.instance() as unknown) as HTMLInputElement;
```
grabkowski, line 276: another instance test

src/components/textarea/textarea.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+import Textarea from ".";
+
+export const DefaultStory: ComponentStory<typeof Textarea> = () => {
+  const [state, setState] = useState("");
+  const setValue = ({ target }: React.ChangeEvent<HTMLInputElement>) => {
+    setState(target.value);
+  };
+  return <Textarea label="Textarea" value={state} onChange={setValue} />;
+};
+DefaultStory.storyName = "default";
```
grabkowski, line 15: this story is already named in `mdx`, so we don't need that to be declared for second time here.

src/components/textarea/textarea.component.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { useRef, useEffect, useContext } from "react";
-import PropTypes from "prop-types";
-import styledSystemPropTypes from "@styled-system/prop-types";
+import { MarginProps } from "styled-system";
 
-import { InputPresentation } from "../../__internal__/input";
+import { IconType } from "components/icon";
```
robinzigmond, line 4:
```suggestion
import { IconType } from "../icon";

Is there a reason we're not using the relative import style for this? I've seen imports break mysteriously in the past when doing it the absolute way, and in any case I prefer my suggestion for consistency if nothing else!
```

src/components/textarea/textarea.component.tsx
```diff
@@ -59,20 +147,21 @@ const Textarea = ({
   "data-role": dataRole,
   helpAriaLabel,
   ...props
-}) => {
+}: TextareaProps) => {
   const { validationRedesignOptIn } = useContext(NewValidationContext);
-  const computeLabelPropValues = (prop) =>
+  const computeLabelPropValues = <T,>(prop: T): undefined | T =>
```
robinzigmond, line 152:
```suggestion
  const computeLabelPropValues = <T>(prop: T): undefined | T =>
```

I haven't tried it but is there any reason we need the comma there? It looks weird to me.

src/components/textarea/textarea.component.tsx
robinzigmond: if you make the change suggested by Kuba I think you might be able to remove the `istanbul ignore next`. If not, it should go just before the `0` rather than ignoring the whole line.

src/components/textarea/textarea.component.tsx
robinzigmond: same comment as above

src/components/textarea/textarea.spec.tsx
robinzigmond: agreed, I didn't even know what this meant, and having looked it up I'm still confused as to how it works with functional components, or (as here) plain HTML elements!

It looks like we want to get the DOM element, so `.getDOMNode()` instead of `.instance()` ought to work I would think.

src/components/textarea/textarea.spec.tsx
```diff
@@ -249,9 +255,11 @@ describe("Textarea", () => {
                 />
               );
 
-              expect(textarea.find(Input).prop("aria-describedby")).toBe(
-                `${id}-field-help ${id}-validation-icon`
-              );
+              it('should render a valid "aria-describeby"', () => {
```
robinzigmond, line 258:
```suggestion
              it('should render a valid "aria-describedby"', () => {
```

src/components/textarea/textarea.spec.tsx
```diff
@@ -460,21 +468,19 @@ describe("componentWillUnmount", () => {
...
-            reverse
           />
         </CarbonProvider>
       );
 
     describe("label width and align props", () => {
       it("default to undefined", () => {
         const wrapper = renderWithNewValidations({});
-        const { labelAlign, labelInline, reverse, labelWidth } = wrapper
+        const { labelAlign, labelInline, labelWidth } = wrapper
```
robinzigmond, line 478: I'm probably missing something but it's not clear to me why you've removed the test for `reverse`, since this isn't a prop that's been removed.

src/components/textarea/textarea.stories.mdx
```diff
@@ -32,160 +32,96 @@ import Textarea from "carbon-react/lib/components/textarea";
...
 ### AutoFocus
 
 <Canvas>
-  <Story name="autoFocus" parameters={{ chromatic: { disable: true } }}>
-    <Textarea label="Textarea" autoFocus />
-  </Story>
+  <Story
+    name="autoFocus"
+    story={stories.autoFocusStory}
+    parameters={{ chromatic: { disable: true } }}
```
robinzigmond, line 62: from previous experience the "disable chromatic" and similar params here don't seem to work when included on the `Story` component here. You instead have to do it in the tsx file: `autoFocusStory.parameters = { chromatic: {disable: true}};`. I could be wrong, but it's safer this way.

src/components/textarea/textarea.stories.mdx
```diff
@@ -32,160 +32,96 @@ import Textarea from "carbon-react/lib/components/textarea";
...
-          value={value}
-          onChange={({ target }) => setValue(target.value)}
-        />
-      );
-    }}
-  </Story>
+  <Story
+    name="expandable"
+    story={stories.expandableStory}
+    parameters={{ chromatic: { disable: true } }}
```
robinzigmond, line 72: see comment above

src/components/textarea/textarea.stories.mdx
```diff
@@ -32,160 +32,96 @@ import Textarea from "carbon-react/lib/components/textarea";
...
+  <Story name="with characterLimit" story={stories.characterLimitStory} />
 </Canvas>
 
 ### With characterLimit - not enforced
 
 <Canvas>
   <Story
     name="with characterLimit - not enforced"
+    story={stories.unenforcedCharacterLimitStory}
     parameters={{ chromatic: { disable: true } }}
```
robinzigmond, line 88: see comment above on disabling chromatic

src/components/textarea/textarea.stories.mdx
```diff
@@ -228,46 +149,14 @@ It is possible to use the `tooltipPosition` to override the default placement of
   <Story
     name="validations - string - with tooltipPosition overriden - component"
     parameters={{ chromatic: { disable: true } }}
```
robinzigmond, line 151: see above comment on disabling chromatic

src/components/textarea/textarea.stories.mdx
```diff
@@ -276,68 +165,20 @@ It is possible to use the `tooltipPosition` to override the default placement of
   <Story
     name="validations - string - with tooltipPosition overriden - label"
     parameters={{ chromatic: { disable: true } }}
```
robinzigmond, line 167: see above comment on disabling chromatic

src/components/textarea/textarea.component.tsx
grabkowski: yes, `istanbul ignore` would not be needed after doing that

src/components/textarea/textarea.spec.tsx
nineteen88: This is what I wanted to speak about. I spoke to @edleeks87 about this in the office, as I also wasn't sure what it meant. He suggested that the test was kind of superfluous and realistically this ticket wasn't to address these kinds of issues - it will be addressed when we move to RTL. However, we can remove the instance check and do `.getDOMNode()` as you suggest Robin and that works. We just wanted to check if this is what you meant here, @grabkowski?

src/components/textarea/textarea.component.tsx
nineteen88: Probably a case of vscode automatically creating the import. Will correct

src/components/textarea/textarea.component.tsx
nineteen88: I too this type definition from @mkrds' code in the [Textbox component](https://github.com/Sage/carbon/blob/FE-4839-convert-textarea-to-typescript/src/components/textbox/textbox.component.tsx#L186). It seems to require the comma.

src/components/textarea/textarea.spec.tsx
nineteen88: I'm honestly not sure why this was removed. Can't work out what the reasoning was. Re-added because doesn't seem to be any issues with having it and I'm with you, it's not clear why it was removed in the first place

src/components/textarea/textarea.spec.tsx
```diff
@@ -23,18 +23,25 @@ import StyledTextarea from "./textarea.style";
...
   describe("when textarea is rendered with default props", () => {
-    let textarea;
+    let textarea: HTMLInputElement;
 
     beforeAll(() => {
       wrapper = renderTextarea();
-      textarea = wrapper.find("textarea").instance();
+      textarea = wrapper
+        .find("textarea")
+        // .instance() as unknown) as HTMLInputElement;
```
robinzigmond, line 43: is there a reason for commenting this line out rather than simply removing it? Git still has the history in the (unlikely) event we want to go back to using `.instance`

robinzigmond (dismissed):

grabkowski (dismissed):

ZhuoyuJin (approved):

grabkowski (approved):

ZhuoyuJin (approved):

PR #5624, opened 2022-11-21
Sidebar Header and Form Footer Padding FE-5391

edleeks87:
### Proposed behaviour

Adds `footerPadding` prop to support padding interface on Form footer component.
Updates `form-style-util` to ensure any padding overrides applied to footer when
a child of `Sidebar` are only applied when no value is passed to `footerPadding`
prop

Surfaces `headerPadding` prop to support the ability to override the padding in the `Sidebar`
`header`

image

### Current behaviour

No support for custom padding in `Sidebar` header or `Form` footer

### Checklist

#### QA

### Testing instructions

See linked sandbox and added stories, check that when Form footer has padding the default behaviour is overridden (no padding is applied by the `form-style-util`)

https://codesandbox.io/s/tender-sound-7blp6k?file=/src/index.js:1122-1123


grabkowski (approved):

src/components/sidebar/__internal__/sidebar-header/sidebar-header.component.tsx
```diff
@@ -1,18 +1,34 @@
...
+  children,
+  id,
+  closeIcon,
+  ...rest
+}: SidebarHeaderProps) => (
+  <StyledSidebarHeader
+    hasClose={!!closeIcon}
+    id={id}
+    data-component="sidebar-header"
+    p="27px 32px 32px"
```
Parsium, line 24: Would we be able to specify the padding in the `style.ts` file instead?

edleeks87: It's possible to move it back to the style file and have it work. I moved it to avoid having the padding style applied twice with one overwritten, basically to avoid convoluting the styles unnecessarily (albeit this time the styles aren't massively so already)

with default in style file: 
image

with default set in component:
image

divyajindel (approved):

PR #5626, opened 2022-11-21
refactor(simple-color-picker): ts conversion - FE-4833

mkrds:
### Proposed behaviour

This PR contains TypeScript conversion of SimpleColorPicker component
### Checklist

#### QA

### Testing instructions

SimpleColorPicker regression testing is recommended


src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -41,21 +68,53 @@ const SimpleColorPicker = (props) => {
...
+  );
+
+  const hasProperChildren = useMemo(() => {
+    let hasSimpleColorChildren = true;
+
+    React.Children.toArray(children).forEach((child) => {
+      if (
+        typeof child === "string" ||
+        (React.isValidElement(child) &&
+          (child.type as React.FunctionComponent).displayName !== "SimpleColor")
```
grabkowski, line 88: I think that we need a safeguard in case a child is not a valid element

grabkowski (changes requested):

src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -122,19 +181,25 @@ const SimpleColorPicker = (props) => {
 
       e.preventDefault();
 
-      let itemIndex;
+      let itemIndex: number | string | null = null;
```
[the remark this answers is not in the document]
mkrds: `target.getAttribute("data-item-down")` returns a string
`navigationGrid.findIndex(position)` returns number

And I have to initialize it with `null` value otherwise I get this error: `Variable 'itemIndex' is used before being assigned.ts(2454)`

src/components/simple-color-picker/simple-color-picker.component.tsx
mkrds: I'm not sure if I understand to be honest

src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -41,21 +68,53 @@ const SimpleColorPicker = (props) => {
...
-  ]);
+  const filteredChildren = useMemo(
+    () =>
+      React.Children.toArray(children).filter((child) =>
+        React.isValidElement(child)
+      ) as React.FunctionComponentElement<
+        SimpleColorProps & { ref: React.RefObject<HTMLInputElement> }
+      >[],
+    [children]
+  );
```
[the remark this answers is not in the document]
mkrds: To be honest I don't really see any advantage of using your proposal over mine

src/components/simple-color-picker/simple-color-picker.component.tsx
grabkowski: if one of the values passed down in an array will be undefined or a boolean it will break the component

src/components/simple-color-picker/simple-color-picker.component.tsx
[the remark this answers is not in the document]
grabkowski: I actually did it the same way as you in `AccordionGroup` long time ago, and I think that solution is not right, as we're only forcing a type here, but still that array could contain elements other than `SimpleColor`. I think that you should at least move this code below the invariant error, so it would be triggered before the array is set.

You should also remove the ref part, as `FunctionComponentElement` already has that prop in it's definition.

src/components/simple-color-picker/simple-color-picker.component.tsx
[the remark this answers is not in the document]
grabkowski: we should avoid having this kind of multi-type variables in TypeScript.
The variable should be a `number` and the result of the `getAttribute` should be converted to that type.
Possibility of that being null shows us that there is something wrong with the code.

src/components/simple-color-picker/simple-color-picker.component.tsx
grabkowski: ok, I forgot that the falsy values would be filtered out in `toArray`

grabkowski (approved):

grabkowski (approved):

src/components/simple-color-picker/simple-color-picker.spec.tsx
```diff
@@ -71,19 +81,23 @@ describe("SimpleColorPicker", () => {
...
         onChange = jest.fn();
-        wrapper = render({ maxWidth: "58", childWith: "58", onChange });
+        wrapper = render({ maxWidth: "58", childWidth: "58", onChange });
       });
       it("fires onKeyDown callback if provided", () => {
         const onKeyDown = jest.fn();
         wrapper = render({
           maxWidth: "58",
-          childWith: "58",
+          childWidth: "58",
```
robinzigmond, line 100: I'm curious how these tests even passed before with the typo in the prop name! I only mention this because it hints to me that the test is not actually very useful.

src/components/simple-color-picker/simple-color-picker.spec.tsx
```diff
@@ -403,16 +489,19 @@ describe("SimpleColorPicker", () => {
...
     beforeEach(() => {
       document.addEventListener = jest.fn((eventName, callback) => {
         if (eventName === "mousedown") {
-          documentMousedownCallback = callback;
+          documentMousedownCallback = callback as EventListener;
         }
       });
       fireDocumentMousedown = () => {
-        const customEvent = { target: document };
+        const customEvent = ({ target: document } as unknown) as Event;
```
robinzigmond, line 504: this double-assertion through `unknown` is pretty ugly. I don't have an easy suggestion but it feels like there should be another better way of doing this? No worries though if you already spent time on this and couldn't find one (as it's only a test file and one we'll eventually be replacing with RTL).

src/components/simple-color-picker/simple-color-picker.style.ts
```diff
@@ -1,10 +1,15 @@
...
 
+import { ValidationProps } from "../../__internal__/validations";
 import StyledValidationIcon from "../../__internal__/validations/validation-icon.style";
 
 const BORDER_WIDTH = 2;
-const getRoundedMaxWidth = (maxWidth, childWidth) =>
-  Math.floor(maxWidth / childWidth) * childWidth;
+
+type MaxWidth = string | number;
+type ChildWidth = string | number;
```
robinzigmond, line 9: is there any reason to use 2 separate types here? Why not just `type CssWidth = string | number` (or whatever name you prefer, I'm not suggesting that's the best name, just the first one I thought of) and use it for both arguments/props?

src/components/simple-color-picker/simple-color-picker.spec.tsx
mkrds: AFAIK theres unfortunately no other way to mock the `target` of the `Event` so I had to brute force mock it  :)
https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

src/components/simple-color-picker/simple-color-picker.spec.tsx
mkrds: Yeah, the reason why it passed before is that the `SimpleColorPicker` has a default value for `childWidth` equal to `58` set already.

src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -150,15 +215,22 @@ const SimpleColorPicker = (props) => {
         }
       }
 
-      const item = navigationGrid[itemIndex].ref.current;
-      item.focus();
-      item.click();
+      /* istanbul ignore else */
+      if (itemIndex || itemIndex === 0) {
+        const item = gridItemRefs[+itemIndex].current;
```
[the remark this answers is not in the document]
mkrds: You're right, I've changed the `number` type as Kuba suggested and you're right, type conversion is redundant in this case

src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -41,21 +68,53 @@ const SimpleColorPicker = (props) => {
...
-  const myRef = useRef(null);
+  invariant(
+    hasProperChildren,
+    `SimpleColorPicker accepts only children of type \`${SimpleColor.displayName}\`.`
+  );
+
+  const filteredChildren = useMemo(
+    () =>
+      React.Children.toArray(children).filter((child) =>
+        React.isValidElement(child)
```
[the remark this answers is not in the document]
mkrds: It is a slight change, you're right, but it's not really breaking anything and it actually might prevent some errors caused by incorrect use of the component.

src/components/simple-color-picker/simple-color-picker.spec.tsx
mkrds: I've changed these values to be slightly different now

robinzigmond (dismissed):

grabkowski (dismissed):

robinzigmond (approved):

grabkowski (dismissed):

divyajindel (dismissed):

grabkowski (approved):

divyajindel (approved):

PR #5628, opened 2022-11-22
refactor(radio-button): convert files to typescript (FE-4825)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/checkbox/checkbox.style.ts
```diff
@@ -12,7 +12,7 @@ import StyledValidationIcon from "../../__internal__/validations/validation-icon
 import baseTheme from "../../style/themes/base";
 import { ValidationProps } from "../../__internal__/validations";
 
-interface StyledCheckboxProps extends ValidationProps, MarginProps {
+export interface StyledCheckboxProps extends ValidationProps, MarginProps {
```
mkrds, line 15: It's only used in this file so this export is probably unnecessary

src/components/radio-button/radio-button.style.ts
```diff
@@ -9,7 +9,15 @@ import { StyledLabelContainer } from "../../__internal__/label/label.style";
 import baseTheme from "../../style/themes/base";
 import FormFieldStyle from "../../__internal__/form-field/form-field.style";
 
-const RadioButtonStyle = styled(CheckboxStyle)`
+export interface RadioButtonProps {
```
mkrds, line 12: For the sake of consistency you should probably import RadioButtonProps from `radio-button.component.tsx` and pick props the ones used by RadioButtonStyle instead of redefining them again

src/components/radio-button/radio-button.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  extends CommonCheckableInputProps,
+    MarginProps {
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-component"?: string;
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-element"?: string;
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-role"?: string;
+  /** When true, sets the component in line (for RadioButtonGroup) */
+  inline?: boolean;
```
mkrds, line 26: `inline` is already defined in `InternalRadioButtonProps`, also I'm not sure what is the reason for either `inline` or `required` being internal props? They can actually be used by library consumers can't they?

src/components/radio-button/radio-button.spec.tsx
```diff
@@ -46,48 +49,44 @@ const borderColorsByValidationTypes = {
 describe("RadioButton", () => {
   describe("propTypes", () => {
     it("does not allow a children prop", () => {
-      const consoleSpy = jest.spyOn(console, "error");
-      render({ children: "someChildren" });
+      const consoleSpy = jest
+        .spyOn(console, "error")
+        .mockImplementation(() => {});
```
mkrds, line 54: You left this mock just to not pollute the logs during running the unit tests?

src/components/radio-button/radio-button.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+      ...props
+    }: RadioButtonProps & InternalRadioButtonProps,
+    ref
+  ) => {
+    const marginProps = filterStyledSystemMarginProps(props);
+    const handleChange = useCallback(
+      (event: React.ChangeEvent<HTMLInputElement>) => {
+        /* istanbul ignore else */
+        if (onChange) {
+          onChange(event);
```
mkrds, line 83: FYI: You could always do `onChange?.(event)`

src/components/radio-button/radio-button.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  onClick?: (ev: React.MouseEvent<HTMLInputElement>) => void;
+  /** the value of the Radio Button, passed on form submit */
+  value: string;
+  /** Overrides the default tooltip position */
+  tooltipPosition?: "top" | "bottom" | "left" | "right";
+  /** Aria label for rendered help component */
+  helpAriaLabel?: string;
+}
+
+const RadioButton = React.forwardRef<
```
mkrds, line 39:
```suggestion
export const RadioButton = React.forwardRef<
```

src/components/radio-button/radio-button.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+          </CheckableInput>
+        </RadioButtonStyle>
+      </TooltipProvider>
+    );
+  }
+);
+
+RadioButton.displayName = "RadioButton";
+
+export { RadioButton as PrivateRadioButton };
```
mkrds, line 162: I think this line might be removed and we could use the named export as suggested above

src/components/radio-button/radio-button.spec.tsx
grabkowski: yes, that's the exact reason for it to be here

src/__internal__/radio-button-mapper/radio-button-mapper.spec.tsx
```diff
@@ -244,36 +237,17 @@ describe("RadioButtonMapper", () => {
...
-        radio
-          .at(0)
-          .props()
-          .onBlur({ target: radio.at(0).getDOMNode() });
-      });
-
-      expect(onBlur).toHaveBeenCalled();
-    });
-
-    it("onChange handler is called when a radio button is clicked", () => {
```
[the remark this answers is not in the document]
grabkowski: this is not a functionality that we should test in this util, as it is dependent on the element that will be passed to this component as a child

src/components/checkbox/checkbox.style.ts
grabkowski: looks like a leftover from my previous refactor approach, removed

src/components/radio-button/radio-button.component.tsx
grabkowski: this prop should be passed down from the `RadioButtonGroup` component, the same goes for the 'required` prop. Looks like it was left there as a mistake.

src/components/radio-button/radio-button.component.tsx
grabkowski: good point - I keep forgetting about optional chaining

src/components/radio-button/radio-button.component.tsx
grabkowski: actually I don't think that it is needed at all

mkrds (approved):

divyajindel (approved):

PR #5651, opened 2022-11-28
Sage/DSY-1757 Breadcrumbs Component (FE-5562)

DobroTora:
### Proposed behaviour

Screenshot 2023-02-07 at 14 24 58

### Current behaviour
This is a new component.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs-test.stories.tsx
+++ b/src/components/breadcrumbs/breadcrumbs-test.stories.tsx
@@ -1,0 +1,52 @@
+import React from "react";
+import { Breadcrumbs } from ".";
+import { Crumb } from "./crumb";
+
+export default {
+  title: "Breadcrumbs/Test",
+  includeStories: "DefaultCrumb",
+  parameters: {
+    info: { disable: true },
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
+  argTypes: {
+    isCurrent: {
+      control: {
+        type: "boolean",
+      },
+    },
+    href: {
+      control: {
+        type: "text",
+      },
+    },
+  },
+};
+
+export const Default = ({ ...args }) => {
+  return (
+    <Breadcrumbs {...args}>
+      <Crumb href="#">Breadcrumb 1</Crumb>
+      <Crumb href="#">Breadcrumb 2</Crumb>
+      <Crumb href="#">Breadcrumb 3</Crumb>
+      <Crumb href="#" isCurrent>
+        Current Page
+      </Crumb>
+    </Breadcrumbs>
+  );
+};
+
+export const DefaultCrumb = ({ ...args }) => {
+  return (
+    <Breadcrumbs>
+      <Crumb href="#" {...args}>
+        Breadcrumb 1
+      </Crumb>
+    </Breadcrumbs>
+  );
+};
+
+Default.storyName = "default";
+DefaultCrumb.storyName = "default crumb";
--- a/src/components/breadcrumbs/breadcrumbs.component.tsx
+++ b/src/components/breadcrumbs/breadcrumbs.component.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import tagComponent, { TagProps } from "../../__internal__/utils/helpers/tags";
+import StyledBreadcrumbs from "./breadcrumbs.style";
+import useLocale from "../../hooks/__internal__/useLocale";
+
+export interface BreadcrumbsProps extends TagProps {
+  /** Child crumbs to display */
+  children: React.ReactNode;
+}
+
+export const Breadcrumbs = React.forwardRef<HTMLElement, BreadcrumbsProps>(
+  ({ children, ...rest }: BreadcrumbsProps, ref) => {
+    const l = useLocale();
+    return (
+      <StyledBreadcrumbs
+        ref={ref}
+        role="navigation"
+        {...tagComponent("breadcrumbs", rest)}
+        aria-label={l.breadcrumbs.ariaLabel()}
+      >
+        <ol>{children}</ol>
+      </StyledBreadcrumbs>
+    );
+  }
+);
+
+Breadcrumbs.displayName = "Breadcrumbs";
+
+export default Breadcrumbs;
--- a/src/components/breadcrumbs/breadcrumbs.spec.tsx
+++ b/src/components/breadcrumbs/breadcrumbs.spec.tsx
@@ -1,0 +1,21 @@
+import React from "react";
+import "@testing-library/jest-dom";
+import { screen, render } from "@testing-library/react";
+import Breadcrumbs from "./breadcrumbs.component";
+import Crumb from "./crumb/crumb.component";
+
+test("renders all child crumbs correctly", () => {
+  render(
+    <Breadcrumbs>
+      <Crumb href="#">Breadcrumb</Crumb>
+      <Crumb href="#">Breadcrumb</Crumb>
+      <Crumb href="#">Breadcrumb</Crumb>
+      <Crumb href="#" isCurrent>
+        Breadcrumb
+      </Crumb>
+    </Breadcrumbs>
+  );
+
+  const crumbElement = screen.getAllByText("Breadcrumb");
+  expect(crumbElement.length).toBe(4);
+});
--- a/src/components/breadcrumbs/breadcrumbs.stories.mdx
+++ b/src/components/breadcrumbs/breadcrumbs.stories.mdx
@@ -1,0 +1,68 @@
+import { Meta, Story, Canvas, ArgsTable } from "@storybook/addon-docs";
+import { Breadcrumbs, Crumb } from "./index";
+import * as stories from "./breadcrumbs.stories.tsx";
+import TranslationKeysTable from "../../../.storybook/utils/translation-keys-table";
+
+<Meta title="Breadcrumbs" parameters={{ info: { disable: true } }} />
+
+# Breadcrumbs
+
+  <a
+  target="_blank"
+  href="https://zeroheight.com/2ccf2b601/p/32c1b6-breadcrumbs/b/87b2c7"
+  style={{ color: "#007E45", fontWeight: "bold", textDecoration: "underline" }}
+  >
+    Product Design System component
+  </a>
+
+Breadcrumbs are a secondary navigation aid that helps users easily understand their location.
+
+## Contents
+
+- [Quick Start](#quick-start)
+- [Examples](#examples)
+- [Props](#props)
+
+## Quick Start
+
+To use Breadcrumbs component you need to import `Breadcrumbs` and pass an array of `Crumb`s.
+
+```javascript
+import { Breadcrumbs, Crumb } from "carbon-react/lib/components/breadcrumbs";
+```
+
+## Examples
+
+### Breadcrumbs
+
+<Canvas>
+  <Story name="default" story={stories.DefaultBreadcrumbs} />
+</Canvas>
+
+## Props
+
+### Breadcrumbs
+
+<ArgsTable
+   of={Breadcrumbs}
+ />
+
+### Crumb
+
+<ArgsTable of={Crumb} />
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object to the
+[i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
+
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "breadcrumbs.ariaLabel",
+      description: "The text for the aria-label attribute",
+      type: "func",
+      returnType: "string",
+    },
+]}
+/>
--- a/src/components/breadcrumbs/breadcrumbs.style.ts
+++ b/src/components/breadcrumbs/breadcrumbs.style.ts
@@ -1,0 +1,11 @@
+import styled from "styled-components";
+
+const StyledBreadcrumbs = styled.nav`
+  ol {
+    list-style: none;
+    display: flex;
+    flex-wrap: wrap;
+  }
+`;
+
+export default StyledBreadcrumbs;
--- a/src/components/breadcrumbs/breadcrumbs.test.js
+++ b/src/components/breadcrumbs/breadcrumbs.test.js
@@ -1,0 +1,96 @@
+import React from "react";
+import * as testStories from "./breadcrumbs-test.stories";
+import CypressMountWithProviders from "../../../cypress/support/component-helper/cypress-mount";
+import {
+  breadcrumbsComponent,
+  crumb,
+} from "../../../cypress/locators/breadcrumbs";
+import { CHARACTERS } from "../../../cypress/support/component-helper/constants";
+
+context("Testing Breadcrumbs component", () => {
+  describe("should render Breadcrumbs component", () => {
+    it("should check Breadcrumbs children is set visible", () => {
+      CypressMountWithProviders(<testStories.Default />);
+
+      breadcrumbsComponent()
+        .find("ol")
+        .children()
+        .should("have.length", 4)
+        .and("be.visible");
+    });
+
+    it("should check Breadcrumbs on hover color", () => {
+      CypressMountWithProviders(<testStories.Default />);
+
+      crumb(1)
+        .realHover()
+        .find("a")
+        .should("have.css", "color", "rgb(0, 103, 56)");
+    });
+
+    it("should check Breadcrumbs on focus", () => {
+      CypressMountWithProviders(<testStories.Default />);
+
+      cy.get("body").tab();
+      crumb(0).find("a").should("be.focused");
+      cy.focused().tab().tab();
+      crumb(2).find("a").should("be.focused");
+    });
+
+    it.each([
+      [true, "aria-current", "page"],
+      [false, "href", "#"],
+    ])(
+      "should check Crumb with isCurrent prop is %s",
+      (boolean, attr, value) => {
+        CypressMountWithProviders(
+          <testStories.DefaultCrumb isCurrent={boolean} />
+        );
+
+        crumb(0).find("a").should("have.attr", attr, value);
+      }
+    );
+
+    it("should check Crumb with href prop set to default val", () => {
+      CypressMountWithProviders(
+        <testStories.DefaultCrumb href={CHARACTERS.STANDARD} />
+      );
+
+      crumb(0).find("a").should("have.attr", "href", CHARACTERS.STANDARD);
+    });
+  });
+
+  describe("should check Crumb props", () => {
+    it.each([
+      ["not.exist", true],
+      ["be.visible", false],
+    ])("should check Crumb divider is %s", (assertion, boolean) => {
+      CypressMountWithProviders(
+        <testStories.DefaultCrumb isCurrent={boolean} />
+      );
+
+      crumb(0).children().eq(1).should(assertion);
+      if (!boolean) {
+        crumb(0)
+          .children()
+          .eq(1)
+          .then(($el) => {
+            expect($el).to.have.text("/");
+            expect($el).to.have.css("color", "rgba(0, 0, 0, 0.55)");
+          });
+      }
+    });
+  });
+
+  describe("Accessibility tests for Breadcrumbs component", () => {
+    it("should pass accessibilty tests for Breadcrumbs default story", () => {
+      CypressMountWithProviders(<testStories.Default />);
+      cy.checkAccessibility();
+    });
+
+    it("should pass accessibilty tests for Crumb default story", () => {
+      CypressMountWithProviders(<testStories.DefaultCrumb />);
+      cy.checkAccessibility();
+    });
+  });
+});
--- a/src/components/breadcrumbs/crumb/crumb.component.tsx
+++ b/src/components/breadcrumbs/crumb/crumb.component.tsx
@@ -1,0 +1,53 @@
+import { LinkProps } from "components/link/link.component";
+import React from "react";
+import tagComponent, {
+  TagProps,
+} from "../../../__internal__/utils/helpers/tags";
+import { StyledCrumb, Divider } from "./crumb.style";
+
+export interface CrumbProps
+  extends Omit<
+      LinkProps,
+      | "tooltipMessage"
+      | "tooltipPosition"
+      | "iconType"
+      | "iconAlign"
+      | "isSkipLink"
+      | "isDarkBackground"
+      | "ariaLabel"
+      | "className"
+      | "variant"
+      | "target"
+      | "rel"
+      | "icon"
+      | "disabled"
+    >,
+    TagProps {
+  /** This sets the Crumb to current, does not render Link */
+  isCurrent?: boolean;
+  /** The href string for Crumb Link */
+  href: string;
+}
+
+const Crumb = React.forwardRef<HTMLLinkElement, CrumbProps>(
+  ({ href, isCurrent, children, ...rest }: CrumbProps, ref) => (
+    <li>
+      <StyledCrumb
+        ref={ref}
+        isCurrent={isCurrent}
+        aria-current={isCurrent ? "page" : undefined}
+        {...tagComponent("crumb", rest)}
+        {...(!isCurrent && {
+          href,
+        })}
+      >
+        {children}
+      </StyledCrumb>
+      {!isCurrent && <Divider />}
+    </li>
+  )
+);
+
+Crumb.displayName = "Crumb";
+
+export default Crumb;
--- a/src/components/breadcrumbs/crumb/crumb.spec.tsx
+++ b/src/components/breadcrumbs/crumb/crumb.spec.tsx
@@ -1,0 +1,62 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import Crumb from "./crumb.component";
+
+const LINK_TEXT = "Link text";
+
+describe("Crumb", () => {
+  it("passes href to the anchor element isCurrent is false", () => {
+    render(<Crumb href="foo">{LINK_TEXT}</Crumb>);
+
+    const link = screen.getByText(LINK_TEXT);
+    expect(link).toBeInTheDocument();
+    expect(link.closest("a")).toHaveAttribute("href", "foo");
+  });
+
+  it("does not pass href to the anchor element when isCurrent is true", () => {
+    render(
+      <Crumb href="foo" isCurrent>
+        {LINK_TEXT}
+      </Crumb>
+    );
+
+    const anchor = screen.getByText(LINK_TEXT).closest("a");
+    expect(anchor).not.toHaveAttribute("href", "foo");
+
+    expect(anchor).toHaveStyle({
+      color: "var(--colorsYin055)",
+      textDecoration: "none",
+      cursor: "text",
+    });
+  });
+
+  it("calls the handleClick callback if one is passed and isCurrent is false", async () => {
+    const handleClickFn = jest.fn();
+
+    render(
+      <Crumb href="#" onClick={handleClickFn}>
+        {LINK_TEXT}
+      </Crumb>
+    );
+
+    const link = screen.getByText(LINK_TEXT);
+    await userEvent.click(link);
+
+    expect(handleClickFn).toHaveBeenCalledTimes(0);
+  });
+
+  it("does not call the handleClick callback if one is passed and isCurrent is true", async () => {
+    const handleClickFn = jest.fn();
+
+    render(
+      <Crumb href="#" onClick={handleClickFn}>
+        {LINK_TEXT}
+      </Crumb>
+    );
+    const link = screen.getByText(LINK_TEXT);
+    await userEvent.click(link);
+
+    expect(handleClickFn).not.toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/components/breadcrumbs/crumb/crumb.style.ts
+++ b/src/components/breadcrumbs/crumb/crumb.style.ts
@@ -1,0 +1,37 @@
+import styled, { css } from "styled-components";
+import Link, { LinkProps } from "../../link";
+
+interface StyleCrumbProps extends LinkProps {
+  isCurrent?: boolean;
+}
+
+export const StyledCrumb = styled(Link)<StyleCrumbProps>`
+  font: var(--typographyLinkTextM);
+  ${({ isCurrent }) =>
+    isCurrent &&
+    css`
+      a {
+        color: var(--colorsUtilityYin090);
+        text-decoration: none;
+        font: var(--typographyBreadcrumbCurrentPageM);
+        cursor: text;
+        :hover {
+          color: var(--colorsUtilityYin090);
+          text-decoration: none;
+          cursor: text;
+        }
+      }
+    `}
+`;
+
+export const Divider = styled.span.attrs({
+  children: "/",
+  "aria-hidden": "true",
+})`
+  margin: 0px var(--spacing050) 0px var(--spacing100);
+  line-height: 16px;
+  font: var(--typographyBreadcrumbSeparatorM);
+  color: var(--colorsUtilityYin055);
+`;
+
+export default { StyledCrumb, Divider };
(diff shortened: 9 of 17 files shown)


edleeks87:
Please can you add description for this PR so reviewers have some context?

ljemmo:
@DobroTora Just reviewed the storybook build and found the following to action:

- Can the dividing line use the following token: **{colors.utility.yin.055}.** It looks a lot lighter in the build.
Screenshot 2022-12-06 at 11 03 45

- can we simply use forward slash characters as opposed to a rotated div to mirror the figma component or is there an accessibility reason to avoid this?
- We need to support wrapping for smaller devices out of the box for accessibility reasons. See[ example on ZH.](https://zeroheight.com/2ccf2b601/v/0/p/32c1b6-breadcrumbs/t/630e7a). I imagine this could be quickly achieved via flexbox?

DobroTora:
> @DobroTora Just reviewed the storybook build and found the following to action:
> 
> * Can the dividing line use the following token: **{colors.utility.yin.055}.** It looks a lot lighter in the build and it's currently not aligned horizontally like in the figma component.
>   Screenshot 2022-12-06 at 11 03 45
> * can we simply use forward slash characters as opposed to a rotated div to mirror the figma component or is there an accessibility reason to avoid this?
> * We need to support wrapping for smaller devices out of the box for accessibility reasons. See[ example on ZH.](https://zeroheight.com/2ccf2b601/v/0/p/32c1b6-breadcrumbs/t/630e7a). I imagine this could be quickly achieved via flexbox?
> * Does your component allow passing in of more than 3 links? We should allow passing in of as many as required by devs.
> * Does your component allow passing in of the current page name - this is probably a prop that will need surfacing?

I did you a flexbox for this component.

edleeks87 (commented):
Hi @DobroTora I realise this is still work in progress but I thought I'd have a look after we chatted yesterday. I've left a couple of comments including a sandbox of a potential implementation for this. Apologies if this was already in hand or you have a different idea for how to achieve the designs. Feel free to reach out if you want to chat further about it 👍

DobroTora:
> Please can you add description for this PR so reviewers have some context?

done

src/components/breadcrumbs/breadcrumbs.spec.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import "@testing-library/jest-dom";
+import { screen, render } from "@testing-library/react";
+import StyledBreadcrumbs from "./breadcrumbs.style";
+import Breadcrumbs from "./breadcrumbs.component";
+import StyledCrumb from "./crumb/crumb.style";
+
+test("Breadcrumbs component", () => {
+  render(
+    <StyledBreadcrumbs>
```
edleeks87, line 10: You should render the React components rather than the styled components
```
<Breadcrumbs>
      <Crumb text="Crumb 1" href="#" />
      <Crumb text="Crumb 2" href="#" />
      <Crumb text="Crumb 3" href="#" />
      <Crumb text="Crumb 4" href="#" isCurrent />
    </Breadcrumbs>
```

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+      onClick: (ev) => onClick?.(ev, crumbId || text, to),
+    })}
+  >
+    {text}
+  </StyledCrumb>
+);
+
+Crumb.displayName = "Crumb";
+
+export default Crumb;
```
edleeks87, line 53: you need a empty line here

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+  it("does not call the onClick callback if one is passed and isCurrent is true", () => {
+    const onClickFn = jest.fn();
+
+    render(<Crumb href="foo" text={LINK_TEXT} onClick={onClickFn} isCurrent />);
+    const link = screen.getByText(LINK_TEXT);
+    fireEvent.click(link);
+
+    expect(onClickFn).not.toHaveBeenCalled();
+  });
+});
```
edleeks87, line 48: empty line

src/components/breadcrumbs/crumb/crumb.style.ts
```diff
@@ -0,0 +1,29 @@
...
+  &:not(:last-of-type)::after {
+    content: "/";
+    margin: 0 8px;
+    line-height: 16px;
+    vertical-align: text-top;
+    color: gray;
+  }
+`;
+
+export default StyledCrumb;
```
edleeks87, line 29: empty line

src/components/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,4 @@
+export { default as Breadcrumbs } from "./breadcrumbs.component";
+export { default as Crumb } from "./crumb";
+export type { BreadcrumbsProps } from "./breadcrumbs.component";
+export type { CrumbProps } from "./crumb";
```
edleeks87, line 4: empty line

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it("does not pass href to the anchor element when isCurrent is true", () => {
+    render(<Crumb href="foo" text={LINK_TEXT} isCurrent />);
+
+    const anchor = screen.getByText(LINK_TEXT).closest("a");
+    expect(anchor).not.toHaveAttribute("href", "foo");
+
+    expect(anchor).toHaveStyle({
+      color: "gray",
```
edleeks87, line 23: this will also need updating to use tokens

src/components/breadcrumbs/crumb/crumb.style.ts
```diff
@@ -0,0 +1,29 @@
+import styled, { css } from "styled-components";
+import { CrumbProps } from "./crumb.component";
+import Link from "../../link";
+
+const StyledCrumb = styled(Link)<Pick<CrumbProps, "isCurrent">>`
+  ${({ isCurrent }) =>
+    isCurrent &&
+    css`
+      a {
+        color: gray;
```
edleeks87, line 10: needs to be a token colour

src/components/breadcrumbs/crumb/crumb.style.ts
```diff
@@ -0,0 +1,29 @@
...
+          cursor: text;
+        }
+      }
+    `}
+  &:not(:last-of-type)::after {
+    content: "/";
+    margin: 0 8px;
+    line-height: 16px;
+    vertical-align: text-top;
+    color: gray;
```
edleeks87, line 25: needs to be a token colour

src/components/badge/badge.stories.mdx
```diff
@@ -73,4 +73,4 @@ If the counter is less than 1 badge will not show itself
 
 ### Badge
 
-<ArgsTable of={Badge} />
+<ArgsTable of={Badge} />
```
edleeks87, line 76: Not sure what was updated here or why it needed to be?

src/components/breadcrumbs/__snapshots__/breadcrumbs.spec.tsx.snap
```diff
@@ -0,0 +1,40 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
edleeks87, line 2: You need to update the snapshots when you run the tests, this should have been deleted

`npm test -- -u`

src/components/badge/badge.stories.mdx
DobroTora: I think it was accidental commit from Designer's Notes Removal ....

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+  /** The onClick callback for the Crumb Link */
+  onClick?: (
+    ev:
+      | React.MouseEvent<HTMLAnchorElement>
+      | React.MouseEvent<HTMLButtonElement>
+      | React.KeyboardEvent<HTMLAnchorElement>
+      | React.KeyboardEvent<HTMLButtonElement>,
+    id: string,
+    to?: string
+  ) => null;
```
Parsium, line 23:
```suggestion
  /** A callback function to call when the Crumb link is clicked */
  handleClick?: () => void;
```
I have a couple of points with regarding this prop:

Since this function is being utilised as a callback, personally I wouldn't call this function `onClick` since technically this isn't a browser event handler and could be confusing.

Since we are also surfacing `crumbId` and `to` as props, consumers should have access to these values already in scope so we don't have to reference these again within the callback itself.

On line 44 I noticed we are also using the `text` prop as an identifier if `crumbId` is not set. I think I understand why you have done it this way since I'm assuming we shouldn't have: 

1. Multiple `Breadcrumb` components on the same page (should be unique to one page)
2. Multiple `Crumb`s with the same text
    
If this is the case, I would be tempted to remove `crumbId` and instead make the `text` prop a required prop, since I assume it shouldn't be possible to have a crumb with no text :)

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const LINK_TEXT = "Link text";
+
+describe("Crumb", () => {
+  it("passes href to the anchor element isCurrent is false", () => {
+    render(<Crumb href="foo" text={LINK_TEXT} />);
+
+    const link = screen.getByText(LINK_TEXT);
+    expect(link).toBeInTheDocument();
+    expect(link?.closest("a")).toHaveAttribute("href", "foo");
```
Parsium, line 13:
```suggestion
    expect(link.closest("a")).toHaveAttribute("href", "foo");
```
This assertion will fail if the link is not in the document anyway, so we can just use it on its own

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+    expect(anchor).not.toHaveAttribute("href", "foo");
+
+    expect(anchor).toHaveStyle({
+      color: "var(--colorsYin055)",
+      textDecoration: "none",
+      cursor: "text",
+    });
+  });
+
+  it("calls the onClick callback if one is passed and isCurrent is false", () => {
```
Parsium, line 29: Personally I would opt to use React Testing Library's `userEvent` helper library instead of `fireEvent`, since the helpers it provides simulate user interactions more closely rather than simulating a dispatched DOM event, giving us greater security that our tests mimic how a user would use the component. 

However, I'm happy to accept these two tests as they are here, since the action here isn't complicated

Parsium:
Thanks @DobroTora for writing the tests in RTL rather than Enzyme! This will prevent extra tech debt for us in the future 👍🏼

src/components/breadcrumbs/crumb/crumb.component.tsx
DobroTora: You can run out of sensible synonyms quite quickly in that case :)

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+      color: "var(--colorsYin055)",
+      textDecoration: "none",
+      cursor: "text",
+    });
+  });
+
+  it("calls the onClick callback if one is passed and isCurrent is false", async () => {
+    const onClickFn = jest.fn();
+
+    render(<Crumb href="foo" text={LINK_TEXT} onClick={onClickFn} />);
```
Parsium, line 33: Currently we get the following warning when these click callback tests are run:

```shell
Error: Not implemented: navigation (except hash changes)
          at module.exports (/Users/james.parslow/Development/carbon/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
          at navigateFetch (/Users/james.parslow/Development/carbon/node_modules/jsdom/lib/jsdom/living/window/navigation.js:76:3)
          at exports.navigate (/Users/james.parslow/Development/carbon/node_modules/jsdom/lib/jsdom/living/window/navigation.js:54:3)
          at Timeout._onTimeout (/Users/james.parslow/Development/carbon/node_modules/jsdom/lib/jsdom/living/nodes/HTMLHyperlinkElementUtils-impl.js:81:7)
          at listOnTimeout (node:internal/timers:564:17)
          at processTimers (node:internal/timers:507:7) undefined
```

This is because [jsdom does not support window navigation](https://github.com/jsdom/jsdom/issues/2112#issuecomment-359297866) and it is seeing `foo` as a url to navigate to. To get around this, I would recommend making the `Crumbs` in these tests a dummy link:

```suggestion
    render(<Crumb href="#" text={LINK_TEXT} handleClick={handleClickFn} />);
```

That way you can still test the click callback behaviour, but jsdom won't try to redirect to a new page when the link is clicked.

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const link = screen.getByText(LINK_TEXT);
+    await userEvent.click(link);
+
+    expect(onClickFn).toHaveBeenCalledTimes(0);
```
Parsium, line 37: Since `handleClick` is the callback prop we want to provide, we'll need to update this test and its sad path counterpart below

```suggestion
  it("calls the handleClick callback if one is passed and isCurrent is false", async () => {
    const handleClickFn = jest.fn();

    render(<Crumb href="#" text={LINK_TEXT} handleClick={handleClickFn} />);
    const link = screen.getByText(LINK_TEXT);
    await userEvent.click(link);

    expect(handleClickFn).toHaveBeenCalledTimes(0);
```

src/components/breadcrumbs/breadcrumbs.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+```javascript
+import { Breadcrumbs, Crumb } from "carbon-react/lib/components/breadcrumbs";
+```
+
+## Examples
+
+### Breadcrumbs
+
+<Canvas>
+  <Story name="default" story={BreadcrumbsDefault} />
```
Parsium, line 38:
```suggestion
  <Story name="default" story={stories.DefaultBreadcrumbs} />
```

Parsium (changes requested):

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isCurrent,
+  handleClick,
+  ...rest
+}: CrumbProps) => (
+  <StyledCrumb
+    isCurrent={isCurrent}
+    {...tagComponent("crumb", rest)}
+    {...(!isCurrent && {
+      href,
+      onClick: handleClick ? () => handleClick() : undefined,
```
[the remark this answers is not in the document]
DobroTora: yes, that looks about right @Parsium

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+  href: string;
+  /** The to string for Crumb Link */
+  to?: string;
+  /** This sets the Crumb to current, does not render Link */
+  isCurrent?: boolean;
+  /** A callback function to call when the Crumb link is clicked */
+  handleClick?: () => void;
+  /** The text to render for the Crumb Link */
+  text: string;
+  onClick?: (ev: React.MouseEvent<HTMLElement>) => void;
```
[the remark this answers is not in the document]
edleeks87: I believe onClick is needed so we can meet the pattern for using routing

https://carbon.sage.com/?path=/docs/documentation-usage-with-routing--page

src/components/breadcrumbs/crumb/crumb.component.tsx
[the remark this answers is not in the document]
edleeks87: `handleClick` implies it's an event handler to me. I think we need the onClick as I've mentioned in another comment

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import tagComponent, {
+  TagProps,
+} from "../../../__internal__/utils/helpers/tags";
+import StyledCrumb from "./crumb.style";
+
+export interface CrumbProps extends TagProps {
```
edleeks87, line 7: You can just extend `LinkProps` here and `omit` the ones we don't want to expose

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import tagComponent, {
+  TagProps,
+} from "../../../__internal__/utils/helpers/tags";
+import StyledCrumb from "./crumb.style";
+
+export interface LinkProps extends TagProps {
```
edleeks87, line 7: This is incorrect. The way you had it previously was closer to what we need, something like this would work as Link already supports the interface

```
export interface CrumbProps extends Omit<LinkProps, "toolTipMessage", "tooltipPosition">, TagProps {
 /** This sets the Crumb to current, does not render Link */
  isCurrent?: boolean;
}
```

src/components/breadcrumbs/crumb/crumb.component.tsx
edleeks87: It needs to be onClick so we can maintain the current approach for using routing solutions, that approach requires the event. Eventually we should investigate if we can adopt the approach used in the VerticalMenu for passing router links in Carbon's Link component but for now we should make sure our doc examples work

src/components/breadcrumbs/crumb/crumb.spec.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { Crumb } from "./crumb.component";
+
+const LINK_TEXT = "Link text";
+
+describe("Crumb", () => {
+  it("passes href to the anchor element isCurrent is false", () => {
+    render(<Crumb href="foo" text={LINK_TEXT} />);
```
edleeks87, line 10: If we extend the `LinkProps` we can just pass the text as children

```
<Crumb href="foo">{LINK_TEXT}</Crumb>
```

src/components/badge/badge.stories.mdx
```diff
@@ -81,4 +81,4 @@ If the counter is less than 1 badge will not show itself
...
```
edleeks87, line 84: This file still needs removing, if it is your editor settings that is formatting it (on save) then you can do the following to check out the master copy:

```
git fetch
git checkout origin/master -- origin/src/components/badge/badge.stories.mdx
```

edleeks87 (changes requested):

src/components/breadcrumbs/crumb/crumb.component.tsx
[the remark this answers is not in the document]
Parsium: I agree with @edleeks87, in that case let's replace the `handleClick` prop with `onClick`

src/components/breadcrumbs/crumb/crumb.spec.tsx
Parsium: In this case we could remove the `text` prop and use `children` instead 👍🏼

src/components/breadcrumbs/crumb/crumb.component.tsx
Parsium: Adding @edleeks87's suggestion should also fix that lint error for you 👍🏼

DobroTora: Sorry @edleeks87 - I misunderstood initially your request.

src/components/breadcrumbs/crumb/crumb.style.ts
```diff
@@ -0,0 +1,31 @@
...
+        cursor: text;
+        :hover {
+          color: var(--colorsUtilityYin090);
+          text-decoration: none;
+          cursor: text;
+        }
+      }
+    `}
+  &:not(:last-of-type)::after {
+    content: "/";
```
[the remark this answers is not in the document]
edleeks87: Very good point @Parsium! As we don't get full browser support, do we need to move away from using a pseudo element here and go with a dedicated component that has [aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) set on it instead then? Worth seeing what UX/ @tempertemper  thinks about it at least imo

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+  extends Omit<LinkProps, "toolTipMessage" | "tooltipPosition">,
+    TagProps {
+  /** This sets the Crumb to current, does not render Link */
+  isCurrent?: boolean;
+  /** The href string for Crumb Link */
+  href: string;
+}
+
+export const Crumb = ({ href, isCurrent, children, ...rest }: CrumbProps) => (
+  <li>
```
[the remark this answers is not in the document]
edleeks87: By adding the Divider as a child of the StyledCrumb you are inheriting the styling, it also needs to go after the text etc

src/components/breadcrumbs/breadcrumbs.style.ts
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+
+const StyledBreadcrumbs = styled.nav`
```
edleeks87, line 3:
```
const StyledBreadcrumbs = styled.nav`
  ol {
    list-style: none;
    display: flex;
    flex-wrap: wrap;
  }
`;
```

edleeks87: You need to apply the styling to the `ol` now and remove the list-style etc

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,33 @@
+import { LinkProps } from "components/link/link.component";
+import React from "react";
+import tagComponent, {
+  TagProps,
+} from "../../../__internal__/utils/helpers/tags";
+import StyledCrumb from "./crumb.style";
+import Divider from "./divider.component";
+
+export interface CrumbProps
+  extends Omit<LinkProps, "toolTipMessage" | "tooltipPosition">,
```
edleeks87, line 10: We might need to omit the `icon` prop here as well as the DS spec says to not use them in Breadcrumbs

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+      {...(!isCurrent && {
+        href,
+      })}
+    >
+      {!isCurrent && <Divider aria-hidden="true" />}
+      {children}
+    </StyledCrumb>
+  </li>
+);
+Crumb.displayName = "Crumb";
```
edleeks87, line 33: lets add a default export here

src/components/breadcrumbs/crumb/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Crumb } from "./crumb.component";
```
edleeks87, line 1: `default as Crumb` when you've added the default export

src/components/breadcrumbs/breadcrumbs.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Props
+
+### Breadcrumbs
+
+<ArgsTable of={Breadcrumbs} />
+
+### Crumb
+
+<ArgsTable of={Crumb} />
```
edleeks87, line 49: Lets add a `TranslationKeysTable` here as well, you can see how it was implemented [here](https://github.com/Sage/carbon/blob/master/src/components/batch-selection/batch-selection.stories.mdx) in the past

src/components/breadcrumbs/breadcrumbs.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface BreadcrumbsProps extends TagProps {
+  /** Child crumbs to display */
+  children: React.ReactNode;
+  /** Aria label for breadcrumbs component */
+  ariaLabel?: string;
+  breadcrumbs?: React.ReactNode;
+}
+
+export const Breadcrumbs = ({ children, ...rest }: BreadcrumbsProps) => {
```
edleeks87, line 14: Should we add a forwardRef here as we do it on a lot of other components?

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+export interface CrumbProps
+  extends Omit<LinkProps, "toolTipMessage" | "tooltipPosition">,
+    TagProps {
+  /** This sets the Crumb to current, does not render Link */
+  isCurrent?: boolean;
+  /** The href string for Crumb Link */
+  href: string;
+}
+
+export const Crumb = ({ href, isCurrent, children, ...rest }: CrumbProps) => (
```
edleeks87, line 18: Should we add a forwardRef here as we do it on a lot of other components?

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /** The href string for Crumb Link */
+  href: string;
+}
+
+const Crumb = React.forwardRef(
+  ({ href, isCurrent, children, ...rest }: CrumbProps) => (
+    <li>
+      <StyledCrumb
+        isCurrent={isCurrent}
+        aria-current={isCurrent}
```
Parsium, line 22: Taking another look at `aria-current`, I think we need to [set it to `"page"` similar to this example.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#example)  The screen reader would then announce that link refers to the name of the page the user is on. (It seems `aria-current` can be used in other contexts as well)

```suggestion
        aria-current={isCurrent ? "page" : undefined}
```

tempertemper: `location` could be equally valid. There's some talk about it in the comments of [this blog post](https://tink.uk/using-the-aria-current-attribute/).

Parsium (dismissed):

edleeks87 (dismissed):

src/components/breadcrumbs/breadcrumbs.test.js
```diff
@@ -0,0 +1,94 @@
...
+      CypressMountWithProviders(
+        <testStories.DefaultCrumb href={CHARACTERS.STANDARD} />
+      );
+
+      crumb(0).find("a").should("have.attr", "href", CHARACTERS.STANDARD);
+    });
+  });
+});
+
+describe("should check Crumb props", () => {
```
stephenogorman, line 64: Should this describe be outside the Context above? Or in its own?

src/components/breadcrumbs/breadcrumbs.test.js
ZhuoyuJin, line 1: Can we add a test for onClick event here?

DlgSHi: there is no such prop for `Breadcrumbs` - those props are passed down from `<Link>`

stephenogorman (dismissed):

DlgSHi (dismissed):

DlgSHi (approved):

src/components/breadcrumbs/breadcrumbs.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+<ArgsTable
+   of={Breadcrumbs}
+ />
+
+### Crumb
+
+<ArgsTable of={Crumb} 
+  exclude={[
+     "className",
+     "onKeyDown",
```
edleeks87, line 55: I think the event handlers should be displayed here so `onKeyDown`, `onClick` and `onMouseDown` shouldn't be excluded. IMO the only props excluded should be the ones omitted from the `CrumbProps` interface

`Omit<LinkProps, "toolTipMessage" | "tooltipPosition" | "iconType" | "iconAlign" | "isSkipLink" | "isDarkBackground" | "ariaLabel" | "className">`

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isCurrent?: boolean;
+  /** The href string for Crumb Link */
+  href: string;
+}
+
+const Crumb = React.forwardRef(
+  ({ href, isCurrent, children, ...rest }: CrumbProps) => (
+    <li>
+      <StyledCrumb
+        isCurrent={isCurrent}
```
Parsium, line 21: As we have defined `Crumb` as a forward ref, we'll need to specify the `ref` on the root element like so:

```suggestion
const Crumb = React.forwardRef<HTMLLinkElement, CrumbProps>(
  ({ href, isCurrent, children, ...rest }: CrumbProps, ref) => (
    <li>
      <StyledCrumb
       ref={ref}
        isCurrent={isCurrent}
```

So consumers would be able to do:

```tsx
const crumbRef = React.useRef<HTMLLinkElement>(null);
return (
    // ....
    <Crumb ref={crumbRef} href="...">Breadcrumb 1</Crumb>
);
```

src/components/breadcrumbs/breadcrumbs.component.tsx
```diff
@@ -0,0 +1,28 @@
...
+        role="navigation"
+        {...tagComponent("breadcrumbs", rest)}
+        aria-label={l.breadcrumbs.ariaLabel()}
+      >
+        <ol>{children}</ol>
+      </StyledBreadcrumbs>
+    );
+  }
+);
+
```
Parsium, line 25: Similar to `Crumb`, since this has been defined as a forward ref we will need to specify `ref` on the root element like so:

```suggestion
export const Breadcrumbs = React.forwardRef<HTMLNavElement, BreadcrumbsProps>(
  ({ children, ...rest }: BreadcrumbsProps, ref) => {
    const l = useLocale();
    return (
      <StyledBreadcrumbs
        ref={ref}
        role="navigation"
        {...tagComponent("breadcrumbs", rest)}
        aria-label={l.breadcrumbs.ariaLabel()}
      >
        <ol>{children}</ol>
      </StyledBreadcrumbs>
    );
  }
);

```

DlgSHi (approved):

src/components/breadcrumbs/breadcrumbs.component.tsx
edleeks87: This is still missing a `ref`

src/components/breadcrumbs/crumb/crumb.component.tsx
edleeks87: This is still missing a `ref`

src/components/breadcrumbs/breadcrumbs-test.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Crumb href="#">Breadcrumb 2</Crumb>
+      <Crumb href="#">Breadcrumb 3</Crumb>
+      <Crumb href="#" isCurrent>
+        Current Page
+      </Crumb>
+    </Breadcrumbs>
+  );
+};
+
+export const DefaultCrumb = ({ ...args }) => {
```
edleeks87, line 41: Sorry I've just noticed this, is there any point in displaying a single crumb? The story looks pretty weird and I don't think we'll ever have it like this in production? I think this story can just be deleted and `includeStories: "DefaultCrumb",` can be updated to `includeStories: "Default",`


image

src/components/breadcrumbs/breadcrumbs.style.ts
edleeks87:
```
const StyledBreadcrumbs = styled.nav`
  ol {
    list-style: none;
    display: flex;
    flex-wrap: wrap;
  }
`;
```
 
Is all you need here, the flex and `flex-wrap` on the `nav` don't do anything here

edleeks87 (changes requested):

src/components/breadcrumbs/breadcrumbs-test.stories.tsx
DlgSHi: this is a `TEST` story :)

edleeks87 (approved):

edleeks87:
LGTM, the commits need to be squashed before this is merged

DlgSHi (approved):

edleeks87 (approved):

DlgSHi (approved):

PR #5656, opened 2022-11-29
feat: add color option to progress tracker

blaky:
### Proposed behaviour

This PR adds a `color` prop to the progress tracker component. If the prop is set, it overrides the default color logic that sets the color based on the value. In some cases we would like to be able to set more indicator colors to the progress tracker bar to indicate different statuses, not only two.

### Current behaviour

The indicator bar has two colors that depend exclusively on whether the bar is completed.

### Checklist

#### QA

### Testing instructions

Use the Progress Tracker / Test / default story, there is a new `color` control that controls the value of the `color` prop.


mkrds (approved):

edleeks87:
We need to clarify with DS if this is a feature they are happy for us to support, there is some work in flight updating the `ProgressTracker` component and we need to confirm that they are happy for this to be added on top of that.

@ljemmo @clairedenning can you advise?

clairedenning:
> 

Can you share a demo link or screengrabs please?

edleeks87:
https://codesandbox.io/s/carbon-quickstart-forked-52e5pu?file=/src/index.js

ljemmo:
> > 
> 
> Can you share a demo link or screengrabs please?

I think the proposed update just allows them to pass in whatever colour they require. @edleeks87  is that correct?

For consistency reasons this concerns me but i can understand what a 3rd party might want this. Let me chat with the DS team further and get back to you.

edleeks87:
Correct @ljemmo, thanks we'll hold off until DS have a decision

ljemmo:
@blaky Unfortunately would like to lock this down to the color tokens specced in our figma library to drive consistency across the ecosystem. However, we are open to adding additional color tokens if necessary for other use-cases. Please can you get your designer to reach out to the DS team and discuss them via our drop in sessions or extended team meeting? Alternatively they can reach out directly via designsystem@sage.com

blaky:
Hi @edleeks87 @ljemmo 

I am fine limiting this to design token values, we only use those values anyway but in general
Carbon doesn't restrict color use anywhere, it is not a specific issue with this change. Wherever colors can be supplied by the consuming products, it is not limited to design token values, but any value, I just used the pattern that was already in place in other components. I suggest if you want to limit the set of colors that can be used, it should be implemented as a global feature, rather than in this pull request for this particular component.

blaky:
Hi @edleeks87 @ljemmo

So then the question is, can we go ahead with this change so we can adopt the progress tracker and then plan another global feature that ensures that color prop values are valid design token variables or values?

Thanks

ljemmo:
@blaky would it be possible for your designer to reach out to me directly so we can discuss a solution that might better work for them? I'd prefer for us to bake in the correct color tokens into the component directly rather than provide carbon with the ability to override colors and therefore potentially drive inconsistency.

edleeks87:
Closing in favour of adopting a more consistent global approach tbd by DS team

blaky:
Given that you closed this PR, our implementation now got blocked. Do you have any estimate when this can be done then? If this change or something else that allows us to change the bar color is not made available soon in Carbon, we will need to hack the library and override the styles, which is something we would like to avoid.

blaky:
> @blaky would it be possible for your designer to reach out to me directly so we can discuss a solution that might better work for them? I'd prefer for us to bake in the correct color tokens into the component directly rather than provide carbon with the ability to override colors and therefore potentially drive inconsistency.

Our developers use this component to indicate the state of certain levels. We use "semantic 500" colors.

PR #5657, opened 2022-11-30
feat(select): add virtual scrolling support - FE-5394

robinzigmond:
### Proposed behaviour

Add support for "virtual scrolling" on Select and its variants.

This PR adds 2 new props to all 3 Select components:
- `enableVirtualScroll`: a boolean prop that when set enables virtual scrolling - meaning not all children of the SelectList will be rendered in the DOM at any one time, but instead only those currently visible on the page, plus a few either side
- `virtualScrollOverscan` is the number of children to be rendered into the DOM either side of the visible ones (this defaults to 5). Has no effect when `enableVirtualScroll` is not used.

Also fix some accessibility issues:
- remove some Axe errors for the `multiColumn` mode on Select
- ensure that the guid IDs of select options in the DOM remains stable under any prop change of Select, and only change in the rare case where children are added or removed dynamically. In particular, in a controlled Select, selecting an option does not change the IDs of the children, and so the Axe error about an invalid value for aria-activedescendant is no longer present.

### Current behaviour

Selects always render all their children into the DOM, leading to severe performance issues when there are thousands of children.

There are several ARIA-related errors when Axe is run with the Select List open in a multicolumn select. Also in a controlled select the IDs of children change when selecting an option, leading to an Axe error about ARIA-activedescendant.

### Checklist

#### QA

### Additional context

Note for UX QA - there is a small difference in behaviour after this PR on multiColumn selects. Previously only the table body was scrollable, so the scrollbar does not go up to the sticky table head. In order to make scrolling work correctly in this PR I had to make the whole table scrollable so the scrollbar extends to the table header, even though that stays fixed as you scroll. I don't believe this is an issue and don't think I can avoid it, but please let me know if you think it's a serious problem.

### Testing instructions

Test with the new `virtualScrollBuffer` prop on Selects with large numbers of options (the new "virtualised" story is a simple example), and ensure that it scrolls smoothly and performance is good. (Compare that to performance without the prop to understand why this was needed!)

Perform full regression testing on all 3 types of Select (Simple, Filterable, Multi). Test all stories and all props, including with keyboard navigation through the list.

Also test with screenreaders, in all supported screenreader/browser combinations, to ensure that when the new prop is set the component is still usable and there are no major accessibility issues (or at least not ones that didn't exist before!).

NOTE: when running Axe with the Select List open there is an error "scrollable region must have keyboard access", which did not exist before (except on a multi-column select, where it did). This refers to the select list itself and is a false positive (I checked this with @tempertemper) because it's possible to navigate to the list with the keyboard using the select textbox, as before. I've tried to alter the markup to remove this error but it isn't possible to do this while still getting the desired behaviour. I believe this is OK as it isn't an actual usability barrier and is also only rated "moderate" by Axe, but we should note that the error exists.


src/components/select/simple-select/simple-select.component.js
```diff
@@ -475,6 +477,12 @@ SimpleSelect.propTypes = {
   listPlacement: PropTypes.oneOf(["top", "bottom", "right", "left"]),
   /** Use the opposite list placement if the set placement does not fit */
   flipEnabled: PropTypes.bool,
+  /** Set this prop to enable a virtualised list of options. It corresponds to the number of options to render into
+   * the DOM at once, either side of the currently-visible ones.
+   * Higher values make for smoother scrolling but may impact performance.
+   * If this prop is not used then all options will be in the DOM at all times, which may cause performance problems
+   * on very large lists */
+  virtualScrollBuffer: PropTypes.number,
```
mkrds, line 485: To be honest I think we should have another prop for enabling virtualScroll, like `virtualScrollEnabled`. `virtualScrollBuffer` should stay as an optional prop though.

robinzigmond: Fair point on the naming - I'm not very good at naming myself! I was thinking overscan might tie it too closely to this one library in case we change it later (I'd still like to move to React Aria in the longer term), but if that's a common term across several libraries it's less of an issue.

Happy to debate having a second prop and having a default for the overscan. I'm not really sure what's best for consumers. I do think they should have the ability to control it though, as I have no confidence in being able to pick a sensible default that works well for all usecases and all devices!

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
+      const idArray = [];
+      for (let i = 0; i < children.length; i++) {
+        idArray.push(guid());
+      }
+      childIdsRef.current = idArray;
+    };
+
+    if (
+      childIdsRef.current === null ||
+      childIdsRef.current.length !== React.Children.count(children)
```
mkrds, line 131:
```suggestion
      if (childIdsRef.current?.length !== React.Children.count(children)) {
```
I think this might be sufficient here

src/components/select/select-list/select-list.component.js
```diff
@@ -431,20 +526,28 @@ const SelectList = React.forwardRef(
...
-              id={id}
               as={multiColumn ? "div" : "ul"}
-              aria-labelledby={labelId}
               data-element="select-list"
-              role="listbox"
-              aria-multiselectable={multiselectValues ? true : undefined}
+              {...(multiColumn
+                ? {}
+                : {
+                    role: "listbox",
```
mkrds, line 539: As stated above, I think it's best to extract these props to a separate object and spread them

src/components/select/select-list/select-list.component.js
```diff
@@ -303,7 +359,30 @@ const SelectList = React.forwardRef(
...
+        const scrollOptions = {
+          smoothScroll: false,
+          align: "end",
+        };
+        if (multiColumn && targetIndex > -1) {
+          // Note: virtualizer.measurementsCache isn't actually documented but isn't marked as private either,
+          // and does actually work at least for now. Unfortunately there's no way with the documented public
+          // API to get this - we can't use virtualizer.getVirtualItems() because the target child may not
+          // be currentlly rendered.
+          const { measurementsCache } = virtualizer;
```
mkrds, line 377: I think this is risky, especially considering that you're using `@tanstack/react-virtual` in beta version

src/components/select/select-list/select-list.component.js
```diff
@@ -391,22 +457,51 @@ const SelectList = React.forwardRef(
...
+      );
+    };
 
-    let selectListContent = childrenWithListProps;
+    let selectListContent = renderedChildren;
 
     if (multiColumn) {
       selectListContent = (
-        <StyledSelectListTable>
+        <StyledSelectListTable ref={tableRef}>
```
mkrds, line 490: `tableRef` is also assigned to `StyledSelectListTableBody`

src/components/select/select-list/select-list.component.js
```diff
@@ -391,22 +457,51 @@ const SelectList = React.forwardRef(
...
-          <StyledSelectListTableBody ref={tableRef}>
-            {childrenWithListProps}
+          <StyledSelectListTableBody
+            role="listbox"
+            aria-multiselectable={
+              multiselectValues ? /* istanbul ignore next */ true : undefined
+            }
+            id={id}
+            aria-labelledby={labelId}
+            ref={tableRef}
```
mkrds, line 501: `tableRef` is also assigned to `StyledSelectListTable`

src/components/select/select-list/select-list.component.js
```diff
@@ -66,6 +70,16 @@ const SelectList = React.forwardRef(
     const listActionButtonRef = useRef();
     const { blockScroll, allowScroll } = useScrollBlock();
 
+    const virtualizer = useVirtualizer({
+      count: React.Children.count(children),
+      getScrollElement: () => listContainerRef.current,
```
mkrds, line 75: Wouldn't you be able to avoid overriding the `virtualizer.scrollToIndex` method if you would pass `tableRef` (`StyledSelectListTableBody`) as a scroll element when in `multiColumn` mode?

src/components/select/select-list/select-list.component.js
```diff
@@ -54,6 +57,7 @@ const SelectList = React.forwardRef(
       flipEnabled = true,
       isOpen,
       multiselectValues,
+      virtualScrollBuffer = React.Children.count(children),
```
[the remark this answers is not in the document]
robinzigmond: this was just the simplest way to guarantee that all options are rendered, which I wanted to be the default. (Although arguably it doesn't need to be - I did this when I was concerned virtual scrolling might harm accessibility but Martin seems to think there shouldn't be an issue.)

Bear in mind that the select list is rendered in the DOM (but hidden) even when the select isn't open, so the overscan needs to be the number of children in order to render them all into the DOM in that situation. It may not matter but it seems to work and I'm not sure what your alternative suggestion is (other than just having a fixed default of 10 or something).

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
-    const childIds = useMemo(() => React.Children.map(children, () => guid()), [
+    const childIdsRef = useRef(null);
+
+    // childIds should be stable except when children are added or removed - can't use useMemo
+    // as that isn't absolutely guaranteed to never rerun when dependencies haven't changed.
+    const setChildIds = () => {
+      const idArray = [];
+      for (let i = 0; i < children.length; i++) {
+        idArray.push(guid());
+      }
```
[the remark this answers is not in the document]
robinzigmond: not sure now if there was a good reason or if I just didn't think of this :D I'll get that changed if I don't find any reason not to - thanks

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
+      };
+
+      return React.cloneElement(child, newProps);
+    });
 
-      return lastIndex;
+    const lastOptionIndex = useMemo(() => {
+      // findLastIndex does not exist in Node till v18, and is only in browsers since the start of 2022
+      // so we can't really use it. This should be updated when there is better support!
+      for (let i = childrenList.length - 1; i >= 0; i--) {
```
edleeks87, line 189: There's obviously a few ways of achieving this so not precious if you prefer to leave it as is

```
const lastIndex = (childrenList.length - 1) - childrenList.reverse().findIndex(child => child.type.displayName === Option.displayName || child.type.displayName === OptionRow.displayName)

```

src/components/select/select-list/select-list.style.js
```diff
@@ -91,10 +91,13 @@ const StyledSelectListTableHeader = styled.thead`
 
 const StyledSelectListTableBody = styled.tbody`
   display: block;
-  overflow-y: auto;
+  //overflow-y: auto;
```
edleeks87, line 94: commented code

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
+        const child = childrenList[i];
+        if (child.type === Option || child.type === OptionRow) {
+          return i;
+        }
+      }
+      return -1;
```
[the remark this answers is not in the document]
edleeks87: I'd go with this if you refactor as it's more readable than what I have above

src/components/select/select-list/select-list.component.js
edleeks87: Inclined to agree with Maciek especially if we use a `beta` version

src/components/select/select-list/select-list.component.js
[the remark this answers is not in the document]
robinzigmond: in the end I've adopted a hybrid of the 2 approaches. The linter doesn't like using `--` but it seems there's an exception for the header of a `for` loop. So I've kept this as a `for` loop and mostly similar to what I had, but I have adopted your suggestion of having a single return and just `break`ing the loop rather than having an early return when the correct child is found.

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
       },
       [onSelect]
     );
 
-    const childIds = useMemo(() => React.Children.map(children, () => guid()), [
+    const childIdsRef = useRef(null);
+
+    // childIds should be stable except when children are added or removed - can't use useMemo
+    // as that isn't absolutely guaranteed to never rerun when dependencies haven't changed.
+    const setChildIds = () => {
```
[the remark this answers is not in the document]
robinzigmond: yes, I realised it did because we don't want to replace the IDs except when the children change - useMemo doesn't guarantee this (even though it likely works in practice), it's supposed to be a performance optimisation, not a guarantee of memoisation for functionality purposes which is what we were using it for here.

src/components/select/select-list/select-list.component.js
```diff
@@ -93,63 +107,92 @@ const SelectList = React.forwardRef(
...
       children,
     ]);
 
-    const childrenWithListProps = useMemo(
-      () =>
-        React.Children.map(children, (child, index) => {
-          if (!child || (child.type !== Option && child.type !== OptionRow)) {
-            return child;
-          }
+    const optionChildrenList = useMemo(
```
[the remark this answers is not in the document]
robinzigmond: there's certainly a string argument for using `displayName` instead of `type` but in the absence of either a specific bug raised for this on Select or a general refactor across all components I'd rather leave this as it was in this PR.

Which yes means including `OptionRow`, I'm not sure how I missed that but Maciek already caught it (above).

src/components/select/select-list/select-list.component.js
```diff
@@ -66,6 +71,45 @@ const SelectList = React.forwardRef(
     const listActionButtonRef = useRef();
     const { blockScroll, allowScroll } = useScrollBlock();
 
+    // small adjustment of react-virtual's default observeElementRect in order to account for the table header in multicolumn mode
+    // (the only change from the default is subtracting the HEIGHT_ADJUSTMENT from the rect's height)
+    /* istanbul ignore next */
```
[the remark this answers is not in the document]
robinzigmond: Yes I wasn't comfortable ignoring the whole thing either - but then I decided it was just basically testing the third-party package. On the other hand, I've had to override this for our purposes so we really should test that. I don't see how I can do a sensible unit test for this but I'll see if I can add a cypress test to ensure that when navigating with the keyboard the selected option is always visible (because that's the issue I was fixing)

src/components/select/simple-select/simple-select.component.js
mkrds: I think it might be confusing for the users that they have to figure out what `overscan` value they have to provide, that's why I think we should provide a default value for that.

I'd name this prop `virtualScrollOverscan` and default it to 3 elements, but it seems like it will break the current implementation where you default the `virtualScrollBuffer` to `React.Children.count(children)`

src/components/select/select-list/select-list.component.js
```diff
@@ -93,62 +115,89 @@ const SelectList = React.forwardRef(
...
+        "aria-posinset": isOption ? optionChildIndex + 1 : undefined,
+        // needed to dynamically compute the size
+        ref: measureCallback,
+        "data-index": index,
+      };
 
+      return React.cloneElement(child, newProps);
+    });
+
+    const lastOptionIndex = useMemo(() => {
```
edleeks87, line 190: Might be worth using the function provided by `lodash`. It's simple enough that I'm happy to leave this as is if you'd prefer

src/components/select/select-list/select-list.component.js
```diff
@@ -351,24 +401,37 @@ const SelectList = React.forwardRef(
...
+          align: "start",
+        });
       }
-    }, [children, currentOptionsListIndex, isLoading, lastOptionIndex]);
+    }, [
+      children,
+      currentOptionsListIndex,
+      isLoading,
+      lastOptionIndex,
+      listContainerRef,
```
edleeks87, line 432: Same question as above

src/components/select/select-list/select-list.component.js
```diff
@@ -351,24 +401,37 @@ const SelectList = React.forwardRef(
...
+      handleListScroll({ target: listContainerRef.current });
     }, [
-      childrenList,
       getIndexOfMatch,
       highlightedValue,
-      multiColumn,
-      optionRefList,
+      virtualizer,
+      handleListScroll,
+      listContainerRef,
```
edleeks87, line 413: Apologies if I missing something, but does having the ref in the dependency array do anything here, I thought mutating refs do not trigger re-renders etc?

robinzigmond: This is more just to keep the linter happy - I know missing dependencies are warnings rather than errors but I still prefer to avoid them.

Although there is a potential need - this is a forwarded ref rather than one created using `useRef`, it's possible that components that consume this pass a `ref` that's conditional and can change from one object to a completely different one, or something weird like that. It likely doesn't matter in practice but I think it's best to keep it in the dependencies. (It's always safer in React to have too many dependencies than not enough - a missing dependency is a bug waiting to happen, while an unnecessary one ought to only cause performance degradations at worst.)

src/components/select/select-list/select-list.component.js
robinzigmond: same answer as below ;)

edleeks87 (approved):

src/components/select/select-list/select-list.component.js
edleeks87: No worries thanks for clarifying 👍

mkrds (dismissed):

mkrds (dismissed):

edleeks87 (dismissed):

divyajindel (approved):

mkrds (approved):

edleeks87 (approved):

divyajindel (approved):

PR #5661, opened 2022-11-30
feat(multi-action-button): surface width prop to allow width to be set for component

### LINKED ISSUE #5547: Allow to set the label and icon on opposite sides in fullWidth button
### Desired behaviour

I would like to be able to set an icon and label in a button with "fullWidth" enabled on opposite sides.

image

### Current behaviour

Currently it is not possible to set the icon and label on opposite sides in button.
If we set fullWidth to true, the button with the icon looks like this:

image

### Suggested Solution

Just add a new property that will set margin-left / margin-right (depending on the position of the icon) icon to "auto". This is just one of the possible solutions.

`
 span[data-component="icon"] {
  margin-left:  'auto'
}
`

`
              <Button
                buttonType="primary"
                iconType="caret_down"
                iconPosition="after"
                fullWidth={isSmallScreen}
                spaceBetween={true} // <== new property to set it
               >
                text
              </Button>
`

_No response_

_No response_

### Confidentiality

### LINKED ISSUE #5672: Make `MultiActionButton` accept `fullWidth` prop
### Desired behaviour

I would love the `MultiActionButton` to accept the `fullWidth` prop similarly to how a regular `Button` does, so it can span the whole available space eg on mobile views.

### Current behaviour

Currently the `MultiActionButton` itself does not accept the `fullWidth` prop and it is only available on the descendant `Button`s which roll down as options. 

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

tomdavies73:
fix #5547, fix #5672

### Proposed behaviour

Adds width prop to component to allow consumers to set a custom width. Also changes justify-content: space-between to default styling behaviour of `Button` components when the prop is passed.

image

### Current behaviour

No support for setting a custom width. Also, without change to default styling, Text and Icons within the `Button` component remain centered when a wider width is used.

image

### Checklist

#### QA

### Testing instructions

Refer to customWidth story


mkrds (dismissed):

robinzigmond (approved):
The changes look good - but it might be worth checking with Luke/Harpal if we definitely want to completely change the styling for all uses of the component. I see the ticket is due for  UX QA so hopefully Harpal can answer then - so I'll approve this.

The reason I have doubts is the original Github issue only mentions adding a prop to enable this new styling. But if the DS team want to change it for all cases that's fine too (but may cause complaints from teams using this component in other products).

src/components/multi-action-button/multi-action-button.stories.mdx
```diff
@@ -47,6 +47,14 @@ import MultiActionButton from "carbon-react/lib/components/multi-action-button";
   <Story name="disabled" story={stories.Disabled} />
 </Canvas>
 
+### With customWidth
+
+Pass any `px` or `%` value to achieve the desired width. `Sizes` prop can still be used to set desired height.
```
robinzigmond, line 52:
```suggestion
Pass any `px` or `%` value to achieve the desired width. `size` prop can still be used to set desired height.
```

Let's name the prop correctly so consumers don't get confused

robinzigmond (approved):

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -672,4 +672,30 @@ describe("MultiActionButton", () => {
       expect(tempWrapper.contains("Some string")).toBe(true);
     });
   });
+
+  describe("when customWidth is passed", () => {
+    it("justify-content on StyledButton set to space-between", () => {
+      wrapper = render({ customWidth: "100%" }, mount);
```
robinzigmond, line 678: we should test the `width` prop now there is no `customWidth` prop - this probably won't even pass

src/components/multi-action-button/multi-action-button.spec.tsx
```diff
@@ -672,4 +672,30 @@ describe("MultiActionButton", () => {
...
+        {
+          justifyContent: "space-between",
+        },
+        wrapper.find(StyledMultiActionButton),
+        { modifier: `${StyledButton}` }
+      );
+    });
+
+    it("width on StyledButton set to 100%", () => {
+      wrapper = render({ customWidth: "100%" }, mount);
```
robinzigmond, line 690: same here

src/components/multi-action-button/multi-action-button.style.ts
```diff
@@ -14,6 +16,17 @@ const StyledMultiActionButton = styled.div<StyledMultiActionButtonProps>`
   display: inline-block;
   position: relative;
 
+  ${({ width }) =>
+    width &&
+    css`
+    ${width && computeWidth({ width })}
```
robinzigmond, line 22:
```suggestion
    ${computeWidth({ width })}
```

you don't need to check `width` as it's already guarded outside this (line 20)

src/components/multi-action-button/multi-action-button.test.js
```diff
@@ -181,290 +181,308 @@ context("Tests for MultiActionButton component", () => {
...
 
-  describe("pressing ArrowUp while MultiActionButton is open", () => {
-    it("should move focus to previous child button and should not loop to last button when first is focused", () => {
-      CypressMountWithProviders(
-        <>
-          <MultiActionButtonList />
-          <MultiActionButtonList />
-        </>
+    it("when customWidth is passed, should render Button with justify-content: space-between", () => {
+      CypressMountWithProviders(<MultiActionButtonList customWidth="70%" />);
```
robinzigmond, line 186: this test also needs `width` instead of `customWidth` (in the test description as well as the actual prop)

src/components/multi-action-button/multi-action-button.test.js
```diff
@@ -181,290 +181,308 @@ context("Tests for MultiActionButton component", () => {
...
-            .eq(0)
-            .trigger("keydown", keyCode("uparrow"));
-          multiActionButtonList().eq(0).should("be.focused");
-          multiActionButtonListContainer()
-            .eq(0)
-            .trigger("keydown", keyCode("uparrow"));
-          multiActionButtonList().eq(0).should("be.focused");
-        });
+    it("when customWidth is passed, should render Button parent with specified customWidth (70% / 945px) ", () => {
+      CypressMountWithProviders(<MultiActionButtonList customWidth="70%" />);
```
robinzigmond, line 196: and this one

robinzigmond (dismissed):

mkrds (dismissed):

src/components/multi-action-button/multi-action-button.test.js
```diff
@@ -181,290 +181,308 @@ context("Tests for MultiActionButton component", () => {
...
+        multiActionButtonList().eq(0).focus();
+        multiActionButtonListContainer()
+          .eq(0)
+          .trigger("keydown", keyCode("End"));
+        multiActionButtonList().eq(2).should("be.focused");
+      });
+  });
+});
+
+describe("Pressing esc while MultiActionButton is open", () => {
```
ZhuoyuJin, line 441: describe("pressing esc while MultiActionButton is open", () => {, change P to p to keep the consistency of all tests please

ZhuoyuJin (approved):

mkrds (approved):

PR #5665, opened 2022-12-01
feat(icon-button): add padding interface to component FE-5567

### LINKED ISSUE #5640: Allowing padding overriding for MenuItem and IconButton
### Desired behaviour

Allow the padding for MenuItem and IconButton components to be overridden and set to 0, if desired, as can be done with Button

### Current behaviour

These components apply a default padding of 16, which may be undesirable

### Suggested Solution

_No response_

_No response_

https://codesandbox.io/s/youthful-darwin-qtjnpk?file=/src/index.js
### Confidentiality

edleeks87:
fix #5640

### Proposed behaviour

The `IconButton` now supports the `padding` interface to allow consuming projects to override the
default padding values.

No longer sets the padding styling on the internal anchor or button elements rendered internally by
the `MenuItem` if both `onClick` and `href` props are not passed

image

### Current behaviour

No support for padding interface in `IconButton`
Default padding is applied to anchor/ button elements when no `onClick`/`href` passed

image

### Checklist

#### QA

### Testing instructions


src/components/menu/menu-item/menu-item.style.js
```diff
@@ -85,11 +87,27 @@ const StyledMenuItemWrapper = styled.a`
...
+        a,
+        ${StyledLink} a,
+        button,
+        ${StyledLink} button {
+          padding: 0 16px;
+        }
+      `
+    }
+
+    ${StyledIconButton} {
```
robinzigmond, line 102: I'm not sure about these styles. Are we sure a StyledIconButton will never be used inside a MenuItem in other situations? At the very least I'd check the `asPassiveItem` here as well so this only applies to the situation we're already considering.

Also seeing this made me check the focus styles on the updated code sandbox and while the lack of outline is probably OK, the icon goes dark/invisible when its focused so that will want fixing...

edleeks87: Fair point I'll add that as an extra guard, and good spot I've missed the colour override on the Icon 👍

Parsium (commented):
Happy to approve once the issue @robinzigmond is fixed 👍🏼 I also noticed the behaviour in the codesandbox where the icon goes dark when the icon button is in focus.

robinzigmond (approved):

Parsium (dismissed):

Parsium (approved):

robinzigmond (approved):

PR #5678, opened 2022-12-06
fix(tab): ensure that validation failures are no longer reported when children unmount FE-5527

### LINKED ISSUE #5591: Form Inputs don't notify parent Tab if suddenly unmounted.
### Current behaviour

Conditionally rendered form components can cause issues with tabs and how they highlight errors.
So form fields will (via context) inform their parent tab if they have an error or not so it will highlight in red.
However, if that field is suddenly removed from the screen (because of conditional rendering), the tab is never notified that there's no longer an error to warn about.

### Expected behaviour

I would expect the field to notify the tab that and clear the error as validation no longer relevant.

https://codesandbox.io/s/tabs-dialogs-i1oo0l?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

I suspect you just need to add a cleanup function to the useEffect here:
https://github.com/Sage/carbon/blob/3238fb0cf79dab4ecd2d7364abf4cf8560055218/src/__internal__/form-field/form-field.component.tsx#L132
So that if a field is unmounted for any reason, it updates the error/warning/info status of the tab one final time?

### Carbon Version

110.0.3

### Design Tokens Version

_No response_

Chrome

Windows

_No response_

### Confidentiality

### LINKED ISSUE #5618: Enhance Tabs so that they can summarise form input error on their page
### Desired behaviour

We had an ask to have the Tab component summarise validation messages from its children like you can see in the screenshot.

image

### Current behaviour

The current behaviour just allows you to provide a string that's independent of anything going on in the tab. 

### Suggested Solution

Since the Tabs component is already keeping a list of errors (as bools), simplest solution seems to be to change that to strings and then allow the user to pass a function instead of a string to format the list.

I've forked carbon and produced a working example as a POC (hacked around TS as not familiar with it) but it shows in principle, how this could be achieved in a fairly straight forward way.

https://github.com/Sage/carbon/commit/9776bf690ea62f00f09a4683492464e99864af35 

_No response_

Discussed with Carbon team members on Slack channel who suggested raising this here.

### Confidentiality

edleeks87:
fix #5591
fix #5618

### Proposed behaviour

Ensures that any validation failures reported to parent Tab components are removed when the
associated child unmounts.

tabs validation fix

Adds `showValidationsSummary` prop to `Tabs` to support the option of rendering a summary of any
validation string failures in the `ValidationIcon` `Tooltip`.

image

### Current behaviour

Validation failures are not removed when an input within a `Tab` unmounts
No support for rendering a summary of all validation string failures in the `ValidationIcon` `Tooltip`

### Checklist

#### QA

### Testing instructions

Use linked sandbox below for testing validation summary (linked issue has different implementation) and sandbox included in #5591

https://codesandbox.io/s/nostalgic-dawn-ubkcu0?file=/src/index.js

### DIFF
--- a/src/__internal__/form-field/form-field.component.tsx
+++ b/src/__internal__/form-field/form-field.component.tsx
@@ -143,9 +143,9 @@ const FormField = ({
   }, []);
 
   useEffect(() => {
-    if (setError) setError(id, !!error);
-    if (setWarning) setWarning(id, !!warning);
-    if (setInfo) setInfo(id, !!info);
+    if (setError) setError(id, error);
+    if (setWarning) setWarning(id, warning);
+    if (setInfo) setInfo(id, info);
 
     return () => {
       if (!isMounted.current) {
--- a/src/__internal__/form-field/form-field.component.tsx
+++ b/src/__internal__/form-field/form-field.component.tsx
@@ -1,4 +1,10 @@
-import React, { useContext, useEffect, useMemo } from "react";
+import React, {
+  useContext,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+} from "react";
 
 import { ValidationProps } from "__internal__/validations";
 import { MarginProps } from "styled-system";
@@ -126,10 +132,28 @@ const FormField = ({
     TabContext
   );
 
+  const isMounted = useRef(false);
+
+  useLayoutEffect(() => {
+    isMounted.current = true;
+
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
   useEffect(() => {
     if (setError) setError(id, !!error);
     if (setWarning) setWarning(id, !!warning);
     if (setInfo) setInfo(id, !!info);
+
+    return () => {
+      if (!isMounted.current) {
+        if (setError && error) setError(id, false);
+        if (setWarning && warning) setWarning(id, false);
+        if (setInfo && info) setInfo(id, false);
+      }
+    };
   }, [id, setError, setWarning, setInfo, error, warning, info]);
 
   const marginProps = filterStyledSystemMarginProps(rest);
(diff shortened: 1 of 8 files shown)


src/__internal__/form-field/form-field.component.tsx
```diff
@@ -126,10 +132,28 @@ const FormField = ({
     TabContext
   );
 
+  const isMounted = useRef(false);
```
edleeks87, line 135: This flag is needed as without it the cleanup function runs every time one of the dependencies changes and causes some side effects in other stories

src/components/tabs/tabs.spec.js
```diff
@@ -931,4 +933,171 @@ describe("Tabs", () => {
...
+    it("any info in a child component is correctly reported in the TabTitle", () => {
+      const wrapper = mount(<MockComponent info />);
+
+      expect(wrapper.find(ValidationIcon).exists()).toBe(true);
+
+      wrapper.setProps({ show: false });
+      wrapper.update();
+      expect(wrapper.update().find(ValidationIcon).exists()).toBe(false);
+    });
+  });
```
nineteen88, line 989: Minor - Could also be refactored into one it.each()

nineteen88 (approved):
Approved with the minor refactor comment open as it's up to you if you want to make that change

src/components/tabs/tabs.spec.js
edleeks87: Yeah good point, will update if I get any further comments

grabkowski (dismissed):

nineteen88 (dismissed):

grabkowski (approved):

DlgSHi (approved):

grabkowski (approved):

PR #5694, opened 2022-12-16
refactor(number): convert files to typescript (FE-4813)

grabkowski:
### Proposed behaviour

### Current behaviour

### Checklist

#### QA

### Testing instructions


src/components/number/number.spec.js
```diff
@@ -6,8 +6,22 @@ import Label from "../../__internal__/label";
 import { assertStyleMatch } from "../../__spec_helper__/test-utils";
 import InputPresentation from "../../__internal__/input/input-presentation.component";
 
+function renderNumberInput(props, renderer = mount) {
```
edleeks87: We don't need the second arg here as mount is the only renderer used

src/components/number/number.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Textbox
+      {...rest}
+      value={value}
+      onChange={handleOnChange}
+      onKeyDown={handleKeyDown}
+    />
+  );
+};
+
+Number.displayName = "Number";
```
robinzigmond, line 55: do we really have to explicitly set the displayName here? I thought React would set that automatically, unless we were using forwardRef or something (which we're not, at least for now).

It won't do any harm (and may save us some work/bugs in future), I'm just surprised it's been added as part of a TS refactor.

src/components/number/number.spec.js
grabkowski: right, removing it now

src/components/number/number.component.tsx
grabkowski: I'm pretty sure that the `Number` component should be wrapped in a `forwardRef`. I'll remove it.

robinzigmond (approved):

nineteen88 (dismissed):

divyajindel (approved):

robinzigmond (approved):

nineteen88 (approved):

nineteen88 (approved):

robinzigmond (approved):

divyajindel (approved):

PR #5711, opened 2022-12-29
docs(): remove args for specialCharacters

DlgSHi:
### Proposed behaviour

After the refactoring, we need to clear up and remove special characters args from Test stories

### Current behaviour

When we were using the `URL` passing approach we were passing `special characters` via URL. Then we implemented util helper. 

### Checklist

#### QA
- [ ] Open Storybook and check all `Test` stories if they are working as previously.


mkrds (dismissed):

src/components/portrait/portrait-test.stories.tsx
```diff
@@ -34,29 +32,21 @@ export default {
...
 } as ComponentMeta<typeof Portrait>;
 
 export const Default = ({
   alt,
-  altSpecialCharacters,
   ...args
 }: {
   alt?: string;
-  altSpecialCharacters?: string;
 } & PortraitProps) => (
```
Parsium, line 43:
```suggestion
}: PortraitProps) => (
```
The `alt` prop is now part of the `PortraitProps` interface, so we can simplify this like so

Parsium (approved):

mkrds (approved):

ZhuoyuJin (approved):

PR #5721, opened 2023-01-03
fix(progress-tracker): add deprecation warnings for orientation and direction props

edleeks87:
### Proposed behaviour

Adds deprecation warning for `orientation` and `direction` props

### Current behaviour

No deprecation warning for `orientation` and `direction` props

### Checklist

#### QA

### Testing instructions

Confirm only one warning displays in storybook (docs tab)


src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -11,6 +11,9 @@ import {
   StyledValue,
 } from "./progress-tracker.style";
 import useResizeObserver from "../../hooks/__internal__/useResizeObserver";
+import Logger from "../../__internal__/utils/logger";
+
+let deprecatedForwardRefWarnTriggered = false;
```
robinzigmond, line 16: can we change the name of this variable? The name should refer to `orientation` and/or `direction`, not `forwardRef` which is irrelevant here.

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -34,6 +37,13 @@ const ProgressTracker = ({
   labelsPosition,
   ...rest
 }) => {
+  if (!deprecatedForwardRefWarnTriggered) {
```
robinzigmond, line 40: this will log the deprecation warning whenever the component is used. Surely we should check for whether the prop(s) are actually used and only show the warning then?

(My preference would be to have 2 separate variables and 2 separate deprecation messages, one for each prop. But I suppose it's OK if the message is logged when either of them are used, that should be understandable enough.)

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -34,6 +37,13 @@ const ProgressTracker = ({
...
+    deprecatedForwardRefWarnTriggered = true;
+    Logger.deprecate(
+      "The `orientation` and `direction` props in `ProgressTracker` component are deprecated and will soon be removed."
```
robinzigmond, line 43: I'm not sure of the background here as to why these are being deprecated, but if there's an alternative prop that can do the same or any other way to achieve it, we should explain how to do that. (I'm not sure that there is but it's going to be worse than a usual breaking change if we're literally removing the ability to achieve particular functionality.)

robinzigmond (changes requested):

src/components/progress-tracker/progress-tracker.component.js
edleeks87: The props have default values so will always display etc, I explained why I've chosen not to add tests or refactor further in the context section above

src/components/progress-tracker/progress-tracker.component.js
edleeks87: The DS have decided we are no longer supporting this functionality

src/components/progress-tracker/progress-tracker.component.js
edleeks87: Yeah sorry copy paste 🤦

src/components/progress-tracker/progress-tracker.component.js
robinzigmond: Yeah I still disagree, I don't think we should tell consumers "you shouldn't use this prop" when they haven't actually used it. That refactor will be needed later anyway when we come to remove the prop so why not do it now.

By all means seek a second opinion if you feel strongly enough about it.

src/components/progress-tracker/progress-tracker.spec.js
```diff
@@ -55,6 +64,23 @@ describe("ProgressBar", () => {
 
   testStyledSystemMargin((props) => <ProgressBar {...props} />);
 
+  describe("deprectaion warning", () => {
```
robinzigmond, line 67:
```suggestion
  describe("deprecation warning", () => {
```

robinzigmond (approved):

nineteen88 (approved):

src/components/progress-tracker/progress-tracker.component.js
```diff
@@ -29,29 +32,39 @@ const ProgressTracker = ({
   showDefaultLabels = false,
   currentProgressLabel,
   maxProgressLabel,
-  orientation = "horizontal",
-  direction = "up",
+  orientation,
+  direction,
```
nineteen88, line 36: Actually, just to check - removing the default here isn't going to break anything for anyone, right? The default will still be these as far as I can tell from the rest of the code and this is purely to prevent the deprecation warning from triggering on every use of the component? Just wanted to double check.

nineteen88 (approved):

src/components/progress-tracker/progress-tracker.component.js
edleeks87: Yeah that's correct

nineteen88: Perfect

DlgSHi (approved):

PR #5726, opened 2023-01-04
Fe 5505 add forwardref

### LINKED ISSUE #5564: All Form input fields should implement forwardRef and give us access to internal input element.
### Desired behaviour

This would be very convenient for legitimate cases where one must interact with the DOM directly, such as focusing or selecting text in a form input (also important for Sage 200, due to third party integration requirements). 

Before switching to function components, this was loosely supported. Class components will always return something to a ref so we could use that the then find the DOM input (see sandbox). It would be nice if all input elements were forwarded refs as a matter of course.

From React docs:

>When to Use Refs
>
> There are a few good use cases for refs:
> 
> Managing focus, text selection, or media playback.
> Triggering imperative animations.
> Integrating with third-party DOM libraries.

### Current behaviour

Currently, there's no implementation of forwardRef in the components so any ref passed will always be undefined. 

### Suggested Solution

wrap every form input in forwardRef and pass the ref down to the form input. 

https://codesandbox.io/s/can-can-focus-on-old-versions-9rn9mb?file=/src/index.js

Here's a different sandbox, showing what I'm trying to do. Other sandbox contains example of how we used to achieve this when things were class based (using the findDomNode escape hatch which is now discouraged).
https://codesandbox.io/s/kind-browser-ps90bb?file=/src/index.js

### Confidentiality

robinzigmond:
### Proposed behaviour

All components which wrap a single input should use `React.forwardRef` to allow consumers to pass a `ref={someRef}` to the component and have that be attached to the underlying DOM element. (It should work with any React ref - so either an object with a `current` property or a function which accepts an HTML element.)

The `inputRef` prop which exists on most such elements and which offers similar functionality should be deprecated as it is no longer needed when `ref` works.

Components such as DateRange and NumeralDate which consist of more than one input should offer props to allow references to be taken to each individual input element as needed.

Fix #5564 

### Current behaviour

Most such components do not allow a `ref` to be attached. In most cases this can be done with the `inputRef` function prop but this has a non-standard shape and isn't obvious (most React developers know how to use `ref` and will expect it to work, whereas knowing about `inputRef` requires a careful reading of the docs - in most cases it only appears once in a big props table towards the bottom of the page).

### Checklist

#### QA

### Additional context

Note that I have not removed the `inputRef` prop from the internal `Input` component as the way this is currently implemented, a non-memoised callback can be passed and will not get called with `null` until the component unmounts - whereas if this is replaced with a "genuine" callback ref, a function that changes every render causes the ref to become "detached" and be called with null before being called again with the DOM element. While not a hard issue to fix, this would count as a breaking change - so I have left this as is, and can be removed later when we finally remove the `inputRef` prop from all components.

### Testing instructions

Careful regression tests are needed on all the following components which have been changed.

Checkbox
Date
DateRange
Decimal
FilterableSelect
GroupedCharacter
MultiSelect
Number
NumeralDate
RadioButton
Search
Select
Switch
Textarea
Textbox
TileSelect

Check that the new props work as intended on DateRange (startRef and endRef) and NumeralDate (dayRef, monthRef, yearRef), and that a plain `ref` works on all the others.


edleeks87 (dismissed):

mkrds:
I think this PR should trigger a major version release considering that removal of the `inputRef` prop is a breaking change.

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -383,8 +415,15 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
                       onChange={(e) =>
                         handleChange(e, datePart as keyof NumeralDateObject)
                       }
-                      inputRef={(ref) => {
-                        refs.current[index] = ref;
+                      ref={(element) => {
+                        refs.current[index] = { current: element };
+                        if (inputRef) {
```
mkrds, line 420: Minor comment, nesting could be avoiding by doing `if (!inputRef) return;`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -383,8 +415,15 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
                       onChange={(e) =>
                         handleChange(e, datePart as keyof NumeralDateObject)
                       }
-                      inputRef={(ref) => {
-                        refs.current[index] = ref;
+                      ref={(element) => {
+                        refs.current[index] = { current: element };
```
mkrds, line 419:
```suggestion
                        refs.current[index] = ref;
```

This change would require following change on line 308

`(ref) => ref === document.activeElement`

src/components/textarea/textarea.component.tsx
```diff
@@ -116,227 +117,266 @@ export interface TextareaProps
...
+      tooltipPosition,
+      value,
+      "data-component": dataComponent,
+      "data-element": dataElement,
+      "data-role": dataRole,
+      helpAriaLabel,
+      inputRef,
+      ...props
+    }: TextareaProps,
+    ref
```
mkrds, line 164:
```suggestion
    ref: React.ForwardedRef<HTMLInputElement>
```

src/components/textbox/textbox.component.tsx
```diff
@@ -123,216 +124,233 @@ export interface TextboxProps extends CommonTextboxProps {
...
+            ) : (
+              input
+            )}
+          </FormField>
+          {characterCount}
+        </InputBehaviour>
+      </TooltipProvider>
+    );
+  }
+);
```
mkrds, line 354:
```suggestion
);

Textbox.displayName = "Textbox";
```

src/components/textarea/textarea.component.tsx
```diff
@@ -116,227 +117,266 @@ export interface TextareaProps
...
+  ) => {
+    const { validationRedesignOptIn } = useContext(NewValidationContext);
+    const computeLabelPropValues = <T,>(prop: T): undefined | T =>
+      validationRedesignOptIn ? undefined : prop;
 
-  const inputRef = useRef<HTMLInputElement>(null);
+    const { current: id } = useRef(idProp || guid());
 
-  const minHeight = useRef(MIN_HEIGHT);
+    const internalRef = useRef<HTMLTextAreaElement>(
```
mkrds, line 172:
```suggestion
    const internalRef = useRef<HTMLTextAreaElement | null>(
```

src/components/textarea/textarea.component.tsx
```diff
@@ -116,227 +117,266 @@ export interface TextareaProps
...
+      null
+    ) as React.MutableRefObject<HTMLTextAreaElement>;
```
mkrds, line 174:
```suggestion
    ) ;
```

src/components/textarea/textarea.component.tsx
```diff
@@ -116,227 +117,266 @@ export interface TextareaProps
...
 
-  const expandTextarea = () => {
-    const textarea = inputRef.current;
+    const callbackRef = useCallback(
+      (inputElement) => {
+        internalRef.current = inputElement;
+        if (ref) {
```
mkrds, line 179: Minor comment, reduce nesting by `if (!ref) return;`

src/components/date/date.component.js
```diff
@@ -24,350 +31,384 @@ import Textbox from "../textbox";
...
+      onKeyDown,
+      pickerProps = {},
+      readOnly,
+      size = "medium",
+      tooltipPosition,
+      value,
+      inputRef,
+      ...rest
+    },
+    ref
```
mkrds, line 70:
```suggestion
    ref: React.ForwardedRef<HTMLInputElement>;
```

mkrds:
I also think it might be beneficial to add `TODO:` comments in places where some code can be removed right after dropping the support for `inputRef` props

src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -62,119 +62,123 @@ export interface CheckableInputProps extends CommonCheckableInputProps {
...
+            <StyledCheckableInput>
+              <HiddenCheckableInput {...inputProps} />
+              {children}
+            </StyledCheckableInput>
+          </FormField>
+        </InputBehaviour>
+      </StyledCheckableInputWrapper>
+    );
+  }
+);
```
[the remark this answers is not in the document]
robinzigmond: yeah, I noticed sometimes it was causing TS errors, but not others. I'm not sure why it would cause a TS error actually, although it's useful when it does! I meant to go back at the end and make sure I added `displayName` to all the components I'd added `forwardRef` to but I think I forgot some - I'll make sure I fix that!

src/components/checkbox/checkbox.component.tsx
```diff
@@ -31,108 +32,126 @@ export interface CheckboxProps extends CommonCheckableInputProps, MarginProps {
...
+      size,
+      tooltipPosition,
+      "data-component": dataComponent = "checkbox",
+      "data-element": dataElement,
+      "data-role": dataRole,
+      helpAriaLabel,
+      inputRef,
+      ...props
+    }: CheckboxProps,
+    ref: React.Ref<HTMLInputElement>
```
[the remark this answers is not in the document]
robinzigmond: and having looked it up it (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts) it seems that `React.Ref` doesn't allow mutable ref objects, whereas `ForwardedRef` does, so `ForwardedRef` is the correct one to ensure consumers don't get errors when trying to update their refs. I'll change all of them to use that type.

src/components/select/simple-select/simple-select.component.js
```diff
@@ -353,21 +364,28 @@ const SimpleSelect = React.forwardRef(
...
+    );
 
     function getTextboxProps() {
       return {
         id,
         name,
         disabled,
         readOnly,
-        inputRef: assignInput,
+        ref: inputRef ? undefined : assignInput,
```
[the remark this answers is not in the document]
robinzigmond: good point, I'd overlooked that. Surprised all tests were passing, I'll take a look at this.

src/components/numeral-date/numeral-date.component.tsx
robinzigmond: I don't think I understand what you're suggesting here. `ref` does not actually exist in scope here after my changes. The whole point of this change was to replace `inputRef`, an internal callback prop which accepts a ref object (whose "current" property is a DOM element, or null), with a standard React callback ref, which takes the DOM element itself (or null) as argument. And because of how `refs` is used, each element of the `refs.current` array has to be a "ref object", hence setting it to `{ current: element }`, where before it was `ref` - they're the same shape, but I now have to build the object myself rather than just getting it from the argument.

I'm not sure what you're suggesting (given that the literal change you suggest would cause an error) but hopefully the above gives more context so you can explain it in a way I can understand :)

src/components/textarea/textarea.component.tsx
robinzigmond: strangely, TS doesn't seem to need this, even though when I hover over it it says React.ForwardedRef<unknown> (but the way it's used implies it's an `HTMLInputElement` inside, as it's passed as `ref` to `Input` which has that typed correctly). Not sure what's going on but I agree it's better to be explicit if only so it's easier for us to see in future what the intended type is.

robinzigmond:
> I also think it might be beneficial to add `TODO:` comments in places where some code can be removed right after dropping the support for `inputRef` props

I'm not sure if there's much benefit in this - surely when we come to remove the prop any code that refers to it will automatically be flagged as errors. I guess maybe there's stuff in the internal `Input` component that won't be obvious it can be removed, so I'll take a look at that again and add TODO comments if I feel it's useful.

src/components/date/date.component.js
robinzigmond: this is still a `.js` file so I can't add a type annotation 😄

src/components/select/simple-select/simple-select.component.js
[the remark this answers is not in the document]
robinzigmond: I can confirm I managed to break it - when adding the prop to the default story, the positioning of the select list is completely incorrect. I'll fix this and also add tests for this behaviour so this regression would get caught.

src/components/date/date.component.js
mkrds: Lol, yeah I've missed that :)

src/components/search/search.component.tsx
```diff
@@ -67,218 +68,238 @@ export interface SearchProps extends ValidationProps, MarginProps {
...
+          inputIcon={iconType}
+          iconTabIndex={iconTabIndex}
+          iconOnClick={handleIconClick}
+          iconOnMouseDown={handleMouseDown}
+          aria-label={ariaLabel}
+          onFocus={handleFocus}
+          onBlur={handleBlur}
+          onChange={handleChange}
+          onKeyDown={handleKeyDown}
+          ref={ref || assignInput}
```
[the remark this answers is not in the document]
edleeks87: We'll likely need some kind of callback here once #5769 is merged as I had to replace the cloning with a context approach

src/components/search/search.component.tsx
```diff
@@ -67,221 +68,229 @@ export interface SearchProps extends ValidationProps, MarginProps {
...
+          inputIcon={iconType}
+          iconTabIndex={iconTabIndex}
+          iconOnClick={handleIconClick}
+          iconOnMouseDown={handleMouseDown}
+          aria-label={ariaLabel}
+          onFocus={handleFocus}
+          onBlur={handleBlur}
+          onChange={handleChange}
+          onKeyDown={handleKeyDown}
+          ref={ref || inputRef}
```
edleeks87, line 271: Worth taking a look at #5769 if you get a chance as I've had to make some changes to how the MenuItem passes a ref to Search using context. I don't think it will create many issues but we'll likely need to use a callback so the ref is registered correctly to the item. The linked PR is obviously in review so may change so not suggesting any updates at this stage but may be needed if it's merged before this one

edleeks87 (dismissed):

src/components/decimal/decimal.component.tsx
```diff
@@ -70,268 +71,290 @@ export interface DecimalProps
...
+      id,
+      name,
+      allowEmptyValue = false,
+      required,
+      locale,
+      value,
+      inputRef,
+      ...rest
+    }: DecimalProps,
+    ref: React.Ref<HTMLInputElement>
```
mkrds, line 96:
```suggestion
    ref: React.ForwardedRef<HTMLInputElement>
```

src/components/search/search.component.tsx
```diff
@@ -67,218 +68,229 @@ export interface SearchProps extends ValidationProps, MarginProps {
...
+      "aria-label": ariaLabel = "search",
+      inputRef,
+      tabIndex,
+      error,
+      warning,
+      info,
+      tooltipPosition,
+      ...rest
+    }: SearchProps,
+    ref: React.Ref<HTMLInputElement>
```
mkrds, line 100:
```suggestion
    ref: React.ForwardedRef<HTMLInputElement>
```

src/components/numeral-date/numeral-date.component.tsx
mkrds: Sorry it should have been:

```suggestion
                        refs.current[index] = element;
```

Right now it looks like we're unnecessarily wrapping the `element` in ref-like `{ current: element }` object.
And then on the line 308, we're accessing it this `(ref) => ref.current === document.activeElement`.
It seems that this one additional level of nesting is unnecesary.

mkrds (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #5730, opened 2023-01-05
[Box, Typography] convert from styled component

nineteen88:
### Proposed behaviour

Convert the Box and Typography components from Styled Components so that we lock down the interface and specify the relevant prop types

### Current behaviour

Box and Typography components are currently Styled Components and will accept any props even if undocumented.

### Checklist

#### QA

### Testing instructions

Regression needed for Box and Typography and any consumers. Chromatic will pick up any issues.


src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,89 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+  /** Override the backgroundColor style */
+  backgroundColor?: string;
+  /** Override the bg value shorthand for backgroundColor */
+  bg?: string;
+  /** Override the opacity value */
+  opacity?: string | number;
+  /** Override the withMargin boolean */
+  withMargin?: boolean;
+  /** Override the className value */
+  className?: string;
```
Parsium, line 75:
```suggestion
/** @private @ignore */
className?: string;
```
I think while some components do support `className`, this is usually a private prop as we don't encourage consumers to use classes for custom css styling.

Parsium (commented):
As some of the newly documented props (like `data-element` and `dataRole`) are not showing in the storybook docs, I would modify the Props section in `typography.stories.mdx` to be the following so all of the props are shown with no duplicates :)

```suggestion
## Props

<StyledSystemProps of={Typography} noHeader spacing />
```

src/components/box/box.component.tsx
```diff
@@ -52,87 +59,70 @@ export interface BoxProps
   rowGap?: Gap;
   /** Design Token for Box Shadow. Note: please check that the box shadow design token you are using is compatible with the Box component. */
   boxShadow?: BoxShadowsType;
+  /** Set the className attribute of the Box component */
+  className?: string;
```
Parsium, line 63:
```suggestion
  /** @private @ignore */
  className?: string;
```
Similar to `Typography`

src/components/card/card.stories.tsx
```diff
@@ -512,7 +512,7 @@ export const WithDraggable: StoryFn = () => {
 
     return (
       <Box width="700px" height="450px">
-        <Box
+        <div
```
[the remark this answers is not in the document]
nineteen88: Yep, fair point. Done.

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,89 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+  textTransform,
+  lineHeight,
+  textDecoration,
+  display,
+  listStyleType,
+  whiteSpace,
+  wordWrap,
+  textOverflow,
+  truncate,
+  color,
```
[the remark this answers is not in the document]
nineteen88: Missed this. Fixed.

src/components/box/box.component.tsx
nineteen88: Makes sense

src/components/typography/typography.component.tsx
nineteen88: Makes sense

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,89 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+  /** Override the color style */
+  color?: string;
+  /** Override the backgroundColor style */
+  backgroundColor?: string;
+  /** Override the bg value shorthand for backgroundColor */
+  bg?: string;
+  /** Override the opacity value */
+  opacity?: string | number;
+  /** Override the withMargin boolean */
+  withMargin?: boolean;
```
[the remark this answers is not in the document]
nineteen88: You're right. I've gone back through the code and tests and I can't see why withMargin was ever added. This has been removed.

nineteen88:
> As some of `Typography`'s newly documented props (like `data-element` and `data-role`) are not showing in the storybook docs, I would modify the Props section in `typography.stories.mdx` to be the following so all of the props are shown with no duplicates :)
> 
> ```
> ## Props
> 
> <StyledSystemProps of={Typography} noHeader spacing />
> ```

I can't close this one because it wasn't a code comment but this has been changed now as well. Resolved.

src/components/box/box.component.tsx
```diff
@@ -52,87 +59,70 @@ export interface BoxProps
   rowGap?: Gap;
   /** Design Token for Box Shadow. Note: please check that the box shadow design token you are using is compatible with the Box component. */
   boxShadow?: BoxShadowsType;
+  /** @private @ignore */
+  className?: string;
+  /** Set the color attribute of the Box component */
+  color?: string;
```
mkrds, line 65: It think that after adding `color`, `bg`, `backgroundColor` and `opacity` props explicitly you can remove the `interface BoxProps extends ColorProps`

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,80 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
-  /** Override the word-wrap type */
+  /** Override the word-wrap */
   wordWrap?: string;
-  /** Override the text-overflow type */
+  /** Override the text-overflow */
   textOverflow?: string;
   /** Apply truncation */
   truncate?: boolean;
+  /** Override the color style */
+  color?: string;
```
mkrds, line 65: Same here, I think `TypographProps extends ColorProps` may be removed

src/components/box/box.component.tsx
nineteen88: Fixed

src/components/typography/typography.component.tsx
nineteen88: Fixed

Parsium:
I think we still have some duplicate in props in the prop tables for `Box` and `Typography` when you load storybook. This is because we have manually defined some prop tables alongside storybook autogenerating one.

src/components/box/box.stories.mdx
```diff
@@ -374,27 +374,4 @@ If no `scrollVariant` prop is set then the scroll bar will resort to the browser
...
-  rows={[
-    {
-      name: "tabIndex",
-      type: { summary: "number | string" },
-      description: "Allows a tabindex to be specified",
-      required: false,
-    },
-  ]}
-/>
+<StyledSystemProps of={Box} spacing color flexBox layout position noHeader />
```
Parsium, line 377:
```suggestion
<StyledSystemProps of={Box} spacing flexBox layout position noHeader />
```
The colour-related props are duplicated in storybook, so given we are defining them ourselves rather than extending styled system's prop interface, this should address it?

mkrds (approved):

Parsium (approved):

ZhuoyuJin (approved):

src/components/box/box.component.tsx
```diff
@@ -52,87 +57,70 @@ export interface BoxProps
...
-    boxShadow &&
-    css`
-      box-shadow: var(--${boxShadow});
-    `}
-`;
-
-Box.defaultProps = {
-  theme: BaseTheme,
-};
+export const Box = React.forwardRef<HTMLDivElement, BoxProps>(
```
przemyslawzalewski, line 72: Hardcoding the `HTMLDivElement` is improper due to `as` support - it will vary with the parameter - just like it previously did with styled-components. This causes issues with `onClick` not being available despite setting `as="button"` - while at the same time the `handleKeyPress` is working fine ⚠️ 

Zrzut ekranu 2023-01-25 o 14 51 33

Zrzut ekranu 2023-01-25 o 14 51 44

Zrzut ekranu 2023-01-25 o 14 53 19

Zrzut ekranu 2023-01-25 o 15 00 23

This is kind of a blocker for us (SBS-49786).

Parsium:
Blocking this PR for now, see tickets FE-5521 and FE-5580 for context

src/components/box/box.component.tsx
```diff
@@ -55,103 +59,73 @@ export interface BoxProps
...
+        overflowWrap={overflowWrap}
+        scrollVariant={scrollVariant}
+        boxSizing={boxSizing}
+        gap={gap}
+        columnGap={columnGap}
+        rowGap={rowGap}
+        tabIndex={tabIndex}
+        className={className}
+        ref={ref}
+        bg={bg}
```
tomdavies73, line 112: **suggestion:** Do we need the `bg` and `backgroundColor` props? I was under the impression that `bg` was just shorthand for `backgroundColor`, and it appears that is still the case. I've done some digging and found the `styledColor` util which accepts either `bg` or `backgroundColor` I'm assuming (possibly incorrectly) due to `bg` or `backgroundColor` being used interchangeably across the codebase. Imo we just need one or the other here

src/components/box/box.component.tsx
```diff
@@ -33,18 +31,24 @@ export interface BoxProps
   extends SpaceProps,
     LayoutProps,
     FlexboxProps,
-    ColorProps,
+    TagProps,
     Omit<PositionProps, "zIndex"> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   as?: keyof JSX.IntrinsicElements | React.ComponentType<any>;
```
tomdavies73, line 37: **issue** (kind-of): So as we're not spreading ...rest onto `<StyledBox>` and we've moved the `calculatePosition` function into `box.style.ts` it might be possible that `as` is not being passed or applied, potentially. I've got quite a bit to take a look at so can't do too much in depth testing but it may be worth taking a look

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,80 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+  /** Override the backgroundColor style */
+  backgroundColor?: string;
```
tomdavies73, line 67: **suggestion**: again, might be wrong here but unsure if we need both `bg` and `backgroundColor` props

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,80 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+  /** Override the bg value shorthand for backgroundColor */
+  bg?: string;
+  /** Override the opacity value */
+  opacity?: string | number;
+  /** @private @ignore */
+  className?: string;
```
tomdavies73, line 73: **question** (non-blocking): same comment applies here, do we need to be adding `className`? Perhaps there's a requirement i'm not privvy too

src/components/typography/typography.component.tsx
```diff
@@ -46,214 +53,80 @@ export interface TypographyProps extends SpaceProps, ColorProps {
...
+      whiteSpace={whiteSpace}
+      wordWrap={wordWrap}
+      textOverflow={textOverflow}
+      truncate={truncate}
+      color={color}
+      backgroundColor={backgroundColor}
+      bg={bg}
+      opacity={opacity}
+      className={className}
+      {...tagComponent(dataComponent, rest)}
```
tomdavies73, line 121: **question** (non-blocking): same comment applies here re `tagProps` etc.

src/components/box/box.component.tsx
tomdavies73: I know this wasn't introduced by you, but might be worth investigating to see if we can remove one

src/components/typography/typography.component.tsx
Parsium: @tomdavies73 because we have supported both `bg` and `backgroundColor` so far, it might be wise to continue supporting both of them for now so we don't break any styling for consumers using the shorthand `bg` prop.

The  shorthand prop is based from styled system similar to `m`, `pr`, etc, so perhaps we can discuss if we want to continue using shorthands once we start working on removing styled system (FE-5531)

tomdavies73: @Parsium fair, might be worth having a discussion about shorthand props now and seeing what we'd like to do. As this PR is already a breaking change, it seems like the perfect opportunity to get rid of `bg` if that's the approach we take. I'm happy with either approach though 👍

nineteen88: Yeah, happy to discuss when we have some time this morning

src/components/box/box.style.ts
```diff
@@ -0,0 +1,112 @@
...
+  flexbox,
+  position as positionFn,
+  PositionProps,
+} from "styled-system";
+import BaseTheme from "../../style/themes/base";
+import styledColor from "../../style/utils/color";
+import { BoxProps } from "./box.component";
+import boxConfig from "./box.config";
+
+export type StyledBoxProps = BoxProps;
```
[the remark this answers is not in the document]
nineteen88: Removed the additional type export

src/components/box/box.component.tsx
```diff
@@ -55,103 +59,73 @@ export interface BoxProps
...
+        tabIndex={tabIndex}
+        className={className}
+        ref={ref}
+        bg={bg}
+        backgroundColor={backgroundColor}
+        boxShadow={boxShadow}
+        borderRadius={borderRadius}
+        color={color}
+        opacity={opacity}
+        {...tagComponent(dataComponent, rest)}
```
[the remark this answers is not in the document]
nineteen88: We don't want to remove existing functionality, and this was possible before and we want it to continue to be supported

src/components/box/box.component.tsx
nineteen88: It's a good suggestion and I think we will streamline this as part of the styled-system removal at a later date. For now supporting both (since it was possible before), will remain

src/components/typography/typography.component.tsx
nineteen88: `className` is used and its current removal will break tests. Potentially we can remove `className` in another ticket though

src/components/box/box.component.tsx
```diff
@@ -55,103 +59,73 @@ export interface BoxProps
   boxShadow?: BoxShadowsType;
   /** Design Token for Border Radius. Note: please check that the border radius design token you are using is compatible with the Box component. **This prop will not do anything if you have the roundedCornerOptOut flag set in the CarbonProvider** */
   borderRadius?: BorderRadiusType;
+  /** @private @ignore */
+  className?: string;
```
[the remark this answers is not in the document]
Parsium: @nineteen88 @tomdavies73 for context FE-6066 is the ticket for remove `className`

src/components/typography/typography.style.ts
```diff
@@ -0,0 +1,194 @@
...
+    variant,
+    as,
+    fontSize,
+    fontWeight,
+    textTransform,
+    lineHeight,
+    textDecoration,
+  }: TypographyProps) => {
+    return {
+      as: as || getAs(variant),
```
Parsium, line 139: **thought**: perhaps when we come to do FE-6177, we should consider deprecating `as` here anyway since we have the `variant` prop

Parsium (dismissed):

src/components/box/box.component.tsx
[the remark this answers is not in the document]
edleeks87: we will need to maintain className on any component that we've used `styled(Box)` on fyi, that is why tests are breaking

src/components/duelling-picklist/picklist/picklist.style.ts
```diff
@@ -1,7 +1,7 @@
 import styled from "styled-components";
-import Box from "../../box";
+import StyledBox from "../../box/box.style";
 
-const StyledPicklist = styled(Box).attrs({ as: "ul" })`
+const StyledPicklist = styled(StyledBox).attrs({ as: "ul" })`
```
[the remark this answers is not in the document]
edleeks87: comment (non-blocking): This could stay as `styled(Box)` fyi.

I don't think removing polymorphism here would actually prevent this approach being used internally by us via `attrs` so not sure the comment is relevant in this use case. Removing/ reducing the ability to use polymorphism would just affect someone trying to it like `<Box as="button">foo</Box>`.

https://codesandbox.io/s/clever-ardinghelli-vjn3qp

edleeks87 (approved):

Parsium (approved):

edleeks87 (approved):

Parsium (approved):

ZhuoyuJin (approved):

PR #5737, opened 2023-01-09
fix(text-editor): add tabIndex 0 to editor container and update types for value and onChange FE-5582

### LINKED ISSUE #5685: TextEditor: on a Dialog component, you cannot access the body via the keyboard
### Current behaviour

When you edit a comment and use the keyboard to move through the edit dialogue you cannot access the body of the text. Focus moves from the title field to the Bold text.

### Expected behaviour

The expected behaviour is that there is a yellow frame around the text edit box and the cursor appears before the first character in the body text box.  

https://codesandbox.io/s/gifted-tree-s6wzy7

### JIRA Ticket (Sage Only)

SBS-41698

### Suggested Solution

_No response_

### Carbon Version

111.16.0

### Design Tokens Version

2.48.0

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5685

### Proposed behaviour

Adds `tabIndex={0}` to `draft-js` `Editor` component so that it is found in the FocusTrap's
focusable elements. Updates the `TextEditor` types for `value` and `onChange` to replace
`Record<string, unknown>` with `EditorState`

Adds cypress test to `Dialog` to catch any potential future regressions
Updates default `Dialog` story to include `TextEditor` as a child

### Current behaviour

The `Editor` component rendered as part of `TextEditor` composition has `tabIndex` `-1` so it is filtered out of focusable elements in the focus trap

### Checklist

#### QA

### Testing instructions

See linked sandbox and default dialog story


src/components/dialog/dialog.stories.tsx
```diff
@@ -54,6 +59,11 @@ export const DefaultStory: StoryFn = () => {
           <Textbox label="Birth Place" />
           <Textbox label="Favourite Colour" />
           <Textbox label="Address" />
+          <TextEditor
+            onChange={handleEditorChange}
+            value={editorState}
+            labelText="Text Editor Label"
+          />
```
grabkowski, line 66: I think that adding that to the story available for users would add more complexity to already complex example. What do you think of adding that to the test story instead?

edleeks87: Yeah sounds like a better approach

robinzigmond: I agree with Kuba that we definitely shouldn't add this to the default story. Leave that as it is and add a new example in test stories with the Text Editor, as he suggests.

src/components/dialog/dialog.test.js
```diff
@@ -274,6 +285,10 @@ context("Testing Dialog component", () => {
       cy.get("body").tab();
       getInput(2).should("be.focused");
       cy.get("body").tab();
+      textEditorInput().should("be.focused");
```
robinzigmond, line 288: similarly to Kuba's comment on the stories, I'd rather add a dedicated test for this than just add it to an existing big test - to make it clear we've got a specific regression test for TextEditor inside a Dialog. I'd also make a new component in this file for testing this rather than hijacking DialogComponent, but I'm happy to leave that up to you, you may disagree.

robinzigmond (approved):

grabkowski (approved):

ZhuoyuJin (approved):

PR #5755, opened 2023-01-16
feat(pager): surface hideDisabledElements prop to ensure disabled elements are hidden

tomdavies73:
### Proposed behaviour

This change will introduce a `hideDisabledElements` prop, which if `true` will set the css `visibility` property of all `disabled`elements to `hidden`. Therefore, this addition will provide consumers with flexibility regarding disabled elements' visibility to all end users.

Screenshot 2023-01-16 at 15 42 10

### Current behaviour

Currently, all disabled elements are visible and presented to users by default.

Screenshot 2023-01-16 at 15 42 21

### Checklist

#### QA

### Testing instructions


src/components/pager/pager.stories.mdx
```diff
@@ -36,6 +36,15 @@ import Pager from "carbon-react/lib/components/pager";
   <Story story={stories.Default} name="default" />
 </Canvas>
 
+### with hideDisabledElements
+
+The example below demonstrates the functionality of the `hideDisabledElements` prop, when `true` all elements that typically would be disabled
+are instead hidden. This can prop be used in conjunction with all other `Pager` props and functionality.
```
DipperTheDan, line 42:
```suggestion
are instead hidden. This prop can be used in conjunction with all other `Pager` props and functionality.
```

DipperTheDan (commented):
Looks good. Just this one minor mistake to correct in the docs please 👍🏻

src/components/pager/__internal__/pager-navigation-link.component.tsx
```diff
@@ -85,8 +88,11 @@ const PagerNavigationLink = ({
 
   const { text } = navLinkConfig[type];
 
+  const hideDisabledButtons = hideDisabledElements && disabled();
+
   return (
     <StyledPagerLink
+      hideDisabledButtons={hideDisabledButtons}
```
Parsium, line 95: I'd recommend inside of hiding the element via css, using the `hidden` attribute like so:
```suggestion
      hidden={hideDisabledButtons}
```
The reason for this is hiding the element via css will visually hide the element, but it will still be recognised by the browser, so the links could potentially still be readable by screen readers. Using `hidden` instead will prevent the browser from rendering it which should avoid this issue.

Parsium (changes requested):

src/components/pager/__internal__/pager-navigation.component.tsx
```diff
@@ -118,13 +121,15 @@ const PagerNavigation = ({
       {!hasTwoPages && showFirstAndLastButtons && (
         <PagerNavigationLink
           type="first"
+          hideDisabledElements={hideDisabledElements}
```
Parsium, line 124: I'd recommend inside of hiding the element via css, using the `hidden` attribute like so:

```suggestion
      hidden={hideDisabledButtons}
```
The reason for this is hiding the element via css will visually hide the element, but it will still be recognised by the browser, so the links could potentially still be readable by screen readers. Using `hidden` instead will prevent the browser from rendering it which should avoid this issue.

Parsium (approved):

edleeks87 (approved):

harpalsingh:
Passed UXQA on this on the request of the task, but something to think about in the future is around how we inform voiceover users when they are shown/removed based on changes on the screen.

divyajindel (approved):

PR #5757, opened 2023-01-17
feat(loader,loader-bar): display loading message if user has opt-ed to reduced motion

Parsium:
### Proposed behaviour

- Alternative loading message for `Loader` and `LoaderBar` 
   - Replace animation with text when user has indicated a preference for reduced motion in their OS settings, or when the user's motion preference cannot be determined (e.g. with older browsers like IE11)
   - Add translation support for loading message (with English and Polish examples added to the respective built-in locales)
   - For `Switch`, remove fix width constraint to fit contents (specifically to support [loading pattern](https://carbon.sage.com/?path=/docs/switch--loading) since alternative text could be long)

#### `Loader` demo

https://user-images.githubusercontent.com/18368713/214578476-5b8f4bb2-910c-409e-9ad6-444c35a47a92.mov

#### `LoaderBar` demo

https://user-images.githubusercontent.com/18368713/214578532-c53f4f7e-15b8-421c-b251-2bf56603fc5c.mov

#### `Button` with `Loader` demo

https://user-images.githubusercontent.com/18368713/214578605-0eba7cb0-d7b1-4299-92ca-6be255500a99.mov

#### `Switch` with `Loader` demo

https://user-images.githubusercontent.com/18368713/214578667-59487ddf-146d-42ba-ac9a-e9bd6f268bab.mov

### Current behaviour

`Loader` and `LoaderBar`'s blinking and scrolling animations cannot be controlled by the user. Since these components could be used in parallel with other content, this could be distracting to users with reading difficulties or attention deficit disorders (see WCAG 2.1 Operable guideline ["Pause, Stop and Hide"](https://www.w3.org/TR/WCAG21/#pause-stop-hide) for details).

### Checklist

#### QA

### Testing instructions

- Open the following stories and with "Reduce Motion" enabled in the OS accessibility preferences, check that animation is replaced with the alternative loading message (for macOS, setting can be found in: `System Preferences > Accessibility > Display > Reduce Motion`):
    - `loader--default-story`
    - `loader-bar--default-story`
    - `switch--loading`
    - `select--with-is-loading-prop`
- Check visuals for loading message in the English and Polish locales, verify no regressions have been introduced.


robinzigmond (commented):
My biggest concern with this PR is setting the container of these 2 components to be an ARIA live region. There are 2 reasons why I believe this is wrong:

1) These components will get dynamically inserted into the page when data is being loaded, then removed afterwards - they won't be on the page permanently. So we'll have a live region with fixed content that itself comes and goes. My understanding is that live regions are designed to work as a fixed container on the page that's there all the time, and it's changing content inside that which should be updated. I see it works on your videos, but these are with Mac and Voiceover, which I think is one of the few screenreaders on which this will work properly. (See https://www.scottohara.me/blog/2022/02/05/are-we-live.html, in the section titled "Non-normative rule for common live region use case".) So it's definitely worth checking this with eg NVDA on Windows with various browsers - at the very least ask QA to do this in the testing instructions. But I have significant enough doubts that this will work that I'd prefer not to pass the review at this stage, as there will be a major rewrite needed if this fails.

2) Even if this works on all browser/OS/screenreader combinations, I still don't think it's the right behaviour. In your examples, the screenreader announces "loading" when the loader appears,  but says nothing when it disappears - the user will have no idea when their data is loaded or what it is (unless the containing component explicitly puts focus there afterward). I think the best practice here will be for the containing region to be a live region, which will ensure that both when the loader appears and when it's replaced with "actual content" a screenreader will inform users as to what's appeared.

src/components/loader-bar/loader-bar.component.tsx
```diff
@@ -1,20 +1,52 @@
...
   InnerBar,
   StyledLoader,
   StyledLoaderBarProps,
 } from "./loader-bar.style";
 
-export interface LoaderBarProps extends StyledLoaderBarProps, MarginProps {}
+export interface LoaderBarProps
+  extends StyledLoaderBarProps,
+    MarginProps,
+    TagProps {
```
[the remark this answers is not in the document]
Parsium: I added this in after noticing `Loader` did not formally have the `data-*` props in its interface and that we didn't have any checks on what props were being spread via `rest` (previously on line 14) so consumers could potentially pass any html attribute to the component. The `data-*` props are needed for consumer props which is why I've decided to add them in here

src/components/loader/loader.component.tsx
```diff
@@ -1,45 +1,69 @@
...
 import StyledLoaderSquare, {
   StyledLoaderSquareProps,
 } from "./loader-square.style";
 
-export interface LoaderProps extends StyledLoaderSquareProps, MarginProps {
-  /** Specify an aria-label for the Loader component */
+export interface LoaderProps
+  extends StyledLoaderSquareProps,
+    MarginProps,
+    TagProps {
```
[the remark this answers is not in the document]
Parsium: For similar reasons as with `Loader` :)

src/components/loader-bar/loader-bar.style.ts
```diff
@@ -23,9 +23,6 @@ const INNER_BAR_LENGTH = "128px";
 const StyledLoader = styled.div<StyledLoaderBarProps & MarginProps>`
   ${margin}
   text-align: center;
```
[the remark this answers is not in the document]
Parsium: These styles here were setting the `StyledLoader`'s height to zero when it does not contain the bar animation as a child - so the alternative text was not showing. I tried to see in the git history why we had these text-based styles originally, my guess is this forces the `StyledLoader` to naturally take the height of bar animation and not the line height of the used font.

src/components/loader-bar/loader-bar.spec.tsx
```diff
@@ -1,22 +1,34 @@
...
+  };
+});
 
 describe("LoaderBar", () => {
-  let wrapper: ReactWrapper;
-  it("renders component as expected", () => {
-    wrapper = mount(<LoaderBar />);
-    const innerBar = wrapper.find(InnerBar);
-    expect(innerBar).toBeTruthy();
+  it("if users disallows animations or their preference cannot be determined, renders alternative loading text", () => {
```
robinzigmond, line 16:
```suggestion
  it("if user disallows animations or their preference cannot be determined, renders alternative loading text", () => {
```

src/components/loader-bar/loader-bar.stories.mdx
```diff
@@ -61,3 +63,20 @@ This is an example of a large LoaderBar component.
...
+
+The following keys are available to override the translations for this component by passing in a custom locale object
+to the [i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
+
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "loader.loading",
+      description:
+        "An alternative message to replace loading animation - for users who have requested minimal motion-based animation to be used via their OS accessibility settings.",
```
robinzigmond, line 77: I would also mention the other thing this is used for - the accessible name of the component itself (some users might be able to see the animated loader bar while also listening to a screenreader describing it)

src/components/loader-bar/loader-bar.component.tsx
```diff
@@ -1,20 +1,44 @@
...
+    "screen and (prefers-reduced-motion: no-preference)"
+  );
+
   return (
-    <StyledLoader {...rest} {...tagComponent("loader-bar", rest)}>
-      <StyledLoaderBar size={size}>
-        <InnerBar size={size} />
-      </StyledLoaderBar>
+    <StyledLoader
+      aria-label={l.loader.loading()}
```
robinzigmond, line 30: sorry I didn't pick this up before - but it seems curious to me that we allow the consumer to override the `aria-label` on `Loader` but not on `LoaderBar`. This seems inconsistent. (I'm not actually sure we need to allow this to be overridden tbh, but there's no harm in it.)

src/components/loader/loader.stories.mdx
```diff
@@ -101,3 +102,20 @@ please remember to pass the `isInsideButton` prop to the `Loader` component.
...
+
+The following keys are available to override the translations for this component by passing in a custom locale object
+to the [i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
+
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "loader.loading",
+      description:
+        "An alternative message to replace loading animation - for users who requested minimal motion-based animation to be used via their OS accessibility settings.",
```
robinzigmond, line 116: same comment here as on LoaderBar

src/components/loader/loader.spec.tsx
```diff
@@ -1,36 +1,57 @@
...
-  testStyledSystemMargin((props) => <Loader {...props} />);
-
-  it("renders as expected", () => {
-    assertStyleMatch(
-      {
-        textAlign: "center",
-      },
-      renderStyles().toJSON()
-    );
+  describe("margins", () =>
```
edleeks87, line 20: Not something to change but we should probably review the text for the styled system utils so we don't need to add the `describe("margins")...` as the output still reads a little funny, but it's better than before etc so can be left as you have it for now

src/components/loader-bar/loader-bar.component.tsx
Parsium: The `aria-label` prop for `Loader` already existed which is why I included it there but not with `LoaderBar` - to avoid adding a prop deprecation or introduce a prop which we don't really need if we use the locale variable

robinzigmond: Ah OK, fair enough - no need to add it now then

robinzigmond (dismissed):

edleeks87 (approved):

edleeks87 (approved):

robinzigmond (approved):

harpalsingh:
Checked the voiceover and motion on Mac, all good.

DlgSHi (approved):

PR #5775, opened 2023-01-23
docs: change 'Storybook' to 'Carbon' in <title>

dcporter65:
### Proposed behaviour

Change the HTML title rendered in the browser to display "Carbon" instead of "Storybook".

Also, change the label on the app icon when you add the Carbon site to a mobile device's Home screen to read "Carbon" instead of "Storybook".

See FE-5630

### Current behaviour

The title is displayed in the current format:

`${storyData.kind} - ${storyData.name} ⋅ Storybook`
 
For example "Welcome - Welcome Page - Storybook".

storybook-tab

Also, when you view the Carbon website on a mobile device and select **Add to Home screen**, the label added to the Home screen appears as "Storybook".

### Checklist

#### QA

### Testing instructions

1. From this branch run `npm start`.
2. In your browser, go to http://localhost:9001
3. The title in the browser tab should initially display "Carbon" and then "Welcome - Welcome page - Carbon"
4. On a mobile device browser go to http://192.168.86.28:9001
5. Select **Add to Home screen**.
6. The icon added to the Home screen should be labelled "Carbon"

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -2,6 +2,7 @@ import { addons, types } from "@storybook/addons";
 import sageTheme from "./sageTheme";
 import { ADDON_ID, TOOL_ID } from "./version-picker/constants";
 import { VersionPicker } from "./version-picker";
+import "./titleAddon";
 
 addons.setConfig({
   theme: sageTheme,
--- a/.storybook/titleAddon.js
+++ b/.storybook/titleAddon.js
@@ -1,0 +1,28 @@
+import addons from "@storybook/addons";
+import { STORY_RENDERED } from "@storybook/core-events";
+
+addons.register("TitleAddon", (api) => {
+  const customTitle = "Carbon";
+  let interval = null;
+  const setTitle = () => {
+    clearTimeout(interval);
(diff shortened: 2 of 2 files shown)


robinzigmond (commented):
Hi @dcporter65 ,

Thanks for working on this. I agree it would be good to have the title showing Carbon rather than Storybook - however I don't know if it's important enough to justify this rather hacky approach (running a function every 100 milliseconds), especially if it's going to stop working when we upgrade storybook.

Having researched this a little myself it does look like this is probably the only way to do this for now - but I don't personally feel the benefit is enough to justify this approach. Especially when we have plans - even though they are some way off (I guess a year minimum, maybe more) - to eventually stop using Storybook for documentation and move to another solution.

I'll leave this here for you and other reviewers to respond to - I don't object to approving this if it's agreed that the end justifies the means, but I think it might need some wider discussion.

dcporter65:
Thanks @robinzigmond 

That's fine. I was concerned that the solution was too hacky. I'll close this PR and we'll live with "Storybook" appearing in the browser tab title.

PR #5792, opened 2023-01-31
feat(tab): introduce titleProps for passing props to corresponding title

Parsium:
### Proposed behaviour

- `Tab`: Add new `titleProps` prop - which accepts an object of props to pass to the tab's corresponding `TabTitle`.
- `TabTitle`: Accepts new `data-role` prop to be used for testing and analytics purposes

### Current behaviour

- Consumers cannot set a `data-role` property for a `Tab`'s corresponding `TabTitle`, which has been raised as a feature request to allow consumers to query a tab's title using testing/analytics tools.

### Checklist

### Testing instructions

- Verify tab title has `data-role` property set when specified
- Double check typescript and docs definitions 

The following CodeSandbox is an example of the broken behaviour: [Go to codesandbox](https://codesandbox.io/s/tab-data-role-6v5wvp?file=/src/App.js)

### DIFF
--- a/src/components/tabs/tabs.component.js
+++ b/src/components/tabs/tabs.component.js
@@ -187,6 +187,7 @@ const Tabs = ({
         infoMessage,
         href,
         customLayout,
+        titleProps,
       } = child.props;
       const refId = `${tabId}-tab`;
       const errors = tabsErrors[tabId];
@@ -234,8 +235,9 @@ const Tabs = ({
         return summaryOfMessages.map((value) => `• ${value}`).join("\n");
       };
 
-      const tabTitle = (
+      return (
         <TabTitle
+          {...titleProps}
           position={isInSidebar ? "left" : position}
           className={child.props.className || ""}
           dataTabId={tabId}
@@ -266,8 +268,6 @@ const Tabs = ({
           align={align}
         />
       );
-
-      return tabTitle;
     });
 
     return (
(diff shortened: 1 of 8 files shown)


src/components/tabs/tabs.component.js
```diff
@@ -42,10 +42,9 @@ const Tabs = ({
   ...rest
 }) => {
   /** The children nodes converted into an Array */
-  const filteredChildren = useMemo(
-    () => Children.toArray(children).filter((child) => child),
-    [children]
-  );
+  const filteredChildren = useMemo(() => Children.toArray(children), [
```
robinzigmond, line 45: is there a reason we're no longer filtering to remove any falsy children? `null` is a valid child and will certainly cause an error when accessing `child.props` below - and this could happen in practice.

If we are sure we don't need to filter then we should change the name to something like `childrenArray`, since we're no longer filtering. Also we could probably do without the array and just use `Children.map` later when we map over the children.

But unless I'm missing something important I think we should keep this filter in place.

src/components/tabs/tabs.component.js
```diff
@@ -234,11 +234,14 @@ const Tabs = ({
...
 
-      const tabTitle = (
+      return (
         <TabTitle
           position={isInSidebar ? "left" : position}
           className={child.props.className || ""}
           dataTabId={tabId}
+          data-role={
+            titleProps !== undefined ? titleProps["data-role"] : undefined
+          }
```
robinzigmond, line 244: since we're limiting `titleProps` to just contain `data-role` (at least for now), I would much prefer to spread props: `{...titleProps}` - I'm pretty sure you can spread undefined as an object and it will add no properties.

I especially think it's right to do it this way since I can only assume the point of having a `titleProps` object with just a single property is to expand this interface later if needed, and we don't want to have to add all those potential props individually.

robinzigmond (commented):
I presume we're adding a `titleProps` prop which holds an object with one key, rather than a single prop called (eg) `titleDataRole` in case we need to expose more props for `TabTitle` later? Seems a good idea to me but if that's not the idea I'm not sure what the advantage is.

src/components/tabs/tabs.component.js
edleeks87: Likewise I think we still need to filter for falsy children here unless I'm missing something somewhere

src/components/tabs/tabs.component.js
edleeks87: Agree with Robin would prefer to see it spread here for simplicity's sake

src/components/tabs/tabs.component.js
```diff
@@ -285,37 +286,32 @@ const Tabs = ({
...
+          ariaLabelledby: `${tab.props.tabId}-tab`,
+          updateErrors,
+          updateWarnings,
+          updateInfos,
+        })
+      );
     }
 
-    const tabs = filteredChildren.map((child) => {
+    return filteredChildren.map((child) => {
```
edleeks87, line 314: I think this could will throw if any falsy children are in the array

src/components/tabs/tabs.component.js
Parsium: Ah this was an oversight on my part, I didn't realise this removed falsy children 👍🏼

src/components/tabs/tabs.component.js
Parsium: Originally I decided to do this so we weren't inadvertently supporting any prop that could be passed via `titleProps`, while still using the `titleProps` approach in case we need to expose more props later. However, I can change this if you think this is excessive and would improve readability

robinzigmond: I see your point - thanks for explaining. On balance I'd rather spread them, if consumers try to pass something else they will hopefully be aware it's undocumented and is being used at their own risk so I don't see it as a problem. But I won't argue any further if you decide to keep it as you had it.

robinzigmond: although even with it this way I would prefer to see it changed to `data-role={titleProps?.["data-role"]}`, which is more readable and should come to the same thing.

edleeks87: Yeah I'd prefer either this or destructure (requires default value of {}) it from `titleProps` as I think someone may come across your implementation in the future and refactor it and miss your valid reasoning you mention above @Parsium 

```
const {dataRole} = titleProps;
```

Parsium: I agree with both of you on this - will push a change now 👍🏼

robinzigmond (approved):

edleeks87 (approved):

divyajindel (approved):

PR #5806, opened 2023-02-03
feat(settings-row, heading): surface headingType prop on components

### LINKED ISSUE #5723: Allow different heading levels to be used in Setting Row component
### Desired behaviour

When using the 'Setting Row' component, have the ability to choose what level of heading you require for the 'Title' element. This will resolve the accessibility issue of using headings in non-sequential order. 

Similar issue to https://github.com/Sage/carbon/issues/4607 

### Current behaviour

The 'Setting Row' component Title always uses a `h1` for the Title element. 

### Suggested Solution

Allowing use of heading styles defined in the Typography component, or more flexibility in the Heading component to choose which level of heading you require. 

https://carbon.sage.com/?path=/docs/setting-row--default-story

### Confidentiality

tomdavies73:
fix #5723

### Proposed behaviour

This change will surface the `headingType` prop on the `SettingsRow` and `Heading` components to provide consumers with the means to explicitly define which `<h>` tag they would like to be rendered when the `title` prop is passed in `SettingsRow` and in the `Heading` component itself. This will ensure a proper order of importance is established within the entire page when the components are used or re-used. Each `string` from `h1-h5` will each represent one of the six HTML Section Heading elements, ensuring a proper web page hierarchy can be created, to improve readability and accessibility. 

### Current behaviour

Currently, when the `title` prop is passed within the `SettingsRow` component, or the `Heading` component is used, a `<h1>` tag is rendered by default. If the components were used multiple times in one page, this would mean multiple `<h1>` tags would be rendered, which goes against the recommended web standard. Furthermore, as there is currently no flexibility regarding what HTML Section Heading elements are used, a proper order of section importance can not be established.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/heading/heading.component.tsx
+++ b/src/components/heading/heading.component.tsx
@@ -20,6 +20,7 @@ import {
 } from "./heading.style";
 import useLocale from "../../hooks/__internal__/useLocale";
 
+export type HeadingType = "h1" | "h2" | "h3" | "h4" | "h5";
 export interface HeadingProps extends MarginProps, TagProps {
   /** Child elements */
   children?: React.ReactNode;
@@ -31,6 +32,8 @@ export interface HeadingProps extends MarginProps, TagProps {
   subheader?: React.ReactNode;
   /** Defines the subtitle id for the heading. */
   subtitleId?: string;
+  /** Defines the HTML heading element of the title. */
+  headingType?: HeadingType;
   /** Defines the help text for the heading. */
   help?: string;
   /** Defines the help link for the heading. */
@@ -66,6 +69,7 @@ export const Heading = ({
   separator = false,
   subheader,
   subtitleId,
+  headingType = "h1",
   title,
   titleId,
   ...rest
@@ -141,7 +145,7 @@ export const Heading = ({
         <StyledHeaderContent>
           <StyledHeadingTitle
             withMargin={!!pills || !!help}
-            variant="h1"
+            variant={headingType}
             data-element="title"
             id={titleId}
           >
--- a/src/components/heading/heading.spec.tsx
+++ b/src/components/heading/heading.spec.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { shallow, mount, ReactWrapper } from "enzyme";
-import Heading from ".";
+import Heading, { HeadingType } from ".";
 import {
   StyledHeader,
   StyledSubHeader,
@@ -39,6 +39,19 @@ describe("Heading", () => {
     it("renders a heading title with text from prop content", () => {
       expect(wrapper.find(StyledHeadingTitle).text()).toEqual("foo");
     });
+  });
+
+  describe("when headingType prop is provided", () => {
+    it.each<HeadingType>(["h1", "h2", "h3", "h4", "h5"])(
+      "HTML heading element is correct when headingType is %s",
+      (headingType) => {
+        const wrapper = mount(
+          <Heading headingType={headingType} title="foo" />
+        );
+
+        expect(wrapper.find(headingType).text()).toBe("foo");
+      }
+    );
   });
 
   describe("when the help prop is provided", () => {
--- a/src/components/heading/heading.stories.mdx
+++ b/src/components/heading/heading.stories.mdx
@@ -42,6 +42,21 @@ import Heading from "carbon-react/lib/components/heading";
   </Story>
 </Canvas>
 
+### Heading with headingType
+
+The `headingType` prop sets the HTML heading element which is rendered on the page. It can take values `"h1"` up to `"h5"`.
+
+<Canvas>
+  <Story name="with headingType">
+    <Heading headingType="h1" title="This is a h1 Title" />
+    <Heading headingType="h2" title="This is a h2 Title" />
+    <Heading headingType="h3" title="This is a h3 Title" />
+    <Heading headingType="h4" title="This is a h4 Title" />
+    <Heading headingType="h5" title="This is a h5 Title" />
+  </Story>
+</Canvas>
+
+
 ### Heading without Divider
 
 <Canvas>
--- a/src/components/heading/heading.test.js
+++ b/src/components/heading/heading.test.js
@@ -152,6 +152,19 @@ context("Testing Heading component", () => {
       }
     );
 
+    describe("when headingType prop is provided", () => {
+      it.each(["h1", "h2", "h3", "h4", "h5"])(
+        "should check HTML heading element is correct when headingType is %s",
+        (headingType) => {
+          CypressMountWithProviders(
+            <HeadingComponent headingType={headingType} title="foo" />
+          );
+
+          cy.get(headingType).contains("foo");
+        }
+      );
+    });
+
     describe("should render Heading component and check accessibility issues", () => {
       it("should check heading accessibility", () => {
         CypressMountWithProviders(<HeadingComponent />);
--- a/src/components/settings-row/settings-row.component.js
+++ b/src/components/settings-row/settings-row.component.js
@@ -17,6 +17,7 @@ const marginPropTypes = filterStyledSystemMarginProps(
 
 const SettingsRow = ({
   title,
+  headingType = "h3",
   description,
   children,
   className,
@@ -28,6 +29,7 @@ const SettingsRow = ({
 
     return (
       <Heading
+        headingType={headingType}
         title={title}
         subheader={description}
         separator={description !== undefined}
@@ -58,6 +60,8 @@ SettingsRow.propTypes = {
   className: PropTypes.string,
   /**  A title for this group of settings. */
   title: PropTypes.string,
+  /** Defines the HTML heading element of the `title` within the component. */
+  headingType: PropTypes.oneOf(["h1", "h2", "h3", "h4", "h5"]),
   /**  A string or JSX object that provides a short description about the group of settings. */
   description: PropTypes.node,
   /** Shows a divider below the component. */
--- a/src/components/settings-row/settings-row.d.ts
+++ b/src/components/settings-row/settings-row.d.ts
@@ -1,10 +1,13 @@
 import * as React from "react";
+import { HeadingType } from "../heading";
 
 export interface SettingsRowProps {
   /**  The CSS classes to apply to the component. */
   className?: string;
   /**  A title for this group of settings. */
   title?: string;
+  /** Defines the HTML heading element of the `title` within the component. */
+  headingType?: HeadingType;
   /**  A string or JSX object that provides a short description about the group of settings. */
   description?: React.ReactNode;
   /** Shows a divider below the component. */
--- a/src/components/settings-row/settings-row.spec.js
+++ b/src/components/settings-row/settings-row.spec.js
@@ -20,6 +20,7 @@ describe("SettingsRow", () => {
   describe("render", () => {
     const name = "foobar-row";
     const title = "Some Title";
+    const description = <span>Some descriptive text</span>;
     const childId = "my_child";
     const children = <span id={childId} />;
     let wrapper = shallow(
@@ -58,7 +59,6 @@ describe("SettingsRow", () => {
     });
 
     describe("when description is provided", () => {
-      const description = <span>Some descriptive text</span>;
       let head;
 
       beforeEach(() => {
@@ -75,6 +75,25 @@ describe("SettingsRow", () => {
       it("passes true as separator prop", () => {
         expect(head.prop("separator")).toBeTruthy();
       });
+    });
+
+    describe("when headingType prop is provided", () => {
+      it.each(["h1", "h2", "h3", "h4", "h5"])(
+        "HTML heading element is correct when headingType is %s",
+        (headingType) => {
+          wrapper = mount(
+            <SettingsRow
+              headingType={headingType}
+              title={title}
+              description={description}
+            >
+              Content for settings
+            </SettingsRow>
+          );
+
+          expect(wrapper.find(headingType).text()).toBe(title);
+        }
+      );
     });
 
     describe("when title is not provided", () => {
--- a/src/components/settings-row/settings-row.stories.mdx
+++ b/src/components/settings-row/settings-row.stories.mdx
@@ -44,6 +44,29 @@ All `children` are rendered in the input column to the right of the header colum
   </Story>
 </Canvas>
 
+### with HeadingType
+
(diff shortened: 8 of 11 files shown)


src/components/heading/heading.component.tsx
```diff
@@ -20,6 +20,7 @@ import {
 } from "./heading.style";
 import useLocale from "../../hooks/__internal__/useLocale";
 
+type HeadingTypes = "h1";
```
Parsium, line 23:
```suggestion
type HeadingTypes = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
As we are giving consumers control over setting the heading type, we should make ensure the typing of the prop reflects this

src/components/heading/heading.component.tsx
```diff
@@ -83,6 +87,8 @@ export const Heading = ({
     );
   };
 
+  console.log(headingLevel);
+
```
Parsium, line 91:
```suggestion
```
Think you may have forgotten to remove this :)

src/components/settings-row/settings-row.component.js
```diff
@@ -17,6 +17,7 @@ const marginPropTypes = filterStyledSystemMarginProps(
 
 const SettingsRow = ({
   title,
+  headingLevel = 1,
```
Parsium, line 20: Personally I would amend this prop to accept strings in the form `h1`, `h2`, ... , `h6` so its consistent with the `Heading`'s `headingLevel` prop and `Typography`'s `variant` prop.

src/components/settings-row/settings-row.spec.js
```diff
@@ -77,6 +77,25 @@ describe("SettingsRow", () => {
...
+            <SettingsRow
+              headingLevel={headingLevels}
+              title={title}
+              description={description}
+            >
+              Content for settings
+            </SettingsRow>
+          );
+
+          expect(wrapper.find(`h${headingLevels}`).exists()).toBeTruthy();
```
Parsium, line 94:
```suggestion
          expect(wrapper.find(`h${headingLevels}`).text()).toBe(title);
```
To guarantee that the heading element is the title and not some other element, you could amend the assertion like so - this would still assert if the title is rendered in the document, but should give you reassurance that you have found the right element.

src/components/settings-row/settings-row.test.js
```diff
@@ -27,6 +27,21 @@ context("Tests for SettingsRow component", () => {
...
+      (headingLevels) => {
+        CypressMountWithProviders(
+          <SettingsRowComponent
+            headingLevel={headingLevels}
+            title="foo"
+            description="bar"
+          />
+        );
+
+        cy.get(`h${headingLevels}`).should("exist");
```
Parsium, line 41:
```suggestion
        cy.get(`h${headingLevels}`).contains("foo");
```
Likewise with the jest test, this could provide some extra specificity in the assertion.

src/components/heading/heading.spec.tsx
```diff
@@ -41,6 +41,33 @@ describe("Heading", () => {
...
+
+    it.each([
+      HeadingOption1,
+      HeadingOption2,
+      HeadingOption3,
+      HeadingOption4,
+      HeadingOption5,
+      HeadingOption6,
+    ])(
+      "HTML section heading element is correct when headingType is %s",
```
Parsium, line 60: What do you think about amending the test description to the following so its clearer which element the test is referring to? This is minor so happy to accept otherwise 👍🏼 
```suggestion
      "title is rendered as the correct heading element when headingType is %s",
```

src/components/heading/heading.component.tsx
```diff
@@ -20,6 +20,7 @@ import {
 } from "./heading.style";
 import useLocale from "../../hooks/__internal__/useLocale";
 
+export type HeadingTypes = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
robinzigmond, line 23:
```suggestion
export type HeadingType = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```

minor thing but I think the name of the type should be singular - a value of this type represents a single `HeadingType`, it shouldn't be plural just because there are a range of options for values it can take. Note that this is consistent with the `IconType` type used in various components (although I haven't checked if we have the plural naming elsewhere - but I think it should be singular and `IconType` is a widely used type in Carbon).

src/components/settings-row/settings-row.component.js
```diff
@@ -17,6 +17,7 @@ const marginPropTypes = filterStyledSystemMarginProps(
 
 const SettingsRow = ({
   title,
+  headingType = "h1",
```
robinzigmond, line 20: I actually made a note in the Jira ticket to say that as well as allowing consumers to override the heading type, I thought we should change the default from `h1`. Admittedly that's just my opinion, and I recognise that it might technically be a "breaking change" to update this. But I don't think it's actually breaking in practice because I don't see how there are *any* legitimate uses of `SettingsRow` where an h1 is remotely sensible.

I think there are advantages in making the default into, say, `h4` (happy for you to consult on the best option), because even though this could still result in incorrect heading levels if consumers don't update it, it's likely to be a much less serious problem for screenreader users than if there's a random h1 in an unexpected place.

src/components/heading/heading.spec.tsx
```diff
@@ -41,6 +41,33 @@ describe("Heading", () => {
...
+      (headingTypes) => {
```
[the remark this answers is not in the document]
tomdavies73: Thanks for this, I knew there would 100% be a better way, but wasn't familiar with these approaches. Will definitely implement!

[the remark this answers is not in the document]
robinzigmond: I agree with James's suggestion - I don't see any value in using constants here for just one group of tests. Further if you do go down that route I don't see why the constant name is in UpperCamelCase, convention I believe is to use either lowerCamelCase or (probably better for constants like this) ALL_CAPS.

src/components/heading/heading.component.tsx
```diff
@@ -31,6 +32,8 @@ export interface HeadingProps extends MarginProps, TagProps {
   subheader?: React.ReactNode;
   /** Defines the subtitle id for the heading. */
   subtitleId?: string;
+  /** Defines the HTML heading element. */
+  headingType?: HeadingTypes;
```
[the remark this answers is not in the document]
tomdavies73: Yeah I agree, I'll change 👍

src/components/heading/heading.spec.tsx
tomdavies73: Yeah I agree, I'll change 👍

src/components/heading/heading.stories.mdx
```diff
@@ -42,6 +42,18 @@ import Heading from "carbon-react/lib/components/heading";
   </Story>
 </Canvas>
 
+### Heading with headingType
+
+The `headingType` prop changes the HTML heading element of the `title` string which is displayed within the component. Each `string` 
+represents a HTML Section Heading element e.g `h1` is equal to `<h1>`, allowing up to six sections of headings from most to least important.
+
```
robinzigmond, line 49: I'm not a writer - this might be worth asking our technical authors about! - but there are a few things about this paragraph that I don't like:

- I'm not sure why `string` is in code formatting. Anyway I'd personally say "each option" rather than "each string".
- "HTML Section Heading element" reads weird to me, particular with the words capitalised which make it sound like an "official" term for something. It's just an HTML heading element so I'd say that.
- "`h1` is equal to `<h1>`" - I know what you mean, and readers likely will too, but this is clumsy: the two things are not "equal".
- I really don't think we need to explain about the HTML heading hierarchy, consumers should be expected to know this, and even if they somehow don't I don't think this brief sentence adds much to our documentation
- I think "an HTML" is more correct than "a HTML" (at least it matches what I say, but I recognise this is potentially controversial so don't worry about it if you much prefer "a")

I would replace all of the above with something like:

```
The `headingType` prop sets the HTML heading element which is rendered on the page. It can take values `"h1"` up to `"h6"`.
```

or words roughly to that effect, but as I said I'm no writer so feel free to make changes or even rewrite it yourself to take account of some of my suggestions (it doesn't have to be all if you don't agree with them!).

src/components/heading/heading.stories.mdx
```diff
@@ -42,6 +42,18 @@ import Heading from "carbon-react/lib/components/heading";
...
+<Canvas>
+  <Story name="with headingType">
+    <Heading headingType="h2" title="This is a h2 Title" />
```
robinzigmond, line 52: a really minor suggestion but it might be clearer what's going on if you change the example from h2 to something more obviously contrasting with h1, such as an h4 or h5.

src/components/heading/heading.test.js
```diff
@@ -152,6 +152,19 @@ context("Testing Heading component", () => {
       }
     );
 
+    describe("when headingType prop is provided", () => {
+      it.each(["h1", "h2", "h3", "h4", "h5", "h6"])(
+        "should check HTML section heading element is correct when headingType is %s",
+        (headingTypes) => {
```
robinzigmond, line 158:
```suggestion
        (headingType) => {
```
this argument represents a single heading type so the singular name is less confusing (`headingType={headingTypes}` on the line below makes my point I feel!)

src/components/settings-row/settings-row.component.js
```diff
@@ -58,6 +60,10 @@ SettingsRow.propTypes = {
   className: PropTypes.string,
   /**  A title for this group of settings. */
   title: PropTypes.string,
+  /** Defines the HTML heading element which is applied to the `title` within the component. */
+  headingType: PropTypes.arrayOf(
```
robinzigmond, line 64: why are we making this an array in PropTypes? It should just be `headingType: PropTypes.oneOf(["h1", "h2", "h3", "h4", "h5", "h6"])` or consumers will see PropTypes errors everywhere when using this component.

Please let me know if I'm missing something obvious!

src/components/settings-row/settings-row.d.ts
```diff
@@ -1,10 +1,13 @@
 import * as React from "react";
 
+type HeadingTypes = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
robinzigmond, line 3: we should import this from the `Heading` component in my opinion, as there's no reason to suppose it will ever be different.

And if there is a good reason to redefine it here then the same comment applies as I made there: the name should be singular (`HeadingType`).

src/components/settings-row/settings-row.stories.mdx
```diff
@@ -44,6 +44,19 @@ All `children` are rendered in the input column to the right of the header colum
   </Story>
 </Canvas>
 
+### with HeadingType
+
+The `headingType` prop changes the HTML heading element of the `title` string which is displayed within the `SettingsRow` component. Each `string` 
+represents a HTML Section Heading element e.g `h1` is equal to `<h1>`, allowing up to six sections of headings from most to least important.
```
robinzigmond, line 50: same comment as for the new story in `Heading` (looks like you copy-pasted this, which is fair enough - just ensure that if you change it the new version is consistent across both components!)

src/components/settings-row/settings-row.stories.mdx
```diff
@@ -44,6 +44,19 @@ All `children` are rendered in the input column to the right of the header colum
...
+
+<Canvas>
+  <Story name="headingType">
+    <SettingsRow headingType="h2" description="Description" title="Title">
```
robinzigmond, line 54: and again I'd prefer to use a "lower" heading for this example - although not the one set as the default if you implement my suggestion to change that from `h1`

src/components/heading/heading.test.js
```diff
@@ -152,6 +152,19 @@ context("Testing Heading component", () => {
       }
     );
 
+    describe("when headingType prop is provided", () => {
+      it.each(["h1", "h2", "h3", "h4", "h5", "h6"])(
+        "should check HTML section heading element is correct when headingType is %s",
```
robinzigmond, line 157:
```suggestion
        "should check HTML heading element is correct when headingType is %s",
```

I'm not sure that the word "section" is really adding anything helpful to this description.

src/components/settings-row/settings-row.test.js
```diff
@@ -27,6 +27,21 @@ context("Tests for SettingsRow component", () => {
       }
     );
 
+    it.each(["h1", "h2", "h3", "h4", "h5", "h6"])(
+      "should check HTML section heading element is correct when headingType is %s",
```
robinzigmond, line 31:
```suggestion
      "should check HTML heading element is correct when headingType is %s",
```

As in `Heading`

src/components/typography/typography.component.tsx
```diff
@@ -11,6 +11,7 @@ const VARIANT_TYPES = [
   "h3",
   "h4",
   "h5",
+  "h6",
```
robinzigmond, line 14: this isn't a suggestion so much as a thought: I recognise we have to add this here to allow `h6` to be used for `Heading` since that uses Typography internally. But was there a reason we didn't allow `h6` for `Typography` before? It's hard to imagine there was a particularly good one, but we should try to find out and look for an alternative if we don't want to expose this on `Typography`.

src/components/settings-row/settings-row.component.js
robinzigmond: I actually prefer the numeric `HeadingLevel` ironically - but I understand the argument about consistency with `Typography`, even though that really has to use strings because it has a much wider range of variants.

But I don't feel strongly so I haven't commented on this elsewhere in the review :-)

Parsium (dismissed):

src/components/settings-row/settings-row.stories.mdx
```diff
@@ -44,6 +44,17 @@ All `children` are rendered in the input column to the right of the header colum
   </Story>
 </Canvas>
 
+### with HeadingType
+
+The `headingType` prop sets the HTML heading element of the `title` within the component which is rendered on the page. It can take values `"h1"` up to `"h5"`.
+<Canvas>
+  <Story name="headingType">
+    <SettingsRow headingType="h4" description="Description" title="This is a h4 Title">
```
robinzigmond, line 52: Since h4 is now the default we should show an example setting it to something different.

robinzigmond (dismissed):

Parsium (approved):

robinzigmond (approved):

Parsium (approved):

Parsium (approved):

robinzigmond (approved):

PR #5809, opened 2023-02-03
WIP docs(select): add aria label prop documentation (FE-5662)

### LINKED ISSUE #5783: `FilterableSelect` to support `aria-label` and `aria-labelledby`
### Desired behaviour

It is desired to be able to pass `aria-label` or `aria-labelledby` to `FilterableSelect` for the use cases like e.g. in the codesandbox attached, where the table header cell would be used as an accessible name for the `role="listbox"` element.

### Current behaviour

As of now, the only way for `FilterableSelect` to satisfy aria's _accessible names for inputs_ is to pass `label` to it, which gets rendered in the UI as input's label. Omitting it causes axe to detect a "Serious issue".

image

### Suggested Solution

Allow `aria-label` or `aria-labelledby` props on the `FilterableSelect` component and pass them to the `role="listbox"` element to render as its attributes.

https://codesandbox.io/s/thirsty-bohr-54slys?file=/src/index.js

_No response_

### Confidentiality

grabkowski:
### Proposed behaviour

add `aria-label` and `aria-labelledby` as props in `PropTypes` to all select components

### Current behaviour

in select components `aria-label` and `aria-labelledby` attributes could be passed down to the input, but props for it are not documented 

### Checklist

#### QA

### Additional context

closes #5783

### Testing instructions

props for those ARIA attributes should be displayed in all select stories


src/components/select/select-textbox/select-textbox.component.js
```diff
@@ -116,13 +118,16 @@ const SelectTextbox = React.forwardRef(
...
+        ? joinIds(
+            ariaLabelledBy || labelId,
+            accessibilityLabelId || textId.current
+          )
+        : joinIds(ariaLabelledBy || labelId, textId.current);
 
       return {
         "aria-expanded": readOnly ? undefined : isOpen,
-        "aria-labelledby": ariaLabelledby || undefined,
+        "aria-labelledby": ariaLabel ? ariaLabelledBy : combinedAriaLabelledBy,
```
robinzigmond, line 130: I've spent a ridiculously long time thinking about this line - I found it very confusing to understand, and on further thought I don't think it's correct in the edge case of both `ariaLabel` and `ariaLabelledBy` props being provided by the consumer.

I think the correct logic is `(ariaLabel && !ariaLabelledBy) ? undefined : combinedAriaLabelledBy` - this is only different from yours in the case where both props are provided, which should be rare if it happens at all, but in that case surely we should use the combined label (which already takes `ariaLabelledBy` into account). I also think this makes the logic clearer - use the computed "combined label" except when an `ariaLabel` prop is provided and not overridden by `ariaLabelledBy`.

src/components/select/select-textbox/select-textbox.spec.js
```diff
@@ -172,38 +172,84 @@ describe("SelectTextbox", () => {
...
+                expect.stringContaining(mockAccessibilityLabelId)
+              );
+            });
+          });
         });
       }
     );
+
+    describe("when the ariaLabelledBy and the ariaLabel prop are set", () => {
+      it("then the ariaLabelledBy prop value should be passed down to the aria-labelledby prop of the Textbox", () => {
```
robinzigmond, line 243: as I explained in the component file, I don't think this behaviour being tested for is quite correct. If you agree with tha tcomment you can just check that the `aria-labelledby` prop contains the `mockAriaLabelledBy` (rather than being equal to it). Or even perhaps remove this test entirely, as both the `aria-label` and `aria-labelledby` attributes should be being covered, separately, by the tests above.

DipperTheDan (commented):
Nothing more to add from me. Just address the comments left by @robinzigmond please.

src/components/select/select-textbox/select-textbox.component.js
grabkowski: my idea was to allow users to completely overwrite the `aria-labelledby` attribute and let them create their own descriptions. I can see that this might be a bit confusing, so I'll just change that to have the combined id's only.

src/components/select/select-textbox/select-textbox.spec.js
grabkowski: this is the expected behaviour,  aria-labelledby in this case should be what user defined, not a combination that is usually passed down from the SelectTextbox, so the attribute should have the same value as the mock, not just contain it.

robinzigmond (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #5814, opened 2023-02-07
fix(box): ensure z-index is set when sticky or fixed positioned

### LINKED ISSUE #5572: inputs overlap fixed positions
### Current behaviour

All kind of inputs (radio buttons, textbox, select...) overlap a box with the `position: sticky` rule, as they have a z-index of 999, while the box doesn't have a custom z-index.

image

### Expected behaviour

For the inputs to go under the sticky box, as the rest of the text.

https://codesandbox.io/s/sidebar-with-sticky-content-forked-xgyefg?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-44457

### Suggested Solution

_No response_

### Carbon Version

111.4.2

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5572

### Proposed behaviour

Ensures that the any `Box` container that has `sticky` or `fixed` positioning has a z-index set to prevent it being overlaid by other elements that are relatively positioned

### Current behaviour

Any `Box` container that has `sticky` or `fixed` positioning has no z-index set

### Checklist

#### QA

### Testing instructions

See sandbox in linked issue below

### DIFF
--- a/src/components/box/box.spec.tsx
+++ b/src/components/box/box.spec.tsx
@@ -14,6 +14,7 @@ import Box, {
   BoxSizing,
   AllowedNumericalValues,
   Gap,
+  BoxProps,
 } from "./box.component";
 import boxConfig from "./box.config";
 
@@ -119,4 +120,30 @@ describe("Box", () => {
       wrapper
     );
   });
+
+  it.each<BoxProps["position"]>(["fixed", "sticky"])(
+    "sets zIndex to 1 when position is %s",
+    (position) => {
+      const wrapper = mount(<Box position={position} />);
+      assertStyleMatch(
+        {
+          zIndex: "1",
+        },
+        wrapper
+      );
+    }
+  );
+
+  it.each<BoxProps["position"]>(["absolute", "relative"])(
+    "does not set zIndex when position is %s",
+    (position) => {
+      const wrapper = mount(<Box position={position} />);
+      assertStyleMatch(
+        {
+          zIndex: undefined,
+        },
+        wrapper
+      );
+    }
+  );
 });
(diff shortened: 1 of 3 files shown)


src/components/box/box.spec.tsx
```diff
@@ -119,4 +120,17 @@ describe("Box", () => {
       wrapper
     );
   });
+
+  it.each<BoxProps["position"]>(["fixed", "sticky"])(
+    "sets zIndex to 1 when position is %s",
```
DipperTheDan, line 125: Is it worth doing a test that covers the undefined path of this too? Might be overkill but I figured I'd ask the question anyway.

Parsium (commented):
Looks good to me 👍🏼 Happy to accept once @DipperTheDan's query has been addressed :)

src/components/box/box.spec.tsx
edleeks87: I feel it's somewhat redundant to assert the zIndex is undefined when the position is undefined if that's what you mean?

DipperTheDan: Not quite. I was referring to if the position css value is `absolute` for example then asserting that the zIndex is `undefined`. Just something that explores the `false` path of the ternary statement.

DipperTheDan (approved):

src/components/box/box.spec.tsx
edleeks87: Those paths are implicitly covered in other test cases but I think there's some value in having an explicit assertion so will add

Parsium (approved):

divyajindel (approved):

PR #5817, opened 2023-02-07
Focus trap fixes

### LINKED ISSUE #5749: Tabbing/focus trap issues persist with editable AG Grid.
### Current behaviour

Follow up from:
https://github.com/Sage/carbon/issues/5544

When this issue was originally logged, we re-produced it in a sandbox with minimal code. As far as that goes, the fix worked.

However, AG grid is mostly used for inline editing and if you enable that in the same sandbox, the focus trap causes further issues with tabbing.

I've modified the original sandbox here:
https://codesandbox.io/s/exciting-dew-tz8pml?file=/src/index.js

I've made the cells editable and changed the carbon-react version to an arbitrary one form the past (75.2.5),

If you double-click the cell with "Toyota" in there and press TAB, you can see the correct behaviour, where it will tab to the next editable cell (in edit mode). 

If you then update the carbon version to the one with this fix (113.0.1) and try the above steps, it will no longer tab to the next editable cell. 

### Expected behaviour

Tabbing in AG Grid should still work as expected

https://codesandbox.io/s/exciting-dew-tz8pml?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

113.0.1

### Design Tokens Version

_No response_

Chrome

Windows

_No response_

### Confidentiality

### LINKED ISSUE #5754: nested modals not treating focus correctly after click on non-focusable content
### Current behaviour

On the attached codesandbox:

- open the full-screen dialog
- open the nested dialog
- click on any non-focusable content (in either child dialog or full-screen one - it doesn't mater)

Now focus has been lost to the document body (which is inevitable here), but pressing tab cycles through the focusable elements in the full-screen dialog, and never reaches the nested dialog.

[Note: I have been told that the issue isn't observable to some people. That's not too surprising as it's basically caused by a race condition I believe - but it always seems to "work" for me!]

Note that the exact types of modal component don't seem to matter - the original bug relates to a full-screen dialog opening on top of a sidebar.

### Expected behaviour

Focus should only ever be on elements in the currently "active" dialog and should return there on pressing tab if lost to the body.

https://codesandbox.io/s/intelligent-gianmarco-7ivt11?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-48145

### Suggested Solution

In `getNextElement` in `focus-trap-utils.js` we check for whether focus is currently on the document body and return the first active element (which the calling code then focuses on). When there are multiple modals open this code is called for each modal and, depending on the order the event handlers fire in, may leave focus in the wrong modal, as has been observed.

I think the fix is to make use of the `TopModalContext` that was introduced for other purposes and return `undefined` from this function if the modal concerned is not the top one.

### Carbon Version

113.0.3

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
### Proposed behaviour

Fixes 2 separate issues within the FocusTrap internal component (which is used by Dialog and related components to trap focus inside when tabbing):

1) Carbon should as far as possible not interfere with the behaviour of any 3rd-party libraries which the consumer is using that, by not calling `preventDefault` on the keyboard event (see first paragraph of "current behaviour" for details). This has been achieved by relying on focusable "tab guard" elements before and after the focus trap which "redirect" the focus as needed when focused, and otherwise relying on browser default behaviour (as modified by any other libraries in use) except when this would result in incorrect behaviour in Carbon (which happens in only one situation - when the consumer is using a custom `focusableSelectors` string - and even there the PR avoids calling `preventDefault` except when absolutely necessary). (fixes #5749 )

2) when multiple dialogs are open and a non-focusable element is clicked, resulting in focus being lost to the document body, tabbing should always result in focus going to the first focusable element of the top modal (fixes #5754 )

### Current behaviour

1) The current implementation relies on calling the `preventDefault` method on the keyboard event whenever the user tabs, in order to work out programatically the next element to focus. This breaks any third-party libraries - such as AG Grid - which implement their own behaviour on tab press and check that the default wasn't prevented before doing anything (thereby effectively making their implementation a new default, as calling `preventDefault` stops the library code from working as intended).

2) when multiple dialogs are open and a non-focusable element is clicked, resulting in focus being lost to the document body, tabbing can lead to a race condition between more than one focus trap both trying to focus their own first element. This leads to the possibility that the wrong element may be focused, as it is only the "top" modal that should get the focus on its first element.

### Checklist

#### QA

### Testing instructions

Ensure both linked issues have been fixed.

Also regression test all components that use FocusTrap - Dialog, DialogFullScree, Sidebar, MenuFullScreen and VerticalMenuFullScreen

### DIFF
--- a/cypress/locators/menu/index.js
+++ b/cypress/locators/menu/index.js
@@ -24,6 +24,7 @@ export const menuComponent = (index) =>
 export const submenuItem = (index) =>
   menuComponent(index).find(SUBMENU).find("ul > li");
 export const menuCanvas = () => cy.get(DLS_ROOT);
+export const fullScreenMenuWrapper = () => cy.get(FULLSCREEN_MENU);
 export const fullscreenMenu = (index) =>
   cy.get(FULLSCREEN_MENU).find("div").eq(index);
 export const fullScreenMenuItem = (index) =>
--- a/src/__internal__/focus-trap/focus-trap-utils.ts
+++ b/src/__internal__/focus-trap/focus-trap-utils.ts
@@ -1,3 +1,7 @@
+type CustomRefObject<T> = {
+  current?: T | null;
+};
+
 const defaultFocusableSelectors =
   'button:not([disabled]), [href], input:not([type="hidden"]):not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]';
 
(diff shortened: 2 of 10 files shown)


edleeks87 (commented):
Not much to comment on past the one Will left (happy for mine to be ignored if you choose to).

I noticed some behaviour that looked odd when testing this out in the sandbox (see gif below): it looks like when you edit a cell the checkboxes are no longer reachable via tab until you leave the gird. I'm not sure if it's anything we've done or if that's how AG-Grid behaves but figured I'd raise it just in case you weren't aware of it

ag-ft-issue

robinzigmond:
> 1. Tabbing/focus trap issues persist with editable AG Grid. #5749

thanks for this Ed - I've taken a look and while my memories of AG Grid are a little hazy, I'm pretty sure this is intended behaviour. When in "edit mode" tab is only supposed to take you through the editable cells, those checkbox cells in the first column don't have `editable` set to true so it's expected they're skipped over.

Even if I'm wrong, I've confirmed that the sandbox from the original issue behaves the same way so at worst this will just be something I've missed (and that wasn't explicitly mentioned in the bug report), not something new we're introducing.

edleeks87:
> > 1. Tabbing/focus trap issues persist with editable AG Grid. [Tabbing/focus trap issues persist with editable AG Grid. #5749](https://github.com/Sage/carbon/issues/5749)
> 
> thanks for this Ed - I've taken a look and while my memories of AG Grid are a little hazy, I'm pretty sure this is intended behaviour. When in "edit mode" tab is only supposed to take you through the editable cells, those checkbox cells in the first column don't have `editable` set to true so it's expected they're skipped over.
> 
> Even if I'm wrong, I've confirmed that the sandbox from the original issue behaves the same way so at worst this will just be something I've missed (and that wasn't explicitly mentioned in the bug report), not something new we're introducing.

Thanks @robinzigmond I'm happy to defer to you on it as it does seem to behave that way in the original issue sandbox

edleeks87 (approved):

nineteen88 (dismissed):

DlgSHi (approved):

edleeks87 (approved):

PR #5826, opened 2023-02-09
fix(link): change display to inline so that text can wrap

### LINKED ISSUE #5774: Allow Link component to display link text inline instead of wrapping onto a new line
### Current behaviour

Link component displays link text in a new line when it is possible to wrap and display link text inline.

### Expected behaviour

Link component should display link inline.

https://codesandbox.io/s/confirm-modal-wv8qgn?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-50071

### Suggested Solution

Update use of `display: inline-block` on both the actual anchor element and the span that Carbon wraps it in.

### Carbon Version

111.17.0

### Design Tokens Version

2.60.0

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
Link components currently use `display: inline-block` for both the anchor element itself and one of the containing components. This means that when the text of a link is long enough to wrap onto a new line, the entire link element is moved to the next line, creating unnecessary blank space. This PR fixes that by removing the `display: inline-block`, except when used in a MenuItem - that component will never be used inline and the inline-block display is needed for the background color on hover to fill the entire vertical space.

fix #5774

### Proposed behaviour

Links should allow text to wrap in the middle of the component's text.

### Current behaviour

If the full link text does not fit in the available space, the entire link is moved onto a new line.

### Checklist

#### QA

### Testing instructions

Note - this sandbox has a slight difference from the one in the Github issue, in that an additional space has been added (rather than just a newline) before the <Link> component. Without this, the bug is still fixed but there is no space between the link text and the previous word. https://codesandbox.io/s/confirm-modal-forked-09n7p7?file=/src/index.js

Also test any relevant stories for components that use Link internally, to ensure that nothing has changed that shouldn't have. The following components are the ones that use Link:
- Heading (if `backLink` prop is used)
- Menu (for all menu items - including those that open submenus, as well as the individual items in submenus)
- Pager (the "links" for moving between pages)


tomdavies73:
Just a question regarding this becoming the new default behaviour, as a small percentage of consumers may be intentionally using `Link` with the previous "incorrect behaviour" as it has been present for a long time, therefore this PR may and could create a visual change for some. This is definitely not a breaking change, but it would be interesting to see what another reviewer thinks. Perhaps an appropriately named prop which removes `display: inline-block` could be a good approach, so consumers have to opt in to the new behaviour. Just a thought, but happy to approve if we feel it's not that big of an issue 👍

grabkowski:
looks good to me, I agree with @tomdavies73  that this should be a breaking change

grabkowski (commented):
looks good to me, I agree with @tomdavies73 that this should be a breaking change

tomdavies73 (approved):

grabkowski (dismissed):

ZhuoyuJin (dismissed):

grabkowski (approved):

ZhuoyuJin (approved):

PR #5835, opened 2023-02-13
fix: resolve issue with multiple execution of onOpen and onFilterChange callbacks - FE-5612

### LINKED ISSUE #5719: FilterableSelect calls onOpen without user interaction
### Current behaviour

Reproduced on codesandbox.

`FilterableSelect` does not have the same behaviour as `Select` when providing the exact same props.

`onOpen` can be repeatedly called without any user input if some props are modified.

### Expected behaviour

`onOpen` is only called when the user clicks on a `FilterableSelect` component.

https://codesandbox.io/s/lucid-waterfall-8thv8k?file=/src/LoadingSelect.tsx:897-909

### JIRA Ticket (Sage Only)

SBS-47539

### Suggested Solution

quote from @robinzigmond :

"
On initial investigation the cause is this useEffect in FilterableSelect:

```
useEffect(() => {
  if (!isOpen) {
    setFilterText("");
  } else if (onOpen) {
    onOpen();
  }
}, [isOpen, onOpen]);
```
    
In your sandbox example, this actually fires on every render because onOpen (your loadItems) is not memoised in any way. 

I've been attempting to fix this with useCallback but I don't think it's entirely solvable within your example, because the callback in your case updates remainingFailedRequests which is impossible to get rid of as a dependency of getItems and therefore of loadItems.

As you observe, simple Select does not have this problem (because it has no similar useEffect) so we should investigate how we might be able to achieve the desired effect in FilterableSelect without running onOpen every time its dependencies change. 
"

### Carbon Version

111.9.1

### Design Tokens Version

_No response_

Firefox

MacOS

111.3.0 on codesandbox but happened originally with 111.9.1

### Confidentiality

mkrds:
### Proposed behaviour

Fixes #5719

This PR also contains a fix for a similar issue happening with the `onFilterChange` callback in both `FilterableSelect` and `MultiSelect`

### Current behaviour

MultiSelect, FilterableSelect: 
The onFilterChange callback function is intended to be triggered only when the filter values are changed by the user. However, onFilterChange callback is being called on the first render of the component and then on each rerender. 

FilterableSelect:
The onOpen callback function is intended to be triggered only when the select list is being opened. Currently though onOpen callback is being called on each rerender of the component.

### Checklist

#### QA

### Testing instructions

Issue codesandbox:
https://codesandbox.io/s/elegant-dream-frnvwm

Codesandbox to test that neither `onChange` nor `onOpen` are triggered on first render or on any subsequential rerenders
https://codesandbox.io/s/carbon-quickstart-typescript-forked-bi1hks


src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -308,11 +310,14 @@ const FilterableSelect = React.forwardRef(
...
 
+    const onFilterChange = useStableCallback(onFilterChangeProp);
+    const isFirstRender = useRef(true);
     useEffect(() => {
-      if (onFilterChange) {
+      if (onFilterChange && !isFirstRender.current) {
         onFilterChange(filterText);
       }
-    }, [filterText, onFilterChange]);
+      isFirstRender.current = false;
```
robinzigmond, line 319: I'd rather keep the update of the `isFirstRender` ref in its own `useEffect` that's not dealing with anything else. It shouldn't change anything about either the component's current behaviour or how we might want to use `isFirstRender` in future, but it would keep each `useEffect` doing separate things (and ensure that in the unlikely event of us removing the `onFilterChange` functionality we don't accidentally forget to update the ref).

src/components/select/filterable-select/filterable-select.spec.js
```diff
@@ -947,6 +947,35 @@ describe("FilterableSelect", () => {
         .simulate("change", { target: { value: filterText } });
       expect(onFilterChangeFn).toHaveBeenCalledWith(filterText);
     });
+
+    it("then it should not be invoked when the component rerenders", () => {
+      const Component = () => {
+        const [onFilterChangeCalled, setOnFitlerChangeCalled] = useState(false);
```
robinzigmond, line 953:
```suggestion
        const [onFilterChangeCalled, setOnFilterChangeCalled] = useState(false);
```

spelling

src/components/select/filterable-select/filterable-select.spec.js
```diff
@@ -947,6 +947,35 @@ describe("FilterableSelect", () => {
...
+              <Option value="opt2" text="green" />
+              <Option value="opt3" text="blue" />
+              <Option value="opt4" text="black" />
+            </FilterableSelect>
+          </>
+        );
+      };
+
+      const wrapper = mount(<Component />);
+      expect(wrapper.find("#on-filter-called").text()).toBe("false");
```
robinzigmond, line 975: not criticising the approach, but I'm curious why you've made it so that calling `onFilterChange` updates the component text and testing for that, rather than using a mock and checking that's not been called, which is our usual approach? They're both testing for the same thing, just curious why you've adopted a different approach here from what we do elsewhere.

src/components/select/filterable-select/filterable-select.spec.js
```diff
@@ -1004,6 +1033,40 @@ describe("FilterableSelect", () => {
...
+              </>
+            );
+          };
+
+          wrapper = mount(<Component />);
+          expect(wrapper.find("#call-counter").text()).toBe("0");
+          wrapper.find("input").simulate("focus");
+          expect(wrapper.find("#call-counter").text()).toBe("1");
+          wrapper.setProps({});
+          expect(wrapper.find("#call-counter").text()).toBe("1");
```
robinzigmond, line 1067: similar here to my question about the other test - why use state and DOM updates to track the number of calls when it can be done directly with `jest.fn()`?

src/components/select/multi-select/multi-select.component.js
```diff
@@ -326,11 +327,14 @@ const MultiSelect = React.forwardRef(
       };
     }, [handleGlobalClick]);
 
+    const onFilterChange = useStableCallback(onFilterChangeProp);
+    const isFirstRender = useRef(true);
```
robinzigmond, line 331: since you've now introduced this `isFirstRender` to two separate components, and I just checked and it's already used in 3 others, it's probably worth extracting to a new custom hook so we can just call `const isFirstRender = useIsFirstRender()` inside any component that needs it. (And this could return the actual bool rather than the ref wrapping it.)

Personally I think that's worth doing as part of this PR, especially as you're already introducing a new custom hook - but I don't mind if we keep this PR as it is and create a new task to refactor this.

src/components/select/filterable-select/filterable-select.component.js
```diff
@@ -281,13 +282,14 @@ const FilterableSelect = React.forwardRef(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [value, onChange, children]);
 
+    const onOpen = useStableCallback(onOpenProp);
```
robinzigmond, line 285: I was wondering why you hadn't used `useStableCallback` on `onOpen` in multi-select - and looking at the code I see why, it already handles this by using a callback argument to `setOpenState` to ensure we don't call `onOpen` if it's already open. I'm wondering if you considered using that approach in filterable-select too? It seems odd to me that these two components which are so similar are using completely different ways to control the calling of `onOpen`, and this may have been an opportunity to make them more consistent with each other. I think both approaches are fine, and the `useStableCallback` is probably simpler to write, but really I think we should use the same one in both components.

src/components/select/multi-select/multi-select.component.js
```diff
@@ -326,11 +327,14 @@ const MultiSelect = React.forwardRef(
...
     useEffect(() => {
-      if (onFilterChange) {
+      if (onFilterChange && !isFirstRender.current) {
         onFilterChange(filterText);
       }
-    }, [filterText, onFilterChange]);
+      isFirstRender.current = false;
```
robinzigmond, line 336: as in filterable-select, if we don't extract this to a custom hook I would much prefer this update is in its own `useEffect`

src/components/select/multi-select/multi-select.spec.js
```diff
@@ -816,6 +816,35 @@ describe("MultiSelect", () => {
...
+              <Option value="opt4" text="black" />
+            </MultiSelect>
+          </>
+        );
+      };
+
+      const wrapper = mount(<Component />);
+      expect(wrapper.find("#on-filter-called").text()).toBe("false");
+      wrapper.setProps({});
+      expect(wrapper.find("#on-filter-called").text()).toBe("false");
```
robinzigmond, line 846: same question as in the new tests for filterable-select

src/components/select/filterable-select/filterable-select.component.js
mkrds: Using the same pattern as in multi-select would require creating a new method which then should be added as a dependency to other methods, meaning I'd have to wrap it in `useStableCallback` either way to prevent some unnecessary renders or method executions. Its obvious that it is possible to implement it in a cleaner way (like extracting some of the methods to a separate `useSelect` hook to reduce redundancy) but I'd rather do it in a separate dedicated `Refactor Select components` task.

src/components/select/multi-select/multi-select.component.js
mkrds: Acutally I change my mind, I've implemented this hook in 2 different ways and each of these implementations has serious drawbacks when compared to just writing that directly in the component.

1. https://usehooks-ts.com/react-hook/use-is-first-render - this solution works as long as you don't use the result in another `useEffect` hook. Reason for that is the fact that the `useEffect` hook inside the `useIsFirstRender` is invoked first, meaning that the returned boolean when referenced in the outside `useEffect` is always `false`

2. Similar implementation as above, but using `useState` - works fine, but introduces unnecessary rerender

src/components/select/filterable-select/filterable-select.spec.js
robinzigmond: oh I see. Sorry I should have thought of that!

Parsium (changes requested):

src/hooks/__internal__/useStableCallback/useStableCallback.ts
```diff
@@ -0,0 +1,16 @@
+import { useRef, useCallback } from "react";
+
+// This is a hook that returns a memoized callback that is guaranteed to be the same reference between renders.
+// This is useful when a stable reference is required to prevent unnecessary re-renders.
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function useStableCallback(callback?: (...args: any[]) => any) {
```
Parsium, line 7:
```suggestion
export function useStableCallback(callback?: (...args: unknown[]) => unknown) {
```
What do you think of the above so we don't have to disable eslint?

src/components/select/filterable-select/filterable-select.spec.js
```diff
@@ -1004,6 +1033,40 @@ describe("FilterableSelect", () => {
...
+                <div id="call-counter">{callCounter}</div>
+                <FilterableSelect
+                  name="testSelect"
+                  id="testSelect"
+                  openOnFocus
+                  onOpen={() => {
+                    if (callCounter < 2) {
+                      setCallCounter(callCounter + 1);
+                    }
+                  }}
```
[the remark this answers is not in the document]
mkrds: Without this condition the component goes into an infinite rerender loop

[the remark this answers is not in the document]
Parsium: Oops I overlooked that the component will re-render when the div text changes 🤦🏼 Though I still don't think that this test is verifying the `onOpen` callback hasn't been invoked again because the changed incremented counter text will trigger a re-render anyway.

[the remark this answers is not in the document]
mkrds: I forgot to mention that it actually ended up in an infinite loop before my fix. So I've left this guard to ensure that if someone in the future breaks it then they get an information about which unit test fails instead of a vague `Maximum call stack size exceeded...` message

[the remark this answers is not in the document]
Parsium: Ah thanks for clarifying this @mkrds 👍🏼 If this test no longer goes into an infinite loop after your fix, what do you think about adding a comment instead of the condition, or changing the condition to `if (callCounter < 3)`? 

That way we can safely assume if the rendered counter text is 2, then `onOpen` has been incorrectly called again after the re-render

Parsium (dismissed):

src/components/select/filterable-select/filterable-select.spec.js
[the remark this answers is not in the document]
Parsium: Realised the text can become 2 and isn't blocked by the condition because its a less-than condition 🤦🏼 This makes my comment irrelevant so happy to accept 👍🏼

robinzigmond (approved):

Parsium (approved):

DlgSHi (approved):

PR #5836, opened 2023-02-14
fix(multi-action-button, split-button): open menu on click

### LINKED ISSUE #5615: Multi Action Button - Accessibility(Speech Recognition) reinforcement request
### Desired behaviour

 For the users who are not able to click using a mouse or a keyboard, **Multi Action Button** should show hidden buttons with onClick event not only with onMouseLeave or Tab key event.

### Current behaviour

**Multi Action Button** currently shows hidden buttons only with mouseover or keyboard events. The users with hand disability should go through one more step, saying 'press enter key' and this is not intuitive. (Video attached) 

https://user-images.githubusercontent.com/90848080/202129174-6bbe9a42-494e-4d17-9896-b347c8f82a4a.mov

### Suggested Solution

 Controlling the state of the hidden buttons under **Multi Action Button** with onClick event along with the current mouseover and keyboard will reinforce the accessibility. 

https://codesandbox.io/s/speech-recognition-issue-8g7v3z?file=/src/index.js

_No response_

### Confidentiality

robinzigmond:
The additional buttons are now shown on click of the toggle button, rather than just on hover. This means that Voice Control users can activate the button by just saying its name (or the appropriate number if going through "show numbers"). Also tidy up by removing aria-haspopup on the surrouding div, this was already on the toggle button itself and from ARIA 1.2 is no longer valid on generic elements like a div.

fix #5615

### Proposed behaviour

There should be a click action on the main button of MultiActionButton, and the toggle button of SplitButton, that opens the child buttons. This enables Voice Control users to open the additional buttons without further instructions needed (see video on the original issue).

The div wrapper of both components should not have aria-haspopup=true, as it is only the (toggle) button that opens the "menu" of additional buttons, and although it's legal as of ARIA 1.1 to have this attribute on any element (and therefore flags no issue in Axe), this is going away in ARIA 1.2 - see https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup and the third "NOTE" green box: "This property is being deprecated as a global property in ARIA 1.2. In future versions it will only be allowed on roles where it is specifically supported." (`button` is listed as one of the allowed roles).

### Current behaviour

The additional buttons open on mouseenter, but not on click.

The `aria-haspopup` property is applied to the wrapping div as well as to the button that actually opens the popup.

### Checklist

#### QA

### Testing instructions

Test that the issue with Voice Control shown in the video attached to the linked issue has been fixed. Test with SpitButton as well as MultiActionButton (as both had the same problem).

Ensure that accessibility has not been impacted - that there are no new Axe errors on either component and there are no issues using them with Voiceover on Safari or NVDA with supported browsers on Windows, that were not present before.

### DIFF
--- a/src/components/multi-action-button/__snapshots__/multi-action-button.spec.tsx.snap
+++ b/src/components/multi-action-button/__snapshots__/multi-action-button.spec.tsx.snap
@@ -358,10 +358,8 @@ exports[`MultiActionButton when rendered should match the snapshot 1`] = `
 }
 
 <div
-  aria-haspopup="true"
   className="c0"
   data-component="multi-action-button"
-  onClick={[Function]}
   onMouseLeave={[Function]}
 >
   <button
@@ -373,6 +371,7 @@ exports[`MultiActionButton when rendered should match the snapshot 1`] = `
     data-element="toggle-button"
     disabled={false}
     draggable={false}
+    onClick={[Function]}
     onKeyDown={[Function]}
     onMouseEnter={[Function]}
     onTouchStart={[Function]}
--- a/src/components/multi-action-button/multi-action-button.component.tsx
+++ b/src/components/multi-action-button/multi-action-button.component.tsx
@@ -34,6 +34,7 @@ export const MultiActionButton = ({
   text,
   subtext,
   width,
+  onClick,
   "data-element": dataElement,
   "data-role": dataRole,
   ...rest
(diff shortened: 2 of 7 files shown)


edleeks87 (dismissed):

tomdavies73 (approved):

sianford (approved):
Tested in Safari and Chrome on macOS and Chrome, Edge and Firefox in Windows 10. VoiceControl works without issue in Chrome on macOS but would not allocate numbers to instruct voice actions in Safari on macOS for these components. Voice activation works correctly in Windows. Testing passed as we cannot fix a Safari-specific issue.

edleeks87 (dismissed):

tomdavies73 (dismissed):

DlgSHi (dismissed):

edleeks87 (approved):

tomdavies73 (approved):

DlgSHi (approved):

PR #5837, opened 2023-02-14
feat(password): add new component

### LINKED ISSUE #5614: Add a new input field for passwords
### Desired behaviour

Have an input component that can be used as a type="password" input.

### Current behaviour

There is currently no way using Sage Carbon Library to create a input of type="password".
Using Carbon Library one can achieve this by using the TextInput component that as the "type" prop to be passed.

The best I can achieve using Sage Carbon is by using the TextBox component which lets me costumize my input as I want but doesn't allow to pass any prop of "type" to the input.

However, if you force the type="password" prop to be passed on the TextBox component it will have effect when the input is rendered in the browser. Typescript is of course not happy with me passing the type="password" prop when it doesn't exist in the component.

### Suggested Solution

Create a specific password component that allows functionallity like icons and clicking the icons to change the type from password to text so the user can see what they have inputed. 

OR

Add the type prop to the Textbox component since it already seems to accept it internally but it is not on the accepted props.

_No response_

Question was asked on Sage Slack channels and it was answered by James Parslow.

### Confidentiality

tomdavies73:
fix #5614

### Proposed behaviour

Screenshot 2023-04-18 at 10 19 01

This pull request aims to add a brand-new component into Carbon named `Password`, the component will render `Textbox` but provide further capacity for obscured text control. The component will also render `ButtonMinor` as a means to change the input type, as well as built-in approaches and attributes to ensure the component is accessible and secure. 

#### Changing type

Screenshot 2023-04-05 at 17 20 54

The `Password` component will surface the HTML input type attribute on `Textbox` but limit the types to be either `text` or `password`, with the latter being the default. The component will also render the `ButtonMinor` carbon component to change type back and forth, allowing users to view and obscure their password when needed. The `ButtonMinor` will utilise a label of either `Show` and `Hide` and icons `View` and `Hide` depending on the current type state of the input. Also, the `buttonType` of `ButtonMinor` will be `tertiary`, with some slight amendments to remove all hover styling changes.

#### Keeping things accessible

A hidden `p` tag element has been created but accessibly hidden, to ensure its contents are still in the accessibility tree and accessible via assistive technology but not visible to users. Within this `p` tag is a string, which will update based on the input type being either `text` or `password`. This `p` tag has also been set as an `aria-live` region, so if the contents changes, assistive technology will recognise this and inform users. 

This ensures users using assistive technology will be informed that they have either shown/hidden their password, so they can focus on the input and have the password read to them when it is safe to do so. This is a much better approach than placing the `aria-live` region on the input itself, as the password can be read out automatically without any warning.

We have also used the `aria-controls` attribute on the input and `ButtonMinor`, so there is a clear association between the two inputs.

#### Keeping things secure

We have also disabled autocompletion on the input, as an initial layer of protection against a user's password accidentally being remembered by a browser. However, we have also placed a reminder within our documentation that consumers must ensure the input type is switched back to `password` on form submission, consumers can achieve this with the `forceObscurity` prop.

#### Next steps

This pull request has been created as an established platform to receive feedback regarding the quality of the code and approach taken, as well as feedback from the Design System team and any relevant accessibility team.

### Current behaviour

Currently, there is no established method for Carbon consumers to provide end users with a dedicated component for password input, as the `Textbox` component does not currently have any supported mechanism to change the input type.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/password/password.component.tsx
+++ b/src/components/password/password.component.tsx
@@ -1,0 +1,62 @@
+import React, { useRef, useState } from "react";
+import { TextboxProps } from "../textbox";
+import { StyledPassword, HiddenAriaLive } from "./password.style";
+import guid from "../../__internal__/utils/helpers/guid";
+import tagComponent, {
+  TagProps,
+} from "../../__internal__/utils/helpers/tags/tags";
+import useLocale from "../../hooks/__internal__/useLocale";
+import ButtonMinor from "../button-minor/button-minor.component";
+
+export interface PasswordProps extends TextboxProps, TagProps {
+  /** When `true` input `type` is `password` and text is obscured. */
+  forceObscurity?: boolean;
+}
+
+export const Password = ({
+  id,
+  disabled,
+  forceObscurity = false,
+  ...rest
+}: PasswordProps) => {
+  const internalInputId = useRef(id || guid());
+  const l = useLocale();
+
+  const [passwordShown, setPasswordShown] = useState(false);
+  const visibleInput = !forceObscurity && passwordShown;
+  return (
+    <>
+      <StyledPassword
+        {...tagComponent("password", rest)}
+        data-element="styled-password-container"
+        id={internalInputId.current}
+        autoComplete="off"
+        type={visibleInput ? "text" : "password"}
+        disabled={disabled}
+        {...rest}
+      >
+        <ButtonMinor
+          aria-label={visibleInput ? "Hide password" : "Show password"}
+          aria-controls={internalInputId.current}
+          onClick={() => setPasswordShown(!passwordShown)}
+          pr={1}
+          buttonType="tertiary"
+          iconType={visibleInput ? "hide" : "view"}
+          iconPosition="before"
+          size="small"
+          disabled={forceObscurity || disabled}
+          isInPassword
+        >
+          {visibleInput ? "Hide" : "Show"}
+        </ButtonMinor>
+      </StyledPassword>
+      <HiddenAriaLive role="status" aria-live="polite">
+        {visibleInput
+          ? l.password.ariaLiveShownMessage()
+          : l.password.ariaLiveHiddenMessage()}
+      </HiddenAriaLive>
+    </>
+  );
+};
+
+export default Password;
--- a/src/components/password/password.spec.tsx
+++ b/src/components/password/password.spec.tsx
@@ -1,0 +1,260 @@
+import React from "react";
+import { mount, ReactWrapper } from "enzyme";
+import { act } from "react-dom/test-utils";
+import Password, { PasswordProps } from "./password.component";
+import { assertStyleMatch } from "../../__spec_helper__/test-utils";
+import { HiddenAriaLive } from "./password.style";
+import StyledButtonMinor from "../button-minor/button-minor.style";
+import Label from "../../__internal__/label";
+import Textbox from "../textbox";
+import ButtonMinor from "../button-minor";
+import guid from "../../__internal__/utils/helpers/guid";
+import StyledInput from "../../__internal__/input/input.style";
+
+const mockedGuid = "guid-12345";
+jest.mock("../../__internal__/utils/helpers/guid");
+
+(guid as jest.MockedFunction<typeof guid>).mockImplementation(() => mockedGuid);
+
+function renderPasswordInput(props: PasswordProps) {
+  return mount(<Password {...props} />);
+}
+
+describe("Password Input", () => {
+  let wrapper: ReactWrapper;
+  let input: ReactWrapper;
+
+  const mockEvent = {
+    nativeEvent: {
+      stopImmediatePropagation: () => {},
+    },
+  };
+
+  describe("when rendered", () => {
+    it("should have the Textbox component as it's child", () => {
+      wrapper = renderPasswordInput({});
+      expect(wrapper.find(Textbox)).toHaveLength(1);
+    });
+    it("should have the ButtonMinor component as it's child", () => {
+      wrapper = renderPasswordInput({});
+      expect(wrapper.find(ButtonMinor)).toHaveLength(1);
+    });
+    it("renders with the expected border radius styling on input", () => {
+      assertStyleMatch(
+        {
+          borderRadius: "var(--borderRadius050)",
+        },
+        mount(<Password />).find(StyledInput)
+      );
+    });
+  });
+
+  describe("required", () => {
+    beforeAll(() => {
+      wrapper = renderPasswordInput({ label: "required", required: true });
+    });
+
+    it("the required prop is passed to the input", () => {
+      input = wrapper.find("input").at(0);
+      expect(input.prop("required")).toBe(true);
+    });
+
+    it("the isRequired prop is passed to the label", () => {
+      const label = wrapper.find(Label).at(0);
+      expect(label.prop("isRequired")).toBe(true);
+    });
+  });
+
+  describe("disabled", () => {
+    beforeAll(() => {
+      wrapper = renderPasswordInput({ label: "disabled", disabled: true });
+    });
+
+    it("the disabled prop is passed to the input", () => {
+      input = wrapper.find("input").at(0);
+      expect(input.prop("disabled")).toBe(true);
+    });
+
+    it("when the disabled is passed to the input, focus is prevented", () => {
+      input = wrapper.find("input").at(0);
+      const firstFocusableElement = document.querySelector("input");
+      expect(document.activeElement).not.toBe(firstFocusableElement);
+    });
+
+    it("the disabled prop is passed to ButtonMinor", () => {
+      const button = wrapper.find(StyledButtonMinor);
+      expect(button.prop("disabled")).toBe(true);
+    });
+
+    it("when 'forceObscurity' is 'true', ButtonMinor is disabled", () => {
+      wrapper = renderPasswordInput({ forceObscurity: true });
+      const button = wrapper.find(StyledButtonMinor);
+      expect(button.prop("disabled")).toBe(true);
+    });
+
+    it("when the disabled prop is passed to ButtonMinor, focus is prevented", () => {
+      input = wrapper.find("input").at(0);
+      const firstFocusableElement = document.querySelector("input");
+      expect(document.activeElement).not.toBe(firstFocusableElement);
+    });
+  });
+
+  describe("password prop checks", () => {
+    beforeAll(() => {
+      wrapper = renderPasswordInput({});
+    });
+
+    it("default input type is 'password'", () => {
+      expect(wrapper.find("input").at(0).prop("type")).toBe("password");
+    });
+
+    it("when 'forceObscurity' is 'true' input type is 'password'", () => {
+      wrapper = renderPasswordInput({ forceObscurity: true });
+      expect(wrapper.find("input").at(0).prop("type")).toBe("password");
+    });
+
+    it("input type should change from 'password' to 'text' on click", () => {
+      wrapper = renderPasswordInput({});
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      wrapper.update();
+      expect(wrapper.find("input").at(0).prop("type")).toBe("text");
+    });
+
+    it("autoComplete attribute is 'off'", () => {
+      expect(wrapper.find("input").at(0).prop("autoComplete")).toBe("off");
+    });
+
+    it("input id should equal value of 'id' prop when passed", () => {
+      wrapper = renderPasswordInput({ id: "foo" });
+      expect(wrapper.find("input").at(0).prop("id")).toBe("foo");
+    });
+
+    it("input id should be a generated guid when 'id' prop is undefined", () => {
+      wrapper = renderPasswordInput({ id: "" });
+      expect(wrapper.find("input").at(0).prop("id")).toBe(mockedGuid);
+    });
+  });
+
+  describe("buttonMinor prop checks", () => {
+    it("aria-controls attribute is correct", () => {
+      wrapper = renderPasswordInput({ id: "foo" });
+      expect(wrapper.find(StyledButtonMinor).prop("aria-controls")).toBe("foo");
+    });
+
+    it("buttonType attribute is correct", () => {
+      expect(wrapper.find(StyledButtonMinor).prop("buttonType")).toBe(
+        "tertiary"
+      );
+    });
+
+    it("iconPosition attribute is correct", () => {
+      expect(wrapper.find(StyledButtonMinor).prop("iconPosition")).toBe(
+        "before"
+      );
+    });
+
+    it("size attribute is correct", () => {
+      expect(wrapper.find(StyledButtonMinor).prop("size")).toBe("small");
+    });
+
+    it("default label text should be 'Show'", () => {
+      expect(wrapper.find(StyledButtonMinor).text()).toBe("Show");
+    });
+
+    it("label text should change from 'Show' to 'Hide' onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).text()).toBe("Hide");
+    });
+
+    it("default aria-label text should be 'Show password'", () => {
+      expect(wrapper.find(StyledButtonMinor).prop("aria-label")).toBe(
+        "Show password"
+      );
+    });
+
+    it("aria-label text should change from 'Show Password' to 'Hide password' onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).prop("aria-label")).toBe(
+        "Hide password"
+      );
+    });
+
+    it("default iconType should be 'view'", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).prop("iconType")).toBe("view");
+    });
+
+    it("iconType should change from 'show' to 'hide' onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).prop("iconType")).toBe("hide");
+    });
+  });
+
+  describe("buttonMinor styling checks", () => {
+    it("default styling should be correct", () => {
+      wrapper = wrapper.find(StyledButtonMinor);
+      assertStyleMatch(
+        {
+          background: "transparent",
+          color: "var(--colorsActionMinor500)",
+        },
+        wrapper
+      );
+    });
+
+    it("styling should be correct onHover", () => {
+      wrapper = wrapper.find(StyledButtonMinor);
+      assertStyleMatch(
+        {
+          background: "transparent",
+          color: "var(--colorsActionMinor500)",
+        },
+        wrapper,
+        { modifier: ":hover" }
+      );
+    });
+  });
+
+  describe("aria-live region checks", () => {
+    it("default aria-live region text should be correct", () => {
+      wrapper = renderPasswordInput({});
+      expect(wrapper.find(HiddenAriaLive).text()).toBe(
+        "Your Password is currently hidden."
+      );
+    });
+
+    it("aria-live region text should be correct onChange", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(HiddenAriaLive).text()).toBe(
+        "Your password has been shown. Focus on the password input to have it read to you, if it is safe to do so."
+      );
+    });
+
+    it("aria-live region should be visually hidden", () => {
+      assertStyleMatch(
+        {
+          border: "0",
+          height: "1px",
+          margin: "-1px",
+          overflow: "hidden",
+          padding: "0",
+          position: "absolute",
+          width: "1px",
+        },
+        wrapper.find(HiddenAriaLive)
+      );
+    });
+  });
+});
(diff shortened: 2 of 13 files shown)


src/components/password/password.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <>
+      <StyledPassword
+        data-element="styled-password-container"
+        id={internalInputId.current}
+        autoComplete="off"
+        type={type || "password"}
+        {...rest}
+      />
+      <HiddenAriaLive role="alert" aria-live="polite">
```
DipperTheDan, line 26: Not sure if `alert` is the correct role for this as `alert` tends to be for time-sensitive _things_. Not really sure what to use instead though as nothing really fits the needs of a password input. Might be a question for the UX/Accessibility lot. Alert also has `aria-live="assertive"` out the box too so you're overriding this too. We'll see what accessibility thinks 😄

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+describe("Password Input", () => {
+  let wrapper: ReactWrapper;
+  let input: ReactWrapper;
+  const TextTypeOptions: TypeOptions = "password";
+  const PasswordTypeOptions: TypeOptions = "text";
+
+  describe("when rendered", () => {
+    it("should have the Textbox component a it's child", () => {
```
DipperTheDan, line 18:
```suggestion
    it("should have the Textbox component as it's child", () => {
```
Spelling mistake?

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+### Changing type with Carbon component
+
+<Canvas>
+  <Story name="with switch" story={stories.ChangingType} />
+</Canvas>
+
+### Overriding aria-live message translations
+
+Various translations can be applied to the `aria-live` region which contains both a shown and hidden message
+variation. Both of these messages will update assistive technology such as screenreaders When
```
DipperTheDan, line 61:
```suggestion
variation. Both of these messages will update assistive technology such as screen readers when 
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+### With required
+
+<Canvas>
+  <Story name="required" story={stories.WithRequired} />
+</Canvas>
+
+
+### Validations
+
+Validation status can be set by passing `error`, `warning` or `info` prop to the component
```
DipperTheDan, line 157:
```suggestion
Validation status can be set by passing `error`, `warning` or `info` prop to the component.
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+Passing a string to these props will display a properly colored border along with a validation icon and tooltip - string value will be displayed as the tooltip message.
+
+Passing a boolean to these props will display only a properly colored border.
+
+For more information check our [Validations](?path=/docs/documentation-validations--string-validation "Validations") documentation page
```
DipperTheDan, line 163:
```suggestion
For more information check our [Validations](?path=/docs/documentation-validations--string-validation "Validations") documentation page.
```

src/components/password/password.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+    warnOverLimit
+  />
+);
+
+export const Prefix: ComponentStory<typeof Password> = () => (
+  <Password label="Password" value="Password" prefix="prefix" />
+);
+
+export const Sizes: ComponentStory<typeof Password> = () => {
+  const sizes: PasswordProps["size"][] = ["small", "small", "large"];
```
DlgSHi, line 102:
```suggestion
  const sizes: PasswordProps["size"][] = ["small", "medium", "large"];
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+## Props
+
+### Password
+
+Due to the `Textbox` component being used internally by the `Password` component all the props of `Textbox` can be applied to `Password` as well.
+
+<StyledSystemProps of={Textbox} noHeader margin />
```
DlgSHi, line 171: `type` as a prop is not shown in `Props`

src/components/password/password.test.js
```diff
@@ -0,0 +1,178 @@
+import React from "react";
```
DlgSHi, line 1: `ariaLive` - props are not covered

src/components/password/password.test.js
```diff
@@ -0,0 +1,178 @@
...
+  });
+
+  describe("check events for Password component", () => {
+    let callback;
+
+    beforeEach(() => {
+      callback = cy.stub();
+    });
+
+    it("should call onBlur callback when a blur event is triggered", () => {
```
DlgSHi, line 166: don't we have `onChange`/`onClick`/`onFocus` events here?

DlgSHi (changes requested):

src/components/password/password.component.tsx
tomdavies73: Thanks, Dan, have gone for `role="status"`, `aria-live="polite"` for the meantime as I feel the update is not important enough to be an alert in retrospect. But happy to amend/change this down the line.

src/components/password/password.test.js
tomdavies73: Yeah good point, in retrospect I think we can remove any event tests as they're all covered in `Textbox`, which this component renders. Happy to include them though if needed 👍

DipperTheDan: Yeah, I don't think you need the event tests because, as you say, they're covered in Textbox. If you were hijacking these events to do something specific to this component, it would be needed then but I think you're fine to not include these.

src/components/password/password.test.js
```diff
@@ -0,0 +1,190 @@
+import React from "react";
```
DipperTheDan, line 1: Might not be for this PR or if one of the QA's is kind enough, this file probably needs an `checkAccessibility()` test(s).

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+        {...rest}
+      />
+      {/* Below hidden aria live <p> may be superseded by 
+       ongoing work to create a dedicated aria live
+       component - FE-5685  */}
+      <HiddenAriaLive role="status" aria-live="polite">
+        {type === "text"
+          ? l.password.ariaLiveShownMessage()
+          : l.password.ariaLiveHiddenMessage()}
+      </HiddenAriaLive>
```
Parsium, line 33: Nice work on the live region stuff @tomdavies73 👍🏼 

This looks good to me, though I'm not sure if we need to add a small delay on rendering the message, as I believe there are some cases where the content update is too quick for a screen reader so it will duplicate the voice announcement. 

However, I can't say this with certainty, so I will approve for now and perhaps this is something @DlgSHi will be able to verify for you? 😄

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,89 @@
...
+        wrapper = renderPasswordInput({ type: inputTypes });
+        expect(wrapper.find("input").at(0).prop("type")).toBe(inputTypes);
+      }
+    );
+
+    it("Password input id should be correct when id is passed as a custom string", () => {
+      wrapper = renderPasswordInput({ id: "foo" });
+      expect(wrapper.find("input").at(0).prop("id")).toBe("foo");
+    });
+  });
```
Parsium, line 53: It might be a good idea to include an additional test for the other case where the `id` prop is not passed and the `input` id is set to the guid, so we are accounting for this scenario too. 

You should be able to do this by mocking the `guid` helper in the file scope like so:
```ts
import guid from "../../__internal__/utils/helpers/guid";

jest.mock("../../__internal__/utils/helpers/guid");
(guid as jest.MockedFunction<typeof guid>).mockReturnValue("guid-12345");
```

src/locales/en-gb.ts
```diff
@@ -88,6 +88,11 @@ const enGB: Locale = {
     pageX: () => "Page",
     ofY: (count) => `of ${count}`,
   },
+  password: {
+    ariaLiveShownMessage: () =>
+      "Your Password has been shown. Focus on the Password Input to have it read to you, if it safe to do so.",
```
Parsium, line 93:
```suggestion
      "Your password has been shown. Focus on the password input to have it read to you, if it is safe to do so.",
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+## Examples
+
+### Default
+
+<Canvas>
+  <Story name="default" story={stories.Default} />
+</Canvas>
+
+### With other input components used to change type
```
Parsium, line 33:
```suggestion
### Controlling visibility of password
```
What do you think about this wording tweak? This might make it clearer about what this pattern is about at a glance from a design perspective

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,165 @@
...
+supported across different browsers and can have unpredictable results.
+
+Also, we ** strongly ** recommend the use of Carbon components: `Switch`, `Checkbox` and `Button` to change the input type of the `Password` component,
+we have provided an example below of how this is possible. Please ensure `aria-controls` are used with the `id` prop 
+and chosen string to ensure there is a clear indication that the chosen input component is responsible for changing type.
+
+### Changing type with Carbon component
+
+<Canvas>
+  <Story name="with switch" story={stories.ChangingType} />
```
Parsium, line 46: Similar to the previous comment, what do you think?

```suggestion
  <Story name="control password visibility" story={stories.ChangingType} />
```

src/components/password/password.test.js
```diff
@@ -0,0 +1,190 @@
...
+          types === "text"
+            ? "Your Password has been shown. Focus on the Password Input to have it read to you, if it safe to do so."
+            : "Your Password is currently hidden.";
+
+        cy.get("p").contains(ariaString);
+      }
+    );
+
+    it.each(["text", "password"])(
+      "aria-live region should also be visible in the accessibility tree but accessibily hidden regardless of type being %s",
```
Parsium, line 60: Personally I would amend the test description to the following, as we aren't verifying that the live region text is in the accessibility tree here, so this should make it clearer what this test is actually verifying

```suggestion
      "when `type` prop is %s, aria-live region text should be visually hidden",
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+Please ensure that the input type is set to `password` on form submission, this will ensure browsers do not remember
+the inputs content. We have set the `autocomplete` attribute to `off` on the component, but this is not widely 
+supported across different browsers and can have unpredictable results.
+
+Also, we ** strongly ** recommend the use of Carbon components: `Switch`, `Checkbox` and `Button` to change the input type of the `Password` component,
+we have provided an example below of how this is possible. Please ensure `aria-controls` are used with the `id` prop 
+and chosen string to ensure there is a clear indication that the chosen input component is responsible for changing type.
+
+### Controlling visibility of password with Carbon component
```
Parsium, line 43:
```suggestion
```

Parsium (approved):

Parsium (approved):

DipperTheDan (dismissed):

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+import ButtonMinor from "../button-minor/button-minor.component";
+
+export type TypeOptions = "text" | "password";
+
+export interface PasswordProps extends TextboxProps {
+  /** HTML type attribute of the input */
+  type?: TypeOptions;
+}
+
+export const Password = ({ id, type, ...rest }: PasswordProps) => {
```
DipperTheDan, line 15: `type` is declared here but I don't think you've used it anywhere.

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <>
+      <StyledPassword
+        data-element="styled-password-container"
+        id={internalInputId.current}
+        autoComplete="off"
+        type={passwordShown ? "text" : "password"}
+        {...rest}
+      >
+        <ButtonMinor
```
DipperTheDan, line 30: Might be an idea to destructure `disabled` prop and pass it ButtonMinor as when the component is disabled, the button still works.

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,201 @@
...
+    });
+
+    it("iconType should change from `show` to `hide` onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).prop("iconType")).toBe("hide");
+    });
+
+    it("default styling should be correct", () => {
```
DipperTheDan, line 145: Might be a good idea to put these styling tests inside their own `describe` block.

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+### Validations
+
+Validation status can be set by passing `error`, `warning` or `info` prop to the component.
+
+Passing a string to these props will display a properly colored border along with a validation icon and tooltip - string value will be displayed as the tooltip message.
+
+Passing a boolean to these props will display only a properly colored border.
+
+For more information check our [Validations](?path=/docs/documentation-validations--string-validation "Validations") documentation page.
+
```
DipperTheDan, line 140: You'll need to add examples here of how `Password` should look with validations. Old and new validations like we have with `Decimal`.
https://carbon.sage.com/?path=/docs/decimal-input--validations-string-component#validations

src/components/password/password.test.js
DipperTheDan, line 1: When you address the disabled button comment, add a test to guard against it happening again in the future. Could be a cypress test that makes sure the button isn't focusable or something along those lines.

src/components/password/password.component.tsx
tomdavies73: Used on `StyledPassword`, which is a styled `Textbox` 👍

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,201 @@
...
+      ["foo", "foo"],
+      [mockedGuid, ""],
+    ])(
+      "Password input id should be custom value or guid depending on passed value",
+      (idType, id) => {
+        wrapper = renderPasswordInput({ id });
+        expect(wrapper.find("input").at(0).prop("id")).toBe(idType);
+      }
+    );
+  });
```
Parsium, line 81: Since we are testing the default behaviour and adding a custom id, I would perhaps split this test into two separate ones, just so if one test fails it has a clearer description of what it is trying to verify, which would help when debugging:

```tsx
it("input id should equal value of `id` prop when passed", () => {/**/});
it("input id should be a generated guid when `id` prop is undefined", () => {/**/});
```

src/components/password/password.test.js
```diff
@@ -0,0 +1,308 @@
...
+          buttonDataComponent()
+            .should("be.visible")
+            .and("have.css", "background-color", transparent)
+            .and("have.css", "color", colorsActionMinorYang100);
+        });
+    });
+  });
+
+  describe("aria-live region checks", () => {
+    it("default aria-live region should be correct", () => {
```
Parsium, line 162:
```suggestion
    it("when password is hidden, aria-live region should contain the correct text", () => {
```

src/components/password/password.test.js
```diff
@@ -0,0 +1,308 @@
...
+      CypressMountWithProviders(<PasswordComponent />);
+
+      cy.get("p").contains("Your Password is currently hidden.");
+    });
+
+    it("default aria-live region should be correct", () => {
```
Parsium, line 168:
```suggestion
    it("when user clicks to show password, aria-live region should contain the correct text", () => {
```

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,150 @@
...
+## Props
+
+### Password
+
+Due to the `Textbox` component being used internally by the `Password` component all the props of `Textbox` can be applied to `Password` as well.
```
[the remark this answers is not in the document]
tomdavies73: In my opinion, we should be surfacing `type` because consumers will want to enter their own mechanisms to change the `type` back to `password` on submission etc.

src/components/password/password.spec.tsx
tomdavies73: Yeah agreed

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+import guid from "../../__internal__/utils/helpers/guid";
+import useLocale from "../../hooks/__internal__/useLocale";
+import ButtonMinor from "../button-minor/button-minor.component";
+
+export type TypeOptions = "text" | "password";
+
+export interface PasswordProps extends TextboxProps {
+  /** HTML type attribute of the input */
+  type?: TypeOptions;
+}
```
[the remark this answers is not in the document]
tomdavies73: Yeah tbh I'd be interested to see what we can do here, so `type` is not in `TextboxProps`, hence half the need to create a dedicated component, however by having `type` in the interface in theory we should be allowing consumers to manually change type on submission etc for security reasons. `StyledPassword` is just a `Textbox` which uses the `PasswordProps` interface, which lets us get around the `type` issue, if `type` wasn't used in this interface and we tried to pass it to `StyledPassword` it would throw an error, similar to how it would if we just rendered `Textbox` and it passed it to that.

src/components/password/password.component.tsx
tomdavies73: Ah nvm I see what you mean, i'm looking at moving towards `type` being used in a hook, opposed to boolean which we use at the minute. That way we can still change the input type and icons based on the current state, but also allow users to change `type` manually, for use cases like changing back to `password` form submission etc

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,239 @@
...
+    });
+
+    it("label text should change from `Show` to `Hide` onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).text()).toBe("Hide");
+    });
+
+    it("default aria-label text should be 'Show password`", () => {
```
DipperTheDan, line 152:
```suggestion
    it("default aria-label text should be `Show password`", () => {
```

src/components/password/password.spec.tsx
```diff
@@ -0,0 +1,239 @@
...
+    it("aria-label text should change from `Show Password` to `Hide password` onClick", () => {
+      act(() => {
+        wrapper.find(StyledButtonMinor).simulate("click", mockEvent);
+      });
+      expect(wrapper.find(StyledButtonMinor).prop("aria-label")).toBe(
+        "Hide password"
+      );
+    });
+
+    it("default iconType should be 'view'", () => {
```
DipperTheDan, line 167: You've used a mixture of back ticks and single quotes. Like here and line 158 for example, best to pick one and stick to that for consistency.

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+  ...rest
+}: PasswordProps) => {
+  const internalInputId = useRef(id || guid());
+  const l = useLocale();
+
+  const [passwordShown, setPasswordShown] = useState(type);
+  return (
+    <>
+      <StyledPassword
+        data-element="styled-password-container"
```
Parsium, line 28: Might be a good idea to use the `tagComponent` util here, so consumers can set the `data-element` and `data-role` props if they choose to as part of their tests
 
```suggestion
        {...tagComponent("password", rest)}
```

DipperTheDan (dismissed):

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,209 @@
...
+import Password from "carbon-react/lib/components/password";
+```
+
+## Examples
+
+### Default
+
+Please ensure that the `forceObscurity` prop is `true` on form submission, this will ensure browsers do not remember
+the inputs content. We have set the `autocomplete` attribute to `off` on the component also, but this is not widely 
+supported across different browsers and can have unpredictable results.
```
Parsium, line 31: 1. As we want to highlight this as an important security disclaimer, what do you think about making this paragraph bold?

```suggestion
**Warning - Please ensure that the `forceObscurity` prop is `true` on form submission, this will ensure browsers do not remember the inputs content.** 

We have set the `autocomplete` attribute to `off` on the component also, but this is not widely supported across different browsers and can have unpredictable results.
```

src/components/password/password.spec.tsx
Parsium, line 1: Might be worth adding a test, either using jest or cypress, that checks its not possible to change the `<input>` `type` to text when `forceObscurity` is set.

src/components/password/password.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <StyledPassword
+        {...tagComponent("password", rest)}
+        data-element="styled-password-container"
+        id={internalInputId.current}
+        autoComplete="off"
+        type={visibleInput ? "text" : "password"}
+        disabled={disabled}
+        {...rest}
+      >
+        <ButtonMinor
```
Parsium, line 42: Given `forceObscurity` overrides the control of this button, we will either need to disable it or not render it, since this could be confusing for users if they attempt to press the button and expect the visibility to change.

src/components/password/password.spec.tsx
tomdavies73: Just had to go with a standard type test, as we've disabled the button we now can't do any mock onClick events for it 💯

src/components/password/password.component.tsx
tomdavies73: Went for `disabled` to ensure we stay within designs 👍

src/components/password/password.stories.mdx
```diff
@@ -0,0 +1,210 @@
...
+</Canvas>
+
+## Props
+
+### Password
+
+<StyledSystemProps of={Password} noHeader margin />
+
+**Any other supplied props will be provided to the underlying HTML input element**
+
```
Parsium, line 210: Apologies for not spotting this previously @tomdavies73, it would be a good idea to document the new translation keys too (`password.ariaLiveShownMessage` and `password.ariaLiveHiddenMessage`).

We have several examples for other components which should help you with how to structure this

Parsium (dismissed):

DipperTheDan (dismissed):

LaiaSage:
**Focus states**
image

DlgSHi (approved):

DipperTheDan (approved):

Parsium (approved):

DlgSHi (approved):

PR #5859, opened 2023-02-22
refactor(draggable): refactor component to be TypeScript

DipperTheDan:
### Proposed behaviour

Refactor component to be TypeScript

### Current behaviour

Component is currently JS

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

Component should work exactly as it did before the refactor.


src/components/draggable/draggable-container.component.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    getOrder(draggableItemIds, item);
+  };
+
+  const marginProps = filterStyledSystemMarginProps(rest);
+
+  return (
+    <DndProvider backend={HTML5Backend}>
+      <DropTarget getOrder={getItemsId} {...marginProps}>
+        {draggableItems.map((item: any) => {
```
DipperTheDan, line 137: I'm not happy with this `any` typing here. The only other away around this I could think of was to use `React.isValidElement()` as this allows the use of `.props.`. The issue with this was that I'd need a way of returning a value for if the element is valid and if it isn't. By the nature of this component, it only allows `DraggableItems` to be children. So if I try to pass in a child that is a string or number, the invariant fires before it will ever reach this code. Open to suggestions.

src/components/draggable/draggable-container.component.tsx
```diff
@@ -0,0 +1,155 @@
...
+  useEffect(() => {
+    if (!isFirstRender.current) {
+      setDraggableItems(React.Children.toArray(children));
+    } else {
+      isFirstRender.current = false;
+    }
+  }, [children]);
+
+  const findItem = (id: string | number) => {
+    const draggableItem = draggableItems.filter((item: any) => {
```
edleeks87, line 100: We need to type this better than `any`, I'm surprised the linter hasn't complained here

src/components/draggable/draggable-container.component.tsx
edleeks87: This should be fine to use unless I've missed something

```
{draggableItems.map((item) => {
          return React.isValidElement(item) && React.cloneElement(
            item as React.ReactElement,
            {
              id: `${item.props.id}`,
              findItem,
              moveItem,
            },
            [item.props.children]
          );
        })}
```

src/components/draggable/draggable-container.component.tsx
edleeks87: This should be fine to use unless I've missed something

```return React.isValidElement(item) && `${item.props.id}` === id;```

File: src/components/draggable/draggable-item.component.tsx [hunk 3]
```diff
@@ -1,12 +1,30 @@
...
+export interface DraggableItemProps extends StyledDraggableItemProps {
+  /**
+   * The id of the `DraggableItem`.
+   *
+   * Use this prop to make `Draggable` work
+   */
+  id: number | string;
+  /** The content of the component. */
+  children: React.ReactNode;
+  findItem?: (
```
edleeks87, line 20: This needs to be `private` and `ignore`d

src/components/draggable/draggable-item.component.tsx
```diff
@@ -1,12 +1,30 @@
...
+    id: string | number
+  ) => { DraggableItemProps: React.ReactElement; index: number };
+  moveItem?: (
```
edleeks87, line 23: This needs to be `private` and `ignore`d

src/components/draggable/draggable-container.component.tsx
```diff
@@ -0,0 +1,155 @@
+import React, { useEffect, useState, useRef, useMemo } from "react";
+import { DndProvider, useDrop } from "react-dnd";
+import { HTML5Backend } from "react-dnd-html5-backend";
+
+import invariant from "invariant";
+import { filterStyledSystemMarginProps } from "../../style/utils";
+import DraggableItem from "./draggable-item.component";
+import {
+  StyledDraggableContainer,
+  StyledDraggableContainerProps,
```
edleeks87, line 10: This can just be `MarginProps`

src/components/draggable/draggable-item.style.ts
```diff
@@ -1,14 +1,20 @@
...
-import { padding, margin } from "styled-system";
+import { padding, margin, PaddingProps, MarginProps } from "styled-system";
 
 import { baseTheme } from "../../style/themes";
 import Icon from "../icon";
 
-const StyledDraggableContainer = styled.div`
+export type StyledDraggableContainerProps = MarginProps;
+
+const StyledDraggableContainer = styled.div<StyledDraggableContainerProps>`
```
edleeks87, line 9: This can just be `MarginProps` imo

src/components/draggable/draggable-item.component.tsx
```diff
@@ -1,12 +1,30 @@
...
 import { filterStyledSystemPaddingProps } from "../../style/utils";
-import { StyledDraggableItem } from "./draggable-item.style";
+import {
+  StyledDraggableItem,
+  StyledDraggableItemProps,
+  StyledIcon,
+} from "./draggable-item.style";
 
-const paddingPropTypes = filterStyledSystemPaddingProps(styledSystemPropTypes);
+export interface DraggableItemProps extends StyledDraggableItemProps {
```
edleeks87, line 11: I think we should just extend `PaddingProps` here, this approach will add `isDragging` to the public interface

src/components/draggable/draggable-item.style.ts
```diff
@@ -1,14 +1,20 @@
...
   ${margin}
 `;
 
-const StyledDraggableItem = styled.div`
+export interface StyledDraggableItemProps extends PaddingProps {
```
edleeks87, line 13: This is fine but if we go with my suggestion about `DraggableItemProps` extending `PaddingProps` then this doesn't need to be exported anymore

src/components/draggable/draggable-test.stories.tsx
```diff
@@ -17,7 +17,7 @@ export default {
 };
 
 export const Default = () => {
-  const handleUpdate = (items) => {
+  const handleUpdate = (items: any) => {
```
edleeks87, line 20:
```suggestion
  const handleUpdate = (items?: (string | number | undefined)[]) => {
```

src/components/draggable/draggable.spec.tsx
```diff
@@ -75,38 +75,19 @@ describe("Draggable", () => {
 
   it("should return an array with id's", () => {
     wrapper.setProps({ getOrder });
-    wrapper.find("DropTarget").at(0).props().getOrder(2);
+    (wrapper.find("DropTarget").at(0).props() as any).getOrder(2);
```
edleeks87, line 78: I think this is due to the way we are finding the `"DropTarget"` as it seems to return `HTMLAttributes` rather than the component type with props etc. I think we can move `DropTarget` into and `internal` directory and import it in the container files and then remove `any`

src/components/draggable/draggable.spec.tsx
```diff
@@ -190,9 +171,9 @@ describe("Draggable", () => {
     });
 
     it("should drag items within container 1", () => {
-      const container1 = mountNode.querySelector("#container-1");
+      const container1 = mountNode?.querySelector("#container-1");
 
-      const draggableItems = getDraggableItems(container1);
+      const draggableItems = getDraggableItems(container1 as never);
```
edleeks87, line 176: If you change `getDraggableItems` to the following you can remove `as never` wherever you've used it

```
const getDraggableItems = (mountNode: Element | null) =>{
    const draggableItems = mountNode?.querySelectorAll('div[data-element="draggable"]')
    return draggableItems ? Array.from(draggableItems) : [];
  }
```

edleeks87 (changes requested):

tomdavies73 (commented):
Nothing else to add really, past Ed's comments. Only thing I found was, perhaps it may be worth moving the draggable component stories into a new .tsx file and import them back into the .mdx 👍

src/components/draggable/draggable.stories.mdx
```diff
@@ -1,5 +1,6 @@
 import { Meta, Story, Canvas } from "@storybook/addon-docs";
 import { ArgsTable } from "@storybook/components";
+import * as stories from "./draggable.stories.tsx";
 
 import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
 import { DraggableContainer, DraggableItem } from ".";
```
tomdavies73, line 6: Think you can remove the imports for DraggableItem and Checkbox here, as only DraggableContainer is needed

edleeks87: I think it is needed, the prop tables little a little weird to me we can remove the ArgsTable and go with the below and it should display the padding props and the `id` and `children` ones etc

```
<StyledSystemProps of={DraggableItem} noHeader padding defaults={{ py: "8px" }} />
```

edleeks87 (approved):

tomdavies73 (dismissed):

DlgSHi (approved):

edleeks87 (approved):

tomdavies73 (approved):

edleeks87 (approved):

PR #5877, opened 2023-02-28
Fe 5652 buttontoggle accessibility

robinzigmond:
Rewrite the ButtonToggle and ButtonToggleGroup components so that they render as a set of HTML
button elements, rather than an HTML radio button group that visually looks like a group of buttons.

BREAKING CHANGE: Changes to props in the ButtonToggle component:
- `defaultChecked`, `guid` and `onChange` have been removed
- `onClick`, `onBlur` and `onFocus` have new types
Changes to props in the ButtonToggleGroup component:
- all validation-related props (`validationOnLabel`, `error`, `warning` and `info`) have been
removed
- `onBlur` and `onChange` have new types
Other important changes to be aware of:
- these components can no longer be used in an "uncontrolled" fashion, the state of which button
(if any) has been selected must be managed by consumers using React
- the `name` prop on ButtonToggleGroup is no longer mandatory. Both this and the corresponding
prop on ButtonToggle now provide nothing except an internal label which can be used in an
`onChange` handler. These have been kept for backwards compatibility and will soon be removed
- the `checked` prop on ButtonToggle has likewise been kept only temporarily for backwards
compatibility, and is an alias for the new `pressed` prop which sets the `aria-pressed`
attribute on the underlying HTML button

### Proposed behaviour

ButtonToggle should be rendered as an HTML button, and ButtonToggleGroup as a container for several HTML buttons.

### Current behaviour

ButtonToggleGroup is rendered as an HTML radio button group that visually looks like a set of buttons, causing confusion for screenreader users.

### Checklist

#### QA

### Testing instructions

Test that all functionality, apart from the removed props detailed above, works as before, and that there is a good experience for screenreader users.


src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -74,47 +117,73 @@ export const ButtonToggle = ({
...
+
+  const isPressed =
+    pressedButtonValue === null
+      ? pressedPropValue
+      : pressedButtonValue === value;
+
   return (
-    <StyledButtonToggle
+    <StyledButtonToggleWrapper
       data-component={dataComponent || "button-toggle"}
```
edleeks87, line 160: As we're making a breaking change with this PR, it might be worth refactoring this so it extends `TagProps` and removes the customisability of `data-component`. There are a few other components that have the same implementation so I'm happy for you to leave this as is and we can do them all in one ticket if that's preferable

src/components/button-toggle-group/button-toggle-group.stories.mdx
```diff
@@ -35,7 +35,6 @@ import ButtonToggleGroup from "carbon-react/lib/components/button-toggle-group";
   <Story name="default" parameters={{ info: { disable: true } }}>
```
edleeks87, line 35: Maybe worth moving this to a `tsx` file so we can get the benefit of type checking, it's probably out of scope for this ticket so don't worry about it you'd prefer to leave it for now.

src/components/button-toggle/button-toggle.stories.mdx
```diff
@@ -39,14 +39,14 @@ import ButtonToggleGroup from "carbon-react/lib/components/button-toggle-group";
```
edleeks87, line 39: Maybe worth moving this to a `tsx` file so we can get the benefit of type checking, it's probably out of scope for this ticket so don't worry about it you'd prefer to leave it for now.

edleeks87 (commented):
Approach on the whole LGTM @robinzigmond. I was wondering, as we are making a breaking change anyway,  would it be worth doing a bit of tidying up pf the files and moving the group ones into the `button-toggle` directory. This seems to be the only component where we export the group and individual input/button etc from separate paths

mkrds:
It looks like the styles have been changed unintentionally
Your branch:
Zrzut ekranu 2023-03-7 o 08 53 49
Master:
Zrzut ekranu 2023-03-7 o 08 53 58

src/components/button-toggle-group/button-toggle-group.component.tsx
```diff
@@ -45,33 +47,46 @@ export interface ButtonToggleGroupProps
...
   labelAlign?: "left" | "right";
-  /** Callback triggered by blur event on the input. */
-  onBlur?: (ev: React.FocusEvent<HTMLInputElement>) => void;
-  /** Callback triggered by change event on the input. */
-  onChange?: (ev: React.ChangeEvent<HTMLInputElement>) => void;
-  /** The value of the Button Toggle Group */
+  /** Callback triggered by blur event on one of the child buttons. */
+  onBlur?: (ev: React.FocusEvent<HTMLButtonElement>) => void;
+  /** Callback triggered by pressing one of the child buttons. Use with controlled components to set the value prop to ev.target.value. */
+  onChange?: (ev: CustomEvent) => void;
```
mkrds, line 53: I'd rather avoid introducing more `CustomEvents` to the library, so far they've been a PITA during TS refactors. I think that better approach would be to pass the normal event (React.MouseEvent<HTMLButtonElement> in this case) as a first argument, and then the selected value as a second argument

src/components/button-toggle-group/button-toggle-group.component.tsx
```diff
@@ -106,50 +115,65 @@ const ButtonToggleGroup = ({
...
+        value={{
+          onButtonClick: setPressedButtonValue,
+          pressedButtonValue: value || pressedButtonValue,
+          onBlur,
+          onChange,
+          name,
+        }}
+      >
+        <StyledButtonToggleGroupWrapper labelInline={labelInline}>
+          {label && (
```
mkrds, line 148: What is the reason for not using the `FormField` here?

src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -25,12 +30,20 @@ export interface ButtonToggleProps
...
+  /** Callback triggered by blur event on the button. */
+  onBlur?: (ev: React.FocusEvent<HTMLButtonElement>) => void;
+  /** Callback triggered by focus event on the button. */
+  onFocus?: (ev: React.FocusEvent<HTMLButtonElement>) => void;
+  /** Callback triggered by click event on the button. */
+  onClick?: (ev: React.MouseEvent<HTMLButtonElement>) => void;
+  /** Set the pressed state of the toggle button */
+  pressed?: boolean;
+  /** An optional string by which to identify the button in either an onClick handler, or an onChange handler on the parent ButtonToggleGroup. */
+  value?: string;
```
mkrds, line 46: IMO `value` should be required as there's no point of using this component without it. Or maybe I'm missing something :)

src/components/button-toggle-group/button-toggle-group.stories.mdx
robinzigmond: I think @nineteen88 was addressing all of these in his PR that's currently in review. There may be some merge conflicts from that but it'll be worse if I attempt to do the same thing.

src/components/button-toggle/button-toggle.stories.mdx
robinzigmond: see reply above

robinzigmond:
> It looks like the styles have been changed unintentionally Your branch: Zrzut ekranu 2023-03-7 o 08 53 49 Master: Zrzut ekranu 2023-03-7 o 08 53 58

thanks, not sure how I missed this - I'll need to take a look at this. (Chromatic would have caught it anyway, but still - better to catch it before it's passed review.)

src/components/button-toggle-group/button-toggle-group.component.tsx
```diff
@@ -45,33 +47,46 @@ export interface ButtonToggleGroupProps
   fullWidth?: boolean;
   /** The alignment for the text in the label. */
   labelAlign?: "left" | "right";
-  /** Callback triggered by blur event on the input. */
-  onBlur?: (ev: React.FocusEvent<HTMLInputElement>) => void;
-  /** Callback triggered by change event on the input. */
-  onChange?: (ev: React.ChangeEvent<HTMLInputElement>) => void;
-  /** The value of the Button Toggle Group */
+  /** Callback triggered by blur event on one of the child buttons. */
+  onBlur?: (ev: React.FocusEvent<HTMLButtonElement>) => void;
```
mkrds, line 51: Is it necessary to still have `onBlur` on the `ButtonToggleGroup` considering that it is no longer meant to be used as a form input and no longer supports validations?

src/components/button-toggle/button-toggle.component.tsx
robinzigmond: that's a fair point, since the component can no longer be used uncontrolled. I'd be wary of adding yet another breaking change but as it's unlikely to break anything that wouldn't already be broken by this PR, I'm happy to do it.

src/components/button-toggle-group/button-toggle-group.component.tsx
robinzigmond: Probably not necessary, no - but again I wanted to avoid removing props if there was some sensible way to keep them. HTML buttons, just like HTML buttons, can be blurred - I agree this is unlikely to be useful in a non-form context but you never know what consumers might be doing. Maybe this should be deprecated for now?

src/components/button-toggle-group/button-toggle-group.component.tsx
robinzigmond: Oh sorry, I've just read your comments/suggested change more carefully and see that you were proposing passing the `value` as a second argument. That's a reasonable idea, although it will mean consumers having to rewrite existing implementations which I'd have preferred to avoid. I'll also have to add a third argument for the name, at least temporarily.

src/components/button-toggle/button-toggle.component.tsx
robinzigmond: So I've realised that I can't force `value` to be a `string` (ie never `null` or `undefined`) because it's possible that no value might be selected - certainly that's possible on initial load, and if Harpal approves your suggestion of having a button be pressable to deselect then it can be true later too. We could allow `null` for that case rather than `undefined` but this requires quite a few changes and I'm not clear that there's any benefit over just keeping `undefined` being possible and having that mean "there is currently nothig selected".

src/__internal__/label/label.component.tsx
```diff
@@ -16,6 +16,8 @@ export interface LabelProps
   extends ValidationProps,
     StyledLabelProps,
     StyledLabelContainerProps {
+  /* To change HTML element from <label> to <span>. To be used only in very special cases! */
+  asSpan?: boolean;
```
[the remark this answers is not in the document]
robinzigmond: actually it seems to work now, so I'll indeed change it to use `as` as you suggest

src/components/button-toggle-group/button-toggle-group.component.tsx
robinzigmond: It seems fine using FormField now I go back to it - I can just leave off the `validationProps` etc. It means I don't have to recreate the styling/placement of things like the labels and fieldhelp - not sure why I didn't use it before but I'll restore its use to this component as I don't see a good reason not to have it.

src/components/button-toggle-group/button-toggle-group.stories.mdx
edleeks87: Good point, happy for this to be addressed later if needs be 👍

cypress/downloads/example-img.jpg
edleeks87, line 1: I don't think this file should be committed, I'd add the path to the `gitignore`

src/components/button-toggle/button-toggle-group/button-toggle-group.stories.mdx
```diff
@@ -174,6 +166,36 @@ When the `fullWidth` prop is `true`, the buttons will expand in size to take up
   </Story>
 </Canvas>
 
+### Allowing deselection
+
+By default, a `ButtonToggleGroup` there is no way to deselect the currently-selected toggle button except by selecting another one.
```
edleeks87, line 171: Think the first sentence is missing a word (in a ButtonToggleGroup?) or needs slightly tweaking

src/components/button-toggle/button-toggle-group/button-toggle-group.stories.mdx
edleeks87, line 1: I wonder if having a flat structure to the stories rather than having the group stories nested here would be better as it would mirror how we've demonstrated the other components that can render in groups such as Radios, Checkboxes, TileSelects, Accordions and so on. Happy for you to leave it this as is for now and the best way to demo this can be left up to UX/ technical writers and updated in a later ticket if preferred.

edleeks87 (approved):

mkrds (approved):

mkrds (dismissed):

edleeks87 (approved):

edleeks87 (approved):

mkrds (dismissed):

harpalsingh:
As discussed solution provided in demo for deselect option as requested so passes UX and Accessibility reviews.

src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -74,47 +130,80 @@ export const ButtonToggle = ({
     );
   }
 
-  function handleClick() {
-    inputRef.current?.focus();
+  function handleClick(ev: React.MouseEvent<HTMLButtonElement>) {
```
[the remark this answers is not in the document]
robinzigmond: I think we need to keep the `value` argument as optional as an `undefined` value corresponds to nothing being selected - even with deselection not allowed the initial value will usually be `undefined`. If consumers have to use guards or coercion I think that's fine as it keeps them honest about the potential types.

edleeks87 (dismissed):

DipperTheDan (approved):

edleeks87 (approved):

DipperTheDan (approved):

edleeks87 (approved):

PR #5881, opened 2023-03-02
fix(validations): fix various accessibility issues (FE-5382, FE-5337)

### LINKED ISSUE #5411: Textbox - Validation icon uses invalid ARIA attribues
### Current behaviour

When using Textbox validation (and presumably others), the validation icon is presented as a span with an `aria-label` attribute set, but no valid role attribute.

This has been detected when running the Axe Dev Tools to check for accessibility requirements and can be recreated in Storybook on the latest Carbon version (see below)

Screenshot 2022-08-24 at 12 28 56

### Expected behaviour

A valid role is applied to the span in question. No error appears when scanning using Axe Dev Tools.

https://carbon.sage.com/?path=/docs/textbox--validations-string-component

### JIRA Ticket (Sage Only)

SKPU-7210

### Suggested Solution

_No response_

### Carbon Version

110.1.1

### Design Tokens Version

_No response_

Chrome

MacOS, Windows

_No response_

### Confidentiality

### LINKED ISSUE #5498: Resolve accessibility violation for validation icon
### Current behaviour

Axe reports accessibility issue for validation icon in Textbox component.

**Issue Description:** Ensures ARIA attributes are allowed for an element's role
**Fix the following:** aria-label attribute cannot be used on a span with no valid role attribute.

#### Axe Report Screenshot
image

### Expected behaviour

Axe does not report accessibility issues for validation icon.

https://codesandbox.io/s/validation-icon-accessibility-m4fu34

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

109.1.2

### Design Tokens Version

2.40.0

Chrome

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #5594: 'Elements must only use allowed ARIA attributes' with labelHelp errors.
### Current behaviour

Business DD > Manage Users > Add user > click 'Add' > 2 labelHelp appear > run axe report.
Axe report returns 'Elements must only use allowed ARIA attributes'
"To solve this problem, you need to...
aria-label attribute cannot be used on a span with no valid role attribute."

### Expected behaviour

No results on the axe report.

https://carbon.sage.com/?path=/story/textbox--validations-string-label

### JIRA Ticket (Sage Only)

SBS-45292

### Suggested Solution

Axe report links to this URL https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?application=AxeChrome

### Carbon Version

111.0.0

### Design Tokens Version

2.42.0

Chrome

Windows

_No response_

### Confidentiality

### LINKED ISSUE #5828: Error unannounced in forms
### Current behaviour

Error message is not read by VoiceOver when they appear

### Expected behaviour

Error message is read by VoiceOver when they appear

https://codesandbox.io/s/admiring-dust-jnecnj

### JIRA Ticket (Sage Only)

SBS-48611, SBS-48601

### Suggested Solution

_No response_

### Carbon Version

114.9.1

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

grabkowski:
Validation icon was labelled by the aria-label attribute, which was not compliant with WCAG standards and resulted in accessibility issues.
Added `aria-describedby` attribute, that points to the tooltip content.
Added id to the validation message to be used in `aria-describedby` on the input.
 
### Proposed behaviour

Resolve accessibility errors on validation icon.
Enable error announcement on redesigned validations.

### Current behaviour

Accessibility issues on validation icon, when rendered in the textField and on the label.
Errors not announced on redesigned validations.

### Checklist

~~- [ ] Related docs have been updated if required~~

#### QA

### Additional context

fixes #5411
fixes #5498
fixes #5594
fixes #5828

### Testing instructions

There is no test for the tooltip content. Previous test just checked the aria-label attribute which should not be placed on an element that triggers a tooltip.


src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -133,7 +133,7 @@ const CheckableInput = React.forwardRef(
       name: id,
       reverse,
       warning,
-      validationIconId,
+      validationId,
```
Parsium, line 136: I don't think this would be recognised by `FormField` as it would expect a `validationIconId` prop instead. What do you think about doing this? 

```suggestion
      validationIconId: validationId,
```

src/components/checkbox/checkbox.test.js
```diff
@@ -238,27 +238,21 @@ context("Testing Checkbox component", () => {
   it("should render Checkbox component with error message", () => {
     CypressMountWithProviders(<CheckboxComponent error="Error has occurred" />);
 
-    checkboxIcon()
-      .should("have.attr", "aria-label", "Error has occurred")
-      .should("have.attr", "data-element", "error");
+    checkboxIcon().should("have.attr", "data-element", "error");
```
Parsium, line 241: Personally I think we still need to verify if the tooltip message has been rendered alongside checking the icon. What do you think about adding the following assertion so we are testing if the icon and tooltip are both rendered? This would also apply to the two succeeding tests as well :) 

```suggestion
    checkboxIcon().should("have.attr", "data-element", "error");
    tooltipPreview().should("have.text", "Error has occurred");
```

src/components/checkbox/checkbox.test.js
```diff
@@ -415,29 +409,23 @@ context("Testing Checkbox component", () => {
           <CheckboxGroupComponent error="Error has occurred" />
         );
 
-        checkboxGroupIcon()
-          .should("have.attr", "aria-label", "Error has occurred")
-          .should("have.attr", "data-element", "error");
+        checkboxGroupIcon().should("have.attr", "data-element", "error");
```
Parsium, line 412: Similar to the single checkbox tests

```suggestion
        checkboxGroupIcon().should("have.attr", "data-element", "error");
        tooltipPreview().should("have.text", "Error has occurred");
```

src/components/radio-button/radiobutton.test.js
```diff
@@ -316,29 +298,23 @@ context("Testing RadioButton component", () => {
       <RadiobuttonComponent error="Error has occurred" />
     );
 
-    radiobuttonIcon()
-      .should("have.attr", "aria-label", "Error has occurred")
-      .should("have.attr", "data-element", "error");
+    radiobuttonIcon().should("have.attr", "data-element", "error");
```
Parsium, line 301: Similar to Checkbox for the changed tests in this file

```suggestion
    radiobuttonIcon().should("have.attr", "data-element", "error");
    tooltipPreview().should("have.text", "Error has occurred");
```

src/__internal__/checkable-input/checkable-input.component.tsx
grabkowski: nice catch, looks like TypeScript compiler did not catch it.
I have fixed that and added a type to that object, so it will not happen in the future. Looks like there was an unnecessary `name` prop that I also removed.

src/components/checkbox/checkbox.test.js
grabkowski: this was not testing that the tooltip was rendered, this was just checking if the aria-label attribute is on the icon, which should not be added here, so I decided to remove it completely. As for the tooltip message, I'll leave it to QA, as this should be added later by them if necessary:
image

src/__internal__/validation-message/validation-message.component.tsx
```diff
@@ -5,17 +5,26 @@ export interface ValidationMessageProps {
   /** Indicate that error has occurred
   Pass string to display hint with error */
   error?: boolean | string;
+  /** Id of the  component, to be used for accessibility purposes */
```
DipperTheDan, line 8:
```suggestion
  /** Id of the component, to be used for accessibility purposes */
```

src/components/checkbox/checkbox.test.js
DipperTheDan: We'll need to make sure that the person QA'ing this definitely does add the needed Cypress tests.

Parsium (dismissed):

DipperTheDan (approved):

src/components/checkbox/checkbox.test.js
DlgSHi: I don't agree with adding additional tests because we:
1. already have tests for tooltip for each component 
2. need to unskip all tests with `validation`

Parsium: Good point @DlgSHi 👍🏼

In that case what do you think @grabkowski of amending the test descriptions to something like "should render Checkbox component with error icon" - since we aren't asserting if the message has been rendered but the icon. Happy to accept if otherwise as this is minor

src/components/date-range/date-range.test.js
```diff
@@ -506,204 +506,193 @@ context("Test for DateRange component", () => {
...
-        "should check accessibility with the tooltipPosition is set to %s",
-        (position) => {
-          CypressMountWithProviders(
-            <DateRangeCustom
-              m={9}
-              tooltipPosition={position}
-              startError={testCypressText}
-            />
-          );
+    it.each([["top"], ["bottom"], ["left"], ["right"]])(
```
Parsium, line 606: This can also be written as this :)
```suggestion
    it.each(["top", "bottom", "left", "right"])(
```

grabkowski: changed

src/components/checkbox/checkbox.test.js
grabkowski: changed

src/__internal__/checkable-input/checkable-input.component.tsx
Parsium, line 1: I noticed we are passing `aria-labelledby` to `HiddenCheckableInput` but the code of `HiddenCheckableInput` uses `ariaLabelledBy` instead. It seems given the component's type interface:

```ts
export interface CommonHiddenCheckableInputProps
  extends Omit<
    React.InputHTMLAttributes<HTMLInputElement>, // extended interface includes `aria-labelledby` and `aria-describedby`
    "value" | "size" | "type"
  > {
  ariaDescribedBy?: string;
  ariaLabelledBy?: string;
  // more props...
}
```

that the component accepts kebab-case and non kebab-case version of the `aria-describedby` and `aria-labelledby` props. This may cause unexpected behaviour since the kebab case props aren't used in the code, so would you be able to modify this? What do you think about using the kebab case version of the props for consistency?

src/__internal__/input/input.component.tsx
Parsium, line 1: Similarly here because `Input` can accept the kebab-case and non kebab-cases versions of `aria-describedby` and `aria-labelledby` (but only use the non kebab prop) this could cause unexpected behaviour. Would you be able to omit the extended props and use the kebab-case versions of the props?

src/hooks/__internal__/useInputAccessibility/useInputAccessibility.ts
```diff
@@ -1,40 +1,47 @@
...
   warning?: string | boolean;
   info?: string | boolean;
   label?: React.ReactNode;
   fieldHelp?: React.ReactNode;
 }): {
   labelId?: string;
-  validationIconId?: string;
+  validationId?: string;
   fieldHelpId?: string;
   ariaDescribedBy?: string;
```
Parsium, line 21: What do you think about doing this so we don't encounter any unexpected behaviour related to prop names?

```suggestion
  "aria-describedby"?: string;
```

src/__internal__/checkable-input/checkable-input.component.tsx
DipperTheDan: I'm all for keeping things consistent across our components. Good thinking 👍🏻

DipperTheDan (commented):
Nothing more to add than what @Parsium has already mentioned.

src/__internal__/checkable-input/checkable-input.component.tsx
grabkowski: this is an internal component, so there is no reason to pass these props as kebab-case and then converting them again to camelCase multiple times. We do already have `ariaLabelledBy` passed down as camelCase in `CheckableInput`.
When extending all HTML attributes we are just confusing users, as a few props we are overwriting internally, like `aria-describedby`, `aria-checked` or `aria-invalid`, and in my opinion we should not extend that in the first place.

src/__internal__/input/input.component.tsx
grabkowski: if I change that , users will still see unexpected behaviour as we are overwriting HTML attributes that are extended in the Input interface. I think that the best way would be to not extend that, or omit these ARIA properties in the interface. In my opinion we should change the other props to camelCase in our internal components to avoid that assignment: `"aria-<prop>": aria<Prop>`

src/hooks/__internal__/useInputAccessibility/useInputAccessibility.ts
grabkowski: I think that it would lead to more confusion, as we would have to convert that prop to camelCase when we want to use or modify it.

Parsium (dismissed):

DipperTheDan (approved):

DipperTheDan (approved):

DipperTheDan (approved):

PR #5892, opened 2023-03-03
test(group-character): add accessibility tests

ZhuoyuJin:
### Proposed behaviour
- Add `accessibility` Cypress tests for the `Group-character` component to use the new cypress-component-react framework for testing.

### Current behaviour
Old approach for accessibility tests

### Checklist

#### QA

### Testing instructions

- [x] Run `npx cypress open --component` to check if there are newly added tests
- [x] Check if the `group-character.test.js` file passed
- [x] Run `npm run build-storybook` -> `npx sb extract` -> `npx http-server storybook-static -p 9001` and run `npx cypress run --e2e` and check that `Group-character` tests are not running twice.


src/components/grouped-character/grouped-character.test.js
```diff
@@ -15,7 +17,7 @@ import {
 
 const specialCharacters = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
 
-const GroupedCharacterComponent = ({ onChange, ...props }) => {
+const GroupedCharacterGroup = ({ onChange, ...props }) => {
```
DlgSHi, line 20: can we move this component into `grouped-character-test.stories.tsx`

DlgSHi (changes requested):

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -77,3 +78,30 @@ NewValidation.args = {
   separator: "-",
   ...getCommonTextboxArgs(),
 };
+
+export const GroupedCharacterComponent = ({
+  onChange,
+  ...props
+}: {
+  onChange?: (modifiedEvent: any) => void;
```
DlgSHi, line 86: is this a proper type of `onChange` ?

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -77,3 +78,30 @@ NewValidation.args = {
...
+}) => {
```
Parsium, line 87: Since we already have a type interface for `GroupCharacter`'s props, doing the following will correctly type the `onChange` prop for you.

This will also type the other remaining props, since currently `GroupCharacterComponent` is typed to only expect a `onChange` prop, when we could be passing the other `GroupCharacter` props as well.

```suggestion
}: GroupCharacterProps) => {
```

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -77,3 +78,30 @@ NewValidation.args = {
...
+  const [state, setState] = React.useState("");
+
+  const setValue = ({ target }: CustomEvent) => {
+    setState(target.value.rawValue);
+    if (onChange) {
+      onChange(target);
+    }
+  };
```
Parsium, line 95: This will throw a type error since the `onChange` function is expecting an event, not the element it is targeting.
```suggestion
  const setValue = (ev: CustomEvent) => {
    setState(ev.target.value.rawValue);
    if (onChange) {
      onChange(ev);
    }
  };
```

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -77,3 +78,30 @@ NewValidation.args = {
...
+
+  return (
+    <GroupedCharacter
+      label="GroupedCharacter"
+      value={state}
+      onChange={setValue}
+      groups={[2, 2, 3]}
+      separator="-"
+      {...props}
```
Parsium, line 104: If you modify the typing of this mock component's props using my suggestion, the typescript compiler will complain that its possible to overwrite the `groups` and `separator` props here, since these could be present in the `props` object.

To fix this, we'll need to pass the `groups` and `separator` props to `GroupCharacter` if they have been defined, else set a default value for them, something like so:

```tsx
export const GroupedCharacterComponent = ({
  onChange,
  groups,
  separator,
  ...props // this will no longer include `groups` and `separator`, since we have already unpacked them in the above two lines
}: GroupedCharacterProps) => {

  /** ...other code... */

  return (
    <GroupedCharacter
      label="GroupedCharacter"
      value={state}
      onChange={setValue}
      groups={groups || [2, 2, 3]} // pass in `groups` if its defined, else set to a default value
      separator={separator || "-"} // pass in `separator` if its defined, else set to a default value
      {...props}
    />
  );
};
```

DlgSHi (approved):

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -77,3 +81,32 @@ NewValidation.args = {
...
+  ...props
+}: GroupedCharacterProps) => {
+  const [state, setState] = React.useState("");
+
+  const setValue = ({ target }: CustomEvent) => {
+    setState(target.value.rawValue);
+    if (onChange) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      onChange(target);
```
Parsium, line 98: I would suggest doing this so we don't have to disable the type check for `onChange` :)

```suggestion
  const setValue = (ev: CustomEvent) => {
    setState(ev.target.value.rawValue);
    if (onChange) {
      onChange(ev);
```

Parsium (approved):

stephenogorman (approved):

PR #5893, opened 2023-03-06
feat(badge): allow component to be display only - FE-5687

### LINKED ISSUE #5706: Badge component should support static informative display
### Desired behaviour

Quite often the `Badge` component is used for plain informative display purposes and is not intended to be interacted with. To achieve that, the `Badge` component (by default?) should support only the display of the counter, not capture focus, and don't show the clear button.

### Current behaviour

However, it still shows the "X" button on hover and the badge is capturing the focus and does not support any option to turn off that behavior.

https://user-images.githubusercontent.com/30114244/208888908-cbcd0f92-b31f-4323-bab1-4c71c6ad2480.mov

### Suggested Solution

The simplest solution would be to turn off the "Clear button" display and interactivity in case the (optional) `onClick` handler is not present:

Zrzut ekranu 2022-12-21 o 11 57 51

_No response_

_No response_

### Confidentiality

DipperTheDan:
This change allows Badge to be display only by not passing an `onClick` prop to the component. In this variant, the badge will render as a `span` element instead of a `button`.

fixes #5706

### Proposed behaviour

Have a display only variant of `Badge` that is not interactive but looks the same as `Badge` currently does.

### Current behaviour

`Badge` component cannot be display only. Component is always interactive, even without an `onClick` function a hover style is applied implying that the component is able to be interacted with. 

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

A new example has been created called `display only`. This can be seen in Storybook. 
A codesanbox can be created if needs be too.

### DIFF
--- a/src/components/badge/badge-test.stories.tsx
+++ b/src/components/badge/badge-test.stories.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { action } from "@storybook/addon-actions";
 
 import Badge from "./badge.component";
+import Box from "../box";
 import Button from "../button";
 
 export default {
@@ -33,9 +34,20 @@ export const DefaultStory = ({ counter, ...args }: BadgeStoryProps) => {
   );
 };
 
-DefaultStory.story = {
-  name: "default",
-  args: {
-    counter: 1,
-  },
+export const DisplayOnlyStory = ({ counter, ...args }: BadgeStoryProps) => {
+  return (
+    <Box margin="40px">
+      <Badge counter={counter} {...args}>
+        <Button mr={0} buttonType="tertiary">
+          Filter
+        </Button>
+      </Badge>
+    </Box>
+  );
 };
+
+DefaultStory.storyName = "default";
+DefaultStory.args = { counter: 1 };
+
+DisplayOnlyStory.storyName = "display only";
+DisplayOnlyStory.args = { counter: 1 };
--- a/src/components/badge/badge.component.tsx
+++ b/src/components/badge/badge.component.tsx
@@ -1,12 +1,14 @@
 import React from "react";
 import {
   StyledBadgeWrapper,
-  StyledButton,
   StyledCrossIcon,
   StyledCounter,
+  StyledBadge,
 } from "./badge.style";
 
 export interface BadgeProps {
+  /** Prop to specify an aria-label for the component */
+  "aria-label"?: string;
   /** The badge will be added to this element */
   children: React.ReactNode;
   /** The number rendered in the badge component */
@@ -15,21 +17,44 @@ export interface BadgeProps {
   onClick?: (ev: React.MouseEvent<HTMLElement>) => void;
 }
 
-export const Badge = ({ children, counter = 0, onClick }: BadgeProps) => {
+export const Badge = ({
+  "aria-label": ariaLabel,
+  children,
+  counter = 0,
+  onClick,
+}: BadgeProps) => {
+  const shouldDisplayCounter = counter > 0;
+  const counterToDisplay = counter > 99 ? 99 : counter;
+
+  const renderCorrectBadge = () => {
+    const props = onClick
+      ? {
+          buttonType: "secondary",
+          onClick,
+        }
+      : {
+          "aria-label": ariaLabel,
+        };
+
+    if (shouldDisplayCounter) {
+      return (
+        <StyledBadge data-component="badge" {...props}>
+          {onClick && (
+            <StyledCrossIcon data-element="badge-cross-icon" type="cross" />
+          )}
+          <StyledCounter data-element="badge-counter">
+            {counterToDisplay}
+          </StyledCounter>
+        </StyledBadge>
+      );
+    }
+
+    return null;
+  };
+
   return (
     <StyledBadgeWrapper>
-      {counter > 0 && (
-        <StyledButton
-          data-component="badge"
-          buttonType="secondary"
-          onClick={onClick}
-        >
-          <StyledCrossIcon data-element="badge-cross-icon" type="cross" />
-          <StyledCounter data-element="badge-counter">
-            {counter > 99 ? 99 : counter}
-          </StyledCounter>
-        </StyledButton>
-      )}
+      {renderCorrectBadge()}
       {children}
     </StyledBadgeWrapper>
   );
--- a/src/components/badge/badge.spec.tsx
+++ b/src/components/badge/badge.spec.tsx
@@ -1,9 +1,10 @@
 import React from "react";
 import { mount } from "enzyme";
 import { ThemeProvider } from "styled-components";
+import { assertStyleMatch } from "__spec_helper__/test-utils";
 import Badge from "./badge.component";
 import Button from "../button";
-import { StyledCounter } from "./badge.style";
+import { StyledCounter, StyledBadge } from "./badge.style";
 import { baseTheme } from "../../style/themes";
 
 const renderComponent = (props = {}) => (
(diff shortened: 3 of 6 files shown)


src/components/badge/badge.stories.mdx
```diff
@@ -70,13 +70,30 @@ If the counter is less than 1 badge will not show itself
...
 </Canvas>
 
+### Display Only
+
+To acheive a `display only` version of this component, you can simply not pass an `onClick` function to the component. This means that instead of rendering as a `button` element, 
+the component will render as a `span`.
+
+<Canvas>
+  <Story name="display only">
+    <div style={{ margin: "40px" }}>
```
mkrds, line 87: Considering that we have [this PR](https://github.com/Sage/carbon/pull/5869) open I'd avoid adding new instances of `style` prop.

src/components/badge/badge.spec.tsx
```diff
@@ -54,4 +61,79 @@ describe("Badge", () => {
...
+          modifier: "&:hover",
+        }
+      );
+    });
+
+    it("should apply the correct style when onClick is not specified", () => {
+      const wrapper = mount(renderComponent({ counter: 9 }));
+      assertStyleMatch(
+        {
+          cursor: `default`,
```
mkrds, line 117: As above, I'm not sure if it makes sense to test styles here

src/components/badge/badge.spec.tsx
```diff
@@ -54,4 +61,79 @@ describe("Badge", () => {
...
+      );
+    });
+  });
+
+  describe("Styles", () => {
+    it("should apply the correct style when onClick is present", () => {
+      const wrapper = mount(renderComponent({ counter: 9, onClick: () => {} }));
+      assertStyleMatch(
+        {
+          padding: `0`,
```
mkrds, line 83: I don't think we should be checking styles here considering that we have it covered with chromatic. I'd just test that depending on wether we have the onClick present or not, we're rendeing either a span or a button

src/components/badge/badge-test.stories.tsx
```diff
@@ -33,9 +33,28 @@ export const DefaultStory = ({ counter, ...args }: BadgeStoryProps) => {
   );
 };
 
+export const DisplayOnlyStory = ({ counter, ...args }: BadgeStoryProps) => {
+  return (
+    <div style={{ margin: "40px" }}>
```
mkrds, line 38: We should avoid adding new `style` prop instances - https://github.com/Sage/carbon/pull/5869

src/components/badge/badge-test.stories.tsx
```diff
@@ -33,9 +33,28 @@ export const DefaultStory = ({ counter, ...args }: BadgeStoryProps) => {
...
+};
+
 DefaultStory.story = {
   name: "default",
   args: {
     counter: 1,
   },
 };
+
+DisplayOnlyStory.story = {
```
mkrds, line 55: This syntax is already deprecated, might be worth to refactor it like this:

DisplayOnlyStory.storyName = ...
DisplayOnlyStory.args = { ... }

I'd do the same for DefaultStory too

src/components/badge/badge.stories.mdx
DipperTheDan: Completely forgot about that PR. I'll swap the `div` out for `box`.

src/components/badge/badge.spec.tsx
DipperTheDan: If you think they aren't needed, I'll get them removed and a test written like you've said here 😄

src/components/badge/badge-test.stories.tsx
DipperTheDan: Same comment as previous. Completely forgot about that PR. I'll swap the `div` out for `Box`.

src/components/badge/badge.style.ts
```diff
@@ -14,6 +14,30 @@ const StyledCounter = styled.div`
   margin-top: -1px;
 `;
 
+const StyledCounterSpan = styled.span`
```
edleeks87, line 17: As @mkrds points out there's a lot of shared styling here, I think we could just use the `as` prop here to achieve polymorphism, when there's an onClick render a `Button` and when there's not a `span`. This will allow you to conditionally apply any diverged styles whilst not repeating the shared parts etc

src/components/badge/badge-test.stories.tsx
DipperTheDan: Completely forgot about this too. I just coped as was in the file already. I'll get this changed too 👍🏻

src/components/badge/badge.style.ts
```diff
@@ -14,6 +14,30 @@ const StyledCounter = styled.div`
...
+  cursor: default;
```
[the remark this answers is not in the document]
DipperTheDan: Good plan. I'll get this done 👍🏻

src/components/badge/badge.style.ts
edleeks87: You could do this in the component or use `attrs` to achieve it in the styled component itself

```
// renamed StyledButton to be more generic
const StyledBadge = styled.span.attrs(({ onClick }) => ({
  as: onClick ? Button : undefined,
}))`
```

src/components/badge/badge.component.tsx
```diff
@@ -15,21 +18,52 @@ export interface BadgeProps {
...
+  children,
+  counter = 0,
+  onClick,
+}: BadgeProps) => {
+  const shouldDisplayCounter = counter > 0;
+  const displayOnly = !onClick && shouldDisplayCounter;
+  const counterToDisplay = counter > 99 ? 99 : counter;
+
+  const renderCorrectBadge = () => {
+    if (displayOnly) {
```
edleeks87, line 32: You can conditionally spread the props here if you went with the polymorphic approach I suggested below (using `as` prop)

```
const props = onClick ? {
      buttonType: "secondary",
      onClick
    } : {
      "aria-label": ariaLabel
    }
    
    if (shouldDisplayCounter) {
      return (
        <StyledBadge
          data-component="badge"
          {...props}
        >
          {onClick && <StyledCrossIcon data-element="badge-cross-icon" type="cross" />}
          <StyledCounter data-element="badge-counter">
            {counterToDisplay}
          </StyledCounter>
        </StyledBadge>
      );
    }
 ```

File: src/components/badge/badge.spec.tsx (hunk 3 above)
Comment (edleeks87) [reply src/components/badge/badge.spec.tsx:83]: If you went with a polymorphic approach you could` find("button")` or `find("span")` and assert the correct element exists and that would suffice imo. Alternatively yo could test the cursor styling here as that won't be caught in chromatic, but the rest of the styling will be as @mkrds says

File: src/components/badge/badge.spec.tsx (hunk 2 above)
Comment (DipperTheDan) [reply src/components/badge/badge.spec.tsx:117]: As per Ed's comment, I'll leave a test for cursor behaviour as Chromatic won't capture that until we implement the Storybook Play functions.

File: src/components/badge/badge.style.ts [hunk 9]
```diff
@@ -1,8 +1,18 @@
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import StyledIcon from "../icon/icon.style";
-import Button from "../button";
 import Icon from "../icon";
```
edleeks87, line 3: I think your current implementation uses the Button.displayName. Whilst that works I think we should avoid it and use the constructor to polymorph the element instead, that way if we change the displayName it won't break this feature

src/components/badge/badge.style.ts
```diff
@@ -14,48 +25,63 @@ const StyledCounter = styled.div`
...
-    ${StyledIcon} {
-      display: block;
-      width: auto;
-      height: auto;
-      margin-right: 0;
-
-      :before {
-        font-size: 16px;
-        color: var(--colorsActionMajorYang100);
+  ${({ onClick }) => css`
```
edleeks87, line 46: Don't think you need to use `css` twice here fyi

DipperTheDan: I certainly did not. This has been removed.

edleeks87 (approved):

mkrds (approved):

ZhuoyuJin (dismissed):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #5899, opened 2023-03-07
Convert Tabs, Tab, and all internal components used by them, to Typescript

robinzigmond:
This PR is for the refactor of the Tabs and Tab components to Typescript.

### Proposed behaviour

Tabs and Tab should be Typescript components.

Note that the type of the `siblings` prop on both `Tab` and the interal `TabTitle` component have been changed from an array of `ReactNode`s to a single `ReactNode`. This is to workaround a bug in the `typescript-to-proptypes` package where it errors out if given the type `ReactNode[]`, or any other type which mentions `ReactNode` except as a top-level prop. This change should not impact consumers as an array of `ReactNode`s is itself a valid `ReactNode`.

For the same reason, the type of the `children` prop in `Tabs` has been simplified from `ReactNode | ReactNode[]` to just `ReactNode`.

### Current behaviour

They are currently written in Javascript.

### Checklist

#### QA

### Testing instructions


src/components/tabs/__internal__/tab-title/tab-title.spec.tsx
```diff
@@ -16,16 +16,33 @@ import StyledValidationIcon from "../../../../__internal__/validations/validatio
 import Icon from "../../../icon";
 import StyledIcon from "../../../icon/icon.style";
 import Tooltip from "../../../tooltip";
-
-function render(props, renderer = shallow, container = {}) {
+import { ThemeObject } from "../../../../style/themes/base";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mkrds, line 21: You should move this comment 3 lines below

src/components/tabs/tab/tab.component.tsx
```diff
@@ -1,17 +1,62 @@
...
+}
 
-const paddingPropTypes = filterStyledSystemPaddingProps(
-  styledSystemPropTypes.space
-);
+export interface TabProps extends PaddingProps {
+  title?: string;
+  /** A unique ID to identify this specific tab. */
+  tabId: string;
+  className?: string;
```
DipperTheDan, line 16:
```suggestion
  /** @ignore @private */
  className?: string;
```

src/components/tabs/tabs-test.stories.tsx
```diff
@@ -1,16 +1,16 @@
...
+import { Tabs, Tab, TabsProps } from ".";
 
-import { Tabs, Tab } from "./tabs.component";
-
-<Meta
-  title="Tabs/Test"
-  parameters={{
+export default {
+  title: "Tabs/Test",
+  includeStories: "Default",
```
mkrds, line 6: Is this line necessary? It should be included automatically

src/components/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,1660 @@
+import React, { useState } from "react";
+import { Tabs, Tab } from ".";
+import { Checkbox } from "../checkbox";
+import Pill from "../pill";
+import Icon from "../icon";
+import Box from "../box";
+
+export const Default = () => {
+  return (
+    <div style={{ padding: "4px" }}>
```
mkrds, line 10: Considering existence of this PR - https://github.com/Sage/carbon/pull/5869 - maybe it's worth replacing the divs with Boxes?

src/components/tabs/tabs.component.tsx
```diff
@@ -8,21 +8,60 @@ import React, {
...
 import TabsHeader from "./__internal__/tabs-header";
 import TabTitle from "./__internal__/tab-title";
 import { DrawerSidebarContext } from "../drawer";
-import { filterStyledSystemMarginProps } from "../../style/utils";
 
-const marginPropTypes = filterStyledSystemMarginProps(
-  styledSystemPropTypes.space
-);
+export interface TabsProps extends MarginProps {
+  className?: string;
```
DipperTheDan, line 23:
```suggestion
/** @ignore @private */
  className?: string;
```

src/components/tabs/tabs.component.tsx
```diff
@@ -40,12 +79,19 @@ const Tabs = ({
   headerWidth,
   showValidationsSummary,
   ...rest
-}) => {
+}: TabsProps) => {
+  if (position !== "left" && headerWidth !== undefined) {
+    // eslint-disable-next-line no-console
+    console.error(
```
DipperTheDan, line 85: Should this be an `invariant` rather than a console error?

src/components/tabs/tabs.component.tsx
```diff
@@ -8,21 +8,60 @@ import React, {
   createRef,
   cloneElement,
   Children,
+  ReactElement,
 } from "react";
-import PropTypes from "prop-types";
-import styledSystemPropTypes from "@styled-system/prop-types";
-import Tab from "./tab";
+import { MarginProps } from "styled-system";
+import Tab, { TabProps } from "./tab";
```
mkrds, line 14: It looks like you're importing `TabProps` just to reexport it from this file. I'd rather export it directly from the `/.tab` in the `index.ts` file

src/components/tabs/tab/tab.component.tsx
robinzigmond: I missed this because it was only in the proptypes, not the previous `.d.ts` file. Thanks.

src/components/tabs/tabs.component.tsx
robinzigmond: and same again - thanks

src/components/tabs/tabs.component.tsx
robinzigmond: Well that's open to debate. My first attempt was to make it an `invariant` but that means the test for this fails, as it simply crashes the component with an error. Apparently `console.error` is called by errors in custom PropTypes which is what we had before so this was the closest equivalent. Consumers have in the past complained at us completely crashing their app when they've done something that's only mildly questionable - so I'm not sure we want to introduce a new reason for that.

src/components/tabs/__internal__/tab-title/tab-title.style.ts
```diff
@@ -250,7 +270,16 @@ const StyledTitleContent = styled.span`
   `}
 `;
 
-const StyledTabTitle = styled.button`
+interface StyledTabTitleProps extends Omit<TabTitleProps, "tabIndex"> {
+  tabIndex?: number;
```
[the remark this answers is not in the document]
robinzigmond: I'll have a look - I'm pretty sure I had TS errors until I listed them but it's possible that other changes I made since make this OK.

src/components/tabs/__internal__/tab-title/tab-title.component.tsx
```diff
@@ -0,0 +1,295 @@
...
+    const hasHover = useRef(false);
+
+    const showTooltip = () => {
+      setShouldShowTooltip(true);
+    };
+
+    const hideTooltip = () => {
+      // we only provide objects as a ref to this internal component so the type assertion is safe
+      if (
+        (ref as React.RefObject<HTMLElement>)?.current !==
```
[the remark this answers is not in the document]
robinzigmond: Good point, I generally avoid type assertions if it's at all possible. I think I did it this way because of what I explained in the comment - it's completely safe here and therefore the check is unncessary. But we don't know that will always be the case so I'm happy to change it.

src/components/tabs/__internal__/tab-title/tab-title.spec.tsx
robinzigmond: ah yes, thanks - it was correct when I wrote it, but since then Prettier has altered what's on each line!

src/components/tabs/tabs.stories.tsx
robinzigmond: Oh yeah I see what you mean, this won't get merged in due to it being a different file. Thanks, I was unaware of this - might as well do it when the change is so small.

src/components/tabs/tabs.stories.mdx
```diff
@@ -1777,97 +222,8 @@ by hovering your mouse pointer over the `ValidationIcon`s in the example below.
...
-                }
-                checked={infos.two}
-              />
-            </Tab>
-          </Tabs>
-        </div>
-      );
-    }}
-  </Story>
+    story={stories.WithStringValidationsSummarised}
```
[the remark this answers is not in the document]
robinzigmond: good catch, thanks. I missed that this was being used here.

src/components/tabs/tabs.component.tsx
robinzigmond: I'd overlooked that, thanks. I must have been using it at some point, but as it is I agree it's pointless to import just to reexport.

src/components/tabs/__internal__/tab-title/tab-title.style.ts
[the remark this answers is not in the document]
robinzigmond: I've looked at this and indeed most of these props don't need to be listed. The exception is `tabIndex` which is a `string` in `TabTitleProps` but needs to be a `number` on the element itself. While I could leave that in I've decided to change the props to `TabTitle` since its only an internal component anyway, which means I can just use `TabTitleProps` here and don't need a separate `StyledTabTitleProps` - thanks.

src/components/tabs/__internal__/tab-title/tab-title.component.tsx
```diff
@@ -0,0 +1,295 @@
...
+          {renderContent()}
+          {isHref && <Icon type="link" />}
+
+          {hasFailedValidation && (
+            <StyledLayoutWrapper
+              position={position}
+              hasCustomSibling={!!customLayout}
+            >
+              {error && (
+                <ValidationIcon tooltipPosition="top" error={errorMessage} />
```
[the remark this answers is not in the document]
robinzigmond: I just tried it, and the reason is that the types of `onClick` are not compatible - the prop on `ValidationIcon` has a `React.MouseEvent<HTMLElement>` as its argument, where on `TabTitle` I've restricted it to be specifically an anchor or a button. While this is probably fixable by changing some of these types, it seemed simpler to just remove the `onclick` here as I don't see what is gained having a specific click handler on the validation icons that's the same as that on the surrounding element. (`ValidationIcon` doesn't `stopPropagation` or anything like that - in fact it simply asses the `onClick` prop to the underlying `span`, with no changes.)

src/components/tabs/tabs-test.stories.tsx
robinzigmond: It seems that it indeed isn't necessary - I'll remove it

src/components/tabs/tabs.component.tsx
robinzigmond: Actually now I come to look at this, it is being used - on line 95, to assert that the `filteredChildren` have type `ReactElement<TabProps>`. But I now see that it's only necessary to assert it's a `ReactElement` (otherwise it complains about `child.props`), so I'll make that change and export `TabProps` directly as you suggest.

mkrds (dismissed):

edleeks87 (approved):

src/components/tabs/tabs.component.tsx
edleeks87: I had a little think about this more (I agree with @robinzigmond) and perhaps a package like [this](https://www.npmjs.com/package/warning) would be better for us to use in instances like this. Not suggesting it as part of this PR as I'd prefer to do it across all components at once if we agree it's a good idea

src/components/tabs/tab/tab.component.tsx
```diff
@@ -23,13 +69,17 @@ const Tab = ({
   updateWarnings,
   updateInfos,
   href,
+  // title is destructured purely to NOT spread it as part of rest to the underlying HTML element.
+  // Both this and titleProps are used as part of child.props inside Tabs component
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   title,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   titleProps,
```
edleeks87, line 77: this can probably not be destructured and then the linter comment isn't needed

robinzigmond: read the comment a few lines above - I spent a while on this but unfortunately both this and `title` *do* need to be destructured, despite not being directly used in this component. Otherwise they end up spread onto the styled component. Maybe that's OK for `titleProps` (as opposed to `title` where it definitely isn't) but I'd still rather not run the risk we add something to that that's a valid HTML attribute that ends up causing some weird effect.

edleeks87 (approved):

mkrds (approved):

PR #5900, opened 2023-03-07
 fix(advanced-color-picker): add visually hidden description list to describe selected color

tomdavies73:
### Proposed behaviour

This pull request will add a visually hidden description list which will describe the currently selected colour, as well as an `aria-describedby` attribute to the button which opens the component, to create an association between the two. However, this is just a part-solution to improve accessibility in the short-term, as an accessible re-design of how this component is opened is due to take place in the foreseeable future, which will integrate with the changes made in this PR.

### Current behaviour

Currently, there is no indication of the button's purpose due to a lack of any proper labelling. Also, the current chosen colour is only returned to the user visually, so users unable to see colour will not know which colour is currently selected.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/advanced-color-picker/advanced-color-picker.component.tsx
+++ b/src/components/advanced-color-picker/advanced-color-picker.component.tsx
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useCallback, useRef } from "react";
 import { MarginProps } from "styled-system";
 import {
   StyledAdvancedColorPickerWrapper,
+  HiddenCurrentColorList,
   StyledAdvancedColorPickerCell,
   StyledAdvancedColorPickerPreview,
   DialogStyle,
@@ -9,6 +10,9 @@ import {
 import { SimpleColorPicker, SimpleColor } from "../simple-color-picker";
 import Events from "../../__internal__/utils/helpers/events";
 import { filterStyledSystemMarginProps } from "../../style/utils";
+import guid from "../../__internal__/utils/helpers/guid";
+import useLocale from "../../hooks/__internal__/useLocale";
+import { Dt, Dd } from "../definition-list";
 
 export interface AdvancedColor {
   label: string;
@@ -74,6 +78,9 @@ export const AdvancedColorPicker = ({
     setSelectedColorRef,
   ] = useState<HTMLInputElement | null>(null);
 
+  const descriptionId = useRef(guid());
+  const l = useLocale();
+
   const simpleColorPickerData = useRef<{
     gridItemRefs: Array<HTMLInputElement | null>;
   }>(null);
@@ -90,6 +97,14 @@ export const AdvancedColorPicker = ({
           : null,
     };
   });
+
+  const currentSelectedColor = () => {
+    const returnedColor = availableColors.find(
+      (color) => color.value === currentColor
+    )?.label as string;
+
+    return returnedColor || currentColor;
+  };
 
   useEffect(() => {
     if (dialogOpen || open) {
@@ -198,11 +213,26 @@ export const AdvancedColorPicker = ({
     >
       <StyledAdvancedColorPickerCell
         data-element="color-picker-cell"
+        aria-label={l.advancedColorPicker.ariaLabel()}
+        aria-describedby={descriptionId.current}
         onClick={handleOnOpen}
         onKeyDown={handleOnKeyDown}
         color={currentColor}
         tabIndex={0}
       />
+      <HiddenCurrentColorList
+        id={descriptionId.current}
+        data-element="current-color-description"
+      >
+        <Dt>
+          {l.advancedColorPicker.currentColorDescriptionTerm(
+            currentSelectedColor()
+          )}
+        </Dt>
+        <Dd>
+          {l.advancedColorPicker.currentColorAssigned(currentSelectedColor())}
+        </Dd>
+      </HiddenCurrentColorList>
       <DialogStyle
         aria-describedby={ariaDescribedBy}
         aria-label={ariaLabel}
--- a/src/components/advanced-color-picker/advanced-color-picker.stories.mdx
+++ b/src/components/advanced-color-picker/advanced-color-picker.stories.mdx
@@ -1,9 +1,11 @@
 import { Meta, Canvas, Story } from "@storybook/addon-docs";
 import AdvancedColorPicker from ".";
 import StyledSystemProps from "../../../.storybook/utils/styled-system-props";
-import { useState } from "react";
+import * as stories from "./advanced-color-picker.stories.tsx"
 
-<Meta title="Advanced Color Picker" />
+import TranslationKeysTable from "../../../.storybook/utils/translation-keys-table";
+
+<Meta title="Advanced Color Picker" parameters={{ info: { disable: true } }} />
 
 # Advanced Color Picker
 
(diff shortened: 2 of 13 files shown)


src/components/advanced-color-picker/advanced-color-picker.component.tsx
```diff
@@ -198,6 +209,9 @@ const AdvancedColorPicker = ({
     >
       <StyledAdvancedColorPickerCell
         data-element="color-picker-cell"
+        aria-label={
+          l.advancedColorPicker.currentColorString() + currentColorName()
```
robinzigmond, line 213: I'm no expert in localisation but I'm pretty sure that constructing localised strings by concatenation like this is not the way to go. The issue is that different languages construct things in different ways - we shouldn't assume that the correct translation for the strong "current color: red" is always going to be a translation for "current color", followed by a translation for "red".

I don't know what the correct solution for this is within Carbon, but the issue could be avoided if using the alternative approach I suggest in my overall PR comment: have a fixed aria label of (something like) "open color picker" and use an aria-describedby (pointing to an appropriate visually-hidden element) that has the name of the color. Both of these strings can be included in the locale file and the relation between them should be obvious from what the screenreader does from reading one as the name and the other as the description.

robinzigmond (commented):
The code looks fine for the change you've made, but I have doubts as to whether it's correct. I'd prefer to use a combination of `aria-label` and `aria-describedby`, with the `aria-label` staying fixed, rather than having a dynamically-changing `aria-label` meaning the accessible name of the button changes every time the user changes colour.

As you'll see from my other comment this approach makes localisation a lot easier as well.

Have you discussed this with the Accessibility team? If so and they've recommended this approach over my suggested one, then no worries and I'll accept their judgement (and I recognise this is only a temporary fix anyway). But I'm really not sure how the localisation should be handled in this case.

robinzigmond (commented):
in addition to the comments already given, as communicated privately I personally don't think having a dynamic `aria-label` that changes with the selected colour is a very good idea. I'd prefer to see a fixed `aria-label` like "choose colour", and an `aria-describedby` pointing to some hidden text which updates with the chosen colour.

src/components/advanced-color-picker/advanced-color-picker.component.tsx
```diff
@@ -91,6 +94,24 @@ export const AdvancedColorPicker = ({
     };
   });
 
+  const CurrentAriaLabelColor = () => {
+    const currentSelectedColorValue = selectedColorRef?.getAttribute(
+      "aria-label"
```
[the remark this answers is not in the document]
robinzigmond: hmm, the comments seem to have ended up in a different order - I was referring to the one on lines 99-102 of the same file, which is showing as above this now in the Github PR!

src/components/advanced-color-picker/advanced-color-picker.component.tsx
```diff
@@ -91,6 +94,24 @@ export const AdvancedColorPicker = ({
...
+    );
+
+    if (!currentSelectedColorValue) {
+      let index = -1;
```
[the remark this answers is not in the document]
tomdavies73: Good shout, definitely better than what was previously written 👍

src/components/advanced-color-picker/advanced-color-picker.component.tsx
```diff
@@ -74,6 +78,9 @@ export const AdvancedColorPicker = ({
     setSelectedColorRef,
   ] = useState<HTMLInputElement | null>(null);
 
+  const selectedColorId = useRef(guid());
```
robinzigmond, line 81: really minor thing but I'm not a fan of the name `selectedColorId` for this const - it made me think for a while it was the ID of the actual selected color input, when it's just of a (hidden) text description of it. I'd prefer a name like `descriptionId` but there's probably something better. (I don't mind if you disagree and prefer to keep it as it is now, btw.)

src/components/advanced-color-picker/advanced-color-picker.component.tsx
```diff
@@ -91,6 +98,13 @@ export const AdvancedColorPicker = ({
     };
   });
 
+  const currentAriaLabelColor = () => {
```
robinzigmond, line 101: this name is definitely wrong as the color is no longer used in the aria label. `currentSelectedColor` maybe?

src/components/advanced-color-picker/advanced-color-picker.stories.mdx
```diff
@@ -76,3 +77,25 @@ Shows how the color picker can be used to select a color from a predefined set.
 ### AdvancedColorPicker
 
 <StyledSystemProps of={AdvancedColorPicker} noHeader margin />
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object to the
+[i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
+
+<TranslationKeysTable
```
robinzigmond, line 86: I know you mentioned this to me yourself but just to make sure it's in writing - you should include the `ariaLabel` key in here as well.

src/components/advanced-color-picker/advanced-color-picker.style.ts
```diff
@@ -11,12 +11,24 @@ import Dialog from "../dialog/dialog.component";
...
 import baseTheme from "../../style/themes/base";
+import { Dl } from "../definition-list";
 
 const StyledAdvancedColorPickerWrapper = styled.div`
   ${margin}
   display: inline-block;
 `;
 
+/** To be replaced by accessibly hidden class added in FE-5503 */
+const HiddenCurrentColorList = styled(Dl)`
```
robinzigmond, line 22: I'd prefer to put these styles in a global location (in this PR I have it in `styles/utils` https://github.com/Sage/carbon/pull/5931 ) as we know we need this in a few different places. I'm a little worried that whichever of these PRs gets merged second will need to refactor, and it's unlikely it will be flagged as a conflict on Github and we'll simply end up with the same styles defined in 2 separate places.

src/components/advanced-color-picker/advanced-color-picker.stories.mdx
```diff
@@ -64,7 +66,6 @@ Shows how the color picker can be used to select a color from a predefined set.
             setOpen(false);
           }}
           onBlur={() => {}}
-          open={open}
```
robinzigmond, line 67: I don't think we should remove this line. I tested the story out and am rather surprised that it seems you can still open and close the picker's dialog, I presume because it's now worked in "uncontrolled" fashion" in which case the whole `open` state variable isn't actually being used.

I know you wanted to change this so it's not open when the page first loads. The correct way to make that change would be to change the initial value in the `useState` call from `true` to `false`.

However both that, and the change you've made here, will mean that Chromatic will stop picking up any visual changes in the picker - it doesn't look like this story has Chromatic disabled (nor should it I don't think). So we should probably have it open by default when Chromatic is running, and not otherwise. We have examples of doing this elsewhere on dialog-like components - using the `isChromatic` util to return a "defaultOpenState` and use that in the `useState` call - if you search the repo for those names you'll find plenty of examples to copy.

tomdavies73: Yeah I agree, I've utilised the util, but I've had to refactor the stories file type from `.mdx` to `.tsx`, and I've also decided to refactor the test story too. I'll do these as a separate commit, so it's slightly easier to track 👍

robinzigmond (dismissed):

src/components/advanced-color-picker/advanced-color-picker.style.ts
DipperTheDan: Agree with Robin here and we will need to be careful we don't merge with same code in two different places.

DipperTheDan (approved):

DlgSHi (dismissed):

robinzigmond (dismissed):

DipperTheDan (dismissed):

DlgSHi (dismissed):

robinzigmond (dismissed):

DipperTheDan (approved):

DipperTheDan (approved):

PR #5916, opened 2023-03-13
Menu ts refactor FE-4806 

edleeks87:
### Proposed behaviour

Converts all JavaScript files (apart from `test` file) in `Menu` directory to TypeScript

Fixes issue with incorrect `Icon` colour when `MenuItem` has focus
 
image

### Current behaviour

`Menu` files are JavaScript

Incorrect `Icon` colour when `MenuItem` has focus

image

### Checklist

#### QA

### Testing instructions


src/components/menu/scrollable-block/scrollable-block.component.tsx
```diff
@@ -35,8 +50,7 @@ const ScrollableBlock = ({
           data-component="scrollable-block-parent"
           overrideColor
           variant={parentVariant}
-          as="div"
-          href="#"
+          // as="div"
```
nineteen88, line 53: Left over commented line

src/components/menu/menu.spec.tsx
```diff
@@ -280,13 +247,17 @@ describe("Menu", () => {
         openSubmenu(wrapper, 0);
 
         act(() => {
-          menuItem.props().onClick();
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          menuItem.props().onClick!({} as React.MouseEvent<HTMLButtonElement>);
```
Parsium, line 251: What do you think of doing this so we are avoiding ignoring typescript and mimicking the user's action?
```suggestion
          menuItem.simulate("click");
```

src/components/menu/menu.spec.tsx
```diff
@@ -280,13 +247,17 @@ describe("Menu", () => {
...
         });
 
         wrapper.update();
 
         act(() => {
-          menuItem.props().onKeyDown(events[key]);
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          menuItem.props().onKeyDown!(
+            (events[key] as unknown) as React.KeyboardEvent<HTMLButtonElement>
+          );
```
Parsium, line 260: Similar to the previous comment
```suggestion
          menuItem.simulate("keydown", { key: events[key] });
```

src/components/menu/menu.spec.tsx
```diff
@@ -297,4 +268,32 @@ describe("Menu", () => {
...
+    };
+    it("unregisters itself to the parent Menu", () => {
+      wrapper = mount(<MockMenu />);
+
+      act(() => {
+        wrapper
+          .find(StyledMenuItemWrapper)
+          .first()
+          .props()
+          .onClick({} as React.MouseEvent<HTMLButtonElement>);
```
Parsium, line 293: Similar to the previous comments
```suggestion
        wrapper
          .find(StyledMenuItemWrapper)
          .first()
          .simulate("click")
```

src/components/menu/menu.spec.tsx
```diff
@@ -297,4 +268,32 @@ describe("Menu", () => {
...
+      });
+
+      wrapper.update();
```
Parsium, line 296: Are we missing an assertion here to check if the `MenuItem` is still on the page?

src/components/menu/menu.stories.mdx
```diff
@@ -53,210 +47,68 @@ import {
...
-              <MenuItem href="#">Item Submenu Two</MenuItem>
-            </MenuItem>
-          </Menu>
-        </div>
-      ))}
-    </Box>
-  </Story>
+  <Story
+    name="default"
+    parameters={{ chromatic: { disableSnapshot: true } }}
```
Parsium, line 53: We may want to move the parameters from this file to the `.stories.tsx` file, as they will be ignored here by storybook since they must be defined the same location as the stories themselves.

src/components/menu/menu.stories.tsx
```diff
@@ -0,0 +1,542 @@
...
+    {menuTypes.map((menuType) => (
+      <Box key={menuType}>
+        <Typography variant="h4" textTransform="capitalize" my={2}>
+          {menuType}
+        </Typography>
+        <Menu menuType={menuType}>{children}</Menu>
+      </Box>
+    ))}
+  </Box>
+);
```
Parsium, line 31: What do you think about moving the JSX from this function into each of the stories. While this would duplicate repeating code it may be less confusing for consumers reading the story source code in the docs who won't know what the `renderMenu` function is.

src/components/menu/scrollable-block/scrollable-block.component.tsx
Parsium, line 53: Not sure if this was meant to stay commented? :)

src/components/menu/scrollable-block/scrollable-block.spec.tsx
```diff
@@ -13,19 +13,31 @@ import { assertStyleMatch } from "../../../__spec_helper__/test-utils";
...
-  let container;
-  let wrapper;
-
-  const render = (menuType, props, blockIndex, focusIndex) => {
+  let container: HTMLDivElement | null;
+  let wrapper: ReactWrapper;
+
+  const render = (
+    menuType: MenuType,
+    props = {},
```
Parsium, line 28: Similar to previous comments
```suggestion
    props: Partial<ScrollableBlockProps> = {},
```

src/components/menu/menu-segment-title/menu-segment-title.component.tsx
```diff
@@ -1,11 +1,16 @@
 import React, { useContext } from "react";
-import PropTypes from "prop-types";
 import StyledTitle from "./menu-segment-title.style";
 import MenuContext from "../menu.context";
 import { StyledMenuItem } from "../menu.style";
+import { VariantType } from "../menu-item";
 
-const MenuSegmentTitle = React.forwardRef(
-  ({ children, variant = "default" }, ref) => {
+export interface MenuTitleProps {
```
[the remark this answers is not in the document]
edleeks87: We were previously exporting an interface called `MenuTitleProps` from the `d.ts` file so we can't rename it without it being a breaking change imo

src/components/menu/menu.spec.tsx
edleeks87: The tests fail when I update it to use simulate, as these are test files I personally don't think it's an issue to use a non-null assertion here as using an `if` will also require me to add a lint ignore statement

src/components/menu/menu.spec.tsx
edleeks87: I don't actually need the type assertion here so I'll just go with removing it for now as this will all be refactored to RTL soon

src/components/menu/menu.spec.tsx
edleeks87: Good point, I originally added this just to get the last bit of coverage but I seem to have missed an assertion 🤦

src/components/menu/menu.stories.tsx
edleeks87: yeah good point I hadn't considered that the code blocks would prevent a DRY approach here

src/components/menu/menu-full-screen/menu-full-screen.component.tsx
```diff
@@ -15,38 +14,38 @@ import Icon from "../../icon";
 import Portal from "../../portal";
 import MenuDivider from "../menu-divider/menu-divider.component";
 
+export interface MenuFullscreenProps {
```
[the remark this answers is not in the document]
edleeks87: Good point, I was going back and forth on this and the rest props you also raised as it was a way of consumers passing the `data-` tags and I thought extending the interface could be done when we remove the rest spread etc. However, these props aren't documented if we don't extend the interface so you've convinced me to add it in 😄

src/components/menu/menu-item/menu-item.component.tsx
```diff
@@ -58,13 +127,15 @@ const MenuItem = ({
...
+  const ref = useRef<HTMLAnchorElement & HTMLButtonElement & HTMLDivElement>(
+    null
+  );
   const focusFromMenu = focusId === menuItemId.current;
   const focusFromSubmenu = submenuFocusId
     ? submenuFocusId === menuItemId.current
     : undefined;
-  const inputRef = useRef(null);
-  const inputIcon = useRef(null);
+  const inputRef = useRef<HTMLInputElement | null | undefined>(null);
```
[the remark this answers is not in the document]
edleeks87: I'm not really sure what you mean tbh, can you provide a code example? The current implementation is necessary as the `querySelector` used to assign the input and icon refs may return an element or null but tsc thinks the main `ref` may not have been initialised and that therefore would cause the other refs to be `undefined`. I can update it to remove the additonal type if I go with something like below, let me know if you have a different approach

```
const inputRef = useRef<HTMLInputElement | null>(null);
  const inputIcon = useRef<HTMLSpanElement | null>(null);

  inputIcon.current = ref.current ? ref.current?.querySelector(
    "[data-element='input-icon-toggle']"
  ) : null;
  inputRef.current = ref.current ? ref.current?.querySelector("[data-element='input']") : null;
  ```

File: src/components/menu/menu-item/menu-item.spec.tsx [hunk 12]
```diff
@@ -38,21 +38,23 @@ const updateFocusId = jest.fn();
 
 const menuContextValues = {
   handleKeyDown: mockMenuhandleKeyDown,
-  menuType: "light",
+  menuType: "light" as MenuType,
```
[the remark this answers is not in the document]
edleeks87: Yeah good idea, `handleKeyDown` isn't a valid prop on the interface as well 😆

src/components/menu/menu-item/menu-item.component.tsx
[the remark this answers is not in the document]
Parsium: I'm happy to go with this example 👍🏼

src/components/menu/__internal__/submenu/submenu.spec.tsx
Parsium, line 1: Just a few comments from me for this file, let me know what you think 👍🏼 
1. Specifying the return type for helper functions where it could be useful to catch incompatible types, for example:
```ts
// line 93
const menuContextValues = (menuType: MenuType) => ({ ... }); // becomes
const menuContextValues = (menuType: MenuType): MenuContextProps => ({ ... });

// line 108
const render = (menuType: MenuType, props = {}) => { ... }; // becomes
const render = (menuType: MenuType, props: Partial<SubmenuProps> = {}) => { ... };
```
2. Using generics where we can instead of type assertions, such as:
```ts
// line 186
(wrapper.find("button").getDOMNode() as HTMLButtonElement).click(); // becomes
wrapper.find("button").getDOMNode<HTMLButtonElement>().click();
```

src/components/menu/menu-item/menu-item.component.tsx
```diff
@@ -33,75 +110,83 @@ const MenuItem = ({
...
     menuType,
     openSubmenuId,
-  } = useContext(MenuContext);
+  } = useContext<MenuContextProps>(MenuContext);
   const menuItemId = useRef(guid());
-  const submenuContext = useContext(SubmenuContext);
+  const submenuContext = useContext<SubmenuContextProps>(SubmenuContext);
   const {
-    registerItem: registerSubmenuItem,
-    unregisterItem: unregisterSubmenuItem,
```
Parsium, line 55: Need we mean to remove these as consuming context values?

src/components/menu/menu-segment-title/menu-segment-title.component.tsx
```diff
@@ -1,17 +1,25 @@
...
+import { VariantType } from "../menu-item";
+import tagComponent, {
+  TagProps,
+} from "../../../__internal__/utils/helpers/tags";
 
-const MenuSegmentTitle = React.forwardRef(
-  ({ children, variant = "default" }, ref) => {
+export interface MenuTitleProps extends TagProps {
+  children: string;
+  variant?: VariantType;
```
Parsium, line 12: Might be worth adding back in the comment descriptions for these props as `MenuSegmentTitle` is a consumer-facing subcomponent

src/components/menu/menu.test.js
```diff
@@ -1399,5 +1448,23 @@ context("Testing Menu component", () => {
           expect(callback).to.have.been.calledOnce;
         });
     });
+
+    it("should have correct keyboard navigation order when children of submenu update", () => {
```
Parsium, line 1452: Are we missing some assertions for this test? I assume we would want to check the correct element is being focused after each arrow press?

src/components/menu/menu.test.js
```diff
@@ -1399,5 +1448,23 @@ context("Testing Menu component", () => {
...
+      CypressMountWithProviders(<MenuWithChildrenUpdating />);
+
+      submenu().eq(positionOfElement("first"), div).trigger("mouseover");
+      cy.wait(500);
```
Parsium, line 1456: If we do want to verify the correct element is focused after each arrow press, what do you think about using aliases with a locator function instead of using `cy.wait()`? Similar to [this example from Cypress](https://docs.cypress.io/guides/references/best-practices#Unnecessary-wait-for-cyget)? This would mean the test wouldn't be dependent on the timeout duration and break if we change it.

src/components/menu/menu.stories.mdx
Parsium, line 1: I noticed the prop tables for `ScrollableBlock` and `MenuFullscreen` are currently missing. This may be because of an error with the imports in the file?

src/components/menu/__internal__/submenu/submenu.component.tsx
Parsium, line 1: I noticed in this file and for `MenuItem`, we have removed the `registerItem` and `unregisterItem` functions from the submenu context, was this an intentional change?

src/components/menu/menu.test.js
edleeks87: It's not really necessary here imo, the full keyboard behaviour is tested by other cases, this test is just to show that when the children update the focusable items in state are updated so that the order is correct

src/components/menu/__internal__/submenu/submenu.component.tsx
edleeks87: yep it's been replaced by the `useEffect`

src/components/menu/menu.test.js
edleeks87: I'm not convinced this approach is necessary for this scenario tbh. The wait is just so the state can update with the setTimeout, it's entirely contrived for this test case rather than baked into the component being tested.

src/components/menu/menu-segment-title/menu-segment-title.component.tsx
edleeks87: Good point thanks

src/components/menu/__internal__/submenu/submenu.spec.tsx
edleeks87: Yeah sorry it looks like I've accidentally reverted these changes when I rebased

src/components/menu/menu.stories.mdx
edleeks87: I think I've missed the named export from those components

src/components/menu/menu.test.js
edleeks87: Yep good point, looks like I've missed it in a few places, likely due to copy pasting 👏

Parsium (approved):

nineteen88 (approved):

nineteen88 (approved):

Parsium (dismissed):

Parsium (commented):
Happy to approve once @nineteen88's comment has been addressed 👍🏼

Parsium (approved):

nineteen88 (approved):

stephenogorman (approved):

PR #5919, opened 2023-03-14
refactor(detail): convert files to typescript

tomdavies73:
### Proposed behaviour

### Current behaviour

Component files are currently JavaScript.

### Checklist

#### QA

### Testing instructions


src/components/detail/detail.component.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { filterStyledSystemMarginProps } from "../../style/utils";
+import tagComponent from "../../__internal__/utils/helpers/tags/tags";
+import {
+  StyledDetail,
+  StyledDetailContent,
+  StyledDetailIcon,
+  StyledDetailFootnote,
+} from "./detail.style";
+
+export interface DetailProps extends MarginProps {
```
Parsium, line 13: Since we are using `tagComponent()` below, we'll want to extend the `TagProps` interface so we are accounting for the `data-element` and `data-role` props.

```suggestion
import { TagProps } from "../../__internal__/utils/helpers/tags";
export interface DetailProps extends MarginProps, TagProps {
```

src/components/detail/detail.style.ts
```diff
@@ -15,7 +18,7 @@ const StyledDetail = styled.div`
   }
 `;
 
-const StyledDetailContent = styled.div`
+const StyledDetailContent = styled.div<IconProps>`
```
Parsium, line 21: What do you think about alternatively doing this? I feel the name `IconProps` may be a bit misleading since this styled component is a container rather than an icon.

```suggestion
const StyledDetailContent = styled.div<{ hasIcon: boolean }>`
```

src/components/detail/detail.style.ts
```diff
@@ -46,7 +49,7 @@ const StyledDetailIcon = styled(Icon)`
   }
 `;
 
-const StyledDetailFootnote = styled.div`
+const StyledDetailFootnote = styled.div<IconProps>`
```
Parsium, line 52: Similar to previous comment :)
```suggestion
const StyledDetailFootnote = styled.div<{ hasIcon: boolean }>`
```

src/components/detail/detail-test.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    },
+  },
+  argTypes: {
+    icon: {
+      options: ICONS,
+      control: {
+        type: "select",
+      },
+    },
+  },
```
Parsium, line 22: Since `children` is typed as a node, the storybook control for editing the `children` prop has been automatically set to accept JSON input - since the prop could be a React node in addition to a string. If you press the `a` key in Canvas mode to open the addons sidebar, you'll see this: 

Screenshot 2023-03-17 at 16 40 10

Since we are only interested in checking if some value passed to `children` renders properly, we can manually change the control input for `children` like so:

```suggestion
  },
  children: {
    control: { type: "text" },
  },
```

This will change the control type to accept text input instead

DipperTheDan (commented):
Nothing more to add from me. Just address the comments left from @Parsium 😃

DipperTheDan (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #5928, opened 2023-03-19
feat(announcer): introduce utility for creating custom screen reader announcements (FE-5685)

Parsium:
### Proposed behaviour

Introduce a new `Announcer` utility to add [ARIA live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) support into Carbon - this allows consumers to make custom screen reader announcements.

- Utility returns two functions `announcePolitely` and `announceAssertively` that both accept a message to dispatch. The first function dispatches a polite announcement - which is only made when the user is idle and any existing announcements have been spoken. The second function dispatches an assertive announcement - which will immediately be spoken by the screen reader.
- Both returned functions have a corresponding live region, where each is attached to the `document.body`. All messages dispatched via these functions are then inserted into the polite or assertive region by mutating the DOM directly.
- Both the polite and assertive live regions are visually hidden from view.
- Both live regions are uniquely identifiable and cannot be duplicated - so micro-frontends can use the same two regions for communication.  

The approach to directly mutate the DOM to create and modify the live regions was opt-ed for two reasons. To ensure both regions are placed high in the DOM tree as possible, and so micro-frontends have an identical, global method of accessing them. For context, here are some examples from other open-source projects which also use this approach:
[@adobe/react-spectrum](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/live-announcer/src/LiveAnnouncer.tsx)
[@chakra-ui](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/live-region/src/live-region.ts)

### Current behaviour

Consumers have no means of specifying ARIA live region properties on Carbon components, meaning they must wrap components in html containers with live region properties themselves:

```tsx
const LoadableContent = () => (
   <div aria-live="polite"> {/* Any newly content added inside here will be announced */}
      <Box>
          {isLoading ? (
              <Loader aria-label="Loading content"/>
          ) : (
              {/* Render loaded content */}
          )}
      </Box>
   </div>
);
```
Live regions must also exist on the page before any changes occur, meaning they cannot be created at the same time when a consumer dispatches a message. To always ensure a live region is rendered, it needs to exist high in the React DOM as possible (closer to the root). A consumer could implement this themselves in their app, but this becomes more challenging in micro-frontend environments, since co-ordination between different micro-frontends is needed to prevent messages from being interrupted, lost among other incoming messages or spoken out of order.

### Checklist

#### QA

### Testing instructions

Open storybook locally with `npm start` and navigate to <http://localhost:9001/?path=/docs/documentation-hooks-useannouncer>, then manually test the storybook examples using the following screen reader and browser combinations:

- Chrome/Edge + NVDA 
- Firefox + NVDA
- Safari + macOS VoiceOver

Using these combinations, verify if the following behaviour is observed for each storybook example:

#### Announce loading a new page
1. When retrieving reports message is shown, should announce *"Retrieving reports, please wait"* one time.
2. Then when reports table is shown, should announce *"Reports loaded"* one time

#### Announce loading new content
1. When placeholder lines underneath "Contacts" heading are shown, should announce *"Loading contacts, please wait"* one time.
2. Then when contacts table is shown, should announce *"Contacts loaded"* one time.

#### Form validation messages
- When save button is pressed and both the name and email fields are empty, should announce *"2 errors found. Please correct the form errors before resubmitting"* one time.
- When save button is pressed and one of the name and email fields are empty, should announce *"1 errors found. Please correct the form errors before resubmitting"* one time.
- When save button is pressed and both the name and email fields are filled, should announce *"Successfully submitted form"* one time.

### DIFF
--- a/src/utils/Announcer/Announcer.spec.ts
+++ b/src/utils/Announcer/Announcer.spec.ts
@@ -1,0 +1,60 @@
+import { screen, waitForElementToBeRemoved } from "@testing-library/dom";
+import Announcer from ".";
+
+describe("Announcer", () => {
+  afterEach(() => {
+    Announcer.cleanup();
+  });
+
+  it("when announcePolitely is called, inserts passed message into polite live region", () => {
+    const message = "foo";
+    Announcer.announcePolitely(message);
+
+    const region = screen.getByRole("status");
+    expect(region).toHaveTextContent(message);
+  });
+
+  it("when announcePolitely is called, inserted message is cleared after delay", async () => {
+    const message = "bar";
+    Announcer.announcePolitely(message);
+
+    const region = screen.getByRole("status");
+    expect(region).toHaveTextContent(message);
+    await waitForElementToBeRemoved(() => screen.queryByText(message));
+  });
+
+  it("when announcePolitely is called and polite live region exists in document, do not recreate", () => {
+    Announcer.announcePolitely("some message");
+
+    expect(screen.queryByRole("status")).toBeInTheDocument();
+
+    Announcer.announcePolitely("new message");
+    expect(screen.queryAllByRole("status")).toHaveLength(1);
+  });
+
+  it("when announceAssertively is called, inserts passed message into assertive live region", () => {
+    const message = "baz";
+    Announcer.announceAssertively(message);
+
+    const region = screen.getByRole("alert");
+    expect(region).toHaveTextContent(message);
+  });
+
+  it("when announceAssertively is called,inserted message is cleared after delay", async () => {
+    const message = "buzz";
+    Announcer.announceAssertively(message);
+
+    const region = screen.getByRole("alert");
+    expect(region).toHaveTextContent(message);
+    await waitForElementToBeRemoved(() => screen.queryByText(message));
+  });
+
+  it("when announceAssertively is called and assertive live region exists in document, do not recreate", () => {
+    Announcer.announceAssertively("some message");
+
+    expect(screen.queryByRole("alert")).toBeInTheDocument();
+
+    Announcer.announceAssertively("new message");
+    expect(screen.queryAllByRole("alert")).toHaveLength(1);
+  });
+});
--- a/src/utils/Announcer/Announcer.ts
+++ b/src/utils/Announcer/Announcer.ts
@@ -1,0 +1,93 @@
+type Politeness = "polite" | "assertive";
+
+/* Due to possibility of having multiple Carbon versions being used on
+ * the same page, it is neccessary to maintain the same values for the
+ * following ids */
+const CARBON_INTERNALS__LIVE_REGION_IDS: Record<Politeness, string> = {
+  polite: "carbon-polite-live-region",
+  assertive: "carbon-assertive-live-region",
+};
+
+const parentNode = document.body;
+const TIMEOUT_DURATION = 200;
+const roles: Record<Politeness, string> = {
+  polite: "status",
+  assertive: "alert",
+};
+
+function getRegion(politeness: Politeness): HTMLElement {
+  const region = document.getElementById(
+    CARBON_INTERNALS__LIVE_REGION_IDS[politeness]
+  );
+
+  if (region) return region;
+
+  const element: HTMLDivElement = document.createElement("div");
+  const visuallyHiddenStyles: Partial<CSSStyleDeclaration> = {
+    clip: "rect(0 0 0 0)",
+    clipPath: "inset(50%)",
+    height: "1px",
+    overflow: "hidden",
+    position: "absolute",
+    whiteSpace: "nowrap",
+    width: "1px",
+  };
+  element.setAttribute("id", CARBON_INTERNALS__LIVE_REGION_IDS[politeness]);
+  element.setAttribute("role", roles[politeness]);
+  element.setAttribute("aria-live", politeness);
+  element.setAttribute("aria-atomic", "false");
+  element.setAttribute("aria-relevant", "additions");
+  Object.assign(element.style, visuallyHiddenStyles);
+
+  return parentNode.appendChild(element);
+}
+
+/**
+ * Dispatches a message for a screen reader to announce to users.
+ * Message will be announced at the next graceful opportunity, such as
+ * when the user is idle or any current announcements have finished being
+ * spoken. */
+function announcePolitely(message: string): void {
+  const politeRegion = getRegion("polite");
+
+  politeRegion.textContent = message;
+
+  window.setTimeout(() => {
+    politeRegion.textContent = "";
+  }, TIMEOUT_DURATION);
+}
+
+/**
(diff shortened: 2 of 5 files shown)


src/hooks/useAnnouncer/useAnnouncer.ts
```diff
@@ -0,0 +1,69 @@
...
+  );
+  element.setAttribute("aria-live", assertiveness);
+  element.setAttribute("aria-atomic", "false");
+  element.setAttribute("aria-relevant", "additions");
+  Object.assign(element.style, visuallyHiddenStyles);
+
+  return document.body.appendChild(element);
+}
+
+function useAnnouncer(): {
```
Parsium, line 41: Currently, I have faced some issues with repeated screen reader announcements with VoiceOver and NVDA. I have a feeling it may be due the hook not cleaning itself up properly - needing to use an effect with a returned cleanup to remove the DOM elements on unmount, something like this:

```tsx
useEffect(() => () => destroyLiveRegions());
```

However, I haven't been able to verify this and encountering more problems with setting this up, especially since we need the live regions available in the DOM for the various micro-frontends to use.

src/hooks/useAnnouncer/useAnnouncer.ts
```diff
@@ -0,0 +1,69 @@
...
+} {
+  const politeRegion = useRef<HTMLElement>(getRegion("polite"));
+  const assertiveRegion = useRef<HTMLElement>(getRegion("assertive"));
+
+  // use .textContent instead of .innerText, which can't be used to get text of elements hidden with css.
+  function announcePolitely(message: string): void {
+    politeRegion.current.textContent = message;
+    setTimeout(() => {
+      politeRegion.current.textContent = "";
+    }, TIMEOUT_DURATION);
```
Parsium, line 53: My intention is to change the design of announcement functions to append `<p>` elements into the live region with the dispatched message inside, as opposed to setting the text content, so we are avoid having a race condition where multiple messages from different micro-frontends are incoming at a similar time.

src/hooks/useAnnouncer/useAnnouncer.ts
```diff
@@ -0,0 +1,69 @@
+import { useRef } from "react";
+
+/* Due to possibility of multiple Carbon versions being used on the page,
+ * it is neccessary to maintain the same values for the following ids */
+const POLITE_REGION_ID = "__CARBON_INTERVALS_POLITE_LIVE_REGION";
```
edleeks87, line 5:
```suggestion
const POLITE_REGION_ID = "__CARBON_INTERNALS_POLITE_LIVE_REGION";
```

src/hooks/useAnnouncer/useAnnouncer.ts
```diff
@@ -0,0 +1,69 @@
...
+const ASSERTIVE_REGION_ID = "__CARBON_INTERVALS_ASSERTIVE_LIVE_REGION";
```
edleeks87, line 6:
```suggestion
const ASSERTIVE_REGION_ID = "__CARBON_INTERNALS_ASSERTIVE_LIVE_REGION";
```

src/hooks/useAnnouncer/useAnnouncer.ts
edleeks87: I agree with @robinzigmond about seeing if we can avoid it using a custom hook. However, if we do and end up needing to use a useEffect we should either use the `useStableCallback` or a similar approach that stores the announcer functions in refs if for some reason it doesn't work

src/hooks/useAnnouncer/useAnnouncer.ts
edleeks87: I think @robinzigmond is right about needing to maintain a global queue here, we'll need to make sure that the timeouts are only triggered when the next message is popped off etc

src/hooks/useAnnouncer/use-announcer.stories.mdx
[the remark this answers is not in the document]
edleeks87: Agree with @robinzigmond, may be worth catching up someone from UX/ accessibility to help construct a valid pattern if you're not sure etc

src/hooks/useAnnouncer/use-announcer.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Usage
+
+### Announce loading a new page
+
+<Canvas>
+  <Story name="announce loading a new page" story={stories.AnnouncePageLoad} />
+</Canvas>
+
+### Announce loading new content
```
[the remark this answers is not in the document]
Parsium: I wanted to include two examples of a full page reload and loading a smaller, local area of content, however they do behave exactly the same.

I could update "Announce loading new content" to demonstrate this better, or I'm happy to just remove it altogether if you feel its not necessary :)

src/hooks/useAnnouncer/use-announcer.stories.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<Canvas>
+  <Story
+    name="announce loading new content"
+    story={stories.AnnounceContentLoad}
+  />
+</Canvas>
+
+### Form validation messages
```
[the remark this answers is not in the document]
Parsium: Good idea 👍🏼

src/hooks/useAnnouncer/useAnnouncer.spec.tsx
```diff
@@ -0,0 +1,145 @@
...
+    (assertiveness, id) => {
+      render(<MockComponent />);
+      const region = screen.queryByRole(regionRoles[assertiveness]);
+      expect(region).toBeEmptyDOMElement();
+      expect(region).toHaveAttribute("id", id);
+    }
+  );
+
+  it.each<Assertiveness>(["polite", "assertive"])(
+    "sets correct ARIA live properties onto %s live region",
```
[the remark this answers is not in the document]
Parsium: Yeah given at the moment only text can be added to the live regions, specifying `aria-atomic='false'` in particular isn't strictly necessary. Though I may end up changing the functionality of adding messages so I'll either keep this or remove it based on the outcome of this 👍🏼

src/hooks/useAnnouncer/useAnnouncer.ts
Parsium: Thanks for this guys 👍🏼 I've scratching my head for a while on this one. I like the idea of using two separate hooks for `announcePolitely` and `announceAssertively` instead

src/hooks/useAnnouncer/use-announcer.stories.mdx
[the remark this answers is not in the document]
Parsium: I'm aware that several teams are moving over to using `react-form-hook` now too, which also allows teams to validate on blur. I'll speak with Accessibility about it as I'm not sure either if using ARIA live in this scenario is good practice.

PR #5931, opened 2023-03-21
feat(help): improve accessibility

### LINKED ISSUE #5569: VoiceOver announces help tooltip as "Button" for Dialog, Dialog FullScreen
### Current behaviour

For stories "with help" in dialog and dialog full screen, the screen reader announces the help tooltip as a button, when it shouldn't as it doesn't do anything on click. 
Image 20-10-2022 at 10 49

### Expected behaviour

It looks like the issue is with our Help component - the help icon has been given a role of button except when it has an href. It shouldn't be announced as a button as there's no interaction can be done to it. 

https://carbon.sage.com/?path=/story/dialog--with-help

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

latest

### Design Tokens Version

_No response_

MacOS

_No response_

### Confidentiality

robinzigmond:
Remove `role="button"` from help icon, and remove `aria-label` and `aria-describedby`. The relevant text is now included in the DOM directly in styles that make it not visible but are picked up by screenreaders - ensuring a good experience for screenreader users. Deprecate the `tooltipId` prop which is no longer necessary for accessibility, and deprecate `aria-label` and replace with `accessibilityLabel` (as the prop's value is no longer used for an `aria-label` attribute)

fix #5569

### Proposed behaviour

The help icon, which is not clickable and merely displays a tooltip on focus/hover, should not be announced as a button. There should be no Axe violations, and screenreaders should read both the supplied `ariaLabel` prop if supplied (otherwise it defaults to "help") and the tooltip text.

Further `ariaLabel` should be renamed (I have chosen `accessibilityLabel` but am open to better suggestions) because the prop is no longer used as an HTML `aria-label` attribute.

### Current behaviour

Help icon is announced as a button by screenreaders, which is incorrect and confusing.

### Checklist

#### QA

### Testing instructions

Do both general regression testing and screenreader testing (both Voiceover with Safari on MacOS and NVDA with supported browsers on Windows) on the Help component and all other components which use it: Heading and the various input components

### DIFF
--- a/src/components/button-toggle-group/button-toggle-group.spec.tsx
+++ b/src/components/button-toggle-group/button-toggle-group.spec.tsx
@@ -21,7 +21,7 @@ import ButtonToggle from "../button-toggle/button-toggle.component";
 import StyledButtonToggleGroup from "./button-toggle-group.style";
 import FormFieldStyle from "../../__internal__/form-field/form-field.style";
 import FormField from "../../__internal__/form-field";
-import StyledHelp from "../help/help.style";
+import { VisuallyHidden as HelpVisuallyHidden } from "../help/help.style";
 
 jest.mock("../../__internal__/utils/helpers/guid");
 (guid as jest.MockedFunction<typeof guid>).mockImplementation(
@@ -59,8 +59,9 @@ describe("ButtonToggleGroup", () => {
       helpAriaLabel: text,
     });
 
-    const ariaLabel = wrapper.find(StyledHelp).prop("aria-label");
-    expect(ariaLabel).toEqual(text);
+    const helpVisuallyHiddenText = wrapper.find(HelpVisuallyHidden).text();
+
+    expect(helpVisuallyHiddenText).toContain(text);
   });
 
   it("when inputWidth prop is passed, renders component with correct width", () => {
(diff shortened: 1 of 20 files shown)


grabkowski (changes requested):
I think that this change will not improve accessibility and will break the ARIA tooltip pattern.
I don't know why would the button role of the tooltip trigger be wrong, almost every component library is using buttons for this purpose.

We should consult that with Accessibility Team before we do implement that change.

robinzigmond:
> I think that this change will not improve accessibility and will break the ARIA tooltip pattern. I don't know why would the button role of the tooltip trigger be wrong, almost every component library is using buttons for this purpose.
> 
> We should consult that with Accessibility Team before we do implement that change.

Thanks Kuba for the review - for some reason I didn't spot these comments till now.

Can you give me a link to somewhere else that is using this pattern - `role="button"` on something that only works on hover, not on llick - or an article sugesting it? I'm not seeing what you think will break here?

I already discussed this with Martin with regard to what to do, and he was the one who suggested this pattern of using visibly-hidden text. (He did also suggest, as an alternative, keeping it as a button but making it work on click rather than hover - but when I discussed the options with Nick we agreed the visibly hidden text would be easier and wouldn't result in any changes for users without accessibility needs, so it seemed the best change to make.)

grabkowski (dismissed):

grabkowski (approved):

Parsium (dismissed):

Parsium (approved):

tomdavies73 (approved):

PR #5934, opened 2023-03-22
feat(pager): surface interactivePageNumber prop to ensure selecting page number is optional

### LINKED ISSUE #5674: Pager - disable free-text page number navigation
### Desired behaviour

The ability to type in the desired page number should be optional (i.e. input containing current page can be disabled)

### Current behaviour

The Pager allows you to navigate to pages which may not necessarily exist for the use case implemented. Further explanation from Slack discussion on the matter:

`We make our first call without tokens, which is interpreted by the database as "get page 1". If there are more records than the page size you've asked for, you'll get a token back for the next X records. Page two would then return a token to get you the previous set of results (and another, separate token if there are records beyond the ones shown - we work with a previousToken and nextToken approach). These tokens control Carbon's navigation buttons, but asking the database for "page 2" couldn't work as we have know way of knowing what the token for page 2's records would be - it's all calculated server-side.`

### Suggested Solution

Allow me to pass a prop to the pager that disables the input and replaces it with a simple text label (`Page X of Y`)

_No response_

_No response_

### Confidentiality

tomdavies73:
fix #5674

### Proposed behaviour

This PR will surface the `interactivePageNumber` prop on the `pager` component, and internal component `pager-navigation`, when `false` one full label will be rendered which includes the current page and total page count all in one string. This will give consumers flexibility of the interactivity of the page number, and limit the means end-users use to navigate the pager.

Screenshot 2023-03-24 at 14 43 21

### Current behaviour

Currently, the page number is returned as an interactive number input, with two labels either side. 

Screenshot 2023-03-22 at 15 47 57

### Checklist

#### QA

### Testing instructions


cypress/locators/pager/index.js
```diff
@@ -21,6 +22,7 @@ export const pageSelect = () =>
 export const pageSelectElement = () => cy.get(PAGE_SELECT);
 export const maxPages = () => cy.get(MAX_PAGES);
 export const currentPageWrapper = () => cy.get(CURRENT_PAGE);
+export const currentPageWLabelWrapper = () => cy.get(CURRENT_PAGE_LABEL);
```
DipperTheDan, line 25:
```suggestion
export const currentPageLabelWrapper = () => cy.get(CURRENT_PAGE_LABEL);
```

DipperTheDan: You'll also need to change the name if you've used it somewhere else too.

tomdavies73: Classic me

src/components/pager/pager.component.tsx
```diff
@@ -47,6 +47,8 @@ export interface PagerProps {
   ) => void;
   /** Current visible page */
   currentPage?: number | string;
+  /** If true, page number navigation will be changed to a non-interactive label */
+  nonInteractivePager?: boolean;
```
edleeks87, line 51: I think this prop name is a bit vague as it's implying that the whole Pager is not interactive but there will still be buttons/ dropdowns potentially. Maybe something like `hasPageInput` which defaults to true or something like that?

src/components/pager/__internal__/pager-navigation.component.tsx
```diff
@@ -161,24 +165,33 @@ const PagerNavigation = ({
...
-            onKeyUp={(ev) =>
-              Events.isEnterKey(ev) ? handlePageInputChange(ev) : false
-            }
-          />
-          <StyledPagerNoSelect>{l.pager.ofY(pageCount)}</StyledPagerNoSelect>
-        </StyledPagerNavInner>
-      )}
+            aria-live="polite"
+          >
+            {l.pager.pageXOfY(currentPage.toString(), pageCount)}
```
edleeks87, line 192: I don't think we need the extra function as we can use the existing ones to build the string

```suggestion
           {l.pager.pageX()} {currentPage} {l.pager.ofY(pageCount)}
```

src/components/pager/pager.component.tsx
tomdavies73: I've went with `nonInteractivePageNumber`, let me know what you think 👍

edleeks87: Sounds good to me 👍

src/components/pager/__internal__/pager-navigation.component.tsx
```diff
@@ -161,24 +168,33 @@ const PagerNavigation = ({
...
-          </label>
-          <NumberInput
-            value={currentPage.toString()}
-            data-element="current-page"
-            onChange={handleCurrentPageChange}
-            onBlur={handlePageInputChange}
+      {showPageCount &&
+        (!nonInteractivePageNumber ? (
+          <StyledPagerNavInner>
+            <label htmlFor={currentPageId}>
```
edleeks87, line 174: This was what I was referring to when I said could we make this a label instead of a div within label etc, will it not work as just a `label` instead of `StyledPagerNoSelect` rendering a `div`?

src/components/pager/pager.component.tsx
DipperTheDan: Agreed @edleeks87 - this seems a lot cleaner and clear way of having/using this prop.

edleeks87 (approved):

DipperTheDan (dismissed):

src/components/pager/pager.test.js
```diff
@@ -255,6 +256,44 @@ context("Test for Pager component", () => {
       firstArrow().should("be.visible");
       previousArrow().should("be.visible");
     });
+
+    it.each([
+      [false, "not.exist"],
+      [true, "be.visible"],
+    ])(
+      "when interactivePageNumber is %s, standard pager nav number input rendered correctly",
```
DipperTheDan, line 264:
```suggestion
      "when interactivePageNumber is %s, standard pager nav number input is rendered correctly",
```
This is literally the pettiest comment ever so I understand if you don't want to change it 😆

DipperTheDan (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #5935, opened 2023-03-22
refactor(progress-tracker): convert ProgressTracker component to Typescript

### LINKED ISSUE #5864: ProgressTrackerProps interface don't have description property
### Current behaviour

ProgressTrackerProps https://github.com/Sage/carbon/blob/master/src/components/progress-tracker/progress-tracker.d.ts don't have description property, so I can't use it in my project with TypeScript like it is described in the documentation: 
https://carbon.sage.com/?path=/docs/progress-tracker--custom-label-values

The behavior of component with description property is correct, but without correct interface it generates error in projects with TypeScript.

### Expected behaviour

The "description" property with value string should be added to ProgressTrackerProps https://github.com/Sage/carbon/blob/master/src/components/progress-tracker/progress-tracker.d.ts. 

https://codesandbox.io/s/fast-waterfall-xs3d0z

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

The "description" property with value string should be added to ProgressTrackerProps https://github.com/Sage/carbon/blob/master/src/components/progress-tracker/progress-tracker.d.ts. 

### Carbon Version

114.14.1

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

nineteen88:
### Proposed behaviour

Convert javascript files to use typescript instead

fixes #5864

### Current behaviour

Currently ProgressTracker component is not using Typescript

### Checklist

#### QA

### Testing instructions


src/components/progress-tracker/progress-tracker.spec.tsx
```diff
@@ -18,17 +18,19 @@ import useResizeObserver from "../../hooks/__internal__/useResizeObserver";
...
+
 describe("ProgressTracker", () => {
-  let wrapper;
+  let wrapper: ReactWrapper;
 
   const originalOffsetHeight = Object.getOwnPropertyDescriptor(
     HTMLElement.prototype,
     "offsetHeight"
-  );
+  ) as PropertyDescriptor | ThisType<unknown>;
```
robinzigmond, line 29: I confess I don't really understand what the `ThisType` is doing here - but I tried removing it, both here and on line 33, and VSCode isn't complaining at me. So it looks like it's not needed.

robinzigmond:
I just realised you'd done this all in 1 commit. I thought the agreed approach was to separate the moving of `mdx` stories into `tsx` into its own commit. I don't personally mind too much though.

src/components/progress-tracker/progress-tracker.spec.tsx
nineteen88: The type expected was `PropertyDescriptor | ThisType<any>`.  This is a leftover though, now it's just being cast as `PropertyDescriptor` to resolve the type issue `ThisType` isn't needed. Good spot.

robinzigmond (dismissed):

DlgSHi (approved):

DlgSHi (approved):

PR #5949, opened 2023-03-30
feat(tabs): enhace tabs component - FE-5558

mkrds:
Tabs header can now be scrolled horizontally when content overflows
Navigating through Tab titles with arrow keys no longer automatically selects the tab

### Checklist

#### QA

### Testing instructions


src/components/tabs/tabs.test.js
```diff
@@ -681,25 +685,23 @@ context("Testing Tabs component", () => {
           });
       });
 
-      it("should call onTabChange callback when right arrow key event is triggered", () => {
+      it("should focus next tab title when right arrow key event is triggered", () => {
         CypressMountWithProviders(<TabsComponent onTabChange={callback} />);
```
robinzigmond, line 689: we don't need the `onTabChange` prop in these tests any more

src/components/tabs/tabs.test.js
robinzigmond, line 1: since we're no longer testing for tab change on the arrow key presses, should we not add new cypress test(s) that pressing enter or space when focused on a tab should trigger the tab change?

Actually, I realised later that theres no specific handling for this, and it works because they're HTML buttons and the updating of the selected tab happens on click. So probably no need for this as long as the click is tested (which it is) - but I'll leave the comment here in case you think it would be better to have.

src/components/tabs/__internal__/tabs-header/tabs-header.component.tsx
```diff
@@ -21,21 +32,50 @@ const TabsHeader = ({
...
+  const [rightScrollOpacity, setRightScrollOpacity] = useState(1);
+
+  const ref = useRef<HTMLDivElement>(null);
+
+  let isScrollable = false;
+
+  const { current } = ref;
+
+  if (position === "top" && current) {
+    isScrollable = current?.scrollWidth > current?.clientWidth;
```
robinzigmond, line 48:
```suggestion
    isScrollable = current.scrollWidth > current.clientWidth;
```

conditional chaining isn't necessary since we already know (from the `if`) that `current` exists

src/components/tabs/__internal__/tabs-header/tabs-header.component.tsx
```diff
@@ -21,21 +32,50 @@ const TabsHeader = ({
...
-        alternateStyling={alternateStyling}
         noRightBorder={noRightBorder}
         isInSidebar={isInSidebar}
+        onScroll={throttledHandleScroll}
+        leftScrollOpacity={leftScrollOpacity}
+        rightScrollOpacity={rightScrollOpacity}
+        isScrollable={isScrollable}
+        ref={ref}
       >
+        {position === "top" && <StyledTabsBottomBorder />}
```
robinzigmond, line 78: I'm curious as to why we're now using a `div` by itself to represet the bottom border, instead of just using CSS. I notice that previously it was done by a box-shadow on the StyledTabsHeaderList. The replacement looks similar to me, but the new styles I see are pretty complex including specific gradients, so I have to ask - did you check this with UX? (I guess Chromatic will tell us if anything has changed - but I'm interested to know the reason you felt you had to change this.)

src/components/tabs/__internal__/tabs-header/tabs-header.component.tsx
```diff
@@ -1,15 +1,26 @@
-import React from "react";
+import React, { useRef, useState } from "react";
 import {
   StyledTabsHeaderWrapper,
   StyledTabsHeaderList,
+  StyledTabsBottomBorder,
 } from "./tabs-header.style";
+import useThrottle from "../../../../hooks/__internal__/useThrottle";
+
+const fullOpacityThreshold = 128;
```
robinzigmond, line 9: I confess I don't really understand what this is doing and in particular why the  value is 128? `getOpacityValue` is calculating its input as a propertion of 128, but that input is actually the `scrollLeft` property of the element - which is a value in pixels. What's special about 128 pixels, and what are we actually trying to achieve here anyway?

src/components/tabs/__internal__/tabs-header/tabs-header.component.tsx
mkrds: 1. I had to change the `overflow-x` of `StyledTabsHeaderList` to `auto` to allow horizontal scrolling. 
2. That caused the 3px orange outline of the currently focused TabHeader to be clipped by the element boundaries. 
3. To amend that I added padding=3px and margin=-3px to the container which misaligned the `box-shadow` acting as a border. 
4. So to fix that I decided to create a custom element that looks like a border and can be positioned however I like.

src/components/tabs/tabs.test.js
mkrds: I've added tests for the enter keypress

src/components/tabs/__internal__/tabs-header/tabs-header.component.tsx
mkrds: One of the requirements of this ticket is to add shadows on the sides of the TabHeader when it is scrollable. There is a prototype of this linked in [this issue](https://github.com/Sage/design-system/issues/7). In this prototype, these shadows, particularly the way that they disappear, is handled by scrolling in another gradient with the color of the background of the Tabs. While this works in prototype, it won't work when implemented in a reusable component because Tabs might be used on various background colors. (Check this example https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/ - there are linear-gradients acting as a `shadow covers` and radial-gradients acting as the shadows themselves). 

Considering all above I had to come up with different solution, so I decided to tie the opacity of the radial-gradient shadow with the current scroll position. Tbh, the magic `128` number is kinda random, I came up with it trying to replicate the way that the shadow disappears in the prototype. I didn't run it with UX yet but this PR is meant to be reviewed by UX anyway.

robinzigmond: Thanks for the explanation! I'll let this pass and see what UX say about it - hopefully Chromatic will catch any future regressions on this.

robinzigmond (dismissed):

DipperTheDan (dismissed):

robinzigmond (approved):

ZhuoyuJin (approved):

DipperTheDan (approved):

robinzigmond (approved):

PR #5952, opened 2023-03-31
fix(message): add conditional padding based on the showCloseIcon props truth value

### LINKED ISSUE #5904: Remove padding-rigth on Message Component
### Desired behaviour

When using a [Message](https://carbon.sage.com/?path=/docs/message--default-story) component and close button is not shown, padding right should be equals than left.

### Current behaviour

When using a [Message](https://carbon.sage.com/?path=/docs/message--default-story) component and close button is not shown, this place is not fulled with text. Padding-right is maintained.
Captura de pantalla 2023-03-08 a las 12 37 28

### Suggested Solution

Maybe close button should not be position absolute.

https://codesandbox.io/s/inspiring-roentgen-zyxj6z

_No response_

### Confidentiality

tomdavies73:
fix #5904

### Proposed behaviour

This PR will introduce conditional padding, when `showCloseIcon` is `true`, css property `padding-right` will be `50px`, which is the current hard-coded pixel value which adds enough space for the close icon. However, when `false`, this will be changed to `20px`, ensuring padding is equal on both sides.

Screenshot 2023-03-31 at 13 00 35

### Current behaviour

Currently, the css property padding-right is hard coded at `50px` regardless of the `showCloseIcon` props truth value.

Screenshot 2023-03-31 at 13 00 24

### Checklist

#### QA

### Testing instructions


src/components/message/message-content/message-content.style.ts
```diff
@@ -1,7 +1,8 @@
 import styled from "styled-components";
 
-const MessageContentStyle = styled.div`
-  padding: 15px 50px 15px 20px;
+const MessageContentStyle = styled.div<{ showCloseIcon?: boolean }>`
```
nineteen88, line 3: It's not the end of the world because it's unlikely to ever change, but it's better to have one source of truth of your Typescript props. So, ideally you would import it from the base component and using Pick to grab the correct type. Happy to approve anyway because as I say, unlikely to ever change.

nineteen88 (approved):

src/components/message/message-content/message-content.style.ts
```diff
@@ -1,7 +1,8 @@
...
+  padding: 15px ${({ showCloseIcon }) => (showCloseIcon ? "50px" : "20px")} 15px
```
Parsium, line 4: Ideally we should use any specified design tokens from DS over hardcoded values, however I couldn't seem to find any relevant tokens in the design from DS, so happy to approve. 

We can always go back and add the tokens in at a later point once we can determine which tokens to use and where to use them :)

Parsium (dismissed):

src/components/message/message.stories.tsx
```diff
@@ -20,6 +20,25 @@ export const Default: ComponentStory<typeof Message> = () => {
...
+          showCloseIcon={false}
+          open={isOpen}
+          onDismiss={() => setIsOpen(false)}
+        >
+          A longer custom message which now shows the close icon is not rendered
+          and padding is equal on both sides
+        </Message>
+      </Box>
+    </>
+  );
```
Parsium, line 39: Given `onDismiss` is only called when the close icon is clicked, by hiding the icon its not possible to hide the message and access the "Open Message" button, so what do you think about this?:

```suggestion
  return (
      <Box width="600px">
        <Message
          showCloseIcon={false}
          open
        >
          A longer custom message which now shows the close icon is not rendered
          and padding is equal on both sides
        </Message>
      </Box>
  );
```

Visually this would achieve the same result but without having the inaccessible button.

Parsium (approved):

nineteen88 (dismissed):

src/components/message/message-content/message-content.spec.tsx
```diff
@@ -10,4 +12,26 @@ describe("MessageContent", () => {
       ).toMatchSnapshot();
     });
   });
+  describe("showCloseIcon prop checks", () => {
+    it.each([
+      [true, "50px"],
+      [false, "20px"],
+    ])(
```
divyajindel, line 19:
```suggestion
      [true, 50],
      [false, 20],
    ])(
```

src/components/message/message-content/message-content.spec.tsx
```diff
@@ -10,4 +12,26 @@ describe("MessageContent", () => {
...
+      "padding changes should be correct when showCloseIcon is %s",
+      (boolVal, paddingVal) => {
+        const wrapper = mount(
+          <MessageContent showCloseIcon={boolVal}>children</MessageContent>
+        );
+
+        assertStyleMatch(
+          {
+            padding: `15px ${paddingVal} 15px 20px`,
```
divyajindel, line 28:
```suggestion
            padding: `15px ${paddingVal}px 15px 20px`,
```

divyajindel (dismissed):

nineteen88 (approved):

Parsium (approved):

PR #5962, opened 2023-04-06
test(split button): refactor cypress accessibility tests

sianford:
### Proposed behaviour

Add accessibility Cypress tests for the Split Button component to use the new cypress-component-react framework for testing.

### Current behaviour

Old approach with separate test files and run jobs for accessibility tests.

### Checklist

#### QA

### Testing instructions

- [ ] Run npx cypress open --component and run the split-button.test.js file
- [ ] Check the split-button.test.js file passed
- [ ] Run npx cypress run --component to check none of the other *.test.js files have regressed
- [ ] Run npm run build-storybook -> npx sb extract -> npx http-server storybook-static -p 9001 and run npx cypress run --e2e and check the Split Button tests do not run.

### DIFF
--- a/src/components/split-button/split-button-test.stories.tsx
+++ b/src/components/split-button/split-button-test.stories.tsx
@@ -1,15 +1,16 @@
-import React from "react";
+import React, { useState } from "react";
 import { action } from "@storybook/addon-actions";
 import Button from "../button";
 import Box from "../box";
 import { ICONS } from "../icon/icon-config";
+import Dialog from "../dialog";
 import {
   SPLIT_BUTTON_ALIGNMENTS,
   SPLIT_BUTTON_ICON_POSITIONS,
   SPLIT_BUTTON_SIZES,
   SPLIT_BUTTON_THEMES,
 } from "./split-button.config";
-import SplitButton from "./split-button.component";
+import SplitButton, { SplitButtonProps } from "./split-button.component";
 
 export default {
   title: "Split Button/Test",
@@ -53,44 +54,31 @@ export default {
   },
 };
 
-/* eslint-disable react/prop-types */
 export const SplitButtonStory = ({
   buttonType,
-  dataElement,
-  dataRole,
-  text,
   subtext,
-  iconType,
-  iconPosition,
   ...args
-  /* eslint-enable */
-}) => {
-  return (
-    <Box height={400} mt={100} ml={100}>
-      <SplitButton
-        buttonType={buttonType}
-        data-element={dataElement}
-        data-role={dataRole}
-        text={text}
-        subtext={subtext}
-        iconType={iconType}
-        iconPosition={iconPosition}
-        {...args}
-        onClick={action("click")}
-      >
-        <Button {...args} onClick={action("click")}>
-          Example Button
-        </Button>
-        <Button {...args} onClick={action("click")}>
-          Example Button with long text
-        </Button>
-        <Button {...args} onClick={action("click")}>
-          Short
-        </Button>
-      </SplitButton>
-    </Box>
-  );
-};
+}: Partial<SplitButtonProps>) => (
+  <Box height={400} mt={100} ml={100}>
+    <SplitButton
+      buttonType={buttonType}
+      text="text"
+      subtext={subtext}
+      {...args}
+      onClick={action("click")}
+    >
+      <Button {...args} onClick={action("click")}>
+        Example Button
+      </Button>
+      <Button {...args} onClick={action("click")}>
+        Example Button with long text
+      </Button>
+      <Button {...args} onClick={action("click")}>
+        Short
+      </Button>
+    </SplitButton>
+  </Box>
+);
 
 SplitButtonStory.story = {
   name: "default",
@@ -106,4 +94,27 @@ SplitButtonStory.story = {
     text: "Example Split Button",
     subtext: "",
   },
+};
+
+export const SplitButtonList = ({ ...props }) => {
+  return (
+    <SplitButton text="default text" {...props}>
+      <Button>Button 1</Button>
+      <Button>Button 2</Button>
+      <Button>Button 3</Button>
+    </SplitButton>
+  );
+};
+
+export const SplitButtonNestedInDialog = () => {
+  const [isOpen, setIsOpen] = useState(true);
+  return (
+    <Dialog open={isOpen} onCancel={() => setIsOpen(false)} title="Dialog">
+      <SplitButton text="default text">
+        <Button>Button 1</Button>
+        <Button>Button 2</Button>
+        <Button>Button 3</Button>
+      </SplitButton>
+    </Dialog>
+  );
 };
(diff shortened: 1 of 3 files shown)


src/components/split-button/split-button-test.stories.tsx
```diff
@@ -53,44 +54,39 @@ export default {
   },
 };
 
-/* eslint-disable react/prop-types */
+type SplitButtonStoryArgs = {
+  buttonType?: SplitButtonProps["buttonType"];
+  textContent: string;
```
robinzigmond, line 59: I'm not sure what `textContent` is doing here - it's not used in the `SplitButtonStory` below (or any of the other stories), nor is it an accepted prop on `SplitButton`

src/components/split-button/split-button-test.stories.tsx
```diff
@@ -53,44 +54,39 @@ export default {
   },
 };
 
-/* eslint-disable react/prop-types */
+type SplitButtonStoryArgs = {
+  buttonType?: SplitButtonProps["buttonType"];
```
robinzigmond, line 58:
```suggestion
  buttonType: SplitButtonProps["buttonType"];
```

I don't think you need a `?` here as `buttonType` is already optional in `SplitButtonProps`.

src/components/split-button/split-button-test.stories.tsx
```diff
@@ -53,44 +54,39 @@ export default {
   },
 };
 
-/* eslint-disable react/prop-types */
+type SplitButtonStoryArgs = {
```
robinzigmond, line 57: Since these prop types are the same as in `SplitButton` (other than `textContent` which doesn't seem to be used - see below), I think it's better to write this as:

`type SplitButtonStoryArgs = Pick<SplitButtonProps, "buttonType" | "text" | "subtext">;`

to be explicit that we're just picking particular props from the `SplitButtonProps`. This also saves us having to update this type separately if we ever decide to change the type of one of these props (which is very unlikely, I admit!)

src/components/split-button/split-button-test.stories.tsx
```diff
@@ -53,44 +54,39 @@ export default {
...
+  buttonType?: SplitButtonProps["buttonType"];
+  textContent: string;
+  text: string;
+  subtext: string;
+};
```
robinzigmond, line 62: Do we definitely only need those specific 3/4 props? I notice the story previously used `iconType`, `iconPosition`, `dataElement` and `dataRole` which we're now excluding. (While they will in theory still be passed down as part of `args`, using this type will mean that the Typescript compiler won't allow them to be supplied.)

src/components/split-button/split-button.test.js
```diff
@@ -184,8 +163,8 @@ context("Tests for Split Button component", () => {
     });
   });
 
-  describe("pressing tab while SplitButton is open", () => {
-    it("should move focus to next child button and to second SplitButton when end of list reached", () => {
+  describe("pressing tab while Split Button is open", () => {
```
robinzigmond, line 166: I notice you've changed `SplitButton` to `Split Button` here and in many places - presumably because we were inconsistent. I'm all for consistency but I wonder if `SplitButton` should be the preferred name, as that's the actual component name.

src/components/split-button/split-button.test.js
```diff
@@ -499,3 +478,64 @@ context("Tests for Split Button component", () => {
     });
   });
 });
+
+describe("check accessibility for Split Button component", () => {
+  it("should pass accessibility tests for Split Button Default story", () => {
```
robinzigmond, line 483: I see we check accessibility on the `Default` story when the additional buttons are opened, as well as shut, which is great 👍 Just wondering if we should do the same on each of the other stories - at the moment you're only testing them when initially rendered, so presumably closed. Not saying we definitely should - it may be overkill, as these additional props shouldn't change the behaviour of the open buttons much or at all. But if we only test one way it might be better to test with them open than closed?

I'm not sure what the agreed QA practice is on this, feel free to ignore this!

src/components/split-button/split-button-test.stories.tsx
sianford: Hmm, I'm not sure. I just pulled this code out of the `split-button.test.js` file and into the `split-button-test.stories.tsx` file. I'll investigate.

src/components/split-button/split-button.test.js
sianford: Yeah, I changed that for consistency and pondered which way it should be done. I checked other multi-worded components and many had spaces in the description so I went with that, but on reflection it does make sense to make it the same as the component.

src/components/split-button/split-button-test.stories.tsx
robinzigmond: ah thanks, I hadn't noticed that 🤦 Worth investigating but probably not worth spending too long on, if we had this before I won't let it stop me approving!

src/components/split-button/split-button.test.js
robinzigmond: Yes, those are all good points. I realised shortly after I left this comment that this would still get QA reviewers, so I'm happy to wait to see what they say and go with the QA concensus 😄

src/components/split-button/split-button-test.stories.tsx
sianford: Removed.

src/components/split-button/split-button.test.js
sianford: More tests added for primary type and hovering on both sections of primary and secondary type buttons.

src/components/split-button/split-button-test.stories.tsx
```diff
@@ -53,44 +54,39 @@ export default {
...
-          Example Button with long text
-        </Button>
-        <Button {...args} onClick={action("click")}>
-          Short
-        </Button>
-      </SplitButton>
-    </Box>
-  );
-};
+}: SplitButtonStoryArgs) => (
```
DlgSHi, line 69: why we just can't use `Partial<SplitButtonProps>`?

src/components/split-button/split-button.test.js
DlgSHi: I agree with @sianford , those `accessibility` are enough IMO

src/components/split-button/split-button.test.js
DlgSHi: I agree with @robinzigmond , the original name of the component should be used like `SplitButton`, could you amend that @sianford ?

DlgSHi (changes requested):

src/components/split-button/split-button.test.js
sianford: Already done, I just hadn't pushed the changes. I was waiting for confirmation of those hover states ;)

src/components/split-button/split-button-test.stories.tsx
sianford: Changed :)

DlgSHi (approved):

robinzigmond (approved):

PR #5963, opened 2023-04-06
refactor(select): convert all select components to typescript

robinzigmond:
### Proposed behaviour

Select and all internal components/utils should be written in Typescript

### Current behaviour

All Select components and internal utils are written in Javascript

### Checklist

#### QA

### Testing instructions


src/components/select/__internal__/select-text/select-text.spec.tsx
```diff
@@ -1,7 +1,11 @@
 import { mount } from "enzyme";
 import React from "react";
 import { assertStyleMatch } from "../../../../__spec_helper__/test-utils";
-import SelectText from "./select-text.component";
+import SelectText, { SelectTextProps } from ".";
+
+function renderSelectText(props: Partial<SelectTextProps> = {}) {
```
nineteen88, line 6: Is the Partial function required? As far as I can see all Props are optional anyway?

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -161,10 +238,11 @@ const FilterableSelect = React.forwardRef(
...
     const setMatchingText = useCallback(
-      (newValue, isClosing) => {
-        const matchingOption = React.Children.toArray(children).find((child) =>
-          isExpectedOption(child, newValue)
-        );
+      (newValue, isClosing = false) => {
+        const matchingOption = React.Children.toArray(children).find(
+          (child) =>
+            React.isValidElement(child) && isExpectedOption(child, newValue)
+        ) as React.ReactElement | undefined;
```
nineteen88, line 245: Do we need the `| undefined` if we're casting it as React.ReactElement?

src/components/select/__internal__/select-text/select-text.spec.tsx
robinzigmond: fair point, looks like you're right - I think I needed this at some point but then must have later had to make one of the props optional which wasn't before

src/components/select/filterable-select/filterable-select.component.tsx
robinzigmond: I'll take a look later but I'm fairly sure we do, because the array `find` method can still return `undefined` if it doesn't find the element it's looking for. And with filterable select we've no control over what the user types so the `matchingOption` could well be `undefined`.

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -19,9 +20,70 @@ import useInputAccessibility from "../../../hooks/__internal__/useInputAccessibi
...
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-component"?: string;
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-element"?: string;
+  /** Identifier used for testing purposes, applied to the root element of the component. */
+  "data-role"?: string;
+  /** Child components (such as Option or OptionRow) for the SelectList */
+  children: React.ReactNode;
+  /** The default selected value(s), when the component is operating in uncontrolled mode */
+  defaultValue?: string | Record<string, unknown>;
```
nineteen88, line 40: I don't know the component, will need to have a little investigation but is it not possible to avoid unknown here (and the few other places)? Would it not be `string | number` for example?

robinzigmond: yeah that's a good point. I'll be honest that I don't really know if it's needed or not - I deliberately adjusted a few of the newer places which didn't have types before so that the types matched the existing props (to avoid TS errors). I mostly didn't want to change the existing TS types (in the `.d.ts` file) to avoid breaking any consumers' implementations, but on reflection you're probably right that this probably shouldn't work (and likely won't work, in the sense of functioning correctly), if the record values are of some strange type.

src/components/select/list-action-button/list-action-button.component.tsx
```diff
@@ -22,6 +31,11 @@ const ListActionButton = React.forwardRef(
         );
       }
 
+      // istanbul ignore if
+      if (!React.isValidElement(listActionButton)) {
```
nineteen88, line 35: We can't test for this?

src/components/select/list-action-button/list-action-button.spec.tsx
```diff
@@ -1,20 +1,35 @@
...
+) {
+  const componentProps = { ...props };
+  if (!componentProps.onListAction) {
+    componentProps.onListAction = () => {};
+  }
+
+  return renderer(
+    <ListActionButton {...(componentProps as ListActionButtonProps)} />
+  );
+}
```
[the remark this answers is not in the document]
nineteen88: Apologies, this now causes a knock on effect for line 22 `expect(renderListActionButton({}, TestRenderer.create)).toMatchSnapshot();` as this is using react-test-renderer rather than Enzyme. This is the issue with using `:any` as it's flagging this up now. I can look at how we can update this further if it seems worthwhile or we can ignore for now as it will all be replaced by RTL in the future I guess.

src/components/select/list-action-button/list-action-button.component.tsx
robinzigmond: We certainly could - but we weren't before, and I'm not convinced it adds much value. The coverage issue only came up when converting to TS because, despite the check for truthiness above, a ReactNode can in theory be a string, so TS complains at the `cloneElement` call below.

I'm happy to add a test if you and/or a second reviewer thinks it's worthwhile, but I'm really not convinced.

src/components/select/utils/highlight-part-of-text.tsx
```diff
@@ -13,28 +16,37 @@ function highlightPartOfText(text, partToHighlight) {
...
   if (typeof child === "string") {
     return highlightPartOfText(child, partToHighlight);
   }
+  /* istanbul ignore if */
+  if (!React.isValidElement(child)) {
+    return child;
+  }
   const highlightedChildren = React.Children.map(
     child.props.children,
     (grandChild) => highlightPartOfTextRecursive(grandChild, partToHighlight)
```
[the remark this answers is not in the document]
robinzigmond: thanks, I'll check this out when I'm back on the branch. I didn't see anything when running the `type-check` script (and any errors there would have failed the build, although a warning is possible I guess).

src/components/select/list-action-button/list-action-button.spec.tsx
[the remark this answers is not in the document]
robinzigmond: As you observe, we do this in pretty much all tests that are written like this, and it's because the `renderer` can typically be either `shallow` or `mount`, and for some reason there doesn't seem to be a way of typing that other than using `any`. (`typeof shallow | typeof mount` seems to not work, although I don't understand why.) Obviously here with using react-test-renderer as well there are probably even more issues.

I'm inclined to ignore it as we do this in many places, and as you say it will go away when we do the RTL refactor anyway - we'll be doing that in TS from the start so hopefully won't be so accepting of `any`.

However looking at your comment has made be realise that prettier has reformatted the code so that the `ignore-next-line` isn't actually ignoring the correct line any more! I'm surprised this isn't causing a TS error...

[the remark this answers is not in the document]
nineteen88: Good spot. And yeah it seems fair enough to be honest, particularly with the "imminent" refactor.

src/components/select/utils/get-next-child-by-text.ts
```diff
@@ -1,27 +1,22 @@
...
-
-  return arrayOfChildren[indexOfMatch];
+function findElementStartingWithText(
+  textToMatch: string,
+  list: React.ReactElement[]
+) {
+  return list.findIndex((child) => {
+    return (
+      child.props.text &&
+      child.props.text.toLowerCase().startsWith(textToMatch.toLowerCase())
```
DipperTheDan, line 10:
```suggestion
      child.props?.text.toLowerCase().startsWith(textToMatch.toLowerCase())
```

src/components/select/utils/get-next-child-by-text.ts
```diff
@@ -1,27 +1,22 @@
...
-  }
-
-  return arrayOfChildren[indexOfMatch];
+function findElementStartingWithText(
+  textToMatch: string,
+  list: React.ReactElement[]
+) {
+  return list.findIndex((child) => {
+    return (
+      child.props.text &&
```
DipperTheDan, line 9:
```suggestion

```

src/components/select/filterable-select/filterable-select.component.tsx
robinzigmond: So you're correct that the `| undefined` isn't required to make this compile, but I don't think we should use type-casts that we know are inaccurate - `.find` might fail to find a matching option, and result in `matchingOption` being `undefined`. And this has a potential danger because if casting to `ReactElement` TS won't complain if the `!matchingOption` check on the next line gets removed - whereas it does when it knows that it might be undefined.

I've found an alternative though - I can cast the `React.Children.toArray(children)` to an array of `ReactElement`s, which I think is the most accurate description of what's actually happening. Then TS correctly infers the type of `matchingOption` (including that it can be `undefined`) and there don't seem to be any further problems.

src/components/select/filterable-select/filterable-select.component.tsx
robinzigmond: so having looked into this a bit more - although in practice consumers are most likely only going to want to use simple types like strings and numbers here, the idea is to allow them to use arbitrary data, so I'm not comfortable restricing this unless there's a very good reason. And I don't really see that. `unknown` is a bit like `any` in that anything can be assigned to it, but it's a more type-safe version that basically forces us to acknowledge in our code that we don't know what it is (unlike with `any` where we can do whatever we like, with the risk of things blowing up at runtime). This feels like the right balance to me both for our internal component code and for consumers - but I'm happy to be talked around if you have different ideas.

src/components/select/list-action-button/list-action-button.component.tsx
robinzigmond: thanks Dan, I looked up the Draggable code that uses `cloneElement` and saw how you did it - it uses `&&`, basically as a substitute for `if`, but it works in such a way as to basically fool the coverage checker (there's nothing "inside" the if so technically all code does get run no matter what) 😆 Happy to copy that here as (as I argued above) I'm really not convinced this needs a test, and it's always nicer not to have to put `istanbul ignore`...

src/components/select/filterable-select/filterable-select.component.tsx
robinzigmond: actually on reflection this change isn't right, at least not on this place. The issue is that if `listActionButton` is something truthy which is still not a React Element - like a non-empty string - we want to just leave it as is, but the `React.isValidElement(listActionButton) && ...` will end up rendering nothing.

So I'm going to put this back to how I had it, as I think it's the only way to make all possible React Node types behave correctly here while also keeping TS happy. Let me know if you think I should add a test for this case and I will, but for now I'll keep it with the istanbul ignore comment.

src/components/select/list-action-button/list-action-button.component.tsx
robinzigmond: actually on reflection this change isn't right, at least not in this place. The issue is that if `listActionButton` is something truthy which is still not a React Element - like a non-empty string - we want to just leave it as is, but the `React.isValidElement(listActionButton) && ...` will end up rendering nothing.

So I'm going to put this back to how I had it, as I think it's the only way to make all possible React Node types behave correctly here while also keeping TS happy. Let me know if you think I should add a test for this case and I will, but for now I'll keep it with the istanbul ignore comment.

src/components/select/simple-select/simple-select.spec.tsx
```diff
@@ -874,11 +908,11 @@ describe("SimpleSelect", () => {
     });
 
     describe("when parent re-renders", () => {
-      const WrapperComponent = (props) => {
+      const WrapperComponent = () => {
         const mockRef = useRef();
 
         return (
-          <span change={props.change}>
```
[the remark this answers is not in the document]
robinzigmond: in a sense it isn't, but if you look at what the test is actually testing, it's checking that when the SimpleSelect is inside a parent component and that parent rerenders, that certain things still work correctly. While setting/updating non-existent props on a `span` is a little daft, it's the simplest way to do that kind of test and I believe it is telling us something, even though the use-case is completely unrealistic. But it definitely wouldn't be testing the same thing if the span was simply removed.

I'm also not sure I want to rewrite this test to something more realistic given the upcoming RTL refactor, but feel free to insist on it! 😆

src/components/select/utils/with-filter.hoc.tsx
```diff
@@ -35,24 +41,70 @@ const filterOptionRows = (optionRow, filterText) => {
   return null;
 };
 
-const filterChildren = (filterText) => (children) => {
+const filterChildren = (filterText: string) => (children: React.ReactNode) => {
   // eslint-disable-next-line consistent-return
```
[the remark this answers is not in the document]
robinzigmond: cheers - you're right, I don't either. I think I must have added the `return null` at the end later to get around another issue, and that made this unnecessary

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -19,9 +20,70 @@ import useInputAccessibility from "../../../hooks/__internal__/useInputAccessibi
...
+  /** A callback that is triggered when a user scrolls to the bottom of the list */
+  onListScrollBottom?: () => void;
+  /** If true the Component opens on focus */
+  openOnFocus?: boolean;
+  /** SelectList table header, should consist of multiple th elements.
+   * Works only in multiColumn mode
+   */
+  tableHeader?: React.ReactNode;
+  /** The selected value(s), when the component is operating in controlled mode */
+  value?: string | Record<string, unknown>;
```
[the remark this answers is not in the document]
robinzigmond: agreed. But as you can see I've decided to keep `defaultValue` as it is so the same can be done for `value`

src/components/select/multi-select/multi-select.component.tsx
```diff
@@ -257,27 +334,32 @@ const MultiSelect = React.forwardRef(
...
           };
         }
 
-        const { title } = pillProps;
-        const key = title + (matchingOption?.props.value || index);
+        const title = pillProps.title || "";
+        const key =
+          title +
+          ((React.isValidElement(matchingOption) &&
+            matchingOption.props.value) ||
```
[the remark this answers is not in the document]
robinzigmond: hmm that's weird - when I hover over `matchingOption.props` I'm getting type `any`, which explains why I'm getting no such error. Different TS or React versions defaulting to different types for the props of an unknown element, perhaps? Anyway, even though it works for me I don't like things being typed as `any` even though there's no error/warning for it here, so I'll see if I can sort this out (same on `filterable-select`) 👍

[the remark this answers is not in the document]
robinzigmond: seeing the same now thanks to help from Will - turns out my VSCode was out of date and not auto-updating due to me having downloaded it into the wrong place and never being told about it till I thought to try updating manually. I'll get all these errors sorted everywhere they crop up 👍

src/components/select/utils/highlight-part-of-text.tsx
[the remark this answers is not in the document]
robinzigmond: seeing it now after updating VSCode - will see if I can remove it

nineteen88 (approved):

DipperTheDan (dismissed):

DipperTheDan (approved):

nineteen88 (approved):

PR #5964, opened 2023-04-11
chore: update crypto-js to latest version (FE-5748)

Parsium:
### Proposed behaviour

Update `crypto-js` dependency to latest version.

### Current behaviour

Dependency is currently on an outdated version.

### Checklist


package.json
```diff
@@ -174,7 +174,7 @@
     "chalk": "^4.1.1",
     "ci-info": "^3.3.2",
     "classnames": "~2.2.6",
-    "crypto-js": "~3.3.0",
+    "crypto-js": "^4.1.1",
```
nineteen88, line 177: Think I'm happy with changing this to caret to pick up the minor releases. Good change.

nineteen88 (approved):

nineteen88:
4.0.0 did introduce a breaking change, I'm not sure what kind of checks we can do regarding the regression required there/whether we need QA to take a look, as I'm not 100% sure what cryptojs is used for.

Parsium:
> 4.0.0 did introduce a breaking change, I'm not sure what kind of checks we can do regarding the regression required there/whether we need QA to take a look, as I'm not 100% sure what cryptojs is used for.

We use [the `MD5` hash function from `crypto-js` in `Portrait`](https://github.com/Sage/carbon/blob/master/src/components/portrait/portrait-gravatar.component.tsx#L31) to support the use of [gravatars](https://en.gravatar.com/).
 If the user has a registered gravatar we can create a URL to it so it will be displayed within `Portrait`

PR #5971, opened 2023-04-14
 feat(minor-button): add position: absolute property to icon when component is icon only

### LINKED ISSUE #5946: Button Minor with Tooltip - Icon moves up vertically when hovered
### Current behaviour

Tooltip Icon moves up vertically when hovered.

Screenshot 2023-03-28 at 17 57 39

Screenshot 2023-03-28 at 17 57 47

### Expected behaviour

Tooltip Icon should remain in the same position when Button Minor is hovered over.

https://carbon.sage.com/?path=/docs/button-minor--primary-button#icon-only-button

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

Latest

### Design Tokens Version

_No response_

Firefox, Chrome, Safari

MacOS

_No response_

### Confidentiality

tomdavies73:
fix #5946

### Proposed behaviour

This PR will add position: absolute CSS property to `Icon` when component is icon only. This prevents the rendered `Icon` from moving position on hover due to a reduction in padding added during `ButtonMinor`s composition, which causes the `Icon` and Portal used to to provide a tooltip message to stack on top of one another

### Current behaviour

Currently, the `Icon` rendered within the component and Portal used to to provide a tooltip message stack on top of one another, this is due to the rendred `Icon` having relative positioning, which is needed for other variants of the button and a reduction in padding implemented during the components composition.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/cypress/components/button-minor/button-minor.test.js
+++ b/cypress/components/button-minor/button-minor.test.js
@@ -94,6 +94,11 @@ context("Test for Button Minor component", () => {
       }
     );
 
+    it("when icon only, icon's position is absolute", () => {
+      CypressMountWithProviders(<ButtonMinor iconType="bin" />);
+      icon().should("have.css", "position", "absolute");
+    });
+
     it.each([
       [BUTTON_SIZES[0], 32],
       [BUTTON_SIZES[1], 40],
--- a/src/components/button-minor/button-minor.spec.tsx
+++ b/src/components/button-minor/button-minor.spec.tsx
@@ -6,6 +6,7 @@ import {
   ButtonTypes,
   SizeOptions,
 } from "../button/button.component";
+import StyledIcon from "../icon/icon.style";
 import ButtonMinor from "./button-minor.component";
 import { assertStyleMatch } from "../../__spec_helper__/test-utils";
 
@@ -87,6 +88,28 @@ describe("Button Minor", () => {
       assertStyleMatch({ ...styles }, wrapper);
     }
   );
+
+  it("with icon and text children, icon's position is undefined", () => {
+    const wrapper = mount(<ButtonMinor iconType="bin">Foo </ButtonMinor>);
+    assertStyleMatch(
+      {
+        position: undefined,
+      },
+      wrapper,
+      { modifier: `${StyledIcon}` }
+    );
+  });
+
+  it("when icon only, icon's position is absolute", () => {
+    const wrapper = mount(<ButtonMinor iconType="bin" />);
+    assertStyleMatch(
+      {
+        position: "absolute",
+      },
+      wrapper,
+      { modifier: `${StyledIcon}` }
+    );
+  });
 
   it("when destructive prop is passed, renders with destructive styling", () => {
     const wrapper = mount(<ButtonMinor destructive>foo</ButtonMinor>);
--- a/src/components/button-minor/button-minor.style.ts
+++ b/src/components/button-minor/button-minor.style.ts
@@ -17,6 +17,14 @@ function makeColors(color: string) {
 }
 
 const StyledButtonMinor = styled(Button)`
+  ${({ children }) =>
+    !children &&
+    css`
+      ${StyledIcon} {
+        position: absolute;
+      }
+    `}
+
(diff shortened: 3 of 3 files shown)


src/components/button-minor/button-minor.style.ts
```diff
@@ -55,24 +53,24 @@ const StyledButtonMinor = styled(Button)`
...
     `}
 
-  ${({ size }) => css`
+  ${({ size, iconTooltipMessage }) => css`
     ${size === "small" &&
     `
         min-height: 32px;
-        padding: var(--spacing000) var(--spacing100) var(--spacing000)
-          var(--spacing100);
+        padding-left: ${iconTooltipMessage ? "none" : "var(--spacing100)"};
```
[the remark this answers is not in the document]
robinzigmond: I actually don't like this approach at all. We shouldn't have to set different styles depending on whether an `iconTooltipMessage` prop is passed in or not - after all, just setting the prop doesn't force the tooltip to appear, as you also have to hover over the button.

Since it seems that this left and right padding is never actually needed, I'd prefer to just remove the styles entirely (assuming that doesn't cause any visual changes - if it breaks tests we can change those, as you've already done).

I've just been looking to see what is causing the icon to jump when the `<span data-portal-entrace>` appears - I haven't figured it out yet but I think it's something to do with the flex styles on the parent button. It's weird that it jumps verticall when *left* and *right* padding are added to the <span> with the icon in it - especially as it only happens when you add both, not just one. And of course the additional child, the portal-entrance span, has a width and height of 0. I'm happy to help out with this if you want me to - but I also think simply removing the padding in all circumstances is a good enough solution unless I've missed some reason why it's sometimes needed. I definitely don't like conditionally removing it only when there's potential for a tooltip to be shown.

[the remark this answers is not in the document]
tomdavies73: I agree, I've went for an absolute position approach on the rendered `Icon` opposed to the conditional padding approach 👍

src/components/button-minor/button-minor.style.ts
```diff
@@ -17,6 +17,16 @@ function makeColors(color: string) {
 }
 
 const StyledButtonMinor = styled(Button)`
+  ${({ iconType, iconTooltipMessage, children }) =>
+    iconType &&
+    iconTooltipMessage &&
```
robinzigmond, line 22: I don't think we need all these conditions.

I see that we need to check for an icon-only button (with the `!children` check), otherwise the icon overlaps the text when both are present. But `iconTooltipMessage` makes no difference here (as far as I can tell), and `iconType` isn't needed simply because without that there's no `StyledIcon` inside for the styles to affect.

I prefer it checking just this one condition as it makes the code simpler with fewer special cases.

src/components/button-minor/button-minor.spec.tsx
```diff
@@ -88,6 +89,19 @@ describe("Button Minor", () => {
     }
   );
 
+  it("when icon only with tooltip message, icon's position is absolute", () => {
+    const wrapper = mount(
+      <ButtonMinor iconType="bin" iconTooltipMessage="foo" />
```
robinzigmond, line 94: If you make the code change I suggested,the `iconTooltipMessage` prop won't be needed in this prop (and we can leave it out of the test description).

cypress/components/button-minor/button-minor.test.js
```diff
@@ -94,6 +94,13 @@ context("Test for Button Minor component", () => {
       }
     );
 
+    it("when icon only with tooltip message, icon's position is absolute", () => {
+      CypressMountWithProviders(
+        <ButtonMinor iconType="bin" iconTooltipMessage="foo" />
```
robinzigmond, line 99: If you make the code change I suggested,the `iconTooltipMessage` prop won't be needed in this prop (and we can leave it out of the test description).

Parsium (approved):

robinzigmond (approved):

ZhuoyuJin (approved):

PR #5973, opened 2023-04-14
test(tooltip): add accessibility tests

stephenogorman:
### Proposed behaviour
- Add `accessibility` Cypress tests for the `Tooltip` component to use the new cypress-component-react framework for testing.
- Refactor `tooltip.stories.test.mdx` to corresponding `*.tsx`file.

### Current behaviour
Old approach for accessibility tests

### Checklist

#### QA

### Testing instructions

- [x] Check if the `tooltip.test.js` file passed
- [x] Run `npm run build-storybook` -> `npx sb extract` -> `npx http-server storybook-static -p 9001` and run `npx cypress run --e2e` and check that `tooltip` tests are not running twice.


src/components/tooltip/tooltip-test.stories.tsx
```diff
@@ -0,0 +1,196 @@
+/* eslint-disable react/prop-types */
+/* eslint-disable react/button-has-type */
+import React, { forwardRef, useState } from "react";
+import Tooltip from ".";
+import { TOOLTIP_POSITIONS, TooltipPositions } from "./tooltip.config";
+import { ButtonProps } from "../button";
+
+export default {
+  title: "Tooltip/Test",
+  includeStories: "DefaultStory",
```
DlgSHi, line 10:
```suggestion
  includeStories: ["DefaultStory"],
```

DlgSHi (changes requested):

DlgSHi (approved):

src/components/tooltip/tooltip-test.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+  isVisible?: boolean;
+  flipOverrides?: TooltipPositions;
+}
+
+export const DefaultStory = ({
+  message,
+  isVisible,
+  flipOverrides,
+  ...args
+}: TooltipStoryProps) => {
```
Parsium, line 65: Currently the `args` object has an empty type `{}`, because `DefaultStory` is typed to only expect any of the three properties: `message`, `isVisible` and/or `flipOverrides`.

To type `DefaultStory` to expect the `Tooltip` props to be passed too, we need to amend the type definition slightly like this:

```suggestion
}: TooltipProps & TooltipStoryProps) => {
```

src/components/tooltip/tooltip-test.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+);
+SecondaryButton.displayName = "Tooltip";
+
+export const TooltipComponent = ({ ...props }) => (
+  <div
+    style={{
+      padding: "60px 60px 60px 160px",
+    }}
+  >
+    <Tooltip message="I am a tooltip!" isVisible {...props}>
```
Parsium, line 163: Similarly here, we need to type the incoming props as `Tooltip` props here.

```suggestion
export const TooltipComponent = ({ message, ...rest }: TooltipProps) => (
  <div
    style={{
      padding: "60px 60px 60px 160px",
    }}
  >
    <Tooltip message={message || "I am a tooltip!"} isVisible {...props}>
```

Since `message` can be passed but has a default value of "I am tooltip", we also need the ability to override its value if `message` has passed to `TooltipComponent` from the cypress tests.

src/components/tooltip/tooltip-test.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+      control: {
+        type: "select",
+      },
+    },
+    flipOverrides: {
+      options: ["left", "right", "top", "bottom"],
+      control: {
+        type: "select",
+      },
+    },
```
Parsium, line 34: Currently, if we were to pass `flipOverrides` directly to `Tooltip` within `DefaultStory`, we would get a type error, since `flipOverrides` passed into the story has the type `TooltipPositions` and not `TooltipPositions[]`.

To fix this, we need to map whatever value we select via the storybook controls into an array like so:

```suggestion
flipOverrides: {
      mapping: (value: TooltipPositions) => [value],
      options: ["left", "right", "top", "bottom"],
      control: {
        type: "select",
      },
    },
```

Which means the type of `flipOverrides` will correctly be `TooltipPositions[]` and not `TooltipPositions`.

src/components/tooltip/tooltip-test.stories.tsx
stephenogorman: Should line 163 be:
<Tooltip message="I am a tooltip!" isVisible {...rest}>

Parsium (approved):

DlgSHi (approved):

PR #5996, opened 2023-04-27
feat(action-popover): align component with design system 

### LINKED ISSUE #5801: ActionPopover items are not aligned if some items have icons while others don't
### Current behaviour

When an `ActionPopover` menu renders with a mix of items with and without icons then the labels are not aligned.

image

### Expected behaviour

When an `ActionPopover` menu renders with a mix of items with and without icons then the labels should be aligned and the icon's space should be preserved on the items without icons.

https://codesandbox.io/s/affectionate-sound-cqi2fv?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

114.7.1

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

tomdavies73:
fix #5801

### Proposed behaviour

Nine new design variations have been implemented for the component, each one varies based on specific props and attributes passed to individual children and siblings.

**Please see new designs below:**

**Number 1: All children and siblings have no icons or submenus.**

Screenshot 2023-06-21 at 13 45 18

**Number 2: All children and siblings have no icons, but only some have submenus.**

Screenshot 2023-06-21 at 13 45 26

**Number 3: All children and siblings have no icons, but all have submenus.**

Screenshot 2023-06-21 at 13 45 33

**Number 4: All children and siblings have no submenus, but only some have icons.**

Screenshot 2023-06-21 at 13 45 40

**Number 5: All children and siblings have no submenus, but all have icons.**

Screenshot 2023-06-21 at 13 45 47

**Number 6: All children and siblings have submenus and icons.**

Screenshot 2023-06-21 at 13 46 06

**Number 7: All children and siblings have icons, but only some have submenus.**

Screenshot 2023-06-21 at 13 45 54

**Number 8: All children and siblings have submenus, but only some have icons.**

Screenshot 2023-06-21 at 13 46 13

**Number 9: Some children and siblings have submenus and icons, but some have no submenus or icons.**

Screenshot 2023-06-21 at 13 46 20

Please note, these designs also have various versions based on the `horizontalAlignment` and `submenuPosition` props, so in theory there are technically 36 new designs. A codeSandbox has been included [here](https://codesandbox.io/s/carbon-quickstart-forked-kctlch?file=/src/App.js) to test the new designs and variations. 

**Support for icon's in submenu's:**

Screenshot 2023-06-21 at 13 46 57

Screenshot 2023-06-21 at 13 47 07

**`submenuPosition` is temporarily switched from "left" to "right" or "right" to "left" when the container becomes too small for submenu:**

### Current behaviour

Currently, the `actionPopover` component is severely unaligned with Sage's design system, meaning specific designs and functionality seen within the design system, can not be replicated using the Carbon component. Currently, there is no support for any submenus to open to the right, and no support for icon's to be rendered within submenus also. Both of these are fairly major misalignments and could make a large difference to how the component is used and implemented.

Also, how text, and icons are aligned based on icon's and submenus being present amongst children and siblings is inconsistent, and also deviates from current design system guidance and best practices.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/action-popover/action-popover-item/action-popover-item.component.tsx
+++ b/src/components/action-popover/action-popover-item/action-popover-item.component.tsx
@@ -11,11 +11,13 @@ import {
   MenuItemIcon,
   SubMenuItemIcon,
   StyledMenuItem,
+  StyledMenuItemInnerText,
+  StyledMenuItemOuterContainer,
   StyledMenuItemWrapper,
 } from "../action-popover.style";
 import Events from "../../../__internal__/utils/helpers/events";
 import createGuid from "../../../__internal__/utils/helpers/guid";
-import ActionPopoverContext from "../action-popover-context";
+import ActionPopoverContext, { Alignment } from "../action-popover-context";
 import useLocale from "../../../hooks/__internal__/useLocale";
 
 import { IconType } from "../../icon";
@@ -47,55 +49,55 @@ export interface ActionPopoverItemProps {
   /** @ignore @private */
   focusItem?: boolean;
   /** @ignore @private */
-  horizontalAlignment?: "left" | "right";
+  horizontalAlignment?: Alignment;
+  /** @ignore @private */
+  childHasSubmenu?: boolean;
+  /** @ignore @private */
+  childHasIcon?: boolean;
+  /** @ignore @private */
+  currentSubmenuPosition?: Alignment;
+  /** @ignore @private */
+  setChildHasSubmenu?: (value: boolean) => void;
+  /** @ignore @private */
+  setChildHasIcon?: (value: boolean) => void;
+  /** @ignore @private */
+  setCurrentSubmenuPosition?: (value: Alignment) => void;
+  /** @ignore @private */
+  isASubmenu?: boolean;
 }
 
 const INTERVAL = 150;
 
 type ContainerPosition = {
-  left: number;
+  left: string | number;
   top?: string;
   bottom?: string;
-  right: "auto";
+  right: string | number;
 };
 
 function checkRef(ref: React.RefObject<HTMLElement>) {
   return Boolean(ref && ref.current);
 }
 
-function leftAlignSubmenu(
+function calculateSubmenuPosition(
   ref: React.RefObject<HTMLElement>,
-  submenuRef: React.RefObject<HTMLElement>
+  submenuRef: React.RefObject<HTMLElement>,
+  submenuPosition: Alignment,
+  currentSubmenuPosition?: Alignment
 ) {
   /* istanbul ignore if */
-  if (!ref.current || !submenuRef.current) return true;
 
-  const { left } = ref.current.getBoundingClientRect();
+  if (!ref.current || !submenuRef.current)
+    return currentSubmenuPosition || submenuPosition;
+
+  const { left, right } = ref.current.getBoundingClientRect();
   const { offsetWidth } = submenuRef.current;
+  const windowWidth = document.body.clientWidth;
 
-  return left >= offsetWidth;
-}
-
-function getContainerPosition(
-  itemRef: React.RefObject<HTMLElement>,
-  submenuRef: React.RefObject<HTMLElement>,
-  placement: "bottom" | "top"
-): ContainerPosition | undefined {
-  /* istanbul ignore if */
-  if (!itemRef.current || !submenuRef.current) return undefined;
-
-  const { offsetWidth: parentWidth } = itemRef.current;
-  const { offsetWidth: submenuWidth } = submenuRef.current;
-  const xPositionValue = leftAlignSubmenu(itemRef, submenuRef)
-    ? -submenuWidth
-    : parentWidth;
-  const yPositionName = placement === "top" ? "bottom" : "top";
-
-  return {
-    left: xPositionValue,
-    [yPositionName]: "calc(-1 * var(--spacing100))",
-    right: "auto",
-  };
+  if (submenuPosition === "left") {
+    return left >= offsetWidth ? "left" : "right";
+  }
+  return windowWidth >= right + offsetWidth ? "right" : "left";
 }
 
 export const ActionPopoverItem = ({
@@ -109,6 +111,13 @@ export const ActionPopoverItem = ({
   download,
   href,
   horizontalAlignment,
+  childHasSubmenu,
+  childHasIcon,
+  currentSubmenuPosition,
+  setChildHasSubmenu,
+  setChildHasIcon,
+  setCurrentSubmenuPosition,
+  isASubmenu = false,
   ...rest
 }: ActionPopoverItemProps) => {
   const l = useLocale();
@@ -124,7 +133,12 @@ export const ActionPopoverItem = ({
     "ActionPopoverItem only accepts submenu of type `ActionPopoverMenu`"
   );
 
-  const { setOpenPopover, isOpenPopover, focusButton } = context;
+  const {
+    setOpenPopover,
+    isOpenPopover,
+    focusButton,
+    submenuPosition,
+  } = context;
   const isHref = !!href;
   const [containerPosition, setContainerPosition] = useState<
     ContainerPosition | undefined
@@ -132,7 +146,6 @@ export const ActionPopoverItem = ({
   const [guid] = useState(createGuid());
   const [isOpen, setOpen] = useState(false);
   const [focusIndex, setFocusIndex] = useState<number>(0);
-  const [isLeftAligned, setIsLeftAligned] = useState(true);
 
   const submenuRef = useRef<HTMLDivElement>(null);
   const ref = useRef<HTMLButtonElement>(null);
@@ -145,21 +158,65 @@ export const ActionPopoverItem = ({
     }
   }, [isOpenPopover]);
 
-  const alignSubmenu = useCallback(() => {
-    if (checkRef(ref) && checkRef(submenuRef) && submenu) {
-      const align = leftAlignSubmenu(ref, submenuRef);
-      setIsLeftAligned(align);
-      setContainerPosition(getContainerPosition(ref, submenuRef, placement));
+  useEffect(() => {
+    if (icon) {
+      setChildHasIcon?.(true);
     }
-  }, [submenu, placement]);
+    if (submenu) {
+      setChildHasSubmenu?.(true);
+    }
+  }, [icon, setChildHasSubmenu, setChildHasIcon, submenu]);
+
+  const alignSubmenu = useCallback(() => {
+    const checkCalculatedSubmenuPosition = calculateSubmenuPosition(
+      ref,
+      submenuRef,
+      submenuPosition,
+      currentSubmenuPosition
+    );
+
+    setCurrentSubmenuPosition?.(checkCalculatedSubmenuPosition);
+
+    return checkRef(ref) && checkRef(submenuRef) && submenu;
+  }, [
+    submenu,
+    setCurrentSubmenuPosition,
+    submenuPosition,
+    currentSubmenuPosition,
+  ]);
 
   useEffect(() => {
-    alignSubmenu();
+    const getContainerPosition = () => {
+      /* istanbul ignore if */
+      if (!ref.current || !submenuRef.current) return undefined;
 
+      const { offsetWidth: submenuWidth } = submenuRef.current;
+
+      const leftAlignedSubmenu = currentSubmenuPosition === "left";
+      const leftValue = leftAlignedSubmenu ? -submenuWidth : "auto";
+      const rightValue = leftAlignedSubmenu ? "auto" : -submenuWidth;
+      const yPositionName = placement === "top" ? "bottom" : "top";
+
+      return {
+        left: leftValue,
+        [yPositionName]: "calc(-1 * var(--spacing100))",
+        right: rightValue,
+      };
+    };
+    setContainerPosition(getContainerPosition);
+  }, [submenu, currentSubmenuPosition, placement]);
+
+  useEffect(() => {
+    if (submenu) {
+      alignSubmenu();
+    }
+  }, [alignSubmenu, submenu]);
+
+  useEffect(() => {
     if (focusItem) {
       ref.current?.focus();
     }
-  }, [alignSubmenu, focusItem]);
+  }, [focusItem]);
 
   useEffect(() => {
     return function cleanup() {
@@ -205,7 +262,7 @@ export const ActionPopoverItem = ({
         e.stopPropagation();
       } else if (!disabled) {
         if (submenu) {
-          if (isLeftAligned) {
+          if (currentSubmenuPosition === "left") {
             // LEFT: open if has submenu and left aligned otherwise close submenu
             if (Events.isLeftKey(e) || Events.isEnterKey(e)) {
               setOpen(true);
@@ -242,7 +299,7 @@ export const ActionPopoverItem = ({
         e.stopPropagation();
       }
     },
-    [disabled, download, isHref, isLeftAligned, onClick, submenu]
+    [disabled, download, isHref, onClick, submenu, currentSubmenuPosition]
   );
 
   const itemSubmenuProps = {
@@ -283,7 +340,21 @@ export const ActionPopoverItem = ({
   };
 
   const renderMenuItemIcon = () => {
-    return icon && <MenuItemIcon as={undefined} type={icon} />;
+    return (
+      icon && (
+        <MenuItemIcon
+          type={icon}
+          data-element="action-popover-menu-item-icon"
+          horizontalAlignment={horizontalAlignment}
+          submenuPosition={currentSubmenuPosition}
+          childHasIcon={childHasIcon}
+          childHasSubmenu={childHasSubmenu}
+          hasIcon={!!icon}
+          hasSubmenu={!!submenu}
+          isASubmenu={isASubmenu}
+        />
+      )
+    );
   };
 
   return (
@@ -298,18 +369,40 @@ export const ActionPopoverItem = ({
           tabIndex={0}
           isDisabled={disabled}
           horizontalAlignment={horizontalAlignment}
+          submenuPosition={currentSubmenuPosition}
+          hasSubmenu={!!submenu}
+          childHasSubmenu={childHasSubmenu}
           {...(disabled && { "aria-disabled": true })}
           {...(isHref && { as: ("a" as unknown) as undefined, download, href })}
           {...(submenu && itemSubmenuProps)}
         >
-          {submenu && checkRef(ref) && isLeftAligned ? (
-            <SubMenuItemIcon type="chevron_left" />
+          {submenu && checkRef(ref) && currentSubmenuPosition === "left" ? (
+            <SubMenuItemIcon
+              data-element="action-popover-menu-item-chevron"
+              type="chevron_left_thick"
+            />
           ) : null}
-          {horizontalAlignment === "left" ? renderMenuItemIcon() : null}
-          {children}
-          {horizontalAlignment === "right" ? renderMenuItemIcon() : null}
-          {submenu && checkRef(ref) && !isLeftAligned ? (
-            <SubMenuItemIcon type="chevron_right" />
+          <StyledMenuItemOuterContainer>
+            {horizontalAlignment === "left" ? renderMenuItemIcon() : null}
+            <StyledMenuItemInnerText
+              data-element="action-popover-menu-item-inner-text"
+              horizontalAlignment={horizontalAlignment}
+              submenuPosition={currentSubmenuPosition}
+              isASubmenu={isASubmenu}
+              childHasSubmenu={childHasSubmenu}
+              childHasIcon={childHasIcon}
+              hasIcon={!!icon}
+              hasSubmenu={!!submenu}
+            >
+              {children}
+            </StyledMenuItemInnerText>
+            {horizontalAlignment === "right" ? renderMenuItemIcon() : null}
+          </StyledMenuItemOuterContainer>
+          {submenu && checkRef(ref) && currentSubmenuPosition === "right" ? (
+            <SubMenuItemIcon
+              data-element="action-popover-menu-item-chevron"
+              type="chevron_right_thick"
+            />
           ) : null}
         </StyledMenuItem>
         {React.isValidElement(submenu)
@@ -325,6 +418,8 @@ export const ActionPopoverItem = ({
                 setOpen,
                 setFocusIndex,
                 focusIndex,
+                isASubmenu: true,
+                horizontalAlignment,
               }
             )
           : null}
--- a/src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx
+++ b/src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useContext } from "react";
+import React, { useCallback, useMemo, useContext, useState } from "react";
 import invariant from "invariant";
 
 import { Menu } from "../action-popover.style";
@@ -7,7 +7,7 @@ import ActionPopoverItem, {
   ActionPopoverItemProps,
 } from "../action-popover-item/action-popover-item.component";
 import ActionPopoverDivider from "../action-popover-divider/action-popover-divider.component";
-import ActionPopoverContext from "../action-popover-context";
+import ActionPopoverContext, { Alignment } from "../action-popover-context";
 
 export interface ActionPopoverMenuBaseProps {
   /** Children for the menu */
@@ -25,19 +25,21 @@ export interface ActionPopoverMenuBaseProps {
   /** Unique ID for the menu's parent */
   parentID?: string;
   /** Horizontal alignment of menu items content */
-  horizontalAlignment?: "left" | "right";
+  horizontalAlignment?: Alignment;
   /** Set whether the menu should open above or below the button */
   placement?: "bottom" | "top";
   /** @ignore @private */
   role?: string;
   /** @ignore @private */
+  isASubmenu?: boolean;
+  /** @ignore @private */
   "data-element"?: string;
   /** @ignore @private */
   style?: {
-    left: number;
+    left: string | number;
     top?: string;
     bottom?: string;
-    right: "auto";
+    right: string | number;
   };
 }
 
@@ -60,6 +62,7 @@ const ActionPopoverMenu = React.forwardRef<
       setFocusIndex,
       placement = "bottom",
       horizontalAlignment,
+      isASubmenu,
       ...rest
     }: ActionPopoverMenuBaseProps,
     ref
@@ -69,7 +72,7 @@ const ActionPopoverMenu = React.forwardRef<
       context,
       "ActionPopoverMenu must be used within an ActionPopover component"
     );
-    const { focusButton } = context;
+    const { focusButton, submenuPosition } = context;
 
     invariant(
       setOpen && setFocusIndex && typeof focusIndex !== "undefined",
@@ -164,6 +167,13 @@ const ActionPopoverMenu = React.forwardRef<
       [focusButton, setOpen, focusIndex, items, setFocusIndex]
     );
 
+    const [childHasSubmenu, setChildHasSubmenu] = useState(false);
+    const [childHasIcon, setChildHasIcon] = useState(false);
+    const [
+      currentSubmenuPosition,
+      setCurrentSubmenuPosition,
+    ] = useState<Alignment>(submenuPosition);
+
     const clonedChildren = useMemo(() => {
       let index = 0;
       return React.Children.map(children, (child) => {
@@ -175,13 +185,30 @@ const ActionPopoverMenu = React.forwardRef<
               focusItem: isOpen && focusIndex === index - 1,
               placement: child.props.submenu ? placement : undefined,
               horizontalAlignment,
+              childHasSubmenu,
+              setChildHasSubmenu,
+              childHasIcon,
+              setChildHasIcon,
+              currentSubmenuPosition,
+              setCurrentSubmenuPosition,
+              isASubmenu,
             }
           );
         }
 
         return child;
       });
-    }, [children, focusIndex, isOpen, placement, horizontalAlignment]);
+    }, [
+      children,
+      focusIndex,
+      isOpen,
+      placement,
+      horizontalAlignment,
+      childHasSubmenu,
+      childHasIcon,
+      currentSubmenuPosition,
+      isASubmenu,
+    ]);
 
     return (
       <Menu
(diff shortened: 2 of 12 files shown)


src/components/action-popover/action-popover.style.ts
```diff
@@ -105,9 +105,17 @@ const StyledButtonIcon = styled.div`
   }
 `;
 
-const MenuItemIcon = styled(Icon)`
-  padding: var(--spacing100);
+const MenuItemIcon = styled(Icon)<StyledMenuItemProps>`
```
robinzigmond, line 108: I'm not sure I like the use of `StyledMenuItemProps` here, as the props aren't the same - `StyledMenuItemProps` also has `isDisabled` which has nothing to do with `MenuItemIcon`.

Better imo would be to type it as either `{ horizontalAlignment: "left" | "right" }`, or perhaps `Pick<StyledMenuItemProps, "horizontalAlignment">` if we're sure that the two styled components should always have the same type for that prop (which they probably should). I'm also happy with making an alias (`StyledMenuItemIconProps` or similar) for either of these options (not too bothered either way) - but I don't think it's right to use the whole of `StyledMenuItemProps`.

robinzigmond (commented):
I expect you're aware, but I thought I'd mention these things just in case you're not. I see you've marked this PR to go through UX QA so happy to wait to see what Harpal says, but these should probably be brought to his attention:

1) although it looks a lot better than it did before, there's still a small gap on the left when the icon's there, which isn't there for just text. (See screenshot - this seems to be due to the flex with `justify-content: center` applied to the icon, whereas the text has no such thing so aligns perfectly to the left.)

image

2) by changing the padding in all cases, this causes visual changes in all uses of this component with icons, even those unaffected by the bug. In particular, when submenus are involved, I think the gap between the icon and the submemu arrow looks much too small (see screenshot below) - but as I'm not a designer I'm happy to leave this for Harpal's judgement. 
image

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -274,7 +274,16 @@ export const ActionPopoverItem = ({
   };
 
   const renderMenuItemIcon = () => {
-    return icon && <MenuItemIcon as={undefined} type={icon} />;
+    return (
+      icon && (
+        <MenuItemIcon
+          data-element="action-popover-menu-item-icon"
+          horizontalAlignment={horizontalAlignment}
+          as={undefined}
```
Parsium, line 282: I don't believe this prop does anything to the icon markup, so I would suggest we remove it.

```suggestion
```

I am assuming this was added to address some typescript errors? If so, we could work around this by following @robinzigmond's suggestion regarding prop types and introducing a default value for the `horizontalAlignment` prop.

Parsium:
I agree with @robinzigmond regarding the gap between the chevron and icon gap for items with a submenu. Looking at the DS docs it seems we only have a design for right-aligned chevrons when the submenu is opened on the right:

Screenshot 2023-05-04 at 10 47 20

But we don't have a left-aligned variant when the submenu is opened on the left, might be worth getting DS's/Harpal's thoughts on this once this is in UX QA :)

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+    return 1;
+  };
+
   useEffect(() => {
     if (!isOpenPopover) {
       setOpen(false);
     }
   }, [isOpenPopover]);
 
+  useEffect(() => {
```
robinzigmond, line 215: **suggestion**: as far as I can tell this `useEffect` is the only place where `setIconState` or `setCaretState` are actually used. If that's the case I don't understand why we need all the complication of passing down `iconState`, `caretState` and the 2 setters down as props just to achieve this - you can just remove all of them and set use `icon` everywhere you're using `iconState` in this child component, and `submenu` everywhere you're using `caretState`. Am I missing something which means we can't do this?

Actually there is one difference between the two - your version will `setIconState` to `true` if `icon` ever becomes `true`, but won't then set it back to `false` if `icon` goes back to being false. (Same with `submenu` and `setCaretState`.) I doubt it matters much in practice as these props aren't likely to change dynamically over the popover menu's lifecycle - but they could, in which case I think your version is bugged. Anyway, if I'm right about that it can be easily fixed, but it would be much better to do the simplification I've suggested.

Please let me know if I'm overlooking something which means we have to do it the complicated way!

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+    if (icon) {
+      setIconState?.(true);
+    }
+    if (submenu) {
+      setCaretState?.(true);
+    }
+  }, [icon, setCaretState, setIconState, submenu]);
+
+  useEffect(() => {
```
robinzigmond, line 224: **suggestion**: exactly the same comment here regarding `isLeftAligned`/`leftAlignmentState` as with `iconState` and `caretState` in the comment above - we can surely avoid any of this state and passing it down as props, when all we need to do is use the `isLeftAligned` state which is internal to `ActionPopoverItem`.

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+    if (!isLeftAligned) {
+      setLeftAlignmentState?.(true);
+    }
+
+    if (isLeftAligned) {
```
robinzigmond, line 229: **suggestion**: if we do need to keep this `useEffect`, this can just be an `if/else` rathre than 2 `if`s - or even just a single line: `setLeftAlignmentState(!isLeftAligned)`.

(Also that makes me reaise something is backwards between these 2 booleans - seems weird, not sure what's going on there...)

src/components/action-popover/action-popover-test.stories.tsx
```diff
@@ -269,11 +270,20 @@ export const ActionPopoverWithProps = ({ ...props }) => {
             <FlatTableCell>Doe</FlatTableCell>
             <FlatTableCell>
               <ActionPopover {...props}>
-                <ActionPopoverItem icon="email" disabled onClick={() => {}}>
+                <ActionPopoverItem
+                  data-element="menu-item1"
```
robinzigmond, line 274: **suggestion** (non-blocking) - it apears these `data-element` props have only been added to the story to allow them to be identified in your new Cypress tests. This shouldn't ideally be necessary as I assume we have cypress locators to find ActionPopoverItem, and if not we should be happy to add them.

Not a big deal when it's only a test story though!

src/components/action-popover/action-popover.spec.tsx
```diff
@@ -180,6 +181,421 @@ describe("ActionPopover", () => {
     );
   }
 
+  function renderDesignVariants(variant: number, props = {}, renderer = mount) {
```
robinzigmond, line 184: **comment**: by now I assume you know my feelings about this numeric `variant` prop in eamples 😅 Let's identify exactly which situations we're testing and create descriptive component/prop names so it's clear what exactly we're testing at each point.

src/components/action-popover/action-popover.spec.tsx
```diff
@@ -1519,36 +2002,353 @@ describe("ActionPopover", () => {
...
-        wrapper.find(StyledMenuItem)
-      );
-    });
+    it.each([
+      ["left", "left", "var(--spacing400)", "var(--spacing100)"],
+      ["left", "right", "var(--spacing100)", "var(--spacing100)"],
+      ["right", "left", "var(--spacing100)", "var(--spacing100)"],
+      ["right", "right", "var(--spacing100)", "var(--spacing400)"],
+    ])(
+      "when horizontalAlignment is %s, submenuPosition is %s and iconState is false, caretState is true, icon is false and submenu is false padding-left is: %s and padding-right is: %s",
```
robinzigmond, line 2035: **comment**: I think similar comments to those I already made on the new Cypress tests apply to these new unit tests too.

Further (and as I hadn't realised when first commenting on the Cypress tests), we can hopefully get rid of a lot of these if my comments about being able to remove `iconState` etc as props turn out to not be crazy 😅

src/components/action-popover/action-popover.style.ts
```diff
@@ -19,12 +20,134 @@ const Menu = styled.div`
     `${theme.zIndex?.popover}`}; // TODO (tokens): implement elevation tokens - FE-4437
 `;
 
+type variants = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;
+
+function getLeftPaddingValues(variant: variants) {
```
robinzigmond, line 25: **comment** - really just repeating what I already put in the parent component: I'm not a fan of using a single numeric `variant` prop to calculate left and right padding and so on. I'd rather calculate what we want in the parent component and pass appropriate padding (etc) props to the styled component.

cypress/components/action-popover/action-popover.cy.js
```diff
@@ -444,6 +447,15 @@ context("Test for ActionPopover component", () => {
         .should("be.visible");
     });
 
+    it("should render ActionPopover with icon's within a submenu", () => {
```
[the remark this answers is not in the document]
tomdavies73: Yeah, agree, I'll change.

src/components/action-popover/action-popover-test.stories.tsx
```diff
@@ -285,6 +295,433 @@ export const ActionPopoverWithProps = ({ ...props }) => {
   );
 };
 
+export const ActionPopoverWithDesignVariants = ({ ...props }) => {
```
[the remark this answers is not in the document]
tomdavies73: Hi Robin, I agree with your point, but there is definitely a need for separate components to test, as there are nine individual variants which all differ based on the truth/false state of `caretState`, `iconState` and the presence of an icon, or submenu. For example, `caretState` or `iconState` will be true even if there is one icon or submenu in the entire component, which means specific separate components need to be used in some test cases. However, we could condense some and make each child represent a specific variant where possible. I also think making these into separate test stories is a good shout too, as I was never really happy with this variant prop approach in the first place 👍. I think we might struggle on what to name them though aha

cypress/components/action-popover/action-popover.cy.js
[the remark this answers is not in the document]
tomdavies73: Technically no this test, doesn't open a submenu but to my knowledge, but it's finding the `MenuItemIcon` and ensuring it exists, and it is only present in a `submenu`. Happy to change this though 👍

cypress/components/action-popover/action-popover.cy.js
```diff
@@ -574,6 +604,605 @@ context("Test for ActionPopover component", () => {
...
   });
 
+  describe("padding checks on 'StyledMenuItemInnerText'", () => {
+    it.each([
+      ["left", "left"],
+      ["left", "right"],
+      ["right", "left"],
+      ["right", "right"],
+    ])(
+      "when horizontalAlignment is %s, submenuPosition is %s and iconState, caretState, icon and submenu are all false is left and right padding is: --spacing100",
```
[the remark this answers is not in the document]
tomdavies73: I agree, I'm happy to change the structure of the tests to maybe nest some `describe`s just to de-clutter each statement as they are fairly long. If we decide against that, though, I'll split up the descriptions the best I can, use commas where possible and remove the colons 👍.

cypress/components/action-popover/action-popover.cy.js
```diff
@@ -574,6 +604,605 @@ context("Test for ActionPopover component", () => {
...
+            submenuPosition={position}
+          />
+        );
+
+        actionPopoverButton()
+          .eq(0)
+          .click()
+          .then(() => {
+            actionPopoverMenuItemInnerText()
+              .should("have.css", "padding-left", "8px")
```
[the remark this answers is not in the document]
tomdavies73: Hi Robin, I agree with you! In theory, the entire rationale for including this in the first place was to stop us changing tests manually if they break. I just included the actual pixel values, as there were similar approaches for other CSS values for `background-color` and `min-height` in `Password` and `Textbox`. I agree though, plus any way to shorten both test files is an absolute win in my view.

cypress/components/action-popover/action-popover.cy.js
```diff
@@ -574,6 +604,605 @@ context("Test for ActionPopover component", () => {
...
+      }
+    );
+
+    it.each([
+      ["left", "left", "--spacing400", "32px"],
+      ["left", "right", "--spacing100", "8px"],
+      ["right", "left", "--spacing100", "8px"],
+      ["right", "right", "--spacing100", "8px"],
+    ])(
+      "when horizontalAlignment is %s, submenuPosition is %s and iconState is false, caretState is true, icon is false and submenu is false padding-left is: %s",
```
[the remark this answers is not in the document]
tomdavies73: Agree completely, as i've replied above i'm going to look at some other solutions to make each test case easier to read, failing that i'll utilise your approach in the descriptions as this is much better and more readable for sure 👍

cypress/components/action-popover/action-popover.cy.js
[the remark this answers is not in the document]
tomdavies73: This is an excellent point, I just focused on testing every variant separately, but in theory we could do one test which covers all `--spacing100` and then separate tests for the cases where it differs from this baseline. Which would massively cut down on code 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+      setLeftAlignmentState?.(false);
+    }
+  }, [setLeftAlignmentState, isLeftAligned]);
+
   const alignSubmenu = useCallback(() => {
+    const align = leftAlignSubmenu(ref, submenuRef);
+    setIsLeftAligned(align);
+
+    return checkRef(ref) && checkRef(submenuRef) && submenu;
+  }, [submenu, placement]);
```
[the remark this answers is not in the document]
tomdavies73: Agreed, think `placement` was there originally, but I can remove 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+
+  useEffect(() => {
     if (checkRef(ref) && checkRef(submenuRef) && submenu) {
-      const align = leftAlignSubmenu(ref, submenuRef);
-      setIsLeftAligned(align);
       setContainerPosition(getContainerPosition(ref, submenuRef, placement));
     }
-  }, [submenu, placement]);
+  }, [submenu, placement, currentSubmenuAlignment()]);
```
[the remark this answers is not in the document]
tomdavies73: Cheers Robin, great point. I'll review all console warnings and see where we can put `currentSubmenuAlignment()` and work out why it's there. I'll try and work out a way to fix, but if I hit a snag i'll drop you a message 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -242,7 +333,15 @@ export const ActionPopoverItem = ({
...
     },
-    [disabled, download, isHref, isLeftAligned, onClick, submenu]
+    [
+      disabled,
+      download,
+      isHref,
+      isLeftAligned,
+      onClick,
+      submenu,
+      currentSubmenuAlignment(),
```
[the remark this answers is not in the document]
tomdavies73: Agreed, i'll remove `isLeftAligned` and try fix the rest 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
   const mouseEnterTimer = useRef<NodeJS.Timeout | null>(null);
   const mouseLeaveTimer = useRef<NodeJS.Timeout | null>(null);
 
+  const currentSubmenuAlignment = () => {
```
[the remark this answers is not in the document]
tomdavies73: Agree, i'll condense 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+    itemRef: React.RefObject<HTMLElement>,
+    containerSubmenuRef: React.RefObject<HTMLElement>,
+    containerPlacement: "bottom" | "top"
+  ): ContainerPosition | undefined {
+    /* istanbul ignore if */
+    if (!itemRef.current || !containerSubmenuRef.current) return undefined;
+
+    const { offsetWidth: submenuWidth } = containerSubmenuRef.current;
+
+    const horizontalAlignmentValue = currentSubmenuAlignment() === "left";
```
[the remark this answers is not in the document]
tomdavies73: Agree, i'll change 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -139,19 +143,106 @@ export const ActionPopoverItem = ({
...
+    const yPositionName = containerPlacement === "top" ? "bottom" : "top";
+
+    return {
+      left: leftValue,
+      [yPositionName]: "calc(-1 * var(--spacing100))",
+      right: rightValue,
+    };
+  }
+
+  const retrieveVariantType = () => {
```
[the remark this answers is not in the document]
tomdavies73: Agreed, I was never a fan of a the `variant` approach, I guess we could just pass `submenu`, `icon`, `caretState` and `iconState` to the `MenuItemInnerText`. That's actually, definitely a much better solution which would reduce the need for both switch cases in the style file too 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
tomdavies73: Hi Robin, so `setIconState` and `setCaretState` which set `iconState` and `caretState` to true respectively is how we've made design changes based on the value of siblings. So let's say we have x8 children and the first has an `icon`, `setIconState` will be triggered and will be changed to `true`, and will not go back to `false`. This is the desired original behaviour and not a bug imo, as now we can make design changes knowing a sibling has an `Icon`. 

Screenshot 2023-06-21 at 13 46 13

The above screenshot is a good example of this, the first child has an `icon` so `iconState` is true, and remains true even though the next three children do not have an `icon`. This means those three children know a sibling has an icon, so padding is applied to each text to ensure padding is the same, which matches designs. This was my solution to make changes based on the value of siblings 👍. Happy to explore alternatives if you're not happy with this, though, or if you still feel it's incorrect.

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
tomdavies73: A similar approach was used with `isLeftAligned` and `leftAlignmentState` as `isLeftAligned` is only false when a childs `submenu` width is smaller than space available on the left-hand side of the container. Technically, `isLeftAligned` will never be false for children without a `submenu`. Therefore, without this approach of changing state once to true for all siblings, only some children would switch to the right. With this approach, as soon as the criteria is met for one child, every single child switches to the right. Which is a lot cleaner imo 👍 , happy to discuss this approach, and the other sibling state approaches too though.

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
tomdavies73: I'll have a look to see if this can be changed to an `else` 👍

src/components/action-popover/action-popover-test.stories.tsx
tomdavies73: I can include a locator for `MenuItem`, I think one or two tests use their own `data-element` like I've done here, but if I add a locator I can update these to use it too 👍

src/components/action-popover/action-popover.spec.tsx
tomdavies73: Yep, agreed 👍

src/components/action-popover/action-popover.style.ts
tomdavies73: Agreed, if we pass props to these styled components, we can do css applications for each one, the styled component might get a bit long, but I agree its a much better approach 👍

src/components/action-popover/action-popover.spec.tsx
tomdavies73: I agree about the descriptions being unreadable, but see my comment explaining my view on the state props. Happy to work out what to do going forward 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
robinzigmond: Thanks, I'll take another look at your code but I think this is raising something important I'd missed - that these are being passed down to all children so any one of them can affect all the others. That definitely makes it more complex than I was initially thinking. I'll have to have a think about if there's any other way which removes the need to have all these `useEffect`s (there may not be though!)

src/components/action-popover/action-popover.style.ts
edleeks87: I agree with Robin, I also think as some of the variants produce the same value for right/ left padding they don't need to be separate in whatever you go for in the end. For example variants 3 and 4 produce the same padding results so you probably don't need both etc.

src/components/action-popover/action-popover.style.ts
```diff
@@ -19,12 +20,134 @@ const Menu = styled.div`
...
+  if (variant === 1) {
+    return "var(--spacing100)";
+  }
+  if (variant === 2) {
+    return "var(--spacing400)";
+  }
+  if (variant === 3 || variant === 4) {
```
edleeks87, line 32: This is what I mean whereby you don't really need separate variants for `3` and `4` or `5`,`6` and `7` as each one produces the same padding left or right token value

src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx
```diff
@@ -162,6 +168,10 @@ const ActionPopoverMenu = React.forwardRef<
       [focusButton, setOpen, focusIndex, items, setFocusIndex]
     );
 
+    const [caretState, setCaretState] = useState(undefined);
```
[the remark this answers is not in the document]
edleeks87: suggestion: it might be a little more intuitive if we named these a little differently but it's not essential if you're happy with them. I was thinking something like:
```suggestion
    const [childHasSubmenu, setChildHasSubmenu] = useState(false);
     const [childHasIcon, setChildHasIcon] = useState(false);
```

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
edleeks87: question: should we handle for the potential that icons could be added/ removed dynamically as well, this current approach doesn't cater for that? Happy to leave it as is for now but figured I'd ask and see what you and Robin thought

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -124,42 +133,96 @@ export const ActionPopoverItem = ({
...
+    setContainerPosition(getContainerPosition(ref, submenuRef, placement));
+  }, [submenu, placement, currentSubmenuPosition]);
 
   useEffect(() => {
-    alignSubmenu();
+    if (submenu && !hasSubmenuAligned.current) {
+      alignSubmenu();
+      hasSubmenuAligned.current = true;
+    }
+  });
```
robinzigmond, line 219:
```suggestion
  }, [alignSubmenu, submenu]);
```

**nitpick/opinion**: I don't think it's all that important, since by design the `alignSubmenu()` can only be run once by this, but I'd prefer not to run the effect itself on every render. (Performance-wise I don't actually know which is better, not running the `if` check after every render should save time but adding the dependencies makes React itself check if they've changed. I doubt it matters noticeable either way - I just prefer to include effect dependencies. Happy to leave this up to you though.)

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -124,42 +133,96 @@ export const ActionPopoverItem = ({
...
+        right: rightValue,
+      };
     }
-  }, [submenu, placement]);
+    setContainerPosition(getContainerPosition(ref, submenuRef, placement));
+  }, [submenu, placement, currentSubmenuPosition]);
 
   useEffect(() => {
-    alignSubmenu();
+    if (submenu && !hasSubmenuAligned.current) {
```
robinzigmond, line 215: **question** (non-blocking): I don't think it's at all likely to happen in practice, but just in case - I'm wondering what happens if the `submenu` prop gets dupdated/toggled during the component's lifecycle. The way you have it now, it looks like it will be fine the first time it comes on, even if that's well ater the first render. But if it gets toggled from off to on to off and then on again(!), then `alignSubmenu` will only be run the first time it turns on, which will set the ref flag to `true` and stop it being run the second time. Might this matter? (I honestly don't know - but it's such an edge case I doubt it's worth worrying much over...)

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -124,42 +133,96 @@ export const ActionPopoverItem = ({
...
+    return checkRef(ref) && checkRef(submenuRef) && submenu;
+  }, [
+    submenu,
+    setCurrentSubmenuPosition,
+    submenuPosition,
+    currentSubmenuPosition,
+  ]);
+
+  useEffect(() => {
+    function getContainerPosition(
```
robinzigmond, line 190: **comment** (non-blocking): I see that this function, which previously was defined outside the scope of the component, is now inside (and correctly inside the `useEffect` since it's not called from anywhere else). It doesn't matter much but this could be moved back outside - while it does use values from the component scope, this could still work with the function defined outside by passing those values in as additional arguments.

I'm not saying this would be better - 6 arguments would be rather a lot! - but it seems inconsistent to pass in 3 values as arguments but not the other 3. I'd prefer to either have the function defined outside the component and pass all 6 in as arguments, or have it here and read everything from the component scope (and add them as effect dependencies).

Not important though so don't feel under any obligation.

src/components/action-popover/action-popover.spec.tsx
```diff
@@ -180,6 +181,249 @@ describe("ActionPopover", () => {
...
+        <>
+          <input id="before" />
+          <ActionPopover {...defaultProps} {...props} />
+          <input id="after" />
+        </>
+      </ThemeProvider>
+    );
+  }
+
+  function renderActionPopoverWithSubmenusAndIconsAndNoSubmenusAndIcons(
```
robinzigmond, line 356:
```suggestion
  function renderActionPopoverWithVariableChildren(
```

**suggestion** (non-blocking): I don't know if I even like this name, there's probably something better. I don't mind the length of the original name at all, if it's descriptive, the problem I have is that it sounds contradictory - and I only discovered by asking you in person that it means that some children have submenu/icon and some don't. (Which is absolutely a case we should test, my comment is purely about the naming.)

src/components/action-popover/action-popover.spec.tsx
```diff
@@ -180,6 +181,249 @@ describe("ActionPopover", () => {
...
+          Return Home
+        </ActionPopoverItem>,
+        null,
+        undefined,
+      ],
+      ...props,
+    };
+
+    renderer(
+      <ThemeProvider theme={mintTheme}>
```
robinzigmond, line 219: **question** - do we need the `mintTheme` (or a `ThemeProvider` at all) for any of these tests? Pretty sure that's a legacy thing. I do see it was in use before for some tests in this file but it looks like you're using it in a lot more now. Now would be a good time to remove all mentions of it if it's not actually needed.

I'm only making this comment once but of course it's used in several of these functions, the same comment applies to each.

src/components/action-popover/action-popover.spec.tsx
edleeks87: Yeah agree this name could be a bit shorter, you could even just go with `renderWithVariableChildren`

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
edleeks87: Agree with Robin, worth seeing if this works with a deps array

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
edleeks87: Yeah I agree with Robin, I don't think we need to pass any arguments here as it will have access to the refs/props

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
tomdavies73: So I've tested this using a `setTimeout` to set the `submenu` prop to either a submenu or null, and it appears to work fine. Even when the default is `null`, and a submenu is passed lets say 5 seconds after render, the submenu is adjusted correctly.  Even when there is not enough space on the right for example, and then submenus are passed 5 seconds later, it inverts as expected. I'm happy to add this into a codeSandbox after the next batch of changes to show this, so you can both play around with it, but to me, it looks ok 👍

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -187,36 +187,32 @@ export const ActionPopoverItem = ({
...
         left: leftValue,
         [yPositionName]: "calc(-1 * var(--spacing100))",
         right: rightValue,
       };
-    }
-    setContainerPosition(getContainerPosition(ref, submenuRef, placement));
-  }, [submenu, placement, currentSubmenuPosition]);
+    };
+    setContainerPosition(getContainerPosition);
+  }, [submenu, currentSubmenuPosition, ref, submenuRef, placement]);
```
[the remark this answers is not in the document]
robinzigmond: agreed - React knows refs can't change so the lint rule for exhaustive dependencies will ignore it if you leave them out. (It's different if it's passed in from another component, but here when defined in the same component via `useRef` you should be fine leaving them out.)

src/components/action-popover/action-popover.spec.tsx
```diff
@@ -462,6 +447,7 @@ describe("ActionPopover", () => {
   });
 
   testStyledSystemMargin((props) => (
+    // ThemeProvider needs to remain here due to this test
```
robinzigmond, line 450: I understand why you added this comment, in the context of my previous comment about the use of `theme`, but in the context of the code itself (rather than the PR and its review comments) I don't think it adds anything and may be confusing. I think it's sufficiently obvious (at least to those who've worked on Carbon for a while) why a theme is needed for the styled system stuff to work correctly.

(Although maybe it's no so obvious after all - I tried the test without the ThemeProvider and you're right, they fail, but ideally we'd update the test utils so they work with the SageTheme... That's not relevant for this PR though! 😅 )

robinzigmond (approved):

divyajindel (approved):

PR #6011, opened 2023-05-09
fix(submenu): ensure that the search cross icon is centred when focused - FE-5694

DipperTheDan:
Currently the `X` icon in `Search` is not centered when used in Menu/Submenu. This is because margin-right is applied to icons in submenu. The fix for this is target the icon inside of `Search` and make sure that margin-right is 0px.

fixes #FE-5694

### Checklist

- [] Screenshots are included in the PR if useful

#### QA

### Testing instructions


edleeks87 (commented):
style changes look good to me, I think it's worth adding some sort of regression test here as I can imagine it'll be rather easy for it to happen

edleeks87 (commented):
duplicate of other comment

edleeks87 (approved):

robinzigmond (commented):
not sure if you plan to do it anyway, but in case not - your 2 commits should really be squashed, as they're not independent of each other (the test is testing the changes in the other commit, and if for some reason we ever reverted the first commit while keeping the one with the new test, that test would then fail - or at least I hope so!)

robinzigmond (approved):

DlgSHi (approved):

PR #6033, opened 2023-05-16
feat(message): enable component to be programmatically focused - FE-5727

### LINKED ISSUE #5924: Add ability to focus on Message component
### Desired behaviour

We need the ability to be able to set a focus state on a Message component.

Context: To follow a11y guidelines in situations where an action is performed by the user and the action has finished, if a message component is then rendered to inform the user of the completed action, the message should be focused on. Anyone using a screen reader in this case would get confirmation that the action has finished.

### Current behaviour

Message component is not focusable.

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

DipperTheDan:
fixes #5924

### Proposed behaviour

Add a ref so that the `Message` component can be focused.

### Current behaviour

Component is not currently focusedable.

### Checklist

#### QA

### Additional context

Inconsistent screenreader behaviour will not be addressed in this ticket. This work is to provide a mechanism for consumers to focus the `Message` component.

### Testing instructions

- `with focus` example: Message component should be focused when first opened and **not** display an outline. 
- `with focus` example: Screenreader should read out the message in the component.
- Should not have effected the functionality of the component.
- Should be no new accessibility warnings.

### DIFF
--- a/cypress/locators/message/index.js
+++ b/cypress/locators/message/index.js
@@ -3,6 +3,7 @@ import {
   MESSAGE_CHILDREN,
   MESSAGE_TITLE,
   MESSAGE_DISMISS_ICON,
+  MESSAGE_DISMISS_ICON_BUTTON,
   MESSAGE_CONTENT,
   VARIANT_PREVIEW,
 } from "./locators";
@@ -12,5 +13,7 @@ export const messagePreview = () => cy.get(MESSAGE_PREVIEW);
 export const messageChildren = () => cy.get(MESSAGE_CHILDREN);
 export const messageTitle = () => cy.get(MESSAGE_TITLE);
 export const messageDismissIcon = () => cy.get(MESSAGE_DISMISS_ICON);
+export const messageDismissIconButton = () =>
+  cy.get(MESSAGE_DISMISS_ICON_BUTTON);
 export const messageContent = () => cy.get(MESSAGE_CONTENT);
 export const variantPreview = () => cy.get(VARIANT_PREVIEW);
--- a/src/components/message/message-test.stories.tsx
+++ b/src/components/message/message-test.stories.tsx
@@ -1,10 +1,11 @@
-import React, { useState } from "react";
+import React, { useState, useRef, useEffect } from "react";
 import { action } from "@storybook/addon-actions";
 import Button from "../button";
 import Message, { MessageProps } from "./message.component";
 
 export default {
   title: "Message/Test",
+  includeStories: ["Default"],
   parameters: {
     info: { disable: true },
     chromatic: {
@@ -61,6 +62,29 @@ export const MessageComponent = (props: MessageProps) => {
     <div>
       {!isOpen && <Button onClick={() => setIsOpen(true)}>Open Message</Button>}
       <Message open={isOpen} onDismiss={() => setIsOpen(false)} {...props}>
+        Some custom message
+      </Message>
+    </div>
+  );
+};
+
+export const MessageComponentWithRef = (props: MessageProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const messageRef: React.Ref<HTMLDivElement> = useRef(null);
+
+  useEffect(() => {
+    if (isOpen) messageRef.current?.focus();
+  });
+
+  return (
+    <div>
+      {!isOpen && <Button onClick={() => setIsOpen(true)}>Open Message</Button>}
+      <Message
+        open={isOpen}
+        onDismiss={() => setIsOpen(false)}
+        ref={messageRef}
+        {...props}
+      >
         Some custom message
       </Message>
     </div>
--- a/src/components/message/message.component.tsx
+++ b/src/components/message/message.component.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useRef } from "react";
 import { MarginProps } from "styled-system";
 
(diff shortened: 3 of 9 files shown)


src/components/message/message.component.tsx
```diff
@@ -39,52 +39,82 @@ export interface MessageProps extends MarginProps {
...
+      variant = "info",
+      children,
+      onDismiss,
+      id,
+      className,
+      closeButtonAriaLabel,
+      showCloseIcon = true,
+      ...props
+    }: MessageProps,
+    ref
```
Parsium, line 57: It might be worth documenting `ref` in storybook so consumers are aware it is available for use for them.

Parsium (commented):
May I ask if we need to consider the case when multiple `Message`s are opened?

cypress/locators/message/index.js
```diff
@@ -12,5 +13,7 @@ export const messagePreview = () => cy.get(MESSAGE_PREVIEW);
 export const messageChildren = () => cy.get(MESSAGE_CHILDREN);
 export const messageTitle = () => cy.get(MESSAGE_TITLE);
 export const messageDismissIcon = () => cy.get(MESSAGE_DISMISS_ICON);
+export const messageDismissIcoButton = () =>
```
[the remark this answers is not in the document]
DipperTheDan: Good spot 👍🏻

src/components/message/message.component.tsx
```diff
@@ -39,52 +39,82 @@ export interface MessageProps extends MarginProps {
...
+  ) => {
+    const messageRef = useRef<HTMLDivElement | null>(null);
+    const focusedElementBeforeOpening = useRef<HTMLElement | null>(null);
+    const [tabIndex, setTabIndex] = useState<number | undefined>(0);
```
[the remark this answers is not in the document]
DipperTheDan: I had thought about that but we do it this way in `Toast`.
https://github.com/Sage/carbon/blob/master/src/components/toast/toast.component.tsx#L80

I figured to keep things consistent, I'd do it this way. Happy to change this though as I'm good with either way.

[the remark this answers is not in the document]
DipperTheDan: I've left this as is for now until a second reviewer appears on the PR.

src/components/message/message.spec.tsx
```diff
@@ -201,4 +202,80 @@ describe("Message", () => {
...
     testStyledSystemMargin((props) => <Message {...props} />);
   });
+
+  describe("focus behaviour", () => {
+    let wrapper: ReactWrapper | ShallowWrapper;
+    let onDismissCallback: jest.Mock;
+
+    beforeEach(() => {
+      const element = document.createElement("div");
+      const htmlElement = document.body.appendChild(element);
```
robinzigmond, line 212: do we need the `appendChild` stuff here? Fair enough if we do, I know we use it elsewhere but I've never fully understood why it's sometimes needed, and I'd rather do without if we can. If it doesn't work without, feel free to leave it in.

src/components/message/message.stories.mdx
Parsium, line 1: **issue:** I noticed that the `ref` prop is not showing in the prop table.

It seems this may be related to this issue on how [storybook is generating the docs internally for React forward refs](https://github.com/storybookjs/storybook/issues/21801). If this is the case this [seems to be fixed in storybook 7.x](https://github.com/storybookjs/storybook/issues/21801#issuecomment-1504325012). In this case what do you think about manually adding this in for now?

Parsium: Agreed documentation is sufficient with the added `withFocus` story

robinzigmond (commented):
Happy enough with this other than the minor things I've replied to.

Just wondering though if it's the right approach to just give consumers the ref and leave it up to them to discover and use this functionality - I feel like much of the time it will be appropriate to have these automatically focus when opening, and we should offer a boolean prop to automatically enable that behaviour (or make it the default and offer a prop to turn it off)?

I realise this would mean a fair rewrite at this stage, maybe this is best addressed in the upcoming work on this component regarding it being a live region - but I wanted to mention it anyway.

src/components/message/message.stories.tsx
```diff
@@ -107,33 +107,112 @@ export const WithRichContent: ComponentStory<typeof Message> = () => {
 };
 
 export const WithMargin: ComponentStory<typeof Message> = () => {
-  const [isOpen, setIsOpen] = useState(true);
+  const [isOpen, setIsOpen] = useState({
```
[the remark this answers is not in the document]
DipperTheDan: Awh yeah. Good point. Been so focused on the other stuff, forgot to put this in it's own commit. I'll do it in the next round of addressing comments.

Parsium (dismissed):

robinzigmond (dismissed):

src/components/message/message-test.stories.tsx
```diff
@@ -66,3 +67,26 @@ export const MessageComponent = (props: MessageProps) => {
     </div>
   );
 };
+
+export const MessageComponentWithRef = ({ ...props }) => {
```
DlgSHi, line 71:
```suggestion
export const MessageComponentWithRef = (props: MessageProps) => {
```

DlgSHi (approved):

Parsium (approved):

robinzigmond (approved):

PR #6034, opened 2023-05-16
fix(menu): ensure onSubmenuOpen is only called once when the submenu opens

robinzigmond:
### Proposed behaviour

The onSubmenuOpen prop, if provided, should only be called once each time the submenu opens - even if subsequent events are triggered while it is open, that would open it if it weren't already open.

### Current behaviour

onSubmenuOpen is called multiple times in some situations:

- when pressing the Enter or Space key to open a submenu with the keyboard, the function is called twice
- when moving the mouse over the item, the function is called multiple times (typically 3 to 4), even though the submenu only opens once

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/trusting-water-yiyzf3?file=/src/App.tsx

### DIFF
--- a/cypress/components/menu/menu.cy.js
+++ b/cypress/components/menu/menu.cy.js
@@ -1554,7 +1554,7 @@ context("Testing Menu component", () => {
     let callback;
 
     beforeEach(() => {
-      callback = cy.stub();
+      callback = cy.stub().as("callback");
     });
 
     it("should call onClick callback when a click event is triggered", () => {
@@ -1579,13 +1579,9 @@ context("Testing Menu component", () => {
         </Box>
       );
 
-      submenu()
-        .eq(positionOfElement("first"), div)
-        .trigger("mouseover")
-        .then(() => {
-          // eslint-disable-next-line no-unused-expressions
-          expect(callback).to.have.been.calledOnce;
-        });
+      submenu().eq(positionOfElement("first"), div).trigger("mouseover");
+
+      cy.get("@callback").should("have.been.calledOnce");
     });
 
     it("should call onSubmenuOpen callback when a click event is triggered", () => {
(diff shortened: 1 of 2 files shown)


cypress/components/menu/menu.cy.js
```diff
@@ -1569,8 +1569,10 @@ context("Testing Menu component", () => {
         .eq(positionOfElement("first"), div)
         .trigger("mouseover")
         .then(() => {
-          // eslint-disable-next-line no-unused-expressions
-          expect(callback).to.have.been.calledOnce;
+          cy.wait(0).then(() => {
```
DlgSHi, line 1572: why do we need to wait for (0) seconds? :) for all tests below?

robinzigmond: hah you're quick Igor! I expected to have this conversation later when it reached QA 😅 

Basically it's because the fix involved adding a `useEffect` which means the submenu only opens on the next render (which happens straight away), rather than straight away in the event handler which it was before. For that reason, without the wait these tests all fail - the callback is called but a tiny fraction of a second after the assertion. Thankfully a wait of 0 does work, it just needs to be told to run asynchronously.

I don't particularly like the way I was forced to write this test though, particularly the nested `then` (just putting `cy.wait(...)` above didn't work) - so I'd be happy to hear better ways of writing it!

DlgSHi (dismissed):

cypress/components/menu/menu.cy.js
Parsium: I've taken a look at the cypress docs and I noticed they generally [don't recommend chaining off of actions](https://docs.cypress.io/guides/core-concepts/retry-ability#Incorrectly-checking-if-the-stub-was-called) like `trigger()`.

Using [an alias for the callback](https://docs.cypress.io/guides/core-concepts/retry-ability#Correctly-waiting-for-the-stub-to-be-called) should work instead to avoid having to use `cy.wait`. This should take into account React re-rendering the DOM because of the effect:

```tsx
// in describe block...
const callback = cy.stub().as("callback");

// in test...
submenu().eq(positionOfElement("first"), div).trigger("mouseover");
cy.get("@callback").should("have.been.calledOnce");
```

edleeks87 (commented):
Nothing else from me, happy to approve once James' comment is addressed

Parsium (dismissed):

edleeks87 (dismissed):

ZhuoyuJin (commented):
Happy to approve it once the comments from Igor have been addressed

DlgSHi (approved):

ZhuoyuJin (approved):

edleeks87 (approved):

Parsium (approved):

PR #6038, opened 2023-05-18
chore(cypress): convert remaining support utils to typescript (FE-5895)

Parsium:
### Proposed behaviour

- Convert all remaining cypress support utils to TypeScript
- Correct any util typing that is causing type errors
- Rename `useJQueryCssValueAndAssert` util to `assertCssValueIsApproximately` to reflect its behaviour
- Remove unused `dragAndDrop` helper

### Current behaviour

- Still have some cypress utils written in JavaScript with no TypeScript type declarations - which will be needed when migrating the cypress `*.cy.*` tests to TypeScript
- Some utils need their types modifying - to address some type errors encountered while migrating the test files to TypeScript

### Checklist

#### QA

### Testing instructions

- Ensure `npm run type-check` passes without errors
- Ensure all cypress tests still pass


cypress/support/helper.ts
```diff
@@ -53,72 +21,78 @@ export function continuePressingTABKey(count: number, shift = false) {
...
+  third: 2,
+  fourth: 3,
+  fifth: 4,
+  sixth: 5,
+  seventh: 6,
+  eighth: 7,
+  ninth: 8,
+  tenth: 9,
+  eleventh: 10,
+  thirteenth: 12,
```
DlgSHi, line 36: why this is changed from value 13 to 12? I think we a safe to delete those values (because we don't use them)

cypress/support/helper.ts
```diff
@@ -53,72 +21,78 @@ export function continuePressingTABKey(count: number, shift = false) {
...
+  nineteenth: 18,
```
DlgSHi, line 37: same here?

cypress/support/helper.ts
```diff
@@ -53,72 +21,78 @@ export function continuePressingTABKey(count: number, shift = false) {
...
+};
 
-export function positionOfPaginationButton(type: string) {
```
DlgSHi, line 74: is this function not used anywhere?

Parsium: Searching with VSCode I didn't find any files with references to this function, which is why I decided to remove it. Looking at the git history, it seems this function was created for use in the old BDD cypress tests, but has since become unused.

cypress/support/helper.ts
Parsium: Ah I can remove these two if you would like, as I also noticed they weren't used anywhere in Carbon. I changed these values as, given this function is used to index an array, this wouldn't return the thirteenth and nineteenth elements as arrays are zero-indexed.

Parsium: Now deleted thirteenth and nineteenth 👍🏼

DlgSHi (approved):

nineteen88 (approved):

tomdavies73 (approved):

PR #6056, opened 2023-05-23
test(help): typescript refactor

divyajindel:
### Proposed behaviour
- Rename the `help.cy.js` to `help.cy.tsx` file in `./cypress/components/help/` folder.

### Current behaviour

Currently Message component is in JS not in TS.

### Checklist

#### QA

### Testing instructions
- [x] Run `npx cypress open --component` to check if the`help.cy.tsx` file passed
- [x] Run `npx cypress run --component` to check none of the other *.cy.tsx files have regressed
- [x] Run `npm run type-check` to make sure there is no type issue


cypress/components/help/help.cy.tsx
```diff
@@ -78,13 +63,16 @@ context("Tests for Help component", () => {
...
+      let errors: any;
       beforeEach(() => {
         cy.window().then((win) => {
           errors = cy.spy(win.console, "error").as("spyWinConsoleError");
         });
       });
+
       it("should throw a console error when invalid html elements are used", () => {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore:next-line
```
Parsium, line 75: For context for other reviewers, I suggested @divyajindel to instruct the TypeScript compiler to ignore this line.

TypeScript consumers will see a compile-time error if they attempt to pass an illegal value as normal, but for non-TypeScript consumers who don't have this ability, we need to make sure a runtime error is raised for this case as well, which is want the test is verifying.

cypress/components/help/help.cy.tsx
```diff
@@ -78,13 +63,16 @@ context("Tests for Help component", () => {
         "Warning: The tag <%s> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.%s",
         asElem
       );
-      let errors;
+      let errors: any;
```
Parsium, line 66: To avoid using `any`, there is an alternative way we could write the test. What do you think about updating the test to be the following?:

```tsx
describe("check as prop for Help component", () => {
  it("should throw a console error when invalid html elements are used", () => {
    cy.spy(window.console, "error");
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore:next-line
    CypressMountWithProviders(<HelpComponent as="foo" />);
    cy.window().its("console.error").should("have.been.called");
  });
});
``` 

This should eliminate the need for you to type that `errors` object :)

cypress/components/help/help.cy.tsx
```diff
@@ -54,12 +39,12 @@ context("Tests for Help component", () => {
...
     });
 
-    it.each(["-1", "0", "1", "5"])(
+    it.each([-1, 0, 1, 5])(
       "should check %s as tabIndex for Help component",
       (tabIndex) => {
         CypressMountWithProviders(<HelpComponent tabIndex={tabIndex} />);
 
-        if (tabIndex === "-1") {
+        if (String(tabIndex) === "-1") {
```
DlgSHi, line 47: I think we can go with
```suggestion
        if (tabIndex == "-1") {
```

cypress/components/help/help.cy.tsx
DlgSHi: still not resolved

cypress/components/help/help.cy.tsx
DlgSHi: same here

DlgSHi (changes requested):

cypress/components/help/help.cy.tsx
divyajindel: Hi @DlgSHi , I've tried your suggested solution but didn't work me. Please let me know if we can fix this using some different method.

DlgSHi: changed as suggested locally and it works as expected

DlgSHi (changes requested):

DlgSHi (approved):

Parsium (approved):

PR #6067, opened 2023-05-26
test(confirm): typescript refactor

ZhuoyuJin:
### Proposed behaviour
- Rename the `confirm.cy.js` to `confirm.cy.tsx` file in `./cypress/components/confirm/` folder.

### Current behaviour

Currently Confirm component is in JS not in TS.

### Checklist

### Testing instructions
- [ ] Run `npx cypress open --component` to check if the`confirm.cy.tsx` file passed
- [ ] Run `npx cypress run --component` to check none of the other `*.cy.tsx` files have regressed

### DIFF
--- a/cypress/components/confirm/confirm.cy.tsx
+++ b/cypress/components/confirm/confirm.cy.tsx
@@ -1,5 +1,7 @@
+/* eslint-disable jest/valid-expect */
+/* eslint-disable no-unused-expressions */
 import React from "react";
-import Confirm from "../../../src/components/confirm";
+import Confirm, { ConfirmProps } from "../../../src/components/confirm";
 import CypressMountWithProviders from "../../support/component-helper/cypress-mount";
 import { ConfirmComponent } from "../../../src/components/confirm/confirm-test.stories";
 import {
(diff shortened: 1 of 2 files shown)


cypress/components/confirm/confirm.cy.tsx
```diff
@@ -196,12 +200,12 @@ context("Testing Confirm component", () => {
     });
 
     it("should check Esc key is disabled", () => {
+      const callback = cy.stub();
```
DlgSHi, line 203: was this callback assigned anywhere?

src/components/confirm/confirm-test.stories.tsx
```diff
@@ -95,7 +95,7 @@ Default.args = {
   cancelButtonType: "secondary",
 };
 
-export const ConfirmComponent = ({ ...props }) => {
+export const ConfirmComponent = ({ ...props }: Partial<ConfirmProps>) => {
```
DlgSHi, line 98:
```suggestion
export const ConfirmComponent = (props: Partial<ConfirmProps>) => {
```

DlgSHi (changes requested):

cypress/components/confirm/confirm.cy.tsx
ZhuoyuJin: Callback is also appear for the event tests, but for this case, this callback is only assigned for this particular test

cypress/components/confirm/confirm.cy.tsx
```diff
@@ -200,7 +200,7 @@ context("Testing Confirm component", () => {
     });
 
     it("should check Esc key is disabled", () => {
-      const callback = cy.stub();
+      const callback: ConfirmProps["onCancel"] = cy.stub();
       CypressMountWithProviders(<ConfirmComponent disableEscKey />);
```
DlgSHi, line 204: but it is still not used here like:
```suggestion
CypressMountWithProviders(<ConfirmComponent disableEscKey onCancel={callback} />);
```

cypress/components/confirm/confirm.cy.tsx
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable jest/valid-expect */
+/* eslint-disable no-unused-expressions */
```
stephenogorman, line 2: Could be one line:
/* eslint-disable jest/valid-expect , no-unused-expressions */

cypress/components/confirm/confirm.cy.tsx
```diff
@@ -323,16 +341,10 @@ context("Testing Confirm component", () => {
       );
     });
 
-    it("should focus the dialog container", () => {
```
stephenogorman, line 326: Has this test been removed?

ZhuoyuJin: Yes, there is no such blockFocusElement prop in confirm component now

cypress/components/confirm/confirm.cy.tsx
ZhuoyuJin: Don't think this will effect the tests, but noted and will make changes in next PR

DlgSHi (approved):

stephenogorman (approved):

tomdavies73 (approved):

PR #6081, opened 2023-06-02
Convert Date and DateRange to TypeScript

edleeks87:
### Proposed behaviour

Converts `Date` and `DateRange` to TypeScript

### Current behaviour

`Date` and `DateRange` are written in JavaScript

### Checklist

#### QA

### Testing instructions


src/components/date-range/date-range.spec.tsx
```diff
@@ -25,36 +25,62 @@ const datePickerWidths = {
...
+}
+
 describe("DateRange", () => {
-  let wrapper;
-  let startInput;
-  let endInput;
-  let customOnChange;
-  let customOnBlur;
+  let wrapper: ReactWrapper;
+  // let startInput: ShallowWrapper | ReactWrapper;
```
nineteen88, line 59: suggestion: remove commented out code

src/components/date/date.spec.tsx
```diff
@@ -503,25 +631,6 @@ describe("Date", () => {
         expect(wrapper.update().find(DayPicker).exists()).toBe(false);
         wrapper.unmount();
       });
-
-      it("the 'DatePicker' should close if it is open", () => {
```
DipperTheDan, line 507: question: has this test been removed intentionally?

src/components/date/date.stories.tsx
```diff
@@ -0,0 +1,478 @@
...
+  const [state, setState] = useState("2022-04-05");
+  const handleChange = (ev: DateChangeEvent) => {
+    setState(ev.target.value.formattedValue);
+  };
+  const [state2, setState2] = useState("2022-04-05");
+  const handleChange2 = (ev: DateChangeEvent) => {
+    setState2(ev.target.value.formattedValue);
+  };
+  return (
+    <div style={{ display: "flex", justifyContent: "space-around" }}>
```
DipperTheDan, line 447: suggestion: Is it worth changing this to `Box`? I believe @Parsium has a PR pending to swap out all instances of HTML elements in out stories.

src/components/date/__internal__/date-formats/date-formats.spec.ts
```diff
@@ -511,7 +511,22 @@ describe.each([...euLocales, ...naLocales, ...cnLocales])(
     });
 
     it("the expected format", () => {
-      expect(format).toEqual(formatMap[locale]);
+      const { format } = getFormatData({ code: locale });
+
+      expect(format).toEqual(formatMap[locale as keyof typeof formatMap]);
+    });
+
+    it("defaults to en-GB locale if no localecode string passed", () => {
```
DipperTheDan, line 519: question/nitpick: is it meant to be `locale code` or is `localecode` correct?

src/components/date/__internal__/date-picker/date-picker.spec.tsx
```diff
@@ -93,7 +127,7 @@ describe("DatePicker", () => {
 
     it(`should pass to the "DayPicker" component the "disabledDays"
         prop containing a null value`, () => {
```
DipperTheDan, line 129: question: does this description need updating as you've changed `null` to `undefined` in the `.toEqual()` assertion?

src/components/date/date.stories.tsx
edleeks87: Yeah that's why I've left it as wasn't sure what the state of his work was in etc, happy to update if James gives the green light

src/components/date/__internal__/date-formats/date-formats.spec.ts
edleeks87: probably either `local code` or `localCode`, I'll update accordingly

src/components/date/date.stories.tsx
DipperTheDan: Wonderful. Either which way is good for me 😄

src/components/date/date.spec.tsx
DipperTheDan: Figured there would have been but just wanted to make sure it was an accidental thing 👍🏻

src/components/date/__internal__/date-picker/date-picker.spec.tsx
```diff
@@ -221,7 +263,7 @@ describe("StyledDayPicker", () => {
       });
     });
 
-    describe.each([
+    describe.each<keyof typeof translations>([
```
robinzigmond, line 266:
```suggestion
    describe.each([
```

**suggestion** (non-blocking): as suggested elsewhere, this isn't necessary as long as you put an `as const` after the array

src/components/date/__internal__/navbar/navbar.component.tsx
```diff
@@ -1,23 +1,27 @@
...
+}
+
+export const Navbar = ({
+  onPreviousClick,
+  onNextClick,
+  ...props
+}: NavbarProps) => (
   <StyledNavbar {...props}>
-    <StyledButton onClick={() => onPreviousClick()}>
+    <StyledButton onClick={() => onPreviousClick?.()}>
```
robinzigmond, line 18:
```suggestion
    <StyledButton onClick={onPreviousClick}>
```

**suggestion** (non-blocking): any good reason why we can't simply do this?

Same for the "next" button of course

src/components/date/__internal__/navbar/navbar.component.tsx
```diff
@@ -1,23 +1,27 @@
...
-import PropTypes from "prop-types";
 import StyledButton from "./button.style";
 import StyledNavbar from "./navbar.style";
 import Icon from "../../../icon";
 
-const Navbar = ({ onPreviousClick, onNextClick, ...props }) => (
+export interface NavbarProps {
+  onPreviousClick?: () => void;
+  onNextClick?: () => void;
+  className?: string;
```
robinzigmond, line 9: **question**: do we need a `className` here? This is an internal component only, and while I see we have tests for the `className` prop working, we don't actually pass this anywhere. (Maybe I missed something.) If this isn't needed, lets just remove this from the interface (I notice it wasn't there before) and remove all the spreading props onto `StyledNavbar`.

src/components/date/__internal__/navbar/navbar.spec.tsx
```diff
@@ -1,12 +1,15 @@
...
+import { shallow, ShallowWrapper } from "enzyme";
 
-import Navbar from "./navbar.component";
+import Navbar, { NavbarProps } from "./navbar.component";
 import StyledButton from "./button.style";
 
 describe("Navbar", () => {
-  let wrapper, onPreviousClick, onNextClick;
+  let wrapper: ShallowWrapper;
+  let onPreviousClick: null | jest.Mock;
```
robinzigmond, line 10:
```suggestion
  let onPreviousClick: jest.Mock;
```

**suggestion** (non-blocking): I see no reason for the `null` option here or for `onNextClick`. If we remove that the `?.` before checking the `mock` property can go.

robinzigmond (commented):
note: this review is incomplete but is as far as I got before the end of the day. (Still Weekday to do, as well as the main Date component and the change inside Dialog - will do all those tomorrow.) 

EDIT: now complete, sorry for the delay

src/components/date/date.spec.tsx
```diff
@@ -263,8 +385,8 @@ describe("Date", () => {
   });
 
   describe("when the blur event is triggered on the input", () => {
-    let onBlurFn;
-    let onChangeFn;
+    let onBlurFn: jest.Mock | null;
```
robinzigmond, line 388:
```suggestion
    let onBlurFn: jest.Mock | undefined;
```

as in an earlier comment, this means that the `|| jest.fn` can be removed (line 466).

Sadly this doesn't seem possible for the `onChange` due to that not being an optional prop.

src/components/date/date.spec.tsx
```diff
@@ -558,10 +667,10 @@ describe("Date", () => {
   describe("initial value as ISO format", () => {
     describe.each(
       Object.keys(locales).filter((l) => !["en-CA", "en-US"].includes(l))
```
robinzigmond, line 669:
```suggestion
      Object.keys(locales).filter((l) => !["en-CA", "en-US"].includes(l))
```
```suggestion
      (Object.keys(locales) as (keyof typeof locales)[])filter((l) => !["en-CA", "en-US"].includes(l))
```

**suggestion** (non-blocking): I prefer the type assertion here instead of line 673 (where it should be removed from if you agree to this change) since here it's pretty obvious when reading it that it's a valid asertion, whereas it's less so on line 673.

src/components/date/date.spec.tsx
```diff
@@ -573,10 +682,10 @@ describe("Date", () => {
 
     describe.each(
       Object.keys(locales).filter((l) => ["en-CA", "en-US"].includes(l))
```
robinzigmond, line 684:
```suggestion
      (Object.keys(locales) as (keyof typeof locales)[])filter((l) => ["en-CA", "en-US"].includes(l))
```

**suggestion** (non-blocking): as above

src/components/date/date.spec.tsx
```diff
@@ -630,10 +741,10 @@ describe("Date", () => {
         simulateFocusOnInput(wrapper);
         act(() => {
           wrapper
-            .update()
+            ?.update()
             .find(DatePicker)
-            .props()
-            .onDayClick(mockDate, {}, { target: {} });
+            ?.props()
+            ?.onDayClick?.(mockDate, mockEvent);
```
robinzigmond, line 747:
```suggestion
            ?.onDayClick?.(mockDate, mockEvent);
```
```suggestion
            .onDayClick?.(mockDate, mockEvent);
```

**suggestion** (non-blocking): think you got carried away with optional chaining here, as this one doesn't seem to be necessary - `props` always returns a props object

src/components/date/date.spec.tsx
```diff
@@ -666,10 +779,10 @@ describe("Date", () => {
         jest.clearAllMocks();
         act(() => {
           wrapper
-            .update()
+            ?.update()
             .find(DayPicker)
-            .props()
-            .onDayClick(mockDate, {}, { target: {} });
+            ?.props()
+            ?.onDayClick?.(mockDate, mockDayModifiers, mockEvent);
```
robinzigmond, line 785:
```suggestion
            .onDayClick?.(mockDate, mockDayModifiers, mockEvent);
```

**suggestion** (non-blocking): as above

src/components/date/date.spec.tsx
```diff
@@ -697,17 +810,21 @@ describe("Date", () => {
           jest.clearAllMocks();
           act(() => {
             wrapper
-              .update()
+              ?.update()
               .find(DayPicker)
-              .props()
-              .onDayClick(mockDate, { disabled: true }, { target: {} });
+              ?.props()
+              ?.onDayClick?.(
```
robinzigmond, line 816:
```suggestion
            .onDayClick?.(
```

**suggestion** (non-blocking): as above

src/components/date/date.spec.tsx
robinzigmond: Actually it looks like nothing is needed here, just `: jest.Mock;` appears to be fine

src/components/date/date.spec.tsx
```diff
@@ -263,8 +385,8 @@ describe("Date", () => {
   });
 
   describe("when the blur event is triggered on the input", () => {
-    let onBlurFn;
-    let onChangeFn;
+    let onBlurFn: jest.Mock | null;
+    let onChangeFn: jest.Mock | null;
```
robinzigmond, line 389:
```suggestion
    let onChangeFn: jest.Mock;
```

**suggestion** (non-blocking): this `null` doesn't seem to be needed

src/components/date/date.spec.tsx
```diff
@@ -380,31 +508,31 @@ describe("Date", () => {
...
         simulateOnKeyDown(wrapper, "Enter");
-        expect(wrapper.update().find(DayPicker).exists()).toBe(true);
+        expect(wrapper?.update().find(DayPicker).exists()).toBe(true);
       });
     });
   });
 
   describe("when a mousedown or click event is triggered", () => {
-    let onClickFn;
+    let onClickFn: jest.Mock | null;
```
robinzigmond, line 535:
```suggestion
    let onClickFn: jest.Mock;
```

**suggestion** (non-blocking): again, the `null` seems unnecessary

src/components/date/date.spec.tsx
```diff
@@ -687,7 +800,7 @@ describe("Date", () => {
       });
 
       describe("when the disabled modifier is set", () => {
-        let onChangeFn;
+        let onChangeFn: jest.Mock | null;
```
robinzigmond, line 803:
```suggestion
        let onChangeFn: jest.Mock;
```

**suggestion** (non-blocking): same again here

src/components/date/date.spec.tsx
```diff
@@ -739,13 +856,15 @@ describe("Date", () => {
...
           (value) => {
-            wrapper.find("input").simulate("change", { target: { value } });
-            expect(wrapper.update().find("input").prop("value")).toEqual(value);
+            wrapper?.find("input").simulate("change", { target: { value } });
+            expect(wrapper?.update().find("input").prop("value")).toEqual(
+              value
+            );
             simulateBlurOnInput(wrapper);
-            const { separator } = locales[localeKey];
+            const { separator } = locales[localeKey as keyof typeof locales];
```
robinzigmond, line 864: **suggestion** (non-blocking): It won't let me comment on line 832, but similarly to changes suggested above, if that is changed to `describe.each(Object.keys(locales) as (keyof typeof locales)[])(` then not only do I think that is more easy to tell the assertion is sensible, it enables it to be removed from both here and line 886

src/components/date/date.spec.tsx
```diff
@@ -881,7 +1004,7 @@ describe("Date", () => {
   });
 
   describe("validation", () => {
-    it.each(["error", "warning", "info"])(
+    it.each<keyof ValidationProps>(["error", "warning", "info"])(
```
robinzigmond, line 1007:
```suggestion
    it.each(["error", "warning", "info"] as const)(
```

**suggestion** (non-blocking): personal preference but I think I prefer this.

Same on lines 1019 and 1045 (not going to bother making the same comment 3 times 😁 )

src/components/date/date.spec.tsx
```diff
@@ -945,32 +1068,34 @@ describe("Date", () => {
...
+      expect(wrapper?.find(DatePicker).props().inputElement.current).toBe(
         wrapper.find(StyledInputPresentation).getDOMNode()
       );
     });
   });
 
   describe("required", () => {
-    let input;
-    let label;
+    let input: ReactWrapper | null;
```
robinzigmond, line 1097:
```suggestion
    let input: ReactWrapper;
```

**suggestion** (non-blocking): as in a few comments above, this seems unnecessary. Same on the next line of course.

This would also mean the optional chaining on lines 1107/1111 isn't needed

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -10,14 +14,49 @@ import Navbar from "../navbar";
...
+
+export interface DatePickerProps {
+  /** Boolean to toggle where DatePicker is rendered in relation to the Date Input */
+  disablePortal?: boolean;
+  /** Minimum possible date YYYY-MM-DD */
+  minDate?: string;
+  /** Maximum possible date YYYY-MM-DD */
+  maxDate?: string;
+  /** Pass any props that match the DayPickerProps interface to override default behaviors */
+  pickerProps?: PickerProps;
```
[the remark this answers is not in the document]
edleeks87: Good question, I'd tested it with the props with the `Modifier/ Modifier[]` type and not found any issues setting it to `undefined`. However, testing it there again I've realised `[undefined]` isn't being accepted now 🤦 . Luckily the issue with the types package is only when unfolding/ reusing third party interfaces/ types so I can just add  `undefined[]` 


image

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -117,9 +160,8 @@ const DatePicker = React.forwardRef(
             fixedWeeks
             initialMonth={selectedDays || undefined}
             disabledDays={getDisabledDays(minDate, maxDate)}
-            inline
             locale={l.locale()}
-            localeUtils={{ formatDay }}
+            localeUtils={{ ...LocaleUtils, formatDay }}
```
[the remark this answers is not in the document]
edleeks87: Looks like a typo, we need to assert the type here as we only need `formatDay` but it's expecting several other utils. It should have been `const localeUtils = {formatDay} as LocaleUtils;`

src/components/date/__internal__/navbar/navbar.component.tsx
edleeks87: We need to support className as `react-day-picker` passes a class down to the navbar to handle the layout, I could probably remove the type and leave it as a spread but I'd prefer to type it and then explicitly set the className so it's more readable for the next person working on this

image

src/components/date/date.component.tsx
```diff
@@ -21,26 +21,84 @@ import {
...
+  onBlur?: (ev: DateChangeEvent) => void;
+  /** The current date string */
+  value: string;
+  /**
+   * Pass any props that match the DayPickerProps interface to override default behaviors
+   * See [DayPickerProps](https://react-day-picker-v7.netlify.app/docs/getting-started/) for a full list of available props
+   * */
+  pickerProps?: PickerProps;
+  /** Specify a callback triggered on keyDown */
+  onKeyDown?: (ev: React.KeyboardEvent<HTMLInputElement>) => void;
```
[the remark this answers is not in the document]
edleeks87: Good point I'd missed it coming from CommonInputProps

src/components/date/date.component.tsx
```diff
@@ -85,7 +143,7 @@ const DateInput = React.forwardRef(
       dateFnsLocale,
     ]);
     const { inputRefMap, setInputRefMap } = useContext(DateRangeContext);
-    const inputName = dataElement?.split("-")[0];
+    const inputName = dataElement?.split("-")[0] || "start";
```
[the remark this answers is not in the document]
edleeks87: I don't think this will create any issues tbh, but I don't really like the approach so I'm going to make a small refactor here and ensure that the it uses a unique id instead rather that the dataElement

src/components/date/date.component.tsx
```diff
@@ -268,22 +333,22 @@ const DateInput = React.forwardRef(
         isBlurBlocked.current = true;
       }
 
-      if (ev.target.type === "text" && !open) {
+      if ((ev.target as HTMLInputElement).type === "text" && !open) {
```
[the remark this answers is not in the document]
edleeks87:
```
const { type } = ev.target as HTMLInputElement;

      if (type !== "text") {
        alreadyFocused.current = true;
        setOpen((prev) => !prev);
      } else if (!open) {
        setOpen(true);
      }
```

src/components/date/date.component.tsx
```diff
@@ -299,8 +364,12 @@ const DateInput = React.forwardRef(
           }
         }
 
-        if (inputRefMap && inputRefMap[inputName]?.setOpen !== setOpen) {
-          setInputRefMap({
+        if (
+          inputRefMap &&
+          inputRefMap[inputName as keyof typeof inputRefMap]?.setOpen !==
```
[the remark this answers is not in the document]
edleeks87: `dataElement` is set in DateRange for each input, it cannot be overridden. However, I don't love the approach here so I'm going to use a different unique id and name it something more intuitive

image

src/components/date/date.component.tsx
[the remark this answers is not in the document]
edleeks87: I'll make this change as a new commit which will go after the ones refactoring the files to TS

src/components/date-range/date-range-test.stories.tsx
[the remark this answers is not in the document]
edleeks87: good spot thanks

src/components/date/__internal__/utils.ts
```diff
@@ -1,30 +1,39 @@
...
+  if (!dateValue || !isDateValid(dateValue)) {
     return null;
   }
 
   return formatISO(dateValue).split("T")[0];
 }
 
-export function formattedValue(formatString, value) {
+export function formattedValue(formatString: string, value?: Date) {
+  if (!value) return "";
```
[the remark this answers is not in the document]
edleeks87: we can't pass undefined to `format`

src/components/date/__internal__/utils.ts
```diff
@@ -1,30 +1,39 @@
...
-export function formatToISO(formatString, valueString) {
+export function isDateValid(date?: Date) {
+  return date && date.toString() !== "Invalid Date";
+}
+
+export function formatToISO(formatString?: string, valueString?: string) {
   const dateValue = parseDate(formatString, valueString);
 
-  if (!isDateValid(dateValue)) {
+  if (!dateValue || !isDateValid(dateValue)) {
```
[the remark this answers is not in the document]
edleeks87: I thought so too but when I don't have it the ts compiler complains about passing undefined to `formatISO`

image

DipperTheDan (approved):

src/components/date-range/date-range.spec.tsx
```diff
@@ -803,10 +796,11 @@ describe("StyledDateRange", () => {
   });
 
   describe("validation layout", () => {
-    it.each(["small", "medium", "large"])(
+    it.each<DateInputProps["size"]>(["small", "medium", "large"])(
```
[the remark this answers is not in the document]
edleeks87: Same comment, I'd prefer to leave it until we agree as a team about the approach as we have utilised both currently

src/components/date-range/date-range.spec.tsx
```diff
@@ -803,10 +796,11 @@ describe("StyledDateRange", () => {
...
       "sets the maxWidth of the %s Date input wrappers to the expected value",
       (size) => {
-        const maxWidth = datePickerWidths[size];
+        const maxWidth =
+          datePickerWidths[size as keyof typeof datePickerWidths];
```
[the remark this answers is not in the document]
edleeks87: see comment above

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -68,19 +107,24 @@ const DatePicker = React.forwardRef(
...
-      if (!disabled) {
-        const { id, name } = inputElement?.current?.firstChild;
+    const handleDayClick = (
+      date: Date,
+      modifiers: DayModifiers,
+      ev: React.MouseEvent<HTMLDivElement>
+    ) => {
+      if (!modifiers.disabled) {
+        const { id, name } = inputElement?.current
+          ?.children[0] as HTMLInputElement;
```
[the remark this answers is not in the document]
edleeks87: reverted, I suspect it was complaining as I refactored but I then amended what it was complaining about

src/components/date/date.component.tsx
```diff
@@ -21,26 +21,84 @@ import {
...
+   * */
+  pickerProps?: PickerProps;
+  /** Specify a callback triggered on keyDown */
+  onKeyDown?: (ev: React.KeyboardEvent<HTMLInputElement>) => void;
+}
 
 let deprecateInputRefWarnTriggered = false;
 
-const DateInput = React.forwardRef(
+export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
```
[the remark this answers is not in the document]
edleeks87: See reply above to similar comment

src/components/date/date.spec.tsx
```diff
@@ -129,7 +251,7 @@ describe("Date", () => {
       ["small", "120px"],
       ["medium", "135px"],
       ["large", "140px"],
-    ])("to %s", (size, expectedValue) => {
+    ] as SizesAndWidths)("to %s", (size, expectedValue) => {
```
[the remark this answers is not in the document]
edleeks87: See other response relating to agreeing a preferred approach

src/components/date/date.spec.tsx
edleeks87: see previous comment

src/components/date/__internal__/navbar/navbar.component.tsx
robinzigmond: hmm, I've looked in the Cypress tests for Date and can't find any that look like they do this - but I'll take your word for it. Happy to accept the change for now but really this is saying to me that we should change that Cypress test. Presumably when the Cypress test is updated to TS it will have to update anyway, as you can't call an optional function prop!

src/components/date/__internal__/utils.ts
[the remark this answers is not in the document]
robinzigmond: oh yes, you're right - I get the same on running the the type-check command. That's really weird as VSCode isn't underlining it, and when I hover over the `formatISO` function it tell me it accepts `Date | undefined`. Not sure what's going on but as the additional check clearly doesn't do any harm I'll just accept this for now.

src/components/date-range/date-range.component.tsx
```diff
@@ -134,26 +221,27 @@ const DateRange = ({
...
 
   const buildCustomEvent = useCallback(
-    (changedDate, newValue) => {
-      const startValue =
-        changedDate === "start" && newValue ? newValue : startDateValue;
-      const endValue =
-        changedDate === "end" && newValue ? newValue : endDateValue;
+    (
+      changedDate: string,
+      newValue: { formattedValue: string; rawValue: string | null }
```
robinzigmond, line 226:
```suggestion
      newValue: DateInputValue
```

**suggestion**: We should do this now that we've extracted it as a separate type.

src/components/date-range/date-range.spec.tsx
```diff
@@ -25,36 +25,58 @@ const datePickerWidths = {
...
+    <DateRange value={initialValues} onChange={() => {}} {...props} />,
+    {
+      attachTo: document.getElementById("enzymeContainer"),
+    }
+  );
+}
+
+function simulateFocusOnInput(
+  container: ReactWrapper | null,
+  dateId: "start" | "end"
```
robinzigmond, line 39:
```suggestion
  dateId: InputName
```

**suggestion** (obviously we'd have to import that type from the context file)

src/components/date-range/date-range.spec.tsx
```diff
@@ -25,36 +25,58 @@ const datePickerWidths = {
...
+) {
+  const input = container?.find("input").at(dateId === "start" ? 0 : 1);
+
+  input?.simulate("focus");
+}
+
+function simulateBlurOnInput(
+  container: ReactWrapper | null,
+  dateId: "start" | "end"
```
robinzigmond, line 48:
```suggestion
  dateId: InputName
```

**suggestion**: as above

src/components/date/date.component.tsx
```diff
@@ -21,26 +20,88 @@ import {
...
+   * Pass any props that match the DayPickerProps interface to override default behaviors
+   * See [DayPickerProps](https://react-day-picker-v7.netlify.app/docs/getting-started/) for a full list of available props
+   * */
+  pickerProps?: PickerProps;
+  /**
+   * @private
+   * @ignore
+   * Name passed from DateRange to allow it to know which input is updating
+   * */
+  inputName?: "start" | "end";
```
robinzigmond, line 99:
```suggestion
  inputName?: InputName;
```

**suggestion**: similarly to the couple of uses in the DateRange spec file (again it will have to be imported)

src/components/date/date.spec.tsx
```diff
@@ -86,20 +86,142 @@ const locales = {
...
+    rawValue: string | null;
+  }) => void | jest.Mock;
+}
+
 describe("Date", () => {
-  let container;
-  let wrapper;
-  let onFocusFn;
+  let container: HTMLDivElement | null;
+  let wrapper: ReactWrapper | null | null;
```
robinzigmond, line 213:
```suggestion
  let wrapper: ReactWrapper | null;
```

**suggestion**: typo I assume?

edleeks87: I hope so 😆

robinzigmond (dismissed):

robinzigmond (approved):

DlgSHi (approved):

PR #6087, opened 2023-06-05
fix(popover-container): prevent click outside calling onclose when container is not open

### LINKED ISSUE #6066: Popover container component calls onClose method even if the popover is not opened.
### Current behaviour

onClose is called each time the use click on the page

HandleClickAway calls "onClose" even if the popover isn't opened.
So each time the user click on a page with a popover (not opened) the onClose is called

popover-container.component.tsx
line 210

const handleClickAway = (e: Event) => {
    if (!isControlled) setIsOpenInternal(false);
    if (onClose) onClose(e);  // <<<<<<<<<<<<<<<<<
  };
  const handleClick = useClickAwayListener(handleClickAway, "mousedown");

### Expected behaviour

Call onClose only if the popover was opened.

https://carbon.sage.com/?path=/docs/popover-container--default

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

118.3.2

### Design Tokens Version

_No response_

Firefox, Chrome

Windows

Obviously there are workarounds in user's code but you can search a while before finding the reason of React refresh.

### Confidentiality

robinzigmond:
Clicking outside the Popover container should call the onClose callback only when the container was previously open and is closed by the click - subsequent clicks should not call the callback.

fix #6066

### Proposed behaviour

The onClose callback should not be called when clicking outside the container if it is not currently open.

### Current behaviour

The onClose callback is called when clicking anywhere on the page other than the button that opens the container.

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/quizzical-tu-2on5yy?file=/src/App.tsx

Test by observing the console output when clicking anywhere on the page (outside the container-open button) when the container is not open.

Ensure the callback still runs as expected when clicking outside to close an open container.


edleeks87 (approved):

robinzigmond:
Please do not review this yet as I just realised an issue with the codesandbox from the PR - the onClose callback is not getting called when actually closing the container via click-away, as I only catered for the controlled case. I'll fix that and put this back in review.

edleeks87 (approved):

cypress/components/popover-container/popover-container.cy.js
Parsium, line 1: **suggestion(non-blocking)**: The cypress assertions we have here could be rewritten using aliases:

```tsx
const callback = cy.stub().as("callback");

CypressMountWithProviders(
   <PopoverContainerComponent onClose={callback} open />
);

cy.root().click();
cy.get("@callback").should("have.been.calledOnce");
```

This would allow us to avoid disabling eslint and ensure we are following [cypress's recommended approach with checking callbacks are called](https://docs.cypress.io/guides/core-concepts/retry-ability#Incorrectly-checking-if-the-stub-was-called).

Parsium (approved):

edleeks87 (approved):

stephenogorman (approved):

PR #6104, opened 2023-06-08
test(form): typescript refactor

stephenogorman:
### Proposed behaviour

- Rename the `form.cy.js` to `form.cy.tsx` file in `./cypress/components/form/` folder.

### Current behaviour

Form tests are in js not ts.

### Checklist

#### QA

### Testing instructions
- [ ] Run `npx cypress open --component` to check if the`form.cy.tsx` file passed
- [ ] Run `npx cypress run --component` to check none of the other *.cy.tsx files have regressed


ZhuoyuJin (approved):

cypress/components/form/form.cy.tsx
```diff
@@ -99,15 +103,12 @@ context("Tests for Form component", () => {
       }
     );
 
-    it("should call onSubmit callback when a click event is triggered", () => {
-      const callback = cy.stub();
+    it("should call onSubmit callback when a submit event is triggered", () => {
```
Parsium, line 106: **issue:** it seems this test currently is [causing Cypress to repeatedly re-run the tests](https://github.com/Sage/carbon/actions/runs/5210847135/jobs/9402423440), causing an infinite loop. 

This is because the browser's default behaviour when a HTML form is submitted is to redirect the user to a new page. Here when we submit the form by clicking the save button, it is causing the page to reload, which seems to result in Cypress re-running the preceding tests.

To fix this, I would suggest doing the following, where we pass a callback function which prevents this reload behaviour, then pass it to the component and spy on it. This should be similar to how the `onSubmit` prop will be used by consumers in practice.

```tsx
it("should call onSubmit callback, when onSubmit prop is passed and save button is triggered", () => {
  const onSubmit: FormProps["onSubmit"] = (event) => {
    event.preventDefault(); // this prevents the default reloading behaviour
  };

  const onSubmitSpy = cy.spy(onSubmit).as("onSubmitSpy");

  CypressMountWithProviders(
    <FormComponent
      onSubmit={onSubmitSpy}
      saveButton={
        <Button buttonType="primary" type="submit">
          Save
        </Button>
      }
    />
  );

  cy.contains("Save").click();
  cy.get("@onSubmitSpy").should("have.been.called");
});
```

DlgSHi (approved):

Parsium (approved):

PR #6110, opened 2023-06-13
fix(select): fix accessible name

### LINKED ISSUE #5945: Accessible name of input does not match visual label
### Current behaviour

While updating to Carbon v116.2.0, specs to get the `<Select />` component within a form using the `getByRole` query were failing as the name option string was required to be the the visual label text and the value text concatenated. This was noticed in the `react-testing-library` output: see attachment

On inspection of the DOM in the browser, the value for the `aria-labelledby` attribute for the `input` element within the `<Select />` component is given the value of the combined `IDs` of the `label` element which holds the label string and the `span` element which holds the value string and so the label text while testing (`screen.getByRole("combobox", { name: "Label text here" })`) is considered to be the string in the label element and in the span element as opposed to the visual label which is not correct. See DOM screenshot: 
image

This also results in the screen reader (VoiceOver) reading the field label including the value (County Please select...).

### Expected behaviour

The accessible name of the input should match the visual label and should not contain the value so while testing, the element can be procured by using the visual label text via the `getByRole` query. Also this will ensure the screen reader reads the label correctly and communicates it to the user.

https://codesandbox.io/s/silly-tu-8msf5x

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

116.2.0

### Design Tokens Version

_No response_

Microsoft Edge

MacOS

_No response_

### Confidentiality

robinzigmond:
Simple select and Filterable select both use a 2-part aria-labelledby that references both the field's label and a hidden element containing the currently-selected  value. This results in the accessible name being different depending on the selected value, which is unexpected to users who will expect that the accessible name is simply the label.

Note that MultiSelect has been unchanged for now, pending further investigation with UX/accessibility of the best way to make this component accessible.

As a result of these changes, Select and FilterableSelect do not automatically get an accessible name provided, so all uses need either a `label`, `aria-label` or `aria-labelledby` to pass automatic accessibility checks. In particular, `pager` has had to be updated to ensure the page size select always has an accessible name even if `showPageSizeLabelBefore` is false.

fix #5945

### Proposed behaviour

When no `aria-label` or `aria-labelledby` is provided, the accessible name of a simple or filterable select should be simply the visual label.

### Current behaviour

The accessible name of a simple or filterable select is the label followed by the currently-selected value (or "Please select" if there is no currently-selected value).

### Checklist

#### QA

### Testing instructions

Test all 3 select variants, with `aria-label`, `aria-labelledby` and neither, and ensure the behaviour is as expected:
- on simple and filterable select, the accessible name should be the visual label if neither `aria-label` or `aria-labelledby` is provided, and the expected value if one of them is
- on multiselect, there should be no change from the previously-released version (the accessible name consists of the visual label followed by all the currently-selected options)

[Note that the "accessible name" is what is announced by a screenreader along with "combobox" and the current value, and can be inspected in the developer tools in Chrome and Firefox.]

Test with a screenreader and ensure the components are still usable and the announcements make sense. (Not needed for MultiSelect other than to ensure there are no serious regressions - accessibility improvements to MultiSelect are currently under discussion.)

Test the Pager component with a screenreader with all combinations of the `showPageSizeLabelBefore` and `showPageSizeLabelAfter` props (with `showPageSizeSelection=true`) to ensure there is an appropriate accessible name for the select in all cases.


src/components/pager/pager.component.tsx
```diff
@@ -225,6 +225,9 @@ export const Pager = ({
           onKeyDown={handleKeyDown}
           data-element="page-select"
           id={pageSizeSelectId}
+          {...(!showPageSizeLabelAfter && !showPageSizeLabelAfter
```
DipperTheDan, line 228: question: Is this a mistake that you're checking `!showPageSizeLabelAfter` twice?

robinzigmond: yes, it is - should be checking ! them both. Worrying that this isn't causing a test failure 🤦

DipperTheDan (dismissed):

src/components/select/select-textbox/select-textbox.component.tsx
```diff
@@ -211,19 +201,11 @@ const SelectTextbox = React.forwardRef(
...
-      const joinIds = (...ids: (string | undefined)[]) =>
-        ids.filter((item) => item !== undefined).join(" ");
-      const combinedAriaLabelledBy = hasTextCursor
-        ? joinIds(
-            ariaLabelledby || labelId,
-            accessibilityLabelId || textId.current
-          )
-        : joinIds(ariaLabelledby || labelId, textId.current);
-
       return {
```
tomdavies73, line 204: **nitpick (non-blocking):** if this function just has a return statement after your changes, could we refactor to an arrow function to be more concise? Again not essential and I may be wrong

src/components/select/select-textbox/select-textbox.component.tsx
```diff
@@ -234,12 +216,7 @@ const SelectTextbox = React.forwardRef(
     function renderSelectText() {
       return (
```
tomdavies73, line 217: **nitpick (non-blocking):** same comment from above might apply here (if its correct in the first place)

src/components/select/select-textbox/select-textbox.spec.tsx
```diff
@@ -185,15 +159,15 @@ describe("SelectTextbox", () => {
...
+      ["not set", undefined],
     ])(
       "when the ariaLabelledBy prop is %s",
-      (description, propName, mockAriaLabelledBy) => {
+      (description, mockAriaLabelledBy) => {
         const mockLabelId = "baz";
 
         describe("with the hasTextCursor prop set to true", () => {
-          it(`then the aria-labelledby in the Textbox should contain ${propName} and the id of the SelectText`, () => {
+          it(`then the aria-labelledby in the Textbox should be the value of the ariaLabelledBy prop`, () => {
```
tomdavies73, line 170: **nitpick (non-blocking):** I know this `it` was a template literal/string for string interpolation, but as you've edited it, and it is now just a string, it might be worth changing it back to quotes, to make sure the use of backticks and quotations are consistent.

src/components/pager/pager.spec.tsx
```diff
@@ -405,6 +405,46 @@ describe("Pager", () => {
 
       expect(wrapper.find(StyledPagerSummary).text()).toBe("");
     });
+
+    it("renders a label correctly-linked to the select when 'showPageSizeLabelBefore' is true, and does not add an aria-label", () => {
```
tomdavies73, line 409: **nitpick (non-blocking):** Could these two `its` be changed to an `it.each`? As far as I can tell, these tests are identical, apart from `showPageSizeLabelBefore` and `showPageSizeLabelAfter` having opposite truth\false values. This could help condense the file, I am aware `it.each`s are less readable at times, but when there will be a small amount of parameters, like here, it could be worth doing.

src/components/select/select-textbox/select-textbox.component.tsx
robinzigmond: it's a fair comment. I personally don't much care either way - note that because this returns an object literal it doesn't work to just do `() => { key: value, etc. }` because JS interprets the `{...}` as the function body and not for an object. It still works though if you do `() => ({...})` instead. I don't think it's better than leaving the `return` in but it's not worse either, happy to make this change if you think it's for the best.

src/components/select/select-textbox/select-textbox.component.tsx
robinzigmond: it's a much fairer point for this one that the one above. In fact since this is only used in one place I can probably remove the function entirely and just inline the JSX at the point it's currently called. Happy to do that 👍

src/components/select/select-textbox/select-textbox.spec.tsx
robinzigmond: Excellent point thanks, I'd missed this. I'll get it changed.

src/components/pager/pager.spec.tsx
robinzigmond: that's another good point, thanks. In my head these were different tests because the label referenced by `selectLabel` is actually in a different place in the DOM in the two cases - but that doesn't actually change how the tests are structured or what they're fundamentally testing. I agree with you and will wrap them in an `it.each`.

tomdavies73 (dismissed):

DipperTheDan (dismissed):

src/components/pager/pager.stories.mdx
```diff
@@ -116,7 +116,7 @@ Due to the fact that it is last page, next and last links are disabled.
 ### Custom responsive example
 
 The`show...` props can also be used to implement responsive behaviour in the `Pager`. Below is an example that will
-conditionally render the interanl elements as the screen size is adjusted. This example is best viewed in the Canvas tab
+conditionally render the internal elements as the screen size is adjusted. This example is best viewed in the Canvas tab
```
DipperTheDan, line 119: **praise:** Good spot 👍🏻

DipperTheDan (approved):

tomdavies73 (approved):

tomdavies73 (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #6114, opened 2023-06-13
fix(button): render falsy children

### LINKED ISSUE #6108: [ Button ] - Children prop evaluate to falsy
### Current behaviour

Given a button component
When the `iconType` prop is not passed
And `children` prop is falsy 
Then the children won't be rendered

### Expected behaviour

Given a button component
When the `iconType` prop is not passed
And `children` prop is falsy 
Then the children is rendered

https://codesandbox.io/s/prod-glitter-ypps3s?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

119.0.1

### Design Tokens Version

4.19.0

Chrome

MacOS

_No response_

### Confidentiality

MarcServat:
fix #6108

### Proposed behaviour
The Button will render all falsy values as a children unless `undefined` and `false`

### Current behaviour
The Button treat falsy values as `false`. As a consequence does not render falsy values, like 0, within the `span` and also the component throws an error when the iconType prop is not passed and a falsy value is passed as children. 

### Checklist

#### QA

### Testing instructions


Parsium (approved):

tomdavies73 (approved):

src/components/button/button.component.tsx
```diff
@@ -131,12 +131,14 @@ RenderChildrenProps) {
     bgSize: "extra-small",
   };
 
+  const isValidChildren = children !== undefined && children !== false;
+
   return (
     <>
-      {iconType && iconPosition === "before" && children && (
+      {iconType && iconPosition === "before" && isValidChildren && (
```
nineteen88, line 138: suggestion: we could just use Boolean(children) to achieve the same goal

MarcServat: not really, because the `Boolean` will convert to `false` a falsy value resulting in the same behavior as before. e.g.
```Boolean(0) // false```

nineteen88: Yeah fair point and forgot the point of the fix. No worries!

nineteen88 (approved):

ZhuoyuJin (approved):

PR #6118, opened 2023-06-14
test(grid): typescript refactor

stephenogorman:
### Proposed behaviour

- Rename the `grid.cy.js` to `grid.cy.tsx` file in `./cypress/components/grid/` folder.

### Current behaviour

Grid tests are in js not ts.

### Checklist

#### QA

### Testing instructions
- [ ] Run `npx cypress open --component` to check if the`grid.cy.tsx` file passed
- [ ] Run `npx cypress run --component` to check none of the other *.cy.tsx files have regressed

### DIFF
--- a/cypress/components/grid/grid.cy.tsx
+++ b/cypress/components/grid/grid.cy.tsx
@@ -7,7 +7,7 @@ import Pod from "../../../src/components/pod";
 import CypressMountWithProviders from "../../support/component-helper/cypress-mount";
 import { assertCssValueIsApproximately } from "../../support/component-helper/common-steps";
 
-const viewportSize = (sizeOfViewport) => {
+const viewportSize = (sizeOfViewport: string) => {
   switch (sizeOfViewport) {
     case "default":
       cy.viewport(1958, 900);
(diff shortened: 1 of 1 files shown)


cypress/components/grid/grid.cy.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/valid-expect, jest/valid-expect-in-promise */
```
ZhuoyuJin, line 1: Can we get rid of this disable line?

stephenogorman: This was added at James's suggestion as the 'expect' which requires it is difficult to replace - when the eslint rules are redefined it shouldn't be required anyway.

ZhuoyuJin: Alright, that makes sense. Alternatively I think we could use ".should("be.gte", number1). and("be.lte", number2)" to avoid 'expect'

ZhuoyuJin (dismissed):

cypress/components/grid/grid.cy.tsx
DlgSHi: actually we, can use `cy.wrap($element).should(...)` instead of except. Or use an `alias`

DlgSHi (changes requested):

cypress/components/grid/grid.cy.tsx
stephenogorman: I had tried numerous combinations of your suggestions yesterday and was getting the same "Promise should be returned to test its fulfillment or rejection" error using cy.wrap. Anyway, what I've pushed a change which seems to be working without the disable line.

DlgSHi (approved):

PR #6125, opened 2023-06-15
feat(switch): add support for new validation - FE-5755

DipperTheDan:
### Proposed behaviour

Screenshot 2023-06-15 at 16 22 07

### Current behaviour

New validations are not currently implemented in `Switch`

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/__internal__/validation-message/validation-message.spec.tsx
+++ b/src/__internal__/validation-message/validation-message.spec.tsx
@@ -35,7 +35,7 @@ describe("ValidationMessage component", () => {
           color: error
             ? "var(--colorsSemanticNegative500)"
             : "var(--colorsSemanticCaution600)",
-          fontWeight: error ? "bold" : "regular",
+          fontWeight: error ? "bold" : "normal",
           marginTop: "0px",
           marginBottom: "8px",
         },
--- a/src/__internal__/validation-message/validation-message.style.ts
+++ b/src/__internal__/validation-message/validation-message.style.ts
@@ -9,7 +9,7 @@ const StyledValidationMessage = styled.p<StyledValidationMessageProps>`
     color: ${isWarning
       ? "var(--colorsSemanticCaution600)"
       : "var(--colorsSemanticNegative500)"};
-    font-weight: ${isWarning ? "regular" : "bold"};
+    font-weight: ${isWarning ? "normal" : "bold"};
     margin-top: 0px;
     margin-bottom: 8px;
   `}


src/__internal__/form-field/form-field.component.tsx
```diff
@@ -190,7 +191,7 @@ const FormField = ({
             inline={inlineLabel}
             width={labelWidth}
             optional={isOptional}
-            useValidationIcon={useValidationIcon}
+            useValidationIcon={!validationRedesignOptIn && useValidationIcon}
```
robinzigmond, line 194: **question**: Again, was this deliberate that it affects other input components besides just `switch`. If you use the `validationOnLabel` prop with the validation redesign flag, the current behaviour is shown in this sandbox https://codesandbox.io/s/gallant-wave-qpr3yk?file=/src/App.tsx - which could well be wrong as it has the old-style tooltip validation despite the flag (but we'd need UX confirmation of that).

In your branch, the same code now shows "new style" validation - so arguably that's a fix/improvement. But it's no longer clear that `validationOnLabel` actually does anything, so I think there's a bit more to it perhaps. And even if this is a needed fix for new validation in general it probably shouldn't be done in a PR that's supposed to be just about updating `switch`.

src/components/switch/switch.component.tsx
```diff
@@ -139,7 +144,9 @@ export const Switch = React.forwardRef(
       error,
       warning,
       info,
-      useValidationIcon: !shouldValidationBeOnLabel && !disabled,
+      useValidationIcon: !validationRedesignOptIn
```
robinzigmond, line 147: **suggestion** (non-blocking) - I personally think this is much easier to understand when written as `useValidationIcon: !validationRedesignOptIn && !shouldValidationBeOnabel && !disabled`

robinzigmond: Further thoughts on this - I could be barking up completely the wrong tree but I thought worth mentioning as it might be helpful:

As far as just this property is concerned, it would be neater if `useValidationIcon` stayed the same as before but the definition of `shouldValidationBeOnLabel` changed so that it took `validationRedesignOptIn` into account.

Then I realised that `shouldValidationBeOnLabel` is also used in `validationOnLabel` so any such change would also affect the value of that prop that's passed to `CheckableInput`.

BUT it's possible that doing this will actually fix `validationOnLabel` so that it's doing what we want in `CheckableInput` and doesn't require messing with `FormField` which, as I observed above, is causing some issues as it stands now.

Not making this an "official" comment but maybe there's something useful in there, even if unlikely 😅

src/components/switch/switch.component.tsx
```diff
@@ -157,8 +164,7 @@ export const Switch = React.forwardRef(
       onChange: isControlled ? onChange : onChangeInternal,
       id,
       name,
-      label,
-      labelHelp,
+      labelHelp: !validationRedesignOptIn ? labelHelp : null,
```
robinzigmond, line 167: **comment** (non-blocking): as far as I can tell the functionality is working fine here, but I find it a bit inconsistent (and was confused at first) how you're handling the `label` prop vs `labelHelp`. In both cases we only want to pass it to `CheckableInput` if `validationRedesignOptIn` is false, but for `labelHelp` you've expressed that with a ternary within `inputProps`, whereas for `label` you've removed it entirely and manually added it as a separate prop in the JSX in the appropriate part of the larger ternary.

I don't have a strong preference for which way is better but the inconsistency bothers me a bit.

src/components/switch/switch.spec.tsx
```diff
@@ -581,6 +593,34 @@ describe("Switch", () => {
     });
   });
 
+  describe("New validations", () => {
+    const validationTypes = ["error", "warning"] as const;
```
robinzigmond, line 597: **comment** (non-blocking): is there a good reason not to also use `info` in these tests?

DipperTheDan: There is no `info` in the new validations. I'll send you a link to the Figma designs in a private chat.

src/__internal__/form-field/form-field.component.tsx
```diff
@@ -160,15 +160,16 @@ const FormField = ({
...
-    <FieldHelp
-      labelInline={inlineLabel}
-      labelWidth={labelWidth}
-      id={fieldHelpId}
-    >
-      {fieldHelpContent}
-    </FieldHelp>
-  ) : null;
+  const fieldHelp =
+    fieldHelpContent && !validationRedesignOptIn ? (
```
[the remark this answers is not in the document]
Parsium: @robinzigmond if you are able to do this with `Textbox`, then it would be a good idea to assess which props related to the old validation pattern remain in Carbon and will need removing in the future.

src/components/switch/switch.component.tsx
DipperTheDan: Good shout. I'll make sure it's changed to `useValidationIcon: !validationRedesignOptIn && !shouldValidationBeOnabel && !disabled`

src/components/switch/switch.component.tsx
DipperTheDan: Good point. I'll pick a way to do both and keep it consistent. I'll pick a way which makes sense for when we remove the old design and the need for the opt in flag.

src/components/switch/switch.component.tsx
Parsium, line 1: **thought:** What do you think about only using `validationRedesignOptIn` to set the relevant props within `switch.component.tsx`, as opposed to using it in the internal components as well?

Given a lot of features like tooltips, validation icons, field help elements and inline variants won't be supported with the new patterns, this could be an opportunity for us to simplify our internal components's code.

If we use `validationRedesignOptIn` to set the relevant props within `Switch` only, then later on it could help us identity which internal components contain useless code and can be refactored (or removed entirely).

Parsium (commented):
**suggestion:** so we don't forget to remove the props related to the old validation pattern at a later date, what do you think about deprecating the props related to field helps, validation icons, tooltips and inline labels?

It might also be helpful for consumers if we provide an explanation for these deprecations in the docs for the newly-added new validation story, so they understand its related to the `validationRedesignOptIn` feature flag.

src/components/switch/switch.component.tsx
```diff
@@ -169,16 +175,34 @@ export const Switch = React.forwardRef(
...
-        tooltipPosition={tooltipPosition}
-      >
-        <StyledSwitch {...switchStyleProps}>
-          <CheckableInput {...inputProps}>
-            <SwitchSlider {...switchSliderProps} />
-          </CheckableInput>
-        </StyledSwitch>
-      </TooltipProvider>
+      <>
+        {validationRedesignOptIn ? (
```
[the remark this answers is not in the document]
DipperTheDan: Thank you. That was the plan. Means we can just delete the `false` path and keep the `true` path.

src/__internal__/form-field/form-field.component.tsx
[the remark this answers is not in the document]
DipperTheDan: `fieldHelp` doesn't render anything in the new style validation. It only works for old style. I agree with a separate task to identify the props related to the old style validation. It'll make decoupling it easier when it comes to deprecation time. 

Screenshot to demonstrate how `fieldHelp` doesn't render in new validations. 
Screenshot 2023-06-22 at 14 29 56

robinzigmond (approved):

src/components/switch/switch.component.tsx
Parsium, line 1: **suggestion:** the `reverse` and `labelSpacing` props can be removed from `switchStylePropsForNewValidation` and `inputPropsForNewValidation` objects.

As these don't seem to be utilised in the new validation pattern.

Parsium (dismissed):

robinzigmond (dismissed):

ZhuoyuJin (dismissed):

Parsium (approved):

Parsium (approved):

robinzigmond (approved):

ZhuoyuJin (approved):

PR #6142, opened 2023-06-22
fix(decimal): ensure empty value is formatted correctly - FE-5892

### LINKED ISSUE #6012: `Decimal`'s emptyValue does not follow `precision` attribute
### Current behaviour

**Given** `Decimal` component
**When** I set `precision` attribute to 0
**And** I empty the value and blur the input
**Then** I see '0.00' filled in

### Expected behaviour

**Given** `Decimal` component
**When** I set `precision` attribute to 0
**And** I empty the value and blur the input
**Then** I see '0' filled in

https://codesandbox.io/s/dreamy-sunset-ne2lq6

### JIRA Ticket (Sage Only)

SBS-58568

### Suggested Solution

 Please could you adjust the default emptyValue 0 to adhere to the precision attribute setting or just 0?

### Carbon Version

117.6.1

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

DipperTheDan:
A bug was present whereby if `allowEmptyValue` was false and precision value was less than 2, the value of Decimal would not be formatted correctly. This has been rectified by making sure that an empty value is formatted  with precision and locale taken in to consideration.

fixes #6012

### Proposed behaviour

When `allowEmptyValue` is false, the value displayed should follow the precision and locale set by the consumer. 

### Current behaviour

When `allowEmptyValue` is false, the value displayed does not follow the precision and locale set by the consumer if precision value is less than 2. 

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- There should be not regressions with functionality.
- CSB link to test functionality. CSB is built from the one provided on the original issue: https://codesandbox.io/s/pensive-edison-lhqw2t

### DIFF
--- a/cypress/components/decimal/decimal.cy.tsx
+++ b/cypress/components/decimal/decimal.cy.tsx
@@ -268,6 +268,92 @@ context("Tests for Decimal component", () => {
     });
   });
 
+  describe("allowEmptyValue", () => {
+    it.each([
+      [0, "en", "0", "0"],
+      [1, "en", "0.0", "0.0"],
+      [2, "en", "0.00", "0.00"],
+      [0, "es-ES", "0", "0"],
+      [1, "es-ES", "0.0", "0,0"],
+      [2, "es-ES", "0.00", "0,00"],
+      [0, "fr", "0", "0"],
+      [1, "fr", "0.0", "0,0"],
+      [2, "fr", "0.00", "0,00"],
+      [0, "pt-PT", "0", "0"],
+      [1, "pt-PT", "0.0", "0,0"],
+      [2, "pt-PT", "0.00", "0,00"],
+    ] as [DecimalProps["precision"], string, string, string][])(
+      "should format an empty value correctly when precision is %s, locale is %s and allowEmptyValue is false",
+      (precisionValue, localeValue, rawValue, expectedValue) => {
+        const callback: DecimalProps["onBlur"] = cy.stub().as("onBlur");
+
+        CypressMountWithProviders(
+          <Decimal
+            precision={precisionValue}
+            locale={localeValue}
+            allowEmptyValue={false}
+            onBlur={callback}
+          />
+        );
+
+        commonDataElementInputPreview().type("100");
+        commonDataElementInputPreview()
+          .clear()
+          .blur({ force: true })
+          .then(() => {
+            cy.get("@onBlur").should(
+              "have.been.calledWith",
+              eventOutput(expectedValue, rawValue)
+            );
+          });
+        expect(
+          commonDataElementInputPreview().should("have.value", expectedValue)
+        );
+      }
+    );
+
+    it.each([
+      [0, "en"],
+      [1, "en"],
+      [2, "en"],
+      [0, "es-ES"],
+      [1, "es-ES"],
+      [2, "es-ES"],
+      [0, "fr"],
+      [1, "fr"],
+      [2, "fr"],
+      [0, "pt-PT"],
+      [1, "pt-PT"],
+      [2, "pt-PT"],
+    ] as [DecimalProps["precision"], string][])(
+      "should format an empty value correctly when precision is %s, locale is %s and allowEmptyValue is true",
+      (precisionValue, localeValue) => {
+        const callback: DecimalProps["onBlur"] = cy.stub().as("onBlur");
+
+        CypressMountWithProviders(
+          <Decimal
+            precision={precisionValue}
+            locale={localeValue}
+            allowEmptyValue
+            onBlur={callback}
+          />
+        );
+
+        commonDataElementInputPreview().type("100");
+        commonDataElementInputPreview()
+          .clear()
+          .blur({ force: true })
+          .then(() => {
+            cy.get("@onBlur").should(
+              "have.been.calledWith",
+              eventOutput("", "")
+            );
+          });
+        expect(commonDataElementInputPreview().should("have.value", ""));
+      }
+    );
+  });
+
   describe("check events for Decimal component", () => {
     const inputValue = "123";
     const iterable = [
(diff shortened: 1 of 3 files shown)


cypress/components/decimal/decimal.cy.tsx
```diff
@@ -268,6 +268,72 @@ context("Tests for Decimal component", () => {
...
+      [1, "es-ES", ""],
+      [2, "es-ES", ""],
+      [0, "fr", ""],
+      [1, "fr", ""],
+      [2, "fr", ""],
+      [0, "pt-PT", ""],
+      [1, "pt-PT", ""],
+      [2, "pt-PT", ""],
+    ] as [DecimalProps["precision"], string, string][])(
+      "should format an empty value correctly when precision is %s, locale is %s and allowEmptyValue is true",
```
DlgSHi, line 318: we don't need here a third parameter `""` because we can simply pass the same down below for the assertion and avoid copy/paste in all examples

src/components/decimal/decimal.component.tsx
```diff
@@ -179,6 +158,30 @@ export const Decimal = React.forwardRef(
       [getSeparator, isNaN, l, locale, precision]
     );
 
+    const emptyValueWithPrecision =
+      precision === 0 ? "0" : formatValue(`0.${"0".repeat(precision)}`);
```
robinzigmond, line 162: **question** (non-blocking) - is there a good reason to build up the precision value manually like this, instead of just using `(0).toFixed(precision)`? (From trying it in the console you unfortunately need those ugly parentheses around the 0, otherwise JS thinks the `.` is a decimal point rather than for object property access...)

It works correctly for precision 0 too so you wouldn't even need the ternary. And you probably don't even need the separate const as it's only used once and the expression will become simple enough.

src/components/decimal/decimal.component.tsx
```diff
@@ -179,6 +158,30 @@ export const Decimal = React.forwardRef(
...
+        // We're intentionally preventing the use of number values to help prevent any unintentional rounding issues
+        invariant(
+          typeof initialValue === "string",
+          "Decimal `value` prop must be a string"
+        );
+
+        if (initialValue && !allowEmptyValue) {
+          invariant(
+            initialValue !== "",
+            "Decimal `value` must not be an empty string. Please use `allowEmptyValue` or `0.00`"
```
robinzigmond, line 177: **question**: not exactly a suggestion as I'm not sure what would be better, but I'm not sure this message is accurate any more - the part about `0.00` I mean.

I think the message is confusing anyway, unless I'm missing something (pretty likely!) I think the best message is just something like "Please either set `allowEmptyValue` or specify a non-empty `initialValue`.".

cypress/components/decimal/decimal.cy.tsx
DipperTheDan: Good point. I copied the format of the set above. Two parameters is sufficient for this and just assert the value is `""` in the actual assertion.

cypress/components/decimal/decimal.cy.tsx
```diff
@@ -268,6 +268,72 @@ context("Tests for Decimal component", () => {
...
+
+        commonDataElementInputPreview().type("100");
+        commonDataElementInputPreview().clear().blur({ force: true });
+        expect(
+          commonDataElementInputPreview().should("have.value", expectedValue)
+        );
+      }
+    );
+
+    it.each([
```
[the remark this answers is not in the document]
DipperTheDan: We have unit tests that covered this kind of scenario but no Cypress tests. I was surprised too 😄 
I'd say we do need them just to guard against any issue that might occur in the future. Decimal is pretty important to accounting so I feel it's better we over test this than not test it. Locales can do funny things to values too due to the different separators and delimiters.

src/components/decimal/decimal.component.tsx
DipperTheDan: I seem to remember trying `.toFixed()` back in 2020 when I refactored Decimal and I seem to remember it rounded up values when we didn't want it to. Might have remember incorrectly though... 

In this case, I don't think accidentally rounding up would be an issue as it's just `0`'s. I'll give it a try and see if the tests pass 😃

src/components/decimal/decimal.component.tsx
DipperTheDan: I think you make a good point. This invariant message comes from when Decimal was first written about 4 years ago now. [See it here](https://github.com/Sage/carbon/blame/5bf51f7a32490113141308b2e02028402fd3fbcc/src/__experimental__/components/decimal/decimal.component.js#L193). We can probably change it to your suggestion as that makes more sense considering the change on this PR.

src/components/decimal/decimal.component.tsx
```diff
@@ -179,6 +158,29 @@ export const Decimal = React.forwardRef(
       [getSeparator, isNaN, l, locale, precision]
     );
 
+    const emptyValue = allowEmptyValue
+      ? ""
+      : formatValue(`${(0).toFixed(precision)}`);
```
robinzigmond, line 163:
```suggestion
      : formatValue((0).toFixed(precision));
```

`toFixed` already returns a string so there's absolutely no need to wrap it in backticks and `${...}`, which for a single value does nothing other than convert it to a string 😉

DipperTheDan: Awh I hadn't realised that. I'll get them removed then 👍🏻

robinzigmond (dismissed):

tomdavies73 (approved):

DlgSHi (approved):

robinzigmond (approved):

tomdavies73 (approved):

DlgSHi (approved):

PR #6170, opened 2023-06-30
chore(eslint): update config so rules are applied to the correct files (FE-5758)

Parsium:
### Proposed behaviour

- Update eslint config by creating new overrides, so that:
   - Jest linting rules are only applied to `*.spec.*` files
   - Cypress linting rules are only applied to files within the `cypress/` directory
- Refactor any jest and cypress test code that threw lint errors due to these new config changes[^1]
- Remove `eslint-disable` comments in `ButtonToggle`'s cypress test

[^1]: Note these errors were not being flagged before since their corresponding rules were being overwritten by our config.

### Current behaviour

- Jest linting rules are being applied to cypress test files, which is incorrect[^2]
- Some recommended rules from both the jest and cypress eslint plugins are being overwritten, so these lint errors remain unflagged.

[^2]: Cypress uses Mocha internally as its test runner, so the cypress tests do not use jest in any way, which is why the jest lint rules should not be applied to cypress files

### Checklist

#### QA

### Testing instructions

Verify that:
1. Any eslint rules that would conflict with prettier are not loaded
2. The correct rules are being applied to jest test files, cypress test files, and other remaining files

A shell script has been created for convenience when testing these two criteria. Please see these notes on how to use it this script:
[Testing eslint config for FE-5758.pdf](https://github.com/Sage/carbon/files/11949567/Testing.eslint.config.for.FE-5758.pdf)


.eslintignore
edleeks87, line 1: 3 years after the last of these files were removed 😆

src/components/date/date.spec.js
```diff
@@ -91,11 +91,14 @@ describe("Date", () => {
   let wrapper;
   let onFocusFn;
 
-  // eslint-disable-next-line react/prop-types
   const MockComponent = ({
+    // eslint-disable-next-line react/prop-types
```
edleeks87, line 95: question (non-blocking): Likewise for spec.js files, maybe we should just turn the rule off altogether?

cypress/components/flat-table/flat-table.cy.js
```diff
@@ -2749,7 +2749,9 @@ const FlatTablePagerStickyHeaderComponent = () => {
 };
 
 const FlatTablePartiallySelectedOrHighlightedRows = ({
+  // eslint-disable-next-line react/prop-types
```
edleeks87, line 2752: question (non-blocking): Would it be better to just ignore this rule all together in any cypress (JS) files

src/components/flat-table/flat-table.spec.js
```diff
@@ -1535,206 +1535,6 @@ describe("FlatTable", () => {
             .at(0)
         ).toBeFocused();
       });
-
```
edleeks87, line 1538: comment: I'm a bit surprised this isn't causing missed coverage

src/components/date/date.spec.js
Parsium: I agree, if we plan to release this work once all our components are migrated to TypeScript, there will be little point of keeping this rule.

Parsium:
Given our immediate plans to move to Playwright, I am closing this ticket as it is no longer required. However, I will keep the branch open as the updates to how our linting rules are organised would still be applicable after moving to Playwright.

PR #6171, opened 2023-07-03
fix(menu-fullscreen): fix focus when menu closed

### LINKED ISSUE #6037: MenuFullscreen traps focus when closed
### Current behaviour

When `Fullscreenmenu` is in the closed state it appears to be trapping focus as if it were open.

The provided codesandbox has a skiplink, `FullScreenmenu` and a number of interactive elements. When triggering the skiplink, it's not possible to navigate via keyboard to the other interactive elements.

### Expected behaviour

To be able to navigate via keyboard past `FullScreenmenu` when closed.

https://codesandbox.io/s/fullscreen-menu-focus-nrk45s

### JIRA Ticket (Sage Only)

SBS-45425

### Suggested Solution

_No response_

### Carbon Version

118.3.2

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
MenuFullScreen renders a FocusTrap with DOM children even when the menu is closed, resulting in some unexpected behaviour with the tab key. To fix this, all FocusTrap behaviour has been  turned off when the surrounding component is closed.

fix #6037

### Checklist

#### QA

### Testing instructions


robinzigmond:
> **issue:** Looks like you've fixed the bug initially raised but seems that it is still possible for tabbing to be lost by tabbing to the end of the menu and back up to the close icon, close the menu and focus seems to go elsewhere on the page. Attached a GIF. This is observable in the CSB built from the original issue. menu-full-screen-bug

Are you sure it's lost completely? I think focus probably goes off the bottom of the page, and can be "got back" by back tabbing - or forward tabbing and reappearing at the top of the browser. Not ideal, I admit, but I don't think it can be avoided easily - and it comes from the fact that the menu is technically at the end of the DOM. The only way around this would be to focus the open button when the menu is closed. Maybe that's not a bad idea actually but it probably needs UX input.

cypress/components/menu/menu.cy.js
```diff
@@ -274,6 +274,23 @@ const MenuFullScreenBackgroundScrollTest = () => {
   );
 };
 
+const ClosedMenuFullScreenWithButtons = () => {
```
tomdavies73, line 277: **suggestion** (non-blocking): could you import `ClosedVerticalMenuFullScreenWithButtons` from `vertical-menu-test.stories` instead of creating it again here?

robinzigmond: totally meant to do that - not sure how I ended up defining it twice and not even using one of them! 🤦 Cheers.

robinzigmond: Oh hang on, I've only just realised that these are 2 separate components - MenuFullScreen and VerticalMenuFullScreen. So I'm not going to change this, even though the 2 stories are effectively testing the same thing - because they're in 2 separate components.

It is inconsistent that one is defined in a test-stories file and the other directly in the Cypress file, I agree - to be honest I was just going with what the 2 existing Cypress files were doing, and that for (non-vertical) Menu already defines a bunch of 'stories" inside the cypress test file. That probably does want changing but I'd rather leave it to a wider piece of tidying-up than introduce one new story in the test-stories file in a component where we otherwise aren't doing it this way. If QA want this done now they can say so when the ticket reaches them! 😄

robinzigmond:
> #6037

I've now checked on an older version from before the bug appeared - https://carbon.sage.com/v/106.6.10/index.html?path=/story/menu--fullscreen-view - and confirmed the tabbing behaviour once the menu has been opened and then shut is exactly as it is as a result of this PR. So while it's worth speaking to UX to see if this is something we can improve, I don't think it needs to (or should) be done in the same PR.

tomdavies73 (dismissed):

edleeks87 (dismissed):

ZhuoyuJin (approved):

edleeks87 (approved):

tomdavies73 (approved):

ZhuoyuJin (approved):

PR #6177, opened 2023-07-06
feat(radio-button): add support for new validation - FE-5753

DipperTheDan:
### Proposed behaviour

Screenshot 2023-07-06 at 12 13 43

### Current behaviour

New style validations are not supported in `Radio Button` and `Radio Button Group`.

### Checklist

#### QA

### Additional context

Validation styles for `large` size will be completed in another ticket due to the way in which `size` is specified in this component.

### Testing instructions

- New validations should match the designs created by DS.
- Old validations should continue to work as normal.
- No other regressions with functionality or styles.

### DIFF
--- a/src/components/checkbox/checkbox.component.tsx
+++ b/src/components/checkbox/checkbox.component.tsx
@@ -11,6 +11,7 @@ import { TooltipProvider } from "../../__internal__/tooltip-provider";
 import { CheckboxGroupContext } from "./checkbox-group.component";
 import Logger from "../../__internal__/utils/logger";
 import useFormSpacing from "../../hooks/__internal__/useFormSpacing";
+import { NewValidationContext } from "../carbon-provider/carbon-provider.component";
 
 export interface CheckboxProps extends CommonCheckableInputProps, MarginProps {
   /** Breakpoint for adaptive spacing (left margin changes to 0). Enables the adaptive behaviour when set */
@@ -81,6 +82,8 @@ export const Checkbox = React.forwardRef(
     }: CheckboxProps,
     ref: React.ForwardedRef<HTMLInputElement>
   ) => {
+    const { validationRedesignOptIn } = useContext(NewValidationContext);
+
     const largeScreen = useIsAboveBreakpoint(adaptiveSpacingBreakpoint);
     const adaptiveSpacingSmallScreen = !!(
       adaptiveSpacingBreakpoint && !largeScreen
@@ -158,6 +161,7 @@ export const Checkbox = React.forwardRef(
           fieldHelpInline={fieldHelpInline}
           reverse={reverse}
           size={size}
+          applyNewValidation={validationRedesignOptIn}
           {...marginProps}
         >
           <CheckableInput {...inputProps}>
--- a/src/components/radio-button/radio-button-group.component.tsx
+++ b/src/components/radio-button/radio-button-group.component.tsx
@@ -1,14 +1,20 @@
-import React from "react";
+import React, { useContext } from "react";
 import { MarginProps } from "styled-system";
 import tagComponent from "../../__internal__/utils/helpers/tags/tags";
 import Fieldset from "../../__internal__/fieldset";
-import RadioButtonGroupStyle from "./radio-button-group.style";
+import RadioButtonGroupStyle, {
+  StyledHintText,
+} from "./radio-button-group.style";
 import RadioButtonMapper from "../../__internal__/radio-button-mapper/radio-button-mapper.component";
 import useIsAboveBreakpoint from "../../hooks/__internal__/useIsAboveBreakpoint";
 import { filterStyledSystemMarginProps } from "../../style/utils";
 import { TooltipProvider } from "../../__internal__/tooltip-provider";
 import { ValidationProps } from "../../__internal__/validations";
 import Logger from "../../__internal__/utils/logger";
+import { NewValidationContext } from "../carbon-provider/carbon-provider.component";
+import ValidationMessage from "../../__internal__/validation-message/validation-message.component";
+import Box from "../../components/box";
+import { ErrorBorder } from "../../components/textbox/textbox.style";
 
 let deprecateUncontrolledWarnTriggered = false;
 export interface RadioButtonGroupProps extends ValidationProps, MarginProps {
@@ -24,6 +30,8 @@ export interface RadioButtonGroupProps extends ValidationProps, MarginProps {
   labelSpacing?: 1 | 2;
   /** The content for the RadioGroup Legend */
   legend?: string;
+  /** The content for the hint text of the RadioGroup Legend */
+  legendHelp?: string;
   /** Text alignment of legend when inline */
   legendAlign?: "left" | "right";
   /** When true, legend is placed in line with the radiobuttons */
@@ -51,6 +59,7 @@ export const RadioButtonGroup = (props: RadioButtonGroupProps) => {
     children,
     name,
     legend,
+    legendHelp,
     error,
     warning,
     info,
@@ -68,6 +77,8 @@ export const RadioButtonGroup = (props: RadioButtonGroupProps) => {
     required,
     tooltipPosition,
   } = props;
+
+  const { validationRedesignOptIn } = useContext(NewValidationContext);
 
   if (!deprecateUncontrolledWarnTriggered && !onChange) {
     deprecateUncontrolledWarnTriggered = true;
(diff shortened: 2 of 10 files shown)


src/components/radio-button/radio-button-group.style.ts
```diff
@@ -1,5 +1,27 @@
...
+  ${({ warning }: { warning: boolean }) =>
+    css`
+      position: absolute;
+      z-index: 6;
+      width: 2px;
+      background-color: ${warning
+        ? "var(--colorsSemanticCaution500)"
+        : "var(--colorsSemanticNegative500)"};
+      left: -12px;
+      bottom: -2px;
```
edleeks87, line 13: Question: Do we need to bottom and top to not be 0px here? I ask as we already have these defined in `textbox.style` maybe they can be used here or at least extended and the necessary styles overridden.

```
const CheckableErrorBorder = styled(ErrorBorder)`
bottom: -2px;
top: 2px;
`
```

src/components/radio-button/radio-button.component.tsx
```diff
@@ -138,32 +141,83 @@ export const RadioButton = React.forwardRef<
       ...props,
     };
 
+    const commonPropsForNewValidation = {
```
edleeks87, line 144: comment: there's a lot of duplication here as most of the props are needed whether the flag is set or not

edleeks87: Suggestion: you could remove the duplication by doing something like the below but happy for you to go with what you prefer as it's not a blocking issue for me

``` 
const validationProps = {
      disabled,
      inputWidth,
      error,
      warning,
      info,
    };

    const commonProps = {
      ...validationProps,
      fieldHelpInline,
      labelSpacing,
    };

    const inputProps = {
      ...(validationRedesignOptIn ? {...commonProps, fieldHelp} : validationProps),
      autoFocus,
      checked,
      fieldHelp,
      name,
      onChange: handleChange,
      onBlur,
      onFocus,
      labelAlign,
      labelInline: true,
      labelWidth,
      label,
      labelHelp,
      id,
      value,
      type: "radio",
      /**
       * Invert the reverse prop, to ensure the FormField component renders the components
       * in the desired order (other elements which use FormField render their sub-components the
       * opposite way around by default)
       */
      reverse: !reverse,
      required,
      ref: ref || inputRef,
      ...props,
    };

    invariant(
      !props.children,
      "This component is meant to be used as a self-closing tag. " +
        "You should probably use the label prop instead."
    );

    const componentToRender = (
      <RadioButtonStyle
            applyNewValidation={validationRedesignOptIn}
            data-component={dataComponent}
            data-role={dataRole}
            data-element={dataElement}
            inline={inline}
            reverse={reverse}
            size={size}
            {...(validationRedesignOptIn ? {...commonProps, fieldHelp} : validationProps)}
            {...marginProps}
          >
            <CheckableInput {...inputProps}>
              <RadioButtonSvg />
            </CheckableInput>
          </RadioButtonStyle>
    )

    return (
      <>
        {validationRedesignOptIn ? (
          componentToRender
        ) : (
          <TooltipProvider
            helpAriaLabel={helpAriaLabel}
            tooltipPosition={tooltipPosition}
          >
            {componentToRender}
          </TooltipProvider>
        )}
      </>
    );
    ```

File: src/components/radio-button/radio-button.stories.tsx
Comment (edleeks87) [inline src/components/radio-button/radio-button.stories.tsx:1]: Question: Could we add something that demos the validation on the buttons rather than group as well?

File: src/components/checkbox/checkbox.component.tsx [hunk 3]
```diff
@@ -37,6 +38,8 @@ export interface CheckboxProps extends CommonCheckableInputProps, MarginProps {
   value?: string;
   /** A callback to retrieve the input reference (deprecated) */
   inputRef?: React.Ref<HTMLInputElement>;
+  /** boolean to apply new validation styles to warning */
```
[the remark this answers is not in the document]
DipperTheDan: Not needed. Removed locally and everything still works as expected so I'll get that sorted.

src/components/radio-button/radio-button-group.style.ts
DipperTheDan: In the designs it looks like the error border extends slightly below the last radio button. Having bottom as 0px makes it look like the border stops too short. I'm happy to run it past UX though to see if 0px would suffice. Using the `ErrorBorder` from textbox might be a good idea even if we just need to potentially override the bottom css value.

src/components/radio-button/radio-button.component.tsx
DipperTheDan: Yeah will be neater way of doing it 👍🏻

src/components/radio-button/radio-button.stories.tsx
DipperTheDan: I can do that but It'll look like this:
Screenshot 2023-07-06 at 16 34 52

Not much visually going on but it will demonstrate the props and have a code example to go with too.

src/components/radio-button/radio-button-group.component.tsx
```diff
@@ -97,53 +108,110 @@ export const RadioButtonGroup = (props: RadioButtonGroupProps) => {
...
-        {...marginProps}
-        ml={marginLeft}
-        blockGroupBehaviour={!(error || warning || info)}
-      >
-        <RadioButtonGroupStyle
-          data-component="radio-button-group"
-          role="radiogroup"
-          inline={inline}
-          legendInline={inlineLegend}
+    <>
```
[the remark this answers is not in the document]
DipperTheDan: I had thought about that but my thought process behind this was when we come to remove the old style validations, whenever that will be, we just need to delete the `false` side of the ternary. I suppose I could go for a mix of that ease in the future whilst trimming out the duplication. I'll see what I can come up with.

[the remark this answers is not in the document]
edleeks87: Yeah that makes sense and I did wonder if that was your reasoning here so happy to leave it as is as 👍

src/components/radio-button/radio-button-group.style.ts
edleeks87: No worries, yeah I couldn't decide if that was the case based on the designs myself so maybe bounce it off UX and go from there

src/components/radio-button/radio-button.stories.tsx
edleeks87: Ah I see, yeah no worries then. Maybe we could add something in the test story for it at least?

src/components/radio-button/radio-button.stories.mdx
```diff
@@ -156,6 +156,18 @@ Validation status could also be passed to the `RadioButtonGroup` component as a
...
+  <Story name="new validation" story={stories.NewValidationDefault} />
+</Canvas>
+
+<Canvas>
+  <Story name="new validation group" story={stories.NewValidationDefaultGroup} />
+</Canvas>
+
+### With custom styled labels
+
+Labels could be customized using the Typography Component
```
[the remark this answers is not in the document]
DipperTheDan: Nope. I think I've somehow copied [this piece of docs here](https://github.com/Sage/carbon/pull/6177/files/61249736d768731635ac482f97a4b21e8daf6bd6#diff-e04cdf55119bcbb7d16f2c70637698bebfbf3645f33cb1f47fbbd8c852a33d0dR139-R141). I'll get it removed.

tomdavies73 (dismissed):

edleeks87 (approved):

DlgSHi (approved):

src/__internal__/validation-message/validation-message.component.tsx
```diff
@@ -10,12 +10,15 @@ export interface ValidationMessageProps {
   /** Indicate that warning has occurred
   Pass string to display hint with warning */
   warning?: boolean | string;
+  /** Used to set font size if large size of a component is used.  */
+  sizeIsLarge?: boolean;
```
tomdavies73, line 14: **question:** (non-blocking):After checking the DS and the corresponding figma, I can see there is a need for a larger text variant, which I personally don't see the need for especially with such a minor `px` difference. But I agree, we should stick to the figma and add what's there. Instead of a`boolean`, could we perhaps use a prop like `fontSize` with a type of `"medium"` | `"large"`, with `"medium"` being the default?

src/components/radio-button/radio-button-group.component.tsx
```diff
@@ -52,6 +52,8 @@ export interface RadioButtonGroupProps extends ValidationProps, MarginProps {
   value?: string;
   /** Overrides the default tooltip position */
   tooltipPosition?: "top" | "bottom" | "left" | "right";
+  /** Set validation styling to large */
+  largeStyleValidation?: boolean;
```
tomdavies73, line 56: **suggestion**(non-blocking): Again, similar sort of comment here, if we do have to go with this approach. Could we use some sort of type approach with size types, instead of boolean. Something like `validationSize` etc. with `"medium"` | `"large"` as types? This would also be future-proof, as if the DS add a small variant in the future, or an even larger variant we'd have scalability to add more types without introducing a breaking change by moving away from the boolean approach, or adding more props 👍

edleeks87 (approved):

DlgSHi (approved):

src/__internal__/validation-message/validation-message.component.tsx
DipperTheDan: Decided that we're doing this in a separate piece of work so I'll resolve this.

src/components/radio-button/radio-button-group.component.tsx
DipperTheDan: Decided that we're doing this in a separate piece of work so I'll resolve this.

tomdavies73 (approved):

PR #6185, opened 2023-07-11
fix(decimal): replace row-reverse with row when text is right aligned

### LINKED ISSUE #6109: `prefix` appears on the right of Decimal and Number
### Current behaviour

Set the `prefix` prop on Decimal to some value. The text appears on the far right on the input rather than at the start.

### Expected behaviour

The prefix to appear on the left.

https://codesandbox.io/s/patient-shape-szmsxk?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

116.2.0

### Design Tokens Version

_No response_

Firefox, Chrome

Windows

_No response_

### Confidentiality

tomdavies73:
fix #6109

### Proposed behaviour

When the `align` prop is "right" on components which render `Textbox` (Decimal is set to "right" by default), then the flex-direction will change from `row-reverse` to `row`. 

Screenshot 2023-07-11 at 12 15 03

### Current behaviour

When the `align` prop is "right" on components which render `Textbox` (Decimal is set to "right" by default), then the flex-direction is `row-reverse`. This means the main-start and main-end points are opposite to the content direction, causing the `prefix` to be placed behind the text.

Screenshot 2023-07-11 at 12 15 19

### Checklist

#### QA

### Testing instructions


src/__internal__/input/input-presentation.style.ts
```diff
@@ -114,7 +115,10 @@ const InputPresentationStyle = styled.div<
       border-color: var(--colorsUtilityReadOnly600);
     `}
 
-  ${({ align }) => align === "right" && "flex-direction: row-reverse"}
+  ${({ align, prefix }) =>
+    align === "right" &&
+    `flex-direction: ${prefix ? "row" : "row-reverse"};
+  `}
```
Parsium, line 121: **suggestion(non-blocking):** What do you think about simplifying this like so?
```suggestion

```
As the `InputPresentationStyle` is the container of the `<input>` and prefix, `flex-direction="row-reverse"` doesn't control the `<input>`'s text alignment. 

This way we can decouple the CSS of the `prefix` and `align` props, which will help with maintenance.

tomdavies73: Hi mate, agreed 👍, was considering just removing the whole thing entirely. After doing a bit of digging, the original code was introduced in [this PR](https://github.com/Sage/carbon/commit/0015029c46d75f68d60884019d59e5d01afd2ce4) as part of input validations, so it may serve some purpose. If we remove and no tests fail, and no chromatic regressions are flagged, I'm happy to do so. Will wait and see what a second reviewer thinks as well :)

cypress/components/decimal/decimal.cy.tsx
```diff
@@ -238,6 +239,27 @@ context("Tests for Decimal component", () => {
...
+
+        textboxPrefix()
+          .should("have.text", prefix)
+          .and("have.css", "font-size", "14px")
+          .and("have.css", "font-weight", "900")
+          .and("have.css", "margin-left", "12px");
+      }
+    );
+
+    it("when prefix prop is set, 'flex-direction' should be 'row'", () => {
```
nineteen88, line 255: nitpick(non-blocking): tests should read in plain English ideally

This is something that is common across the project, so I wouldn't worry about changing it but something to keep in mind potentially. 

```suggestion
    it("should set `flex-direction` to `row` when a prefix prop is used", () => {
```

nineteen88 (approved):

divyajindel (approved):

PR #6196, opened 2023-07-13
test(password): refactor to ts

DlgSHi:
### Proposed behaviour

- Rename the `password.cy.js` to the `password.cy.tsx` file in the `./cypress/components/password/` folder.

### Current behaviour

`Password ` tests are in js not ts.

### Checklist

#### QA

### Testing instructions
- [x] Run `npx cypress open --component` to check if the `password.cy.tsx` file passed
- [x] Run `npx cypress run --component` to check none of the other `*.cy.tsx` files have regressed

### DIFF
--- a/cypress/components/password/password.cy.tsx
+++ b/cypress/components/password/password.cy.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import Password from "../../../src/components/password/password.component";
+import { PasswordProps } from "../../../src/components/password/password.component";
+import { PasswordComponent } from "../../../src/components/password/password-test.stories";
 import * as stories from "../../../src/components/password/password.stories";
 import CypressMountWithProviders from "../../support/component-helper/cypress-mount";
 import {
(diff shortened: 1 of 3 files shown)


stephenogorman (approved):

cypress/components/password/password.cy.tsx
```diff
@@ -208,27 +172,22 @@ context("Tests for Password component", () => {
...
-        .getDesignTokensByCssProperty("color")
-        .should(($el) => {
-          expect($el[3]).to.equal("--colorsUtilityMajor300");
-        });
+      icon().getDesignTokensByCssProperty("color").as("colorToken");
+
+      cy.get("@colorToken").its("3").should("equal", "--colorsUtilityMajor300");
     });
 
     it("icon color is 'colorsActionMajorYang300' onHover'", () => {
```
tomdavies73, line 180: **suggestion:** (non-blocking): Think this was actually my fault originally, but may be worth renaming this test during this refactor. As test description references `colorsActionMajorYang300`, but we're actually testing against `colorsUtilityMajor300`

divyajindel (approved):

src/components/password/password-test.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+    },
+    labelWidth: {
+      control: {
+        min: 0,
+        max: 100,
+        step: 1,
+        type: "range",
+      },
+    },
+    children: {
```
tomdavies73, line 41: **suggestion:** I think using `children` here is redundant, instead we can replace it with `value`, which will actually change the value of the input. Something like:  `value: {
      control: {
        type: "text",
      },
    },`

src/components/password/password-test.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+export default {
+  title: "Password/Test",
+  includeStories: ["Default"],
+  parameters: {
+    info: { disable: true },
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
+  argTypes: {
```
tomdavies73, line 13: **suggestion:** I agree we shouldn't surface every potential prop as an argType, as consumers will probably only use a few, and I think your used props are pretty much spot on. The only other ones I'd suggest adding are `inputHint` and maybe `prefix` at a push. Imo `inputHint` should definitely be added as it's a key part of the `Password component`

tomdavies73 (approved):

divyajindel (approved):

PR #6201, opened 2023-07-14
feat(option-group-header, option-row, option): allow custom IDs and surface data selectors

### LINKED ISSUE #5810: Carbon Select list of options is not related to the Select component
### Desired behaviour

The list of options associated with a Carbon Select should be located within the Carbon Select div (<div data-element="select-input">) and ultimately the select-input div should include a data role attribute that can be set by the developer. 

### Current behaviour

The list of options for a Carbon Select is located in a completely separate area of the DOM to the div that contains the Select UI elements. 
This means automation/unit test code, on opening the select, has no way to click the desired option other than to search for an li element with matching text. This is a very unreliable solution and runs into issues where there are two or more Selects that contain options with the same text content. I have encountered this situation so this is not just a hypothetical concern. 
Developers sometimes encapsulate the Select div within an outer div that contains a data role to facilitate automation but this advantage is lost when interacting with the drop down options because of the situation described above. 

### Suggested Solution

The list of options associated with a Carbon Select should be located within the Carbon Select div and ultimately the select-input div should include a data role attribute that can be set by the developer. 

_No response_

I have also tried to select an option by inserting the relevant text into the hidden input associated with the Select but if the options contain two items that begin with the same character then that will not be reliable since it only seems to search out the option that matches the first character of the input string. 

### Confidentiality

### LINKED ISSUE #6126: ID Props applied to Select Options do not carry through, leading to dynamic IDs
### Current behaviour

If I add an ID as a prop to the children of a Select (Option element) That ID is then not present in the devTools. This means certain methods of finding the elements that our (Delta Squad) QA use cannot locate the elements.

### Expected behaviour

Id's supplied as props to the Options of a parent Select element are applied to the component.

https://codesandbox.io/s/distracted-mahavira-k23nr3?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

npm:carbon-react@106.6.8

### Design Tokens Version

_No response_

Firefox, Chrome, Safari, Microsoft Edge, Other

Windows

_No response_

### Confidentiality

tomdavies73:
fix #6126, fix #5810

### Proposed behaviour

A condition has been added to `SelectList` which checks if the `id` prop is present on any children, if it is the GUID is not applied. This provides consumers with flexibility to pass specific IDs to `OptionGroupHeader`, `OptionRow`, and `Option` component children. Also all interfaces in `Option`, `OptionRow` and `OptionGroupHeader` now extend `TagProps`, giving consumers access to the `data-role` and `data-element` attributes for testing purposes

### Current behaviour

Currently, a GUID is applied to every potential child of the `SelectList` component `(OptionGroupHeader, OptionRow, Option)`. Even when consumers pass the `id` prop, the id attribute is still overwritten with a GUID.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/cypress/components/select/filterable-select/filterable-select.cy.tsx
+++ b/cypress/components/select/filterable-select/filterable-select.cy.tsx
@@ -46,6 +46,7 @@ const addElementText = "Add a New Element";
 const columns = 3;
 const icon = "add";
 const keyToTrigger = ["downarrow", "uparrow", "Home", "End", "Enter"] as const;
+const listOption = "Amber";
 
 context("Tests for FilterableSelect component", () => {
   describe("check props for FilterableSelect component", () => {
@@ -95,7 +96,7 @@ context("Tests for FilterableSelect component", () => {
       );
     });
 
-    it("should render FilterableSelect with id prop prop set to test value", () => {
+    it("should render FilterableSelect with id prop set to test value", () => {
       CypressMountWithProviders(
         <stories.FilterableSelectComponent id={testPropValue} />
       );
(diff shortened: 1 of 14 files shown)


src/components/select/select-list/select-list.component.tsx
```diff
@@ -196,7 +196,11 @@ const SelectList = React.forwardRef(
     // as that isn't absolutely guaranteed to never rerun when dependencies haven't changed.
     const setChildIds = () => {
       childIdsRef.current =
-        React.Children.map(children, () => guid()) ||
+        React.Children.map(
+          children,
+          (child) => (React.isValidElement(child) && child?.props.id) || guid()
+        ) ||
+        // React.Children.map(children, () => guid()) ||
```
DipperTheDan, line 203: **issue:** Left some commented out code in.

Parsium (commented):
Happy to approve once the comment from @DipperTheDan has been addressed 👍🏼

Parsium (approved):

DipperTheDan (approved):

cypress/components/select/filterable-select/filterable-select.cy.tsx
```diff
@@ -824,6 +824,84 @@ context("Tests for FilterableSelect component", () => {
...
+      dropdownButton().click();
+      multiColumnsSelectListBody()
+        .parent()
+        .should("have.attr", "data-element", "option-row");
+    });
+
+    it("should contain custom option id option1", () => {
+      CypressMountWithProviders(<stories.FilterableSelectComponent />);
+
+      const option = "Amber";
```
DlgSHi, line 873: can we move this to a consts at the top of the file? and then use it below

sianford (dismissed):
Tested in codesandbox and Storybook. No issues found.

cypress/components/select/multi-select/multi-select.cy.tsx
```diff
@@ -965,6 +965,76 @@ context("Tests for MultiSelect component", () => {
       selectList().should("not.be.visible");
       commonDataElementInputPreview().should("not.be.focused");
     });
+
+    it("should contain custom option id option1", () => {
+      CypressMountWithProviders(<stories.MultiSelectComponent />);
+
+      const option = "Amber";
```
DlgSHi, line 972: same here

stephenogorman (changes requested):

DlgSHi (approved):

DipperTheDan (approved):

PR #6203, opened 2023-07-17
fix(switch): ensure invariant is not thrown when loading and validation is set

tomdavies73:
### Proposed behaviour

A context approach has been used to pass down the value of `loading`, so when `true`, the invariant within `FormField` will not be thrown when validations are also set. An invariant has also been added to `Switch` to stop consumers passing `loading`, and `disabled` at the same time on the component. As when `loading` is true, `disabled` is also set to `true`. This also makes firing the invariant within `FormField`, easier and less complex. 

### Current behaviour

Currently, an approach in `Switch` sets `disabled` to true whenever `disabled` or `loading` is true: `disabled: disabled || loading,`. This is necessary, as the component should be `disabled` when `loading` is true, to stop users interacting with the component during a loading state. However, this also means when `loading` is true an invariant is triggered as `FormField` recognises `disabled` as being true.

### Checklist

#### QA

### Testing instructions


DipperTheDan (dismissed):

src/components/switch/switch.context.ts
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+type SwitchContextType = {
+  loading?: boolean;
+};
+
+const SwitchContext = React.createContext<SwitchContextType | null>(null);
```
robinzigmond, line 7:
```suggestion
const SwitchContext = React.createContext<SwitchContextType>({});
```

I don't see any reason to compicate matters by bringing `null` into it. The context type is an object with an (optional) boolean inside it, so you can make the default just an object that doesn't have it. (You could also do this with an explicit `{ loading: false }`, which would also allow you to make the type non-optional - but I don't care much whether you do that or not.)

src/__internal__/form-field/form-field.spec.tsx
```diff
@@ -242,4 +254,22 @@ describe("FormField", () => {
       consoleSpy.mockRestore();
     }
   );
+
+  it.each(["error", "warning", "info"])(
```
robinzigmond, line 258: **comment**: I don't think there's any need for this test to be here in `FormField`. It's testing what `Switch` does or doesn't do with particular prop combinations - that should be in the `Switch` tests (and I think you do already have this there 👍 ).

There's nothing wrong with testing how `FormField` works with your new context (although I don't think it's really needed), but in that case I think rather than rendering `Switch` explicitly we should render `FormField` inside just the context provider. But as I said I don't think this is necessary here.

robinzigmond (commented):
**comment**: I'm not saying this to ask you to redo the whole PR, especially as I can't give a super-clear justification for my preference - but I personally don't feel that Context is the right tool here.

I think it should be possible to solve this without Context, by adding `loading` as a prop to FormField (it's an internal component anyway so we shouldn't have to worry about whether the new prop is public or not). Then we can can access it inside `FormField` directly as well as `disabled`, removing the need to use both to decide whether to pass `disabled` to `FormField` and losing the knowledge of which one was actually set by the consumer. If necessary we can use an internal `const isLoadingOrDisabled = disabled || loading` inside `FormField` for what we currently use `disabled` for (which, other than the `invariant` stuff that we're changing here anyway, looks like it's just a prop passed to `Label`).

Would be interested to hear your thoughts, I think this change is probably about the same in terms of amount of code but might be easier to understand as well as potentially a bit more performant (context is never going to be great for performance because whenever a context value updates every single child of the Provider is rerendered).

But I don't hate the context approach either, and my specific comments on the code are obviously under the assumption that we do take this forward with your context approach.

src/components/switch/switch.component.tsx
```diff
@@ -154,6 +154,7 @@ export const Switch = React.forwardRef(
       warning,
       info,
       disabled: disabled || loading,
```
robinzigmond, line 156:
```suggestion
      disabled,
```

The whole point of introducing the `loading` prop to `FormField` was to not have to pretend the component is `disabled` when `loading` is set

src/components/switch/switch.component.tsx
```diff
@@ -199,6 +200,7 @@ export const Switch = React.forwardRef(
       error,
       warning,
       disabled: disabled || loading,
```
robinzigmond, line 202:
```suggestion
      disabled,
```

as above

robinzigmond (approved):

ZhuoyuJin (approved):

PR #6204, opened 2023-07-18
fix(split-button, multi-action-button): remove children iteration

### LINKED ISSUE #4784: Unable to select options from SplitButton with keyboard if children are link buttons
### Current behaviour

On Sage 200 some of our SplitButtons contain children which are links. For this we make carbon buttons into links following the example given in the [usage with routing](https://carbon.sage.com/?path=/docs/documentation-usage-with-routing--page) documentation.

When a user tries to use the keyboard to select an option from the split button, they are unable to. In the console we see `Uncaught TypeError: additionalButtons.current[nextIndex].focus is not a function` from within the Carbon SplitButton component.

Impacts accessibility and slows down our power users who like to use the keyboard.

### Expected behaviour

User should be able to select options from the dropdown with their keyboard as when the split button has only buttons as children.

https://codesandbox.io/s/bold-sunset-ztpoc?file=/src/index.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

104.0.0

Firefox, Chrome

Windows

Also applies to the MultiActionButton

### Confidentiality

robinzigmond:
SplitButton and MultiActionButton used to clone their children and pass particular props, including refs - assuming each child was a Button component. This meant that the components failed to work properly when using custom child components that wrap one or more Carbon Buttons (at least unless forwardRef was used). This has now been replaced by a context-based approach using DOM selectors to achieve the same functionality no matter how the children are composed.

fix #4784

### Proposed behaviour

All functionality in SplitButton and MultiActionButton should work when consumers wrap Carbon Button in a custom child component

### Current behaviour

There are warnings in the console when using such a wrapper, and keyboard navigation fails to work

### Checklist

#### QA

### Testing instructions

Ensure that the original issue is fixed, and perform regression testing on the SplitButton and MultiActionButton components.

### DIFF
--- a/src/components/button/button.component.tsx
+++ b/src/components/button/button.component.tsx
@@ -9,6 +9,7 @@ import { TooltipProvider } from "../../__internal__/tooltip-provider";
 import Logger from "../../__internal__/utils/logger";
 import { TooltipPositions } from "../tooltip/tooltip.config";
 import { ButtonBarContext } from "../button-bar/button-bar.component";
+import SplitButtonContext from "../split-button/__internal__/split-button.context";
 
 export type ButtonTypes =
   | "primary"
@@ -194,6 +195,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       iconTooltipMessage,
       iconTooltipPosition,
       fullWidth: fullWidthProp = false,
+      onClick,
       ...rest
     }: ButtonProps,
     ref
@@ -236,6 +238,10 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
     }
 
     const [internalRef, setInternalRef] = useState<HTMLButtonElement>();
+
+    const { inSplitButton, onChildButtonClick } = useContext(
+      SplitButtonContext
+    );
 
     let paddingX;
 
@@ -280,11 +286,12 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         }
         as={!disabled && href ? "a" : "button"}
         onKeyDown={href ? handleLinkKeyDown : undefined}
+        onClick={inSplitButton ? onChildButtonClick?.(onClick) : onClick}
         draggable={false}
         buttonType={buttonType}
         disabled={disabled}
         destructive={destructive}
-        role="button"
+        role={inSplitButton ? "menu-item" : "button"}
         type={href ? undefined : "button"}
         iconType={iconType}
         size={size}
--- a/src/components/multi-action-button/multi-action-button.component.tsx
+++ b/src/components/multi-action-button/multi-action-button.component.tsx
@@ -1,20 +1,20 @@
-import React, { useCallback, useState, useRef, useMemo } from "react";
+import React, { useRef } from "react";
 import { WidthProps } from "styled-system";
 
 import useClickAwayListener from "../../hooks/__internal__/useClickAwayListener";
 import { SplitButtonProps } from "../split-button";
+import SplitButtonContext from "../split-button/__internal__/split-button.context";
 import {
   StyledMultiActionButton,
   StyledButtonChildrenContainer,
 } from "./multi-action-button.style";
 import Button from "../button";
-import Events from "../../__internal__/utils/helpers/events";
 import Popover from "../../__internal__/popover";
 import {
   filterStyledSystemMarginProps,
   filterOutStyledSystemSpacingProps,
 } from "../../style/utils";
-import useMenuKeyboardNavigation from "../../hooks/__internal__/useMenuKeyboardNavigation";
+import useChildButtons from "../../hooks/__internal__/useChildButtons";
 
 export interface MultiActionButtonProps
   extends WidthProps,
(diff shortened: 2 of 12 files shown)


src/components/button/button.component.tsx
```diff
@@ -9,6 +9,7 @@ import { TooltipProvider } from "../../__internal__/tooltip-provider";
 import Logger from "../../__internal__/utils/logger";
 import { TooltipPositions } from "../tooltip/tooltip.config";
 import { ButtonBarContext } from "../button-bar/button-bar.component";
+import splitButtonContext from "../split-button/split-button.context";
```
edleeks87, line 12: **nitpick**: `splitButtonContext` -> `SplitButtonContext`

src/components/split-button/split-button.context.ts
edleeks87, line 1: **comment**: I realise we're pretty inconsistent about this (as usual 🤦 ) but I think we should make this an internal file as we don't ever intend consumers to use it etc

src/components/button/button.component.tsx
```diff
@@ -237,6 +239,10 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
 
     const [internalRef, setInternalRef] = useState<HTMLButtonElement>();
 
+    const { inSplitButton, onChildButtonClick } = useContext(
```
edleeks87, line 242: **comment**(non-blocking): I wonder if we need `inSplitButton` here at all. As `onChildButtonClick` is optional we could derive `inSplitButton` from it like so `const inSplitButton = !!onChildButtonClick;`.

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -41,17 +42,21 @@ export const MultiActionButton = ({
...
-  const buttonContainer = useRef<HTMLDivElement>(null);
-  const buttonChildren = useMemo(() => React.Children.toArray(children), [
-    children,
-  ]);
-  const buttonChildrenRefs = useMemo(
-    () => buttonChildren.map(() => React.createRef<HTMLButtonElement>()),
-    [buttonChildren]
+  const childrenContainer = useRef<HTMLDivElement>(null);
+
+  const getButtonChildren = useCallback(
```
edleeks87, line 47: **question**: as we're doing the same thing in both Multi and Split components, I wonder if it would be worth moving this into a custom hook or a provider component? This is non-blocking for me as it may be more complicated than to integrate it with `useMenuKeyboardNavigation` being used here as well but figured I'd ask anyway 😄

src/components/button/button.component.tsx
robinzigmond: Oh yeah, I was thinking uppercase was only for components but not sure why I didn't think to check other contexts that we define and use in Carbon 🤦

src/components/split-button/split-button.context.ts
robinzigmond: I don't object, but I thought the convention was that anything not exported from the `index.ts` (which this isn't) was considered internal and use-at-your-own-risk anyway. But sure we have `__internal__` subfolders in other components - happy to put this in one of those if it makes things clearer.

src/components/button/button.component.tsx
robinzigmond: It's a fair comment. I like having the explicit boolean here rather than assume the presence of another property implies anything - as what if we later we want to make `onChildButtonClick` optional even in the SplitButton case. As ever I'm probably worrying about something that will never happen 😂 - but I don't see much harm in being explicit. Happy to be persuaded otherwise though.

src/components/multi-action-button/multi-action-button.component.tsx
robinzigmond: Fair comment. I did realise there was a lot of duplication between the two components but realised a lot of it already existed so didn't really explore further - but now you mention it I agree it's worth exploring whether some of this duplication can be put into a custom hook 👍

src/components/button/button.component.tsx
edleeks87: Happy to let the second reviewer split the difference 😄

edleeks87 (approved):

src/components/button/button.component.tsx
DipperTheDan: Personally, I prefer `const inSplitButton = !!onChildButtonClick;` approach for this particular need. As we don't have any future plans plans to change `onChildButtonClick` and if we did, we could always implement what you have here but seeing as @edleeks87 has already approved the PR, I'm not going to block the PR/request changes over it. Just my preference and not a hill I want to die on 😆

DipperTheDan: Personally, I prefer `const inSplitButton = !!onChildButtonClick;` approach for this particular need. As we don't have any future plans plans to change `onChildButtonClick` and if we did, we could always implement what you have here but seeing as @edleeks87 has already approved the PR, I'm not going to block the PR/request changes over it. Just my preference and not a hill I want to die on 😆

DipperTheDan (approved):

edleeks87 (approved):

divyajindel (approved):

PR #6217, opened 2023-07-25
Sage/DSY-1746 Button Toggle styling changes

DobroTora:
### Proposed behaviour
New Styling Added to Button Toggle component

### Current behaviour
New Styling added to Button Toggle

### Checklist

#### QA

### Testing instructions


src/components/button-toggle/button-toggle.stories.tsx
```diff
@@ -2,27 +2,57 @@ import React from "react";
 import { ComponentStory } from "@storybook/react";
 
 import { ButtonToggle, ButtonToggleGroup } from ".";
-import Box from "../box";
 
 export const Default: ComponentStory<typeof ButtonToggle> = () => (
-  <ButtonToggleGroup id="button-toggle-group-id" label="Default example">
+  <ButtonToggleGroup
+    variant="default"
```
tomdavies73, line 8: **suggestion**: the `variant` prop doesn't need to be passed here

src/components/button-toggle/button-toggle.stories.tsx
```diff
@@ -75,6 +157,24 @@ export const small: ComponentStory<typeof ButtonToggle> = () => (
   </ButtonToggleGroup>
 );
 
+export const smallMinor: ComponentStory<typeof ButtonToggle> = () => (
```
tomdavies73, line 160: **issue**: Doesn't look like this story has been included into `button-toggle.stories.mdx`

tomdavies73: **suggestion**: include story like so `story={stories.smallMinor}`(even though i'm against this approach)

src/components/button-toggle/button-toggle-group/button-toggle-group.stories.tsx
```diff
@@ -167,3 +302,27 @@ export const AriaLabel: ComponentStory<typeof ButtonToggleGroup> = () => {
     </ButtonToggleGroup>
   );
 };
+
+export const AriaLabelMinor: ComponentStory<typeof ButtonToggleGroup> = () => {
```
tomdavies73, line 306: **isue:**: story is not used in `button-toggle.stories.mdx`

tomdavies73: **suggestion**: add story like so `groupStories.AriaLabelMinor` (even though i'm against this approach)

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -126,6 +128,31 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
       }
       cursor: not-allowed;
     `};
+
+  ${({ variant }) =>
```
tomdavies73, line 132: **thought**: this all looks good, and just testing the storybook it looks fine, but i'll leave the specific checking of these tokens against the DS to QA

src/components/button-toggle/button-toggle.stories.tsx
```diff
@@ -201,6 +421,24 @@ export const disabled: ComponentStory<typeof ButtonToggle> = () => (
   </ButtonToggleGroup>
 );
 
+export const disabledMinor: ComponentStory<typeof ButtonToggle> = () => (
```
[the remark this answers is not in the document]
DobroTora: FYI, there is a difference between minor and major buttons in terms of padding for example, not just a hover state. Scope of this ticket is to add minor child of every story, so there is a need to add disabled state as well.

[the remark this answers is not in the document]
tomdavies73: Hi Debra, I agree when it comes to other variations, but here with this approach on `ButtonBar` there is no difference padding wise etc, just various different colours, therefore when both are disabled there is no visal difference.
Screenshot 2023-07-31 at 12 27 52

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -126,6 +128,31 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+    variant === "minor" &&
```
tomdavies73, line 133: **issue**: a check against the `disabled` prop should be added here. To stop any styling being added when `disabled` is true. If you hover on a `disabled` button with the minor variant, the hover colors are still applied

src/components/button-toggle/button-toggle.stories.tsx
edleeks87: We should follow the naming convention so it should be `SmallMinor`

src/components/button-toggle/button-toggle.style.ts
edleeks87: There are some differences between the two variants fro example the border-radius is smaller when `minor` is set, it would probably work better to move the block setting the `disabled` styling is applied after the `minor` styling

src/components/button-toggle/button-toggle.stories.tsx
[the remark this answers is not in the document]
edleeks87: There is a difference based on the designs so IMO we should keep this, the styling will need to be updated to ensure it looks right though

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -65,6 +67,40 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
   box-sizing: border-box;
   max-width: 100%;
 
+  ${({ variant }) =>
+    variant === "minor" &&
+    css`
+      border: 1px solid var(--colorsActionMinor500);
+
+      &&&&&& {
```
edleeks87, line 75: Comment: This doesn't feel like the right approach to me, it feels quite brittle and we could end up needing to raise specificity to ensure this works. I think it would be better to move the border radius styling into the `StyleButtonToggleWrapper` element and pass `variant` into that so we can set the radius accordingly.

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -65,6 +67,40 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+        border-radius: var(--borderRadius050);
+        padding: var(--spacing000) var(--spacing100) var(--spacing000)
+          var(--spacing100);
+      }
+
+      &:focus {
```
edleeks87, line 81: Comment: you don't need this as far as I can tell. You should style it using the `aria-pressed=true` attribute similar to how it is achieved above. You'll also need to the padding imo as when the toggle button is active the increased border thickness will cause the layout to jump so the padding will need to be reduced when that happens to avoid it. Something like the below will work:

```
color: var(--colorsActionMinor500);
      border: 1px solid var(--colorsActionMinor500);

      &[aria-pressed="true"] {
        border: 4px solid var(--colorsActionMinor500);
        background-color: transparent;
        padding: 0 ${paddingConfig[size] - 3}px;
        
        :hover {
          background-color: var(--colorsActionMinor600);
          color: var(--colorsUtilityYang100);
          border-color: var(--colorsActionMinor600);
        }
      }

      :not([aria-pressed="true"]):not(:disabled):hover {
        color: var(--colorsActionMinorYang100);
        background-color: var(--colorsActionMinor600);
        border-color: var(--colorsActionMinor600);
      }
```

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -65,6 +67,40 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+        border: 4px solid var(--colorsActionMinor500);
+        background-color: var(--colorsActionMinor600);
+      }
+
+      &&&:hover {
```
edleeks87, line 86: comment: see above for suggested change, I don't think we should use `&&&` but instead use the same style attribute to target it that we are trying to override etc

src/components/button-toggle/button-toggle.style.ts
edleeks87:
```
const StyledButtonToggleWrapper = styled.div<StyledButtonToggleWrapperProps>`
  display: inline-block;
  vertical-align: middle;

  ${({ grouped, variant }) =>
    css`
      ${!grouped &&
      css`
        &&&& {
          ${StyledButtonToggle} {
            border-radius: ${variant === "default"
              ? "var(--borderRadius400)"
              : "var(--borderRadius050)"};
          }
        }
      `}

      ${grouped &&
      css`
        &&&& {
          :first-of-type {
            ${StyledButtonToggle} {
              border-top-left-radius: ${variant === "default"
                ? "var(--borderRadius400)"
                : "var(--borderRadius050)"};
              border-bottom-left-radius: ${variant === "default"
                ? "var(--borderRadius400)"
                : "var(--borderRadius050)"};
            }
          }

          :last-of-type {
            ${StyledButtonToggle} {
              border-top-right-radius: ${variant === "default"
                ? "var(--borderRadius400)"
                : "var(--borderRadius050)"};
              border-bottom-right-radius: ${variant === "default"
                ? "var(--borderRadius400)"
                : "var(--borderRadius050)"};
            }
          }
        }
      `}
    `}

  &:not(:first-of-type) {
    margin-left: 8px;
  }

  ${({ grouped }) =>
    grouped &&
    css`
      &:not(:first-of-type) {
        margin-left: -1px;
      }
    `};
`;
```

src/components/button-toggle/button-toggle-group/button-toggle-group.component.tsx
```diff
@@ -111,6 +113,7 @@ const ButtonToggleGroup = ({
   name,
   onChange,
   value,
+  variant,
```
edleeks87, line 116:
```suggestion
  variant = "default",
```

edleeks87: We should default the value imo

src/components/button-toggle/button-toggle.style.ts
edleeks87: https://codesandbox.io/s/carbon-quickstart-forked-f82m7p?file=/src/App.js

If you click one of the buttons you will see the content jumping

edleeks87 (changes requested):
We'll need to prevent the content jumping when the minor buttons are pressed/ activated

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -220,6 +220,25 @@ context("Testing Button-Toggle component", () => {
     });
   });
 
+  describe("should render Button Toggle with variant prop", () => {
+    it("should render Button Toggle with variant prop set to default value", () => {
+      CypressMountWithProviders(
```
edleeks87: there's no assertion here

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -220,6 +220,25 @@ context("Testing Button-Toggle component", () => {
...
+        <stories.ButtonToggleGroupComponent
+          id="button-toggle-group-id"
+          variant="default"
+        />
+      );
+    });
+    it("should render Button Toggle with variant prop set to minor value", () => {
```
edleeks87: there's also no assertion here

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -220,6 +224,22 @@ context("Testing Button-Toggle component", () => {
     });
   });
 
+  describe("should render Button Toggle with variant prop", () => {
+    const variants = ["default", "minor"];
+
+    it.each(variants)(
```
edleeks87, line 230: There's no assertion here

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -1,10 +1,14 @@
 /* eslint-disable no-unused-expressions, jest/valid-expect, jest/valid-expect-in-promise */
 import React from "react";
+import { mount } from "cypress/angular";
```
edleeks87, line 3: This is not the correct mount

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -220,6 +224,22 @@ context("Testing Button-Toggle component", () => {
     });
   });
 
+  describe("should render Button Toggle with variant prop", () => {
+    const variants = ["default", "minor"];
+
+    it.each(variants)(
+      "renders with variant prop, which is set to %s",
+      (variantType) => {
+        mount(
```
edleeks87, line 233: We should be using `CypressMountWithProviders`

cypress/components/button-toggle/button-toggle.cy.tsx
edleeks87: Maybe assert it has the correct radius/ background styling etc?

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -65,9 +67,39 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+          color: var(--colorsActionMinorYang100);
+        }
+      }
+    `}
+
+  ${({ size, variant }) => css`
     height: ${heightConfig[size]}px;
-    padding: 0 ${paddingConfig[size]}px;
+    padding: ${variant === "default"
+      ? `0 ${paddingLargeIconConfig[size]}px;`
```
edleeks87, line 101: Where does this come from, I think you can just use `paddingConfig` that's already defined in this file

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -65,9 +67,39 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
   box-sizing: border-box;
   max-width: 100%;
 
-  ${({ size }) => css`
+  ${({ variant }) =>
```
edleeks87, line 70:
```suggestion
  ${({ variant, size }) =>
```

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -106,26 +138,49 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
       height: ${heightLargeIconConfig[size]}px;
-      padding: 0 ${paddingLargeIconConfig[size]}px;
+      padding: ${variant === "default"
+        ? `0 ${paddingLargeIconConfig[size]}px;`
+        : `var(--spacing000) var(--spacing100) var(--spacing000) var(--spacing100)`};
       flex-direction: column;
     `}
-  ${({ disabled }) =>
+
+  ${({ disabled, variant }) =>
```
edleeks87, line 152: you'll need to destructure size here

edleeks87: It would probably be good to look through this file and consolidate it a bit as it there's multiple points where we are destructuring props (for example `size` and `variant`) and then later in the file doing the same thing which makes it more difficult to read maintain imo. If/ when you do be careful with the ordering as we will need the `minor` styling to override the default in most cases

src/components/button-toggle/button-toggle-group/button-toggle-group.spec.tsx
```diff
@@ -487,3 +509,6 @@ describe("ButtonToggleGroup", () => {
     (component) => component.find(StyledButtonToggleGroup)
   );
 });
+function CypressMountWithProviders(arg0: React.JSX.Element) {
```
edleeks87, line 512: You don't need this in a unit test

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +138,46 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+
+      :not([aria-pressed="true"]):not(:disabled):hover {
+        color: var(--colorsActionMinorYang100);
+        background-color: var(--colorsActionMinor600);
+        ${StyledIcon} {
+          color: var(--colorsActionMinorYang100);
+        }
+      }
+
+      ${!disabled &&
```
nineteen88, line 164: suggestion:  we probably should approach this differently as it's quite confusing

Personally I think any default styles should be applied first i.e. with your `variant === minor` check, and then we override the necessary styles if the `disabled` prop is true. The way we're currently approaching it feels like a bit of a double negative and it makes it more challenging to understand what's going on.

src/components/button-toggle/button-toggle.component.tsx
```diff
@@ -107,8 +107,8 @@ export const ButtonToggle = ({
     isInGroup,
     firstButton,
     childButtonCallbackRef,
+    variant,
```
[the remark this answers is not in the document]
nineteen88: This hasn't been resolved

nineteen88 (commented):
An addition to these comments there's also an issue with the minor variant's focus border. On previous button toggle examples, when you have selected a button the focus border will show, however, whenever you select a minor variant button the focus border never shows until it's deselected again. It seems like the selected border is overriding the focus border styles, but I haven't determined the specific piece of code that is causing this.

cypress/components/button-toggle/button-toggle.cy.tsx
edleeks87: Yeah agree we can delete this one

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +139,43 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
         }
       }
       cursor: not-allowed;
-    `};
+    `}
+
+  ${({ variant, disabled }) =>
```
edleeks87, line 144: comment: `disabled` prop isn't being used here

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +139,43 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+        background-color: var(--colorsActionMinor600);
+        color: var(--colorsActionMinorYang100);
+
+        ${StyledIcon} {
+          color: var(--colorsActionMinorYang100);
+        }
+      }
+
+      &[aria-pressed="true"]:not(:hover) {
+        box-shadow: inset 0px 0px 0px 3px var(--colorsActionMinor500);
```
edleeks87, line 175: comment: the focus styling is still not correct when a button has been pressed, you'll need to add something like this below line 176:

```

        :focus {
          box-shadow: inset 0px 0px 0px 3px var(--colorsActionMinor500), 0px 0px 0px var(--borderWidth300) var(--colorsSemanticFocus500), 0px 0px 0px var(--borderWidth600) var(--colorsUtilityYin090);
        }

File: src/components/button-toggle/button-toggle.style.ts [hunk 24]
```diff
@@ -176,19 +217,22 @@ StyledButtonToggle.defaultProps = { theme: baseTheme };
...
 
-  ${({ grouped }) =>
+  ${({ grouped, variant }) =>
     css`
       ${!grouped &&
       css`
         &&&& {
           ${StyledButtonToggle} {
-            border-radius: var(--borderRadius400);
+            border-radius: ${variant === "default"
```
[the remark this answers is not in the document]
edleeks87: I agree with @nineteen88  but I don't mind moving the reorganisation of the styling for `StyledButtonToggleWrapper` to a new ticket if he agrees

src/components/button-toggle/button-toggle.component.tsx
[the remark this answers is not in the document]
edleeks87: This will get the default value from the context as it's defaulted in the group component

cypress/components/button-toggle/button-toggle.cy.tsx
edleeks87: This can be deleted @DobroTora as the tests below covers this what we need

edleeks87 (changes requested):
We'll need to ensure the focus state is applied to `pressed` buttons when the `variant` is `minor`

src/components/button-toggle/button-toggle.style.ts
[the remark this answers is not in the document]
nineteen88: I partially agree that we can do this later, but the only part that's holding me back on fully agreeing is the snapshot changes. It's not the end of the world of course

edleeks87 (commented):
@DobroTora I'm happy to approve this but it looks like several of @tomdavies73's comments haven't been addressed/ resolved

DobroTora:
> @DobroTora I'm happy to approve this but it looks like several of @tomdavies73's comments haven't been addressed/ resolved

I have requested a review and received it from @tomdavies73  :)

cypress/components/button-toggle/button-toggle.cy.tsx
```diff
@@ -94,40 +97,6 @@ context("Testing Button-Toggle component", () => {
       });
     });
 
-    describe("rounded corners", () => {
```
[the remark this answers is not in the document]
DobroTora: Rounded corners will conditionally change if the prop will be passed and css variable for design token will be different.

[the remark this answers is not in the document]
tomdavies73: Okay that's fine, but we should be creating tests for either scenario instead of just deleting the rounded corners tests completely, imo

src/components/button-toggle/button-toggle.component.tsx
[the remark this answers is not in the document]
tomdavies73: This only gets the default value from context if its wrapped in a `ButtonToggleGroup`, if its being used independently, or not wrapped it could cause issues. Looks like this is actually causing the cypress tests which Debra removed to fail, when its added they pass

cypress/components/button-toggle/button-toggle.cy.tsx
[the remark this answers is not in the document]
tomdavies73: After doing some digging it looks like this was just failing because the variants were getting messed up. The `ButtonToggleComponent` we use on the failing test returns multiple `ButtonToggle`'s which are wrapped by a `<div />` and not a `ButtonToggleGroup`. If we ensure the `variant` prop in `button-toggle.component.tsx` is default like so `variant="default"` the tests pass. Alternatively, if you would like you can wrap the `ButtonToggle`s rendered in `ButtonToggleComponent` with a `ButtonToggleGroup`, instead of a `<div />` to achieve the same result :)

src/components/button-toggle/button-toggle.stories.tsx
tomdavies73: @DobroTora just to state that we definitely don't need to be passing `variant="default"` here, your context approach has ensured that when no `variant` prop is present it will always default to `"default"`. If `variant="default"` was needed to be passed in order to ensure correct behaviour, you would have to do this on every story. Furthermore, this would be a massive issue for consumers as they'd have to opt back in to default functionality, causing huge visual changes to their UI. I've tested this locally with your branch and it looks spot on compared to master :). So yeah you're good to remove this imo

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -176,19 +220,22 @@ StyledButtonToggle.defaultProps = { theme: baseTheme };
...
```
edleeks87, line 236: Has this been unintentionally reverted? Weren't we checking for minor here before like we are below?

DobroTora: It was definitely not intentional - thank you for spotting that -  I will fix it.

tomdavies73: Nothing else from me, apart from this

edleeks87 (dismissed):

tomdavies73 (dismissed):

DobroTora:
As per last week DS request:

- minor/default prop are going to be removed
- current Button Minor Toggle will have minor styling.
- Figma Designs: [(https://www.figma.com/file/Tau6qv5iyr8wthWNqYktej/DS-Official-Components?type=design&node-id=25270-303744&mode=design&t=rSp19yUK20wrler2-0)]

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +136,71 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+  }
+
+  & ${StyledIcon} {
+    color: var(--colorsActionMinor500);
+    :not([aria-pressed="true"]):not(:disabled):hover {
+      color: var(--colorsActionMinorYang100);
+    }
+  }
+
+  color: var(--colorsActionMinor500);
```
edleeks87, line 185: Comment: You can delete this as this is the same style set on line 148

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +136,71 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+
+      & ${StyledIcon} {
+    color: var(--colorsActionMinor500);
+    :not([aria-pressed="true"]):not(:disabled):hover {
+      color: var(--colorsActionMinorYang100);
+    }
+  }
+
+  color: var(--colorsActionMinor500);
+  border: 1px solid var(--colorsActionMinor500);
```
edleeks87, line 149: Comment: The border styles are set on line 85 already so you can delete this

src/components/button-toggle/button-toggle.stories.tsx
```diff
@@ -13,16 +12,14 @@ export const Default: ComponentStory<typeof ButtonToggle> = () => (
 );
 
 export const DefaultWrappedText: ComponentStory<typeof ButtonToggle> = () => (
```
edleeks87, line 14: comment: this changes the layout of the story

master:
image

proposed changes:

image

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +136,71 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+    background-color: transparent;
+    :focus {
+      box-shadow: inset 0px 0px 0px 3px var(--colorsActionMinor500),
+        0px 0px 0px var(--borderWidth300) var(--colorsSemanticFocus500),
+        0px 0px 0px var(--borderWidth600) var(--colorsUtilityYin090);
+    }
+  }
+
+  & ${StyledIcon} {
+    color: var(--colorsActionMinor500);
```
edleeks87, line 179: comment: you don't need this line the colour is set on line 116

edleeks87 (dismissed):

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +136,63 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+    box-shadow: inset 0px 0px 0px 3px var(--colorsActionMinor500);
+    background-color: transparent;
+    :focus {
+      box-shadow: inset 0px 0px 0px 3px var(--colorsActionMinor500),
+        0px 0px 0px var(--borderWidth300) var(--colorsSemanticFocus500),
+        0px 0px 0px var(--borderWidth600) var(--colorsUtilityYin090);
+    }
+  }
+
+  border: 1px solid var(--colorsActionMinor500);
```
nineteen88, line 178: comment: another place where we have the duplicate border definition and should be removed

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -136,7 +136,63 @@ const StyledButtonToggle = styled.button<StyledButtonToggleProps>`
...
+
+  :not([aria-pressed="true"]):not(:disabled):hover {
```
nineteen88, line 180: comment: in fact all of this is duplicated. Can you go through the styles and make sure we have removed all duplicate styling definitions?

src/components/button-toggle/button-toggle.style.ts
```diff
@@ -188,7 +252,7 @@ const StyledButtonToggleWrapper = styled.div<StyledButtonToggleWrapperProps>`
       css`
         &&&& {
           ${StyledButtonToggle} {
-            border-radius: var(--borderRadius400);
+            border-radius: var(--borderRadius050);
```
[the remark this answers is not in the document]
nineteen88: If this is done, can we resolve this comment?

src/components/button-toggle/button-toggle-test.stories.tsx
```diff
@@ -41,6 +41,34 @@ export const DefaultStory = () => {
   );
 };
 
+export const ButtonToggleGroupMinorStory = () => {
```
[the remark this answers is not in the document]
nineteen88: @DobroTora is there any reason this hasn't been removed yet?

src/components/button-toggle/button-toggle.style.ts
nineteen88: @DobroTora the duplicate styles are still there

src/components/button-toggle/button-toggle.style.ts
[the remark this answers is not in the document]
nineteen88: @edleeks87 are you happy with this?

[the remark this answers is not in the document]
edleeks87: Yeah looks good to me I'll resolve it now

src/components/button-toggle/button-toggle.style.ts
DobroTora: I just updated and hopefully removed duplicated styles.

nineteen88 (approved):

edleeks87 (dismissed):

edleeks87 (dismissed):

nineteen88 (approved):

stephenogorman (dismissed):

edleeks87 (approved):

stephenogorman (approved):

PR #6231, opened 2023-08-04
 fix(textarea, textbox): improve character count accessibility and align with design system

tomdavies73:
### Proposed behaviour

Major accessibility improvements have been made on both `Textarea` and `Textbox` as well as design changes to ensure both components character count functionality is aligned with  the design system. The default hint message which was applied when the `characterLimit` prop was passed has been removed. 

Consumers now have the option to pass a hint with the `inputHint` prop. This hint is intended to help contextualise the input, and not inform users of any character count functionality. This has been achieved with a mandatory auto-applied hint div which is visually hidden. This is referenced via a aria-describedby attribute on both `Textarea` and `Textbox`'s inputs. This ensures screen reader users will know of any character count functionality, and character limits. This hint is auto applied and mandatory, however we support customising this hint via translations with the `I18nProvider` component. 

Also, to ensure screen reader users do not have their count announced every time a character is typed or removed, a seperate debounced visually hidden aria-live region has been added. Therefore the current character count will only be announced after a reasonable period has passed after typing. Also, the count itself underneath the input's text has been left aligned opposed to right aligned, to improve readability and accessibility. 

Finally, the character count functionality on `Textarea` and `Textbox` is now also fully aligned with the design system, with all character count messages matching by default. However, these can also be overridden via a `I18nProvider`.

Screenshot 2023-09-22 at 17 23 19

Screenshot 2023-09-22 at 17 23 37

### Current behaviour

Currently, there is a default hint string above the character counter which users can change by wrapping the component in an `I18nProvider`, however it does not inform users about any restrictions on the input. Also text on the actual counter underneath the input is right-aligned.

Screenshot 2023-08-04 at 11 24 55

### Checklist

#### QA

### Testing instructions


DipperTheDan (commented):
**issue(blocking):** There seems to be a bug whereby if you set the character as a value that requires a separator, you're given two differently formatted values.
   
Screenshot 2023-08-08 at 16 44 44

src/components/textarea/textarea.spec.tsx
```diff
@@ -315,38 +315,76 @@ describe("Textarea", () => {
   });
 
   describe(`when the characterLimit prop is passed`, () => {
-    it.each([2, 3, 4])("renders a character counter", (characterLimit) => {
+    let characterLimit = 4;
+    let hintString =
```
Parsium, line 319: **polish**: while moving these two variables outside of the three tests on lines 345, 353 and 371 follows DRY principles, it does make the tests a little bit harder to read.

src/components/textarea/textarea.stories.mdx
Parsium, line 1: **issue**: it seems we never actually had the translation keys for `characterCount` documented for `Textarea` 😬

 What do you think about adding them in as part of this PR?

He is one [example of a translation docs section](https://carbon.sage.com/?path=/docs/numeral-date--default#translation-keys) for reference.

Parsium (commented):
**suggestion**: I do like the idea of allowing consumers to modify the hint text with the character count in it, though I think we could simplify the approach slightly for consumers.

Instead of creating a new prop and replacing the count with regex, we could document to consumers to pass the `hintString` translation key into the `labelHelp` prop like so:

```tsx
// inside consumer's component code
<i18nProvider
   locale={locale}
>
   <Textbox
      label="Textbox"
      value={state}
      onChange={setValue}
      characterLimit={10}
      labelHelp={getTranslationKey("characterCount.hintString")(10)}
    />
</i18nProvider>

/* where getTranslation() refers to some method consumers use to return a
translation key, in accordance to their app setup. The character count is
passed to the translation key */
```
This would simplify the process of modifying the hint text for consumers, as they will be already using translation keys like this elsewhere in their app's code.

I'm happy to talk about this some more externally if you would like too :)

src/components/textarea/textarea.component.tsx
```diff
@@ -379,7 +400,9 @@ export const Textarea = React.forwardRef(
             >
               {characterLimit || inputHint ? (
                 <StyledInputHint id={hintIdValue} data-element="input-hint">
-                  {characterCountHint || inputHint}
+                  {formattedCharacterCountHint ||
```
edleeks87, line 403: Question: do we need to have 3 string props for this? Couldn't this be achieved via a single `inputHint` string prop, if it needs to be formatted you could check if the `characterLimit` is set etc

```
const formattedHint = characterLimit ? inputHint?.replace(
      /\{.*?\}/g,
      formattedCharacterLimit
    ) : inputHint;
```

src/components/textarea/textarea.component.tsx
```diff
@@ -278,6 +287,18 @@ export const Textarea = React.forwardRef(
...
+      new Intl.NumberFormat(locale).format(rawValue);
+    const capturedLocale = characterCountHint?.match(/{([^']+)}/)?.[1];
+    const formattedCharacterLimit = getFormatNumber(
+      characterLimit as number,
+      capturedLocale as string
+    );
+    const formattedCharacterCountHint = characterCountHint?.replace(
+      /\{.*?\}/g,
+      formattedCharacterLimit
+    );
```
Parsium, line 300: **suggestion**: Extending on @edleeks87's suggestion, we could also simplify this code to something like:

```tsx
const getInputHintWithCharacterLimit = (rawString: string, countLimit: number) => {
      const capturedLocale = rawString.match(/{([^']+)}/)?.[1];
      const format = new Intl.NumberFormat(capturedLocale).format(countLimit);
      return rawString.replace(/\{.*?\}/g, format);
};
```

which would allow us to do this when rendering the input hint:

```ts
const formattedHint = (inputHint && characterLimit) ? getInputHintWithCharacterLimit(inputHint, characterLimit) : inputHint;
```

src/components/textbox/textbox.component.tsx
```diff
@@ -358,7 +379,9 @@ export const Textbox = React.forwardRef(
           >
             {characterLimit || inputHint ? (
               <StyledInputHint id={hintIdValue} data-element="input-hint">
-                {characterCountHint || inputHint}
+                {formattedCharacterCountHintString ||
+                  characterCountHintString ||
+                  inputHint}
```
Parsium, line 384: **suggestion**: likewise you might be able to use @edleeks87's suggestion for `Textbox` too :)

src/components/textbox/textbox.spec.tsx
```diff
@@ -68,37 +68,81 @@ describe("Textbox", () => {
   );
 
   describe(`when the characterLimit prop is passed`, () => {
-    it.each([2, 3, 4])("renders a character counter", (characterLimit) => {
+    let wrapper: ReactWrapper;
+    let characterLimit = 4;
+    let hintString =
+      "this is a test string where the limit will be nested here {}";
```
Parsium, line 74: **suggestion**: likewise I would define these locally in each test to help with readability, especially since not every test in this block uses all three variables.

src/hooks/__internal__/useCharacterCount/useCharacterCount.tsx
```diff
@@ -16,7 +16,13 @@ const useCharacterCount = (
...
     typeof characterLimit === "number" && !Number.isNaN(characterLimit);
   const l = useLocale();
-  const hintString = l.characterCount.hintString();
+
+  const getFormatNumber = (rawValue: number, locale: string) =>
+    new Intl.NumberFormat(locale).format(rawValue);
+
+  const hintString = l.characterCount.hintString(
+    getFormatNumber(characterLimit as number, l.locale())
+  );
```
Parsium, line 25: **issue**: I think we need a check here in case `characterLimit` is not undefined rather than doing a type assertion, since if `characterLimit` is undefined, there is a chance the code on line 21 could throw an error.

```tsx
const hintString =
     isCharacterLimitValid 
        ? l.characterCount.hintString(getFormatNumber(characterLimit, l.locale())
        : undefined;
```

src/components/textarea/textarea.spec.tsx
Parsium, line 319: **suggestion**: while moving these two variables outside of the three tests on lines 345, 353 and 371 follows DRY principles, it does make the tests a little bit harder to read, so I would suggest making these local to each test.

src/components/textarea/textarea.stories.mdx
Parsium, line 1: **issue**: it seems we never actually had the translation keys for `characterCount` documented for `Textarea` 😬

 What do you think about adding them in as part of this PR?

He is one [example of a translation docs section](https://carbon.sage.com/?path=/docs/numeral-date--default#translation-keys) for reference.

src/components/textbox/textbox.component.tsx
Parsium, line 1: **issue**: it seems previously we did not update the documentation for the `characterCount` translation keys for `Textbox`.

If you look at `Textbox`'s docs, you will see the currently-used keys `characterCount.hintString`, `characterCount.tooManyCharacters` and `characterCount.charactersLeft` are not shown:

Screenshot 2023-08-16 at 14 19 55

What do you think about replacing the table with these three keys as part of this PR?

Parsium:
**note**: I just noticed we never added any documentation for the three character count translation keys 😬

- `characterCount.tooManyCharacters`
- `characterCount.hintString`
- `characterCount.charactersLeft`

I've created a separate ticket for this though:

**FE-6201**

tomdavies73:
> **issue(blocking):** There seems to be a bug whereby if you set the character as a value that requires a separator, you're given two differently formatted values.
> 
> Screenshot 2023-08-08 at 16 44 44

So this one looks like it's not a bug, it's just because it hasn't been wrapped in a I8n provider, if it was and the `"es-ES"` locale was passed to it the translations

src/components/textarea/textarea.component.tsx
tomdavies73: Yeah, this is a good point, even though the intentions were good by making sure a dedicated hint was provided via the `characterCountHint` which supersedes an `inputHint`, I feel like we can just use the `inputHint` prop and trust consumers to do the right thing here

src/hooks/__internal__/useCharacterCount/useCharacterCount.tsx
tomdavies73: We already have a similar check in the file on line 45

DipperTheDan (approved):

Parsium (approved):

DipperTheDan (approved):

src/hooks/__internal__/useCharacterCount/useCharacterCount.tsx
```diff
@@ -1,22 +1,29 @@
...
-  string | undefined,
-  string | null
-] => {
+): [number | undefined, JSX.Element | null, string | undefined] => {
   const isCharacterLimitValid =
     typeof characterLimit === "number" && !Number.isNaN(characterLimit);
-  const l = useLocale();
-  const hintString = l.characterCount.hintString();
+
+  const [debouncedValue, setDebouncedValue] = useState(value);
```
DipperTheDan, line 14: **suggestion(non-blocking):** Not going to make you do this now as I know we have time constraints on this. It might be worth considering in the future to potentially use a ref here instead of `useState`. For now the performance benefits will be negligible though. I'd rather the work be done and we can focus on incremental improves in the future 😃

src/components/textarea/textarea.stories.mdx
```diff
@@ -82,6 +83,9 @@ import Textarea from "carbon-react/lib/components/textarea";
 
 ### With characterLimit
 
+If you use the `inputHint` prop to provide the user with a hint before the input, please use a full stop `.` at the end,
+as it forces a pause before any other announcements, this well help screan reader users understand the hint fully.
```
DipperTheDan, line 87:
```suggestion
as it forces a pause before any other announcements, this well help screen reader users understand the hint fully.
```

tomdavies73: Brilliant spot 👍, blaming the new keyboard for this one :/

src/__internal__/character-count/character-count.spec.tsx
```diff
@@ -17,13 +21,75 @@ describe("CharacterCount", () => {
...
+      );
+    });
+
+    it("visually hidden character counter has aria-live='polite'", () => {
+      expect(wrapper.find(VisuallyHiddenCharacterCount).prop("aria-live")).toBe(
+        "polite"
+      );
+    });
+
+    it("visually hidden character counter has the same count message", () => {
```
Parsium, line 52: **suggestion(non-blocking)**: might be worth us tweaking this description slightly to clear up any potential ambiguities

```suggestion
    it("visually hidden character counter has the same count message as visual character counter", () => {
```

src/__internal__/character-count/character-count.spec.tsx
```diff
@@ -17,13 +21,75 @@ describe("CharacterCount", () => {
...
+      expect(wrapper.find(VisuallyHiddenCharacterCount).text()).toBe(
+        "5 characters left"
+      );
+    });
+
+    it("visually hidden character counter has visuallyHidden class applied", () => {
```
Parsium, line 58: **suggestion(non-blocking)**: likewise here, since we aren't importing the `visuallyHidden` class for use inside the test, this might help another developer understand the context behind this test if we decide amend/remove the `visuallyHidden` util in the future.

```suggestion
    it("visually hidden character counter has css properties to visually hide it from view", () => {
```

src/components/textbox/textbox.stories.mdx
```diff
@@ -52,6 +52,9 @@ import Textbox from "carbon-react/lib/components/textbox";
 
 ### Character counter
 
+If you use the `inputHint` prop to provide the user with a hint before the input, please use a full stop `.` at the end,
+as it forces a pause before any other announcements, this well help screan reader users understand the hint fully.
```
Parsium, line 56: **suggestion**: similar to @DipperTheDan's comment :)

```suggestion
as it forces a pause before any other announcements, this well help screen reader users understand the hint fully.
```

src/components/textarea/textarea.stories.mdx
```diff
@@ -214,3 +218,34 @@ It is possible to use the `tooltipPosition` to override the default placement of
...
+        +" This will also be announced to screenreaders.",
+      type: "func",
+      returnType: "string",
+    },
+        {
+      name: "characterCount.charactersLeft",
+      description:
+        "The message displayed below the input which will inform"
+        +" users how many characters they have before they reach  or exceed the"
+        +" set `characterLimit`. This will also be announced to screenreaders.",
```
Parsium, line 238: **suggestion**

```suggestion
        +" users if they have exceeded the set `characterLimit` and how much by." 
        +" This will also be announced to screen readers.",
      type: "func",
      returnType: "string",
    },
        {
      name: "characterCount.charactersLeft",
      description:
        "The message displayed below the input which will inform"
        +" users how many characters they have before they reach  or exceed the"
        +" set `characterLimit`. This will also be announced to screen readers.",      
```

src/components/textbox/textbox.stories.mdx
```diff
@@ -229,11 +232,30 @@ The following keys are available to override the translations for this component
...
       description:
-        "The locale code to display the correct number format in the counter",
+        "The message which will be read out to screen reader users, informing them"
+        +" of the set `characterLimit`.",
       type: "func",
       returnType: "string",
     },
   ]}
-/>
+/>
```
Parsium, line 261: **suggestion**

```suggestion
      name: "characterCount.tooManyCharacters",
      description:
        "The message displayed below the input which will inform"
        +" users if they have exceeded the set `characterLimit` and how much by." 
        +" This will also be announced to screen readers.",
      type: "func",
      returnType: "string",
    },
        {
      name: "characterCount.charactersLeft",
      description:
        "The message displayed below the input which will inform"
        +" users how many characters they have before they reach  or exceed the"
        +" set `characterLimit`. This will also be announced to screen readers.",      
        type: "func",
      returnType: "string",
    },
        {
      name: "characterCount.visuallyHiddenHint",
      description:
        "The message which will be read out to screen reader users, informing them"
        +" of the set `characterLimit`.",
      type: "func",
      returnType: "string",
    },
  ]}
/>
```

src/locales/locale.ts
Parsium, line 1: **suggestion**: since we are adding and removing some translation keys here, might be worth us making this change in a separate commit, so semantic release mentions this as a change in our release notes.

@edleeks87 @DipperTheDan would you class this as a breaking change given we are removing `characterCount.hintString`?

cypress/locators/index.js
```diff
@@ -60,8 +60,7 @@ export const pressESCKeyOntoFocusedElement = () => {
   cy.focused().type("{esc}");
 };
 export const commonInputPrefix = () => cy.get(COMMON_INPUT_PREFIX);
-export const commonInputCharacterLimit = () =>
-  cy.get(COMMON_INPUT_CHARACTER_LIMIT);
+export const characterCount = () => cy.get(CHARACTER_COUNT);
```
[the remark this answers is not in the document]
tomdavies73: I'm working on the `Textbox` refactor and I think that may be merged after this, if that's the case i'll just add it during the refactor

src/__internal__/character-count/character-count.spec.tsx
```diff
@@ -17,13 +21,75 @@ describe("CharacterCount", () => {
...
+      );
+    });
+
+    it("visually hidden hint has correct message based on 'characterLimit", () => {
+      expect(wrapper.find(VisuallyHiddenHint).text()).toBe(
+        "You can enter up to 10 characters"
+      );
+    });
+
+    it("visually hidden hint has visuallyHidden class applied", () => {
```
Parsium, line 80: **suggestion**: we should update this test in a similar fashion to the one on line 52.

```suggestion
    it("visually hidden hint has css properties to visually hide it from view", () => {
```

src/hooks/__internal__/useCharacterCount/useCharacterCount.spec.tsx
```diff
@@ -1,24 +1,18 @@
 import * as React from "react";
```
Parsium, line 1: **issue**: since we have updated the spec for `useCharacerCount` like so:

```ts
const countValue = 10;
const countLimit = 20;
const [CharacterCount, visuallyHiddenHintId] = useCharacterCount(countValue, countLimit); 
```

these tests won't be functioning as intended, as on line 15, `maxLength` will actually be the `CharacterCount` component, and `characterCount` will actually be the id of the visually hidden hint text.

tomdavies73: Yeah this is a brilliant spot mate, obviously the tests were still passing but that's down to sheer luck on our end haha

DipperTheDan (approved):

Parsium (dismissed):

DipperTheDan (approved):

Parsium (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #6237, opened 2023-08-08
fix: set rootDir to src in tsconfig

robinzigmond:
The `build:types` script ran as part of the release process is broken as it outputs the `.d.ts` files inside `src` directories in the output folders. On investigation, this has happened because of the addition of `.ts` files inside the new `playwright` folder which are not excluded from the ts build and which cause tsc to default the `rootDir` compiler option to the root directory, rather than the `src` folder as it needs to be. To fix this, `rootDir` has been explicitly set to `./src/` in the build config, as well as the playwright ts files being excluded (without that tsc throws errors about trying to build files outside of `rootDir`). To avoid breaking the local `type-check` script which still needs to typecheck the playwright ts files, the new config has only been added inside the `tsconfig-build.json`.

### Proposed behaviour

`npm run precompile` should output `.d.ts` files alongside the corresponding `.js` files in the appropriate subdirectories of the output folders (`/lib/` and `/esm/`), as can be seen from the file structure of a working NPM release https://www.npmjs.com/package/carbon-react/v/119.10.0?activeTab=code . There should also be no `playwright` folder in the build output.

### Current behaviour

There is a separate `src` directory inside each output folder, and this contains just the `.d.ts` files (in normal folder structure). The `.js` files are still in the normal place, but TS intellisense (etc) is broken for consumers because there are no type declaration files alongside them (see broken structure here https://www.npmjs.com/package/carbon-react/v/119.10.1?activeTab=code).

A `playwright` folder is also found in the top level of these output directories.

### Checklist

#### QA

### Testing instructions

Run `npm run precompile` from the root directory (after first deleting the local `lib` and `esm` folders, if present), and ensure the build succeeds with no errors, and the correct directory structure is found in the two output directories.

Also try introducing TS errors into the playwright ts files (esp in the `playwright` directory) and ensuring these are picked up by `npm run type-check`, as a sanity check that local type-checking is not excluding files it shouldn't be. (Feel free to do any more tests along these lines that you can think of!)

### DIFF
--- a/tsconfig-build.json
+++ b/tsconfig-build.json
@@ -1,4 +1,7 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*"]
+  "compilerOptions": {
+    "rootDir": "./src/",
+  },
+  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*", "./src/**/*.pw.ts*", "./src/**/*.test-pw.ts*"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,5 +22,5 @@
     "target": "es6"
   },
   "include": ["./src/global.d.ts", "./src/**/*.ts*"],
-  "exclude": ["./node_modules", "./cypress", "./cypress.config.ts"]
+  "exclude": ["./node_modules", "./cypress", "./cypress.config.ts", "./playwright", "./playwright-ct.config.ts"]
 }


tsconfig-build.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*"]
+  "compilerOptions": {
+    "rootDir": "./src/",
+  },
+  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*", "./src/**/*.pw.ts*", "./playwright", ]
```
DlgSHi, line 6:
```suggestion
  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*", "./src/**/*.pw.ts*", "./src/**/*.test-pw.ts*", "./playwright"]
```

edleeks87, line 6: comment: additional whitespace can be removed

tsconfig.json
```diff
@@ -22,5 +22,5 @@
     "target": "es6"
   },
   "include": ["./src/global.d.ts", "./src/**/*.ts*"],
-  "exclude": ["./node_modules", "./cypress", "./cypress.config.ts"]
+  "exclude": ["./node_modules", "./cypress", "./cypress.config.ts", "./playwright-ct.config.ts"]
```
DlgSHi, line 25:
```suggestion
  "exclude": ["./node_modules", "./cypress", "./cypress.config.ts", "./playwright", "./playwright-ct.config.ts"]
```

DlgSHi (changes requested):
I think we need also add that excludes for `*.test-pw.tsx` and `*.pw.tsx` for `build.js` file

tsconfig.json
robinzigmond: It looks like you're right, actually. I deliberately added `./playwright` and `./src/**/*.test-pw.ts*"` to only the `tsconfig-build` file because I thought with them excluded from tsconfig they wouldn't be type-checked when we do `npm run type-check`, but I've just tested it and it looks like they still are (which is a bit surprising to me, actually).

tsconfig-build.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["src/**/*.stories.*", "src/**/*.spec.*"]
+  "compilerOptions": {
+    "rootDir": "./src/",
```
edleeks87, line 4: comment (non-blocking): I don't think we need the trailing `/` here but it's not critical so happy for you to leave

DipperTheDan (commented):
Nothing more to add from me. Just address what has been raised by @edleeks87 & @DlgSHi 👍🏻

tsconfig-build.json
robinzigmond: I never really know if it's needed but it's normally to specify the trailing `/` to make clear that something's a directory, so imo it does no harm and is potentially beneficial - unless you know a reason it's better not to have it?

tbh I think I put it there mainly beacuse the `outDir` already has a trailing slash in our tsconfig and I assumed they would use the same convention 🤷

edleeks87 (approved):

DipperTheDan (approved):

DlgSHi (approved):

PR #6239, opened 2023-08-08
Playwright events

DlgSHi:
### Proposed behaviour

- add reporting statuses back to github
Screenshot 2023-08-09 at 15 12 16

- add helper for events 

### Current behaviour

- no reporting statuses back to github
- no helper for events 

### Checklist

### Testing instructions

- [ ] check if all tests still pass
- [ ] check if the `Playwright tests workflow completed successfully.` appears in checks below the PR
- [ ] check the newly added helper functions for events

### DIFF
--- a/playwright/support/helper.ts
+++ b/playwright/support/helper.ts
@@ -91,3 +91,34 @@ export const checkDialogIsNotInDOM = async (page: Page) => {
   await checkElementIsNotInDOM(page, OPEN_MODAL);
   await checkElementIsInDOM(page, CLOSED_MODAL);
 };
+
+/**
+ * Asserts if an element has event was calledOnce
+ * @param callbackData an array with callback data
+ * @param eventName {string} event name
+ * @example await expectEventWasCalledOnce(messages, "onClick");
+ */
+export const expectEventWasCalledOnce = async (
+  callbackData: string[],
+  eventName: string
+) => {
+  const count = JSON.stringify(callbackData.length);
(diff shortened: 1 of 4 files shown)


edleeks87 (dismissed):

playwright/support/helper.ts
```diff
@@ -66,3 +66,34 @@ export const checkGoldenOutline = async (
...
 };
+
+/**
+ * Asserts if an element has event was calledOnce
+ * @param callbackData an array with callback data
+ * @param eventName {string} event name
+ * @example await expectEventWasCalledOnce(messages, "onClick");
+ */
+export const expectEventWasCalledOnce = async (
+  callbackData: string[],
```
edleeks87, line 77: We can sort out the types for these utils at a later date imo. I wasn't able to find any way of pulling in `SyntheticBaseEvent` which seems to be part of the internals of playwright but we could do something like the following which would allow us to remove the ts ignores:

```
export interface Foo extends SyntheticEvent<HTMLElement> {
  _reactName: string;
}
```

DlgSHi: we don't need to do that at least here, we can specify which type of event we are passing inside the `onClick`/`onBlur` etc. function. But then we need to somehow be able to pass it and `.push` it to the array

DipperTheDan (approved):

edleeks87 (approved):

DipperTheDan (approved):

stephenogorman (approved):

sianford (approved):

PR #6242, opened 2023-08-09
console fixes for unit tests

robinzigmond:
### Proposed behaviour

There should be no errors or warnings in the console when running our unit test suite with `npm run test`.

(This PR does not remove every single one - see detailed notes below - but should remove the vast majority. In particular it removes any which were due to incorrect setup of the tests themselves - those remaining are due to actual issues in components which have been logged separately, again see the details below.)

### Current behaviour

When running the unit test suite the console is flooded with errors and warnings, even as the tests pass with 100% coverage - making it very hard to pick out any useful information (particularly if there are failing tests which you are trying to understand).

### Checklist

#### QA

### Testing instructions

Since unfortunately a large number of console errors/warnings still remain - more than fit in a single terminal window in VS Code - it is unfortunately not practical to just run `npm run test` and check that only specific ones remain. Instead it is easiest to identify those test suites that do have remaining issues and filter those out.

The errors/warnings that remain fall into 4 separate categories - all have been logged to work on separately as they arise from actual errors

A) a warning "uncontrolled behavioiur in textbox is deprecated" coming from use of Select. Most such deprecation warnings have been removed in this PR by changing to controlled usage, but there is a particular issue with Select where the component itself renders an uncontrolled Textbox, even if the Select is controlled. This is logged as FE-5988.

B) "React does not recognise the placeholderTabIndex prop on a DOM element". This comes from an underlying issue on (I believe) the Link component which passes through this prop to a DOM element when it should not, and affects many unit tests in components that render Link. Logged as FE-5989.

C) In MultiSelect, there is a warning "received `true` for a non-boolean attribute `fill`" which I believe comes from incorrect prop-spreading in the Option component (but alternative possible fixes should be considered). Logged as FE-6039.

D) In MenuFullScreen, there is a warning that "each child in a list should have a unique key prop", because this has been omitted in the component. Logged as FE-6075.

In order to filter out certain tests, the easiest way is to go into "watch mode". From the terminal, first run `npm run test -- --watch`. Press p to filter filepaths by a regular expression, and enter the following, to run all component test suites except for those in the 7 components affected by the above issues: `src/components/(?!form/)(?!heading/)(?!link/)(?!menu/)(?!pager/)(?!pages/)(?!select/)`. This should run 134 test suites in total and generate no unwanted console output.

Then do the same again runnning all tests in the `src` folder but outside `components` (eg in the `style` and `__internal__` folders), using the following regex: `src/(?!components)`. This runs another 53 test suites and again there should be no console output generated.

There are 2 test suites outside `src`, in the `scripts` folder - use the regex `scripts` to run these. (Note that this and the following runs can be done from the normal command-line, outside watch mode, if you find it more comfortable - eg `npm run test -- scripts/` for this one.)

Finally use the following regular expressions to run just the test suites in each of the affected components originally, and verify that the only console errors/warnings are those due to the specific problems A-D identified above. (Exactly which ones occur is noted afterwards, as well as how many test suites are ran in total.)

- `src/components/form` (A) (1 test suite)
- `src/components/heading` (B) (1 test suite)
- `src/components/link` (B) (1 test suite)
- `src/components/menu` (B, D) (8 test suites)
- `src/components/pager` (A, B) (2 test suites)
- `src/components/pages` (B) (2 test suites)
- `src/components/select` (C) (17 suites)

As a final sanity check, notice that the total number of different test suites ran through all of the above is 134 + 53 + 2 + 1 + 1 + 1 + 8 + 2 + 2 + 17, which comes to 221 - and 221 is the total number of unit test suites in the project (as can be verified by running a global `npm run test`).

Note for QA: I do not believe any formal QA is needed on this PR, as the only files actually changed are unit test files. Github is still flagging a Carbon/QA review as required - I think this is because the `button-toggle-group.component.tsx` file has technically been edited, although that was only to add `istanbul ignore` comments (to cover some edge-case code that really doesn't need to be covered, and was only being covered previously due to an accident which also happened to cause a console warning to be generated). Please do feel free to look through the file changes, and approve if you are satisfied that no component behaviour can have been affected by this PR.


tomdavies73 (commented):
**suggestion**: (non-blocking) Looks good to me, obviously I still need to go through the testing instructions tomorrow and have a good play around, but every commit and the changes look good to me, I've spent a good minute or two on each commit as well. 

Only thing I'm wondering is if we need this many commits. For example, there are three separate commits for the `Tabs` component. Combining these into one commit might be a good idea. Alternatively, you could create one separate commit for one type of fix, perhaps one commit for "remove console warnings", "mock.logger deprecate", "fix uncontrolled deprecation messages" etc. Granted there may be duplicated components in these commits, and a fair few files, but it would reduce the amount of commits in total, and may even make testing easier.

However, if this is going to be too much work or inpractical, I'm more than happy with the current approach commit-wise 👍.

DipperTheDan (commented):
Nothing more to add from me. Let me know if you decide to implement commit changes requested by @tomdavies73. If not, I'm still happy to approve.

DipperTheDan (approved):

tomdavies73 (dismissed):

DipperTheDan (approved):

divyajindel (approved):

PR #6246, opened 2023-08-09
feat(pod): allow title and subtitle to be passed as a node

### LINKED ISSUE #4607: Accessibility Pod Title prop always shown H4
### Desired behaviour

Be able to use a similar prop as ["as" present in typography](https://carbon.sage.com/iframe.html?id=typography--variants&viewMode=docs#typography-1) to modify the rendered title element to solve Accessibility issues coming from SBS-24478
### Current behaviour

Always returning an H4 element besides his internal rendered node which may be even an H3, H2.

### Suggested Solution

_No response_

https://carbon.sage.com/iframe.html?id=pod--default-story&viewMode=docs#pod-1

SBS-24478

### Confidentiality

tomdavies73:
fix #4607

### Proposed behaviour

This PR allows `title` and `subtitle` props to be passed as a node, providing consumers with the ability
to pass components such as `Typography` to control what element is rendered and its appearance.
Also, conditional rendering has been used to ensure when `title` is passed as a node is not wrapped
in any other section headings.

Screenshot 2023-08-10 at 11 44 22

### Current behaviour

Currently, every title within the component is rendered as a `h4` section heading. We allow the `title` prop to be passed as a string or as a node, however whatever is passed is still rendered as a `h4`.

### Checklist

#### QA

### Testing instructions


src/components/pod/pod.component.tsx
```diff
@@ -39,9 +39,9 @@ export interface PodProps extends MarginProps {
   /** Prop to apply a theme to the Pod */
   variant?: PodVariant;
   /** Title for the pod h4 element always shown */
```
nineteen88, line 41: suggestion: Comment needs to be updated

src/components/pod/pod.component.tsx
```diff
@@ -39,9 +39,9 @@ export interface PodProps extends MarginProps {
...
-  title?: string | React.ReactNode;
+  title?: React.ReactNode;
```
nineteen88, line 42: suggestion: Potential side effect that something other than a JSX element or string is passed

This will have existed before but React.ReactNode type will also allow unsupported types through like `number`. We could purely check for `string | React.ReactElement` or potentially allow them through and handle them in the code? We can discuss or see what the next reviewer suggests?

src/components/pod/pod.component.tsx
```diff
@@ -39,9 +39,9 @@ export interface PodProps extends MarginProps {
...
   /** Optional subtitle for the pod */
-  subtitle?: string;
+  subtitle?: React.ReactNode;
```
nineteen88, line 44: suggestion: Potential side effect that something other than a JSX element or string is passed

React.ReactNode type will also allow unsupported types through like `number`. We could purely check for `string | React.ReactElement` or potentially allow them through and handle them in the code? We can discuss or see what the next reviewer suggests?

src/components/pod/pod.component.tsx
edleeks87: A fair point, I think we'd need to treat this as a breaking change if we did go with this approach. I'm inclined to say leave it as @tomdavies73 has it currently as it won't crash if anything other than a string of node is passed etc

edleeks87 (commented):
nothing else from me

src/components/pod/pod.component.tsx
nineteen88: Yep, happy to leave it.

nineteen88 (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6280, opened 2023-08-29
feat(definition-list): allow children to be wrapped FE-6179

### LINKED ISSUE #6216: Using a Box component with DL results in content disappearing
### Current behaviour

I am using a Definition List and need some sections to be a different background colour. When wrapping the `dt` and`dd` in a `Box` the content of the `dt` and`dd` disappears. 

### Expected behaviour

Wrap `dt` and`dd`  in a Box to change the background colour or add a new prop to `dt` and`dd`  ensure background colour can be changed. 

https://codesandbox.io/s/nice-lederberg-hkqmky?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

119.6.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #6243: Unable to compose a `Dl` with exported components
### Current behaviour

When we use an exported component within a `Dl` it doesn't render.

### Expected behaviour

Be able to use exported components to compose a `Dl`

https://codesandbox.io/s/summary-with-dl-forked-j7kylw?file=/src/App.tsx

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

119.4.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #6216, fix #6243

### Proposed behaviour

Removes children iteration which was constraining the ability to wrap children in custom components.
Adds a test story for chromatic to capture any regressions to supporting using `Box` as a children of `Dl`

### Current behaviour

Iterates over children and checks the type of them to compose an array, which prevents implementing teams from wrapping them in any Carbon or custom component (or additional valid HTML elements)

### Checklist

#### QA

### Testing instructions


src/components/definition-list/definition-list-test.stories.tsx
DipperTheDan, line 1: **issue: (non-blocking)** you've got quite a few declared but unused props in the `DefaultStory`

DipperTheDan:
**question:** Just had a quick look at the codesandboxes and [objective-visvesvaraya-lwdd23](https://codesandbox.io/s/objective-visvesvaraya-lwdd23) looks like it not longer stays within the bounds of the `Tile`. Should this be the case with the changes you've made?

Updated CSB
Screenshot 2023-08-30 at 14 32 06

Original CSB
Screenshot 2023-08-30 at 14 33 39

src/components/definition-list/definition-list-test.stories.tsx
edleeks87: Fair point, I didn't modify that story but it's no effort to fix the lint warnings 👍

src/components/definition-list/definition-list-test.stories.tsx
```diff
@@ -89,3 +90,32 @@ export const DLBoxComponent = () => {
     </div>
   );
 };
+
+export const UsingBoxToOverrideBackgroundColor = () => (
```
robinzigmond, line 94: **comment**: although you've added this story, it appears to have no use right now as it's not being used in any Cypress tests nor is it actually displayed in storybook - which means, contrary to what you said in the PR description about why you added this, Chromatic won't run on it.

You'll need to add this to the `includeStories` array to get any benefit, unless I've missed something.

cypress/components/definition-list/definition-list.cy.tsx
```diff
@@ -44,9 +44,9 @@ context("Testing Definition List component", () => {
     );
 
     it.each([
-      [135, 1229, 10],
-      [683, 683, 50],
-      [1229, 135, 90],
+      [113, 1229, 10],
```
robinzigmond, line 47: **question**: I'm sure you know what you're doing, but it's not obvious to me why the pixel widths of the elements should have changed (by over 20 pixels!) in a fix that surely shouldn't introduce any functional or design changes?

src/components/definition-list/definition-list.stories.tsx
```diff
@@ -0,0 +1,189 @@
...
+        <Box mr={1}>Details example</Box>
+        <Icon type="tick" />
+      </Box>
+    </Dd>
+    <Dt>Third</Dt>
+    <Dd>Description</Dd>
+  </Dl>
+);
+
+export const MultipleSingleColumnsWithSegments = () => (
```
robinzigmond, line 82: **comment**: you should add `MultipleSingleColumnsWithSegments.parameters = { chromatic: { disableSnapshot: true } }` because the previous MDX version had this in the Story props.

src/components/definition-list/definition-list.stories.tsx
robinzigmond, line 1: **comment** (non-blocking): I thought we'd agreed ages back that any conversions of stories from MDX to TSX done as part other work on the component should be done in a separate commit. (To be fair I'm surprised to see there were any components we hadn't done this on yet!)

src/components/definition-list/definition-list.style.ts
```diff
@@ -39,49 +29,42 @@ StyledDl.defaultProps = {
...
+  Pick<DlProps, "asSingleColumn" | "dtTextAlign">
+>`
   ${space}
   font-size: var(--fontSizes100);
   font-weight: 700;
   color: var(--colorsUtilityYin090);
+  ${({ asSingleColumn }) =>
+    asSingleColumn &&
+    css`
+      grid-column: 1;
```
robinzigmond, line 42: **comment**: I'm a bit confused by this `grid-column` property - both here and on line 63. They're only applied when `asSingleColumn` is true, but that's precisely the case (line 13 of this same file) that `display: grid;` is *not* applies, which makes it do nothing. See screenshot from CSS inspector:
image

I'm not sure why any of these changes are necessary - things look visually OK to me (although I've only done a quick side-by-side comparison with master) even with this clearly not working, and since we're not trying to produce visual changes I don't see the need for any new CSS (the only change to styles in this PR might be where various props come from).

cypress/components/definition-list/definition-list.cy.tsx
robinzigmond, line 1: **question**: I'd feel more comfortable if we had added some tests (whether browser tests or unit tests I don't mind, but browser tests are likely better/easier which is why I'm putting this comment here) for the children-wrapped-in-Box scenario. I think this is a better means of guarding against regressions in this case than just relying on Chromatic (although I don't mind doing that as well).

src/components/definition-list/definition-list.style.ts
edleeks87: Yep I have the logic inverted 🤦 they should be applied when the grid layout is set not when it is a single column

src/components/definition-list/definition-list.stories.tsx
edleeks87: I wasn't aware of that, I'll separate them 👍

cypress/components/definition-list/definition-list.cy.tsx
edleeks87: I'll see what I can come up with but I'm not convinced we can test this very well beyond testing the background colour of a given `Box` inside a `Dl`, I don't particularly like the idea of arbitrarily testing the positions of the elements etc

cypress/components/definition-list/definition-list.cy.tsx
edleeks87: Fair point to raise, previously each element was wrapped in div so the width included the 24px of right padding. With these divs removed it isn't including that padding anymore. I've modified the test slightly to set the viewport width to 1000px and then added a constant for the padding that is subtracted in the assertion so hopefully it's a little clearer now what's going on

cypress/components/definition-list/definition-list.cy.tsx
edleeks87: I've just asserted that the children wrapped in Boxes are now visible

robinzigmond (approved):

DipperTheDan (dismissed):

DipperTheDan (dismissed):

robinzigmond (approved):

cypress/components/definition-list/definition-list.cy.tsx
```diff
@@ -22,9 +23,18 @@ const specialCharacters = [
 ];
 
 const alignValue = ["left", "center", "right"];
+const PADDING_RIGHT = 24;
+const VIEWPORT_WIDTH = 1000;
 
 context("Testing Definition List component", () => {
   describe("should render Definition List component", () => {
+    it("with grid display layout when asSingleColumn prop is not set", () => {
```
ZhuoyuJin, line 31: **non-blocking**: Can we rename this test to something like "should render definition list with grid display layout when asSingleColumn prop is not set", but definitely non-blocking comment as we are going to refactor cypress tests to playwright anyway, so I'm happy if it keeps like this for now.

edleeks87: I'm not convinced that it would improve what's outputted in the console tbh but it's partly down to how the rest of the descriptions are 

How it looks currently
```
Testing Definition List component
  should render Definition List Component
   with grid display layout when asSingleColumn prop is not set
```

How it would be if I make this change
```
Testing Definition List component
  should render Definition List Component
   should render definition list with grid display layout when asSingleColumn prop is not set
```

edleeks87: I've made some small tweaks to the wording, I don't really want to go too far beyond that as it's out of scope to update all test descriptions here etc

DipperTheDan (approved):

robinzigmond (approved):

ZhuoyuJin (approved):

PR #6291, opened 2023-08-31
fix(flat-table): ensure tabindex is set on clickable elements when table has pager and loading state FE-6194

edleeks87:
### Proposed behaviour

Updates component to handle loading states and pagination to ensure tab stop is put on correct row.
Ensure tab index is placed on correct cell when first column is expandable and is in selected or highlighted row

### Current behaviour

Tab index is not updated on first/ highlighted/ selected row when there is a loading state and pagination is used
When first column is expandable and part of the a highlighted/ selected row the tab index is set incorrectly

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/rough-paper-nmdxzk?file=/src/App.js
(may be easier to run locally so you can trigger the loading state by refreshing page etc)

Ensure first row/ cell is focused when tabbing or that the the row/ cell in the highlighted/ selected row is when tabbing. Ensure row/ cell can be focused after pagination via tabbing etc

### DIFF
--- a/cypress/components/flat-table/flat-table.cy.tsx
+++ b/cypress/components/flat-table/flat-table.cy.tsx
@@ -2274,6 +2274,88 @@ context("Tests for Flat Table component", () => {
       flatTableBodyRowByPosition(5).should("be.focused");
     });
 
+    it("should set tabIndex 0 on the first row when after the loading state has finished", () => {
+      CypressMountWithProviders(<stories.KeyboardNavigationWithPagination />);
+      cy.wait(300);
+
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).then(checkNewFocusStyling);
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).should("not.be.focused");
+      flatTableBodyRowByPosition(1).should("not.be.focused");
+      flatTableBodyRowByPosition(2).should("not.be.focused");
+      flatTableBodyRowByPosition(3).should("not.be.focused");
+      flatTablePageSelectNext().click();
+      cy.wait(300);
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).then(checkNewFocusStyling);
+    });
+
+    it("should set the tabIndex on the highlighted row when the loading state has finished", () => {
+      CypressMountWithProviders(
+        <stories.KeyboardNavigationWithPagination highlighted />
+      );
+      cy.wait(300);
+
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).should("not.be.focused");
+      flatTableBodyRowByPosition(2).then(checkNewFocusStyling);
+      cy.focused().tab();
+      flatTableBodyRowByPosition(3).should("not.be.focused");
+      flatTablePageSelectNext().click();
+      cy.wait(300);
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).then(checkNewFocusStyling);
+    });
+
+    it("should set the tabIndex on the highlighted row when the loading state has finished and remove it when row is no longer highlighted", () => {
+      CypressMountWithProviders(
+        <stories.HighlightedRowWithLoadingState expandableArea="wholeRow" />
+      );
+      cy.wait(300);
+
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).should("not.be.focused");
+      flatTableBodyRowByPosition(2).should("be.focused");
+      cy.focused().tab();
+      flatTableBodyRowByPosition(3).should("not.be.focused");
+      flatTableBodyRowByPosition(2).click();
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).should("be.focused");
+    });
+
+    it("should set the tabIndex on the first cell in a highlighted row when the loading state has finished and remove it when row is no longer highlighted", () => {
+      CypressMountWithProviders(
+        <stories.HighlightedRowWithLoadingState expandableArea="firstColumn" />
+      );
+      cy.wait(300);
+
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableCell(0).should("not.be.focused");
+      flatTableCell(8).should("be.focused");
+      cy.focused().tab();
+      flatTableCell(12).should("not.be.focused");
+      flatTableCell(8).click();
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableCell(0).should("be.focused");
+    });
+
     it("should render Flat Table with action popover in a cell opened by mouse", () => {
       CypressMountWithProviders(
         <stories.FlatTableAllSubrowSelectableComponent />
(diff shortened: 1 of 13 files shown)


cypress/components/flat-table/flat-table.cy.tsx
```diff
@@ -2274,6 +2274,47 @@ context("Tests for Flat Table component", () => {
...
+    it("should set tabIndex 0 on the first row when after the loading state has finished", () => {
+      CypressMountWithProviders(<stories.KeyboardNavigationWithPagination />);
+      cy.wait(300);
+
+      cy.get("body").tab();
+
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).then(checkNewFocusStyling);
+      cy.focused().tab();
+      flatTableBodyRowByPosition(0).should("not.be.focused");
```
robinzigmond, line 2286: **question** (non-blocking): I'm a bit surprised we're explicitly checking rows 0, 1 and 3 for not being focused, but not the other two rows? (This after tabbing off the focused row, so  Seems a bit arbitrary to me.

src/components/flat-table/__internal__/use-calculate-sticky-cells.ts
```diff
@@ -28,5 +31,8 @@ export default (id: string) => {
     onKeyDown,
     isFirstCell,
     isExpandableCell,
+    highlighted,
```
robinzigmond, line 34: **comment**: I understand why you've added these props here, but it seems to me that this hook is now about much more than just "sticky cells". Maybe rename it to something more generic like `useTableCell`? (But there's likely something much better than that.)

src/components/flat-table/flat-table-cell/flat-table-cell.component.tsx
```diff
@@ -59,8 +59,17 @@ export const FlatTableCell = ({
...
     isExpandableCell,
     makeCellSticky,
+    highlighted,
+    selected,
+    setSelectedId,
   } = useCalculateStickyCells(internalId.current);
 
+  useEffect(() => {
+    if (isExpandableCell && (highlighted || selected)) {
+      setSelectedId?.(internalId.current);
```
robinzigmond, line 69: **comment**: I don't know for sure (it's not trivial to debug as the story you've added for this doesn't precisely replicate the bug scenario), but I suspect this *might* be where the bug I've noted in the overall review comment is coming from. I think if this `if` fails, we need to set the selected ID to something else (ideally to `null` or `undefined` becase surely nothing is selected at that point, although the types don't currently allow that...)

src/components/flat-table/flat-table-row-header/flat-table-row-header.component.tsx
```diff
@@ -71,8 +71,17 @@ export const FlatTableRowHeader = ({
...
     isFirstCell,
     isExpandableCell,
+    highlighted,
+    selected,
+    setSelectedId,
   } = useCalculateStickyCells(internalId.current);
 
+  useEffect(() => {
+    if (isExpandableCell && (highlighted || selected)) {
+      setSelectedId?.(internalId.current);
```
robinzigmond, line 81: **comment**: could be a similar thing going wrong here to what I've called out in the `flat-table-cell` component. Maybe.

src/components/flat-table/flat-table.component.tsx
```diff
@@ -203,12 +203,23 @@ export const FlatTable = ({
...
     const findSelectedId = () => {
-      const firstfocusableElement = tableRef.current?.querySelector(
-        FOCUSABLE_ROW_AND_CELL_QUERY
+      const focusableElements = Array.from(
+        tableRef.current?.querySelectorAll(FOCUSABLE_ROW_AND_CELL_QUERY) ||
+          /* istanbul ignore next */ []
       );
 
+      if (
+        !focusableElements.length ||
```
robinzigmond, line 212: **suggestion**: I don't think you need the `length` check, empty arrays will never pass a `.some` check (they don't have any elements at all so can't have any elements that pass a particular test). See here from the browser console even with an "autopass" check:

image

robinzigmond (commented):
There's unfortunately a bug even in the fixed version of the codesandbox.

If you click the "set highlighted" button, so that the 3rd row gets `tabindex={0}` this tabindex seems to remain even after removing the highlighting, meaning that row is always the one that gets tab focus until you refresh the sandbox.

DipperTheDan (commented):
Nothing more to add from me. I think @robinzigmond has commented everything already.

src/components/flat-table/flat-table-cell/flat-table-cell.component.tsx
edleeks87: Correct we need to unset the id when it's no longer highlighted

src/components/flat-table/__internal__/use-calculate-sticky-cells.ts
edleeks87: valid point I'll have a little think about it

src/components/flat-table/flat-table.component.tsx
```diff
@@ -201,35 +199,23 @@ export const FlatTable = ({
...
-      if (firstfocusableElement && !selectedId) {
-        const currentlySelectedId = firstfocusableElement?.getAttribute("id");
-
-        /* istanbul ignore else */
-        if (currentlySelectedId && selectedId !== currentlySelectedId) {
-          setSelectedId(currentlySelectedId);
-        }
-      }
-    };
+  const getTabStopElementId = () => {
```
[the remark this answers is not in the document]
edleeks87: Good point, I tried to useCallback and couldn't get it working but that was before I simplified some of it further so I'll give it a go again and see if it works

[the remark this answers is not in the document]
edleeks87: Yeah it still doesn't work as it uses stale values. I've stored it in a ref to avoid adding it as a dependency of the useEffect insted

src/components/flat-table/flat-table-cell/flat-table-cell.component.tsx
```diff
@@ -83,6 +86,8 @@ export const FlatTableCell = ({
       expandable={expandable}
       {...(colspan !== undefined && { colSpan: Number(colspan) })}
       {...(rowspan !== undefined && { rowSpan: Number(rowspan) })}
+      data-selected={selected && isExpandableCell ? true : undefined}
```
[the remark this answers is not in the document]
edleeks87: Yeah my intention was to avoid having it have false but I'm not precious tbh so happy to update

DipperTheDan (commented):
I have nothing more to contribute to this than what @robinzigmond has already said.

src/components/flat-table/__internal__/use-table-cell.ts
```diff
@@ -19,6 +26,16 @@ export default (id: string) => {
   const isFirstCell = id === firstCellId;
   const isExpandableCell = expandable && isFirstCell && firstColumnExpandable;
 
+  useEffect(() => {
+    if (tabstopTimer.current) {
+      clearTimeout(tabstopTimer.current);
+    }
+
+    tabstopTimer.current = setTimeout(() => {
```
DipperTheDan, line 34: **suggestion(not blocking):** Reckon it's worth putting a comment here explaining why the `setTimeout` is needed here? Might be helpful to future maintainers when we're long gone and retired.

robinzigmond (approved):

DipperTheDan (approved):

stephenogorman (approved):

PR #6300, opened 2023-09-06
test(note): refactor cypress to playwright

sianford:
### Proposed behaviour

Refactor Note tests from Cypress to Playwright

### Current behaviour

Tests are currently Cypress

### Checklist

#### QA

### Testing instructions

- [ ] Run Note tests in Playwright Test Runner to check if the note.pw.tsx file passed
- [ ] Run all other tests in Playwright Test Runner to check none of the other *.pw.tsx files have regressed


divyajindel (changes requested):

divyajindel (approved):

src/components/note/note.pw.tsx
```diff
@@ -0,0 +1,252 @@
...
+    mount,
+    page,
+  }) => {
+    await mount(<NoteComponentWithPreviews />);
+
+    await checkAccessibility(page);
+  });
+});
+
+test("should render with expected border radius styling", async ({
```
ZhuoyuJin, line 245: non-blocking: think it could look better if this test could under something like test.describe("check expected border radius styling")

sianford: I did wonder about this, as did Divya. It was out on its own in the original Cypress file, presumably because there's not much benefit in having another describe for one test. If there was another styling type test then I'd have put it in a describe. I'll ask in our QA channel for others' opinions on this. We should have a standard.

src/components/note/components.test-pw.tsx
```diff
@@ -0,0 +1,92 @@
...
+      />
+    </div>
+  );
+};
+
+export {
+  NoteComponent,
+  NoteComponentWithInlineControl,
+  NoteComponentWithPreviews,
+};
```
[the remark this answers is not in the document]
sianford: You're are correct. Removed :)

src/components/note/note.pw.tsx
sianford: Following discussion in the Slack channel, I've decided to leave that test outside of a `describe` as there are no others like it to group it with. I would definitely have put it into a `describe` if there was another radius test case or some other styling test case to group it with though.

ZhuoyuJin (approved):

src/components/note/note.pw.tsx
```diff
@@ -0,0 +1,252 @@
...
+      page,
+    }) => {
+      await mount(<NoteComponent name={name} />);
+
+      await expect(noteFooterCreatedBy(page)).toHaveText(name);
+    });
+  });
+
+  test("should render Note with createdDate prop", async ({ mount, page }) => {
+    const createdDate = "25 June 2022, 11:57 AM" as string;
```
Parsium, line 95: **suggestion(non-blocking)**: You shouldn't need the type assertion here, as TypeScript will infer `createdDate` as a string constant automatically for you.

```suggestion
    const createdDate = "25 June 2022, 11:57 AM";
```

src/components/note/note.pw.tsx
```diff
@@ -0,0 +1,252 @@
...
+            text: value,
+            timeStamp: `${CHARACTERS.STANDARD}`,
+          }}
+        />
+      );
+
+      await noteStatus(page).hover();
+      await expect(noteStatus(page)).toHaveText(value);
+      await expect(tooltipPreview(page)).toHaveText(CHARACTERS.STANDARD);
+      await dlsRoot(page).hover({ position: { x: 0, y: 0 } });
```
Parsium, line 119: **suggestion**: As we are not asserting anything else after this action, we should be good to remove it.

```suggestion
```

src/components/note/note.pw.tsx
```diff
@@ -0,0 +1,252 @@
...
+      <NoteComponent
+        text="https://carbon.s"
+        onLinkAdded={() => {
+          hasOnLinkAddedBeenCalledCount += 1;
+        }}
+      />
+    );
+
+    await linkPreview(page).waitFor();
+    expect(hasOnLinkAddedBeenCalledCount).toBe(1);
```
Parsium, line 148: **praise**: I like the simplicity of this approach!

src/components/note/note.pw.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+      await checkAccessibility(page);
+    });
+  });
+
+  test("should render Note with createdDate prop for accessibility tests", async ({
+    mount,
+    page,
+  }) => {
+    const createdDate = "25 June 2022, 11:57 AM" as string;
```
Parsium, line 210: **suggestion(non-blocking)**: likewise here, the TypeScript compiler should automatically infer this as a string constant, so we don't necessarily need the type assertion. 

```suggestion
    const createdDate = "25 June 2022, 11:57 AM"
```

src/components/note/note.pw.tsx
sianford: After thinking about this, and the fact that the only action event test case is within its own `describe`, I've moved this to within its own `describe` reading to styling. I think we should adopt this pattern for other components. So thanks for mentioning it, and also to @divyajindel for querying the same :)

Parsium (dismissed):

divyajindel (approved):

Parsium (approved):

ZhuoyuJin (approved):

PR #6302, opened 2023-09-07
test(portrait): migrate cypress tests to playwright

tomdavies73:
### Proposed behaviour

The Portrait component should have in-browser tests using the Playwright framework

### Current behaviour

Portrait currently has Cypress tests instead of Playwright tests

### Checklist

#### QA

### Testing instructions


cypress/locators/portrait/index.js
```diff
@@ -1,4 +1,4 @@
-import { PORTRAIT_PREVIEW, PORTRAIT_INITIALS } from "./locators";
+import { PORTRAIT_PREVIEW, PORTRAIT_INITIALS } from "../../../playwright/components/portrait/locators";
```
robinzigmond, line 1: **comment** - I'm not sure what's going on here. Why do we have **cypress** locators that are importing stuff from the corresponding **playwright** locators?

As far as I can tell from a quick search, we don't seem to be importing the cypress portrait locators into any Cypress tests any more (but I might have missed something!) - if so then we can just delete this file and the `cypress/locators/portrait` folder. But if not then we should keep the `cypress/locators/portrait/locators.ts` file as well (just those locators that we need) and this file should continue to import what it needs from there.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+
+      await expect(portraitPreview(page).locator("span")).toHaveCSS(
+        "background-color",
+        color
+      );
+    });
+  });
+
+  testData.forEach((tooltipMessage) => {
+    test(`should render with toolTipMessage as ${tooltipMessage}`, async ({
```
divyajindel, line 163:
```suggestion
    test(`should render with tooltipMessage as ${tooltipMessage}`, async ({
```

src/components/portrait/portrait.config.ts
```diff
@@ -1,6 +1,19 @@
 export const PORTRAIT_SHAPES = ["circle", "square"];
-export const PORTRAIT_SIZES = ["XS", "S", "M", "ML", "L", "XL", "XXL"];
-export const PORTRAIT_SIZE_PARAMS = {
+type PortraitSize = "XS" | "S" | "M" | "ML" | "L" | "XL" | "XXL";
+export const PORTRAIT_SIZES: PortraitSize[] = [
```
robinzigmond, line 3: **comment** (non-blocking): I don't mind what you've done here - there's no harm in making explicit types for these, especially if we end up wanting to export them later. But you didn't actually need to do it to fix the type errors you get in your Playwright stuff if you remove them - just adding `as const` to both `PORTRAIT_SIZES` and `PORTRAIT_SIZE_PARAMS` is enough. I'm not saying this is better though, it's just an alternative.

divyajindel (changes requested):

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
+import React from "react";
+import { test, expect } from "@playwright/experimental-ct-react17";
+import { getDataElementByValue } from "../../../playwright/components";
+import {
+  portraitImage,
+  portraitInitials,
+  portraitPreview,
+} from "../../../playwright/components/portrait/index";
+
```
robinzigmond, line 9: **nitpick** (non-blocking) - I don't really care that much, but I'm not sure what the reason is for the blank lines here on lines 9 and 11. They don't seem to separate logical blocks of imports for example.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+import {
+  PORTRAIT_SIZE_PARAMS,
+  PORTRAIT_SIZES,
+} from "../../../src/components/portrait/portrait.config";
+import {
+  PortraitComponent,
+  PortraitDefaultComponent,
+} from "./components.test-pw";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
```
robinzigmond, line 29: **nitpick** (non-blocking): ironically here I would rather see some blank lines separating the definitions of these consts. Really just personal opinion as I think it's easier on the eye - feel free to ignore if you disagree. (These blank lines were there in the original Cypress version - I think my automatic conversion script strips them out so you can blame me!)

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+  size,
+  PORTRAIT_SIZE_PARAMS[size].dimensions,
+]);
+
+const visibilityValues: [boolean, string][] = [
+  [true, "with"],
+  [false, "without"],
+];
+
+test.describe("Portrait component", () => {
```
robinzigmond, line 53:
```suggestion
test.describe("check props for Portrait component", () => {
```

**suggestion** (non-blocking): This is how we've described this in other components, and also how it was done in Cypress. I don't think we can just use "Portrait component" because this isn't the only `test.describe` block, and all of them are obviously testing the portrait component.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+
+    await expect(portraitPreview(page)).toHaveCSS("height", "40px");
+    await expect(portraitPreview(page)).toHaveCSS("width", "40px");
+    await expect(portraitPreview(page).locator("img")).toHaveAttribute(
+      "src",
+      testImage
+    );
+  });
+
+  imageURL.forEach((urlVals) => {
```
robinzigmond, line 65:
```suggestion
  imageURL.forEach((url) => {
```

**suggestion**: each element of `imageURL` is just a url (or a string represetting one), so why not just call it that? The original Cypress test used `scrImage`, which I think is fine - and I don't even mind `urlVal` if you really insist. What I definitely don't want is to use a plural (as you have with `urlVals`), when each one is a single URL/string/image.

On a similar note I dislike `imageURL` for the array because there's more than one, I'd call it `imageURLs` myself (or similar). But I realise that in the Cypress test it also had this name, so while I would change it personally I don't feel as strongly about it.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+  ["brown", COLOR.BROWN],
+];
+const testImage =
+  "https://www.gravatar.com/avatar/cceff1ad774bf89b1f75cb6e5005333c?s=40&d=404";
+const imageURL = [
+  "https://avataaars.io/?avatarStyle=Transparent&topType=LongHairStraight&accessoriesType=Blank&hairColor=BrownDark&facialHairType=Blank&clotheType=BlazerShirt&eyeType=Default&eyebrowType=Default&mouthType=Default&skinColor=Light",
+  "https://www.gravatar.com/avatar/05c1c705ee45d7ae88b80b3a8866ddaa?s=24&d=404",
+];
+
+const portraitSizes = PORTRAIT_SIZES.map((size) => [
```
robinzigmond, line 43: **praise**: well done for extracting this as a const and showing clearly how it's constructed. The version we had in the Cypress test is truly gruesome to even look at 😂

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+  size,
+  PORTRAIT_SIZE_PARAMS[size].dimensions,
+]);
+
+const visibilityValues: [boolean, string][] = [
```
robinzigmond, line 48: **comment** (non-blocking): personally I don't think there's sufficient benefit in making this array its own const. I know it's used in 2 different tests, but 2 isn't so many and it isn't a big array - and this way the test can't be understood without going away from it to see what the definition is. I think it's much more readable to just pass this array literally into each of the `forEach`'s where it's used. But you don't have to do this if you disagree.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+      await portraitPreview(page).hover();
+
+      const tooltip = getDataElementByValue(page, "tooltip");
+
+      await expect(tooltip).toBeVisible();
+      await expect(tooltip).toHaveId(tooltipId);
+    });
+  });
+
+  visibilityValues.forEach(([boolVal, renderState]) => {
```
robinzigmond, line 196: **comment** (non-blocking): I see you've combined what was 2 separate Cypress tests here into one with a `forEach` to cover both cases. I'm not sure this is an improvement - there wasn't a great deal of repetition between the two tests, and the tests are in fact different (as we only check the text in the visible case). Personally I prefer them separate, but again I'm happy to leave up to you.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+  colors.forEach(([names, color]) => {
+    test(`should render with a tooltip with a ${names} background color`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<PortraitComponent tooltipBgColor={color} />);
+
+      const tooltip = getDataElementByValue(page, "tooltip");
+
+      await expect(tooltip).toBeVisible();
```
robinzigmond, line 268: **question** (non-blocking): I notice this visibility check wasn't in the equivalent Cypress tests, so while I don't object particularly, I'm wondering why you've introduced it here?

I'm actually a bit concerned because I thought the tooltip shouldn't be visible till you focus/hover it (and it shouldn't be possible to check the background color until it is!), yet this apparently all works here without doing that! Kind of makes me wonder what the value was in those tests further up where we explicitly hover before asserting the tooltip is visible.

Mind you I don't know the component very well, and if these are problems then they existed before your Playwright refactor 😄 Mostly I'm wondering why we need the visibility check here - the rest can be investigated and tickets raised if something needs fixing.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+  colors.forEach(([names, color]) => {
+    test(`should render with a tooltip with a ${names} font color`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<PortraitComponent tooltipFontColor={color} />);
+
+      const tooltip = getDataElementByValue(page, "tooltip");
+
+      await expect(tooltip).toBeVisible();
```
robinzigmond, line 282: **question** (non-blocking): same as in previous test

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+      await expect(tooltip).toHaveCSS("color", color);
+    });
+  });
+
+  test("should render and call onClick callback when a click event", async ({
```
robinzigmond, line 287: **comment** (non-blocking): I see you've removed a `describe` that was previously wrapping this test. I guess you decided it was pointless with just one test but having a separate "events" block made sense to me, and I'd have preferred to keep it. But honestly not a big deal.

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+    mount,
+    page,
+  }) => {
+    let callBackCount = 0;
```
robinzigmond, line 291:
```suggestion
    let callbackCount = 0;
```

**nitpick** (non-blocking) - "callback" is a single word (even though I guess it's originally made up of 2), maybe it's just me but camel-casing it as `callBack` looks really weird to me

src/components/portrait/portrait.pw.tsx
```diff
@@ -0,0 +1,486 @@
...
+    await mount(<PortraitDefaultComponent onClick={(callBackCount += 1)} />);
+
+    await portraitPreview(page).click();
+    expect(callBackCount).toBe(1);
+  });
+});
+
+test.describe("Accessibility tests for Portrait component", () => {
```
robinzigmond, line 299: **comment** (non-blocking): maybe it's not needed, but I feel like this is a good/legitimate case to add a TODO comment to remove the "color-contrast" exceptions when that eventually gets fixed

cypress/locators/portrait/index.js
tomdavies73: Good point, i'll just delete the file

src/components/portrait/portrait.config.ts
tomdavies73: Yeah fair, I think I'll still leave this in anyways, as I don't think there's any harm in adding some types, especially as they might be needed further down the line

src/components/portrait/portrait.pw.tsx
tomdavies73: Yeah fair point, I personally prefer using a `forEach` here as I know technically we're testing different things (being visible and not being visible), but I feel as though there is enough overlap elsewhere to justify it

src/components/portrait/portrait.pw.tsx
tomdavies73: Yeah, looks like these were just added to the colour related tests in error, I'll do some further tests on tests which require visibility tests and `.hover()` just to make sure there aren't any issues and the tests aren't redundant

tomdavies73: Turns out this passed because the pre-made `Portrait` component within `components.test-pw.stx` had `tooltipVisibility` set to true. Turns out this component was also used in other tests where we're using focus and checking visibility, so I'll change that

divyajindel (approved):

PR #6304, opened 2023-09-08
test(card): migrate tests from cypress to playwright - FE-6102

DipperTheDan:
### Proposed behaviour

Migrate tests from cypress to playwright

### Current behaviour

Tests are currently written in cypress.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/card/card.pw.tsx
```diff
@@ -0,0 +1,353 @@
...
+  footerCard,
+} from "../../../playwright/components/card/index";
+import { CHARACTERS, SIZE } from "../../../playwright/support/constants";
+import {
+  assertCssValueIsApproximately,
+  checkAccessibility,
+} from "../../../playwright/support/helper";
+import { Card, CardProps } from "../../../src/components/card";
+import {
+  DifferentCardFooterPadding,
```
sianford, line 18: Why are these all from lines 18-33 in upper camel case? I can see that they're like that in the Cypress tests but they should be lower camel case to follow our current standard. Can you change them please while you're in the files?

playwright/components/card/index.ts
```diff
@@ -0,0 +1,19 @@
...
+export const draggableCard = (page: Page, index: number) =>
+  page
+    .locator(`[data-element="draggable-card-${index}"]`)
+    .locator('[data-element="drag"]');
+
+export const draggableContainer = (page: Page, index: number) =>
+  page.locator(`[data-element="draggable-container-${index}"]`);
+
+export const columnCard = (page: Page) =>
+  page.locator(`[data-element="card-column"]`).locator("div:nth-child(2)");
```
sianford, line 16: Should normal single quotes be needed here rather than backticks?

playwright/components/card/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const footerCard = (page: Page) =>
+  page.locator(`[data-element="card-footer"]`);
```
sianford, line 19: Should normal single quotes be needed here rather than backticks?

src/components/card/card.pw.tsx
```diff
@@ -0,0 +1,353 @@
...
+import {
+  CardComponent,
+  CardTextAlignment,
+  DraggableExample,
+} from "../../../src/components/card/components.test-pw";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+const textAlignment = ["center", "left", "right"];
+
+test.describe("Should check Card component properties", () => {
```
sianford, line 39: Can you remove the "Should" from the `describe` level please? The `should` is generally just used at test case level.

src/components/card/card.pw.tsx
```diff
@@ -0,0 +1,353 @@
...
+    await cardElement.click({
+      force: true,
+    });
+
+    await cardElement.waitFor();
+    expect(setClickCounter).toEqual(1);
+  });
+});
+
+(["default", "large"] as CardProps["roundness"][]).forEach((roundness) => {
```
sianford, line 235: Can this test be moved to above the action event one above please? Or even better, can the styling tests be placed in their own `describe` and the action event in its own? I'm of the mind now that they should be in meaningful `describes. Also, can the "S" in "Should" be changed to lower case to follow the same standard please.

src/components/card/card.pw.tsx
```diff
@@ -0,0 +1,353 @@
...
+  });
+});
+
+test.describe("Accessibility tests for Card component", () => {
+  test("should pass accessibility tests for default example", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<CardComponent />);
+    await checkAccessibility(page);
```
sianford, line 263: I think there should be a space between the mount and the test for the accessibility tests to make them follow the same pattern as the other tests in the file and of other components.

src/components/card/card.pw.tsx
```diff
@@ -0,0 +1,353 @@
...
+    page,
+  }) => {
+    await mount(
+      <CardComponent
+        boxShadow="boxShadow400"
+        hoverBoxShadow="boxShadow200"
+        interactive
+      />
+    );
+    const cardElement = await card(page);
```
sianford, line 169: **suggestion: (non-blocking)** a space between the mount and the coasts, assertions etc would look better to follow the same standard as the other tests.

playwright/components/card/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { Page } from "@playwright/test";
+import CARD from "./locators";
+
+// component preview locators
+export const card = (page: Page) => page.locator(CARD).nth(0);
+
+export const draggableCard = (page: Page, index: number) =>
+  page
+    .locator(`[data-element="draggable-card-${index}"]`)
```
sianford, line 9: Is it possible get the full element locators out of the `index` file and into the `locators` file (will need new locators creating obv) so the `card.pw.tsx` file then calls a neater locator? Having the full element at this level makes the `card.pw.tsx` file look messy and not as easy to read. I know this is how the previous `card.cy.tsx` file was but it would be a good opportunity to tidy that up.

src/components/card/card.pw.tsx
DipperTheDan: These are custom components imported from the stories files/test-pw files. They need to pascal case. We do this in the other Playwright files too. [Here is Batch-Selection](https://github.com/Sage/carbon/blob/master/src/components/batch-selection/batch-selection.pw.tsx#L4) and the [custom component in the test-pw file here](https://github.com/Sage/carbon/blob/master/src/components/batch-selection/components.test-pw.tsx#L6) too.

playwright/components/card/index.ts
DipperTheDan: Yeah it should be single quotes. I had issues here and I experimented with passing an `${index}` in the `page.locator` hence the need for back ticks. I'll get them changed to single quotes 👍🏻

playwright/components/card/index.ts
DipperTheDan: Same answer as previous. Yeah it should be single quotes. I had issues here and I experimented with passing an ${index} in the page.locator hence the need for back ticks. I'll get them changed to single quotes 👍🏻

src/components/card/card.pw.tsx
DipperTheDan: Will do. This is the kind of thing I figured the linter would do. Especially if we do have this as a pattern.

src/components/card/card.pw.tsx
DipperTheDan: Same answer as previous too. Will do. This is the kind of thing I figured the linter would do. Especially if we want or do have this as a pattern.

sianford (approved):

playwright/components/card/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export const card = (page: Page) => page.locator(CARD).nth(0);
+
+export const draggableCard = (page: Page, index: number) =>
+  page.locator(DRAGGABLE_CARD_AT_IDEX(index)).locator(CARD_DRAGGABLE_ELEMENT);
+
+export const draggableContainer = (page: Page, index: number) =>
+  page.locator(DRAGGABLE_CONTAINER_AT_INDEX(index));
+
+export const columnCard = (page: Page) =>
+  page.locator(CARD_COLUMN).locator("div:nth-child(2)");
```
Parsium, line 21: **suggestion**: Likewise we could also simplify this locator like so:

```suggestion
  page.locator(CARD_COLUMN);
```

Parsium (approved):

sianford (approved):

PR #6305, opened 2023-09-08
test(box): convert cypress tests to playwright

robinzigmond:
### Proposed behaviour

Box in-browser tests should be written using Playwright

### Current behaviour

Tests currently use Cypress

### Checklist

#### QA

### Additional context

The test that a Sticky box covers up radio buttons has been removed - it proved impossible to rewrite this in a useful way in Playwright, and on further investigation it was discovered that the Cypress test wasn't testing what it should have done. (It only worked because the scroll container was scrolled to (0, 300) which would position the radio buttons off-screen - changing to (0, 200) to position them behind the sticky box causes the test to fail, and further the test still passes even when the bugfix to add a z-index was reverted.)

I have also removed the remaining Radiobutton Cypress locators as these were only being used in the Box Cypress tests, which have now been removed.

### Testing instructions


src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,897 @@
...
+
+test.describe("should render Box component", () => {
+  colorConstants.forEach(([color, rgbValue]) => {
+    test(`should verify text is ${color} when color prop is passed as string`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<Default color={color} />);
+      const boxElement = await getDataElementByValue(page, "box");
+      await expect(boxElement).toHaveAttribute("color", color);
```
Parsium, line 83: **question**: Do we need these attribute assertions here and in the other tests where we are checking CSS values, when we also have CSS assertions in these tests as well?

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,897 @@
...
+    }) => {
+      await mount(<Default bg="primary" width={percentage} />);
+      const boxElement = await getDataElementByValue(page, "box");
+      await expect(boxElement).toHaveAttribute("width", String(percentage));
+      await assertCssValueIsApproximately(boxElement, "width", parseInt(width));
+    });
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  widthConstants.forEach(([width, percentage, number]) => {
```
Parsium, line 199: **suggestion**: For parameterised tests where we have unused test parameters, what do you think about storing the parameters for each test in an object instead of an array? That way we can access the values we need without having to disable the linting rule for the unused ones.

```suggestion
  widthConstants.forEach(({ width, number }) => {
```  
where:
```ts
const widthConstants = [
  { width: "135px", percentage: 0.1, number: 135 },
  { width: "683px", percentage: 0.5, number: 683 },
  { width: "1366px", percentage: 1, number: 1366 },
];
```

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,897 @@
...
+        await expect(boxElement).toHaveCSS("overflow-y", overflow);
+      });
+    }
+  );
+
+  [
+    [200, 300],
+    [400, 400],
+  ].forEach(([width, minWidth]) => {
+    test(`should verify when number Width is set to ${width} that width is not less than minWidth ${minWidth}`, async ({
```
Parsium, line 366: **nitpick**: What do you think about amending the test description slightly, so its a bit clearer what the preconditions are and what is being asserted? 

```suggestion
    test(`when minWidth prop is 300 and width prop is ${width}, the calculated width should be greater than ${minWidth}px`, async ({
```

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,897 @@
...
+      const boxElement = await getDataElementByValue(page, "box");
+      await expect(boxElement).toHaveCSS("overflow-wrap", wrap);
+    });
+  });
+
+  ([
+    ["light", "rgb(102, 132, 148)", "rgb(242, 245, 246)"],
+    ["dark", "rgb(153, 173, 183)", "rgb(51, 91, 112)"],
+  ] as [BoxProps["scrollVariant"], string, string][]).forEach(
+    ([variant, thumb, track]) => {
```
Parsium, line 698: **suggestion(non-blocking)**: For parameterised tests where we are having to explicitly set types on the test parameters, what do you think about using objects for storing the parameters for each test instead of an array?

```suggestion
  [
    {variant: "light", thumbColor: "rgb(102, 132, 148)", trackColor: "rgb(242, 245, 246)"},
    {variant: "dark", thumbColor: "rgb(153, 173, 183)", trackColor: "rgb(51, 91, 112)"},
  ].forEach({variant, thumbColor, trackColor}) => {
```

That way we won't have to worry about explicitly defining the types, as the type of each parameter will be a union of constant values.

Parsium:
**suggestion**: now we have moved any non-storybook examples from the test stories file, we will need to also remove [these old stories](https://github.com/Sage/carbon/blob/5d71afa71405feb50a7d826bbb44b8750a22590b/src/components/box/box-test.stories.tsx#L82-L140)now that they are no longer used

src/components/box/box.pw.tsx
robinzigmond: When I first read this comment I thought it was a good suggestion - but now I'm coming to actually address it I'm having significant doubts.

I certainly would rather not have to disable the linting, and destructuring objects rather than arrays looks quite nice too. But as you say, this would mean redefining `widthConstants` to be an array of objects rather than an array of arrays - which mean not just htis test, but the others that use `widthConstants`, would need changing from array to object destructuring. And this in turn would make the entire test suite quite inconsistent, in a way that could confuse future readers.

The entire file could be converted to use objects rather than arrays of course, but then that would be inconsistent with all the other conversions we've got going on (and completed), where we use arrays 😅 

For this reason I'm going to leave this as it is for now - but very happy to have a conversation about this.

robinzigmond: I've also just remembered that I can remove the ESLint warning without disabling it, if I change the destructuring to `[width, , number]`. This is valid syntax - although I have to admit, I find it pretty gross 😬 But I can live with it if you (and the other reviewer) prefer it!

src/components/box/box.pw.tsx
robinzigmond: I'll stick with "not less than" rather than "greater" (which isn't strictly accurate) - but otherwise I like your suggested rewriting 👍 I'll amend the other descriptions which are like this too.

robinzigmond: no, in fact it's an equality assertion in both cases (albeit an approximate one), so I'll remove any "greater/less than" in the description

src/components/box/box.pw.tsx
robinzigmond: I like this idea in principle, but given my comments above on inconsistency of using objects for some tests and arrays for other ones, I'm not so sure. Although I am more inclined to go with this than with your other suggestion whether the array was a const defined outside the tests. I'll ask about this in general I think 👍

src/components/box/box.pw.tsx
Parsium: Cool thanks for this Robin 👍🏼 Personally I do prefer the object-based syntax as it is easier to read and handle typing, but happy to go with the majority if others disagree :)

src/components/box/box.pw.tsx
Parsium: Ah good point, double checking the mock components it seems we may be accidentally passing CSS attributes as HTML attributes which is incorrect, so we will need to amend this at a later stage

Screenshot 2023-09-15 at 14 25 38

playwright/support/helper.ts
```diff
@@ -217,7 +219,7 @@ export function keyCode(
   release?: boolean;
 } {
   return keys[type];
-};
+}
```
ZhuoyuJin, line 222: Why ";" is removed here?

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,887 @@
...
+      const boxElement = await getDataElementByValue(page, "box");
+      await assertCssValueIsApproximately(boxElement, "width", actualWidth);
+    });
+  });
+
+  ([
+    ["200px", 300],
+    ["400px", 400],
+  ] as [string, number][]).forEach(([width, actualWidth]) => {
+    test(`when minWidth prop is 300 and width prop is ${width}, the calculated width should be ${actualWidth}`, async ({
```
ZhuoyuJin, line 358: This test description is identical to the test above, could we change it to something like "when minWidth prop is 300 and width prop is ${width} **as a string**, the calculated width should be ${actualWidth}", and similar for the above test, which we can add in "**as a number**"

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,887 @@
...
+      const boxElement = await getDataElementByValue(page, "box");
+      await assertCssValueIsApproximately(boxElement, "width", actualWidth);
+    });
+  });
+
+  [
+    [0.1, 300],
+    [0.3, 409],
+  ].forEach(([width, actualWidth]) => {
+    test(`when minWidth prop is 300 and width prop is ${width}, the calculated width should be ${actualWidth}`, async ({
```
ZhuoyuJin, line 372: same comment as above

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,887 @@
...
+    test(`when maxWidth prop is 600 and width prop is ${width}, the calculated width should be ${actualWidth}`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<Default maxWidth={600} width={width} />);
+      const boxElement = await getDataElementByValue(page, "box");
+      await assertCssValueIsApproximately(boxElement, "width", actualWidth);
+    });
+  });
+
```
ZhuoyuJin, line 423: same here, for tests between line 382-423

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,887 @@
...
+  ] as [string, number][]).forEach(([height, actualHeight]) => {
+    test(`when maxHeight prop is 600 and height prop is ${height}, the calculated width should be ${actualHeight}`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<Default maxHeight={600} height={height} />);
+      const boxElement = await getDataElementByValue(page, "box");
+      await assertCssValueIsApproximately(boxElement, "height", actualHeight);
+    });
+  });
```
ZhuoyuJin, line 478: same here

src/components/box/box.pw.tsx
```diff
@@ -0,0 +1,887 @@
...
+  ] as [string, number][]).forEach(([height, actualHeight]) => {
+    test(`when minHeight prop is 600 and height prop is ${height}, the calculated width should be ${actualHeight}`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<Default minHeight={600} height={height} />);
+      const boxElement = await getDataElementByValue(page, "box");
+      await assertCssValueIsApproximately(boxElement, "height", actualHeight);
+    });
+  });
```
ZhuoyuJin, line 450: same here

src/components/box/box.pw.tsx
robinzigmond: no, on reflection I think I do prefer your suggestion of using an array of object for at least some of these, and that the inconsistency isn't really a problem. But I'll have another think about this on Monday morning 😄

src/components/box/box.pw.tsx
robinzigmond: so it turns out that, in trying this, the type annotation is still needed, because the "variant" type is inferred as string, rather than the specific strings needed for he `scrollVariant` prop. While this can be fixed in the object version too, with type annotations/assertions or an `as const`, this seems to remove the advantage of using an object over an array here so I'm going to keep this one as an array of arrays.

ZhuoyuJin (approved):

PR #6312, opened 2023-09-12
test(link-preview): migrate tests from cypress to playwright 

ZhuoyuJin:
### Proposed behaviour

- Refactor cypress tests to playwright

### Current behaviour

- Tests are currently using cypress

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- [ ] Run Link Preview tests in Playwright Test Runner to check if the link-preview.pw.tsx file passed
- [ ] Run all other tests in Playwright Test Runner to check none of the other *.pw.tsx files have regressed


src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,266 @@
+/* eslint-disable jest/valid-expect, no-unused-expressions */
```
divyajindel, line 1: Could you please remove this line ?

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,266 @@
...
+} from "../../../playwright/support/helper";
+import { CHARACTERS } from "../../../playwright/support/constants";
+import { HooksConfig } from "../../../playwright";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+const testCypress = "test-cypress";
+const urlProp = "./carbon-by-sage-logo.png";
+const keysToTrigger = ["Space", "Enter"];
+
+test.describe("when focused", () => {
```
divyajindel, line 23: All the test should not come under this describe block so could you please make it separate describe block

divyajindel (changes requested):

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,266 @@
...
+      mount,
+      page,
+    }) => {
+      await mount(<LinkPreviewComponentTest isLoading />);
+
+      const previewComponent = page.locator(`[data-component="preview"]`);
+
+      const count = await previewComponent.count();
+
+      expect(count).toBe(4);
```
divyajindel, line 103: I've observed that attribute checking has been done in cypress but missed here  "("have.attr", "data-component", "preview")"

divyajindel (changes requested):

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,266 @@
...
+    });
+
+    test("should verify border outline color of close icon on focus", async ({
+      mount,
+      page,
+    }) => {
+      await mount<HooksConfig>(
+        <LinkPreviewComponentTest
+          as="div"
+          // eslint-disable-next-line no-console
```
divyajindel, line 165: do we need this line here?

divyajindel (changes requested):

src/components/link-preview/link-preview.pw.tsx
ZhuoyuJin: Hi @divyajindel, this is because the line for testing "have.attr" becomes a bit useless, as I already have the line99, which states that "page.locator(`[data-component="preview"]`);", we used this way to locate [data-component="preview"], which we should definitely have attribute "preview" as our "data-component"

src/components/link-preview/link-preview.pw.tsx
ZhuoyuJin: I think so, otherwise it returns warning for the next line because of the "console.log"

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  test("should render Link Preview with url prop", async ({ mount, page }) => {
+    await mount(<LinkPreviewComponentTest url="https://www.foo.com" />);
+
+    await expect(page.getByText("www.foo.com")).toBeAttached();
+  });
+});
+
+test.describe("check functionality for Link Preview component", () => {
+  test("should verify hover color of Link Preview component", async ({
```
DipperTheDan, line 127: **suggestion: (non-blocking)** might be nice to trim these tests descriptions a bit. Otherwise we'll have `Link Preview` mentioned three times in the testing block. In your version it'll end up being `Link Preview -> Check functionality of Link Preview component -> should verify hover color for Link Preview component` as an example. 

```suggestion
  test("should verify hover color", async ({
```

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,258 @@
...
+      await page.keyboard.down(key);
+
+      await linkPreviewCloseIcon(page).waitFor();
+      expect(callbackCount).toEqual(1);
+    });
+  });
+});
+
+test.describe("Accessibility tests for LinkPreview component", () => {
+  test("should pass accessibilty tests for default story", async ({
```
DipperTheDan, line 219: **suggestion: (non blocking)** Might be best to make sure these descriptions do not refer to stories as they are no longer stories.

Something like this should do but up to you what you want to go for 😄 

```suggestion
  test("should pass accessibility tests for default example", async ({
```

src/components/link-preview/link-preview.pw.tsx
```diff
@@ -0,0 +1,258 @@
...
+    mount,
+    page,
+  }) => {
+    await mount(<LinkPreviewComponentTest onClose={() => {}} />);
+
+    await checkAccessibility(page);
+  });
+});
+
+test("should render with the expected border radius styling", async ({
```
DipperTheDan, line 247: **suggestion: (non blocking)** Would you be able to move this in to the main block with the rest of the tests? Seems a bit odd to have it after the accessibility tests. I know you've copied this from the Cypress tests as that's what we've done there but it might be nice to tidy them up in to the main block.

playwright/components/link-preview/index.ts
```diff
@@ -0,0 +1,30 @@
+import type { Page } from "@playwright/test";
+import { PILL_CLOSE_ICON } from "../locators";
+import { dlsRoot } from "../../../playwright/components/index";
+import { LinkPreviewProps } from "../../../src/components/link-preview";
+
+// component preview locators
+export const linkPreview = (page: Page) => {
+  return dlsRoot(page).locator("a");
```
sianford, line 8: I don't think you need to specify `dlsRoot()` here as `locator("a")` is in there and it should find it. This was a comment James made on my Note PR which uses LinkPreview :)

playwright/components/link-preview/index.ts
```diff
@@ -0,0 +1,30 @@
...
+      selector = "a";
+      break;
+    case "div":
+      selector = "div";
+      break;
+    default:
+      throw new Error(`Invalid 'as' value: ${as}`);
+  }
+
+  return dlsRoot(page).locator(selector).first();
```
sianford, line 25: As above, I don't think `dlsRoot(page)` needs to be specified here. You can just have `return locator(selector).first();` I think.

src/components/link-preview/link-preview.pw.tsx
sianford: @DipperTheDan I don't think that test case should be in with any of the other describes, unless those others relating to things like border outline colours etc are put in their own `describe` for, say, styling. This is liked with our discussion in the channel about test cases being out on their own or in a `describe` where we decided if they're not related to the other cases in the existing `describe` blocks then they can remain out on their own.

sianford: @DipperTheDan I've since decided that we should wrap all of our tests in a `describe` regardless of number to keep the format and standard uniform throughout :)

DipperTheDan (approved):

divyajindel (approved):

PR #6316, opened 2023-09-14
fix(menu-item): address aria role accessibility error with icon only menu items - FE-6055

### LINKED ISSUE #4986: Menu with icon - Accessibility violations "Elements must only use allowed ARIA attributes"
### Current behaviour

One accessibility violation in "with icon" story for Menu componen
Image 22-03-2022 at 12 37
t

### Expected behaviour

No Accessibility violations

https://carbon.sage.com/?path=/story/menu--with-icon

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

latest

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #6175: Menu - Accessibility violation in the 'with icon' story
### Current behaviour

There is an accessibility violation for reason "Elements must only use allowed ARIA attributes" in the 'with icon' story.

Screenshot 2023-07-05 at 16 44 45

### Expected behaviour

There are no accessibility violations in the Menu component.

https://carbon.sage.com/?path=/story/menu--with-icon-story

### JIRA Ticket (Sage Only)

FE-6048

### Suggested Solution

Resolve violation.

### Carbon Version

119.5.0

### Design Tokens Version

_No response_

MacOS

_No response_

### Confidentiality

DipperTheDan:
`Elements must only use allowed ARIA attributes` error was occurring in an example of Menu with icon only menu items. This was due to the aria-label being specified on the anchor element and also the underlying span element of the icon itself.

fixes #6175
fixes #4986

### Proposed behaviour

Address "Elements must only use allowed ARIA attributes" warning on `with icon` Menu story.

### Current behaviour

`with icon` Menu story has an accessibility error of "Elements must only use allowed ARIA attributes".

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- `with icon` Story should have no accessibility errors. 
- Should be no regression of features or screenreader behaviour in `menu-item`, `link` and `icon`.

### DIFF
--- a/src/components/link/link.component.tsx
+++ b/src/components/link/link.component.tsx
@@ -49,6 +49,8 @@ export interface LinkProps extends StyledLinkProps, React.AriaAttributes {
   rel?: string;
   /** @ignore @private internal prop to be set when no href or onClick passed */
   placeholderTabIndex?: boolean;
+  /** @ignore @private internal prop to be set when no aria-label should be specified */
+  removeAriaLabelOnIcon?: boolean;
 }
 
 export const Link = React.forwardRef<
@@ -75,6 +77,7 @@ export const Link = React.forwardRef<
       variant = "default",
       isDarkBackground,
       placeholderTabIndex,
+      removeAriaLabelOnIcon,
       ...rest
     }: LinkProps,
     ref
@@ -109,7 +112,7 @@ export const Link = React.forwardRef<
           type={icon}
           bgSize="extra-small"
           disabled={disabled}
-          ariaLabel={ariaLabel}
+          ariaLabel={removeAriaLabelOnIcon ? undefined : ariaLabel}
           tooltipMessage={tooltipMessage}
           tooltipPosition={tooltipPosition}
         />
--- a/src/components/link/link.spec.tsx
+++ b/src/components/link/link.spec.tsx
@@ -293,6 +293,19 @@ describe("Link", () => {
         expect(anchor.getAttribute("aria-label")).toEqual("test");
       });
     });
+
+    describe("when removeAriaLabelOnIcon is true", () => {
+      it("should set aria-label as undefined on icon", () => {
+        wrapper = renderLink({
+          onClick: () => null,
+          icon: "home",
+          "aria-label": "test",
+          removeAriaLabelOnIcon: true,
+        });
+
+        expect(wrapper.find(Icon).props().ariaLabel).toBe(undefined);
+      });
+    });
   });
 
   describe("'negative' variant", () => {
--- a/src/components/menu/menu-item/menu-item.component.tsx
+++ b/src/components/menu/menu-item/menu-item.component.tsx
@@ -242,6 +242,7 @@ export const MenuItem = ({
     rel,
     onClick,
     icon,
+    removeAriaLabelOnIcon: true,
     selected,
     variant,
     onKeyDown: !inFullscreenView ? handleKeyDown : undefined,
--- a/src/components/menu/menu-item/menu-item.spec.tsx
+++ b/src/components/menu/menu-item/menu-item.spec.tsx
@@ -748,9 +748,10 @@ describe("MenuItem", () => {
       expect(wrapper.find(StyledIcon).exists()).toBe(true);
     });
 
-    it("add aria-label when it is set", () => {
+    it("add aria-label to Link and Icon aria-label to be undefined when it is set on menu item", () => {
       wrapper = mount(<MenuItem icon="settings" ariaLabel="Settings" />);
-      expect(wrapper.find(Icon).props().ariaLabel).toBe("Settings");
+      expect(wrapper.find(Link).props().ariaLabel).toBe("Settings");
+      expect(wrapper.find(Icon).props().ariaLabel).toBe(undefined);
     });
 
     it("give error when `aria-label` is not set and menu item has no child text", () => {


src/components/icon/icon.component.tsx
```diff
@@ -156,6 +159,7 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
       disabled: disabledFromContext || disabled,
       fontSize,
       hasTooltip,
+      hideAriaLabel,
```
robinzigmond, line 162: **question**: I'm probably being stupid again here - but why does `hideAriaLabel` need to go down to the underlying styled component (a `span` in this case)? It doesn't appear to use it, and no styles will depend on it as far as I can see - nor is it a valid HTML attribute.

src/components/link/link.component.tsx
```diff
@@ -109,6 +112,7 @@ export const Link = React.forwardRef<
           type={icon}
           bgSize="extra-small"
           disabled={disabled}
+          hideAriaLabel={hideAriaLabel}
```
robinzigmond, line 115: **question**: I could well be missing something, but can't we just change the `ariaLabel={ariaLabel}` to `ariaLabel={hideAriaLabel ? undefined : ariaLabel}`, to avoid needing a `hideAriaLabel` prop at all on `Icon`? (You'd still need it on Link though, assuming you ever wanted a <span> with an incorrect aria-label for an icon inside a Link.)

src/components/menu/menu-item/menu-item.component.tsx
```diff
@@ -242,6 +242,7 @@ export const MenuItem = ({
     rel,
     onClick,
     icon,
+    hideAriaLabel: true,
```
robinzigmond, line 245: **suggestion**: (this isn't about this line of code in particular but seems the best place I can find to put it)

I think it's worth adding a test for MenuItem specifically, that we don't end up with an aria-label on the StyledIcon when rendering an icon. I know you have the test for the new prop on Icon itself, and that will satisfy coverage, but you're nowhere testing that we're actually applying this in MenuItem, and that was after all what the ticket is really about.

src/components/icon/icon.component.tsx
```diff
@@ -55,6 +55,8 @@ export interface IconProps extends Omit<StyledIconProps, "type">, MarginProps {
   inputSize?: "small" | "medium" | "large";
   /** @ignore @private */
   tabIndex?: number;
+  /** @ignore @private internal prop to be used in the event no aria-label should be specified */
+  hideAriaLabel?: boolean;
```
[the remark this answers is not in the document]
DipperTheDan: Yeah I had considered tearing out the aria-label out of `Icon` all together but this would be a breaking change as aria-label is advertised as a prop. I figured for now being able to hide the aria-label where we needed to would suffice. I had checked that if I removed the aria-label from `Icon` would it also address the `Pod` issue too but it didn't. Hence I went for this approach instead.

src/components/link/link.component.tsx
```diff
@@ -49,6 +49,8 @@ export interface LinkProps extends StyledLinkProps, React.AriaAttributes {
   rel?: string;
   /** @ignore @private internal prop to be set when no href or onClick passed */
   placeholderTabIndex?: boolean;
+  /** @ignore @private internal prop to be set when no aria-label should be specified */
+  hideAriaLabel?: boolean;
```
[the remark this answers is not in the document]
DipperTheDan: I'll go with `removeAriaLabelFromIcon` 👍🏻

src/components/icon/icon.spec.tsx
```diff
@@ -470,6 +470,18 @@ describe("Icon component", () => {
       );
     });
 
+    it("does not set an aria label if hideAriaLabel is true", () => {
+      const wrapper = mount(
+        <Icon type="home" hideAriaLabel ariaLabel="home" />
+      );
+
+      expect(wrapper.find(StyledIcon).props()).toEqual(
```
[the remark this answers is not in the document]
DipperTheDan: I went for this because it's the same format that is used to test the `aria-label` further up in the tests so I wanted to keep things congruent. I am happy to change it your way though.

src/components/link/link.component.tsx
DipperTheDan: I'll message you privately about this as It'll need some further explaining.

src/components/menu/menu-item/menu-item.component.tsx
DipperTheDan: The reason I tested it on `Icon` was that it wasn't technically `MenuItem` that had the bug. The bug came from `Link` which in turn came from `Icon` getting a aria-label set. I'll get a test added to `MenuItem` though as it's not that big of a job to do for the extra piece of mind it can give us 👍🏻

src/components/icon/icon.spec.tsx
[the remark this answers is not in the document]
robinzigmond: fair enough, I'd missed that. Personally I'd still write this test what I think is the better way, and maybe even change the other one - but it doesn't really matter. (The one scenario that does matter is if somehow the icon ended up with an aria-label taht was different from the one passed in the prop - then your test would pass while my suggestion would fail. But that's such an out-there situation I'm not going to try to use this as a serious argument in favour of mine 😂 )

src/components/menu/menu-item/menu-item.component.tsx
robinzigmond: Actually you're right that it's Link where the bug actually was, so a test on that is not a bad idea. I'd definitely prefer to have one on either that or Menuitem (or even both, but probably no need), than neither, though 👍

DipperTheDan: I've amended one on `MenuItem` and added one to `Link` based on the changes made in the last commit. So we're covered for both now.

`MenuItem` checks `Link` and `Icon`.
`Link` checks `Link` and `Icon`.

So if a regression happens in `Link` or `MenuItem`, we'll know about it 👍🏻

src/components/icon/icon.spec.tsx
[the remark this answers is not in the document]
DipperTheDan: Changed the unit tests as per the other suggested changes so this was kind of redundant now.

src/components/icon/icon.component.tsx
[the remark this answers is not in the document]
DipperTheDan: Changed this approach to having `aria-label={undefined}` now passed from `Link`. I've removed this prop from `Icon` now based on the other suggested changes.

src/components/link/link.spec.tsx
```diff
@@ -293,6 +293,18 @@ describe("Link", () => {
         expect(anchor.getAttribute("aria-label")).toEqual("test");
       });
     });
+
+    describe("when removeAriaLabelOnIcon is true", () => {
+      it("should set aria-label as undefined on icon", () => {
+        wrapper = renderLink({
```
robinzigmond, line 299: **suggestion** (non-blocking): I would feel this test was much more robust if we also passed an `ariaLabel` prop to the Link, to ensure that the `removeAriaLabelOnIcon` is really doing its job

DipperTheDan: Not an unreasonable suggestion and makes sense from a robustness perspective. I'll get it done now and pushed back up 👍🏻

robinzigmond (approved):

Parsium (approved):
Looks good to me 👍🏼 

As it seems `MenuItem` is coupled to the internal styled components of `Link` and `Icon`, which appears to be partly responsible for the bug, it might be worth us doing some refactoring in the future to decouple `MenuItem` from these internals.

divyajindel (approved):
Looks good to me!

Issue #6240, opened 2023-08-08
Tile component - add height prop


harpalsingh:
### Desired behaviour

When using the tile component, we should allow the ability to pass a height value.

### Current behaviour

You can not add a height prop to the Tile component.

### Suggested Solution

Similar to the implementation with the Box component, allow height to be a useable prop on Tile as it has many uses within design.

### CodeSandbox or Storybook URL

https://carbon.sage.com/?path=/docs/tile--default-story

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

Parsium:
FE-6180

edleeks87:
released in https://github.com/Sage/carbon/commit/123630b62c53b264606298a28f99a65cd136a39a

PR #6326, opened 2023-09-22
feat(tile): add support for passing custom height to component

edleeks87:
re #6240

### Proposed behaviour

Adds `height` prop to `Tile` component to support setting custom heights.

### Current behaviour

No support for custom heights

### Checklist

#### QA

### Additional context

Migration of stories from MDX will be done as part of FE-5681, it will be a breaking change that surfaces `TileContent` and there is type issues with current implementation that uses `Content` with `width` set etc
### Testing instructions


src/components/tile/tile.style.ts
```diff
@@ -76,10 +77,13 @@ const TileContent = styled.div<TileContentProps>`
       }
     `}
 
-    ${width &&
+    ${(width || height) &&
     css`
       flex-grow: 0;
-      ${computeWidth({ width })}
+      ${computeSizing({
+        width: width || /* istanbul ignore next */ undefined,
```
edleeks87, line 84: This is fully tested in cypress so ignored here, implementation is likely to slightly change as part of FE-5681 but it was preferred that this prop was surfaced before the breaking change is made

src/components/tile/tile.stories.mdx
```diff
@@ -257,6 +256,52 @@ Please note that it will not have any effect if the Tile orientation is set to `
   </Story>
 </Canvas>
 
+### With custom heights
+
+The Tile component can also have a custom `height` which accepts either numbers or strings: any decimal between 0 and 1.0
+will be converted into a percentage, any number grater will be be treated as pixels and any valid CSS string will also work.
```
DipperTheDan, line 262:
```suggestion
will be converted into a percentage, any number greater will be be treated as pixels and any valid CSS string will also work.
```

src/style/utils/element-sizing.ts
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line no-restricted-imports
```
Parsium, line 1: **suggestion**: this line doesn't seem to raise this same linting error when this comment is removed.


```suggestion
```

src/components/tile/tile.stories.mdx
Parsium: It seems we had a similar typo for the `With custom widths` description too, would you be happy to amend this @edleeks87?

src/components/tile/tile.component.tsx
```diff
@@ -35,15 +35,21 @@ export interface TileProps extends SpaceProps, WidthProps {
     | "info";
   /** Sets the level of roundness of the corners, "default" is 8px and "large" is 16px */
   roundness?: "default" | "large";
+  /**
+   * Set a percentage-based height for the whole Tile component, relative to its parent.
+   * If unset or zero, this will default to 100%.
```
Parsium, line 40: **question**: It seems we don't have a default value for `height`, which I assume is because we don't want the `Tile` completely filling its parent container by default?

src/components/multi-action-button/multi-action-button.style.ts
```diff
@@ -4,7 +4,7 @@ import StyledButton from "../button/button.style";
 import baseTheme from "../../style/themes/base";
 import StyledIcon from "../icon/icon.style";
 import { MultiActionButtonProps } from "./multi-action-button.component";
-import computeWidth from "../../style/utils/width";
+import computeWidth from "../../style/utils/element-sizing";
```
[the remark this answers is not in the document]
edleeks87: It's a default import so the name here isn't important imo, I decided to leave it as computeWidth as that's what it's doing here

src/components/tile/tile.component.tsx
edleeks87: Yes that was my thinking behind it 👍

src/components/tile/tile.stories.mdx
edleeks87: Yep that's what I get for copying and pasting 🤦

src/components/multi-action-button/multi-action-button.style.ts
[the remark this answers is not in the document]
edleeks87: I've updated as it's not a big deal

DipperTheDan (approved):

Parsium (dismissed):

DipperTheDan (dismissed):

DipperTheDan (approved):

Parsium (approved):

divyajindel (approved):

PR #6334, opened 2023-09-28
feat:(Typography) add text-align prop and support span variant  <FE-6204 & FE-6220>

### LINKED ISSUE #6331: `textAlign` attribute for `Typography` component
### Desired behaviour

`Typography` content can be aligned by one prop, `textAlign`

### Current behaviour

Currently, to align `Typography`, it should have Flex-`Box` wrapper.
In the case of wrapping `Typography` with `Link`, this does not work as shown in code-sandbox example

### Suggested Solution

Please could you enable `textAlign` attribute for `Typography` so one prop can do the text alignment? 

https://28tpvm.csb.app/

_No response_

### Confidentiality

nineteen88:
### Proposed behaviour

- add prop support for `text-align`
- include support for using `span` as a variant

fixes #6331 

### Current behaviour

Currently there is no support for the above props and span variant

### Checklist

#### QA

https://codesandbox.io/s/carbon-quickstart-forked-24w5sc

### DIFF
--- a/cypress/components/typography/typography.cy.tsx
+++ b/cypress/components/typography/typography.cy.tsx
@@ -248,6 +248,24 @@ context("Tests for Typography component", () => {
       }
     );
 
+    it.each(VARIANT_TYPES as TypographyProps["variant"][])(
+      "should check text-align for %s variant prop for Typography component when set",
+      (variant) => {
+        CypressMountWithProviders(
+          <Typography variant={variant}>{testCypress}</Typography>
+        );
+
+        const variantElem = getAs(String(variant));
+        const textDecorationLine = getDecoration(String(variant));
+
+        cy.get(variantElem).should(
+          "have.css",
+          "text-decoration-line",
+          textDecorationLine
+        );
+      }
+    );
+
     it.each(["ol", "ul"])(
       "should check as prop set to %s for List component",
       (as) => {
@@ -279,6 +297,18 @@ context("Tests for Typography component", () => {
         cy.get("h1")
           .should("have.text", testCypress)
           .and(`${assertion}.css`, "text-overflow", "ellipsis");
+      }
+    );
+
+    it.each([["left"], ["center"], ["right"], ["justify"]])(
+      "should adjust the text alignment when textAlign prop is set to %s",
+      (textAlignment) => {
+        CypressMountWithProviders(
+          <Typography variant="h1" textAlign={textAlignment}>
+            {testCypress}
+          </Typography>
+        );
+        cy.get("h1").and("have.css", "text-align", textAlignment);
       }
     );
 
--- a/src/components/typography/typography.component.tsx
+++ b/src/components/typography/typography.component.tsx
@@ -58,6 +58,8 @@ export interface TypographyProps extends SpaceProps, TagProps {
   whiteSpace?: string;
   /** Override the word-wrap */
   wordWrap?: string;
+  /** Override the text-align */
+  textAlign?: string;
   /** Override the text-overflow */
   textOverflow?: string;
   /** Apply truncation */
@@ -91,6 +93,7 @@ export const Typography = ({
   listStyleType,
   whiteSpace,
   wordWrap,
+  textAlign,
   textOverflow,
   truncate,
   color = "blackOpacity90",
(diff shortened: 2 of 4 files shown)


src/components/typography/typography.component.tsx
```diff
@@ -39,6 +40,8 @@ export interface TypographyProps extends SpaceProps, TagProps {
   children?: React.ReactNode;
   /** The visual style to apply to the component */
   variant?: VariantTypes;
+  /** Override the ref value */
+  ref?: React.Ref<HTMLElement>;
```
Parsium, line 44: **issue**: I assume this was added so we could have the forward ref documented? 

Unfortunately this will actually add support for a new `ref` React prop that will be separate to the forwarded ref.

```suggestion
```

Due to how the typing of `React.forwardRef()` is defined, the type of the component it returns should automatically type the forwarded `ref`.

However, storybook won't be able to generate any documentation for the `ref`, so we may need to just document it via a story as a workaround.

src/components/typography/typography.spec.tsx
```diff
@@ -627,6 +679,37 @@ describe("Typography", () => {
...
+            FooBar
+          </Typography>
+        </ThemeProvider>
+      );
+    };
+    render(<MockComponent />);
+
+    expect(typeof typographyHandle.current?.focus).toBe("function");
+
+    await typographyHandle.current?.focus();
```
Parsium, line 709: **suggestion(non-blocking)**: `focus()` doesn't return a promise so we don't necessarily need the `await` here.

src/components/typography/typography.spec.tsx
```diff
@@ -47,6 +49,20 @@ const pStyling = {
...
+function render(
+  ui: React.ReactElement,
+  options?: RTLRenderOptions,
+  document = globalThis.document
+) {
+  if (document.body.innerHTML.length >= 1) {
+    throw new Error(
+      "Found DOM to be non-empty before rendering. Please make sure to call cleanup() after each test."
+    );
+  }
```
Parsium, line 61: **thought**: it might be worth us writing a custom render function for cleaning up React Portals when we properly start migrating tests over to RTL. Just so we don't have to duplicate this in several test files.

tomdavies73 (commented):
**suggestion**: we have some variant tests already in `typography.cy.tsx`, where we check the variant is applied, font-size, line-height, font-weight, text-transform and text-decoration-line. It might be worth adding `"span"` to these tests. We can do this by adding `"span"` to `VARIANT_TYPES`, `getAs, getSize, getLineHeight, getWeight` functions etc. Could be a good shout to ensure span is added with other variants as a baseline.

src/components/typography/typography.spec.tsx
nineteen88: Removed all the ref/tab-index work from the PR so this won't be required here but it would be worth doing sooner or later, yeah.

src/components/typography/typography.spec.tsx
nineteen88: Outdated/removed now

src/components/typography/typography.component.tsx
nineteen88: This work has been removed from the PR so it's no longer applicable

nineteen88:
> **suggestion**: we have some variant tests already in `typography.cy.tsx`, where we check the variant is applied, font-size, line-height, font-weight, text-transform and text-decoration-line. It might be worth adding `"span"` to these tests. We can do this by adding `"span"` to `VARIANT_TYPES`, `getAs, getSize, getLineHeight, getWeight` functions etc. Could be a good shout to ensure span is added with other variants as a baseline.

Added this now.

tomdavies73 (dismissed):

Parsium (approved):

divyajindel:
Hi @nineteen88 , could you please add cypress tests for TextAlign prop?

Parsium (approved):

cypress/components/typography/typography.cy.tsx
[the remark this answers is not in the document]
nineteen88: Yeah, I've spoken to @edleeks87 about this and we will do a refactor of the tests later. You're definitely correct.

divyajindel (approved):

PR #6337, opened 2023-09-29
feat(label, Form): ensure all inputs support isOptional and required props

DobroTora:
### Proposed behaviour

Ensure that all form inputs support `required` and `isOptional` prop (no longer legacy).
Add stories and unit tests for all inputs
Add a Form story with some required inputs and `RequiredFieldsIndicator` (new component)
Any input not `required` has been made optional in our docs

Small updates to spacing of `*` and colour of `(option)` content appended to label to match DS design

### Current behaviour

Some inputs do not support required or isOptional props
No Form story for required etc

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/fieldset/fieldset.style.ts
+++ b/src/components/fieldset/fieldset.style.ts
@@ -50,7 +50,8 @@ const LegendContainerStyle = styled.div<StyledFieldsetProps>`
 
   legend {
     font-size: 20px;
-    font-weight: 600;
+    font-weight: var(--fontWeights700);
+    color: var(--colorsUtilityYin090);
     line-height: 24px;
     margin-right: 4px;
 
--- a/src/__internal__/fieldset/fieldset.style.ts
+++ b/src/__internal__/fieldset/fieldset.style.ts
@@ -76,7 +76,8 @@ const StyledLegend = styled.legend<StyledLegendProps>`
   align-items: center;
   margin-bottom: var(--spacing100);
   padding: 0;
-  font-weight: 600;
+  font-weight: var(--fontWeights700);
+  color: var(--colorsUtilityYin090);
   ${({ inline, width, align, rightPadding }) =>
     inline &&
     css`
--- a/src/__internal__/label/label.style.ts
+++ b/src/__internal__/label/label.style.ts
@@ -10,7 +10,7 @@ export interface StyledLabelProps {
 const StyledLabel = styled.label<StyledLabelProps>`
   color: var(--colorsUtilityYin090);
   display: block;
-  font-weight: 600; //TODO: (tokens) use token var(--fontWeights500)
+  font-weight: var(--fontWeights700);
 
   ${({ isRequired }) =>
     isRequired &&


src/components/form/form.component.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useRef, useContext } from "react";
 import { SpaceProps, PaddingProps } from "styled-system";
-
+import invariant from "invariant";
```
edleeks87: comment: unused import

src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -62,12 +64,15 @@ export interface CheckableInputProps extends CommonCheckableInputProps {
   labelAlign?: "left" | "right";
   /** When true label is inline */
   labelInline?: boolean;
+  /** [Legacy] Flag to configure component as optional in Form */
```
edleeks87, line 67: comment: you don't need to define this twice, the common interface is extended so the type exists on the `CheckableInputProps` interface as a result

src/components/form/form.stories.mdx
```diff
@@ -72,6 +72,12 @@ When either `errorCount` or `warningCount` or both are provided, summary with th
   <Story name="with errors summary" story={stories.WithErrorsSummary} />
 </Canvas>
 
+When `required` and `isOptional` are provided.
```
edleeks87, line 75: Comment: we shouldn't have this story as when the invariant is updated it will throw and all we'll see is an error etc

src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -157,6 +163,7 @@ const CheckableInput = React.forwardRef(
       onChange,
       onFocus,
       required,
+      isOptional,
```
edleeks87, line 166: comment: I don't think you don't need to pass this to the input

src/__internal__/form-field/form-field.component.tsx
```diff
@@ -129,6 +129,11 @@ const FormField = ({
       "Use `readOnly` if you require users to see validations with a non-interactive field"
   );
 
+  invariant(
```
edleeks87, line 132: praise: good idea to add a guard against setting both props 👏

edleeks87: suggestion: I think the conditional check needs tweaking slightly though. This will actually only fire when neither is set, we actually want it to fire if **both** are set at the same time

```
isRequired && isOptional

File: src/components/form/form.stories.tsx [hunk 6]
```diff
@@ -21,73 +20,19 @@ import isChromatic from "../../../.storybook/isChromatic";
 const isOpenForChromatic = isChromatic();
 
 export const DefaultWithStickyFooter = () => (
```
edleeks87, line 22: Comment: There seems to be a lot of changes to the stories here that go beyond the scope of this ticket imo. I don't think we should be removing the tabs for example as they were added to catch regressions after issues were raised. We will move them away from the public docs at some point but I don't think it should be done now

src/components/form/form.stories.tsx
```diff
@@ -96,161 +41,230 @@ export const WithFullWidthButtons = () => (
...
+        </Button>
+      }
+      errorCount={1}
+    >
+      <Textbox label="Textbox" />
+    </Form>
+  </Box>
+);
+
+export const WithOptionalAndRequired = () => (
```
edleeks87, line 149: comment: See comment below about this story

edleeks87 (commented):
Just some small tweaks needed @DobroTora. I've done an initial pass here but I'm a little stretched atm so it's probably better to try and get two other reviewers as I'm not convinced I can commit the time on this (I don't want to hold it up waiting for me etc)

src/components/textbox/textbox.component.tsx
```diff
@@ -233,7 +233,7 @@ export const Textbox = React.forwardRef(
       error,
       warning,
       info,
-      label,
+      label: "fill in all fields marked with *",
```
edleeks87, line 236: Question: why are we not passing the label text passed by the implementing teams?

src/__internal__/form-field/form-field.component.tsx
edleeks87: sorry it needs to be `!(isRequired && isOptional)`

src/components/textbox/textbox.component.tsx
edleeks87, line 1: You'll need to move `isOptional` from `TextboxProps` into `CommonTextboxProps` so that all Textbox based components include it in the interface, right now you can't pass it to `Decimal` for example. I'd also remove `[Legacy]` part of the description as it no longer is etc

edleeks87 (commented):
Sorry I had the invariant the wrong way round @DobroTora, if you change it the tests should no longer fail. We now just need to add tests for the invariant firing for all inputs

src/__internal__/checkable-input/checkable-input.component.tsx
edleeks87: This still needs doing, when I pass `isOptional` to Checkbox the TS compiler complains etc. Also noticed `InlineInputs` doesn't support it but probably should

src/components/textbox/textbox.component.tsx
edleeks87: It looks like there's still some `isOptional` props marked as `Legacy` (FormField), we should remove that part of the descriptions

edleeks87 (commented):
I've noticed `InlineInputs` doesn't support `isOptional` at all currently and that when you set the `required` prop it only adds a `*` to the `Label` it does not make the inputs required 🤦 . I think we need to make those updates in this PR, I'll ping you on slack and we can discuss potential approaches 😄

src/components/inline-inputs/inline-inputs.spec.tsx
```diff
@@ -194,12 +194,31 @@ describe("Inline Inputs", () => {
...
+    it("when there are no inputs set", () => {
+      const container = mount(
+        <InlineInputs required label="foo">
+          <div>foo</div>
+        </InlineInputs>
+      );
+
+      const label = container.find(Label);
+      expect(label.prop("isRequired")).toBe(true);
     });
```
Parsium, line 222: **comment**: I understand why we are trying to cover a sad path here, but from a user's perspective they won't be aware of React components and props, so I would remove this if its not needed for coverage.

```suggestion
```

Parsium:
The approach with `InlineInputs` looks good to me @DobroTora 👍🏼 Just a few things from me to add that I think we should address:

1. `NumeralDate` and `Select` are currently missing the `isOptional` prop
2. Docs examples showcasing optional form fields
3. The `isOptional` prop of `FormField` is [incorrectly marked as legacy](https://github.com/Sage/carbon/blob/878d79622a514f58d086329ac79f2bf1cb767ee3/src/__internal__/form-field/form-field.component.tsx#L73)

For 1. this is because of the way the components use the `FormField` component, so the `isOptional` prop of `FormField` unfortunately doesn't cascade to either component. We'll need to add a `isOptional` to the prop interfaces for both `NumeralDate` and `Select` and make sure it is passed to the underlying `Textbox`.

If you need anything, don't hesitate to get in touch 😄

Parsium:
Just a few comments from me @DobroTora. If we could also add some stories demoing the `isOptional` prop for the components that support it, that would be fab 👍🏼

src/components/select/filterable-select/filterable-select.spec.tsx
```diff
@@ -1298,6 +1298,44 @@ describe("coverage filler for else path", () => {
...
+        id="testSelect"
+        openOnFocus
+        isOptional
+      >
+        <Option value="opt1" text="blue" key="blue" />
+      </FilterableSelect>
+    );
+
+    const input = propWrapper.find(FilterableSelect);
+    expect(input.prop("isOptional")).toBe(true);
```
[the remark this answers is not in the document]
edleeks87: I think this is pretty close to true === true here as we aren't setting the `required` prop and are asserting it's not set. I agree the test needs updating but I think asserting that the label has `(Optional)` appended on the label element

DobroTora:
> 2\. Docs examples showcasing optional form fields

I really like the neat and clear format of your review @Parsium . It is a really a good guide to follow for me.

src/components/select/multi-select/multi-select.spec.tsx
```diff
@@ -1117,6 +1117,23 @@ describe("MultiSelect", () => {
...
+          <Option value="opt1" text="red" borderColor="red" fill />
+          <Option value="opt2" text="green" borderColor="green" />
+          <Option value="opt3" text="blue" />
+        </MultiSelect>
+      );
+
+      const inputs = propWrapper.find("input");
+      inputs.forEach((input) => {
+        expect(input.prop("isOptional")).toBe(true);
+      });
```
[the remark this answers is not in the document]
edleeks87: See comment above, we should assert the label is appended (you'll need to set label text)

src/components/select/filterable-select/filterable-select.spec.tsx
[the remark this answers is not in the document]
edleeks87: I think this is pretty close to true === true here as we aren't setting the `required` prop and are asserting it's not set. I agree the test needs updating but I think asserting that the label has `(Optional)` appended on the label element

src/components/select/multi-select/multi-select.spec.tsx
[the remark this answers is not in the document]
edleeks87: I think this is pretty close to true === true here as we aren't setting the `required` prop and are asserting it's not set. I agree the test needs updating but I think asserting that the label has `(Optional)` appended on the label element


```
const propWrapper = mount(
        <MultiSelect name="testSelect" id="testSelect" isOptional label="label">
          <Option value="opt1" text="red" borderColor="red" fill />
          <Option value="opt2" text="green" borderColor="green" />
          <Option value="opt3" text="blue" />
        </MultiSelect>
      );

File: src/components/numeral-date/numeral-date.component.tsx [hunk 12]
```diff
@@ -226,6 +229,11 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
 
   invariant(hasCorrectDateFormat, incorrectDateFormatMessage);
 
+  invariant(
```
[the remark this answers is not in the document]
DobroTora: Cool - happy to rm it.

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -451,6 +457,13 @@ export const FilterableSelect = React.forwardRef(
       }
     }, [textValue, filterText, textboxRef, disabled, readOnly]);
 
+    useEffect(() => {
```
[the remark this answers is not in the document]
DobroTora: Cool - happy to rm it.

src/components/textbox/textbox.component.tsx
```diff
@@ -334,7 +334,7 @@ export const Textbox = React.forwardRef(
             labelWidth={computeLabelPropValues(labelWidth)}
             id={uniqueId}
             reverse={computeLabelPropValues(reverse)}
-            isOptional={isOptional}
+            isOptional={false}
```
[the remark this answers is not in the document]
DobroTora: I think I did it for a testing purposes.

Parsium (commented):
Thanks for adding in the stories @DobroTora 👍🏼  Just a few things from me:

1. Ensure only `FormField` has an invariant error message that throws if consumer tries to make an input field both optional and required
2. Amending the `isOptional` prop tests slightly to assert that the "(optional)" suffix is added to the end of the input label.

edleeks87:
Nothing much from me that @Parsium hasn't raised already. It might also be worth adding a `Form` story with some of these inputs in with `required`/ `isOptional` set

src/components/select/multi-select/multi-select.spec.tsx
```diff
@@ -1099,6 +1100,37 @@ describe("MultiSelect", () => {
...
+      const propWrapper = mount(
+        <MultiSelect name="testSelect" id="testSelect" isOptional label="label">
+          <Option value="opt1" text="red" borderColor="red" fill />
+          <Option value="opt2" text="green" borderColor="green" />
+          <Option value="opt3" text="blue" />
+        </MultiSelect>
+      );
+      const input = propWrapper.find("input");
+      expect(input.prop("required")).toBeFalsy();
+    });
```
edleeks87, line 1114: comment (non-blocking): this might be a bit more readable with a blank line between each `it` block

src/components/select/simple-select/simple-select.component.tsx
edleeks87, line 1: comment: I think it's worth adding the same test that we have in the other Selects for Simple

src/components/numeral-date/numeral-date.stories.mdx
```diff
@@ -115,6 +115,12 @@ You can use the `size` prop to set the size of the field.
   <Story name="size" story={stories.Size} />
 </Canvas>
 
+### With isOptional label on inputs
```
edleeks87, line 118: Question: Is there a corresponding story in the relevant `tsx` file for this?

src/components/select/filterable-select/filterable-select.spec.tsx
```diff
@@ -1341,6 +1342,24 @@ describe("coverage filler for else path", () => {
   simulateSelectTextboxEvent(wrapper, "blur");
 });
 
+describe("when the isOptional prop is passed to the input", () => {
```
Parsium, line 1345: **comment**: For readability, I think just the `it` block should be fine here.

Parsium, line 1345: **comment**: For readability, I think just the `it` block should be fine here.

src/components/form/form.stories.mdx
```diff
@@ -77,7 +77,31 @@ When either `errorCount` or `warningCount` or both are provided, summary with th
...
 
 <Canvas>
-  <Story name="with both errors and warnings summary" story={stories.WithBothErrorsAndWarningsSummary} />
+  <Story
+    name="with both errors and warnings summary"
+    story={stories.WithBothErrorsAndWarningsSummary}
+  />
+</Canvas>
+
+### Optional and Required Fields
```
Parsium, line 86: **suggestion**: as the relevant optional and required props belong to `Textbox` instead of `Form`, I would amend the story description slightly so its a bit clearer, as to me it currently reads that `Form` provides these props which may be confusing.

One way we could amend this is to tweak the description like so:

```mdx
### Example with optional and required fields

<Canvas>
  <Story name="example with optional and required fields" story={/* Story with optional and required fields */>}>
</Canvas>
```

so it just demonstrates how the `Textbox`es would look in a typical form.

src/components/form/form.stories.mdx
```diff
@@ -77,7 +77,31 @@ When either `errorCount` or `warningCount` or both are provided, summary with th
...
+
+<Canvas>
+  <Story name="with optional" story={stories.WithOptionalLabel} />
+</Canvas>
+
+When `required` prop is passed.
+
+<Canvas>
+  <Story name="required" story={stories.WithRequiredLabel} />
+</Canvas>
```
Parsium, line 98: **suggestion**: I think just having the story with both the optional and required fields should be sufficient here:


```suggestion
```

Parsium (commented):
Thanks for the changes @DobroTora :+1:

Just a few small suggestions from me in addition to what @edleeks87 has mentioned :)

src/components/form/form.stories.mdx
```diff
@@ -77,7 +77,31 @@ When either `errorCount` or `warningCount` or both are provided, summary with th
...
+
+<Canvas>
+  <Story
+    name="with optional or required label"
+    story={stories.WithBothOptionalOrRequired}
+  />
```
Parsium, line 104: suggestion: I think just having the story with both the optional and required fields should be sufficient here:

```suggestion
<Canvas>
  <Story
    name="with optional or required label"
    story={stories.WithBothOptionalOrRequired}
  />
```

src/components/numeral-date/numeral-date.spec.tsx
```diff
@@ -651,6 +651,29 @@ describe("NumeralDate", () => {
...
+      const PropWrapper = shallow(
+        <NumeralDate
+          value={{ dd: "12", mm: "", yyyy: "" }}
+          isOptional={false}
+        />
+      );
+
+      const textbox = PropWrapper.find("Textbox").first();
+      expect(textbox.prop("isOptional")).toBeUndefined();
+    });
```
Parsium, line 674: **suggestion**: thanks for adding these tests @DobroTora.

As `NumeralDate` has three inputs instead of one, I think it would be good to test each of them to make sure they are all covered by tests. Would you be able to modify these two tests slightly?:

```ts
it("the isOptional prop is passed to the Textbox", () => {
  const PropWrapper = mount(
    <NumeralDate value={{ dd: "12", mm: "", yyyy: "" }} isOptional />
  );

  const textbox = PropWrapper.find("Textbox");
  textbox.forEach((node) => {
    expect(node.prop("isOptional")).toBeTruthy();
  });
});

it("the isOptional prop is passed to the Textbox and set to false", () => {
  const PropWrapper = mount(
    <NumeralDate
      value={{ dd: "12", mm: "", yyyy: "" }}
      isOptional={false}
    />
  );

  const textbox = PropWrapper.find("Textbox");
  textbox.forEach((node) => {
    expect(node.prop("isOptional")).toBeUndefined();
  });
});
```

edleeks87: I actually think we shouldn't pass [isOptional](https://github.com/Sage/carbon/blob/645b53b9953a45d26013d0b4a8b769761c9341b1/src/components/numeral-date/numeral-date.component.tsx#L509) to the individual inputs here as all we need is it append to the label which it does etc. I think if we remove this line the above test won't work, we should do the same test as we did for the Selects where we assert `(optional)` is added `::after` the label imo

```
   assertStyleMatch(
        {
          content: '"(optional)"',
          fontWeight: "350",
          marginLeft: "4px",
        },
        wrapper.find(StyledLabelContainer),
        { modifier: "::after" }
      );

**Review by edleeks87 — DISMISSED** [review]

**Review by Parsium — DISMISSED** [review]

**Comment by harpalsingh** [comment]
@DobroTora Just to check, is the "Optional" text Black 90% on purpose, as its different in the designs, but from an accessbility pov the only token is the Yin 90 so it works against the app background, but we are missing the 74 value (or we can make it 75). For now I am happy to pass the as built and we can review the color of that text with DS team if we need to add another token value, because it has to handle app background color.

**Comment by harpalsingh** [comment]
Also to add the following acceptance was not included and there not part of this review and will be moved into another ticket.

- Add the required field key to the top of every form example: 'fill in all fields marked with *'

**Comment by harpalsingh** [comment]
@DobroTora Should Checkbox / Radio (Maybe switch also) not include the required/optional work?

**Comment by harpalsingh** [comment]
As discussed with several developers all the outstanding tasks related to this work will be added to a new ticket and so we will approve this PR to be merged to get this work started.

**Comment by harpalsingh** [comment]
@edleeks87  We should not have IsOptional on ALL inputs by default, that prop should still be something a dev/designer add themselves and by default no optional/required is displayed, as that could be very confusing in layouts to see that everywhere all the time in particular right away. Spoken with @ljemmo from DS on this and confirmed that it should not be on by default in this release as we do more testing/demos.

File: src/components/checkbox/checkbox.stories.mdx [hunk 22]
```diff
@@ -137,6 +137,33 @@ This is an example of `Checkbox` with validations passed as boolean values.
   <Story name="New Validations - boolean" story={validationStories.NewBooleanValidation} />
 </Canvas>
 
+### Required
+
+You can use the `required` prop to indicate if the fields are mandatory. This can be done on 
+an idividual checkbox input or on the group level.
```
Parsium, line 143: **typo**:

```suggestion
an individual checkbox input or on the group level.
```

src/components/checkbox/checkbox.stories.mdx
```diff
@@ -137,6 +137,33 @@ This is an example of `Checkbox` with validations passed as boolean values.
...
+</Canvas>
+
+<Canvas>
+  <Story name="required - group" story={stories.CheckboxGroupRequired} />
+</Canvas>
+
+### IsOptional
+
+You can use the `isOptional` prop to indicate if the fields are optional. This can be done on 
+an idividual checkbox input or on the group level.
```
Parsium, line 156: **typo**:

```suggestion
an individual checkbox input or on the group level.
```

src/components/date/date.spec.tsx
```diff
@@ -1167,22 +1169,34 @@ describe("Date", () => {
...
-  describe("required", () => {
-    let input: ReactWrapper;
-    let label: ReactWrapper;
-
-    beforeAll(() => {
-      wrapper = render({ label: "required", required: true });
-      input = wrapper.find("input");
-      label = wrapper.find(Label);
-    });
+  it("should set the required attribute on the input when prop is true", () => {
```
Parsium, line 1172: **praise**: like how clear these tests are :+1:

src/components/form/required-fields-indicator/required-fields-indicator.spec.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import RequiredFieldsIndicator from ".";
+import { testStyledSystemMargin } from "../../../__spec_helper__/test-utils";
+
+describe("RequiredFiledsIndicator", () => {
```
Parsium, line 5: **typo**:

```suggestion
describe("RequiredFieldsIndicator", () => {
```

Parsium (dismissed):
Looks good to me @edleeks87 👍🏼 

| **Component** | **Has Required** | **Required demo** | **Has Optional** | **Optional demo** |
| ---- | ---- | ---- | ---- | ---- |
| Checkbox / Checkbox Group | ✅ | ✅ | ✅ | ✅ |
| Decimal Input | ✅ | ✅ | ✅ | ✅ |
| Date Input | ✅ | ✅ | ✅ | ✅ |
| Date Range | ✅ | ✅ | ✅ | ✅ |
| Fieldset | ✅ | ✅ | ✅ | ✅ |
| File Input | ✅ | ✅ | ✅ | ✅ |
| Grouped Character | ✅ | ✅ | ✅ | ✅ |
| Inline Inputs | ✅ | ✅ | ✅ | ✅ |
| Number Input | ✅ | ✅ | ✅ | ✅ |
| Numeral Date | ✅ | ✅ | ✅ | ✅ |
| Radio Button / Radio Button Group | ✅ | ✅ | ✅ | ✅ |
| SimpleSelect | ✅ | ✅ | ✅ | ✅ |
| MultiSelect | ✅ | ✅ | ✅ | ✅ |
| FilterableSelect | ✅ | ✅ | ✅ | ✅ |
| Switch | ✅ | ✅ | ✅ | ✅ |
| Text Editor | ✅ | ✅ | ✅ | ✅ |
| Textarea | ✅ | ✅ | ✅ | ✅ |
| Textbox | ✅ | ✅ | ✅ | ✅ |

src/components/fieldset/fieldset.style.ts
```diff
@@ -49,6 +53,31 @@ const LegendContainerStyle = styled.div<StyledFieldsetProps>`
...
+        }
+      `}
+
+    ${({ isOptional }) =>
+      isOptional &&
+      css`
+        ::after {
+          content: "(optional)";
+          font-weight: 350; //TODO: (tokens) use token var(--fontWeights400) - FE-6022
+          margin-left: 4px;
```
DipperTheDan, line 78: **suggestion(non-blocking):** The DS token for 4px is `spacing050` if you wanted to use that here as you've used the DS token for margin-left on the `::after` styling. Just thinking for consistency.

Parsium (dismissed):

DipperTheDan (approved):

DipperTheDan (approved):

DipperTheDan (approved):

Parsium (approved):

Parsium (approved):

DipperTheDan (approved):

sianford (approved):

DipperTheDan (dismissed):

Parsium (dismissed):

DipperTheDan (dismissed):

Parsium (dismissed):

DipperTheDan (approved):

sianford (commented):
Testing passed.

PR #6341, opened 2023-10-02
Playwright refactor textbox

tomdavies73:
### Proposed behaviour

The Textbox component should have in-browser tests using the Playwright framework

### Current behaviour

Textbox currently has Cypress tests instead of Playwright tests

### Checklist

#### QA

### Testing instructions


src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+  test("should render with prefix prop whilst align prop is also set to 'right'", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<TextboxComponent prefix="foo" align="right" />);
+
+    await expect(textbox(page)).toHaveCSS("flex-direction", "row");
+  });
+
+  /* This test could be deleted, as its specific test case cannot be properly replicated using Playwright,
```
robinzigmond, line 633: **suggestion**: as mentioned a couple of times above, this comment should be removed, even if the test is decided to be left in

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+          { textboxParentParent, secondElement }
+        );
+
+        expect(domElement1).toBeTruthy();
+        expect(domElement2).toBeTruthy();
+      });
+    }
+  );
+
+  /* This test could be deleted, as its specific test case cannot be properly replicated using Playwright,
```
robinzigmond, line 684: **suggestion**: as mentioned a couple of times above, this comment should be removed, even if the test is decided to be left in

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+
+    await expect(textbox(page)).toHaveCSS("flex-direction", "row");
+  });
+
+  /* This test could be deleted, as its specific test case cannot be properly replicated using Playwright,
+  we can assert it exists in the DOM in the specified DOM structure, but cannot properly test its visibility
+  without added complexity or adding new data tags to `Textbox`s DOM. */
+
+  ([
+    [true, "input", "label"],
```
robinzigmond, line 638: **comment**: In terms of the tests themselves - I've not looked at it in detail but again I think this can be done a bit better using a single `evaluate` to get the DOM element using the various DOM queries, and then check that's truthy (since it seems we can't check visibility directly unless we have a playwright locator).

As you imply in the comment though, I'm not sure these tests add much value so I'd not be sad to see them removed 👍

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+        expect(domElement2).toBeTruthy();
+      });
+    }
+  );
+
+  /* This test could be deleted, as its specific test case cannot be properly replicated using Playwright,
+  we can assert it exists in the DOM in the specified DOM structure, but cannot properly test its visibility
+  without added complexity or adding new data tags to `Textbox`s DOM. */
+
+  test("should render with positionedChildren prop", async ({
```
robinzigmond, line 688: **suggestion**: I could be wrong, but looking at the original Cypress test for this it seems quite different from the `reverse` one above - and I think it's possible to get a reasonable Playwright equivalent here. Again use a single `.evaluate` to get the parent element, and call `querySelectorAll` with the button locator as you're doing here, checking the length is > 0, or do as the original test did and get the child element and check that attribute directly.

Sadly we can't easily check visibility, but it's still probably worth keeping this in imo. (Happy to be disagreed with though - I don't feel that strongly.)

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+
+    expect(buttonExistsInTextboxParent).toBeTruthy();
+  });
+
+  ([
+    ["flex", 399],
+    ["flex", 400],
+    ["block", 401],
+  ] as [string, TextboxProps["adaptiveLabelBreakpoint"]][]).forEach(
+    ([displayValues, breakpoint]) => {
```
robinzigmond, line 718:
```suggestion
    ([displayValue, breakpoint]) => {
```

Again, let's use singular names for single values.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+    }
+  );
+
+  test("should render with labelId prop", async ({ mount, page }) => {
+    await mount(<TextboxComponent id={CHARACTERS.STANDARD} />);
+
+    const label = getDataElementByValue(page, "label");
+
+    await expect(label).toHaveAttribute("id", `${CHARACTERS.STANDARD}-label`);
+    await expect(label).toHaveAttribute("for", `${CHARACTERS.STANDARD}`);
```
robinzigmond, line 749:
```suggestion
    await expect(label).toHaveAttribute("for", CHARACTERS.STANDARD);
```

**suggestion** (non-blocking): really minor thing and I know it was like this in the Cypress test - but there is zero value in wrapping a strong in backticks and `${...}`.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      CHARACTERS.STANDARD
+    );
+  });
+
+  (["add", "filter", "play"] as TextboxProps["inputIcon"][]).forEach((icon) => {
+    test(`should render with inputIcon as ${icon}`, async ({ mount, page }) => {
+      await mount(<TextboxComponent inputIcon={icon} />);
+
+      const inputIcon = getDataElementByValue(page, "input-icon-toggle");
+      await expect(textbox(page).locator(inputIcon)).toBeVisible();
```
robinzigmond, line 777:
```suggestion
      await expect(inputIcon).toBeVisible(;
```

**question** (non-blocking): maybe I'm missing something but this seems much simpler. The only difference is this way doesn't need the icon to be inside the textbox div, whereas yours does - but does this actually matter? (I would think not but I'm not sure.)

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+    let callbackCount = 0;
+    await mount(
+      <TextboxComponent
+        onMouseDown={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+
+    await textboxInput(page).click({ button: "left" });
```
robinzigmond, line 955: **comment**: I'm not sure this is doing what we want - from the Playwright docs it seems that the `button` option defaults to `left` anyway (as you would expect), so this is the same as just calling `click()`.

To simulate a mousedown event it looks like you can use `page.mouse.down()` - see https://playwright.dev/java/docs/api/class-mouse#mouse-down . Although note that you probably first need to use the `move` method to get the mouse over the textbox - this can likely be done by using `getBoundingClientRect` but I'm beginning to think this is more complexity than it's worth and maybe the whole test can just be removed...

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+        await mount(<TextboxComponent size={size} />);
+
+        await expect(textbox(page)).toHaveCSS("min-height", height);
+
+        const tokenValue = await getDesignTokensByCssProperty(
+          page,
+          textbox(page),
+          "min-height"
+        );
+        expect(tokenValue).toContain(token);
```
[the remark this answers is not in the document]
tomdavies73: Good spot, I'll change. I'll keep it as `tokenValues` but just make sure we're testing against the first array in the expect.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+    const contentValue = await getStyle(
+      (labelParent as unknown) as Locator,
+      "content",
+      "after"
+    );
+
+    expect(contentValue).toContain("(optional)");
+  });
+
+  /* All data tag tests below: component, element and role can be removed as we're
```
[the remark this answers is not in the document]
tomdavies73: No worries, was just a comment for review transparency was never to be merged with master.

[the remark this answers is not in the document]
tomdavies73: Also FYI i'm going to leave these tests in, even though they don't provide a huge amount of value I feel the locators i've added will in the future.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+  });
+
+  test("should render with id prop", async ({ mount, page }) => {
+    await mount(<TextboxComponent id={CHARACTERS.STANDARD} />);
+
+    await expect(textboxInput(page)).toHaveId(CHARACTERS.STANDARD);
+  });
+
+  testData.forEach((labelVals) => {
+    test(`should render with label prop as ${labelVals}`, async ({
```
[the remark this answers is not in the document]
tomdavies73: Agree, i'll remove all examples where i've used "as" in test descriptions.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+          inputWidth={inputVals}
+        />
+      );
+
+      const label = getDataElementByValue(page, "label");
+      const labelParent = await label.evaluateHandle((element: Element) => {
+        return element.parentElement;
+      });
+
+      /* Had to create my own textbox locator here instead of using textbox(page), as that seemed to encounter
```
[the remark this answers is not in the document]
tomdavies73: Yeah added this comment for transparency, i've removed it and adopted the new styled approach.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      const inputValue = "12345678901";
+      const underCharacters =
+        limit - inputValue.length === 1 ? "character" : "characters";
+      const overCharacters =
+        inputValue.length - limit === 1 ? "character" : "characters";
+
+      await textboxInput(page).fill(inputValue);
+
+      if (inputValue.length > limit) {
+        console.log(`THE CHARACTER LIMIT FOR THIS TEST IS ${limit}`);
```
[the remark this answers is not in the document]
tomdavies73: Yeah my bad, i've removed these.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      page,
+    }) => {
+      await mount(
+        <TextboxComponent
+          enforceCharacterLimit={false}
+          characterLimit={limit}
+        />
+      );
+
+      const inputValue = "12345678901";
```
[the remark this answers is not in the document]
tomdavies73: Agreed, i've added 11 to the test params.

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+    });
+
+    await expect(textboxInput(page)).toBeFocused();
+    await checkGoldenOutline(textbox(page));
+  });
+
+  test("should render with readOnly prop", async ({ mount, page }) => {
+    await mount(<TextboxComponent readOnly />);
+
+    await expect(textboxInput(page)).toHaveAttribute("readonly", "");
```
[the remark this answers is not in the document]
tomdavies73: Agreed

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+        />
+      );
+
+      const label = getDataElementByValue(page, "label");
+
+      const labelParent = await label.evaluateHandle((element: Element) => {
+        return element.parentElement;
+      });
+
+      const dataElementExistsInLabelParent = await labelParent.evaluate(
```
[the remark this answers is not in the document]
tomdavies73: I feel like in the equivalent test we are asserting that the label exists inside the label's parent element ` getDataElementByValue("label")
          .parent()
          .find(ICON)
          .should("have.attr", "data-element", type);`

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      );
+
+      await expect(textbox(page)).toHaveCSS("border-bottom-color", borderColor);
+    });
+  });
+
+  test("should render with leftChildren prop", async ({ mount, page }) => {
+    await mount(<TextboxComponentWithLeftChildren />);
+
+    await expect(textbox(page).locator(BUTTON)).toHaveAttribute(
```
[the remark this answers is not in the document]
tomdavies73: Agree, i'll just keep the visibility check

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      await mount(<TextboxComponent prefix={prefix} />);
+
+      await expect(textboxPrefix(page)).toHaveText(prefix);
+      await expect(textboxPrefix(page)).toHaveCSS("font-size", "14px");
+      await expect(textboxPrefix(page)).toHaveCSS("font-weight", "900");
+      await expect(textboxPrefix(page)).toHaveCSS("margin-left", "12px");
+    });
+  });
+
+  test("should render with prefix prop whilst align prop is also set to 'right'", async ({
```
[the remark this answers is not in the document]
tomdavies73: So the component code doesn't make CSS attribute changes based on the align prop being either left or right in this scenario. Instead, It's if the `prefix` is present or not, with the align prop always set to `"right"`. I'll amend the test to check the CSS attribute value when the prefix is passed or not.

src/components/textbox/textbox.pw.tsx
tomdavies73: Technically no, but I don't see the harm in keeping this in either. I might just leave it as i think its easier to understand then just locating the inputIcon

src/components/textbox/textbox.pw.tsx
tomdavies73: Yeah, going to just remove this one, test is adding no real value and its scenario is probably adequately covered already via unit test

src/components/textbox/textbox.pw.tsx
tomdavies73: Good shout, managed to cut down the entire test into this `    const dataElementExistsInLabelParent = await textbox(page).evaluate((element) => element.parentElement?.querySelectorAll(`[data-component="button"]`)?.length);
    expect(dataElementExistsInLabelParent).toBeGreaterThan(0);`

src/components/textbox/textbox.pw.tsx
robinzigmond: LGTM!

src/components/textbox/textbox.pw.tsx
```diff
@@ -247,47 +241,40 @@ test.describe("Prop checks for Textbox component", () => {
...
-          (textboxElement as unknown) as Locator,
-          "width",
-          inputRatio
-        )
+      const inputRatioValue = await page.evaluate(
+        (element) => window.getComputedStyle(element).getPropertyValue("width"),
+        textboxJsHandle
       );
+
+      expect(parseInt(labelRatioValue)).toBeGreaterThanOrEqual(labelRatio - 2);
```
robinzigmond, line 276: **question** (non-blocking): are we sure we need to allow a window of 2 either side of the actual value? I know we can't guarantee it being bang on but this seems a wider range than I'm comfortable with. Have we tried it with a gap of say 1, or even 0.5, on either side?

tomdavies73: Yeah I don't object to making this a little stricter

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1125 @@
...
+      );
+      expect(tokenValues[0]).toBe(token);
+    });
+  });
+
+  test("should render with isOptional prop", async ({ mount, page }) => {
+    await mount(<TextboxComponent isOptional />);
+
+    const label = getDataElementByValue(page, "label");
+    const labelParent = await label.evaluateHandle((element: Element) => {
```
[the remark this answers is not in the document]
tomdavies73: Yeah i'll see if I can make this more succinct, if I can't i'll keep the same approach

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1125 @@
...
+        await mount(<TextboxComponent size={size} />);
+
+        await expect(textbox(page)).toHaveCSS("min-height", height);
+
+        const tokenValues = await getDesignTokensByCssProperty(
+          page,
+          textbox(page),
+          "min-height"
+        );
+        expect(tokenValues[0]).toContain(token);
```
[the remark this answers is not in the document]
robinzigmond: I can't quite remember how the util formats it, I think it strips out everything but the token name. But regardless, we should check for an exact match here, otherwise we don't know it's right - same as we do in the other token tests. That means checking against something like `(${token})` if the output does contain brackets, or whatever.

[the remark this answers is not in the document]
tomdavies73: Yeah ignore me, `toBe` works fine

robinzigmond (approved):

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1127 @@
...
+      }
+    });
+  });
+
+  ([
+    [4, "with", "is"],
+    ["", "without", "is not"],
+  ] as [TextboxProps["characterLimit"], string, string][]).forEach(
+    ([characterLimitVal, renderStatus, characterLimitStatus]) => {
+      test(`should render ${renderStatus} a character count hint when the characterLimit prop ${characterLimitStatus} passed`, async ({
```
stephenogorman, line 308: There were three cypress tests for this prop, but only one in playwright. Have the others been removed because they're unnecessary?

src/components/textbox/textbox.pw.tsx
```diff
@@ -0,0 +1,1127 @@
...
+    const labelParent = await label.evaluateHandle((element: Element) => {
+      return element.parentElement;
+    });
+
+    const justifyContentValue = await page.evaluate(
+      (element) =>
+        window.getComputedStyle(element).getPropertyValue("justify-content"),
+      labelParent
+    );
+    expect(justifyContentValue).toEqual("flex-end");
```
stephenogorman, line 164: Regarding our conversation yesterday, the following code does the same verification more simply, with the caveat that it uses `locator("..")` to find the parent element:
``const labelParent = getDataElementByValue(page, "label").locathor("..");
await expect(labelParent).toHaveCSS("justify-content", "flex-end");``

src/components/textbox/textbox.pw.tsx
tomdavies73: Good spot, I actually added these in during my character count work which was in progress during this refactor so they've been missed. I'll add them in in a seperate commit

robinzigmond (dismissed):

robinzigmond (approved):

stephenogorman (approved):

PR #6345, opened 2023-10-03
feat(tile): refactor component to support more flexible composition of children FE-5681

### LINKED ISSUE #5760: Tile component rendering empty element when child element is null
### Current behaviour

When the Tile component is rendered with any JSX child element other than the first that is null, it renders an empty element with a border-top set.

### Expected behaviour

The component should not render anything when a child component is null.

https://codesandbox.io/s/carbon-tile-test-forked-wcc7rc?file=/src/index.js

### JIRA Ticket (Sage Only)

SBS-49777

### Suggested Solution

_No response_

### Carbon Version

114.0.0

### Design Tokens Version

_No response_

MacOS

_No response_

### Confidentiality

edleeks87:
fix #5760

### Proposed behaviour

Surfaces a new `TileContent` component to support the same layout as previous `Tile` component.
Removes the iteration over the `children` to better support more flexibility in consumers’ implementations and fix issue where returning `null` in a child causes additional dividers to be rendered.

BREAKING CHANGE: You will need to wrap children in `TileContent` to achieve the layouts of the previous component where they receive padding relative to that of the `Tile` parent's.

Converts cypress to playwright.

### Current behaviour

Component does not support compositional approach, children are iterated over internally, children that return a falsy value cause layout issues.

### Checklist

#### QA

### Testing instructions


src/components/tile/tile-content/tile-content.component.tsx
Parsium, line 1: **suggestion**: what do you think about removing spreading the remaining props for the exported components?

While we already provide a bunch of props for `Tile`, `TileContent` and `TileFooter`, this may deter consumers from passing `style` and other html attributes in that we don't want to encourage.

src/components/tile/tile-content/tile-content.style.ts
```diff
@@ -0,0 +1,73 @@
...
+        margin-top: 0;
+        border-top: solid 1px var(--colorsUtilityMajor050);
+      }
+    `}
+
+    ${(width || height) &&
+    css`
+      flex-grow: 0;
+      ${computeSizing({
+        width: width || /* istanbul ignore next */ undefined,
```
DipperTheDan, line 62: **suggestion/question:** What was the reason for including the `undefined` or ignoring this? I've checked the `default` example in Storybook whilst console logging the width value for this component and it returns `undefined`. If there is a genuine reason that's all good.

Screenshot 2023-10-09 at 11 25 29

edleeks87: This is just maintaining the existing implementation that ignores when the width is set to 0 so that it fits the content etc

DipperTheDan (dismissed):

Parsium (approved):

src/components/tile/tile.pw.tsx
```diff
@@ -0,0 +1,263 @@
...
+  }) => {
+    await mount(<TileComponentWithFalsyChildren data-element="tile" />);
+
+    const tileContentCount = await getDataElementByValue(
+      page,
+      "tile-content"
+    ).count();
+    expect(tileContentCount).toEqual(1);
+  });
+});
```
sianford, line 263: Can you add the accessibility tests in please?

src/components/tile/tile.pw.tsx
```diff
@@ -0,0 +1,263 @@
...
+  ["borderWidth200", "2px"],
+  ["borderWidth300", "3px"],
+  ["borderWidth400", "4px"],
+];
+const tileRoundness: TileProps["roundness"][] = ["default", "large"];
+
+test.describe("Tile component", () => {
+  tileVariants.forEach(([variant, borderColor, backGroundColor]) => {
+    test(`check ${variant} variant`, async ({ mount, page }) => {
+      await mount(<TileComponent variant={variant} data-element="tile" />);
```
sianford, line 77: Can you please add an empty line after the mount to keep it in line with the style of the rest of the suite?

src/components/tile/tile.pw.tsx
```diff
@@ -0,0 +1,263 @@
...
+      await expect(tile).toHaveCSS("border-color", borderColor);
+      await expect(tile).toHaveCSS("background-color", backGroundColor);
+    });
+  });
+
+  tileOrientations.forEach(([orientation, height]) => {
+    test(`check ${orientation} orientation`, async ({ mount, page }) => {
+      await mount(
+        <TileComponent orientation={orientation} data-element="tile" />
+      );
```
sianford, line 88: As above, can you please add an empty line after the mount to keep it in line with the style of the rest of the suite?

src/components/tile/tile.pw.tsx
```diff
@@ -0,0 +1,263 @@
...
+      const tile = getDataElementByValue(page, "tile");
+
+      await assertCssValueIsApproximately(tile, "height", heightInPixel);
+    });
+  });
+
+  testData.forEach((align) => {
+    test(`check ddTextAlign as ${align}`, async ({ mount, page }) => {
+      await mount(<DlTileComponent ddTextAlign={align} data-element="tile" />);
```
sianford, line 122: As above, can you please add an empty line after the mount to keep it in line with the style of the rest of the suite?

src/components/tile/tile.pw.tsx
```diff
@@ -0,0 +1,263 @@
...
+      const dd = getDataElementByValue(page, "dd").nth(0);
+
+      await expect(dd).toHaveCSS("text-align", align);
+    });
+  });
+
+  testData.forEach((align) => {
+    test(`check dtTextAlign as ${align}`, async ({ mount, page }) => {
+      await mount(<DlTileComponent dtTextAlign={align} data-element="tile" />);
```
sianford, line 131: As above, can you please add an empty line after the mount to keep it in line with the style of the rest of the suite?

src/components/tile/tile.pw.tsx
edleeks87: I've missed some but I'd argue most add no value, what accessibility issue are we testing for when a `Dl` is in a `Tile` with text-align updated for example?

DipperTheDan (approved):

Parsium (approved):

src/components/tile/tile.pw.tsx
sianford: Can you wrap the accessibility tests in a `describe` please to group them together? It makes them more readable and easier to run just that group of tests in either VS Code or the Playwright UI if need be.

In terms of the value of these, some may not be needed. They would need analysing on a case by case basis to determine which ones were, and how we can set up a component to check for things that may not be detectable in its default state. Such as making sure the Accordion example is extended first before asserting (which is not included as a test here), otherwise it would bring no more value than the default test of the Accordion itself. We may miss regressions if we make assumptions on what could possibly cause a violation. For instance, not having a test for the `with TileFooter` example containing buttons as it does in Storybook would result in missing the fact that there is currently a violation in that example due to the contrast ratio between the buttons and the background in the footer. This is the only opportunity we have of catching those regressions other than during a manual test which will not be done unless this is the component that changes are being made on.

sianford (approved):

Parsium (approved):

DipperTheDan (approved):

src/components/tile/tile.pw.tsx
edleeks87: Agreed on slack to wait until we have a standardised approach from QA before making this change

PR #6348, opened 2023-10-05
fix(link): refactor styles to remove focus-within due to issues with JSDOM FE-6217

### LINKED ISSUE #6320: Focus-Within usage within new Link styling
### Current behaviour

119.11.0 introduced new focus sttlying for Link:
https://github.com/Sage/carbon/commit/ae68d8debbfcfa4239096569e738de95773d3d16#diff-4bc76552ffc1900364db75add9c449feb3aedea8878cb1f0255d5816f8c5247fR169

and added focus-within.
This has known issues with jsdom and React Testing Library meaning the UI cannot query role/text of any Link component without opting out of the redesign.

### Expected behaviour

Link styling to use focus or other alternative without known issues with jsdom

https://carbon.sage.com/?path=/docs/link--default-story

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

Link styling to use focus or other alternative without known issues with jsdom

### Carbon Version

119.11.0 and higher

### Design Tokens Version

_No response_

Chrome

MacOS

Chatted with @Parsium on this matter

### Confidentiality

edleeks87:
fix #6320

### Proposed behaviour

Refactor new focus styling for `Link` to avoid `focus-within` as it creates issues when running tests in JSDOM.
Ensures `text-decoration` is `none` when new focus styling applied
Fixes test story `icon` dropdown

image

### Current behaviour

New focus styling uses `focus-within`
`text-decoration` is `underline` when new focus styling applied
Test story `icon` dropdown is broken

image

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/link/link.component.tsx
+++ b/src/components/link/link.component.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo } from "react";
+import React, { useContext, useMemo, useState } from "react";
 
 import Icon, { IconType } from "../icon";
 import MenuContext from "../menu/menu.context";
@@ -82,6 +82,7 @@ export const Link = React.forwardRef<
     }: LinkProps,
     ref
   ) => {
+    const [hasFocus, setHasFocus] = useState(false);
     const l = useLocale();
     const { inMenu } = useContext(MenuContext);
     const handleOnKeyDown = (
@@ -141,6 +142,8 @@ export const Link = React.forwardRef<
       rel,
       "aria-label": ariaLabel,
       ...ariaProps,
+      onFocus: () => setHasFocus(true),
+      onBlur: () => setHasFocus(false),
     };
 
     const buttonProps = {
@@ -160,7 +163,6 @@ export const Link = React.forwardRef<
           ? {
               ...componentProps,
               ...buttonProps,
-              placeholderTabIndex,
             }
           : {
               ...componentProps,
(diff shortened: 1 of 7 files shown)


DipperTheDan (dismissed):

src/components/link/link.component.tsx
```diff
@@ -160,13 +161,16 @@ export const Link = React.forwardRef<
           ? {
               ...componentProps,
               ...buttonProps,
-              placeholderTabIndex,
+              onFocus: () => setHasFocus(true),
```
robinzigmond, line 164: **suggestion** (non-blocking): you could move these `onFocus` and `onBlur` into `componentProps` so you don't have to repeat them in both objects. This seems especially sensible to me since event handlers like `onClick` and `onKeyDown` are already there.

robinzigmond (commented):
Don't know if this is an oversight or something we've got another ticket for, but just in case - I've seen we still have `focus-within` styles inside `MenuItem`. I'd rather get that fixed in the same PR, unless it's significantly more effort to do that (happy to leave it up to you, I'm mostly mentioning it in case it's something you're not aware of).

robinzigmond (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6350, opened 2023-10-05
fix(pager): make top corners rounded (FE-6234)

### LINKED ISSUE #6308: Pager header not displayed correctly with Pod
### Current behaviour

The is an issue with Pager if we want to display it with the Pod component. 
See the expected design

### Expected behaviour

The expected design : 
Capture d’écran 2023-09-08 à 14 19 33

https://codesandbox.io/s/nostalgic-star-5vlw5m?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

119.10.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

Parsium:
closes #6308

### Proposed behaviour

- Modify `Pager` to have a border and rounded corners around the entire component:
Screenshot 2023-10-05 at 17 17 32

- For the edge case where `Pager` is used within the footer of a `FlatTable`, ensure this new styling is overridden, so top border and rounded corners are removed: 
Screenshot 2023-10-05 at 17 16 43

### Current behaviour

- `Pager` has no top border and its top corners are not rounded:
Screenshot 2023-10-05 at 17 22 35

### Checklist

#### QA

### Testing instructions

- Verify that Chromatic has:
   - [ ] Has not flagged any `FlatTable` stories - since styling should remain unchanged.
   - [ ] Flagged changes in all the other stories where `Pager` is used. The top border corners should be rounded.


edleeks87 (dismissed):

src/components/flat-table/flat-table.style.ts
```diff
@@ -364,6 +364,14 @@ StyledFlatTableWrapper.defaultProps = {
 const StyledFlatTableFooter = styled.div<
   Pick<FlatTableProps, "hasStickyFooter">
 >`
+  & > ${StyledPagerContainer} {
+    border-top: none;
+    border-top-width: 0;
```
robinzigmond, line 369: **suggestion** (non-blocking): I don't believe either `border-top-width: 0;` or `border-top-style: none;` are needed when we already have `border-top: none;`. And I find it clearer just to have the one style rather than 3!

src/components/flat-table/flat-table.style.ts
```diff
@@ -364,6 +364,14 @@ StyledFlatTableWrapper.defaultProps = {
...
+    border-top-style: none;
+    border-top-left-radius: var(--borderRadius000);
```
robinzigmond, line 371: **comment** (non-blocking): I know we're supposed to use design tokens where possible as opposed to hard-coded pixel values - but I'm not sure about using a token here to say "remove the border radius". Will we ever want this to be a value other than `0px`? More to the point, will the DS?

I'm not saying to change this necessarily, particularly if we use the `borderRadius000` token elsewhere - but I'd rather be clear and put just `0px` rather than make readers of the code look up a token value, since what we actually want in this case is to have no radius in the top corners (or at least I thought so)... Worth thinking about I think - but happy for you to decide either way.

src/components/flat-table/flat-table.style.ts
```diff
@@ -364,6 +364,14 @@ StyledFlatTableWrapper.defaultProps = {
 const StyledFlatTableFooter = styled.div<
   Pick<FlatTableProps, "hasStickyFooter">
 >`
+  & > ${StyledPagerContainer} {
```
robinzigmond, line 367: **comment** (non-blocking): I'm just wondering if we should unit-test these styles, since we were doing that before, and are also still unit-testing the non-table-specific Pager styles.

It's likely not actually necessary since Chromatic should capture any regressions in table pagination styles, so I'm happy to not do it - I just thought I'd mention it in case this was an oversight and you'd rather have the (slight) added confidence of some unit tests here.

edleeks87 (approved):

robinzigmond (approved):

robinzigmond (approved):

stephenogorman (approved):

PR #6355, opened 2023-10-09
feat: add new file-input component

robinzigmond:
Only supports input of a single file at present - multiple mode will follow in the future.

### Proposed behaviour

A FileInput component should exist in the Carbon library, matching the Design System designs.

### Current behaviour

There is no FileInput component

### Checklist

#### QA

### Testing instructions


src/__internal__/utils/helpers/tags/tags-specs/tags-specs.ts
```diff
@@ -22,5 +22,16 @@ const rootTagTest = (
   expect(rootNode.prop("data-role")).toEqual(role);
 };
 
+const rootTagTestRtl = (
```
edleeks87, line 25: Comment: clearly great minds think alike 😄, I've actually added the same test as part of this open PR https://github.com/Sage/carbon/pull/6345/files#diff-1bd31fff78b0f0250dcf9b00bd118dee6dbf37b35d805e3258793861ad870162. I guess whichever gets merged last should just consume the one from master

src/components/file-input/file-input.stories.tsx
```diff
@@ -0,0 +1,320 @@
...
+    <FileInput
+      label="Only accepts image files"
+      accept="image/*"
+      error={error}
+      onChange={onChange}
+    />
+  );
+};
+
+export const UploadStatusClient: ComponentStory<typeof FileInput> = () => {
```
edleeks87, line 111: comment: I've been going backwards and forwards on this a bit so happy for this to be a bit more of an open discussion around it so we can decide etc. I'm leaning towards this being baked into the component as most use cases will need this and adding an override to allow for any cases that don't.

As I say happy for to open this into a wider discussion as there's potentially scenarios I'm missing etc. I had a little go at moving it into the component and fixing the linting complaints but it may be there's a better way of doing it 🤷 (if we all agree it should be baked in obviously)

https://gist.github.com/edleeks87/a3968fbbd2684868150d5a29e0d5ef5a

src/components/file-input/file-input.style.tsx
```diff
@@ -0,0 +1,91 @@
+import styled, { css } from "styled-components";
+import type { InputProps } from "../../__internal__/input";
+import type { ValidationProps } from "../../__internal__/validations";
+import StyledTypography from "../typography/typography.style";
+
+export const StyledHintText = styled.div`
```
edleeks87, line 6: comment: I think we should just re-use the styled-components from https://github.com/Sage/carbon/blob/69729437c5b9bc5c9b449557ac09d203ec764033/src/components/textbox/textbox.style.ts. We should move these to a shared file but lets do that in a new PR

edleeks87 (commented):
Great work @robinzigmond 👍. I've noticed some potentially minor styling that might need updating but as you have a UX review/ chat with Ds planned I'll leave that stuff up to them as I've confirmed it all functions as I'd expect etc

src/components/file-input/file-input.stories.tsx
nineteen88: I definitely think it's worth a discussion, but I am wondering whether leaving out this additional complexity is the better approach - as it's an extra aspect to maintain beyond primarily just the UI of it. Thinking about this some more I think we did say that the implementation detail should be down to whichever product is consuming it, and I think offering an example of how it could be used (as Robin has done nicely), should be enough.

But yeah, let's get something in the calendar though as I'm not beholden to that - and it would be good to shore up our approach to handling implementation detail like this in the future.

src/components/file-input/file-input.spec.tsx
```diff
@@ -0,0 +1,390 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { rootTagTestRtl } from "../../__internal__/utils/helpers/tags/tags-specs/tags-specs";
+import { assertStyleMatch } from "../../__spec_helper__/test-utils";
+
+import FileInput, { FileUploadStatusProps } from ".";
+
+describe("FileInput component", () => {
```
nineteen88, line 9: suggestion(non-blocking): superfluous describe block

As mentioned before

src/components/file-input/file-input.style.tsx
```diff
@@ -0,0 +1,91 @@
...
+  margin-top: 0px;
+  margin-bottom: 8px;
+  color: var(--colorsUtilityYin055);
+  font-size: 14px;
+`;
+
+export const ErrorBorder = styled.span`
+  position: absolute;
+  z-index: 6;
```
nineteen88, line 15: question: what is this z-index used for?

It stood out because 6 seems to be a bit random to me, but I've tried removing it on storybook and it doesn't make a difference.

src/components/file-input/file-input.style.tsx
```diff
@@ -0,0 +1,91 @@
...
+      css`
+        display: flex;
+        flex-direction: column;
+        flex-wrap: wrap;
+        justify-content: center;
+        align-content: center;
+        align-items: center;
+        text-align: center;
+        gap: 12px;
+        border-radius: 4px;
```
nineteen88, line 79: question: could this be a design token?

I'm just wondering if it's possible that there's consistency here that DS want standard rounded corners across components. It's more in case they want to square off corners again it would be easier with a design token... But maybe I'm just worrying about something that won't happen

nineteen88 (commented):
Praise: Looks really good Robin. Top work!

Just a few nitpicky comments more than anything

src/components/file-input/file-input.component.tsx
```diff
@@ -0,0 +1,256 @@
...
+              {textOnButton}
+            </ButtonMinor>
+          </StyledFileInputPresentation>
+        </Box>
+      </>
+    );
+
+    return (
+      // do not support old-style validation for File Input component
+      <CarbonProvider validationRedesignOptIn>
```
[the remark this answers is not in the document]
robinzigmond: good question - I added this at an early stage after seeing that things like InputPresentation checked the flag. But I didn't end up using that component - so it may be that this isn't necessary any more. I'll check - it would be nice to get rid of it as it will hopefully be going away soon(ish) anyway!

src/components/file-input/file-input.stories.tsx
robinzigmond: Happy to put a meeting in for next week on this - I agree it's a tricky question.

For what it's worth I started off wanting to have this baked in as a default, that consumers are allowed to override - but as I started to implement it the level of complexity became such that I really wasn't comfortable putting so much code in basically a big `if` statement (it would have ended up being more than half the component's code I think!), especially as I suspect there would have been other complications. And I'm not even sure that many consumers will want to do this anyway (I really have no idea though), so is there massive benefit? Documenting it in Storybook seemed - and still seems - to be the best compromise to me.

Regarding the eslint-disable - if you know a way to get around that @edleeks87 please let me know! I soon convinced myself it wasn't possible because there are functions that need to reference each other (not by direct recursion, but by adding/removing event listeners) so that whichever order they're defined in, eslint isn't happy. I don't think it's a big deal here to be honest but if there is a way to keep eslint happy without disabling it and without introducing more problems I'd be very happy to put that in!

src/components/file-input/file-input.style.tsx
robinzigmond: yeah it's a good point, that I'm happy to raise as a ticket. I added it again here  because we already have this exported from many places (TextBox, Switch, CheckboxGroup and RadioButtonGroup) so I thought this was the standard way. Happy to tidy up to make it DRYer if we agree that's the better approach (I'm pretty sure I do btw), but it shouldn't be part of the FileInput work I don't think.

src/components/file-input/file-input.stories.tsx
```diff
@@ -0,0 +1,320 @@
...
+import React, { useState, useRef } from "react";
+import { ComponentStory } from "@storybook/react";
+
+import FileInput, { FileUploadStatusProps } from ".";
+
+export const Default: ComponentStory<typeof FileInput> = () => {
+  return <FileInput label="File input" />;
+};
+
+export const Margins: ComponentStory<typeof FileInput> = () => {
```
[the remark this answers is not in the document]
robinzigmond: Fair comment. I added it because basically the first several stories I added were "copies" of those in Textbox that still seemed relevant for FileInput. I wasn't even sure whether to have the Margin Props, or whether I should add Padding Props too - I expected quite a discussion on that tbh 😁 I'll hold my hands up and say I just went for consistency with Textbox (which also has this story), without thinking about it too much - that could be the wrong approach as they're obviously pretty different as components!

Having made my excuses, I agree this story isn't really adding much for consumers so happy to remove it 👍

src/components/file-input/file-input.stories.mdx
```diff
@@ -0,0 +1,221 @@
...
+
+The `uploadStatus` prop is used to indicate when one or more files have been added, as well as the current status/progress of
+the upload - and allowing users to cancel or remove uploaded files. It is required to set this (to an actual props object, or to
+a non-empty array for multiple files) in order for the users to see any indication of the uploaded file(s).
+
+Here is a relatively simple example of using a [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) to track progress purely
+on the client side, resulting in storing the file locally as a `data:` URL. This is a way to show users a preview of their file if you do not intend
+to track upload progress on the server, or do not actually send the files to a server until form submission.
+
+Note the use of a `ref` to store the `FileReader` object - this is not essential but avoids having to create a new such object more than once in the
```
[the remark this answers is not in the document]
robinzigmond: Yeah, I agree it reads awkwardly with that word in 👍 Not sure if I like just removing it though, as "a new object" isn't really being specific enough imo - the idea was to basically say "a new FileReader object" without repeating myself. I'll have a bit of a play and hopefully come up with a better way of phrasing it.

src/components/file-input/__internal__/file-upload-status/file-upload-status.component.tsx
```diff
@@ -0,0 +1,135 @@
...
+  (StatusUploadingProps | StatusErrorProps | StatusDoneProps);
+
+export const FileUploadStatus = ({
+  status,
+  filename,
+  message,
+  onAction,
+  ...statusProps
+}: FileUploadStatusProps) => {
+  const l = useLocale();
```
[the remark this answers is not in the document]
robinzigmond: While I completely agree in general, `l` is what we universally use for this locale object in other components so I was just copying that. I'd be more than  happy to change it but I think this needs a wider discussion - not just on what to call it but as to whether I use a different name here or we stick with consistency and perhaps update them all at once in a later piece of work.

src/components/file-input/__internal__/file-upload-status/file-upload-status.spec.tsx
```diff
@@ -0,0 +1,313 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import FileUploadStatus from ".";
+import { assertStyleMatch } from "../../../../__spec_helper__/test-utils";
+
+describe("FileUploadStatus component", () => {
```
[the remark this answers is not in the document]
robinzigmond: fair point, I was again just going for consistency, but I don't really see any issues with removing this so I'll likely just do it

src/components/file-input/file-input.component.tsx
```diff
@@ -0,0 +1,256 @@
...
+      minWidth = "273px",
+      name,
+      onChange,
+      required,
+      uploadStatus = [],
+      ...rest
+    }: FileInputProps,
+    ref: React.ForwardedRef<HTMLInputElement>
+  ) => {
+    const l = useLocale();
```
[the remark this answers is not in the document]
robinzigmond: same response as above

src/components/file-input/file-input.pw.tsx
```diff
@@ -0,0 +1,719 @@
...
+  selector: string;
+  filePath: string;
+  fileName: string;
+  fileType?: string;
+}) => {
+  const buffer = readFileSync(filePath).toString("base64");
+
+  const dataTransfer = await page.evaluateHandle(
+    async ({ bufferData, localFileName, localFileType }) => {
+      const dt = new DataTransfer();
```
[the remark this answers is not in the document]
robinzigmond: I'll admit this was basically a copy-paste from a solution I found in someone's comment on a Github issue. I'm happy to change the variable name (to something like `dataTransferObject` I guess?)

src/components/file-input/file-input.style.tsx
robinzigmond: Excellent question that I can't answer 😅 I'll hold my hand up and admit this was copy-pasted from an equivalent styled component elsewhere (almost certainly Textbox but I don't specifically remember). Similar to what @edleeks87  commented on the hint text, that could arguably be moved to a central location (but as I argued there I'd rather not do it in this PR). Most likely the z-index can be removed from the other ones too, or maybe it's specific to that component that I copied it from and just not needed here?

For now I think I'll keep things separate and just remove the z-index from this component after confirming it's not needed. Thanks for spotting it.

src/components/file-input/file-input.style.tsx
robinzigmond: Good question. I grabbed the ones I could from Figma but mostly it's only colours that that seems to show tokens for. I've now had a catch up with Laia from the DS team and she's pointed out a few tokens I missed (in terms of the colour etc actually being wrong, I mean) and has sent me a full list of them all - so absolutely any styles that come from tokens will be done properly.

src/components/file-input/file-input.stories.tsx
robinzigmond: actually @edleeks87 I only just looked at your gist and it seems pretty much OK 👍 (Not that I've tested it yet, but I'd be surprised if there were any major problems.) The only thing is it will still run all this code if consumers provide their own `onChange`, which it shouldn't - however now I think about it, just adding an `if (onChange) return;` early in `onInputChange` should be all it takes to do that. I'm now coming round more to your point of view and am in 2 minds - going to put in a short meeting to discuss this.

src/components/file-input/file-input.style.tsx
robinzigmond: in fact, similar to Ed's suggestion above with the StyledHintText, I'm just going to remove this entirely and inport this styled component from `textbox.style`, as several components already do (and pending a later tidy-up that moves these to a common file).

src/components/file-input/file-input.style.tsx
edleeks87: Yeah agree not part of this work but definitely we should raise a new ticket to consolidate those (I missed that it had already been duplicated elsewhere tbh)

src/components/file-input/__internal__/file-upload-status/file-upload-status.component.tsx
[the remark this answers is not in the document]
edleeks87: Agree with both of you if that's possible, we should be as descriptive as possible but Robin has followed how we have implemented this elsewhere and how we've documented in our own docs so we should look to update them in separate ticket

nineteen88 (dismissed):

src/components/file-input/__internal__/file-upload-status/file-upload-status.style.tsx
```diff
@@ -65,6 +65,7 @@ export const StyledFileUploadStatus = styled.div<StyledFileUploadStatusProps>`
       ${hasError &&
       `&& ${StyledTypography} {
         color: var(--${colorToken});
+        font-weight: 500;
```
edleeks87, line 67: Comment: This could be `fontWeights500`

robinzigmond: We apparently can't do this yet due to an ongoing issue with the font-weight tokens - so I'll leave this as it is for now and we'll update this later along with all the other components using hard-coded font weights.

edleeks87 (dismissed):

nineteen88 (dismissed):

edleeks87 (approved):

nineteen88 (approved):

src/components/file-input/__internal__/file-upload-status/file-upload-status.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ) : (
+    <>
+      <Icon type={iconType} />
+      <span>{filename}</span>
+    </>
+  );
+  const mainRow =
+    status !== "previously" ? (
+      <StyledFileUploadStatusRow>
+        <StyledTypography as="p" mb={0} aria-live="polite">
```
edleeks87, line 108: question: is this something we could just support in general for `Typography` rather than importing the SC here?

src/components/file-input/file-input.spec.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+  it("accepts data tag props", () => {
+    render(
+      <FileInput
+        data-element="element-test"
+        data-role="role-test"
+        onChange={() => {}}
+      />
+    );
+    const wrapperElement = screen.getByText("or drag and drop your file")
```
edleeks87, line 66: Suggestion (non-blocking): might be worth storing this in a variable as it's used quite a lot, that way if the text is ever updated it only needs changing once in this file.

src/components/file-input/__internal__/file-upload-status/file-upload-status.component.tsx
edleeks87: Let's address this in the future (if we need it at all) as this work is urgent etc

src/components/file-input/file-input.spec.tsx
edleeks87: Likewise happy to leave this as is and it can be addressed in the future if needed

edleeks87 (dismissed):

nineteen88 (dismissed):

harpalsingh:
Had an additional UXQA on this after updates and happy with the work again.

nineteen88 (dismissed):

edleeks87 (dismissed):

nineteen88 (dismissed):

edleeks87 (approved):

nineteen88 (approved):

divyajindel (approved):

PR #6361, opened 2023-10-11
fix(pod): make edit button an HTML button rather than an anchor tag

robinzigmond:
Fix accessibility by ensuring the edit button is an actual button, rather than an anchor tag with no href. This makes it possible to interact with the button with the keyboard as well as with a screenreader.

BREAKING CHANGE: The onEdit prop is now restricted to being an event handler function - the options for it to be a string or an object have been removed.

### Proposed behaviour

When the `onEdit` prop is provided, the Edit button should be an HTML `<button>` element, allowing it to be focused with the keyboard and to be accessible using a screenreader. There should be no Axe errors from such a component.

Only functions should be valid values for this prop.

### Current behaviour

The edit button is rendered as an HTML `<a>` tag but without an `href`, which is a non-interactive element - therefore it is not possible to trigger the edit action using a screenreader (Voiceover on Mac announces it as simply "empty group"), nor can it be focused by using the tab/shift-tab keys on a keyboard. Axe reports an error "elements must only use valid ARIA attributes", as we pass an `aria-label` which is not valid (and not even read out by many screenreaders) for such a non-semantic element.

Strings and objects are valid to use for the `onEdit` prop's value, although they do result in a deprecation warning.

### Checklist

#### QA

### Testing instructions


DipperTheDan (commented):
Everything looks OK to me. Just this what we discussed privately basically. Looks like the `edit` button can now be focused but it appears to be the incorrect focus styling (Focus Redesign set to true on the screenshot btw). It looks like this will have existed before your change but now that `edit` button can be focused, this has been brought to light. 

Screenshot 2023-10-12 at 13 17 55

Parsium:
Looks good to me, nothing more to add in addition to @DipperTheDan's comment

Parsium (approved):

DipperTheDan (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6364, opened 2023-10-11
test(decimal): refactor cypress to playwright

sianford:
### Checklist

#### QA

### Testing instructions

- [x] Run Note tests in Playwright Test Runner to check if the decimal.pw.tsx file passed
- [x] Run all other tests in Playwright Test Runner to check none of the other *.pw.tsx files have regressed


playwright/components/textbox/index.ts
DipperTheDan, line 1: **suggestion(non-blocking):** Just a word of warning here. I think you're going to have some conflicts with [the Textbox RW refactor PR ](https://github.com/Sage/carbon/pull/6341/files). So it might be best aligning with that PR on these just so you work or @tomdavies73 doesn't get in each other's way.

src/components/decimal/decimal.pw.tsx
```diff
@@ -0,0 +1,630 @@
...
+          "value",
+          ""
+        );
+      });
+    }
+  );
+});
+
+test.describe("check events for Decimal component", () => {
+  const inputValue = "123";
```
DipperTheDan, line 410: **suggestion(non-blocking):** As `inputValue` is only used in the `should call onBlur callback when a blur event is triggered` test, you could move this from here to inside of that test block. Having it here does no harm, so if the QA reviewer wants you to keep it here as you've agreed this pattern of declaring `const`'s, I'm fine with that.

divyajindel:
Hi @sianford , could you please delete decimal.cy.tsx file in this PR?

src/components/decimal/decimal.pw.tsx
sianford: Well spotted, I've moved that const within that test case now.

DipperTheDan (approved):

divyajindel (approved):
Looks good to me

PR #6368, opened 2023-10-12
fix(date): add aria labels to picker navigation buttons and fix other accessibility issues FE-4075

### LINKED ISSUE #3969: Date Range not accessible through keyboard
### Current behaviour
See https://carbon.sage.com/?path=/docs/experimental-date-range--default-story
When navigating using keyboard, dates can't be selected using tab key. 

### Expected behaviour
When navigating using keyboard, a specific date can be selected.

### Reproducible example
Go to https://carbon.sage.com/?path=/docs/experimental-date-range--default-story and use tab key to navigate in the page.

### Your environment

| Browser          | Google Chrome|
| Operating System | Wondows 10|
DatePickerNotAccessible

### LINKED ISSUE #5804: Accessibility issues on DateInput component
### Current behaviour

Have few Axe Accessibility Issues mentioned in SBS-50673

image

### Expected behaviour

Resolved Accessibility Issues

https://carbon.sage.com/?path=/docs/date-input--default-story

### JIRA Ticket (Sage Only)

SBS-50673

### Suggested Solution

_No response_

### Carbon Version

114.2.0

### Design Tokens Version

2.59.0

Chrome

Windows

_No response_

### Confidentiality

### LINKED ISSUE #6324: DatePicker: support closure on ESC press
### Desired behaviour

The DatePicker should close when ESC is pressed

### Current behaviour

The DatePicker can only be closed by clicking outside of the popover, which makes it difficult for keyboard users to close the modal.

### Suggested Solution

_No response_

https://codesandbox.io/s/musing-shtern-67p2mk?file=/src/App.js

_No response_

### Confidentiality

edleeks87:
fix #5804
fix #6324
fix #3969

### Proposed behaviour

Adds aria-label support to `next` and `previous` month navigation buttons in `Date` picker component, these can be set in the locale object passed via the `I18nProvider`. Removes invalid role and aria-live attributes from caption element to fix some axe warnings.

Adds support for allowing users to keyboard navigate into the picker from the input element and around the focusable elements within it. The behaviour is as follows:

- Tabbing from the input will focus the `previous` month navigation button.
- Tabbing again will focus the `next` month navigation button.
- Pressing enter or space on either of these buttons will change the currently displayed month.
- Pressing tab again will focus the day element corresponding to either the selected or current (today's) date.
- Pressing enter or space on a day element will select it, update the input value and close the picker.
- Pressing tab will close the picker and focus the next focusable element in the DOM

Increases spacing between input and picker when new focus styles are applied

### Current behaviour

Several axe warnings when picker is open.
No support for keyboard access to picker or navigation within.
Picker partially covers the focus outlines when open

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/musing-payne-hp43wf


src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -60,121 +66,209 @@ const popoverMiddleware = [
...
-        localize
-          ?.day(
-            i,
-            ["de", "pl"].some(isGivenLocale)
-              ? { width: "wide" }
-              : { width: "abbreviated" }
-          )
-          .substring(0, isGivenLocale("de") ? 2 : 3)
+  }, [l, localize]);
+  const ref = useRef<HTMLDivElement>(null);
```
nineteen88, line 117: question: if ref is being moved internal does that make this a breaking change?

Albeit it was undocumented afaik...

src/components/date/__internal__/navbar/navbar.component.tsx
```diff
@@ -13,15 +15,40 @@ export const Navbar = ({
...
-    <StyledButton onClick={() => onPreviousClick?.()}>
-      <Icon type="chevron_left" />
-    </StyledButton>
-    <StyledButton onClick={() => onNextClick?.()}>
-      <Icon type="chevron_right" />
-    </StyledButton>
-  </StyledNavbar>
-);
+}: NavbarProps) => {
+  const locale = useLocale();
```
nineteen88, line 19: praise: using a descriptive variable name rather than just `l`

src/components/date/date.spec.tsx
```diff
@@ -515,20 +524,85 @@ describe("Date", () => {
       });
     });
 
-    describe('and with the "Tab" key', () => {
-      it('then the "DatePicker" should be closed', () => {
-        expect(wrapper.update().find(DayPicker).exists()).toBe(true);
-        simulateOnKeyDown(wrapper, "Tab");
-        expect(wrapper.update().find(DayPicker).exists()).toBe(false);
+    it('the "DatePicker" should remain open and the previous month navigation button should be focused when the "Tab" key is pressed', () => {
```
nineteen88, line 527: nitpick(non-blocking): if not in a describe block then probably should just use `test` rather than `it`

This would also read better than `it` which is more of a user story structure

src/components/date/__internal__/date-picker/date-picker.component.tsx
edleeks87: No the ref was only previously set in the date component (I don't even think it was passed to the picker), so there's no change in the exposed interface here

src/components/date/date.spec.tsx
edleeks87: I agree with the point you're making, I'd prefer to leave it and make the changes across the whole file rather than this one block as I think there's likely a fair few similar one's in the rest of the file

nineteen88: I guess the reason I mentioned these ones was because it looked like new tests and only ones out of the describe block but happy to defer

src/components/date/__internal__/date-picker/date-picker.component.tsx
nineteen88: I guess I missed that this was an internal component only as well. No problem!

nineteen88 (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6369, opened 2023-10-12
test(icon-button): add playwright tests

divyajindel:
### Proposed behaviour

- Refactor cypress tests to playwright

### Current behaviour

- Tests are currently using cypress

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- [x] Run Icon-Button tests in Playwright Test Runner to check if the icon-button.pw.tsx file passed
- [x] Run all other tests in Playwright Test Runner to check none of the other *.pw.tsx files have regressed


src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }) => {
+    await mount(<IconButtonComponent aria-label={CHARACTERS.STANDARD} />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
+
+    await expect(iconLocator).toHaveAttribute(
+      "aria-label",
+      CHARACTERS.STANDARD
+    );
+    await iconLocator.isVisible();
```
stephenogorman, line 56: I think `await expect(iconLocator).toBeVisible()` is more reliable than this.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  test(`should render IconButton with disabled prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent disabled />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
+    await iconLocator.isDisabled();
+    await iconLocator.getAttribute("disabled");
```
stephenogorman, line 77: This would be preferable:
`await expect(iconLocator).toBeDisabled()`
`await expect(iconLocator).toHaveAttribute("disabled", /.*/);`

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+    await mount(
+      <IconButtonComponent
+        onFocus={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    const iconLocator = page.getByTestId("icon").locator("..");
+    await iconLocator.focus();
+    await iconLocator.blur();
```
stephenogorman, line 114: This line `await iconLocator.blur();` is not needed.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent disabled />);
+
+    await checkAccessibility(page);
+  });
+});
+
+test(`render with the expected border radius`, async ({ mount, page }) => {
```
stephenogorman, line 220: You could move this test to the first describe, with the border radius tests.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { checkAccessibility } from "../../../playwright/support/helper";
+
+test.describe("when focused", () => {
+  test(`should have the expected styling when the focusRedesignOptOut is false`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent aria-label={CHARACTERS.STANDARD} />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 15: This const is used several times. Can it be declared once outside of the tests?

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+    page,
+  }) => {
+    await mount<HooksConfig>(
+      <IconButtonComponent aria-label={CHARACTERS.STANDARD} />,
+      {
+        hooksConfig: { focusRedesignOptOut: true },
+      }
+    );
+
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 36: As above, this const is used several times. Can it be declared once earlier then just iconLocator called where needed?

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+});
+
+test.describe("check props for IconButton component", () => {
+  test(`should render IconButton with aria-label prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent aria-label={CHARACTERS.STANDARD} />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 50: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+    await iconLocator.isVisible();
+  });
+
+  test(`should render IconButton with disabled prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent disabled />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 75: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }) => {
+    let callbackCount = 0;
+    await mount(
+      <IconButtonComponent
+        onBlur={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 94: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }) => {
+    let callbackCount = 0;
+    await mount(
+      <IconButtonComponent
+        onFocus={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 112: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }) => {
+    let callbackCount = 0;
+    await mount(
+      <IconButtonComponent
+        onMouseEnter={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 130: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+  }) => {
+    let callbackCount = 0;
+    await mount(
+      <IconButtonComponent
+        onMouseLeave={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 147: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }) => {
+      let callbackCount = 0;
+      await mount(
+        <IconButtonComponent
+          onClick={() => {
+            callbackCount += 1;
+          }}
+        />
+      );
+      const iconLocator = page.getByTestId("icon").locator("..");
```
sianford, line 184: As above, this const is used several times.

src/components/icon-button/icon-button.pw.tsx
divyajindel: Hi Sian, IMO we can't do this as we are using page to create this locator inside the test.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,233 @@
...
+      );
+    });
+
+    test(`render with the expected border radius`, async ({ mount, page }) => {
+      await mount(<IconButtonComponent />);
+
+      const iconLocator = page.getByTestId("icon").locator("..");
+      const element = await iconLocator;
+      await element.focus();
+      await expect(iconLocator).toHaveCSS("border-radius", "4px");
```
stephenogorman, line 54: The unnecessary line `const element = await iconLocator` was removed from the previous two tests, but it's still in this test.

stephenogorman (changes requested):

stephenogorman (dismissed):

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,232 @@
...
+  test(`should pass accessibility tests for aria-label prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent aria-label={CHARACTERS.STANDARD} />);
+
+    await checkAccessibility(page);
+  });
+
+  test(`should pass accessibility tests for children prop`, async ({
```
tomdavies73, line 215: **suggestion** (non-blocking): i know it was written like this in the cypress test, but i'd change this test description. There is no `children` prop and in the test there's only one child. Something like `should pass accessibility tests when rendered with a child` should do

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const iconLocator = page.getByTestId("icon").locator("..");
+
+    await expect(iconLocator).toHaveAttribute(
+      "aria-label",
+      CHARACTERS.STANDARD
+    );
+    await expect(iconLocator).toBeVisible();
+  });
+
+  test(`should render IconButton with children prop`, async ({
```
tomdavies73, line 74: **suggestion** (non-blocking): i know it was written like this in the cypress test, but i'd change this test description. There is no `children` prop and in the test there's only one child. Something like `should render with children` should do

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  test(`should render IconButton with disabled prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<IconButtonComponent disabled />);
+
+    const iconLocator = page.getByTestId("icon").locator("..");
+    await expect(iconLocator).toBeDisabled();
+    await expect(iconLocator).toHaveAttribute("disabled", /.*/);
```
tomdavies73, line 92: **suggestion**: I think the expect above should be enough for this, instead of the attribute tests

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,232 @@
...
+import { checkAccessibility } from "../../../playwright/support/helper";
+
+test.describe(
+  "check IconButton component focus outlines and border radius",
+  () => {
+    test(`should have the expected styling when the focusRedesignOptOut is false`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<IconButtonComponent aria-label={CHARACTERS.STANDARD} />);
```
tomdavies73, line 15: **question** (non-blocking): I can see we're passing an `aria-label` for these focus and radius tests, is there a specific reason for this, or can they be removed. I can see we do this in the cypress test too btw, so I understand why you've included them. Just could get rid if they're not neccessary

playwright/components/index.ts
```diff
@@ -25,6 +26,10 @@ export const commonDataElementInputPreview = (page: Page) => {
   return page.locator(COMMMON_DATA_ELEMENT_INPUT);
 };
 
+export const iconButton = (page: Page) => {
```
tomdavies73, line 29: **suggestion**: this can just be called `button` instead of `iconButton` as the locator is just finding `type="button"` and has no overlap with the `IconButton` component, i'd rename this here at least 👍

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,215 @@
+import React from "react";
+import { test, expect } from "@playwright/experimental-ct-react17";
+import IconButtonComponent from "./component.test-pw";
+import { iconButton } from "../../../playwright/components/index";
```
tomdavies73, line 4: **suggestion**: if you've changed the index/locator name from `iconButton` to `button` you can then do this for the import `import { button as iconButton } from "../../../playwright/components/index"`. This way you can keep the name of the locator in your tests and ensure they're still clear what's being tested etc.

src/components/icon-button/icon-button.pw.tsx
```diff
@@ -0,0 +1,215 @@
...
+      );
+
+      await iconButton(page).press(key);
+      expect(callbackCount).toBe(1);
+    });
+  });
+});
+
+test.describe("check accessibility tests for IconButton component", () => {
+  test(`should render aria-label prop for accessibility`, async ({
```
tomdavies73, line 189: **suggestion**: I  think these test descriptions have been changed and i'm not 100% sure why, I think they should all follow something like `should pass accessibility tests when rendered with an aria-label`, `should pass accessibility tests when rendered with a child`, `should pass accessibility tests when disabled` etc.

tomdavies73 (approved):

stephenogorman (approved):

PR #6373, opened 2023-10-16
feat(portrait, profile): align with design system and reduce code complexity

### LINKED ISSUE #6191: Portrait size changes when initials prop is used
### Current behaviour

When an image url is passed to the portrait of size "ML” from the props, the size of the component is 56x56   . But when there is no image url passed and just initials prop is used to show the initials of the user, the size with the same "ML" variant is 54x54 .

### Expected behaviour

The size of the portrait should be same in every case

https://carbon.sage.com/?path=/docs/portrait--default

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

118.4.0

### Design Tokens Version

4.17.0

MacOS

_No response_

### Confidentiality

tomdavies73:
fix #6191

### Proposed behaviour

Screenshot 2023-10-18 at 11 47 18

Screenshot 2023-10-18 at 11 46 06

Colour tokens, and background colour tokens have been aligned with the design system for `Profile`, and subsequently for `Portrait`. This ensures current colour contrast issues which have been observed in both components have been addressed. Also, when the `email` prop is passed on `Portrait` a `Link` component is now rendered with a `mailto: href` allowing users to send emails to the specified email addressed when clicked. Support has also been added for the `text` prop, which is additional text which appears under the email link, this is typically used for phone numbers, but can also be any other relevant read only text. 

Screenshot 2023-10-18 at 11 45 57

A new `darkBackground` variant has been added to `Profile`. This was already present in `Portait` but has been amended to represent new DS designs. This will allow consumers to use `Profile` or `Portait` in Ui which has a dark background. 

Finally, the DS class all text on `Profile` to be optional, when previously in the component they were mandatory. To address this, conditional rendering has been implemented to determine if any text children are present, if true the `Profile` component will be rendered. If false, the `Portrait` component is rendered instead. However, additional guards have also been placed to make sure the `name` prop is passed if consumers do pass text children.

Code complexity has also been greatly reduced, with both .style files for each component being simplified, with the removal of unnecessary functions, types and interfaces. Resulting in a much cleaner structure and reduction in complexity. Also, both `Portrait` subcomponents `portrait-initials` and `portrait-gravatar` have been removed, with the formers functionality replaced with just rendering a `div` with centred text, and the latters functionality being added to the `Portrait` component.

### Current behaviour

Screenshot 2023-10-18 at 11 47 32

Screenshot 2023-10-18 at 11 46 52

Currently, the `Profile` component is not aligned with the DS, and is missing some significant functionality and tokens, which result in some colour contrast issues which have been observed in our Playwright tests. `Portrait` is not in the DS technically, but we rely on `Portrait` for guidance on tokens etc. `Portrait` also supported a wide array of borders from dashed and its subcomponents were also overly complex, with two subcomponents and a style file which utilised unnecessary functions, types and interfaces. Resulting in a fairly complex component for what is only essentially a rounded or square image or `<div>`.

Now all gravatar functionality has been moved inside the component file, and the `initials` prop no longer renders an ``, instead it renders a `div` with text centred. 
### Checklist

#### QA

### Testing instructions


src/components/profile/profile.component.tsx
```diff
@@ -52,45 +63,65 @@ export const Profile = ({
...
-      <ProfileDetailsStyle size={size} hasSrc={!!src}>
-        <ProfileNameStyle size={size} data-element="name">
-          {name}
-        </ProfileNameStyle>
-        <ProfileEmailStyle size={size} data-element="email">
-          {email}
-        </ProfileEmailStyle>
-      </ProfileDetailsStyle>
-    );
+  invariant(
```
robinzigmond, line 87: **comment** (non-blocking): I'd be cautious about using `invariant` here unless you have very clear requirements that consumers should absolutely not do this. If the condition fires the entire app will crash.

This isn't a breaking change (assuming consumers were using typescript), because both `name` and `email` were required props before - but even so I'd be personally more inclined to suggest this more gently, perhaps with a `console.warn`, because I don't personally see a good reason why you couldn't leave out `name` and just have the corresponding element  be empty or not there.

src/components/profile/profile.pw.tsx
```diff
@@ -59,58 +61,178 @@ test.describe("Prop checks for Profile component", () => {
     });
   });
 
-  ([
```
robinzigmond, line 62: **question**: it looks like these tests, checking the appropriate `width` and `height` styles exist depending on the `size` prop, have been removed. Is there a good reason for that? Particularly as I see you've added tests for things like `line-height`, which I'm not saying not to do but seem a less important CSS property to test than width or height.

src/components/profile/profile.spec.tsx
```diff
@@ -37,40 +44,346 @@ describe("Profile", () => {
...
-        instance = shallow(<Profile name="" email="foo@bar.com" />);
+      it("returns an empty string if no name is found", () => {
+        instance = shallow(<Profile name="" />);
         expect(instance.find(ProfileAvatarStyle).props().initials).toEqual("");
       });
 
-      it("returns empty string when name is a space", () => {
-        instance = shallow(<Profile name=" " email="foo@bar.com" />);
-        expect(instance.find(ProfileAvatarStyle).props().initials).toEqual("");
+      test("acronymize function generates correct acronym", () => {
```
robinzigmond, line 58: **comment**: I'm  not sure I like explicitly testing the `acronymise` function individually, particularly when that means exporting it from the component file. I see it as more of an implementation detail. Can we not test the same thing by just checking the text content of the appropriate element when various `name` values are provided as prop?

src/components/profile/profile.stories.mdx
```diff
@@ -51,6 +63,14 @@ To use an image, simply pass any valid image URL as a `src` prop.
   <Story name="src" story={stories.Src} />
 </Canvas>
 
+### Gravatar
```
robinzigmond, line 66: **question**: probably one for the wider team, but even though this has always worked and we definitely shouldn't remove it as part of this PR, I'm not sure explicitly documenting the `gravatar` functionality for the first time at this stage is a good idea. I'm pretty sure we have mentioned removing/deprecating this feature (from `Portrait`, which would obviously also stop it working here).

src/components/portrait/portrait.style.tsx
```diff
@@ -1,157 +1,78 @@
...
+    height: inherit;
+    width: inherit;
 
-type StyledPortraitContainerProps = MarginProps & {
-  onClick?: (ev: React.MouseEvent<HTMLElement>) => void;
-};
+    ::before {
+      font-size: ${({ size }) => PORTRAIT_SIZE_PARAMS[size].iconDimensions}px;
+    }
+`;
```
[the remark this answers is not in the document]
tomdavies73: Good shout, i'll sort this and install the plugin too so I catch things like this going forward

src/components/portrait/portrait.pw.tsx
```diff
@@ -126,35 +133,145 @@ test.describe("Prop checks for Portrait component", () => {
...
+      }) => {
+        await mount(<PortraitDefaultComponent darkBackground={boolVal} />);
+
+        const backgroundColorTokens = await getDesignTokensByCssProperty(
+          page,
+          portraitPreview(page),
+          "background-color"
+        );
+
+        console.log(backgroundColorTokens);
```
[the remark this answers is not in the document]
tomdavies73: Doh! Yeah, not sure how I managed this but i'll make sure I remove them both

src/components/portrait/portrait.pw.tsx
```diff
@@ -392,7 +508,6 @@ test.describe("Accessibility tests for Portrait component", () => {
       await mount(
         <PortraitComponent tooltipIsVisible tooltipMessage={tooltipMessage} />
       );
-
       await checkAccessibility(page, "color-contrast");
```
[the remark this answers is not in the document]
tomdavies73: So I've left the disabled rule param in for the tooltip tests, as locally these all seem to fail to me. I know for sure that there aren't any colour contrast issues on the default tooltip colour but these still seem to fail, so I left these parameters in. Happy to remove them if you think its best, but just a bit uneasy about leaving in potentially flaky tests

src/components/portrait/portrait.pw.tsx
```diff
@@ -449,29 +564,7 @@ test.describe("Accessibility tests for Portrait component", () => {
...
       page,
     }) => {
-      await mount(<PortraitComponent size={tooltipSize} />);
-
-      await checkAccessibility(page, "color-contrast");
-    });
-  });
-
-  colors.forEach(([names]) => {
-    test(`should pass accessibility checks with a tooltip, with a ${names} background color`, async ({
```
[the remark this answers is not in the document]
tomdavies73: So for both we were just passing random colours/background-colours, a lot of which had colour contrast issues, e.g blue and red etc. I also felt these tests really held no value for the component, as these should really be tested in `Tooltip`. I felt a few tests with a tooltip rendered was enough and this was a tad overboard as well as failing colour contrast guidelines

src/components/portrait/portrait.style.tsx
```diff
@@ -1,157 +1,78 @@
...
+      ? "var(--colorsUtilityYin090)"
+      : "var(--colorsUtilityReadOnly400)"};
+  width: ${({ size }) => PORTRAIT_SIZE_PARAMS[size].dimensions}px;
+  height: ${({ size }) => PORTRAIT_SIZE_PARAMS[size].dimensions}px;
+  overflow: hidden;
+  border-radius: ${({ shape }) =>
+    shape === "square" ? "0px" : "var(--borderRadiusCircle)"};
+  border: 1px solid var(--colorsUtilityReadOnly600);
   display: inline-block;
+  flex-shrink: 0;
```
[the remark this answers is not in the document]
tomdavies73: Good question, so as a standalone component this attribute is ineffectual, but when used in `Portrait` it is wrapped by `StyledPortraitContainer` which is a flex-container. Having this attribute means the rendered `Portrait` doesn't shrink and cause a visual bug

src/components/profile/profile.component.tsx
```diff
@@ -52,45 +63,65 @@ export const Profile = ({
       return (
         <ProfileAvatarStyle
           src={src}
```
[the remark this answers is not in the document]
tomdavies73: Yeah, I agree with this, definitely something I should have seen sooner will make the component file a bit less complex 👍

src/components/profile/profile.pw.tsx
tomdavies73: Good question, I felt these were redundant though as we're testing the `Portait` component here, when I feel there's already enough tests in that file for sizes, we actually cover every type of prop in `portrait.pw.tsx` whereas this previous test was only testing the `initials` prop

tomdavies73: Essentially, I felt repeating the test here was redundant

src/components/profile/profile.stories.mdx
tomdavies73: Agree, i'll remove this story example

src/components/profile/profile.component.tsx
tomdavies73: Yeah fair point, having a name is pretty essential to the component as an email or text underneath really doesn't make much sense without it. I'll leave this one for the next reviewer to get their opinion if that's ok as i'm on the fence

src/components/portrait/portrait.style.tsx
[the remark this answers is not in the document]
robinzigmond: no worries. For what it's worth, I don't remember installing that plugin and I can't see it listed in my VSCode extensions, so I'm not actually sure where it comes from 😅

src/components/portrait/portrait.pw.tsx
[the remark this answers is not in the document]
robinzigmond: we should try to fix the flakiness but that's probably a separate piece of work. I agree with not wanting to add more to it so happy to leave the color-contrast exception in for now, as long as we know it's definitely in tests which do sometimes fail for no apparent reason.

src/components/portrait/portrait.pw.tsx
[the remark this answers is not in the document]
robinzigmond: fair enough - thanks for explaining!

src/components/portrait/portrait.style.tsx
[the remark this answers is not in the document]
robinzigmond: This is already in `StyledPortraitContainer` so I don't think this is quite right. Although having looked further, I see this is used as a child inside `Profile` so I assume that's what you meant?

In which case thanks, not sure how I missed that 👍 Although now I look again at `ProfileStyle` (in the `profile` component obviously) I now see that also has a `flex-shrink: 0` and that also seems to be an outer wrapper component?

src/components/profile/profile.component.tsx
robinzigmond: yeah that's fair enough, I just thought I'd ask the question. If we're sure it makes no sense to have text/email with no name then I'm happy to keep this.

src/components/profile/profile.pw.tsx
robinzigmond: I see, I'd missed that these were effectively testing `portrait`. Happy to leave this for the next reviewer to comment on, on the one hand I agree with you, but on the other the fact that `profile` sometimes renders `portrait` could be argued to be just an implementation detail, and that if we ever refactored that we'd still want to be sure `profile` was rendering at the correct size. (Mind you Chromatic should catch any changes there.) So there's an argument both ways - personally given the uncertainty I'd always rather have too many tests than too few, but as I said I'm happy to wait for another opinion and don't object to dropping this.

src/components/portrait/portrait.style.tsx
```diff
@@ -1,157 +1,78 @@
...
-}
-
-type PortraitSizeAndShape = {
-  size: PortraitSizes;
-  shape?: PortraitShapes;
-};
-
-type StyledPortraitInitialsProps = PortraitSizeAndShape & {
-  initials?: string;
+  type: IconType;
```
edleeks87, line 15: Question: Is this used anywhere? I think you get this for free by using `styled(Icon)`

src/components/portrait/portrait.style.tsx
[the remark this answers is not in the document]
edleeks87: Question: if this is only needed when in `Profile` should we not apply it there? I understand it's not doing anything in isolation but seems pointless to apply superfluous styles

src/components/profile/profile.component.tsx
edleeks87: Yep agree invariant feels a little heavy handed here

src/components/profile/profile.pw.tsx
edleeks87: As long as we have some way of catching if this were to regress (ie passing size to Profile doesn't set Portrait) I don't mind where they are

src/components/profile/profile.stories.mdx
edleeks87: Yeah I agree, I think we can look to deprecate this soon so lets not document it and inadvertently encourage it's use

src/components/portrait/portrait.style.tsx
```diff
@@ -1,157 +1,78 @@
...
-  onClick?: (ev: React.MouseEvent<HTMLElement>) => void;
-};
+    ::before {
+      font-size: ${({ size }) => PORTRAIT_SIZE_PARAMS[size].iconDimensions}px;
+    }
+`;
 
-export const StyledPortraitContainer = styled.div<StyledPortraitContainerProps>`
+export const StyledPortraitContainer = styled.div<
+  Pick<StyledPortraitProps, "darkBackground" | "size" | "shape"> & MarginProps
```
edleeks87, line 58:
```suggestion
  StyledPortraitProps & MarginProps
```

src/components/profile/profile.pw.tsx
tomdavies73: I'll look into this as flex-shrink: 0 is definitely needed now we've moved to a new approach for `initials`, moving away from rendering an image and instead rendering a div with centred text etc. I'll make sure its applied in the correct container, and on @edleeks87's point only apply it when `Portrait` is a child of `Profile`. Apologies for mixing the names up but as you can imagine working on Profile/Portrait for a while soon its quite easy to mix them up haha

DipperTheDan (commented):
I have noticed a potential bug with this PR. We previously would only allow three characters to be displayed as `initials` in `Profile` and `Portrait` Even if you passed in more than three characters. Like this:

Screenshot 2023-10-30 at 11 04 44

However in this PR, it now appears more than three will be displayed: Like this:

Screenshot 2023-10-30 at 11 05 52

Is this behaviour expected?

Tom: Yeah just noticed we sliced the initials in the now deleted `portrait-initials.component.tsx`, i'll make sure we do the same with the new approach

src/components/portrait/portrait-initials.component.tsx
```diff
@@ -1,96 +0,0 @@
...
-    if (context) {
-      context.font = `${Math.round(canvas.width / 2.4)}px "Sage UI", Arial`;
-      context.textAlign = "center";
-
-      // Setup background and front color
-      context.fillStyle = tokens[bgColor];
-      context.fillRect(0, 0, dimensions, dimensions);
-      context.fillStyle = tokens[textColor];
-      context.fillText(
-        initials.slice(0, 3).toUpperCase(),
```
DipperTheDan, line 67: **reminder:** You'll need this so the `initials` only render three, even if a million characters are passed in 😄

src/components/portrait/portrait.pw.tsx
```diff
@@ -47,11 +50,20 @@ const portraitSizes = PORTRAIT_SIZES.map((size) => [
 ]);
 
 test.describe("Prop checks for Portrait component", () => {
+  ["SPM", "JM", "AR", "MJ"].forEach((passInitials) => {
```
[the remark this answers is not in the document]
tomdavies73: Yeah i've covered this in a unit test but I don't mind adding a Playwright test too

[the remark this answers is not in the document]
DipperTheDan: Actually, a unit test should be enough for this for now. If we discover in future, that some sort of user interaction can cause more than three to appear in the browser, then it would warrant a Playwright test.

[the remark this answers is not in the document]
tomdavies73: Just added and amended some playwright tests anyways to test the functionality, can't hurt 👍

edleeks87 (dismissed):

DipperTheDan (approved):

edleeks87 (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6379, opened 2023-10-19
docs(carousel): deprecation message for docs

nuria1110:
### Proposed behaviour

Screenshot 2023-10-19 at 14 15 18

Deprecation warning for Carousel. 

### Current behaviour

Component is currently not deprecated. 

### Checklist

#### QA

### Testing instructions


src/components/carousel/carousel.spec.tsx
```diff
@@ -26,6 +29,27 @@ function renderCarousel(props: CarouselProps, renderer = mount) {
 }
 
 describe("Carousel", () => {
+  describe("Deprecation warning", () => {
+    it("when user uses the component, a deprecation warning is raised in the console", () => {
```
nineteen88, line 33: suggestion: add `only once` to the test title

The test seems to also check if we only log out the deprecation warning once, despite the component being used twice, so I think it would be good to add that aspect to the title

```suggestion
    it("when user uses the component, a deprecation warning is raised only once in the console", () => {
```

src/components/carousel/carousel.spec.tsx
```diff
@@ -26,6 +29,27 @@ function renderCarousel(props: CarouselProps, renderer = mount) {
 }
 
 describe("Carousel", () => {
+  describe("Deprecation warning", () => {
+    it("when user uses the component, a deprecation warning is raised only once in the console", () => {
+      const loggerSpy = jest.spyOn(Logger, "deprecate");
+      jest.restoreAllMocks();
```
Parsium, line 35: **suggestion**: we don't need to restore the mocks created with `spyOn` here. Restoring them at the end of the test should be sufficient:

```tsx
const loggerSpy = jest.spyOn(Logger, "deprecate");

/* mount and expects... */

loggerSpy.mockRestore()
```

Parsium (approved):

nineteen88 (approved):

src/components/carousel/carousel.spec.tsx
```diff
@@ -26,6 +29,25 @@ function renderCarousel(props: CarouselProps, renderer = mount) {
 }
 
 describe("Carousel", () => {
+  describe("Deprecation warning", () => {
+    it("when user uses the component, a deprecation warning is raised only once in the console", () => {
+      const loggerSpy = jest.spyOn(Logger, "deprecate");
+      mount(
+        <>
+          <Carousel />
+          <Carousel />
```
divyajindel, line 38: Question: could you please tell me the reason of using <Carousel /> twice ?

nuria1110: Its to check that even if there are multiple instances of the component, the deprecation warning on the console is only fired once - this is mainly to reduce spam for consumers.

PR #6385, opened 2023-10-25
fix(popover-container): ensure that component is closed when Esc key is pressed - FE-6072

### LINKED ISSUE #6195: Escape key should close the Popover Container
### Desired behaviour

The escape key doesn't close the container, like it does with the Dialog component. It would be great if it could to make keyboard navigation easier.

### Current behaviour

The escape key does not close the container.

### Suggested Solution

_No response_

_No response_

### Confidentiality

DipperTheDan:
This fix ensures that the PopoverContainer component close when the escape key is pressed. The only exception to this is if Select is currently focused/open as we want Select to still function as it's default behaviour when escape key is pressed.

fixes #6195

### Proposed behaviour

- Component should be closed when Esc key is pressed.

### Current behaviour

- Component does not close when Esc key is pressed. 

### Checklist

#### QA

### Additional context

If the `PopoverContainer` contains a `Select` component, then the popover should not close when `Select` is focused/ `SelectList` is open as we want to preserve the default functionality of `Select` in this particular case.

### Testing instructions

- Esc key closes the `PopoverContainer`. 
- In the `complex` example, then the correct behaviour is that the `PopoverContainer` should not close when `Select` is focused and the `SelectList` is open as we want to preserve the default functionality of `Select` in this particular case.
- No other functional or styling regressions have occurred to `PopoverContainer`.

### DIFF
--- a/cypress/components/popover-container/popover-container.cy.tsx
+++ b/cypress/components/popover-container/popover-container.cy.tsx
@@ -20,7 +20,7 @@ import {
 } from "../../locators/popover-container/index";
 import { selectListText, selectText } from "../../locators/select/index";
 import CypressMountWithProviders from "../../support/component-helper/cypress-mount";
-import { keyCode } from "../../support/helper";
+import { keyCode, pressESCKey } from "../../support/helper";
 import { CHARACTERS } from "../../support/component-helper/constants";
 
 const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
@@ -280,12 +280,41 @@ context("Test for Popover Container component", () => {
       }
     );
 
+    it("should close Popover Container using escape keyboard key", () => {
+      CypressMountWithProviders(
+        <PopoverContainer title="Cypress is awesome">Contents</PopoverContainer>
+      );
+
+      popoverSettingsIcon().click();
+      pressESCKey();
+      popoverContainerContent().should("not.exist");
+    });
+
     it("should not close Popover Container when an option is selected from Select component inside", () => {
       CypressMountWithProviders(<PopoverContainerWithSelect />);
       popoverSettingsIcon().click();
       selectText().click();
       selectListText("green").click();
       popoverContainerContent().should("be.visible");
+    });
+
+    it("should not close Popover Container when the escape key is pressed and the Select List is open", () => {
+      CypressMountWithProviders(<PopoverContainerWithSelect />);
+      popoverSettingsIcon().click();
+      selectText().click();
+      selectListText("red").trigger("keydown", keyCode("downarrow"));
+      pressESCKey();
+      selectListText("red").should("not.be.visible");
+      popoverContainerContent().should("be.visible");
+    });
+
+    it("should close Popover Container when the escape key is pressed with focus on the Select component", () => {
+      CypressMountWithProviders(<PopoverContainerWithSelect />);
+      popoverSettingsIcon().click();
+      selectText().click();
+      pressESCKey();
+      pressESCKey();
+      popoverContainerContent().should("not.exist");
     });
   });
 
@@ -338,6 +367,18 @@ context("Test for Popover Container component", () => {
       }
     );
 
+    it("should call onClose callback when the escape is pressed", () => {
+      const callback: PopoverContainerProps["onClose"] = cy
+        .stub()
+        .as("onClose");
+      CypressMountWithProviders(
+        <PopoverContainerComponent onClose={callback} open />
+      );
+
+      pressESCKey();
+      cy.get("@onClose").should("have.been.calledOnce");
+    });
+
     it("should call onClose callback when a click event is triggered outside the container", () => {
       const callback: PopoverContainerProps["onClose"] = cy
         .stub()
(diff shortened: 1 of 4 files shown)


cypress/components/popover-container/popover-container.cy.tsx
```diff
@@ -280,13 +280,33 @@ context("Test for Popover Container component", () => {
       }
     );
 
+    it("should close Popover Container using escape keyboard key", () => {
+      CypressMountWithProviders(
+        <PopoverContainer title="Cypress is awesome">Contents</PopoverContainer>
+      );
+
+      popoverSettingsIcon().click();
+      popoverCloseIcon().trigger("keydown", keyCode("Esc"));
```
robinzigmond, line 289:
```suggestion
      pressESCKey();
```

**suggestion**: I think this is the better test here, and not just because it's less code 😉 

You're specifically triggering `keydown` on the close icon, but (inlike pressing return with the close icon focused) which element the event is triggered on doesn't matter, and in the component you've (correctly imo) added the `keydown` handler to the whole `document`. So we should test that Escape works from anywhere on the page, not just when the close icon is focused.

cypress/components/popover-container/popover-container.cy.tsx
```diff
@@ -338,6 +358,18 @@ context("Test for Popover Container component", () => {
       }
     );
 
+    it("should call onClose callback when the escape is pressed and component is controlled", () => {
```
robinzigmond, line 361:
```suggestion
    it("should call onClose callback when the escape is pressed", () => {
```

**suggestion** (non-blocking): I don't think it makes the test any more descriptive if it specifically refers to being controlled, as `onClose` only exists when it's controlled anyway

cypress/components/popover-container/popover-container.cy.tsx
```diff
@@ -338,6 +358,18 @@ context("Test for Popover Container component", () => {
...
+      const callback: PopoverContainerProps["onClose"] = cy
+        .stub()
+        .as("onClose");
+      CypressMountWithProviders(
+        <PopoverContainerComponent onClose={callback} open />
+      );
+
+      popoverCloseIcon().trigger("keydown", keyCode("Esc"));
```
robinzigmond, line 369:
```suggestion
      pressESCKey();
```

as in an earlier comment, I think this is a  better test than triggering on the close icon

src/components/popover-container/popover-container.component.tsx
```diff
@@ -168,6 +169,33 @@ export const PopoverContainer = ({
       setTimeout(() => closeButtonRef.current?.focus(), 0);
   }, [isOpen]);
 
+  const handleEscKey = useCallback(
+    (ev) => {
+      const eventIsFromSelectInput = Events.composedPath(ev).find((p) => {
```
robinzigmond, line 174: **comment**:  not a fan of the name `p` for this variable, although it looks like you took it from `submenu` where we're doing something very similar. Short variable names are almost always bad (counter variables in loops being the main exception), and I can't even see what `p` stands for (not `path` I hope because `p` here isn't the path, it's an element that's part of the path. For that reason I'd much rather call it something like `element`, or `target` (since technically its type is `EventTarget`).

src/components/popover-container/popover-container.component.tsx
```diff
@@ -168,6 +169,33 @@ export const PopoverContainer = ({
       setTimeout(() => closeButtonRef.current?.focus(), 0);
   }, [isOpen]);
 
+  const handleEscKey = useCallback(
```
robinzigmond, line 172: **suggestion**: this is mostly/entirely doing the same thing as the `handleCloseButtonClick` function that we already have. Unless you can think of a good reason not to I'd want to extract it to a common function (called something simple like `closePopover` and just call that from both places).

The one difference that I can find in current behaviour is that in your version the Escape press doesn't refocus the close button, whereas clicking the close icon (including via the return key) does. I really think this refocus behaviour should happen with escape too, so that's another good reason for reusing the code imo.

src/components/popover-container/popover-container.spec.tsx
```diff
@@ -314,6 +315,22 @@ describe("PopoverContainer", () => {
...
+
+      act(() => {
+        wrapper.find(PopoverContainerOpenIcon).props().onClick();
+      });
+
+      wrapper.update();
+      expect(wrapper.find(PopoverContainerOpenIcon).props().tabIndex).toBe(-1);
+      expect(wrapper.find(PopoverContainerContentStyle).exists()).toBe(true);
+
+      act(() => {
```
robinzigmond, line 329: **comment**: is this test complete? We seem to be opening the container and checking that worked (above) - probably not an assertion we need in this test as it's tested elsewhere but it doesn't hurt. But now (in this `act`) we're pressing escape, but there are no assertions afterwards. Presumably we should check the container has actually closed?

src/components/popover-container/popover-container.spec.tsx
```diff
@@ -785,6 +802,73 @@ describe("open state when click event triggered", () => {
...
+          onClose={(e) => {
+            setOpen(false);
+            onCloseFn(e);
+          }}
+        />
+      );
+    };
+    const wrapper = mount(<MockWrapper />);
+
+    expect(wrapper.update().find(PopoverContainer).prop("open")).toBe(true);
```
robinzigmond, line 823: **comment**: do we need this assertion? You've designed the mock component to start off with `open={true}` so I'm not sure we need to assert on that - especially as the rest of the test doesn't check that it's changed to `false` (which I agree it doesn't need to), instead it checks that the `onCloseFn` has been called.

src/components/popover-container/popover-container.component.tsx
robinzigmond, line 174: **comment**: I'm not sure how to fix it but I just tested this with the "complex example" story you kindly added the Select to, and notice that with that the Escape key doesn't close the PopoverContainer whenever the select is focused, even if it's shut. I assume the intended behaviour is that when the select list is open, escape only closes that, but that a subsequent escape press should close the whole thing? As it is you need to tab/click off the select for that to work. Not critical perhaps but something to think about.

src/components/popover-container/popover-container.spec.tsx
DipperTheDan: Yeah, my bad. It should be checking the `PopoverContainer` is closed. I'll get that added in.

src/components/popover-container/popover-container.component.tsx
DipperTheDan: That seems reasonable. I can certainly do that 👍🏻

src/components/popover-container/popover-container.spec.tsx
DipperTheDan: I can have that removed. I only had this in as a few other tests in this spec file also have this assertion, so I went with the pattern that was already in the file.

src/components/popover-container/popover-container.component.tsx
DipperTheDan: I can certainly change the variable to something a bit descriptive and I most certainly stole the code from `submenu`. I'll get `submenu` updated too but in a separate PR.

cypress/components/popover-container/popover-container.cy.tsx
DipperTheDan: Thank you. I completely forgot that `pressESCKey()` even existed. I'll get that changed.

src/components/popover-container/popover-container.component.tsx
DipperTheDan: I think I've managed to address this using an `aria-expanded` DOM query.

src/components/popover-container/popover-container.spec.tsx
```diff
@@ -785,6 +808,119 @@ describe("open state when click event triggered", () => {
...
+    };
+    mount(<MockWrapper />);
+
+    document.dispatchEvent(
+      new KeyboardEvent("keydown", {
+        key: "Escape",
+        bubbles: true,
+      })
+    );
+    expect(onCloseFn).toHaveBeenCalled();
```
Parsium, line 835: **question**: even the test description, is this test missing an assertion for verifying the `PopoverContainer` has been closed?

```suggestion
    expect(wrapper.update().find(PopoverContainer).exists()).toBe(false);
    expect(onCloseFn).toHaveBeenCalled();
```

src/components/popover-container/popover-container.spec.tsx
```diff
@@ -785,6 +808,119 @@ describe("open state when click event triggered", () => {
...
+    );
+
+    selectInput?.dispatchEvent(
+      new KeyboardEvent("keydown", {
+        key: "Escape",
+        bubbles: true,
+      })
+    );
+
+    expect(wrapper.update().find(PopoverContainer).exists()).toBe(true);
```
Parsium, line 875: **question**: even the test description, is this assertion meant to be checking the `PopoverContainer` has been closed?

```suggestion
    expect(wrapper.update().find(PopoverContainer).exists()).toBe(false);
```

src/components/popover-container/popover-container.spec.tsx
DipperTheDan: Yeah I can put `expect(wrapper.update().find(PopoverContainerContent).exists()).toBe(false);` here as the `PopoverContainer` is in the DOM even when the container is closed.

src/components/popover-container/popover-container.spec.tsx
DipperTheDan: Same again here. I'll check that the content does not exist.

Parsium (approved):

robinzigmond (approved):

robinzigmond (approved):

divyajindel (approved):

PR #6391, opened 2023-10-26
feat(toast): toast vertical align prop

### LINKED ISSUE #6301: Allow Toast components to be show at the bottom of the screen.
### Desired behaviour

Toast components can be shown at the bottom of the screen as well as at the top

### Current behaviour

Currently, the Toast component will only render at the bottom of the screen if using the `notice` variant, at which point you lose the ability to show toasts at the bottom of the screen for successful/erroneous operations. It would be good to be able to position "standard variant" toasts at the bottom of the screen.

The [Sage Design System](https://zeroheight.com/2ccf2b601/p/31d5ea-toast/b/227fa0/t/06fffd) notes the following:
>  Typically, a toast covers part of the app bar and content, about 32 pixels from the top of the screen. But it can appear at the bottom or centre of the screen if required.

### Suggested Solution

_No response_

https://carbon.sage.com/?path=/docs/toast--default

Our old approach used to use `Portal` which is now an internal component. The Carbon storybook suggests that this is still doable via the `targetPortalId` prop but does not give examples of how this can be set up - perhaps the documentation needs to be updated with said changes in mind?

### Confidentiality

nuria1110:
fix #6301

### Proposed behaviour

- Added `alignY` prop which allows to vertically align the component at the top, center and bottom of the screen. 
- The `notice` variant will still render at the bottom of the screen by default but you can set `alignY` set to "top" to render it at the top of the screen. 
- Fixed visual bug on when `align` was set to left. 

### Current behaviour

- Toast currently does not support vertical alignment. 
- There is a visual bug when `align` is set to "left" and the screen size is small:

Screenshot 2023-10-26 at 17 03 27

### Checklist

#### QA

### Testing instructions

Added a test story for cases where there is more than one toast on the screen at the same time and have different vertical alignments. 


src/components/toast/toast.style.ts
```diff
@@ -27,8 +28,9 @@ const StyledPortal = styled(Portal)<{
 
     ${align === "left" &&
     css`
-      left: 12%;
-      transform: translateX(-50%);
+      display: flex;
```
edleeks87, line 31: Question: Do we need `flex` here?

src/components/toast/toast.style.ts
```diff
@@ -49,6 +51,26 @@ const StyledPortal = styled(Portal)<{
...
+    css`
+      top: 50%;
+      transform: translate(${align === "left" ? "50%" : "-50%"}, -50%);
+    `}
+
+    ${alignY === "bottom" &&
+    css`
+      bottom: 0;
+      top: auto;
+      display: flex;
```
edleeks87, line 72: Question: Do we need flex here?

src/components/toast/toast.style.ts
```diff
@@ -129,24 +158,24 @@ const ToastStyle = styled(MessageStyle)<{
       }
 
       &${alternativeAnimationName}-appear, &${alternativeAnimationName}-enter {
-        bottom: -40px;
+        ${alignY === "top" ? "top: -40px;" : "bottom: -40px;"};
```
edleeks87, line 161:
```suggestion
        ${alignY === "top" ? "top: -40px" : "bottom: -40px"};
```
Comment (non-blocking): this id adding an additional `;` here, I believe v4 of styled-components handles these but v5 doesn't so we'd have issues if/ when we upgraded

src/components/toast/toast.stories.tsx
```diff
@@ -354,6 +354,111 @@ export const AlignedRight = () => {
   );
 };
 
+export const AlignedYTop = () => {
```
edleeks87, line 357: Suggestion: I think it would be worth having some demos with the different permutations of `align` and `alignY` set

src/components/toast/toast.spec.tsx
```diff
@@ -593,6 +593,53 @@ describe("Align", () => {
   );
 });
 
+describe("Align vertical", () => {
```
edleeks87, line 596: Comment: we could probably add some tests that capture the different permutations of the alignment props, but maybe that's better suited to cypress or chromatic as that's rendered in browser

edleeks87 (commented):
Some minor comments from me but overall great work @nuria1110 👏 . 

I think we could also slightly tweak the commit message here as it will form the release note when this is merged. Something like below might work but happy for you to go with what you want here 😄 

```
feat(toast): add alignY prop to allow vertical alignment to be configurable 

adds `alignY` prop to `Toast` which allows to consumers to vertically align the component on the screen

fix #6301
```

src/components/toast/toast.style.ts
nuria1110: Good point, no I don't think there is a need for it here. I took it from the styles that were set for `align === "right"` and just flipped it for left, but I think it should be fine to remove `flex` from both left and right as it doesn't seem to affect the styling.

src/components/toast/toast.style.ts
nuria1110: I do think `flex` needs to be added here for `flex-direction` to take effect - this is mainly for stacked toasts to appear in bottom to top order when `alignY` is set to "bottom". I could remove this if its preferred that stacked toasts have the same order from top to bottom in all cases.

cypress/components/toast/toast.cy.tsx
```diff
@@ -204,6 +207,19 @@ context("Testing Toast component", () => {
...
+    it.each(["top", "bottom"] as ToastProps["alignY"][])(
+      "should render Toast component alignY prop set to %s",
+      (alignY) => {
+        CypressMountWithProviders(<ToastComponent alignY={alignY} />);
+
+        toastComponent()
+          .parent()
+          .parent()
+          .parent()
+          .parent()
```
Parsium, line 219: **thought**: it might be worth when we migrate these tests to Playwright to amend how we are locating the portal element here.

It's not essential now, but if we can directly locate the portal element instead of traversing the DOM, this should make the test less fragile if we ever change the component's markup.

Parsium (dismissed):

edleeks87 (dismissed):

harpalsingh:
Approved PR but wasnt sure we wanted centre aligned, however we are considering some updates around that placement, so good to allow that to be included and we can test.

cypress/components/toast/toast.cy.tsx
```diff
@@ -204,6 +207,19 @@ context("Testing Toast component", () => {
           .should("have.css", "justify-content", align);
       }
     );
+    it.each(["top", "bottom"] as ToastProps["alignY"][])(
```
ZhuoyuJin, line 210: Should we have the case for prop "alignY" as centre here too? it.each(["top","centre", "bottom"]

ZhuoyuJin (dismissed):

Parsium (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6395, opened 2023-10-27
feat(select): add selectionConfirmed property to event emitted on change

### LINKED ISSUE #6330: Add the possibility to determinate if an onChange event in FilterableSelect comes from a confirmed choice
### Desired behaviour

The current implementation of `FilterableSelect` component emits an `onChange` event in every change detected. So, if a user scrolls through the list of options using the keyboard or if he selects any option pressing the `Enter` key or clicking the option with the mouse the event is fired.

It should be useful having the possibility to know if the event comes from a confirmed selection or not, so we can know if the user has done the final choice. In that example it would be when the `Enter` key is pressed of when the option is clicked via mouse.

### Current behaviour

The current implementation of `FilterableSelect` component emits an `onChange` event in every change of option

### Suggested Solution

_No response_

https://codesandbox.io/s/pensive-chihiro-n56txf?file=/src/App.js

_No response_

### Confidentiality

edleeks87:
fix #6330

### Proposed behaviour

Adds a `CustomSelectChangeEvent` type which defines the shape of the new change event emitted when the input value changes, backwards compatibility is maintained so it is still possible to continue to use React's ChangeEvent. Within the new custom event there is a `selectionConfirmed` property that indicates when a user has pressed `Enter` or clicked on a given `Option`.

Filter out `fill` prop so it isn't set on the underlying DOM element when passed to an `Option`

### Current behaviour

No `CustomSelectChangeEvent` type exists, React's ChangeEvent is what is emitted when input value is updated. Therefore there is currently no way for implementation teams to know when a user has made a selection via keyboard unless the input is blurred when using `Filterable` and `Simple` select components.

When `fill` prop is passed on `Option` it is set on the underlying DOM element.

### Checklist

#### QA

### Additional context

**There wasn't really much need to add the new `selectionConfirmed` event property to `MultiSelect` as it doesn't update the input value as the user keyboard navigates. However, I decided add it anyway for the sake of consistency but I haven't added any public documentation for it.**

### Testing instructions

See new `selection-confirmed-story` added to storybook for Filterable and Simple Select


DipperTheDan (approved):

src/components/select/simple-select/simple-select-test.stories.tsx
```diff
@@ -702,3 +706,43 @@ SelectWithOptionGroupHeader.args = {
   listPlacement: undefined,
   flipEnabled: true,
 };
+
+export const SelectionConfirmed = () => {
+  const [value, setValue] = React.useState("");
```
nineteen88, line 711: suggestion(minor, non-blocking): We can remove the `React.` here too

nineteen88 (approved):
Approved - up to you if you want to tidy the couple of redundant `React.`s

nineteen88 (approved):

DipperTheDan (approved):

nineteen88 (approved):

DipperTheDan (approved):

stephenogorman (approved):

PR #6414, opened 2023-11-07
fix(menu-full-screen): fix keys in rendering of children

robinzigmond:
Currently the component gives a warning in the console for the list of children not having keys. And due to the lack of keys, complex situations involving a child being added or removed while another changes state are not handled correctly. This has been fixed by using React's build in toArray utility which automatically handles the keys.

### Proposed behaviour

No warnings should be seen in the console when rendering the component.
No warnings should be seen in the terminal when running the unit tests.
Adding or removing child items should maintain any state in the other children.

### Current behaviour

When rendering the component, a warning about missing keys is visible in the console (this can be seen on Storybook - go to the "full screen" example and shrink the screen enough that a MenuFullScreen is rendered).
The same warning is seen in the terminal when running the unit tests for MenuFullScreen.
Adding a child item at the start of the menu resets state in the other children (see codesandbox linked below).

### Checklist

#### QA

### Testing instructions

Ensure that no console warnings about keys appear when viewing MenuFullScreen in storybook, or when running unit tests.

Retest the fix in https://github.com/Sage/carbon/pull/6155 to ensure that nothing has regressed from that (the changes in that PR are what inadvertently caused this problem with keys - note that a Cypress test added as part of that PR was unintentionally removed when reverting the Menu Cypress tests to Typescript, this has been reinstated as part of this PR).

https://codesandbox.io/s/dreamy-hoover-zv2z35

[explanation of Sandbox - this is a contrived example where one menu item maintains an internal state that regularly counts from 0 up to 2, while after 5 seconds a new menu item appears at the start of the menu. Observe that when the new item appears, the last item has its count reset to 0. This would not be expected to happen, and should be fixed by the build from this PR.]


src/components/menu/menu-full-screen/menu-full-screen.component.tsx
```diff
@@ -77,9 +77,15 @@ export const MenuFullscreen = ({
...
-  );
+  const flattenedChildren = React.Children.toArray(children);
+  const childrenWithMenuDividers: React.ReactNode[] = [];
+  flattenedChildren.forEach((child, index) => {
+    childrenWithMenuDividers.push(child);
+    if (index < flattenedChildren.length - 1) {
+      childrenWithMenuDividers.push(<MenuDivider />);
+    }
+  });
+  const childArray = React.Children.toArray(childrenWithMenuDividers);
```
edleeks87, line 88: suggestion (non-blocking): happy for you to leave this as you have it already but we could probably reduce the amount of arrays here by using `map`

```
const childArray = React.Children.toArray(flattenedChildren.map((child, index) => {
    if (index < flattenedChildren.length - 1) {
      return (
        <>
          {child}
          <MenuDivider />
        </>
      )
    }

    return child;
  }));

DipperTheDan (commented):
Nothing more to add from me 😄👍🏻

src/components/menu/menu-full-screen/menu-full-screen.component.tsx
robinzigmond: Thanks - this seems to work just as well as far as assigning keys correctly is concerned. Happy to make this change.

DipperTheDan (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6420, opened 2023-11-09
feat: responsive tile

### LINKED ISSUE #6419: Responsive tile
### Desired behaviour

We want the Tile to be responsive. 
When there is not enough space, the elements inside the Tile should wrap on a new line and a divider should appear between the lines.

### Current behaviour

Currently, the Tile is not responsive and the elements will shrink until they reach theirs min-width and they overflow the Tile.

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

gusch:
closes #6419

### Proposed behaviour

We want the Tile to be responsive.
When there is not enough space, the elements inside the Tile should wrap on a new line and a divider should appear between the lines.

image
image

### Current behaviour

The content of the Tile is not responsive.
The elements will shrink until they reach theirs min-width and may overflow the Tile.

image
image

### Checklist

#### QA

### Testing instructions


src/components/tile/responsive-cell/responsive-cell.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export interface ResponsiveCellProps
+  extends TagProps,
+    PaddingProps,
+    FlexGrowProps,
+    FlexBasisProps,
+    JustifyContentProps,
+    MaxWidthProps {
+  /** The content to render within the responsive cell. */
+  children?: React.ReactNode;
```
edleeks87, line 24:
```suggestion
  children: React.ReactNode;
```

I think we can just make `children` required here

src/components/tile/components.test-pw.tsx
```diff
@@ -77,3 +85,39 @@ export const TileComponentWithFalsyChildren = () => {
...
+        <ResponsiveCell py={2}>Test Body Two</ResponsiveCell>
+        <ResponsiveCell py={2}>
+          Test Body Three With a very very long text
+        </ResponsiveCell>
+      </ResponsiveTileContainer>
+    </Tile>
+  );
+};
+
+export const ResponsiveTileContainerWithFalsyChildren = () => {
```
edleeks87, line 103: Comment: I think this can be a test story or we can just rely on unit/ Ci tests to capture this

src/components/tile/components.test-pw.tsx
```diff
@@ -77,3 +85,39 @@ export const TileComponentWithFalsyChildren = () => {
...
+    <>
+      <ResponsiveTileContainer />
+      <ResponsiveTileContainer>Foo</ResponsiveTileContainer>
+      <ResponsiveTileContainer>{null}</ResponsiveTileContainer>
+      {false && <ResponsiveTileContainer>Bar</ResponsiveTileContainer>}
+    </>
+  );
+};
+
+export const ResponsiveCellWithFalsyChildren = () => {
```
edleeks87, line 114: Comment: I think this can be a test story or we can just rely on unit/ Ci tests to capture this

src/components/tile/responsive-cell/responsive-cell.style.ts
edleeks87, line 1: comment: I don't think we need this, see comment about using a compositional approach above

src/components/tile/responsive-tile-container/responsive-tile-container.component.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import Box from "../../box";
+import boxGap, { Gap } from "../../../style/utils/box-gap";
+
+export interface ResponsiveTileContainerProps {
+  /** The child elements of ResponsiveTileContainer need to be ResponsiveCell components. */
+  children?: React.ReactNode;
```
edleeks87, line 7:
```suggestion
  children: React.ReactNode;
```

src/components/tile/responsive-tile-container/responsive-tile-container.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  it("does not render when no children are passed", () => {
+    render(<ResponsiveTileContainer data-testid="responsive-tile-container" />);
+
+    expect(
+      screen.queryByTestId("responsive-tile-container")
+    ).not.toBeInTheDocument();
+  });
+
+  it("does not render when falsy children are passed", () => {
```
edleeks87, line 30: comment: if we make `children` required we can remove this test

edleeks87 (commented):
I'm not entirely sold on the naming for these new components as it kind of feels like we're implying that the existing `Tile` and `TileContainer` aren't responsive wen they are. I think calling them something like `FlexTileContainer` and `FlexTileCell` or some variation of that would be a bit better as it's a little more intuitive what their purpose is. This isn't a blocker for me though so if other reviewers disagree I'm happy to be overruled

src/components/tile/responsive-cell/responsive-cell.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <Box
+      display="flex"
+      position="relative"
+      flexGrow={rest.flexGrow || 1}
+      flexBasis={rest.flexBasis || "160px"}
+      flexShrink={0}
+      {...rest}
+      {...tagComponent("responsive-cell", rest)}
+    >
+      <StyledDivider>
```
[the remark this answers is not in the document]
gusch: "We haven't had any guidance from DS here" : The props and their values to display the divider correctly was the trickiest part to do during the POC. The divider was clearly part of the wanted result.

Would you agree with a divider prop (like for heading) on each cell to show/hide it ?

src/components/tile/components.test-pw.tsx
gusch: Done in UT

src/components/tile/components.test-pw.tsx
gusch: Done in UT

src/components/tile/responsive-cell/responsive-cell.style.ts
gusch: removed

src/components/tile/responsive-tile-container/responsive-tile-container.spec.tsx
gusch: Children is a ReactNode and can be null for example.
I have this test to cover 
```
if (!children) {
    return <></>;
  }
```

.storybook/utils/styled-system-props.js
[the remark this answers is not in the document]
nineteen88: I think we should not import these props from styled-system and assign them in the interface ourselves. This way we keep styled-system (which is on the roadmap to be removed) for the prop arrays. The alternative would be to allow consumers more flexibility and support all of the flexbox props - but this may need further thought...

src/components/tile/flex-tile-container/flex-tile-container.component.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import Box, { BoxProps } from "../../box";
+
+export interface FlexTileContainerProps extends Pick<BoxProps, "columnGap"> {
+  /** The child elements of ResponsiveTileContainer need to be ResponsiveCell components. */
```
nineteen88, line 5: comment: this comment no longer reflects the component names

src/components/tile/responsive-cell/responsive-cell.component.tsx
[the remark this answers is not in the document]
edleeks87: I don't think we should bake anything in when we're still waiting for DS to provide responsive guidance. There's a chance that guidance could contradict what we implement here and we'd need to make a breaking change to remove it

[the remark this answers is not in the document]
edleeks87: Maybe a compromise here is to move the divider composition into a component within the `Tile` folder

```
const FlexTileDivider = () => (
   <Box
            position="absolute"
            bottom="0px"
            left="0px"
            width="100vw"
            m="0px 0px -1px 0px"
        >
            <Hr m={0} />
          </Box>
)
```

Which can then be used like:

```
<Tile m={0} py={0}>
    <Container>
      <Cell py={2}>
        Test Body One
        <FlexTileDivider />
      </Cell>
  </Container>
</Tile>

File: src/components/tile/tile.stories.tsx [hunk 8]
```diff
@@ -582,3 +589,191 @@ export const WithAccordionAndTileFooter: ComponentStory<typeof Tile> = () => (
...
+        <FlexTileDivider />
+        <Box backgroundColor="#dedede" width="100%">
+          Flex 160px extra-wide
+        </Box>
+      </FlexTileCell>
+    </FlexTileContainer>
+  </Tile>
+);
+
+export const Align: ComponentStory<typeof Tile> = () => (
```
edleeks87, line 764: Question: Should this not have a `Tile` parent?

src/components/tile/tile.stories.tsx
```diff
@@ -582,3 +589,191 @@ export const WithAccordionAndTileFooter: ComponentStory<typeof Tile> = () => (
 WithAccordionAndTileFooter.parameters = {
   chromatic: { disableSnapshot: true },
 };
+
+export const ResponsiveDefaultStory: ComponentStory<typeof Tile> = () => (
```
edleeks87, line 593: Suggestion: These are great for illustrating these enhancements to Carbon consumers. I wonder if it might be worth adding similar ones to the `test` stories with some viewport adjustments so chromatic can capture when the containers wraps and dividers etc are shown. We don't need to worry about them being public facing here so we can just focus on making sure the we have any future visual regressions captured. If you're happy to do that I would disable chromatic for these ones and let it take baselines from the test ones

```
StoryName.parameters = {
  chromatic: {
    viewports: [1500, 900],
  },
};

edleeks87: This is non-blocking for me, if it's going to cause issues with you're deadline we can make these changes in a new PR

edleeks87 (approved):

nineteen88 (dismissed):

stephenogorman (approved):

edleeks87 (approved):

PR #6422, opened 2023-11-10
feat(badge): add color prop

### LINKED ISSUE #5839: Add a Color property for Badge component
### Desired behaviour

I would like the Badge component to have the same color as its children by adding a Color property.

### Current behaviour

The Badge component always have 'primary' color.

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

nuria1110:
fix #5839

### Proposed behaviour

Added an optional `customColor` prop to Badge which allows consumers to change the colour of the component instead of the default one. 

### Current behaviour

Badge currently does not support changing the colour of the component.  

### Checklist

#### QA

### Testing instructions


src/components/badge/badge.component.tsx
```diff
@@ -34,11 +37,16 @@ export const Badge = ({
         }
       : {
           "aria-label": ariaLabel,
+          customColor,
```
edleeks87, line 40: comment: as we're explicitly passing this to `StyledBadge` on line 47 we don't need to conditionally add it to the props here

src/components/badge/badge.style.ts
```diff
@@ -36,14 +40,17 @@ const StyledBadge = styled.span.attrs(({ onClick }) => ({
   width: 22px;
   min-height: 22px;
   border: solid 2px transparent;
-  border-color: var(--colorsActionMajor500);
-  color: var(--colorsActionMajor500);
+
+  ${({ customColor }) => css`
+    border-color: ${customColor || "var(--colorsActionMajor500)"};
+    color: ${customColor || "var(--colorsActionMajor500)"};
```
edleeks87, line 46: Suggestion: it might be worth using the [toColor util](https://github.com/Sage/carbon/blob/750c09398da34ec8f5de5e8f80794975f2b583cc/src/style/utils/color.ts#L26) here and setting the `border` and `color` defaults accordingly. I twould match how we've implemented similar features in other components such as `verticalBorderColor` in the [FlatTableRowHeader](https://github.com/Sage/carbon/blob/847c7c3dd7cd7fb11ebbe26f575c85dfffec5c5e/src/components/flat-table/flat-table-row-header/flat-table-row-header.style.ts). Doing this would mean that implementing teams can simply pass a value like the example below.


```
 ${({ customColor }) => css`
    border-color: ${toColor(theme, customColor || "--colorsActionMajor500")};
    color: ${toColor(theme, customColor || "--colorsActionMajor500")};
`}

...

 <Badge
      counter={9}
      onClick={() => {}}
      customColor="--colorsSemanticNegative500"
    >
      <Button mr={0} buttonType="tertiary" destructive>
        Filter
      </Button>
    </Badge>

File: src/components/badge/badge.style.ts [hunk 3]
```diff
@@ -36,14 +42,19 @@ const StyledBadge = styled.span.attrs(({ onClick }) => ({
   width: 22px;
   min-height: 22px;
   border: solid 2px transparent;
-  border-color: var(--colorsActionMajor500);
-  color: var(--colorsActionMajor500);
+
+  ${({ customColor, theme }) => css`
```
robinzigmond, line 46: **suggestion**: I find all the conditional checks on `customColor` in this component a bit ugly to read. I think we can avoid all that by making `customColor` a mandatory prop (for `StyledBadge` I mean, not for `Badge` itself! note that this will mean you can't use my earlier suggestion of `Pick<BadgeProps, "customColor">` if you do this), and then when rendering the styled component pass it as `customColor={customColor || "--colorsActionMajor500"}`.

This way we avoid a lot of the complexity in this component that's not really needed, and instead will just be saying "we'll render whatever color is passed in, but if one isn't it'll be `colorsActionMajor500`". Which I think is what we want?

src/components/badge/badge.spec.tsx
```diff
@@ -110,5 +110,33 @@ describe("Badge", () => {
         wrapper.find(StyledBadge)
       );
     });
+
+    it("should render badge with default style when no customColor is specified", () => {
```
robinzigmond, line 114:
```suggestion
    it("should render badge with default style when no color is specified", () => {
```

**nitpick** (non-blocking): I don't think we should refer to `customColor` any more now the prop has been renamed

src/components/badge/badge.spec.tsx
```diff
@@ -110,5 +110,33 @@ describe("Badge", () => {
...
+      assertStyleMatch(
+        {
+          borderColor: "var(--colorsActionMajor500)",
+          color: "var(--colorsActionMajor500)",
+        },
+        wrapper.find(StyledBadge)
+      );
+    });
+
+    it("should render badge with correct style when customColor is specified", () => {
```
robinzigmond, line 125:
```suggestion
    it("should render badge with correct style when color is specified", () => {
```

**nitpick** (non-blocking): as above

robinzigmond (approved):

edleeks87 (approved):

edleeks87 (approved):

robinzigmond (approved):

divyajindel (approved):

PR #6427, opened 2023-11-14
Improve markup for menu titles FE-5106

### LINKED ISSUE #6010: Menu segment title doesn't render as a header
### Current behaviour

The segment titles within menus are not rendered as headings. This is an accessibility issue and fails against 4.1.2 of [Web Content Accessibility Guidelines (WCAG) 2.1 (w3.org)](https://www.w3.org/TR/WCAG21/).

### Expected behaviour

The segment title should be rendered as the appropriate heading. 

https://codesandbox.io/s/menu-headers-0655gh?file=/src/App.js:605-694

### JIRA Ticket (Sage Only)

SBS-47524

### Suggested Solution

There should be an option to pass a `headingType` prop to `MenuSegmentTitle `which sets the HTML heading element which is rendered on the page. 

### Carbon Version

117.4.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #6010

### Proposed behaviour

Refactors `MenuSegmentTitle` component to recieve associated list items as children. This ensures
that the heading and list items are held within a sublist so that the heading is treated correctly
by voiceover software. Updates the title element to support the `as` prop to allow it to be set to
any valid heading element but defaults to `h2` if no value is passed. Removes `display:
inline-block` from list items and makes main menu list a flexed layout.

BREAKING CHANGE: You will need to pass any associated list items as children of the
`MenuSegmentTitle` so they are rendered in a sublist with the segment heading

Ensures `Icon`s rendered within `MenuItem` have `display: inline-block`

image

### Current behaviour

MenuSegmentTitles are not rendered as part of a sublist with associated menu items meaning they are treated as independent by VO software.
MenuSegmentTitle is rendered as a div rather than a heading element

Icons in menu items are display inline
image

### Checklist

#### QA

### Testing instructions


DipperTheDan (commented):
Everything looks good on the `Menu` side of things. `ShowEditPod` is still referenced and linked to in the `Pod` stories though. Attached the screenshot and highlighted them. 

Screenshot 2023-11-20 at 11 07 27

src/components/menu/menu-segment-title/menu-segment-title.spec.tsx
```diff
@@ -39,6 +41,17 @@ describe("Title", () => {
...
 
+  it("should title as an h2 element by default and render list items as children of the ul", () => {
+    wrapper = render("light");
+    openSubmenu(wrapper);
+    const title = wrapper.find(StyledTitle);
+    const segmentList = wrapper.find(StyledSegmentChildren);
+
+    expect(title.getDOMNode().tagName).toBe("H2");
+    expect(segmentList.getDOMNode().tagName).toBe("UL");
+    expect(segmentList.find("li").text()).toBe("bar");
```
Parsium, line 52: **thought**: would it be worth us as a team adopting an approach of querying DOM elements first over React DOM nodes where possible?

This would make migrating to RTL easier given it forces you to write tests in a way that tests the final markup a user would interact with.

```suggestion
    expect(title.getDOMNode().tagName).toBe("H2");
    expect(segmentList.getDOMNode().tagName).toBe("UL");
    expect(segmentList.find("li").text()).toBe("bar");
```

src/components/menu/menu-segment-title/menu-segment-title.component.tsx
```diff
@@ -8,25 +8,32 @@ import tagComponent, {
...
+  (
+    { children, variant = "default", text, as = "h2", ...rest }: MenuTitleProps,
+    ref
+  ) => {
     const menuContext = useContext(MenuContext);
 
     return (
       <StyledMenuItem inSubmenu>
         <StyledTitle
+          as={as}
```
Parsium, line 28: **question**: would it be worth filtering the `as` prop so only the supported heading level elements are accepted?

As its possible, despite this resulting in a TypeScript type error, that a JavaScript consumer could try to render the title as any element they like:

```tsx
<MenuSegmentTitle text="I am a title" as="div" /> 
``` 

since `as` is passed directly to the styled component `StyledTitle`

edleeks87: I don't mind doing it but I'm of the opinion that if someone wants to explicitly ignore the types is kind of on them when it breaks etc

Parsium: I agree, though I wonder if some of our JavaScript-only consumers could accidentally do this without repercussions and flag an issue with `as` set to a non-heading element

src/components/menu/menu-segment-title/menu-segment-title.spec.tsx
edleeks87: yeah my thinking is this can easily be converted when we move to RTL

src/components/menu/menu-segment-title/menu-segment-title.component.tsx
Parsium: Ah that's a good point about prop-types 👍🏼 If we come across the same pattern elsewhere in the future, let's leave it in that case

Parsium (dismissed):

DipperTheDan (approved):

Parsium (approved):

DipperTheDan (approved):

divyajindel (approved):

Parsium (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #6428, opened 2023-11-15
fix(select): ensure re-opening the select resets scroll position the top

### LINKED ISSUE #6399: Scroll position reset to second item in list
### Current behaviour

Selecting drop down > scrolling to bottom of list > clicking off the drop down so it closes > reselect the drop down > the scroll has reset back to the second item in the list.

### Expected behaviour

Reset to the first item in the list. 

https://carbon.sage.com/?path=/docs/select-filterable--default

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

120.1.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
Updates the react-virtual dependency, which was directly responsible for the bug. The new version makes the virtualiser return 0 items whenever the select is closed - this change in behaviour resulted in some problems after the update, particularly with tests, which have been addressed. An accessibility error with long virtualised lists, where selecting one option and scrolling away from it so that the selected option is no longer in the DOM, resulting in an invalid aria-activedescendant attribute, has also been addressed.

fix #6399

### Proposed behaviour

Whenever opening a select, the top item in the list should be visible.

No critical accessibility errors should be flagged by Axe on any Select components.

### Current behaviour

When opening a select, scrolling to the bottom, closing and then reopening, the top item cannot be seen (but can be brought into view by scrolling up).

In some circumstances (see below codesandbox and instructions to reproduce) it is possible for the select input to have an `aria-activedescendant` attribute that doesn't correspond to anything in the DOM, causing Axe to flag a critical error.

### Checklist

#### QA

### Testing instructions

Test that both the linked Github issue, and the accessibility issue in the below sandbox, are fixed.

Perform regression testing on all 3 types of Select, as well as the Pager component which uses a Select internally.

The following CodeSandbox shows the accessibility error with virtualised lists https://s5nzqg.csb.app/

To reproduce:
- open the select and select "option 1"
- close the select, reopen it and scroll all the way to the bottom
- with the list still open, run Axe

You should see a critical issue "ARIA attributes must conform to valid values" (see screenshot - note that the other error is a known false positive) 
image


src/components/select/select-list/select-list.component.tsx
```diff
@@ -236,35 +265,39 @@ const SelectList = React.forwardRef(
...
     };
 
-    const renderedChildren = items.map((item) => {
-      const { index, start } = item;
-      const child = childrenList[index];
+    // the rangeExtractor above can cause an undefined value to be appended to the return items.
+    // Easiest way to stop that crashing is just to filter it out.
+    const renderedChildren = items
+      .filter((item) => item !== undefined)
+      .map((item) => {
```
edleeks87, line 272: suggestion (non-blocking): as we're not using `item` anywhere else we could destructure `index` and `start` in the map callback parameter

```suggestion
      .map(({index, start}) => {
```

src/components/select/select-list/select-list.spec.tsx
```diff
@@ -1067,6 +1068,13 @@ describe("SelectList", () => {
       expect(optionChildren.length).toBeGreaterThanOrEqual(5);
       expect(optionChildren.length).toBeLessThan(10);
     });
+
+    it("when an option is selected, it is always in the DOM even when out of view", () => {
```
edleeks87, line 1072: question (minor): is it worth having a check on visibility here as well?

src/components/select/filterable-select/filterable-select.spec.tsx
```diff
@@ -40,6 +43,45 @@ function renderSelect(props = {}, renderer = mount) {
   return renderer(getSelect(props));
 }
 
+function simulateSelectTextboxEvent(
```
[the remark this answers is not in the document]
robinzigmond: Fair point - I of course noticed the repetition but didn't really think seriously about moving them. Not sure if `spec_helper` is appropriate as they're very select-specific, but that's probably better than creating a new test helper file specifically within select. Happy to do as you suggested but will wait for comments from a second reviewer 👍

src/components/select/select-list/select-list.spec.tsx
robinzigmond: is that possible in a Jest/Enzyme test? If this were a cypress/Playwright test I'd agree that asserting that it isn't visible would be useful - but it isn't (and I'm testing this differently in Cypress, by just running the accessibility checks).

edleeks87: No problem it's "nice" rather then "need" to have imo

edleeks87 (approved):

tomdavies73 (approved):

edleeks87 (approved):

ZhuoyuJin (dismissed):

tomdavies73 (approved):

edleeks87 (approved):

tomdavies73 (approved):

stephenogorman (approved):

PR #6435, opened 2023-11-17
feat(dialog): add greyBackground prop

### LINKED ISSUE #6257: Dialog background colour prop
### Desired behaviour

Currently, Dialog only supports a grey background variant, so this appears to be a misalignment with Product Design System which suggests we should be able to use a White one. 

This is a blocker to addressing an accessibility issue where the colour of the Stepper on a Dialog does not have the required colour contrast requirement. 

### Current behaviour

Only grey background on Dialog is available. 

### Suggested Solution

A new prop to select grey or white background. Perhaps backgroundGrey default true if set to false then white?

https://carbon.sage.com/?path=/docs/dialog--default-story

_No response_

### Confidentiality

nuria1110:
fix #6257

### Proposed behaviour

- `Dialog` default background color is set to white. 
- `hasGreyBackground` prop to allow consumers to set the background to grey. 

Screenshot 2023-11-17 at 15 26 45

### Current behaviour

Current `Dialog` default background color is set to grey.

Screenshot 2023-11-17 at 15 22 27

### Checklist

#### QA

### Testing instructions


edleeks87:
I just want to check with @harpalsingh about whether a feature flag passed via the `CarbonProvider` is a better approach here to avoid a future breaking change

src/components/dialog/dialog.component.tsx
```diff
@@ -91,6 +91,8 @@ export interface DialogProps extends ModalProps, TagProps {
   role?: string;
   /** Padding to be set on the Dialog content */
   contentPadding?: ContentPaddingInterface;
+  /** Allows developers to change the background color of the content to grey */
+  hasGreyBackground?: boolean;
```
Parsium, line 95: **suggestion(non-blocking)**: amending prop name to exclude token value

Given we have no control over the token values from the `@sage/design-tokens` package, there is potential that the `colorsUtilityMajor025` token could be updated in the future to be no longer grey. So I would amend the prop name slightly to exclude token value, to avoid the need to update the prop name later as a breaking change.

```suggestion
  /** Change the background color of the content to old grey color */
  legacyBackgroundColor?: boolean;
```

edleeks87:
https://github.com/Sage/carbon/pull/6435#issuecomment-1818849949

Spoken to Harpal and confirmed we want a prop for this

src/components/dialog/dialog.component.tsx
Parsium, line 95: **nitpick(non-blocking)**: a minor suggestion to make the prop name and description more concise

```suggestion
  /** Change the background color of the content to grey */
  greyBackground?: boolean;
```

Parsium (commented):
Approach looks good to me @nuria1110, nice work 👍🏼

src/components/dialog/dialog.pw.tsx
```diff
@@ -496,6 +497,19 @@ test.describe(
       await checkAccessibility(page);
     });
 
+    // TODO: (FE-6245) - this test should be failing color contrast accessibility test with grey background
+    test.skip("DialogWithStepSequence mock component should pass accessibility checks", async ({
```
robinzigmond, line 501: **comment** (non-blocking) - I don't think the test description is very accurate here. I would have thought the important point is that the `hasGreyBackground` prop passes accessibility tests, with the "step sequence" part not being very relevant here? Yet the title only means the step sequence and not the `hasGreyBackground`.

Note that with the TODO comment saying this should actually fail, I'm not sure what the actual intention is, since obviously we don't want failing accessibility tests!

src/components/dialog/components.test-pw.tsx
```diff
@@ -99,3 +100,64 @@ export const DialogWithOpenToastsBackgroundScrollTest = () => {
     </Box>
   );
 };
+
+export const DialogWithStepSequence = (props: Partial<DialogProps>) => {
```
robinzigmond, line 104: **comment** (non-blocking): I'm not sure why we've introduced a more complex story, and highlighted the `StepSequence` part in its name, when the purpose seems to be to use it to test the new `greyBackground` prop. If that's the intention I'd make a very simple dialog example, with the `greyBackground` prop passed explicitly, and call it something like `DialogWithGreyBackground`.

src/components/dialog/dialog.component.tsx
```diff
@@ -305,6 +308,11 @@ export const Dialog = forwardRef<DialogHandle, DialogProps>(
             role={role}
             tabIndex={-1}
             {...contentPadding}
+            backgroundColor={
+              hasGreyBackground
+                ? "var(--colorsUtilityMajor025)"
```
robinzigmond, line 313: **question**: is this definitely the right token? I ask because on the Github issue Luke mentions `colorsUtilityMajor150` as  being correct for the grey background, and you've got 025 rather than 150. No worries if the designs have changed since then (Luke's comment was in August), but I thought it was worth raising this.

src/components/dialog/dialog.pw.tsx
nuria1110: I may have not explained myself very well in the comment sorry - this test was added to address the issue raised regarding the color contrast issues with the green text in `StepSequence` and the previous default grey background. When making sure that the test failed when `hasGreyBackground` prop was passed, the test still passed. I added the ticket number to look into possible issues with the `checkAccessibility()` util, this is also why the test is being skipped at the moment.

src/components/dialog/dialog.component.tsx
nuria1110: This was the previous design token being used as the background colour and I also checked in the DS designs and this seems to be the token being used for the grey background.

src/components/dialog/dialog.pw.tsx
robinzigmond: thanks @nuria1110 - I now see this is my own fault for misunderstanding the original ticket 😅 I'd thought we already had white as the unchangeable default and you were adding a grey option, but now I look more closely I see that we're changing the default to white, and adding a prop to get back to the grey colour. Which of course will change the background color of most dialogs used by consumers, but that's fine if it's what the DS team are asking for 👍 

Regarding the test, I'd make it so that it's testing without the grey background, and ensuring the accessibility is OK. That should presumably pass so imo it's better to not skip it. We should still keep your comment though so we know to investigate it not failing even when the grey background is used and the contrast should fail.

src/components/dialog/dialog.component.tsx
robinzigmond: Fair enough. It was [this comment](https://github.com/Sage/carbon/issues/6257#issuecomment-1688299121) that I was referring to, so you may want to check this with Luke - but if you've checked the most recent designs and confirmed you're using the right token as per them then I'm happy enough 👍 (Especially as Harpal will do UX QA anyway.)

robinzigmond: I've now seen the 025 token in both the Jira ticket and on Figma so a happy to leave this as being correct 👍

robinzigmond (dismissed):

Parsium (dismissed):

robinzigmond (dismissed):

src/components/dialog/dialog-test.stories.tsx
```diff
@@ -175,3 +176,50 @@ Default.args = {
   disableEscKey: false,
   stickyFooter: false,
 };
+
+export const WithGreyBackground = () => {
+  const [isOpen, setIsOpen] = useState(false);
```
Parsium, line 181: **comment**: we may want to set the default to `true`, just so the `Dialog` is open by default when Chromatic sees it 
```suggestion
  const [isOpen, setIsOpen] = useState(true);
```

robinzigmond (approved):

robinzigmond (approved):

PR #6442, opened 2023-11-20
test(date): playwright refactor

stephenogorman:
### Proposed behaviour

Refactor Date tests from Cypress to Playwright
Add handling of locales to playwright/index.tsx

### Current behaviour

Tests are currently Cypress

### Checklist

#### QA

### Additional context

N/A

### Testing instructions
- [ ] Run Form tests in Playwright Test Runner to check if the `date.pw.tsx` file passed
- [ ] Run all other tests in Playwright Test Runner to check none of the other `*.pw.tsx` files have regressed


src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+  });
+
+  test(`should close the picker and focus the next element in the DOM when focus is on a day element and tab pressed`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<DateInputWithButton />);
+
+    await page.focus("body");
+    await page.keyboard.press("Tab");
```
robinzigmond, line 306: **suggestion**: the Cypress version of this test checked which elements were focused at each of these tab stops. I'd prefer we did that in Playwright too, not just to check that the picker is tabbed through as expected but because the way you have it it's not obvious that the picker was ever opened, in which case the assertion that it's closed after all those tabs isn't really very useful. (This test could still pass even if we broke everything so that the picker never opened.)

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+      }
+    });
+  });
+
+  ([
+    [SIZE.SMALL, 30],
+    [SIZE.MEDIUM, 38],
+    [SIZE.LARGE, 46],
+  ] as [DateInputProps["size"], number][]).forEach(([size, approxValue]) => {
+    test(`should check the ${size} of the DateInput`, async ({
```
robinzigmond, line 641:
```suggestion
    test(`should check the ${size} size of the DateInput`, async ({
```

**suggestion** - again I know this wasn't you, but as it is this will just output eg "should check the small of the DateInput"` as a test description

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+      mount,
+      page,
+    }) => {
+      await mount(<DateInputCustom size={size} />);
+
+      const inputParent = getDataElementByValue(page, "input").locator("..");
+      await inputParent.click();
+      await inputParent.waitFor();
+      await assertCssValueIsApproximately(inputParent, "height", approxValue);
```
robinzigmond, line 650: **comment** (non-blocking): I see our Playwright `assertCssValueIsApproximately`  checks a range of 2 either side of the value, while the original Cypress version of this test checks a much narrowere range with an explicit minimum and maximum. I'd personally prefer to use those values than relax the strictness of the test without good reason. (I'm not sure why the util allows things to be up to 2 each side tbh.)

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+
+      const labelElement = getDataElementByValue(page, "label").locator("..");
+      await assertCssValueIsApproximately(labelElement, "width", labelRatio);
+      const inputElement = getDataElementByValue(page, "input").locator("..");
+      await assertCssValueIsApproximately(inputElement, "width", inputRatio);
+    });
+  });
+
+  ["10%", "30%", "50%", "80%", "100%"].forEach((maxWidth) => {
+    test(`should check maxWidth as ${maxWidth} for Date Input component`, async ({
```
robinzigmond, line 675:
```suggestion
    test(`should check maxWidth is ${maxWidth} for Date Input component`, async ({
```

**nitpick** (non-blocking)

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+    const calendarIcon = getDataElementByValue(page, "calendar");
+    await expect(calendarIcon).toBeVisible();
+    await expect(calendarIcon).toHaveCSS("color", "rgba(0, 0, 0, 0.3)");
+  });
+
+  test(`should render Date with read only prop`, async ({ mount, page }) => {
+    await mount(<DateInputCustom readOnly />);
+
+    const input = getDataElementByValue(page, "input");
+    await expect(input).toHaveAttribute("readonly", /.*/);
```
robinzigmond, line 722:
```suggestion
    await expect(input).not.toBeEditable();
```

**suggestion** - I think this is the best way in Playwright of checking something is readonly. And I particularly dislike using a regular expression to say the attribute can have absolutely any value.

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1349 @@
...
+        await input.fill(getDatetoEnter());
+        await input.click();
+
+        const pickerByText = page
+          .locator(".DayPicker-Day")
+          .filter({ hasText: "12" });
+        await pickerByText.click();
+        await expect(input).toHaveValue(formattedValueParam);
+
+        expect(callbackCount).toBe(4);
```
robinzigmond, line 992: **question** (non-blocking): obviously the test passes but I'm not sure why the callback count is exactly 4 - my understanding is that it gets called on each keypress, and since the "date to enter" has many digits, I'd expect the call count to be a lot bigger than 4.

Regardless of whether this is correct or not, I'm not sure the precise number of calls matters here in a complex scenario like this. I'd rather keep this test just for the formatted value, and for checking `onChange` is called do a test with a single call for each scenario we want it called for.

I'd make similar comments on the next 2 groups of tests as well.

playwright/index.tsx
```diff
@@ -11,11 +11,27 @@ import {
...
   focusRedesignOptOut?: boolean;
   theme?: string;
+  localeName?: keyof typeof dateLocales;
+};
+
+const computedLocale = (str: keyof typeof dateLocales) => {
+  return {
+    locale: () => str,
+    date: {
+      dateFnsLocale: () => dateLocales[str as keyof typeof dateLocales],
```
robinzigmond, line 28:
```suggestion
      dateFnsLocale: () => dateLocales[str],
```

**suggestion**: this assertion is unnecessary now (and this was one of the main reasons I suggested changing the type of `str` on line 24, which you've now done)

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1352 @@
...
+    await page.keyboard.press("Tab");
+    const arrowLeft = getDataElementByValue(page, "chevron_left").locator("..");
+    await expect(arrowLeft).toBeFocused();
+    await page.keyboard.press("Tab");
+    const arrowRight = getDataElementByValue(page, "chevron_right").locator(
+      ".."
+    );
+    await expect(arrowRight).toBeFocused();
+    await page.keyboard.press("Tab");
+    const dayPicker = page.locator(".DayPicker-Day--selected");
```
robinzigmond, line 291:
```suggestion
    const dayPicker = page.locator(`${DAY_PICKER_PREFIX}selected`);
```

**suggestion** (non-blocking): I'd missed this in the first review, but since you've bothered making a `DAY_PICKER_PREFIX` const we might as well use it wherever possible (you only seem to have used it in one place).

There are a few more places where this can be used too, just search for `.DayPicker-Day--` (I won't comment on each one individually!)

robinzigmond (approved):

src/components/date/components.test-pw.tsx
```diff
@@ -0,0 +1,153 @@
...
+      name="date-input"
+      value={state}
+      onChange={handleOnChange}
+      onBlur={handleOnBlur}
+      {...props}
+    />
+  );
+};
+
+export const DateInputValidationNewDesign = () => {
```
ZhuoyuJin, line 42: same as above

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1352 @@
...
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("Tab");
+    await page.keyboard.press(arrowKeys[0]);
+    const focusedElement = page
+      .getByRole("row")
+      .nth(1)
+      .locator("div")
+      .filter({ hasText: "1" })
+      .filter({ hasNotText: "31" });
```
ZhuoyuJin, line 608: **(question)** : Why do we need to filter hasNotText: "31"?

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1352 @@
...
+    ["fi-FI", "27.05.2022", "fi"],
+    ["de-AT", "27.05.2022", "deAT"],
+    ["ko-KR", "2022. 05. 27.", "ko"],
+    ["ar-EG", "27/05/2022", "arEG"],
+    ["hi-HI", "27/05/2022", "hi"],
+    ["sl-SI", "27. 05. 2022", "sl"],
+    ["lv", "27.05.2022.", "lv"],
+  ] as const).forEach(
+    ([localeValue, formattedValueParam, dateFnsLocaleKey]) => {
+      test(`should use ${localeValue} locale and change the formattedValue to ${formattedValueParam} after type the date`, async ({
```
ZhuoyuJin, line 1024: small typo: after **typing** the date

src/components/date/date.pw.tsx
```diff
@@ -0,0 +1,1352 @@
...
+    const input = getDataElementByValue(page, "input");
+    await input.fill(MIN_DATE);
+    await input.clear();
+    await expect(input).toHaveAttribute("Value", "");
+    expect(callbackCount).toBe(2);
+  });
+});
+
+test.describe("Accessibility tests", () => {
+  test(`should check accessibility the default component`, async ({
```
ZhuoyuJin, line 1179:
```suggestion
  test(`should check accessibility for the default component`, async ({
```

ZhuoyuJin (dismissed):

robinzigmond (approved):

ZhuoyuJin (approved):

PR #6445, opened 2023-11-22
refactor(toast): Decouple Toast component from Message component - FE-6228

DipperTheDan:
It was decided in 2019 that `Toast` should share the same functionality and styling as `Message`. This was the correct decision at the time but with the new Design System and the changes coming to these components, it will be easier to maintain these as two separate components.

### Proposed behaviour

- Split `Toast` and `Message` so that `Toast` no longer relies on `Message` for styling, icons and colours.
- Tidying up the naming used in the `Toast` style file.

### Current behaviour

`Toast` relies on `Message` for styling, icons and colours.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

There should be no functional or styling regressions with Toast.

### DIFF
--- a/src/components/toast/toast.stories.mdx
+++ b/src/components/toast/toast.stories.mdx
@@ -23,6 +23,14 @@ import * as stories from "./toast.stories";
 Toast is part of Carbon's Flashes components.
 Its purpose is to present quick confirmation of success or failure, relating to the task the user has just performed. Alternatively, use Toast messages for a longer, timely message that the user must dismiss.
 
+Various types are available:
+
+- **Error** - Shows an error or failed action or task. Includes next steps.
+- **Info** - Shows additional information when a task is completed.
+- **Success** - Shows a successful action or completion of a task.
+- **Warning** - Shows information about external factors that impact a task. For example, the user does not have access rights.
+- **Neutral** - Shows information or advice, or provides context.
+
 ## Contents
 
 - [Quick Start](#quick-start)
--- a/src/components/toast/toast.style.ts
+++ b/src/components/toast/toast.style.ts
@@ -1,12 +1,10 @@
 import styled, { css } from "styled-components";
-
-import MessageStyle from "../message/message.style";
-import MessageContentStyle from "../message/message-content/message-content.style";
 import TypeIcon from "../message/type-icon/type-icon.style";
 import StyledIconButton from "../icon-button/icon-button.style";
 import Portal from "../portal/portal";
 import baseTheme from "../../style/themes/base";
 import StyledIcon from "../icon/icon.style";
+import { MessageVariant } from "../message/message.component";
 
 const StyledPortal = styled(Portal)<{
   align?: "left" | "center" | "right";
@@ -17,7 +15,6 @@ const StyledPortal = styled(Portal)<{
   ${({ theme, isCenter, isNotice, align, alignY }) => css`
     position: fixed;
     top: 0;
-
     z-index: ${theme.zIndex.notification};
 
     ${isCenter &&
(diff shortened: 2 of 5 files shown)


src/components/toast/toast.style.ts
```diff
@@ -79,17 +75,66 @@ StyledPortal.defaultProps = {
...
+
+const ToastColourVariants = {
+  error: "var(--colorsSemanticNegative500)",
+  info: "var(--colorsSemanticInfo500)",
+  success: "var(--colorsSemanticPositive500)",
+  warning: "var(--colorsSemanticCaution500)",
+  neutral: "var(--colorsSemanticNeutral500)",
+};
+
+type ToastVariants = "error" | "info" | "success" | "warning" | "neutral";
```
edleeks87, line 87: Comment (non-blocking): we're already exporting `MessageVariants` as a type, I think it would be fine to re-use that here as it's not really coupling imo

src/components/toast/toast.style.ts
```diff
@@ -180,11 +222,21 @@ const ToastStyle = styled(MessageStyle)<{
...
-const ToastContentStyle = styled(MessageContentStyle)<{
+const StyledToastContent = styled.div<{
   isNotice?: boolean;
   isDismiss?: boolean;
 }>`
   padding: 8px 16px 8px 16px;
+  white-space: pre-wrap;
+  flex: 1;
+
+  .carbon-content__title {
```
edleeks87, line 233: Question: are these actually set anywhere? I think these have been missed when we converted [Content](https://github.com/Sage/carbon/blob/c94999f26e0bef0c72a6388afbb78c303100bdf1/src/components/content/content.component.tsx) to a styled component, the fact no one has complained leads me to think no one uses it or cares that it's not got the margin so maybe we just remove from both Toast and Message?

edleeks87 (commented):
I notice that we don't detail the `variant`s for `Toast` like we do for `Message`, do you think it's worth adding that as part of this PR?

https://carbon.sage.com/iframe.html?viewMode=docs&id=message--default&args=#designer-notes

src/components/toast/toast.style.ts
DipperTheDan: When I did see it, I did think it was odd/legacy related. I'll remove it and see what breaks.

src/components/toast/toast.style.ts
DipperTheDan: I can get that done. If they diverge even more in the future, we can always define a new type here like this.

src/components/toast/toast.style.ts
DipperTheDan: Nothing broke. Guess it was just legacy left over code.

edleeks87 (dismissed):

src/components/toast/toast.stories.mdx
```diff
@@ -23,6 +23,14 @@ import * as stories from "./toast.stories";
...
 Its purpose is to present quick confirmation of success or failure, relating to the task the user has just performed. Alternatively, use Toast messages for a longer, timely message that the user must dismiss.
 
+Various types are available:
+
+- **Error** - Shows an error or failed action or task. Includes next steps.
+- **Info** - Shows additional information when a task is completed.
+- **Success** - Shows a successful action or completion of a task.
+- **Warning** - Shows information about external factors that impact a task. For example, the user does not have access rights.
+- **Neutral** - Shows information or advice, or provides context.
+
```
Parsium, line 33: **praise**: great idea of including these from the DS 👍🏼

Parsium (approved):

Parsium (approved):

PR #6449, opened 2023-11-23
feat(numeral-date): add date format and update to align with DS

nuria1110:
### Proposed behaviour

- Component has 3 separate fields with distinct labels for 'Day', 'Month' and 'Year'. 
- Fields have a gap between them. 
- Supports YYYY/MM/DD date format. 
- New validation side border reaches the message. 
- `labelHelp` tooltip does not appear with new validation hint text. 

Screenshot 2023-11-23 at 09 35 48

Screenshot 2023-11-23 at 09 36 13

Screenshot 2023-11-23 at 11 12 40

### Current behaviour

- Component does not support YYYY/MM/DD date format. 
- Fields do not have distinct labels.
- Fields don't have a gap between them. 
- New validation side border does not reach the message. 
- `labelHelp` tooltip still shows with new validation hint text. 

Screenshot 2023-11-23 at 09 35 31

Screenshot 2023-11-23 at 09 36 02

Screenshot 2023-11-23 at 11 08 26

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/cypress/components/numeral-date/numeral-date.cy.tsx
+++ b/cypress/components/numeral-date/numeral-date.cy.tsx
@@ -23,6 +23,7 @@ import {
 import {
   numeralDateComponent,
   numeralDateInputByPosition,
+  numeralDateLabelByPosition,
 } from "../../locators/numeralDate";
 
 import {
@@ -288,26 +289,27 @@ context("Tests for NumeralDate component", () => {
       CypressMountWithProviders(
         <NumeralDateComponent dateFormat={["dd", "mm", "yyyy"]} />
       );
-      numeralDateInputByPosition(0).should("have.attr", "placeholder", "dd");
-      numeralDateInputByPosition(1).should("have.attr", "placeholder", "mm");
-      numeralDateInputByPosition(2).should("have.attr", "placeholder", "yyyy");
+
+      numeralDateLabelByPosition(0).should("have.text", "Day");
+      numeralDateLabelByPosition(1).should("have.text", "Month");
+      numeralDateLabelByPosition(2).should("have.text", "Year");
     });
 
     it('should render NumeralDate with `["mm", "dd", "yyyy"]` dateFormat prop', () => {
       CypressMountWithProviders(
         <NumeralDateComponent dateFormat={["mm", "dd", "yyyy"]} />
       );
-      numeralDateInputByPosition(0).should("have.attr", "placeholder", "mm");
-      numeralDateInputByPosition(1).should("have.attr", "placeholder", "dd");
-      numeralDateInputByPosition(2).should("have.attr", "placeholder", "yyyy");
+      numeralDateLabelByPosition(0).should("have.text", "Month");
+      numeralDateLabelByPosition(1).should("have.text", "Day");
+      numeralDateLabelByPosition(2).should("have.text", "Year");
     });
 
     it('should render NumeralDate with `["dd", "mm"]` dateFormat prop', () => {
       CypressMountWithProviders(
         <NumeralDateComponent dateFormat={["dd", "mm"]} />
       );
-      numeralDateInputByPosition(0).should("have.attr", "placeholder", "dd");
-      numeralDateInputByPosition(1).should("have.attr", "placeholder", "mm");
+      numeralDateLabelByPosition(0).should("have.text", "Day");
+      numeralDateLabelByPosition(1).should("have.text", "Month");
       numeralDateInputByPosition(2).should("not.exist");
     });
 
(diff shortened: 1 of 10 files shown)


cypress/components/numeral-date/numeral-date.cy.tsx
```diff
@@ -440,16 +440,6 @@ context("Tests for NumeralDate component", () => {
         CHARACTERS.STANDARD
       );
     });
-
-    it("should have the expected border radius styling when no search button enabled", () => {
```
DipperTheDan, line 444: **question(non-blocking):** I can see you've removed this test (we'll just just gloss over the reference to a different component altogether) but do we still have a test to check the `border-radius` on this component?

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -145,6 +148,7 @@ const incorrectDateFormatMessage =
   "Only one of these date formats is allowed: " +
   "['dd', 'mm', 'yyyy'], " +
   "['mm', 'dd', 'yyyy'], " +
+  "['yyyy', 'mm', 'dd'], " +
```
DipperTheDan, line 151: **suggestion(non-blocking):** You'll need to update the comment for `dateFormat` so that the prop tables have the correct description. Lines 82 -> 89.

src/components/numeral-date/numeral-date.spec.tsx
```diff
@@ -544,15 +542,12 @@ describe("NumeralDate", () => {
...
   describe("new validation design", () => {
-    describe("hint text/labelHelp", () => {
-      it("is visible when the prop is passed", () => {
-        wrapper = renderNewValidationsWrapper({
-          labelHelp: "Example hint text",
-        });
-        expect(wrapper.find(StyledHintText).text()).toEqual(
-          "Example hint text"
-        );
+    it("hint text is visible and tooltip is not rendered when the prop is passed", () => {
```
DipperTheDan, line 545: **question(non-blocking):** missing word in this test?

```suggestion
    it("hint text is visible and tooltip is not rendered when the labelHelp prop is passed", () => {
```

src/components/numeral-date/numeral-date.spec.tsx
```diff
@@ -691,25 +686,4 @@ describe("NumeralDate", () => {
       });
     });
   });
-
-  it("renders with the expected border radius styling on first and last inputs", () => {
```
DipperTheDan, line 695: **suggestion(non-blocking):** We'll probably still need a test asserting the border-radius which has been changed to `border-radius: var(--borderRadius050);`

src/components/numeral-date/numeral-date.stories.tsx
```diff
@@ -63,6 +65,96 @@ export const InternalValidationWarning: ComponentStory<
...
+
+      <NumeralDate
+        label="Validation as string"
+        labelHelp="Label help"
+        warning="Warning Message (Fix is optional)"
+        onChange={(e) => setValue(e.target.value)}
+        value={value}
+      />
+
+      <NumeralDate
```
DipperTheDan, line 147: **suggestion(non-blocking):** It might be good to put in the label what validation these are demonstrating. In this screenshot, warning doesn't have any real styling changes as a `boolean` value, so it just likes it usually does. 

Screenshot 2023-11-24 at 09 41 55

DipperTheDan (commented):
Good work on this, @nuria1110 😄  
Just a few minor comments but nothing major at all.

cypress/components/numeral-date/numeral-date.cy.tsx
nuria1110: There is no other reference to the search button so I think this was added by mistake to the description of the test. 

I'm also going to completely remove setting the `border-radius` from the style file as it is the same one as `Textbox`.

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -357,7 +373,7 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
           labelInline={labelInline}
           labelWidth={labelWidth}
           labelAlign={labelAlign}
-          labelHelp={labelHelp}
+          labelHelp={!validationRedesignOptIn && labelHelp}
```
edleeks87, line 376: comment: technically this is a breaking change as in theory someone _could_ have used this prop with the new validation pattern. I'll have a little investigation on this tomorrow and get back to you on it

nuria1110: I added this due to the tooltip and the hint text both being rendered with the same `labelHelp` prop, this is just meant to stop the tooltip being rendered along with the hint text as they would have the same text value.

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -368,101 +384,116 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
...
+                  <ErrorBorder
+                    warning={!!(!internalError && internalWarning)}
+                  />
+                )}
+              </>
+            )}
+
+            <StyledNumeralDate
+              name={name}
+              onKeyPress={onKeyPress}
```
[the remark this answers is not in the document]
nuria1110: Some jest tests seem to be failing so will have to raise a new ticket.

src/components/numeral-date/numeral-date.component.tsx
edleeks87: Yeah makes sense, technically it's a breaking change but tbh no one should have used it when using the new pattern and their project won't break now with the guard added, it just won't render the help etc, so let's leave it as is

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6453, opened 2023-11-23
chore: address flaky playwright tests

edleeks87:
### Proposed behaviour

Adds a animation wait to the check accessibility util
Refactors some other flaky tests
Ignores the `color-contrast` rule for `DialogFullscreen`, to be investigated separately

### Current behaviour

Several flaky playwright tests that intermittently fail the build

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/help/help.pw.tsx
+++ b/src/components/help/help.pw.tsx
@@ -6,6 +6,7 @@ import Box from "../../../src/components/box";
 import {
   getDataElementByValue,
   icon,
+  tooltipPreview,
 } from "../../../playwright/components/index";
 import helpComponent from "../../../playwright/components/help";
 import {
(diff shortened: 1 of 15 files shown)


src/components/help/help.pw.tsx
```diff
@@ -293,7 +294,10 @@ test.describe("Accessibility tests for Help component", () => {
         <HelpComponentTest isFocused={boolVal}>{tooltipText}</HelpComponentTest>
       );
 
-      await checkAccessibility(page);
+      await checkAccessibility(
+        page,
+        boolVal ? tooltipPreview(page) : undefined
+      );
```
Parsium, line 300: **suggestion**: Given we now have a conditional statement in the test, what do you think about splitting this test in two?

```
test("when isFocused prop is true, tooltip passes accessibility checks", () => { /* ... */ });

test("when isFocused prop is false, component passes accessibility checks", () => { /* ... */ });
```

Parsium (commented):
**question**: Would it be helpful, for the disabled colour contrast checks, to add TODO comments with reference to FE-6245?

Given these checks would ideally be temporarily disabled only, it may be helpful to highlight this so we can come back and re-enable them later after investigating. What do you think?

edleeks87:
> **question**: Would it be helpful, for the disabled colour contrast checks, to add TODO comments with reference to FE-6245?
> 
> Given these checks would ideally be temporarily disabled only, it may be helpful to highlight this so we can come back and re-enable them later after investigating. What do you think?

Ah I've missed some I think

src/components/help/help.pw.tsx
edleeks87: I've gone with `should check when isFocused prop is` as it's already got the describe that talks about accessibility etc

ZhuoyuJin (approved):

Parsium (approved):

PR #6469, opened 2023-12-05
feat(filterable-select): add disableDefaultFiltering prop

stripeyjumper:
### Proposed behaviour

Adds a new boolean prop to the FilterableSelect component, `disableDefaultFiltering`.

- By default `disableDefaultFiltering` is `false`, and filtering and highlighting of options is handled by the component itself. 
 - If `disableDefaultFiltering` is set to `true`, the component does not automatically filter or highlight options as the user enters a text filter.

This allows use-cases like server-side filtering of options, or rich formatting of options.  Our use-case is an address search dropdown.

The new implementation uses a `FilterableSelectList` by default, and a `SelectList` if `disableDefaultFiltering` is `true`.  It is still appropriate to use a `FilterableSelect` in this case, because we still want access to the filter which is typed into the textbox.

Screenshot of the storybook demo:
custom-filterable-select

### Current behaviour

Custom filtering and styling is not possible, because the `FilterableSelectList` component manipulates the child `Option` components as the user enters a text filter.

### Checklist

#### QA

### Testing instructions

The following CodeSandbox is an example of the previous behaviour, where filtering breaks the option styling:
https://codesandbox.io/s/carbon-quickstart-j5pb2

You can see the new behaviour by looking at the version below:
https://codesandbox.io/p/sandbox/carbon-quickstart-typescript-forked-lr2pr9

### DIFF
--- a/src/components/select/filterable-select/filterable-select.component.tsx
+++ b/src/components/select/filterable-select/filterable-select.component.tsx
@@ -84,6 +84,9 @@ export interface FilterableSelectProps
    * Higher values make for smoother scrolling but may impact performance.
    * Only used if the `enableVirtualScroll` prop is set. */
   virtualScrollOverscan?: number;
+  /** Boolean to disable automatic filtering and highlighting of options.
+   * This allows custom filtering and option styling to be performed outside of the component when the filter text changes. */
+  disableDefaultFiltering?: boolean;
 }
 
 export const FilterableSelect = React.forwardRef(
@@ -125,6 +128,7 @@ export const FilterableSelect = React.forwardRef(
       inputRef,
       enableVirtualScroll,
       virtualScrollOverscan,
+      disableDefaultFiltering = false,
       ...textboxProps
     }: FilterableSelectProps,
     ref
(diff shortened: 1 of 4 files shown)


src/components/select/filterable-select/filterable-select.stories.tsx
```diff
@@ -753,3 +753,65 @@ export const SelectionConfirmedStory = () => {
...
+        { text: "Green", color: "#008000" },
+        { text: "Orange", color: "#FFA500" },
+        { text: "Pink", color: "#FFC0CB" },
+        { text: "Purple", color: "#800080" },
+        { text: "Red", color: "#FF0000" },
+        { text: "White", color: "#FFFFFF" },
+        { text: "Yellow", color: "#FFFF00" },
+      ].filter(
+        ({ text }) =>
+          !filterText || text.toLowerCase().includes(filterText.toLowerCase())
```
edleeks87, line 777: Suggestion (non-blocking): it might be worth adding a `trim` to the `filterText` here to account for whitespace etc. It's not massively relevant to this example but likely will be in more detailed implementations so could be worth having in the code example etc

```
!filterText || (filterText.trim().length && text.toLowerCase().includes(filterText.trim().toLowerCase()))

File: src/components/select/filterable-select/filterable-select.stories.tsx [hunk 2]
```diff
@@ -753,3 +753,65 @@ export const SelectionConfirmedStory = () => {
 };
 
 SelectionConfirmedStory.parameters = { chromatic: { disableSnapshot: true } };
+
+export const CustomFilterAndOptionStyle = () => {
```
edleeks87, line 757: Praise: this is a really good demo of what this feature can support 👏

nineteen88 (approved):

nineteen88:
Looks good, and solid reasoning for this proposal. Just the one non-blocking comment to resolve from Ed should you choose to do so then.

src/components/select/filterable-select/filterable-select.stories.tsx
edleeks87: Thinking about this some more I'm happy with how it's implemented so lets just leave it and if we need to make a change later it's only docs etc

edleeks87 (dismissed):

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -125,6 +128,7 @@ export const FilterableSelect = React.forwardRef(
       inputRef,
       enableVirtualScroll,
       virtualScrollOverscan,
+      disableFiltering = false,
```
sianford, line 131: Can I suggest this prop name is renamed to something like `disableDefaultFiltering`? I was struggling to understand the purpose of this, initially thinking filtering was disabled altogether and therefore what was the purpose of this change to Filterable Select at all if no filtering was then allowed until it was explained to me.

edleeks87 (dismissed):

nineteen88 (approved):

edleeks87 (approved):

PR #6475, opened 2023-12-07
chore(card, card-row, card-column, card-footer): add data attribute support FE-4916

### LINKED ISSUE #4800: [Card] Interactive Card doesn't perform action when enter key pressed
### Current behaviour

Interactive Card component with the focus does not call its `action` when the "Enter" key is pressed.

### Expected behaviour

The `action` prop should be called when the "Enter" key is pressed

https://codesandbox.io/s/billowing-night-nzyon

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

102.16.0

Chrome

MacOS

_No response_

### Confidentiality

edleeks87:
fix #4800

### Proposed behaviour

- Adds support for Carbon `data-` attributes
- Replaces use of `cloneElement` and `Children.map` from `Card`, replacing them with context and DOM
querying. Removes `spacing` prop from `CardRow` and `CardFooter`, this value is now derived from the
value passed into `Card` via `CardContext`.

BREAKING CHANGE: The `spacing` prop is no longer surfaced on `CardRowProps` and `CardFooterProps`
interfaces

- Deprecates the old `dataRole` prop, use `data-role` instead

BREAKING CHANGE: `dataRole` has been deprecated use `data-role` instead

- Adds `onClick` and `href` props to component. When an `href` prop is passed an `a`nchor element is
rendered and the Card behaves like a link clicking or pressing the enter key when focused will
trigger the action. When an `onClick` prop is passed with no `href` then a `button` element is
rendered, clicking or hitting enter/space key will trigger the action. Passing the `draggable` prop
will negate both the `href` and `onClick` props. Surfaces a `footer` props for passing `CardFooter`
to avoid potential accessibility issues.
- new focus styling added when opt out flag not set

BREAKING CHANGE: Removes `action` and `interactive` props. Renames `cardWidth` prop to `width`.

### Current behaviour

- Component relies on cloning and mapping of children to pass props to sub-components
- `dataRole` prop has not been deprecated
- No support for rendering `button` or `a`nchor elements
- Does not trigger action when correct key presses are made
- Does not support `onClick` or `href` props, instead has `interactive` and `action` props
- Does not support new focus styling

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/hopeful-bash-58ckd9?file=/src/App.js


src/components/card/card.stories.tsx
```diff
@@ -460,151 +519,165 @@ DifferentCardFooterPadding.parameters = {
...
-        </Box>
-      </CardFooter>
     </Card>
-    <Card>
+    <Card
+      footer={
+        <CardFooter p={2}>
+          <Box display="flex" width="100%" justifyContent="center">
+            <Link icon="link" href="https://carbon.sage.com/">
+              View Stripe Dashboard
```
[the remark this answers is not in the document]
edleeks87: I agree these examples need refactoring, I intentionally only touched what I needed to as I want a full review of our stories to be done before I go through and make changes that might contradict its findings etc. I also want to avoid the mass conflicts it will cause for the storybook upgrade work

src/components/card/card.component.tsx
```diff
@@ -1,134 +1,142 @@
...
-  /** Flag to indicate if card is interactive */
-  interactive?: boolean;
-  /** Size of card for applying padding */
-  spacing?: CardSpacing;
   /** Design token for custom Box Shadow. Note: please check that the box shadow design token you are using is compatible with the Card component. */
   boxShadow?: BoxShadowsType;
   /** Design token for custom Box Shadow on hover. Interactive prop must be True. Note: please check that the box shadow design token you are using is compatible with the Card component. */
   hoverBoxShadow?: BoxShadowsType;
+  spacing?: CardContextProps["spacing"];
+  roundness?: CardContextProps["roundness"];
```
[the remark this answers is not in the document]
edleeks87: Good spot, for some reason I thought the comment on the Context interface would be pulled through here

src/components/card/card.stories.tsx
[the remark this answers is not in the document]
edleeks87: https://github.com/Sage/carbon/pull/6475#discussion_r1423729976

src/components/card/components.test-pw.tsx
```diff
@@ -10,7 +10,18 @@ import Box from "../box";
...
 export const CardComponent = (props: Partial<CardProps>) => {
   return (
-    <Card {...props}>
+    <Card
+      {...props}
+      footer={
+        <CardFooter>
+          <CardColumn>
+            <Link icon="link" href="https://carbon.sage.com/">
+              View Stripe Dashboard
```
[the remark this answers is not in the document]
edleeks87: https://github.com/Sage/carbon/pull/6475#discussion_r1423729976

src/components/card/card-test.stories.tsx
```diff
@@ -26,44 +26,58 @@ export default {
...
-          </Link>
-        </CardColumn>
-      </CardFooter>
     </Card>
   );
 };
 DefaultStory.storyName = "default";
+DefaultStory.argTypes = {
+  locale: {
+    options: ["en", "fr", "no", "es-ES", "pt-PT", "it"],
```
Parsium, line 77: **question**: I'm not sure I understand the context behind the `locale` arg here, would you be able to clarify this please?

src/components/card/card.component.tsx
```diff
@@ -65,18 +88,28 @@ const Card = ({
     }
   }, [ref]);
 
+  useLayoutEffect(() => {
```
Parsium, line 91: **question**: as this is a warning rather than a critical error, could we use a regular React effect here?

That way we wouldn't need to store `interactive` as an imperative value.

src/components/card/card.style.ts
```diff
@@ -74,8 +83,84 @@ const StyledCard = styled.div<StyledCardProps>`
...
+  outline: none;
+  text-decoration: none;
+  background-color: inherit;
+  border: none;
+  padding: 0;
+
+  ${({ spacing }) => `
+    padding: ${paddingSizes[spacing]};
+    margin: ${marginSizes[spacing]};
+  `}
```
Parsium, line 131: **suggestion**: it appears as `StyledCard` already provides padding on line 61, removing the negative margins and padding achieves the same positioning as with it.

```suggestion
```

src/components/card/card-test.stories.tsx
edleeks87: This looks incorrect, IIRC I was trying to get the controls to update but realised you need to rebuild storybook but forgot to delete this

src/components/card/card.component.tsx
edleeks87: Good point, I don't think we need the effect at all tbh

src/components/card/card.style.ts
edleeks87: Unfortunately that doesn't work as you lose the cursor and focus styling when you don't have the above. I also tried removing the styling on line 61 but it causes the footer to look misaligned.

image

src/components/card/card-row/card-row.component.tsx
```diff
@@ -1,21 +1,37 @@
...
+    }
+
+    return {};
+  };
+
   return (
     <StyledCardRow
+      id={id.current}
       spacing={spacing}
+      {...getComputedPadding()}
```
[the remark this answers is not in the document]
edleeks87: Good point there's more than likely a pure CSS solution we can go with here

src/components/card/card-row/card-row.spec.tsx
```diff
@@ -19,22 +22,29 @@ describe("CardRow", () => {
...
+      {
+        py: "var(--spacing300)",
+      },
+      undefined,
+      { modifier: "&&" }
+    );
   });
 
-  it.each<[Exclude<CardRowProps["spacing"], undefined>, string]>([
+  it.each<[Exclude<CardContextProps["spacing"], undefined>, string]>([
```
Parsium, line 35: **nitpick**: would it be worth renaming the test description slightly to reflect that `spacing` is passed via context inside of a prop?

src/components/card/card.spec.tsx
```diff
@@ -40,15 +52,37 @@ describe("Card", () => {
     "large",
   ])(
     "when spacing prop is set to %s, CardRow and CardFooter children should have a spacing prop injected with same value",
+    (spacing) => {
```
Parsium, line 55: **nitpick**: would it be worth modifying the test description slightly here, since `spacing` is no longer injected?

src/components/card/card.stories.tsx
```diff
@@ -15,12 +15,24 @@ import VerticalDivider from "../vertical-divider";
...
+    onClick={() => {}}
+    // footer={
+    //   <CardFooter>
+    //     <CardColumn>
+    //       <Link icon="link" href="https://carbon.sage.com/">
+    //         Footer link
+    //       </Link>
+    //     </CardColumn>
+    //   </CardFooter>
+    // }
```
Parsium, line 29: **issue**: I assume we would want this commented code instead of lines 51-57 given the tests now we have?

src/components/card/card-row/card-row.spec.tsx
edleeks87: I'm not convinced it matters here as it will be set via prop in the parent

src/components/card/card.spec.tsx
edleeks87: Agree this can be tidied up slightly 👍

Parsium (approved):

DipperTheDan (approved):

Parsium (approved):

DipperTheDan (dismissed):

src/components/card/card.stories.mdx
```diff
@@ -64,15 +64,18 @@ import {
...
-  <Story name="with cardWidth provided" story={stories.WithCardWidthProvided} />
+  <Story name="with width provided" story={stories.WithWidthProvided} />
 </Canvas>
 
 ### Interactive
 
-Click on the card to see the effect
+The card is "interactive" when you pass either or both of `onClick` and `href` props. 
+Passing just an `onClick` prop will render a `button` element within the `Card`. 
+Passing an `href` (with or without an `onClick`) will render an `a`nchor element.
```
ZhuoyuJin, line 77:
```suggestion
Passing an `href` (with or without an `onClick`) will render an `anchor` element.
```

ZhuoyuJin (approved):

src/components/card/card.stories.mdx
edleeks87: This was intentional as `a` is the tag name but it's called and anchor etc

ZhuoyuJin (approved):

Parsium (approved):

PR #6481, opened 2023-12-08
feat(step-flow): add new component

tomdavies73:
### Proposed behaviour

Screenshot 2023-12-08 at 16 27 29

This PR will add the `StepFlow` component which aims to help users advance through a user journey, and keep track of their progress. This is done initially with a very simple default configuration, with just the current title, current step and the total steps in the journey. This can also be contextualised further with an optional `category` prop.

Screenshot 2023-12-08 at 16 27 40

Screenshot 2023-12-08 at 16 33 37

However, a progress indicator can also be rendered via the `showProgressIndicator` prop, providing an added decorative indicator of progress, with indicators holding three individual states: completed, in progress, not completed ; each with their own background colours.

A close icon can also be rendered, with its own `onDismiss` prop, allowing consumers to pass functions to be called onClick such as closing a modal.

Screenshot 2023-12-08 at 16 27 47

The component is also flexible, allowing for multiple steps to be set with the `currentStep` and `totalSteps` props, both can be between 1-8. Also `currentStep` can never exceed the value of `totalSteps`. When these are changed, the label underneath the title will change, as well as the indicator bar (if rendered).

 The component also has built-in accessibility considerations with a focus ref being provided which provides consumers with the ability to pull focus to a visually hidden  DOM element containing a formatted string with all of the relevant titles, category, step progress contained within.

 This ensures screen reader users are given all the relevant information through semantic HTML on first navigation, and then when they progress through the journey, using buttons, for example focus is then pulled to the DOM element shown above. This will announce all of the relevant text content and then allow them to navigate down the page through all the relevant content below.

### Current behaviour

Currently there is no accessible component which provides means to track and display a step journey or sequence.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/step-flow/step-flow.component.tsx
+++ b/src/components/step-flow/step-flow.component.tsx
@@ -1,0 +1,230 @@
+import React, { useImperativeHandle, useRef, forwardRef } from "react";
+import { MarginProps } from "styled-system";
+import Icon from "../icon";
+import IconButton from "../icon-button";
+import {
+  StyledStepFlow,
+  StyledStepContent,
+  StyledStepContentText,
+  StyledStepLabelAndProgress,
+  StyledProgressIndicatorBar,
+  StyledProgressIndicator,
+  StyledTitleFocusWrapper,
+} from "./step-flow.style";
+import tagComponent, {
+  TagProps,
+} from "../../__internal__/utils/helpers/tags/tags";
+import Typography from "../typography";
+import useLocale from "../../hooks/__internal__/useLocale";
+
+export type Steps = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
+
+export interface StepFlowProps extends MarginProps, TagProps {
+  /** A category for the user journey.  */
+  category?: string;
+  /** The title of the current step.  */
+  title: string;
+  /** Set the variant of the internal 'Typography' component which contains the title.
+   * However, despite the chosen variant the styling will always be overridden.
+   */
+  titleVariant?: "h1" | "h2";
+  /** The total steps in the user journey.  */
+  totalSteps: Steps;
+  /**
+   * The current step of the user journey. If the set `currentStep` is higher than
+   * `totalSteps`the value of `currentStep` will be that of `totalSteps` instead.
+   */
+  currentStep: Steps;
+  /** Determines if the progress indicator is shown. */
+  showProgressIndicator?: boolean;
+  /** Determines if the close icon button is shown */
+  showCloseIcon?: boolean;
+  /** function runs when user click dismiss button */
+  onDismiss?: (
+    e:
+      | React.KeyboardEvent<HTMLButtonElement>
+      | React.MouseEvent<HTMLButtonElement>
+  ) => void;
+}
+
+export type StepFlowHandle = {
+  /** Programmatically focus on root container of Dialog. */
+  focus: () => void;
+} | null;
+
+export const StepFlow = forwardRef<StepFlowHandle, StepFlowProps>(
+  (
+    {
+      category,
+      title,
+      titleVariant,
+      totalSteps,
+      currentStep,
+      showProgressIndicator = false,
+      showCloseIcon = false,
+      onDismiss,
+      ...rest
+    },
+    ref
+  ) => {
+    const totalStepsArray = Array.from(
+      { length: totalSteps },
+      (_, index) => index + 1
+    );
+
+    const validatedCurrentStep =
+      currentStep > totalSteps ? totalSteps : currentStep;
+
+    let currentStepWarnTriggered = false;
+    let noRefWarnTriggered = false;
+
+    /* eslint-disable no-console */
+    if (!currentStepWarnTriggered && currentStep > totalSteps) {
+      currentStepWarnTriggered = true;
+      console.warn(
+        "[WARNING] The `currentStep` prop should not be higher than the `totalSteps`prop in `StepFlow`." +
+          " Please ensure `currentStep`s value does not exceed that of `totalSteps`, in the meantime" +
+          " we have set `currentStep` value to that of `totalSteps`, and all indicators have been marked as completed."
+      );
+    }
+    if (!noRefWarnTriggered && !ref) {
+      noRefWarnTriggered = true;
+      console.warn(
+        "[WARNING] A `ref` should be provided to ensure focus is programmatically focused back to a title div," +
+          " this ensures screen reader users are informed regarding any changes and can navigate back down the page."
+      );
+    }
+
+    const progressIndicators = totalStepsArray.map((step) => {
+      const generateDataState = () => {
+        if (step === validatedCurrentStep) {
+          return "in-progress";
+        }
+        if (step < validatedCurrentStep) {
+          return "is-completed";
+        }
+        return "not-completed";
+      };
+
+      return (
+        <StyledProgressIndicator
+          key={step}
+          aria-hidden="true"
+          data-element="progress-indicator"
+          isCompleted={step < validatedCurrentStep}
+          isInProgress={step === validatedCurrentStep}
+          data-state={generateDataState()}
+        >
+          &nbsp;
+        </StyledProgressIndicator>
+      );
+    });
+
+    const locale = useLocale();
+
+    const closeIcon = (
+      <IconButton
+        data-element="close"
+        aria-label={locale.stepFlow.closeIconAriaLabel?.()}
+        onClick={onDismiss}
+      >
+        <Icon type="close" />
+      </IconButton>
+    );
+
+    const titleRef = useRef<HTMLDivElement>(null);
+
+    useImperativeHandle<StepFlowHandle, StepFlowHandle>(
+      ref,
+      () => ({
+        focus() {
+          titleRef.current?.focus();
+        },
+      }),
+      []
+    );
+
+    const stepFlowTitle = (
+      <StyledTitleFocusWrapper
+        data-element="title-text-wrapper"
+        tabIndex={-1}
+        ref={titleRef}
+      >
+        <Typography variant={titleVariant || "h1"} data-element="title-text">
+          <Typography
+            fontWeight="900"
+            fontSize="var(--fontSizes600)"
+            lineHeight="var(--sizing375)"
+            variant="span"
+            aria-hidden="true"
+            data-element="visible-title-text"
+          >
+            {title}
+          </Typography>
+          <Typography
+            variant="span"
+            data-element="visually-hidden-title-text"
+            screenReaderOnly
+          >
+            {locale.stepFlow.screenReaderOnlyTitle(
+              title,
+              validatedCurrentStep,
+              totalSteps,
+              category
+            )}
+          </Typography>
+        </Typography>
+      </StyledTitleFocusWrapper>
+    );
+
+    const stepFlowLabel = (
+      <Typography
+        variant="span"
+        fontWeight="400"
+        fontSize="var(--fontSizes200)"
+        lineHeight="var(--sizing300)"
+        data-element="step-label"
+        aria-hidden="true"
+      >
+        {locale.stepFlow.stepLabel(validatedCurrentStep, totalSteps)}
+      </Typography>
+    );
+
+    return (
+      <StyledStepFlow {...rest} {...tagComponent("step-flow", rest)}>
+        <StyledStepContent>
+          {category ? (
+            <StyledStepContentText>
+              <Typography
+                fontWeight="500"
+                fontSize="var(--fontSizes100)"
+                lineHeight="var(--sizing250)"
+                variant="span"
+                data-element="category-text"
+                aria-hidden="true"
+              >
+                {category}
+              </Typography>
+              {stepFlowTitle}
+            </StyledStepContentText>
+          ) : (
+            stepFlowTitle
+          )}
+          {showCloseIcon ? closeIcon : null}
+        </StyledStepContent>
+        {showProgressIndicator ? (
+          <StyledStepLabelAndProgress>
+            {stepFlowLabel}
+            <StyledProgressIndicatorBar data-element="progress-indicator-bar">
+              {progressIndicators}
+            </StyledProgressIndicatorBar>
+          </StyledStepLabelAndProgress>
+        ) : (
+          stepFlowLabel
+        )}
+      </StyledStepFlow>
+    );
+  }
+);
+
+export default StepFlow;
--- a/src/components/step-flow/step-flow.pw.tsx
+++ b/src/components/step-flow/step-flow.pw.tsx
@@ -1,0 +1,290 @@
+import React from "react";
+import { test, expect } from "@playwright/experimental-ct-react17";
+import {
+  checkAccessibility,
+  getDesignTokensByCssProperty,
+} from "../../../playwright/support/helper";
+import {
+  StepFlowComponent,
+  StepFlowComponentWithRefAndButtons,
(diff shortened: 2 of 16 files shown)


src/components/step-flow/step-flow.style.ts
```diff
@@ -0,0 +1,91 @@
...
+  display: flex;
+  justify-content: space-between;
+`;
+
+const StyledStepContentText = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+const StyledCategoryText = styled(Typography)`
```
edleeks87, line 16: Comment: we can probably just use `Typography` here directly as it already supports the interface for all this styling

src/components/step-flow/step-flow.style.ts
```diff
@@ -0,0 +1,91 @@
...
+  line-height: 21px;
+  margin: 0px;
+  padding: 0px;
+`;
+
+const StyledTitleText = styled(Typography)``;
+
+const StyledTitleTextContainer = styled.span``;
+
+const StyledVisibleTitleText = styled.span`
```
edleeks87, line 29: Comment: if you simplify the HTML like Ive suggested above you can probably just pass these to `Typography` directly

src/components/step-flow/step-flow.style.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  if (isCompleted) {
+    return "var(--colorsSemanticPositive500)";
+  }
+  return "var(--colorsActionDisabled600)";
+}
+
+const StyledProgressIndicator = styled.span<StyledProgressIndicatorProps>`
+  background-color: ${({ isCompleted, isInProgress }) =>
+    calculateProgressIndicatorColor({ isCompleted, isInProgress })};
```
edleeks87, line 71: suggestion (non-blocking): you don't need to destructure here `props` are passed implicitly so you could just do:

```
background-color: ${calculateProgressIndicatorColor};

File: src/components/step-flow/step-flow.style.ts [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+  width: 100%;
+  height: 8px;
```
edleeks87, line 73: comment: wherever possible we should use tokens such as `--spacing150`, `--borderRadius100` etc.

Comment relates to whole file (font sizes etc)

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+export type Steps = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
+
+export interface StepFlowProps extends MarginProps, TagProps {
+  /** A category for the user journey.  */
+  category?: string;
+  /** Set the variant of the internal 'Typography' component which contains the category.
+   * However, despite the chosen variant the styling will always be overridden.
+   */
+  categoryTypographyVariant?: VariantTypes;
```
edleeks87, line 33:
```suggestion
  categoryVariant? :Extract<VariantTypes, "h1" | "h2" | "h3" | "h4" | "h5" | "p">;
```

Naming is non-blocking for me but I think we should limit the elements here as some don't seem viable for this use case

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** The title of the current step.  */
+  title: string;
+  /** Set the variant of the internal 'Typography' component which contains the title.
+   * However, despite the chosen variant the styling will always be overridden.
+   */
+  titleTypographyVariant?: VariantTypes;
```
edleeks87, line 39:
```suggestion
  titleVariant? :Extract<VariantTypes, "h1" | "h2" | "h3" | "h4" | "h5" | "p">;
```

Naming is non-blocking for me but I think we should limit the elements here as some don't seem viable for this use case

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+  /** A ref which will be applied to the DOM element which contains the component's title.
+   * Used to focus on the title, to ensure screen reader user's are aware of title changes,
+   * and allow navigation from the StepFlow component to the contents below.
+   */
+  titleRef: React.RefObject<HTMLDivElement>;
```
edleeks87, line 44: Comment: lets go with the `useImperativeHandle` approach we used in [dialog](https://github.com/Sage/carbon/blob/9d4252a4ad69274be2eff6bc11d5f0f240ec6ffc/src/components/dialog/dialog.component.tsx#L142) here. That way we can support a `forwardRef` (instead of `titleRef`) and only surface the functionality we want here. It will allow us to remove the `titleTabIndex` prop as well (should just be set to -1 imo)

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+  showCloseIcon?: boolean;
+  /** function runs when user click dismiss button */
+  onDismiss?: (
+    e:
+      | React.KeyboardEvent<HTMLButtonElement>
+      | React.MouseEvent<HTMLButtonElement>
+  ) => void;
+}
+
+let currentStepWarnTriggered = false;
```
edleeks87, line 68: comment: should this be within the component as in theory we'd want it to apply for any incorrect implementation of `StepFlow` not just once for all of them. We don't want it to repeatedly warn for the same one so the flag is needed

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+              aria-hidden="true"
+            >
+              {category}
+            </StyledCategoryText>
+          ) : null}
+          <StyledTitleText
+            variant={titleTypographyVariant || category ? "h2" : "h1"}
+            data-element="title-text"
+          >
+            <StyledTitleTextContainer
```
edleeks87, line 151: comment: I'm not sure we need `StyledTitleTextContainer` or `StyledVisibleTitleText` here. If we add `aria-hidden` to `Typography` (probably private prop for now) we can use it directly (see comment in style file around this)

src/components/step-flow/step-flow.component.tsx
```diff
@@ -0,0 +1,193 @@
...
+      <Icon type="close" />
+    </IconButton>
+  );
+
+  const locale = useLocale();
+
+  return (
+    <StyledStepFlow {...rest} {...tagComponent("step-flow", rest)}>
+      <StyledStepContent>
+        <StyledStepContentText>
```
edleeks87, line 137: Comment: this would render an empty container if no `category` is set. We could probably get away with using `Typography` directly here is we add support for `aria-hidden` etc

src/components/step-flow/step-flow.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+            totalSteps: 6,
+            currentStep: 1,
+            showCloseIcon: true,
+            titleRef: mockRef,
+          }}
+        />
+      );
+
+      expect(
+        container.querySelector('[data-element="close"]')
```
edleeks87, line 223: comment: if we set and aria-label (see above) we can get by label here instead of using a container to query.

edleeks87 (commented):
Great effort @tomdavies73, especially on the considerations you've given to accessibility. I've left some comments but they're fairly straightforward to address, on the whole really good work 👏

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+  stepFlowDismissIcon,
+} from "../../../playwright/components/step-flow";
+
+import { button } from "../../../playwright/components";
+
+import { CHARACTERS } from "../../../playwright/support/constants";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+
+test.describe("Prop checks for Portrait component", () => {
```
DipperTheDan, line 31: **typo:** 

```suggestion
test.describe("Prop checks for Step Flow component", () => {
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  test("background colour token should be correct when 'showCloseIcon' is true and the indicator is in progress", async ({
+    mount,
+    page,
+  }) => {
+    await mount(
+      <StepFlowComponent currentStep={1} totalSteps={1} showProgressIndicator />
+    );
+
+    const backgroundColorToken = await getDesignTokensByCssProperty(
```
DipperTheDan, line 151: **suggestion(non-blocking):** You may not need this await here.

```suggestion
    const backgroundColorToken = getDesignTokensByCssProperty(
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  test("background colour token should be correct when 'showCloseIcon' is true and the indicator is completed", async ({
+    mount,
+    page,
+  }) => {
+    await mount(
+      <StepFlowComponent currentStep={2} totalSteps={3} showProgressIndicator />
+    );
+
+    const backgroundColorToken = await getDesignTokensByCssProperty(
```
DipperTheDan, line 168: suggestion(non-blocking): Same again. You may not need this await here.

```suggestion
    const backgroundColorToken = getDesignTokensByCssProperty(
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  test("background colour should be correct when 'showCloseIcon' is true and the indicator is not completed", async ({
+    mount,
+    page,
+  }) => {
+    await mount(
+      <StepFlowComponent currentStep={2} totalSteps={3} showProgressIndicator />
+    );
+
+    const backgroundColorToken = await getDesignTokensByCssProperty(
```
DipperTheDan, line 185: suggestion(non-blocking): Same again. You may not need this await here.

```suggestion
    const backgroundColorToken = getDesignTokensByCssProperty(
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+    await mount(
+      <StepFlowComponent
+        showCloseIcon
+        onDismiss={() => {
+          callbackCount += 1;
+        }}
+      />
+    );
+    await stepFlowDismissIcon(page).click();
+    expect(callbackCount).toBe(1);
```
DipperTheDan, line 210: **nitpick(non-blocking):** space here otherwise the QA's will break your legs.

```suggestion
    await stepFlowDismissIcon(page).click();
    
    expect(callbackCount).toBe(1);
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+test.describe("Ref checks for StepFlow component", () => {
+  test("should focus on title DOM element when a focus ref is passed", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<StepFlowComponentWithRefAndButtons />);
+
+    await expect(stepFlowTitleTextWrapper(page)).not.toBeFocused();
+    await button(page).nth(1).click();
+    await expect(stepFlowTitleTextWrapper(page)).toBeFocused();
```
DipperTheDan, line 223: nitpick(non-blocking): Same again with the broken legs.
```suggestion
    await button(page).nth(1).click();
    
    await expect(stepFlowTitleTextWrapper(page)).toBeFocused();
```

src/components/step-flow/step-flow.pw.tsx
```diff
@@ -0,0 +1,275 @@
...
+  test("should pass accessibility checks when component is rendered with required props", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<StepFlowComponent />);
+
+    await checkAccessibility(page);
+  });
+
+  test("should pass accessibility checks when component is rendered with a ref", async ({
```
DipperTheDan, line 237: suggestion(non-blocking): Just because the example is called `StepFlowComponentWithRefAndButtons`

```suggestion
  test("should pass accessibility checks when component is rendered with a ref and buttons", async ({
```

DipperTheDan (commented):
Excellent work on this, @tomdavies73. Just a few comments from me, nothing serious though 😄👍🏻

src/components/step-flow/step-flow.spec.tsx
```diff
@@ -49,60 +53,134 @@ describe("Step Flow component", () => {
...
       render(
+        <StepFlow title="baz" currentStep={5} totalSteps={6} category="bar" />
+      );
+
+      expect(screen.getByText("baz")).toBeInTheDocument();
+    });
+
+    it("when the 'titleVariant' prop is not passed, the variant is h1 by default", () => {
+      const { container } = render(
+        <StepFlow title="this title is a h1" currentStep={5} totalSteps={6} />
```
edleeks87, line 72: comment (non-blocking, don't actually change anything): I suppose depending if you say aitch or haitch could affect whether it's `a h1` or `an h1` here 😆

src/components/step-flow/step-flow.spec.tsx
```diff
@@ -358,4 +423,87 @@ describe("Step Flow component", () => {
...
+              Press me to refocus on Dialog
+            </Button>
+          </div>
+        );
+      };
+
+      const user = userEvent.setup();
+      const { container } = render(<MockComponent />);
+      const button = screen.getByRole("button");
+      button.focus();
```
edleeks87, line 450: comment: I don't think you need to focus the button here (just click) and given the what this is testing it may be a little confusing

src/components/step-flow/step-flow.stories.mdx
```diff
@@ -39,9 +40,9 @@ import StepFlow from "carbon-react/lib/components/step-flow";
 
 ## Examples
 
-The `titleRef` prop is mandatory, and must be passed when the component is used to ensure focus is pulled to the components title whenever
+A `ref` is **strongly** recommended to be used, to ensure focus is programmatically moved to a a title div when
```
edleeks87, line 43: Suggestion (non-blocking):

```
To ensure focus is programmatically moved to the `title` of a given `StepFlow` when a user advances through each step,  pass a `ref` is  **strongly** recommended. This ensures screen reader users are made aware of all information throughout the user journey

edleeks87 (commented):
Real good updates @tomdavies73, especially how thorough you've been in adding additional testing 👍

src/components/step-flow/step-flow.pw.tsx
tomdavies73: I think we do need an await here as its a promise, without it `backgroundColorToken` returns `undefined`

src/components/step-flow/step-flow.pw.tsx
tomdavies73: I think we do need an await here as its a promise, without it `backgroundColorToken` returns `undefined`

src/components/step-flow/step-flow.pw.tsx
tomdavies73: I think we do need an await here as its a promise, without it `backgroundColorToken` returns `undefined`

src/components/step-flow/step-flow.pw.tsx
tomdavies73: Genuinely do not know how i missed this

edleeks87 (approved):

DipperTheDan (dismissed):

stephenogorman (approved):

edleeks87 (approved):

stephenogorman (approved):

DipperTheDan (approved):

PR #6482, opened 2023-12-11
feat(option, option-row): add disabled prop

nuria1110:
### Proposed behaviour

Add `disabled` prop to `Option` and `OptionRow` in `Select` component. A disabled option will:
- Have disabled styles
- Have an `aria-disabled` attribute
- The user cannot click or keyboard navigate to this option
- The option will still appear in a filtered select but cannot be selected

### Current behaviour

`Option` and `OptionRow` in `Select` component do not have the option to be disabled. 

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/select/filterable-select/filterable-select-test.stories.tsx
+++ b/src/components/select/filterable-select/filterable-select-test.stories.tsx
@@ -650,3 +650,27 @@ export const FilterableSelectNestedInDialog = ({
     </Dialog>
   );
 };
+
+export const FilterableSelectWithDisabledOption = () => {
+  const [value, setValue] = useState("");
+
+  function onChangeHandler(event: React.ChangeEvent<HTMLInputElement>) {
+    setValue(event.target.value);
+  }
+
+  return (
+    <FilterableSelect
+      name="testing"
+      value={value}
+      onChange={onChangeHandler}
+      openOnFocus
+      label="Test"
+      placeholder=" "
+    >
+      <Option value="1" text="One" />
+      <Option value="2" text="Two" disabled />
+      <Option value="3" text="Three" />
+      <Option value="4" text="Four" />
+    </FilterableSelect>
+  );
+};
--- a/src/components/select/multi-select/multi-select-test.stories.tsx
+++ b/src/components/select/multi-select/multi-select-test.stories.tsx
@@ -598,3 +598,27 @@ export const MultiSelectErrorOnChangeNewValidation = () => {
     </CarbonProvider>
   );
 };
+
+export const MultiSelectWithDisabledOption = () => {
+  const [value, setValue] = useState<string[]>([]);
+
+  function onChangeHandler(event: React.ChangeEvent<HTMLInputElement>) {
+    setValue((event.target.value as unknown) as string[]);
+  }
+
+  return (
+    <MultiSelect
+      name="testing"
+      value={value}
+      onChange={onChangeHandler}
+      openOnFocus
+      label="Test"
+      placeholder=" "
+    >
+      <Option value="1" text="One" />
+      <Option value="2" text="Two" disabled />
+      <Option value="3" text="Three" />
+      <Option value="4" text="Four" />
+    </MultiSelect>
+  );
+};
--- a/src/components/select/multi-select/multi-select.pw.tsx
+++ b/src/components/select/multi-select/multi-select.pw.tsx
@@ -16,6 +16,7 @@ import {
   MultiSelectNestedInDialog,
   MultiSelectErrorOnChangeNewValidation,
   SelectionConfirmed,
+  MultiSelectWithDisabledOption,
 } from "../../../../src/components/select/multi-select/components.test-pw";
 import {
   commonDataElementInputPreview,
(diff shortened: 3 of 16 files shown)


src/components/select/select-list/select-list.component.tsx
```diff
@@ -292,6 +292,10 @@ const SelectList = React.forwardRef(
           nextIndex = getNextHighlightableItemIndex(key, nextIndex);
         }
 
+        if (nextElement.props.disabled) {
```
edleeks87, line 295: suggestion: you could add this check to the `if` above

```
if (
          React.isValidElement(nextElement) &&
          ((nextElement.type !== Option &&
          nextElement.type !== OptionRow) ||
          nextElement.props.disabled)
        ) {
          nextIndex = getNextHighlightableItemIndex(key, nextIndex);
        }

File: src/components/select/select-list/select-list.component.tsx [hunk 2]
```diff
@@ -390,6 +394,10 @@ const SelectList = React.forwardRef(
             return;
           }
 
+          if (currentOption.props.disabled) {
```
edleeks87, line 397: Praise: good catch adding this guard, I was initially confused by it but see it prevents a filter text selection

edleeks87: @harpalsingh @ljemmo do you know if we should be excluding disabled Options from the filter? Currently they are displayed when there's a match but obviously can't be selected

image

edleeks87 (commented):
Really good work @nuria1110 👏

src/components/select/filterable-select/filterable-select-test.stories.tsx
```diff
@@ -695,3 +695,38 @@ export const SelectionConfirmed = () => {
     </>
   );
 };
+
+export const FilterableSelectWithDisabledOption = () => {
```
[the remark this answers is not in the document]
nuria1110: @harpalsingh @ljemmo currently the input field will autofill with the first Option available when typing, even if it is a disabled Option. Should we prevent this from happening if the first match is not available or completely excluding it from the filtered list as Ed suggested above?

Parsium (commented):
Great work on this @nuria1110 👏🏼 - really impressed how you've accounted for the possible edge cases given the complexity involved with these components.

src/components/select/select-list/select-list.component.tsx
harpalsingh: I think the requirement of a diabled select option isnt great imo, but based on the specs are someone can still see the option, then in theory it should be displayed since it "exists" but cant be selected. I will have a chat with @tempertemper and @ljemmo about this.

src/components/select/filterable-select/filterable-select-test.stories.tsx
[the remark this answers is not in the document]
harpalsingh: It should not select or auto fill a disabled option, because a user would not be able to do it with keyboard/mouse either. As above, will discuss if we need to show in filter list when type ahead filter is being applied.

edleeks87 (approved):

src/components/select/multi-select/multi-select-test.stories.tsx
```diff
@@ -643,3 +643,38 @@ export const SelectionConfirmed = () => {
...
 };
+
+export const MultiSelectWithDisabledOption = () => {
+  const [value, setValue] = useState<string[]>([]);
+  const [confirmedSelections, setConfirmedSelections] = useState<string[]>([]);
+
+  const handleChange = (event: CustomSelectChangeEvent) => {
+    setValue((event.target.value as unknown) as string[]);
+    if (event.selectionConfirmed) {
+      setConfirmedSelections((event.target.value as unknown) as string[]);
```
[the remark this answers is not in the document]
Parsium: Going through this again, it seems the `Select` components have been designed in a way that it expects `value` to be `string[]`. I'd recommend when we refactor the `Select` components as a whole later, that we revisit this.

src/components/select/multi-select/multi-select.pw.tsx
```diff
@@ -1551,6 +1552,29 @@ test("should not add an empty Pill when filter text does not match option text",
...
 
+test("should not select a disabled option when a filter is typed", async ({
+  mount,
+  page,
+}) => {
+  await mount(<MultiSelectWithDisabledOption />);
+
+  await dropdownButton(page).click();
+  const inputElement = selectInput(page);
+  await inputElement.type("t");
```
Parsium, line 1563: **note**: not essential to add right now, but it seems [Playwright has deprecated the `locator.type` action in a later version](https://playwright.dev/docs/release-notes#deprecations). So going forward I'd recommend using the `locator.fill` action instead for typing text into input fields.

Parsium (dismissed):

edleeks87 (approved):

Parsium (approved):

harpalsingh:
Worth looking at adding to main stories later,  but will confirm if we want any other updates to them before we do that, so this is good go as is.

stephenogorman (approved):

PR #6483, opened 2023-12-12
feat(progress-tracker): add inline label and label width prop

nuria1110:
### Proposed behaviour

- Add "left" option to `labelsPosition` prop to position the labels to the left of the progress bar. 
- Add `labelWidth` prop to enable consumers to set the width of the label, which can help to ensure consistent label lengths and facilitate comparison between multiple progress bars when the label is inline. 

Screenshot 2023-12-12 at 10 54 28

### Current behaviour

Progress Tracker component does not support inline labels. 

### Checklist

#### QA

### Testing instructions


src/components/progress-tracker/progress-tracker.style.ts
```diff
@@ -79,16 +85,22 @@ const StyledDescription = styled.span`
...
 const StyledValuesLabel = styled.span<
-  Pick<ProgressTrackerProps, "size" | "labelsPosition">
+  Pick<ProgressTrackerProps, "size" | "labelsPosition" | "labelWidth">
 >`
   text-align: start;
   display: flex;
   justify-content: flex-start;
   gap: 4px;
   font-size: ${({ size }) => size && fontSizes[size]};
   ${({ labelsPosition }) => labelsPosition === "bottom" && "margin-top: 8px"};
```
edleeks87, line 95: suggestion: you could refactor this to destructure it once rather than three times if you want (non-blocking). We should use the token value though.

```
  const labelsPositionMargin = { top: "bottom", bottom: "top", left: "right" };

 ...
 const StyledValuesLabel = styled.span<
  Pick<ProgressTrackerProps, "size" | "labelsPosition" | "labelWidth">
>`
...
  ${({ labelsPosition }) => `margin-${labelsPositionMargin[labelsPosition]}: var(--spacing100)`};
```

src/components/progress-tracker/progress-tracker.stories.tsx
```diff
@@ -102,6 +102,44 @@ export const DefaultLabelValueLabelsPositionBottom: ComponentStory<
   );
 };
 
+export const DefaultLabelValueLabelsPositionLeft: ComponentStory<
```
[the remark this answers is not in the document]
nuria1110: I've left the story as it is and added a test story with a few different `labelWidth` options. In doing this I noticed the width of the label wasn't behaving as intended, hence adding `flex-grow` & `flex-shrink` to set the width of the bar and the label so that the bar just takes up the remaining space.

src/components/progress-tracker/progress-tracker.style.ts
```diff
@@ -58,7 +64,7 @@ const StyledProgressBar = styled.span<
       : "var(--borderRadius400)"};
     overflow-x: hidden;
     height: ${getHeight(size)};
-    width: 100%;
+    flex-grow: 1;
```
tomdavies73, line 67: **question**: I think flex-grow here and flex-shrink on the label may be cancelling each other out? It may be worth playing around to see if both attributes are necessary

src/components/progress-tracker/progress-tracker.pw.tsx
```diff
@@ -243,6 +246,23 @@ test.describe("Check props for ProgressTracker component", () => {
     );
   });
 
+  test(`should position current value label to the left of tracker line when labelsPosition prop is "left"`, async ({
```
tomdavies73, line 249: **nit-pick** (non-blockiing): as we're not using string interpolation here we can use speech marks for the test description instead of an apostrophe

tomdavies73: I know this is the same for the other tests around this one, happy to leave as is

src/components/progress-tracker/progress-tracker.style.ts
nuria1110: yes you seem to be right, flex-shrink by itself seems to work fine

tomdavies73 (approved):

src/components/progress-tracker/progress-tracker-test.stories.tsx
```diff
@@ -51,6 +51,43 @@ export const Default = (args: ProgressTrackerProps) => {
 
 Default.storyName = "default";
 
+export const LeftLabelWithLabelWidth = (args: ProgressTrackerProps) => {
```
edleeks87, line 54: praise: Good idea having it as a test story

edleeks87 (approved):

src/components/progress-tracker/progress-tracker.pw.tsx
```diff
@@ -243,6 +246,23 @@ test.describe("Check props for ProgressTracker component", () => {
     );
   });
 
+  test("should position current value label to the left of tracker line when labelsPosition prop is left", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<ProgressTrackerComponent labelsPosition="left" />);
+
+    const currentValueLabel = progressTrackerMinVal(page);
```
sianford, line 255: Question: Is there any reason why these `const` can't be pulled out to the top of the file as they're used in three test cases.

sianford (approved):

PR #6486, opened 2023-12-14
feat(popover-container): add prop to disable animation - FE-5881

### LINKED ISSUE #6002: Allow popover container content to be fixed
### Desired behaviour

Allow users to set the "transition" style that's applied to the "popover-container-content" when using the PopoverContainer component, or remove/disable it.
QuickPanelStyle

### Current behaviour

Popover containers can jump around the screen when scrolling up or down: 

### Suggested Solution

Allow the style to be disabled/overridden if the desire is to have a fixed popover container, which is added here: https://github.com/Sage/carbon/blob/master/src/components/popover-container/popover-container.style.ts#L45

_No response_

### Confidentiality

DipperTheDan:
Introduces the `disableAnimation` prop to the component so that the transitions can be disabled.

fixes #6002

### Proposed behaviour

- Add `disableAnimation` prop
- Disable animation if the user specifies that they want reduced motion
- Skip flaky Cypress test that uses `getBoundingClientRect()` to assert whether the button is covered or not.

### Current behaviour

- No way to disable animation
- Component does not cater for users who prefer no motion.

### Checklist

#### QA

### Additional context

Comment out of the flaky Cypress test that uses `getBoundingClientRect()` to assert whether the button is covered or not is covered by a Chromatic snapshot. This means we'll be able to see in future if this feature happens to regress. It will also be converted to PW.

### Testing instructions

- No functional regressions with component.
- If reduce motion is set, there should be no animations. 
- If `disableAnimation` is set, there should be no animations.

### DIFF
--- a/cypress/components/popover-container/popover-container.cy.tsx
+++ b/cypress/components/popover-container/popover-container.cy.tsx
@@ -200,31 +200,33 @@ context("Test for Popover Container component", () => {
       popoverContainerContent().should("be.visible");
     });
 
-    it.each([
-      [true, 93, 170],
-      [false, 117, 194],
-    ])(
-      "should render Popover Container with shouldCoverButton prop set to %s",
-      (boolean, yAndTopValueMin, bottomValueMin) => {
-        CypressMountWithProviders(
-          <div
-            style={{
-              height: 330,
-            }}
-          >
-            <PopoverContainerComponent shouldCoverButton={boolean} />
-          </div>
-        );
+    // TODO: Test is flaky. Investigate whether this is resolved when converting to Playwright as `.not.toBeInViewport()` could possibly be used instead of `getBoundingClientRect()`.
+    // We do have a Chromatic snapshot for this test case also.
+    // it.each([
+    //   [true, 93, 170],
+    //   [false, 117, 194],
+    // ])(
+    //   "should render Popover Container with shouldCoverButton prop set to %s",
+    //   (boolean, yAndTopValueMin, bottomValueMin) => {
+    //     CypressMountWithProviders(
+    //       <div
+    //         style={{
+    //           height: 330,
+    //         }}
+    //       >
+    //         <PopoverContainerComponent shouldCoverButton={boolean} />
+    //       </div>
+    //     );
 
-        popoverContainerContent().then(($el) => {
-          const position = $el[0].getBoundingClientRect();
+    //     popoverContainerContent().then(($el) => {
+    //       const position = $el[0].getBoundingClientRect();
 
-          cy.wrap(position.bottom).should("be.lessThan", bottomValueMin);
-          cy.wrap(position.top).should("be.lessThan", yAndTopValueMin);
-          cy.wrap(position.y).should("be.lessThan", yAndTopValueMin);
-        });
-      }
-    );
+    //       cy.wrap(position.bottom).should("be.lessThan", bottomValueMin);
+    //       cy.wrap(position.top).should("be.lessThan", yAndTopValueMin);
+    //       cy.wrap(position.y).should("be.lessThan", yAndTopValueMin);
+    //     });
+    //   }
+    // );
 
     it.each([...keyToTrigger])(
       "should open Popover Container using %s keyboard key",
--- a/src/components/popover-container/popover-container-test.stories.tsx
+++ b/src/components/popover-container/popover-container-test.stories.tsx
@@ -1,5 +1,6 @@
-/* eslint-disable react/prop-types */
-import React from "react";
+import React, { useState } from "react";
+import Button from "../button";
+import Box from "../box";
 import PopoverContainer, {
   PopoverContainerProps,
 } from "./popover-container.component";
@@ -7,7 +8,7 @@ import { Select, Option } from "../select";
 
 export default {
   title: "Popover Container/Test",
-  includeStories: ["Default", "WithSelect"],
+  includeStories: ["Default", "WithSelect", "InAScrollableBlock"],
   parameters: {
     info: { disable: true },
     chromatic: {
@@ -53,7 +54,7 @@ Default.story = {
 export const PopoverContainerComponent = (
   props: Partial<PopoverContainerProps>
 ) => {
-  const [isOpen, setIsOpen] = React.useState(true);
+  const [isOpen, setIsOpen] = useState(true);
 
   const onOpen = () => setIsOpen(isOpen);
   const onClose = () => setIsOpen(!isOpen);
@@ -95,5 +96,45 @@ export const PopoverContainerWithSelect = () => {
         </Select>
       </PopoverContainer>
     </div>
+  );
+};
+
+export const InAScrollableBlock = () => {
+  return (
+    <Box>
+      <Box bg="#ccd6dbff" height={500} width={1100} />
+      <Box height={400} overflow="scroll">
+        <Box height={400} position="fixed">
+          <PopoverContainer
+            title="This is the title"
+            disableAnimation
+            renderOpenComponent={({
+              "data-element": dataElement,
+              onClick,
+              ref,
+              "aria-label": ariaLabel,
+              id,
+              "aria-expanded": ariaExpanded,
+              "aria-haspopup": ariaHasPopup,
+            }) => (
+              <Button
+                iconType="settings"
+                iconPosition="after"
+                data-element={dataElement}
+                aria-label={ariaLabel}
+                aria-haspopup={ariaHasPopup}
+                aria-expanded={ariaExpanded}
+                ref={ref}
+                id={id}
+                onClick={onClick}
+                m={2}
+              />
+            )}
+          >
+            <Button>View all notifications</Button>
+          </PopoverContainer>
+        </Box>
+      </Box>
+    </Box>
   );
 };
--- a/src/components/popover-container/popover-container.component.tsx
+++ b/src/components/popover-container/popover-container.component.tsx
@@ -3,6 +3,7 @@ import { PaddingProps } from "styled-system";
 import { Transition, TransitionStatus } from "react-transition-group";
 import { flip, offset } from "@floating-ui/dom";
 
+import useMediaQuery from "../../hooks/useMediaQuery";
 import {
   PopoverContainerWrapperStyle,
   PopoverContainerHeaderStyle,
@@ -123,6 +124,8 @@ export interface PopoverContainerProps extends PaddingProps {
   closeButtonAriaLabel?: string;
   /** Container aria label */
   containerAriaLabel?: string;
+  /** Disables the animation for the component */
+  disableAnimation?: boolean;
 }
 
 const popoverMiddleware = [
@@ -148,6 +151,7 @@ export const PopoverContainer = ({
   openButtonAriaLabel,
   closeButtonAriaLabel = "close",
   containerAriaLabel,
+  disableAnimation = false,
   ...rest
 }: PopoverContainerProps) => {
   const isControlled = open !== undefined;
@@ -163,6 +167,10 @@ export const PopoverContainer = ({
     : undefined;
 
   const isOpen = isControlled ? open : isOpenInternal;
+
+  const reduceMotion = !useMediaQuery(
+    "screen and (prefers-reduced-motion: no-preference)"
+  );
 
   useEffect(() => {
     if (isOpen && closeButtonRef.current)
@@ -264,6 +272,9 @@ export const PopoverContainer = ({
         {(state: TransitionStatus) =>
           isOpen && (
             <Popover
+              popoverStrategy={
+                disableAnimation || reduceMotion ? "fixed" : "absolute"
+              }
               reference={popoverReference}
               placement={position === "right" ? "bottom-start" : "bottom-end"}
               {...(shouldCoverButton && { middleware: popoverMiddleware })}
@@ -277,6 +288,7 @@ export const PopoverContainer = ({
                 aria-describedby={ariaDescribedBy}
                 p="16px 24px"
                 ref={popoverContentNodeRef}
+                disableAnimation={disableAnimation || reduceMotion}
                 {...filterStyledSystemPaddingProps(rest)}
               >
                 <PopoverContainerHeaderStyle>
--- a/src/components/popover-container/popover-container.stories.mdx
+++ b/src/components/popover-container/popover-container.stories.mdx
@@ -122,6 +122,17 @@ You can do it easly in this way:
   <Story name="filter" story={stories.Filter} />
 </Canvas>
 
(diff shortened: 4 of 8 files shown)


cypress/components/popover-container/popover-container.cy.tsx
```diff
@@ -200,32 +200,6 @@ context("Test for Popover Container component", () => {
       popoverContainerContent().should("be.visible");
     });
 
-    it.each([
```
robinzigmond, line 203: **comment**: I'm not a fan of just completely removing tests provided they were there for good reason, which it seems like this one is (it's testing that a prop works as expected).

If the test was flaky I'd rather fix that up if we easily can - and if not then at least leave the test there in an `it.skip` with a comment explaining why it's been skipped and a note to investigate this in the future. (In this case I assume we could leave it till we convert the tests to Playwright.)

src/components/popover-container/popover-container-test.stories.tsx
```diff
@@ -97,3 +104,131 @@ export const PopoverContainerWithSelect = () => {
     </div>
   );
 };
+
+export const InAScrollableBlock = () => {
```
robinzigmond, line 108: **question** (non-blocking): I'm not really sure what value this test story is adding, or what its relevance is to the PR since it doesn't use the new prop you've added. (But I'm fine with keeping it if we do want to manually test how PopoverContainer behaves in a scrollable container - I understand the original issue arise from a similar scenario.)

src/components/popover-container/popover-container-test.stories.tsx
```diff
@@ -97,3 +104,131 @@ export const PopoverContainerWithSelect = () => {
...
+  const [value1, setValue1] = useState(false);
+  const [value2, setValue2] = useState(false);
+  const [value3, setValue3] = useState(false);
+  const [value4, setValue4] = useState(false);
+  const Logo = () => <img height={28} src={carbonLogo} alt="Carbon logo" />;
+  return (
+    <Box>
```
robinzigmond, line 115: **comment** (non-blocking): I'm probably missing something, but as far as I understand it we only need the PopoverContainer here and the scrollable Box that wraps it. I don't see what value the GlobalHeader, menus and TileSelect add to the story - indeed the TileSelect confused me when i was first looking at it (especially as I was on the Docs tab and there you don't even seem to be able to see or scroll to the PopoverContainer that's below it! It's fine on Canvas though, so not an issue as this is only for internal QA purposes.)

src/components/popover-container/popover-container.component.tsx
```diff
@@ -164,6 +168,10 @@ export const PopoverContainer = ({
 
   const isOpen = isControlled ? open : isOpenInternal;
 
+  const reduceMotion = !useMediaQuery(
+    "screen and (prefers-reduced-motion: reduce)"
```
robinzigmond, line 172:
```suggestion
    "screen and (prefers-reduced-motion: no-preference)"
```

**suggestion**: I may well be getting confused as I didn't get much sleep last night and I've only had one coffee so far 😅 , but if you look up how we're using this media query in `Loader` and `LoaderBar`, you'll see it's as I've suggested, which is nearly the opposite of what you've got here. ("nearly" because the `screen` check is in both - I'm not sure that's needed but I'm pretty sure it does no harm and in any case we've done that in the other components too and we should try to be consistent). And I'm pretty sure those other components are the ones that have this right - the logic seems to work that way for me, and also testing it out with the OS preference on and off, Loader/LoaderBar are working as intended. (I'd like to say which works better with PopoverContainer but unfortunately the animation there is so small I can't actually personally tell the difference!)

src/components/popover-container/popover-container.stories.mdx
```diff
@@ -122,6 +122,17 @@ You can do it easly in this way:
   <Story name="filter" story={stories.Filter} />
 </Canvas>
 
+### Animation Disabled  
+
+It is possible to disable to animations applied to the `PopoverContainer` by setting the `disableAnimation` prop to `true`.
```
robinzigmond, line 127:
```suggestion
It is possible to disable the animations applied to the `PopoverContainer` by setting the `disableAnimation` prop to `true`.
```

**suggestion** (typo)

src/components/popover-container/popover-container.stories.tsx
```diff
@@ -314,3 +314,15 @@ export const Filter: ComponentStory<typeof PopoverContainer> = () => {
     </Box>
   );
 };
+
+export const DisableAnimation: ComponentStory<typeof PopoverContainer> = () => (
+  <div style={{ height: 100 }}>
+    <PopoverContainer
+      disableAnimation
```
robinzigmond, line 321: **suggestion** (non-blocking): I'd personally add a `title` to this story - most of the others have one - as this should remove the need to add the `containerAriaLabel` and `openButtonAriaLabel` to stop Axe flagging accessibility errors, as well as more generally show the best practices. Also because the story is only meant to demonstrate the `disableAnimation` prop so any other props are just distraction - I personally find a `title` prop less distracting than those 2 aria props. Not a big issue though if you disagree.

src/components/popover-container/popover-container.style.ts
```diff
@@ -32,29 +57,12 @@ const PopoverContainerContentStyle = styled.div<PopoverContainerContentStyleProp
...
-        return `
-        opacity: 0; 
-        transform: translateY(-8px);
-        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
-          `;
-      default:
-        return "opacity: 0";
-    }
-  }}
+  @media screen and (prefers-reduced-motion: no-preference) {
```
robinzigmond, line 60: **question**: is this media-query in the CSS actually necessary? It seems redundant to me, since we're already setting `disableAnimation` to `true` if the user prefers reduced motion (although as in the the JS file, shouldn't this query be for `reduce` rather than `no-preference`?). If so then on line 65 we're already setting `opacity: 1`.

(It also occurred to me that it's better to use CSS than JS given the choice, so maybe we could remove the media query from the JS file and just use this instead. But I don't think it will as then if `disableAnimation` is false then line 65 will end  up turning the animation on. It might work if you wrap the `animationToRender` call in the appropriate media query, but feel free to not do that as I'm not sure there's much gain.)

cypress/components/popover-container/popover-container.cy.tsx
DipperTheDan: I deleted these tests as they were flaky. They were relying on getting the DOMRect for the position and then asserted if that covered the button. Ideally, I could add these back in and skip them until we refactor to Playwright. As we'd be able to assert `.not.toBeInViewport` which is a lot neater than this in Cypress. I'll caveat this with we do also have a Chromatic snapshot of a story with the button covered, so it isn't like removing this test means we aren't covered for this case at all.

src/components/popover-container/popover-container-test.stories.tsx
DipperTheDan: Yeah, my thinking was we can have an example of it in a scrollable block so we can manually test for it in future. If you think I can improve the example, feel free to add suggestions and I'll get the changes made 😄

src/components/popover-container/popover-container.stories.mdx
DipperTheDan: Well spotted.

src/components/popover-container/popover-container.stories.tsx
DipperTheDan: I agree with you. A `title` is a lot better approach than this. Thanks for pointing it out. God knows what I was thinking here 😆

src/components/popover-container/popover-container.component.tsx
DipperTheDan: Good point and thanks for noticing this. I'll get it changed so that it's consistent with `Loader` and `LoaderBar`.

src/components/popover-container/popover-container.style.ts
DipperTheDan: I had thought originally that if I could go with a CSS only approach but figured I'd need a prop to allow devs to control if they want the animation to happen or not too. I'll just make the change [you've suggested here](https://github.com/Sage/carbon/pull/6486#discussion_r1440246342)

src/components/popover-container/popover-container-test.stories.tsx
DipperTheDan: The original issue was composed using the `GlobalHeader`. The original issue had a GIF and screenshot of a live product so I removed it. I'll send you the examples privately and that might give you more context.

src/components/popover-container/popover-container.stories.tsx
robinzigmond: looks like you've still left the `openButtonAriaLabel`, while getting rid of the other one. Not a big deal but my idea was that they should both be able to go away, leaving only a `title` prop and the `disableAnimation`.

robinzigmond (dismissed):

cypress/components/popover-container/popover-container.cy.tsx
edleeks87: Agree with the above, I'd prefer to have it `skip`ped rather than commented out though as even with the TODO it may be seen as something someone meant to remove etc

src/components/popover-container/popover-container-test.stories.tsx
```diff
@@ -97,3 +98,43 @@ export const PopoverContainerWithSelect = () => {
...
+              "aria-haspopup": ariaHasPopup,
+            }) => (
+              <Button
+                iconType="settings"
+                iconPosition="after"
+                data-element={dataElement}
+                aria-label={ariaLabel}
+                aria-haspopup={ariaHasPopup}
+                aria-expanded={ariaExpanded}
+                forwardRef={ref}
```
edleeks87, line 127:
```suggestion
                ref={ref}
```

forwardRef is marked for deprecation

src/components/popover-container/popover-container.style.ts
```diff
@@ -6,6 +6,30 @@ import { baseTheme } from "../../style/themes";
 import IconButton from "../icon-button";
 import StyledIcon from "../icon/icon.style";
 
+function animationToRender(animationState?: TransitionStatus) {
+  switch (animationState) {
+    case "entering":
```
edleeks87, line 11: comment (non-blocking): the indentation on this is off

src/components/popover-container/popover-container.style.ts
```diff
@@ -32,29 +57,8 @@ const PopoverContainerContentStyle = styled.div<PopoverContainerContentStyleProp
...
-        opacity: 0; 
-        transform: translateY(-8px);
-        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
-          `;
-      default:
-        return "opacity: 0";
-    }
-  }}
+  ${({ animationState, disableAnimation }) =>
+    disableAnimation ? `opacity: 1` : animationToRender(animationState)}
```
edleeks87, line 61: suggestion (non-blocking): you could remove the need for a ternary here and push it to the `animationToRender`

```
function animationToRender(animationState?: TransitionStatus, disableAnimation?: boolean) {
  switch (animationState) {
    case "entering":
      return `
        opacity: 0;
        transform: translateY(-8px);
      `;
    case "entered":
      return `
        opacity: 1; 
        transform: translateY(0);
        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
      `;
    case "exiting":
      return `
        opacity: 0; 
        transform: translateY(-8px);
        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
      `;
    default:
      return `opacity: ${disableAnimation ? 1 : 0}`;
  }
}
...
  ${({ animationState, disableAnimation }) => animationToRender(animationState, disableAnimation)}

**Review by edleeks87 — DISMISSED** [review]

**Review by robinzigmond — APPROVED** [review]

File: src/components/popover-container/popover-container.style.ts [hunk 11]
```diff
@@ -6,6 +6,38 @@ import { baseTheme } from "../../style/themes";
...
 import StyledIcon from "../icon/icon.style";
 
+function animationToRender({
+  animationState,
+  disableAnimation,
+}: {
+  animationState?: TransitionStatus;
+  disableAnimation?: boolean;
+}) {
+  if (disableAnimation) return "opacity: 1";
```
robinzigmond, line 16:
```suggestion
  if (disableAnimation) return "opacity: 1;";
```

**suggestion**: I don't know if it matters here, but I'm pretty sure we've had issues in the past with forgetting the trailing `;` in lines of CSS in styled-components. Seems safer to include it to me.

src/components/popover-container/popover-container.style.ts
```diff
@@ -6,6 +6,38 @@ import { baseTheme } from "../../style/themes";
...
+        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
+      `;
+    case "exiting":
+      return `
+        opacity: 0; 
+        transform: translateY(-8px);
+        transition: all 0.3s cubic-bezier(0.25, 0.25, 0, 1.5);
+      `;
+    default:
+      return "opacity: 0";
```
robinzigmond, line 37:
```suggestion
      return "opacity: 0;";
```

**suggestion**: as above

src/components/popover-container/popover-container.component.tsx
```diff
@@ -264,6 +272,7 @@ export const PopoverContainer = ({
         {(state: TransitionStatus) =>
           isOpen && (
             <Popover
+              popoverStrategy={disableAnimation ? "fixed" : "absolute"}
```
robinzigmond, line 275: **comment** (non-blocking): I'm not sure if this matters or not (it seems OK to me in manual testing), but I'm uneasy about the fact that if the user has the Reduced Motion preference on then the animation gets disabled no matter what the value of the `disableAnimation` prop. This feels wrong if the intention was to change the strategy to `fixed` when, and only when, animation is disabled. (And if that wasn't the intention, what was it?)

I'm not sure you have to change anything but it would feel more consistent to me to either change this so that it checks `reduceMotion` as well, or just permanently set `popoverStrategy="fixed"`, since it doesn't seem that there are any scenarios where anything goes wrong with that?

DipperTheDan: I feel you're correct about the need to check `reduceMotion` in this too. I'll get it added. I'd feel more comfortable with this way for now rather than outright changing it be `fixed`. I'd rather do that in separate piece of work that has DS/UX and Accessibility involved. Even if we implement that in the future, the `disableAnimation` prop will still be available as that bypasses the transition CSS stuff.

robinzigmond (approved):

stephenogorman (approved):

PR #6494, opened 2023-12-21
test(pages): convert tests to playwright - FE-6142

DipperTheDan:
### Proposed behaviour

Refactor tests to Playwright

### Current behaviour

Tests currently use Cypress

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,194 @@
...
+          await expect(getDataElementByValue(page, "title")).toHaveText(
+            "My Second Page"
+          );
+        });
+      });
+    });
+  });
+
+  // TODO: Playwright is not returning the full class attribute, so this test is not working as expected.
+  (["slide", "fade"] as PagesProps["transition"][]).forEach((transition) => {
```
robinzigmond, line 50: **comment** - we shouldn't skip tests unless there's something making it fail that defies easy explanation.

In this case I think the issue is that we're not actually clicking the "next page" button, so the browser never actually moves on to the next page and so no animations or related CSS classes happen.

src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,194 @@
...
+  waitForAnimationEnd,
+} from "../../../playwright/support/helper";
+import {
+  PageComponent,
+  PageComponentWithTitle,
+  PagesComponent,
+} from "./components.test-pw";
+
+test.describe("Prop checks for Pages component", () => {
+  ["initialPageIndex", "pageIndex"].forEach((propName) => {
```
robinzigmond, line 25: **suggestion** (non-blocking): I'm not sure of what the "right" thing to do is, and I know this structure was just copied from the Cypress tests - but personally I'm not a fan of the nested `describe`s, with each being in a `forEach` loop. Since there are only 2 prop names being tested I think I'd actually prefer to just repeat the inner block twice, once for each of the 2 prop names we want to test.

But it's not something I'm going to hold this up over - just a thought.

src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,194 @@
...
+      mount,
+      page,
+    }) => {
+      await mount(<PagesComponent transition={transition} />);
+
+      await dataComponentButtonByText(page, "Open Preview").click();
+      const pageComponent = getDataElementByValue(page, "visible-page");
+      await waitForAnimationEnd(pageComponent);
+
+      const classAttribute = pageComponent.getAttribute("class");
```
robinzigmond, line 61:
```suggestion
      const classAttribute = await pageComponent.getAttribute("class");
```

As well as other changes as noted above, we need to `await` this otherwise the test fails because `classAttribute` is an object.

src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,194 @@
...
+    await backArrow(page).click();
+
+    await expect(getDataElementByValue(page, "title").nth(0)).toHaveText(
+      "My First Page"
+    );
+  });
+
+  [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS].forEach(
+    ([testTitle]) => {
+      test(`Component should render as expected with title prop set to ${testTitle}`, async ({
```
robinzigmond, line 124:
```suggestion
      test(`Page Component should render as expected with title prop set to ${testTitle}`, async ({
```

**suggestion** (non-blocking): while I don't think this needs to be in its own `describe` block as it was in the Cypress tests, I think it's maybe worth highlighting that this test is specifically about the Page (sub)component

src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,194 @@
...
+        await waitForAnimationEnd(pageComponent);
+
+        await expect(pageComponent).toHaveText(`${testTitle}`);
+      });
+    }
+  );
+});
+
+test.describe("Accessibility tests for Pages component", () => {
+  ["initialPageIndex", "pageIndex"].forEach((propName) => {
```
robinzigmond, line 140: **comment** (non-blocking): same as above re the nested `describe`'s here.

robinzigmond (approved):

src/components/pages/pages.pw.tsx
```diff
@@ -0,0 +1,218 @@
...
+    }
+  );
+});
+
+test.describe("Accessibility tests for Pages component", () => {
+  [
+    ["number", 1],
+    ["string", "1"],
+  ].forEach(([type, propValue]) => {
+    test(`When initialPageIndex prop is passed as ${type}, should show the title that matches the one on the page with that index`, async ({
```
ZhuoyuJin, line 157: Can we change the test description to include "check accessibility" which is the main purpose for the test. 
Same as below test

ZhuoyuJin (approved):

PR #6497, opened 2023-12-22
test(sidebar): convert tests to playwright - FE-6165

DipperTheDan:
### Proposed behaviour

Refactor tests from Cypress to Playwright.

### Current behaviour

Tests currently use Cypress.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


playwright/components/sidebar/index.ts
```diff
@@ -0,0 +1,7 @@
+import type { Page } from "@playwright/test";
```
ZhuoyuJin, line 1:
```suggestion
import { Page } from "@playwright/test";
```

src/components/sidebar/components.test-pw.tsx
```diff
@@ -0,0 +1,216 @@
...
+        topModalOverride
+      >
+        foo
+      </Sidebar>
+      <Dialog open={isOpen2} onCancel={() => setIsOpen2(false)} title="Dialog">
+        foo
+      </Dialog>
+    </>
+  );
+};
```
ZhuoyuJin, line 216: We can then delete them from `sidebar-test.stories.tsx`, since these mock examples are moved to `components.test-pw.tsx`.

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,441 @@
...
+  SidebarComponentWithHeading,
+  SidebarComponentWithOnCancel,
+  TopModalOverride,
+} from "./components.test-pw";
+import { SidebarProps } from "./sidebar.component";
+import { SIDEBAR_SIZES, SIDEBAR_SIZES_CSS } from "./sidebar.config";
+
+test.describe("Prop tests for Sidebar component", () => {
+  ([true, false] as const).forEach((boolVal) => {
+    test(`check enableBackgroundUI prop set to ${boolVal}`, async ({
```
ZhuoyuJin, line 36:
```suggestion
    test(`verify visibility of backgroundUILocatorElement when enableBackgroundUI prop is set to ${boolVal}`, async ({
```

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,441 @@
...
+        await expect(backgroundUILocatorElement).toBeVisible();
+      }
+    });
+  });
+
+  ([
+    ["left", 0, 852],
+    ["right", 852, 0],
+  ] as [SidebarProps["position"], number, number][]).forEach(
+    ([boolVal, left, right]) => {
```
ZhuoyuJin, line 56:
```suggestion
    ([position, left, right]) => {
``` just the variable name "boolVal" sounds like it should be either true/false

File: src/components/sidebar/sidebar.pw.tsx [hunk 5]
```diff
@@ -0,0 +1,441 @@
...
+  test("check component has expected data-role", async ({ mount, page }) => {
+    await mount(<SidebarComponent data-role={CHARACTERS.STANDARD} />);
+
+    await expect(sidebarComponent(page)).toHaveAttribute(
+      "data-role",
+      CHARACTERS.STANDARD
+    );
+  });
+
+  test("pressing the Escape key closes the component when disableEscKey is false", async ({
```
ZhuoyuJin, line 170: For this test and the one below, can we have something like: 
```
([true, false] as const).forEach((boolVal) => {
    test(`verify visibility of sidebar component when disableEscKey is ${boolVal} and Escape key is pressed`, async ({ ....
```

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,441 @@
...
+  test("tabbing forward through the sidebar and other focusable containers back to the start should not make the background scroll to the bottom", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<SidebarBackgroundScrollWithOtherFocusableContainers />);
+
+    await continuePressingTAB(page, 6);
+    const closeIconButtonElement = closeIconButton(page).nth(0);
+
+    await expect(closeIconButtonElement).toBeFocused();
```
ZhuoyuJin, line 421: **question**: strange for the below GithubActions failure, doesn't seem like there is a value "not focused", any idea?

DipperTheDan: No idea on this as all the jobs have passed 🤷🏻‍♂️ 
How odd. I'll rebase with master and push back up just try to run it with a fresh commit.

DipperTheDan: Fixed this btw. It needed the `waitForAnimationEnd()`as it was flaking.

ZhuoyuJin (approved):

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+  test("check component has expected data-role", async ({ mount, page }) => {
+    await mount(<SidebarComponent data-role={CHARACTERS.STANDARD} />);
+
+    await expect(sidebarComponent(page)).toHaveAttribute(
+      "data-role",
+      CHARACTERS.STANDARD
+    );
+  });
+
+  ([true, false] as const).forEach((boolVal) => {
```
robinzigmond, line 170: **suggestion** (non-blocking): as above, I'd prefer to call this boolean something like `disableEscKey`, rather than the non-meaningful `boolVal`

robinzigmond, line 170:
```suggestion
  ([true, false]).forEach((boolVal) => {
```

**suggestion**: as above

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+      if (boolVal) {
+        await expect(sidebarPreviewElement).toBeInViewport();
+      } else {
+        await expect(sidebarPreviewElement).not.toBeInViewport();
+      }
+    });
+  });
+
+  ([true, false] as const).forEach((boolVal) => {
```
robinzigmond, line 189:
```suggestion
  ([true, false]).forEach((boolVal) => {
```

**suggestion**: as above

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+    test(`check component with open prop set to ${boolVal}`, async ({
```
robinzigmond, line 190: **suggestion** (non-blocking): as above, I'd prefer to call this something like `open` or maybe `isOpen`, or similar - rather than `boolVal`

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+    const toastElement = getComponent(page, "toast");
+
+    await expect(toastElement).not.toBeVisible();
+
+    const openToastElement = getDataElementByValue(page, "open-toast");
+    await openToastElement.click();
+
+    await expect(toastElement).toBeVisible();
+
+    const toastElementChildrenChildren = page
```
robinzigmond, line 221: **suggestion**: lol at `toastElementChildrenChildren` as a const name - I assume that's the nonsense output by the automated conversion script I wrote 😅 But I'm sure you can come up with a better name (`toastCloseButton` maybe?)

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+  test("should check accessibility for SidebarComponent example", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<SidebarComponent />);
+
+    await checkAccessibility(page);
+  });
+
+  (["left", "right"] as SidebarProps["position"][]).forEach((boolVal) => {
```
robinzigmond, line 335: **comment**: quite apart from my usual complainat about `boolVal` as a variable name, this isn't even boolean! Let's just call it `position` or something similar.

src/components/sidebar/sidebar.pw.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+  ([
+    SIDEBAR_SIZES[0],
+    SIDEBAR_SIZES[1],
+    SIDEBAR_SIZES[2],
+    SIDEBAR_SIZES[3],
+    SIDEBAR_SIZES[4],
+    SIDEBAR_SIZES[5],
+    SIDEBAR_SIZES[6],
+  ] as SidebarProps["size"][]).forEach((size) => {
```
robinzigmond, line 363: **suggestion** (non-blocking): similar to my earlier comment where we used the `SIDEBAR_SIZES` array, but even simpler here - this can be simplified a lot. This array is just a very longhand way of writing `SIDEBAR_SIZES`!

src/components/sidebar/components.test-pw.tsx
```diff
@@ -0,0 +1,216 @@
...
+      </Sidebar>
+    </>
+  );
+};
+
+export const SidebarBackgroundScrollTestComponent = () => {
+  return (
+    <Box height="2000px" position="relative">
+      <Box
+        data-element="test-box"
```
robinzigmond, line 108: **suggestion** (non-blocking): I like the idea of adding a `data-element` to identify these boxes in the test 👍 But if we're adding this I think you should remove the `id="bottom-box"` because unless I'm forgetting something, the only reason I added that ID was to locate it in the Cypress test, and we're now locating it by data-element instead. The examples would be cleaner without an un-needed ID.

src/components/sidebar/components.test-pw.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+export const SidebarBackgroundScrollWithOtherFocusableContainers = () => {
+  const toast1Ref = useRef(null);
+  const toast2Ref = useRef(null);
+  return (
+    <Box height="2000px" position="relative">
+      <Box
+        data-element="test-box"
+        height="100px"
+        id="bottom-box"
```
robinzigmond, line 131: **suggestion** (non-blocking): as above, the `id` can be removed from this example

robinzigmond (approved):

ZhuoyuJin (approved):

PR #6500, opened 2024-01-02
docs: replace Formik/Yup example and add explanatory notes

robinzigmond:
### Proposed behaviour

The docs on Validation should link to a working example of Carbon with Formik and Yup validation, and clearly explain any important points.

### Current behaviour

The linked sandbox crashes when updating almost any of the component values. No explanation is given of the CSB example despite some of the code being quite complex.

### Checklist

#### QA

### Testing instructions

Check the documentation added in the `validation.stories.mdx` file - suggest any changes necessary to make it more readable and understandable, or any key points missed.

Also carefully inspect/test the codesandbox to ensure the validation of each component behaves as desired and that there are no other bugs.


docs/validations.stories.mdx
```diff
@@ -203,4 +203,26 @@ Formik supports three kinds of validation:
...
+### Form validation using Field-level validation with Formik and Yup
+
+See the [example codesandbox code](https://codesandbox.io/s/carbon-with-formik-new-example-2cqdxm?file=/src/App.tsx) showing how to
+integrate various different Carbon components into Formik/Yup.
+
+Note the following points about the implementation:
+- Carbon components are wrapped in [Formik's Field component](https://formik.org/docs/api/field), passing the actual Carbon component as the `as` prop.
+- a `validate` prop is also required as well as warning/error/info props as appropriate to pass to the Carbon component. In the example these have been
+returned from a custom Hook that takes the Yup validation schemas - this should be adapted to your actual use case if not appropriate.
+- note the use of an explicit `checked` prop on the checkbox and switch components, with `value` set to a hardcoded string. These is required for these
```
edleeks87, line 215:
```suggestion
- note the use of an explicit `checked` prop on the checkbox and switch components, with `value` set to a hardcoded string. This is required for these
```

docs/validations.stories.mdx
```diff
@@ -203,4 +203,26 @@ Formik supports three kinds of validation:
...
+checkbox-based components to work properly (Formik `Field` by default will try to pass `value` but `checked` is the important property on these components,
+with `value` having no purpose other than allowing the field to be identified on the server side after form submission)
+- date-related components require more careful handling - an explicit `onChange` is required that calls Formik's `setFieldValue` function with the
+`formattedValue` property of the custom event object's `target`
+- there is also a custom Yup transform to ensure the date is parsed in `dd/MM/yyyy` format, which is the default for Carbon date components but not accepted
+by Yup. The implementation in the sandbox uses the `date-fns` library but you can use whichever date-parsing library you choose, and of course use whichever
+date-format is appropriate to the locale you're using
+- `DateRange` requires particularly special handling, as this is a single Carbon component which renders 2 separate HTML input fields. In the example
+this is done by using a modified form of the custom hook mentioned above that returns two separate props objects, one for each input, and passes them
```
edleeks87, line 224:
```suggestion
this is done by using a modified form of the custom hook mentioned above that returns two separate prop objects, one for each input, and passes them
```

non-blocking as it's probably fine the way it's currently written in this context

edleeks87 (commented):
Really good work to get this updated @robinzigmond given how it's been neglected for a while now. I noticed that the validation doesn't set an error on both DateRange inputs when you follow the instructions about setting a start after the end etc


image

docs/validations.stories.mdx
robinzigmond: meh, I personally think "props objects" reads better (there are 2 separate objects, both of which contain props), but I don't hate your suggestion. I can probably word it better anyway to avoid the debate - I'll have a think.

robinzigmond:
> Really good work to get this updated @robinzigmond given how it's been neglected for a while now. I noticed that the validation doesn't set an error on both DateRange inputs when you follow the instructions about setting a start after the end etc
> 
> image

yeah you're right, I'd forgotten about this but it's a minor issue imo, and importantly one I remember deciding I couldn't really do anything about, given the difficulties/contraints around getting DateRange to work with Yup/Formik. I'm probably missing something though - I'll take another look at this as I did this before Christmas and I should get my head back in it.

DipperTheDan (commented):
Looks OK and I have nothing further to add. I did notice that the `Grouped Inputs` components have a bug present because the two share the same `id` and `value` strings. Nothing you've done as it was here before this PR. Should it be addressed in this PR or do it in a different piece of work?

https://github.com/Sage/carbon/assets/56251247/e404faf5-bcfe-45ea-9cac-0d8c91844b48

robinzigmond:
> Looks OK and I have nothing further to add. I did notice that the `Grouped Inputs` components have a bug present because the two share the same `id` and `value` strings. Nothing you've done as it was here before this PR. Should it be addressed in this PR or do it in a different piece of work?
> 
>  Screen.Recording.2024-01-04.at.13.54.29.mov

oops 🤦‍♂️ 

Thanks for reviewing and great spot - I'll get this fixed

edleeks87 (approved):

DipperTheDan (approved):

PR #6506, opened 2024-01-05
feat(accordion): add subtle variant and update default styles

nuria1110:
### Proposed behaviour

- Introduced `variant` prop with "standard" and "subtle" options, default accordion is now standard variant. 
- "chevron_down_thick" added to `iconType` options.
- Update standard accordion styles to align with DS. 
- `scheme` and `buttonHeading` have been deprecated. 

Subtle variant:
- `iconAlign` is set to "left" by default
- `iconType` is set to "chevron_down_thick" by default
- `borders` is set to "none"
- `subTitle` will not be rendered
- Does not support validation

Screenshot 2024-01-05 at 16 49 28

Screenshot 2024-01-05 at 16 35 56

### Current behaviour

- Accordion component does not have "subtle" variant. 
- Default styles do not align with DS. 

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/accordion/accordion-test.stories.tsx
+++ b/src/components/accordion/accordion-test.stories.tsx
@@ -25,7 +25,7 @@ export default {
       },
     },
     iconType: {
-      options: ["chevron_down", "dropdown"],
+      options: ["chevron_down", "chevron_down_thick", "dropdown"],
       control: {
         type: "select",
       },
@@ -48,6 +48,12 @@ export default {
         type: "select",
       },
     },
+    variant: {
+      options: ["standard", "subtle"],
+      control: {
+        type: "select",
+      },
+    },
   },
 };
 
@@ -55,7 +61,6 @@ export const Default = ({ ...args }) => (
   <Accordion
     onChange={action("expansionToggled")}
     {...{
-      iconAlign: "right",
       customPadding: 0,
       title: "Title",
       subTitle: "Sub Title",
@@ -71,19 +76,31 @@ export const Default = ({ ...args }) => (
 
 Default.storyName = "default";
 
-export const Grouped = () => (
+export const Grouped = ({ ...args }) => (
   <AccordionGroup>
-    <Accordion title="First Accordion" onChange={action("expansionToggled")}>
+    <Accordion
+      title="First Accordion"
+      onChange={action("expansionToggled")}
+      {...args}
+    >
       <Box p={2}>
         <Textbox label="Textbox in an Accordion" />
       </Box>
     </Accordion>
-    <Accordion title="Second Accordion" onChange={action("expansionToggled")}>
+    <Accordion
+      title="Second Accordion"
+      onChange={action("expansionToggled")}
+      {...args}
+    >
       <Box p={2}>
         <Textbox label="Textbox in an Accordion" />
       </Box>
     </Accordion>
-    <Accordion title="Third Accordion" onChange={action("expansionToggled")}>
+    <Accordion
+      title="Third Accordion"
+      onChange={action("expansionToggled")}
+      {...args}
+    >
       <Box p={2}>
         <Box mt={2}>Content</Box>
         <Box>Content</Box>
--- a/src/components/accordion/accordion.pw.tsx
+++ b/src/components/accordion/accordion.pw.tsx
@@ -121,34 +121,37 @@ test.describe("should render Accordion component", () => {
     await expect(accordionContent(page)).toBeVisible();
   });
 
-  (["chevron_down", "dropdown"] as const).forEach((iconType) => {
-    test(`should set iconType to ${iconType} when Accordion row is closed`, async ({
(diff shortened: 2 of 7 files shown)


src/components/accordion/accordion.pw.tsx
```diff
@@ -733,4 +748,13 @@ test.describe("Accessibility tests for Accordion", () => {
 
     await checkAccessibility(page);
   });
+
+  test("should pass accessibility tests for AccordionSubtle", async ({
```
DipperTheDan, line 752: **suggestion(non-blocking): ** I feel something like this is a better description for the test.

```suggestion
  test("should pass accessibility tests for Accordion when variant is subtle", async ({
```

src/components/accordion/accordion.pw.tsx
```diff
@@ -121,34 +121,37 @@ test.describe("should render Accordion component", () => {
...
+    (iconType) => {
+      test(`should set iconType to ${iconType} and transform when Accordion row is closed`, async ({
+        page,
+        mount,
+      }) => {
+        await mount(<AccordionComponent iconType={iconType} />);
+
+        await expect(accordionIcon(page)).toHaveAttribute("type", iconType);
+        await expect(accordionIcon(page)).toBeVisible();
+        const transformValue = await getStyle(accordionIcon(page), "transform");
```
DipperTheDan, line 134: suggestion(non-blocking): I don't _think_ you need the `await` here. 

```suggestion
        const transformValue = getStyle(accordionIcon(page), "transform");
```

src/components/accordion/accordion.pw.tsx
```diff
@@ -121,34 +121,37 @@ test.describe("should render Accordion component", () => {
...
+    (iconType) => {
+      test(`should set iconType to ${iconType} and transform when Accordion row is open`, async ({
+        page,
+        mount,
+      }) => {
+        await mount(<AccordionComponent iconType={iconType} expanded />);
+
+        await expect(accordionIcon(page)).toHaveAttribute("type", iconType);
+        await expect(accordionIcon(page)).toBeVisible();
+        const transformValue = await getStyle(accordionIcon(page), "transform");
```
DipperTheDan, line 150: suggestion(non-blocking): Same again. I don't _think_ you need the `await` here. Try removing it, if it breaks then I am obviously wrong 😄 

```suggestion
        const transformValue = getStyle(accordionIcon(page), "transform");
```

src/components/accordion/accordion.style.ts
```diff
@@ -165,7 +182,26 @@ const StyledAccordionTitleContainer = styled.div<StyledAccordionTitleContainerPr
...
 
+    ${variant === "subtle" &&
+    css`
+      color: var(--colorsActionMajor500);
+      padding: var(--spacing025);
+      margin-bottom: ${isExpanded && "var(--spacing200)"};
+      ${StyledAccordionIcon} {
+        color: var(--colorsActionMajor500);
+        ${iconAlign === "left" && "margin-right: var(--spacing050)"};
+      }
```
DipperTheDan, line 193: **suggestion(non-blocking):** very minor formatting change here just to keep things consistent. 
```suggestion
      margin-bottom: ${isExpanded && "var(--spacing200)"};
      
      ${StyledAccordionIcon} {
        color: var(--colorsActionMajor500);
        ${iconAlign === "left" && "margin-right: var(--spacing050)"};
      }
```

DipperTheDan (commented):
Good work @nuria1110. Just a couple of questions:

1. Is the expected behaviour that if someone were to pass `variant="standard"` and `scheme="transparent"` that the `variant` prop takes precedence? If so, I think we should document or warn consumers of that in someway if they do accidentally use both. 
2. You may need to add the `variant` prop to the `argTypes` in the Accordion test story. Double check with QA first on that though.

src/components/accordion/accordion.pw.tsx
nuria1110: I'm afraid it is in fact needed

src/components/accordion/accordion.pw.tsx
nuria1110: answered above

DipperTheDan (dismissed):

Parsium (dismissed):

src/components/accordion/accordion-test.stories.tsx
```diff
@@ -71,19 +76,31 @@ export const Default = ({ ...args }) => (
 
 Default.storyName = "default";
 
-export const Grouped = () => (
+export const Grouped = ({ ...args }) => (
```
Parsium, line 79: **praise**: nice spot that the `args` weren't being passed properly 👍🏼

Parsium (approved):

divyajindel (approved):

DipperTheDan (approved):

PR #6510, opened 2024-01-08
fix(flat-table): ensure that pressing arrow and space keys scroll table when it is scrollable FE-6273

edleeks87:
### Proposed behaviour

Ensures that arrow and space keys scroll `FlatTable` that has overflow set. The tab stop has been
moved to container further in the HTML tree to ensure that the scroll behaviour works as expected.

### Current behaviour

Scrollable tables are not scrolled when arrow or space key is pressed and table wrapper is focused.

### Checklist

#### QA

### Testing instructions

`flat-table--with-sticky-head-row-span-and-colspan` story should scroll the table when arrow or space key pressed
Confirm the other stories behave as expected (eg the expandable or clickable row stories still work)


src/components/flat-table/flat-table.spec.tsx
```diff
@@ -724,9 +726,45 @@ describe("FlatTable", () => {
             </FlatTableBody>
           </FlatTable>
         );
-        const tableWrapper = await screen.findByRole("region");
+        const tableWrapper = container.querySelectorAll(
+          "div"
+        )[1] as HTMLElement; // screen.findByRole("region");
```
nineteen88, line 731: question: is this comment actually needed?

src/components/flat-table/flat-table.spec.tsx
```diff
@@ -724,9 +726,45 @@ describe("FlatTable", () => {
...
+                  <FlatTableCell>three</FlatTableCell>
+                  <FlatTableCell>four</FlatTableCell>
+                </FlatTableRow>
+              </FlatTableBody>
+            </FlatTable>
+          </CarbonProvider>
+        );
+        const tableWrapper = container.querySelectorAll(
+          "div"
+        )[2] as HTMLElement; // screen.findByRole("region");
```
nineteen88, line 762: question: same as before

But because there is two I'm probably just being silly and it's good to have?

nineteen88 (dismissed):

tomdavies73 (dismissed):

src/components/flat-table/flat-table.spec.tsx
edleeks87: good spot I'll get these removed

tomdavies73 (approved):

nineteen88 (approved):

stephenogorman (approved):

PR #6515, opened 2024-01-11
feat(inline-inputs): add labelAlign prop

### LINKED ISSUE #6498: Add labelAlign prop to the Inline Inputs component
### Desired behaviour

`labelAlign` prop to the Inline `Inputs component` to be able to provide the same functionality as to align the Label which the textbox component has. 

### Current behaviour

No `labelAlign` prop  present. 

### Suggested Solution

Add `labelAlign` prop. 

_No response_

_No response_

### Confidentiality

nineteen88:
### Proposed behaviour

Add a labelAlign prop that will take "left" or "right" params. 
Should replicate the behaviour of the inline label labelAlign for the Textarea component.

fix #6498

### Current behaviour

No such prop exists.

### Checklist

#### QA

### DIFF
--- a/src/components/inline-inputs/inline-inputs.spec.tsx
+++ b/src/components/inline-inputs/inline-inputs.spec.tsx
@@ -203,6 +203,21 @@ describe("Inline Inputs", () => {
     });
   });
 
+  describe("when providing the labelAlign prop", () => {
+    beforeEach(() => {
+      wrapper = render({ label: "foo", labelAlign: "left", labelWidth: 30 });
+    });
+
+    it("sets the inline label to the provided alignemnt", () => {
+      assertStyleMatch(
+        {
+          justifyContent: "flex-start",
+        },
+        wrapper.find(Label)
+      );
+    });
+  });
+
   describe("when the inputWidth prop is not passed in", () => {
     beforeEach(() => {
       wrapper = render({});
--- a/src/components/inline-inputs/inline-inputs.component.tsx
+++ b/src/components/inline-inputs/inline-inputs.component.tsx
@@ -35,6 +35,8 @@ export interface InlineInputsProps
   htmlFor?: string;
   /** Defines the label text for the heading. */
   label?: string;
+  /** Inline label alignment */
+  labelAlign?: "left" | "right";
   /**
    * Custom label id, could be used in combination with aria-labelledby prop of each input,
    * to make them accesible for screen readers.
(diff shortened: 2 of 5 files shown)


src/components/inline-inputs/inline-inputs.spec.tsx
```diff
@@ -203,6 +203,21 @@ describe("Inline Inputs", () => {
     });
   });
 
+  describe("when providing the labelAlign prop", () => {
+    beforeEach(() => {
+      wrapper = render({ label: "foo", labelAlign: "left", labelWidth: 30 });
+    });
+
+    it("sets the inline label to the provided alignemnt", () => {
```
tomdavies73, line 211: **question**: could we do an` it.each` here to check `labelAlign` being `"left"`/`"right"` and the css value of justify-content being `"flex-start"`/`"flex-end"`, just like its done in `inline-inputs.pw.tsx`?

tomdavies73 (commented):
Apart from that comment, looks good to me! Also, just worth noting, you've mentioned the relevant fix in the PR, but can't seem to see it in the feat commit. Might be worth making sure It's in there too.

src/components/inline-inputs/inline-inputs.spec.tsx
nineteen88: I could do but I personally just wanted to check that the prop is passing through as expected for unit test coverage and then the css checks are reserved for playwright. If it's felt that we should be css checking here then I'll make that change

tomdavies73: Yeah, that's fair, I think the playwright check is enough actually in retrospect

DipperTheDan: Traditionally we'd have also tested the CSS in the unit tests but I think there was a discussion around not doing this when we move to RTL? I'm OK doing it the way you have and you seem to have full coverage too.

DipperTheDan (commented):
Everything looks OK to me. I just have a question regarding the new `labelAlign` prop used in conjunction with the `adaptiveLabelBreakpoint` prop. It seems that the label is always on the left. Is this to be expected? 

Screenshot 2024-01-16 at 09 21 17

harpalsingh:
> Details

Once it is responsive and stacked above the text input the left/right align wouldnt be needed and left aligned as what it does now makes sense.

tomdavies73 (approved):

divyajindel (approved):

PR #6524, opened 2024-01-16
fix(flat-table): ensure that focus size is correctly applied in Safari - FE-6315

### LINKED ISSUE #6496: Incorrect focus on element inside flat table row with modified height.
### Current behaviour

Focus has incorrect height in the table row (just Safari browser):
image

### Expected behaviour

Focus should have same height like the row.

https://codesandbox.io/p/sandbox/safari-focus-bug-qzf5rl

### JIRA Ticket (Sage Only)

https://jira.sage.com/browse/SBS-70188

### Suggested Solution

_No response_

### Carbon Version

124.4.1

### Design Tokens Version

_No response_

Safari

MacOS

_No response_

### Confidentiality

DipperTheDan:
This fix uses the offset height of the flat-table-row element to correctly apply the correct height when the row is focused. This is fix is only for the newer focus styling and the old style will continue to work as it did previously.

fixes #6496

### Proposed behaviour

https://github.com/Sage/carbon/assets/56251247/8140b568-257d-4301-83d7-3d4eff7cc86e

### Current behaviour

https://github.com/Sage/carbon/assets/56251247/7a0e677e-b6ac-4f99-8ba2-943ba308528a

### Checklist

#### QA

### Additional context

This piece of work is purely to address the static styling of Flat Table Row in Safari. A further piece of work to address any dynamic styles will be addressed in FE-6334.

### Testing instructions

- Navigate to the new `Flat Table Size Focus` test story in Safari and ensure that the focus border is the correct size for the new focus styling. 
- There should be no other functional or styling regressions with regards to focus behaviour in any browser with this component.

### DIFF
--- a/src/components/flat-table/flat-table-row/flat-table-row.component.tsx
+++ b/src/components/flat-table/flat-table-row/flat-table-row.component.tsx
@@ -293,6 +293,7 @@ export const FlatTableRow = React.forwardRef<
         id={internalId.current}
         data-selected={selected && expandableArea === "wholeRow"}
         data-highlighted={highlighted && expandableArea === "wholeRow"}
+        rowHeight={rowRef?.current?.offsetHeight}
         {...interactiveRowProps}
         {...rest}
       >
--- a/src/components/flat-table/flat-table-row/flat-table-row.style.ts
+++ b/src/components/flat-table/flat-table-row/flat-table-row.style.ts
@@ -11,7 +11,6 @@ import { FlatTableProps } from "..";
 import { FlatTableRowProps } from "./flat-table-row.component";
 import addFocusStyling from "../../../style/utils/add-focus-styling";
 import { isSafari } from "../../../__internal__/utils/helpers/browser-type-check";
-import cellSizes from "../cell-sizes.style";
 
 const horizontalBorderSizes = {
   medium: "2px",
@@ -112,13 +111,6 @@ const verticalBorderColor = (colorTheme: FlatTableProps["colorTheme"]) => {
   }
 };
 
-const getFocusHeight = (size: StyledFlatTableRowProps["size"]) => {
-  if (!size) return "40px";
-  const { height } = cellSizes[size];
-
-  return height;
-};
-
 interface StyledFlatTableRowProps
   extends Pick<
     FlatTableRowProps,
@@ -144,6 +136,7 @@ interface StyledFlatTableRowProps
   isSubRow?: boolean;
   isFirstSubRow?: boolean;
   stickyOffset?: number;
+  rowHeight?: number;
 }
 
 const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
@@ -170,6 +163,7 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
     theme,
     isDragging,
     draggable,
+    rowHeight,
   }) => {
     const backgroundColor = bgColor ? toColor(theme, bgColor) : undefined;
     const customBorderColor = horizontalBorderColor
@@ -293,6 +287,7 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
               top: 0px;
             }
           }
+
           /* Styling for safari. Position relative does not work on tr elements on Safari  */
           ${isSafari(navigator) &&
           css`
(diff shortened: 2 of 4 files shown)


src/components/flat-table/flat-table-row/flat-table-row.component.tsx
```diff
@@ -105,8 +105,14 @@ export const FlatTableRow = React.forwardRef<
     const [firstCellId, setFirstCellId] = useState<string | null>(null);
     const [cellsArray, setCellsArray] = useState<Element[]>([]);
     const [tabIndex, setTabIndex] = useState(-1);
+    const [rowHeight, setRowHeight] = useState(0);
     let interactiveRowProps = {};
 
+    useLayoutEffect(() => {
```
robinzigmond, line 111: **comment** (non-blocking): I'm not a big fan of using state to update the height, and especially not when that is only updated in an effect that runs only once when the component is first rendered.

I don't think there's a perfect solution here, because in order to catch any changes in height (eg due to the user resizing the window, when the cell has enough text to run to several lines), we'd ideally need to use a `ResizeObserver`, which is rather more complex than I'd like to do. (And this behaviour never worked correctly anyway, it's certainly not a priority.)

My suggestion of using the `ref` directly for the `rowHeight` prop (see suggestion below) has a slight advantage over your solution in at least (potentially) updating the height on every render, so if this is in a table that dynamically changes quite a bit then there's more chance of the focus styles being adjusted to the correct height where things change. On the other hand, there might be performance implications (I doubt it, reading the `offsetHeight` shouldn't be expensive, but I can't guarantee that it's never an issue), so I'm not suggesting which way is better.

src/components/flat-table/flat-table-row/flat-table-row.component.tsx
```diff
@@ -293,6 +299,7 @@ export const FlatTableRow = React.forwardRef<
         id={internalId.current}
         data-selected={selected && expandableArea === "wholeRow"}
         data-highlighted={highlighted && expandableArea === "wholeRow"}
+        rowHeight={rowHeight}
```
robinzigmond, line 302:
```suggestion
        rowHeight={rowRef?.current?.offsetHeight}
```

**suggestion** (non-blocking): as talked about in comment above, this is an alternative to using state

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -11,7 +11,6 @@ import { FlatTableProps } from "..";
 import { FlatTableRowProps } from "./flat-table-row.component";
 import addFocusStyling from "../../../style/utils/add-focus-styling";
 import { isSafari } from "../../../__internal__/utils/helpers/browser-type-check";
-import cellSizes from "../cell-sizes.style";
```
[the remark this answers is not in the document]
DipperTheDan: My approach with this piece of work was to touch as little as possible as we've had situations with Flat Table where a simple change has been made and it's introduced more bugs than it fixed. I wanted to avoid that 😆  
Like you, I realised it was used somewhere else so it was left as is.

src/components/flat-table/flat-table-row/flat-table-row.component.tsx
DipperTheDan: The performance side of things is a valid consideration. I'd like to think that if Flat Table is used and it has to render that much data, it would be done so using a Pager component to limit the amount of rows rendered to 10, 25 etc.

robinzigmond (approved):

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -312,10 +307,9 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
             css`
               position: -webkit-sticky;
               :after {
-                content: none;
                 border: none;
                 content: "";
-                height: ${getFocusHeight(size)} ${newFocusStyling};
+                height: ${rowHeight}px ${newFocusStyling};
```
edleeks87, line 312: shouldn't this be
```suggestion
height: ${rowHeight};
${newFocusStyling}
```
]

DipperTheDan: It needs the `px` on the height otherwise it applies it as a percentage. The moving of `${newFocusStyling}` out of the height works though. I hadn't moved this as I figured it had to be done that way for a reason. Guess it was just a mistake originally that still worked 🤷🏻

edleeks87: Previously it was a function call that added the `;` as part of the string it returned, now it's the actual pixel value so will need adding

robinzigmond (approved):

edleeks87 (approved):

stephenogorman (approved):

PR #6528, opened 2024-01-18
fix(icon): bgSize and fontSize bug fix

### LINKED ISSUE #5812: Icon - cannot use fontSize prop along with bgSize
### Current behaviour

In carbon Icon - if fontSize prop is used along with bgSize prop, the design of the icon is disturbed 

### Expected behaviour

Required - The size should maintain 

https://codesandbox.io/s/carbon-icon-issue-z0jpr6?file=/src/App.js

### JIRA Ticket (Sage Only)

SBS-47530

### Suggested Solution

_No response_

### Carbon Version

112.0.2

### Design Tokens Version

2.45.0

Firefox, Chrome

MacOS

_No response_

### Confidentiality

### LINKED ISSUE #6353: Focus on link with icon is partially covering icon
### Current behaviour

When the link with an icon is focused, the focus is partially covering the icon 

### Expected behaviour

The focus for a link with an icon should not cover the icon

https://carbon.sage.com/?path=/docs/link--with-icon-align

### JIRA Ticket (Sage Only)

SBS-72761

### Suggested Solution

_No response_

### Carbon Version

120.2.0

### Design Tokens Version

4.23.0

MacOS

_No response_

### Confidentiality

nuria1110:
fix #5812
fix #6353 

### Proposed behaviour

- If the `bgSize` value is smaller than the `fontSize` value, the background size will be adjusted to be equal to the icon size. Otherwise, the `fontSize` value has no effect on the background size. 
- "extra-small" option for `bgSize` prop has been deprecated, components internally using this size have been updated. 

### Current behaviour

When using `bgSize` and `fontSize` props together on an `Icon`, if `fontSize` is not small its value will always override the `bgSize` value.

### Checklist

#### QA

### Testing instructions

### DIFF
--- a/src/components/icon/icon-test.stories.tsx
+++ b/src/components/icon/icon-test.stories.tsx
@@ -60,6 +60,11 @@ export default {
         type: "select",
       },
     },
+    bg: {
+      control: {
+        type: "text",
+      },
+    },
   },
 };
 
@@ -112,6 +117,7 @@ export const All = () => (
               fontSize={fontSize}
               bgShape={bgShape}
               bgSize={bgSize}
+              bg="#00b000"
             />
           ));
         });
--- a/src/components/icon/icon.component.tsx
+++ b/src/components/icon/icon.component.tsx
@@ -10,6 +10,7 @@ import { ICON_TOOLTIP_POSITIONS } from "./icon-config";
 import { IconType } from "./icon-type";
 import { TooltipPositions } from "../tooltip/tooltip.config";
 import { TabTitleContext } from "../tabs/__internal__/tab-title";
+import Logger from "../../__internal__/utils/logger";
 
 export type LegacyIconTypes =
   | "help"
@@ -57,6 +58,8 @@ export interface IconProps extends Omit<StyledIconProps, "type">, MarginProps {
   tabIndex?: number;
 }
 
+let deprecatedExtraSmallBgSizeTriggered = false;
+
 const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
   (
     {
@@ -64,7 +67,7 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
       ariaLabel,
       bg,
       bgShape,
-      bgSize = "small",
+      bgSize,
       className,
       color,
       disabled,
@@ -92,6 +95,13 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
       tooltipFlipOverrides.every((override) =>
         ICON_TOOLTIP_POSITIONS.includes(override)
       );
+
+    if (!deprecatedExtraSmallBgSizeTriggered && bgSize === "extra-small") {
+      deprecatedExtraSmallBgSizeTriggered = true;
+      Logger.deprecate(
+        "The `extra-small` variant of the `bgSize` prop for `Icon` component has been deprecated and will soon be removed."
+      );
+    }
 
     if (tooltipFlipOverrides) {
       invariant(
@@ -148,7 +158,7 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
       "aria-hidden": ariaHidden,
       "aria-label": ariaLabel,
       bg,
-      bgSize,
+      bgSize: bgSize || fontSize,
       bgShape,
       className: className || undefined,
       color,
--- a/src/components/icon/icon.spec.tsx
+++ b/src/components/icon/icon.spec.tsx
@@ -25,6 +25,7 @@ import getColorValue from "../../style/utils/get-color-value";
 import { IconType } from "./icon-type";
 import { TooltipPositions } from "../tooltip/tooltip.config";
 import { TabTitleContext } from "../tabs/__internal__/tab-title/tab-title.component";
+import Logger from "../../__internal__/utils/logger";
 
 interface MismatchedPairs {
   prop: LegacyIconTypes;
@@ -60,6 +61,34 @@ describe("Icon component", () => {
     { prop: "success", rendersAs: "tick" },
     { prop: "messages", rendersAs: "message" },
   ];
+
+  describe("console warnings", () => {
+    it("should display 'extra-small' deprecation warning once", () => {
+      const loggerSpy = jest.spyOn(Logger, "deprecate");
+
+      mount(<Icon type="home" bgSize="extra-small" />);
+
+      expect(loggerSpy).toHaveBeenCalledWith(
+        "The `extra-small` variant of the `bgSize` prop for `Icon` component has been deprecated and will soon be removed."
+      );
+      expect(loggerSpy).toHaveBeenCalledTimes(1);
+
+      loggerSpy.mockRestore();
+    });
+
+    it("should display larger bgSize warning once", () => {
+      const consoleSpy = jest.spyOn(console, "warn");
+
+      mount(<Icon type="home" bgSize="small" fontSize="medium" />);
+
+      expect(consoleSpy).toHaveBeenCalledWith(
+        '[WARNING - Icon] The "small" `bgSize` is smaller than "medium" `fontSize`, the `bgSize` has been auto adjusted to a larger size.'
+      );
+      expect(consoleSpy).toHaveBeenCalledTimes(1);
+
+      consoleSpy.mockRestore();
+    });
+  });
 
   describe.each(mismatchedPairs)(
     "mismatched pairs of props and icons retrieved",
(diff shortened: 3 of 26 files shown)


src/components/icon/icon.style.ts
```diff
@@ -57,9 +57,24 @@ export interface StyledIconInternalProps {
   theme?: ThemeObject;
 }
 
+const sizeValues: Record<BgSize | FontSize, number> = {
```
tomdavies73, line 60: **praise**: good use of this here, the numerical values assigned to each give you an easy way to do calculations based on size further down the line

src/components/icon/icon.style.ts
```diff
@@ -57,9 +57,24 @@ export interface StyledIconInternalProps {
...
+  "extra-small": 1,
+  small: 2,
+  medium: 3,
+  large: 4,
+  "extra-large": 5,
+};
+
 function adjustIconBgSize(fontSize?: FontSize, bgSize?: BgSize) {
```
tomdavies73, line 68: **praise**: function looks good, and all potential branches are covered nicely

src/components/icon/icon-test.stories.tsx
```diff
@@ -112,6 +112,7 @@ export const All = () => (
               fontSize={fontSize}
               bgShape={bgShape}
               bgSize={bgSize}
+              bg="#00b000"
```
tomdavies73, line 115: **question**: just wondering if there's a reason why this background colour has been changed

tomdavies73 (commented):
Looks good to me, just wondering if you've had any thoughts about adding a console warning to let consumers know they've passed a bgSize higher than their fontSize?

A console warning in icon component would suffice, plus one unit test

src/components/icon/icon-test.stories.tsx
robinzigmond: I assume it's so the background is actually visible, to better see the different sizes of background vs icon. (It took me a little while to realise that you have to scroll down a fair way to see the icons with this green background, as there are lot of icons before these ones in the story!)

src/components/icon/icon.stories.mdx
```diff
@@ -73,6 +73,8 @@ such as `tooltipPosition`, `tooltipVisible` , `tooltipBgColor`, `tooltipFontColo
 
 ### Various bgSizes
 
+If using the `bgSize` prop in combination with `fontSize`, the `fontSize` prop value will take precedence and the background size may be adjusted to accommodate for a larger icon.
```
robinzigmond, line 76:
```suggestion
If using the `bgSize` prop in combination with `fontSize`, the `fontSize` prop value will take precedence and the background size may be adjusted to accommodate a larger icon. 
```

**suggestion** (non-blocking): may just be me but "accomodate for" seems weird, I think it reads fine without the "for"

src/components/icon/icon.style.ts
robinzigmond: **suggestion** (non-blocking): while I agree with Tom that this is a nice way to handle it, since this object is only used inside `adjustIconBgSize` I think it's better in principle for this `const` definition to be inside that, so it's clear that it's only intended to be used in that util.

That doesn't stop us moving it outside later if it turns out we want to use the object in other tests.

src/components/icon/icon.style.ts
```diff
@@ -57,9 +57,24 @@ export interface StyledIconInternalProps {
...
-  if (fontSize && fontSize !== "small") {
-    return iconConfig.backgroundSize[fontSize];
+  if (fontSize && bgSize && fontSize !== "small") {
```
robinzigmond, line 69: **question**: do we need the `fontSize !== "small"` check here at all any more? I think prior to your change that was necessary in order to keep the background as per the `bgSize` prop when `fontSize` was small, but the logic you have here now I don't think needs any exception. (If `fontSize` is `"small"` then `bgSizeValue < fontSizeValue` can never be true so we get `iconConfig.backgroundSize[bgSize]` returned, which is correct and the exact same as we'd get from line 80.

Not a huge deal but I find it more readable without the extra condition, which makes you stop and think "hang on, why do we have a special case for small font-size"? 😅

robinzigmond: The above comment was made before I noticed that `extra-small` was a possible value for `bgSize` (but not `fontSize`)! In this case we can have `bgSizeValue < fontSizeValue` with `fontSizeValue="small"` - which means what you've got here would ignore that case. So if the consumer provides `<Icon fontSize="small" bgSize="extra-small" />`, your code here would end up with an `extra-small` background size, whereas my suggestion would result in `small`. I think `small` is correct here as otherwise the background is smaller than the font, which I thought is what we're trying to avoid?

You can actually see this case in the icons you added to the `All` story, here's a screenshot highlighting the problematic size combination (small icon in extra-small background): 
image
Maybe it doesn't look so bad but I'll leave that with you and/or Harpal to judge!

src/components/icon/icon-test.stories.tsx
nuria1110: Robin is correct here - I noticed that the background was not visible even if it has been set so adding a background colour made changes easier to see.

src/components/icon/icon.stories.mdx
nuria1110: I agree, adding a new story to demonstrate this might be the better option.

src/components/icon/icon.style.ts
nuria1110: Harpal has suggested removing the "extra-small" `bgSize`, so I've deprecated it and added a warning for this. 
This prop was only being used in components internally so they have been adjusted slightly to avoid any large visual changes.

src/components/icon/icon.spec.tsx
```diff
@@ -61,6 +62,35 @@ describe("Icon component", () => {
...
+      mount(<Icon type="home" bgSize="extra-small" />);
+
+      expect(loggerSpy).toHaveBeenCalledWith(
+        "The `extra-small` variant of the `bgSize` prop for `Icon` component has been deprecated and will soon be removed."
+      );
+      expect(loggerSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it("should display larger bgSize warning once", () => {
+      loggerSpy = jest.spyOn(console, "warn");
```
robinzigmond, line 84: **suggestion** (non-blocking): I don't think we should call the mock `loggerSpy` in this case as it's not for anything to do with the `Logger`, it's for `console.warn`. I'd call it something like `warnSpy` instead. Obviously that's a bit less convenient because you'd need two separate variables and to `mockRestore` on both - so don't feel obliged to do it.

src/components/icon/icon.component.tsx
```diff
@@ -93,6 +96,13 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
         ICON_TOOLTIP_POSITIONS.includes(override)
       );
 
+    if (!deprecatedExtraSmallBgSizeTriggered && bgSize === "extra-small") {
```
tomdavies73, line 99: **comment**: I still think its worth adding a deprecation warning if `fontSize` has exceeded `bgSize` to let consumers now that the `bgSize` has been adjusted accordingly

src/components/icon/icon.style.ts
```diff
@@ -58,8 +58,27 @@ export interface StyledIconInternalProps {
...
+  };
+
+  if (fontSize && bgSize) {
+    const fontSizeValue = sizeValues[fontSize];
+    const bgSizeValue = sizeValues[bgSize];
+
+    if (bgSizeValue < fontSizeValue) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `The "${bgSize}" \`bgSize\` is smaller than "${fontSize}" \`fontSize\`, the \`bgSize\` has been auto adjusted to a larger size.`
```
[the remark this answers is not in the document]
tomdavies73: Agree, also might be a good idea to follow a similar sort of structure we use for other warnings, `[WARNING - Icon]` etc.

src/components/icon/icon.stories.tsx
```diff
@@ -87,6 +87,32 @@ export const VariousBgSizes: ComponentStory<typeof Icon> = () => {
   );
 };
 
+export const BgSizesAndFontSizes: ComponentStory<typeof Icon> = () => {
```
tomdavies73, line 90: **suggestion** (non-blocking): I'd consider adding a story where the `fontSize` value is larger than that of `bgSize` too

robinzigmond (approved):

tomdavies73 (approved):

divyajindel (commented):
Looks good to me!

divyajindel (approved):

robinzigmond (approved):

PR #6535, opened 2024-01-22
test(text-editor): playwright refactor

stephenogorman:
### Proposed behaviour

Refactor Text-Editor tests from Cypress to Playwright

### Current behaviour

Tests are currently Cypress

### Checklist

#### QA

### Additional context

N/A

### Testing instructions
- [x] Run Form tests in Playwright Test Runner to check if the `text-editor.pw.tsx` file passed
- [x] Run all other tests in Playwright Test Runner to check none of the other `*.pw.tsx` files have regressed


src/components/text-editor/text-editor.pw.tsx
```diff
@@ -0,0 +1,590 @@
...
+  test(`should render TextEditor with required prop`, async ({
+    mount,
+    page,
+  }) => {
+    await mount(<TextEditorCustom required />);
+
+    await verifyRequiredAsteriskForLabel(page);
+  });
+
+  ([
```
edleeks87, line 283: Comment (non-blocking): some of these tests are quite complex and verbose, I wonder if it would be better to separate it all rather than loop over the various test cases, it would help remove the `if` statements etc

edleeks87 (dismissed):

edleeks87 (approved):

src/components/text-editor/text-editor.pw.tsx
```diff
@@ -0,0 +1,613 @@
...
+      "rgb(255, 188, 25) 0px 0px 0px 3px, rgba(0, 0, 0, 0.9) 0px 0px 0px 6px"
+    );
+    await expect(toolbarNumber).toHaveCSS(
+      "outline",
+      "rgba(0, 0, 0, 0) solid 3px"
+    );
+    await expect(toolbarNumber).toHaveCSS("position", "relative");
+  });
+});
+
```
sianford, line 554: We should have tests here for the optional buttons too (e.g. Save and Cancel). Mainly for the operation using the keyboard and focus states as Chromatic can snapshot the default state. When they're present, the right arrow key should cycle back to the styling keys but pressing Tab on the numbered-list key should move focus to the first optional Button.

src/components/text-editor/text-editor.pw.tsx
```diff
@@ -0,0 +1,613 @@
...
+      }
+      const beforePseudoElement = window.getComputedStyle(
+        validationIcon,
+        "::before"
+      );
+      return beforePseudoElement ? beforePseudoElement.color : null;
+    });
+    expect(iconColor).toBe(VALIDATION.INFO);
+  });
+
```
sianford, line 377: I think it would be useful to also have a test for the input hint text prop in the new validation example using the `testdata` chars.

src/components/text-editor/text-editor.pw.tsx
edleeks87: We do already have this unit tested so I would be inclined for now to leave it as is until we can do a full review of what we test where when this playwright epic is finished.

src/components/text-editor/text-editor.pw.tsx
```diff
@@ -0,0 +1,613 @@
...
+    await page.keyboard.press("Tab");
+    await expect(textEditorToolbar(page, "bold")).toBeFocused();
+    await page.keyboard.press("ArrowRight");
+    await expect(textEditorToolbar(page, "italic")).toBeFocused();
+    await page.keyboard.press("Tab");
+    await page.keyboard.press("Shift+Tab");
+    await expect(textEditorToolbar(page, "bold")).toBeFocused();
+  });
+});
+
```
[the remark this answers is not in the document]
edleeks87: Good point we should cover this in a browser 👍

src/components/text-editor/text-editor.pw.tsx
edleeks87: This is also tested in unit tests so I'm not overly concerned about it being tested in a browse. If you want to add it for the sake of consistency with the other pw files then that makes sense, but these type of tests will be unit tests in the long run though

edleeks87 (approved):

src/components/text-editor/text-editor.pw.tsx
sianford: Does the unit test cover the special char strings that we use in Playwright tests? If so, then I'm happy to leave it in unit test. The more that can be shifted left while providing full coverage, the better for me.

edleeks87: No it just tests that it is configurable, ie I pass foo and label is foo type thing. Is there any reason why we need to test these like this as if they render or not isn't down to the component but browser/ project that determines it? As I say don't mind it being added just not sure there's any increased value beyond consistency

sianford: In the past, we had issues where some special characters in labels, hint fields etc, didn't perform the correct escape action and littered them with a load of `&` chars instead of the correct ones. I remember it regressed at one point, so we tend to test all of these in the browser now.

sianford (approved):

edleeks87 (approved):

edleeks87 (approved):

sianford (approved):

PR #6542, opened 2024-01-24
fix(pages): maintain height of content when page slides in

### LINKED ISSUE #6492: Pages Component visible scroll in page then it is not necessary
### Current behaviour

Then you pass to page with bigger height you can see scroll during 1 second at the left.

### Expected behaviour

Scroll should not be visible if all pages has smaller height then the parent element.

https://codesandbox.io/p/sandbox/pages-example-2xzqq2

### JIRA Ticket (Sage Only)

SBS-78793

### Suggested Solution

Detect biggest height of pages and asign this value to the height of all pages OR modify css overflow property.

### Carbon Version

124.4.1

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
Currently, when moving from a page with a small amount of content to one with more content, the height of the new page content is truncated (with a scrollbar) and suddenly "jumps" to the new height when the transition is completed - a rather unexpected effect. This is due to the absolute position on the page content as it slides in, which stops its height being taken into account when calculating the parent's height. The fix is to remove this absolute positioning (moving it up to the page wrapper so that the slide effect still works).

fix #6492

### Proposed behaviour

When a new page slides in which is taller than the previous one, it should take up the full space it needs while sliding.

### Current behaviour

The new page is truncated to the height of previous page while sliding, suddenly "jumping" to its full height only when the slide has finished

### Checklist

#### QA

### Testing instructions

https://codesandbox.io/s/pages-scroll-forked-c22mv3


src/components/pages/pages-test.stories.tsx
```diff
@@ -125,3 +126,402 @@ export const DefaultStory = ({
...
+            moderate but not too-big amount of text here.
+          </p>
+        </Page>
+        <Page title={<Heading title="Page 2" />}>
+          <div>
+            <Button onClick={() => setPageIndex(0)}>Previous page</Button>
+            <Button onClick={() => setPageIndex(2)}>Next page</Button>
+          </div>
+          <p>
+            Lots and lots of text here. Lots and lots of text here. Lots and
```
DipperTheDan, line 155: Nobody can accuse you of not putting a lot of text here 😆

src/components/pages/pages.pw.tsx
```diff
@@ -161,9 +161,8 @@ test.describe("Accessibility tests for Pages component", () => {
       await mount(<PagesComponent initialpageIndex={propValue} />);
 
       await dataComponentButtonByText(page, "Open Preview").click();
-      const pageComponentTitle = getDataElementByValue(page, "title").nth(1);
 
-      await checkAccessibility(page, pageComponentTitle);
```
DipperTheDan, line 166: suggestion(non-blocking): I had to add these `pageComponentTitle` to avoid some flakiness I witnessed when doing the refactor. The accessibility checks were running before the component had loaded.

robinzigmond: Hmm, I removed these only because these tests started failing (I'd raised the PR initially without checking PW locally and they failed on here, and then failed again locally when I ran them), and it turned out removing that second paramater made them pass (and several times in a row, so I hoped it wasn't just flakiness). I didn't dig too deeply but I think I assumed my CSS changes had altered how the animations were happening enough that waiting for that element, which was necessary before, was now somehow breaking things.

I'll take a closer look locally at what is happening in the browser, both with and without that removal, to see if it's OK to remove - and if not I'll see if there's another solution to whatever was making it fail.

edleeks87 (approved):

src/components/pages/pages.pw.tsx
robinzigmond: Thanks for making me look at this again. From watching the replay, the accessibility tests in my version definitely don't seem to be running at the right time - it seems to be after PW has already navigated to a new page for some reason.

I've looked more deeply at why the tests were failing with these additional waits in, and believe I've come up with a better solution for those (for this one it's rewriting the `pageComponentTitle` locator, for the others it's changing a trivial bit of CSS in the component). Will get those changes pushed this afternoon 👍

edleeks87 (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #6544, opened 2024-01-24
fix(filterable-select, multi-select): ensure whitespace is trimmed fom the input when filtering

### LINKED ISSUE #6446: Filterable select should trim the value
### Current behaviour

https://carbon.sage.com/?path=/docs/select-filterable--default#default
In the example above if we type `Amber` we see on the list the `Amber` option
If we type through ` Amber` or `Amber ` we do not see the option on the list

### Expected behaviour

I think we should trim that value, or at least trim start and the user can see the `Amber` option

https://codesandbox.io/s/brave-chatelet-zvlmrq?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

latest

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

tomdavies73:
fix #6446

### Proposed behaviour

Ensures that values typed into the input with leading and trailing whitespace, do not affect the filtering of any options in both the `FilterableSelect` and `MultiSelect` components

### Current behaviour

Currently, any whitespace leading or trailing in the respective components' input results in no options being filtered

### Checklist

#### QA

### Testing instructions


src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -442,17 +444,17 @@ export const FilterableSelect = React.forwardRef(
...
-        .startsWith(filterText?.toLowerCase());
+        .startsWith(filterText?.trim().toLowerCase());
       const isTextboxActive = !disabled && !readOnly;
 
       if (
         isTextboxActive &&
         textboxRef &&
-        filterText?.length &&
-        textValue?.length > filterText?.length &&
+        filterText?.trimStart().length &&
```
edleeks87, line 453: suggestion: you call this a few times here so it might be a but more readable to store it in a variable like you did above

``` 
const trimmed = filterText?.trimStart()

File: src/components/select/filterable-select/filterable-select.spec.tsx [hunk 2]
```diff
@@ -613,6 +617,204 @@ describe("FilterableSelect", () => {
     });
   });
 
+  describe("when the filter text contains whitespace,", () => {
```
edleeks87, line 620: comment: I think it would be better to avoid over nesting the describe blocks here. Obviously there's already a lot of nesting in the spec files for the Select components but I think we should avoid adding more. The console output will also be a little confusing as right now it will output like so:

```
when the filter text contains whitespace
  which is leading whitespace:
     the input value is trimmed and filled correctly

Comment (edleeks87) [reply src/components/select/filterable-select/filterable-select.spec.tsx:620]: Maybe something like this would work:

```
it("should render the input with value trimmed and filled correctly when the filterText contains leading whitespace", () => {
...
it("should render the input with selection range set correctly when the filterText contains leading whitespace", () => {

edleeks87: Happy to chat about this on slack tomorrow, also don't mind one `describe` if you prefer to encapsulate it a bit more than the above would, but I think the console output needs tidying up and some nesting removed

src/components/select/multi-select/multi-select.spec.tsx
```diff
@@ -779,6 +783,122 @@ describe("MultiSelect", () => {
     });
   });
 
+  describe("when the filter text contains whitespace,", () => {
```
edleeks87, line 786: Same comment as above wrt to tidying up the output when these tests run and maybe removing some nesting of describes

src/components/select/filterable-select/filterable-select.spec.tsx
Parsium: **suggestion(non-blocking)**: what do you think about the following structure so we could remove the top `describe` block? ("when the filter text contains whitespace,")

```
describe("when filter text contains leading whitespace", () => { /* tests... */});
describe("when filter text contains whitespace that is neither leading or trailing", () => { /* tests... */});
describe("when filter text is only whitespace", () => { /* tests... */});
describe("when filter text contains trailing whitespace", () => { /* tests... */});
describe("when filter text contains both leading and trailing whitespace", () => { /* tests... */});
```

Parsium (approved):

edleeks87 (approved):
Good work @tomdavies73 the tests are a lot more readable now 👏

ZhuoyuJin (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6549, opened 2024-01-29
test(popover-container): convert tests to playwright - FE-6146

DipperTheDan:
### Proposed behaviour

Convert tests to Playwright

### Current behaviour

Tests currently use Cypress.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    await expect(popoverContainerContent(page)).toHaveAttribute(
+      "aria-describedby",
+      testString
+    );
+  });
+
+  test("should render with openButtonAriaLabel", async ({ mount, page }) => {
+    await mount(<PopoverContainerComponent openButtonAriaLabel={testString} />);
+
+    const popoverContainerOpenComponent = getDataElementByValue(
```
robinzigmond, line 199: **suggestion** (non-blocking): I'm a little curious why you haven't just used `popoverSettingsIcon` here (btw that seems a strange name for what is the icon to open it...), and `popoverCloseIcon` in the following test, since you've bothered defining them and have even imported them to use in other tests

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+      "aria-label",
+      testString
+    );
+  });
+
+  test("should render with renderOpenComponent", async ({ mount, page }) => {
+    await mount(<WithRenderOpenButtonComponent />);
+
+    const buttonComponent = getComponent(page, "button");
+    await buttonComponent.getByText("Test");
```
robinzigmond, line 241: **suggestion**: I might be being dense and in need of a coffee - but what is this actually `awaiting`? It certainly doesn't modify what `buttonComponent` is referring to, which is the thing we're asserting against on the next line.

Assuming we want to ensure we're getting the button with the specific text "test", I'd just do that directly, with a locator like `const buttonComponent = page.getByRole("button", { name: "Test" });`, or even `const buttonComponent = page.getByText("Test")` if we're confident there's no other element with that text.

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    const popoverContainerContentElement = popoverContainerContent(page);
+
+    await expect(popoverContainerContentElement).not.toBeVisible();
+  });
+
+  test("should render with renderCloseComponent", async ({ mount, page }) => {
+    await mount(<WithRenderCloseButtonComponent />);
+
+    const getComponentElementButton = getComponent(page, "button");
+    await getComponentElementButton.getByText("Test");
```
robinzigmond, line 258: **suggestion**: same comment/suggestion here as on the previous test

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    locator.evaluate((element) => element.getBoundingClientRect().top);
+
+  const getYValue = (locator: Locator) =>
+    locator.evaluate((element) => element.getBoundingClientRect().y);
+
+  ([
+    [true, 102, 178],
+    [false, 126, 202],
+  ] as const).forEach(([coverButton, yAndTopValueMin, bottomValueMin]) => {
+    test(`should render with shouldCoverButton prop set to ${coverButton}`, async ({
```
robinzigmond, line 280: **praise**: well done for rewriting the flaky test in a way that's easy to read and (I assume!) actually works

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+      );
+      const positionY = await getYValue(popoverContainerContentElement);
+
+      await expect(positionBottom).toBeLessThan(bottomValueMin);
+      await expect(positionTop).toBeLessThan(yAndTopValueMin);
+      await expect(positionY).toBeLessThan(yAndTopValueMin);
+    });
+  });
+
+  ([...keyToTrigger] as const).forEach((key) => {
```
robinzigmond, line 307:
```suggestion
  keyToTrigger.forEach((key) => {
```

**suggestion**: various suggestions here, which apply to all the other tests using this array:

- there's absolutely no need to spread it, we can just refer to the array itself
- the `as const` is unnecessary, there's already an `as const` in the definition (I'm not even sure if that's necessary but I haven't checked, and even if it's not I don't think it does any harm)
- not shown here, but as I've said in other PRs, I'd rather rename it `keysToTrigger` because it's an array and has more than one key in it

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    await mount(<PopoverContainerWithSelect />);
+
+    await popoverSettingsIcon(page).click();
+    await selectText(page).click();
+    await selectListText(page, "green").click();
+
+    await expect(popoverContainerContent(page)).toBeVisible();
+  });
+
+  // Due to a bug in Playwright whereby the Escape key fires a keyup event before the keydown event.
```
robinzigmond, line 361: **praise**: great comment to explain exactly why we're skipping the test and what needs to happen before we re-enable it.

It does give rise to a **question** though - since it seems 1.40.0 has been released, and I assume (perhaps foolishly!) it shouldn't have any major breaking changes from what we're currently using (1.36.2), is there a good reason not to just bite the bullet and upgrade to take the fix as part of this PR?

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    );
+
+    const popoverCloseIconElement = popoverCloseIcon(page);
+    await popoverCloseIconElement.click();
+
+    await expect(callbackCount).toBe(1);
+  });
+
+  ([...keyToTrigger] as const).forEach((key) => {
+    test(`should call onClose callback when a ${key} keyboard event is triggered`, async ({
```
robinzigmond, line 462:
```suggestion
    test(`should call onClose callback when a keyboard event is triggered by ${key} key`, async ({
```

**suggestion** (non-blocking): this is the test description you used for the `onOpen` version a couple of tests up from here, and I much prefer it. Mainly so you don't get "when a Enter keyboard event is triggered" in one of the tests 😅  - but also for consistency.

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+      );
+
+      const popoverCloseIconElement = popoverCloseIcon(page);
+      await popoverCloseIconElement.press(key);
+
+      await expect(callbackCount).toBe(1);
+    });
+  });
+
+  test("should call onClose callback when the escape is pressed", async ({
```
robinzigmond, line 483:
```suggestion
  test("should call onClose callback when escape is pressed", async ({
```

**suggestion** (non-blocking): yes I know this was like this in the Cypress version too, but it reads oddly to me. Either remove "the", as I've suggested, or do "when the escape key is pressed" instead

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    const checkboxElementOne = page.locator(`[role="checkbox"]`).nth(0);
+    const checkboxElementTwo = page.locator(`[role="checkbox"]`).nth(1);
+    const checkboxElementThree = page.locator(`[role="checkbox"]`).nth(2);
+
+    await checkboxElementOne.check();
+    await checkboxElementTwo.check();
+    await checkboxElementThree.check();
+    const mainTextElement = getDataElementByValue(page, "main-text").nth(1);
+    await mainTextElement.click();
+    const pillComponentOne = getComponent(page, "pill").nth(0);
```
robinzigmond, line 653: **suggestion**: we could run these pill checks in parallel rather than in sequence, to speed up the test (we don't do any actions in between so these should all be independent of each other). Something like:

```
const assertions = [0, 1 , 2].map(index => expect(getComponent(page, "pill").nth(index)).toBeVisible();
await Promise.all(assertions);
```

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+  WithRenderCloseButtonComponent,
+  PopoverContainerComponentCoverButton,
+} from "../popover-container/components.test-pw";
+import Portrait from "../portrait";
+
+const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
+const testString = "test-string";
+const keyToTrigger = ["Space", "Enter"] as const;
+
+test.describe("Check props component", () => {
```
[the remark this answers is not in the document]
DipperTheDan: I forgot to make this make actual sense. I'll get it addressed.

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -0,0 +1,683 @@
...
+    test(`should render when open prop is set to ${openValue}`, async ({
+      mount,
+      page,
+    }) => {
+      await mount(<PopoverContainerComponent open={openValue} />);
+
+      const popoverContainerContentElement = popoverContainerContent(page);
+
+      if (openValue) {
+        await expect(popoverContainerContentElement).toBeInViewport();
```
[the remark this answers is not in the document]
DipperTheDan: I'll get it changed to `toBeVisible` as you make a valid point.

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: Good point. I'll make these follow the same pattern of using the settings icon rather than the data element value.

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: I'll get it changed to `const buttonComponent = page.getByText("Test")` as that makes more sense. There are no other specific text using `Test` that could cause issues.

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: Thanks. Well, time will tell on if it actually works 😆

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: Consistently inconsistent. I'll get it changed.

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: I prefer `when the escape key is pressed` so I'll go with that.

src/components/popover-container/popover-container.pw.tsx
DipperTheDan: Big brain thinking 🧠 
Good idea, I'll get it changed to that 👍🏻

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -304,7 +292,7 @@ test.describe("Check props component", () => {
     });
   });
 
-  ([...keyToTrigger] as const).forEach((key) => {
+  [...keysToTrigger].forEach((key) => {
```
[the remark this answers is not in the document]
DipperTheDan: Yep. Forgot about removing the spread syntax. I'll get on it now.

robinzigmond (approved):

sianford (approved):

PR #6553, opened 2024-01-30
feat: accessibility improvements to ActionPopover, SplitButton and MultiActionButton

robinzigmond:
### Proposed behaviour

- ActionPopover, SplitButton and MultiActionButton should not use the `menu` and `menuitem` roles, but simply present buttons in a semantic HTML list
- the consumer should be able to override the `aria-label` attribute for the main button on ActionPopover should
- no `aria-label` should be placed on menu items that open a submenu, so that screenreaders read the text of the item rather than a vague word like "actions"

### Current behaviour

- ActionPopover, SplitButton and MultiActionButton use ARIA menu roles, which are not semantically appropriate for a simple list of actions/buttons
- It is not possible to override the `aria-label` of "actions" on the main button for ActionPopover, which may lead to a poor experience for users (potentially many duplicate labels on a page, none of them meaningful for what the component is actually used for)
- There is the same `aria-label="actions"` on each menu item that opens a submenu, which causes screenreaders to announce the item as "actions" rather than the actual text that a sighted user can read

### Checklist

#### QA

### Testing instructions

NOTE: marked as requiring UX QA, this is for an accessibility review rather than more general UX. (Nothing should have changed for users who don't use assistive technology.)

### DIFF
--- a/src/components/action-popover/action-popover.component.tsx
+++ b/src/components/action-popover/action-popover.component.tsx
@@ -53,6 +53,8 @@ export interface ActionPopoverProps extends MarginProps {
   renderButton?: (buttonProps: RenderButtonProps) => React.ReactNode;
   /** Boolean to control whether menu should align to right */
   rightAlignMenu?: boolean;
+  /** Prop to specify an aria-label for the component */
+  "aria-label"?: string;
 }
 
 const onOpenDefault = () => {};
@@ -68,6 +70,7 @@ export const ActionPopover = ({
   placement = "bottom",
   horizontalAlignment = "left",
   submenuPosition = "left",
+  "aria-label": ariaLabel,
   ...rest
 }: ActionPopoverProps) => {
   const l = useLocale();
@@ -221,7 +224,7 @@ export const ActionPopover = ({
         "data-element": "action-popover-button",
         ariaAttributes: {
           "aria-haspopup": "true",
-          "aria-label": l.actionPopover.ariaLabel(),
+          "aria-label": ariaLabel || l.actionPopover.ariaLabel(),
           "aria-controls": menuID,
           "aria-expanded": `${isOpen}`,
         },
@@ -232,7 +235,7 @@ export const ActionPopover = ({
       <StyledButtonIcon
         role="button"
         aria-haspopup="true"
-        aria-label={l.actionPopover.ariaLabel()}
+        aria-label={ariaLabel || l.actionPopover.ariaLabel()}
         aria-controls={menuID}
         aria-expanded={isOpen}
         tabIndex={isOpen ? -1 : 0}
--- a/src/components/action-popover/action-popover.component.tsx
+++ b/src/components/action-popover/action-popover.component.tsx
@@ -75,7 +75,7 @@ export const ActionPopover = ({
   const [focusIndex, setFocusIndex] = useState(0);
   const [guid] = useState(createGuid());
   const buttonRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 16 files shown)


DipperTheDan (dismissed):

src/components/action-popover/action-popover.pw.tsx
```diff
@@ -475,7 +475,7 @@ test.describe("check functionality for ActionPopover component", () => {
     await actionPopoverButton(page).click();
 
     // check download item has 'download' property
-    const downloadItem = page.getByRole("menuitem", { name: "Download" });
+    const downloadItem = page.getByRole("link", { name: "Download" });
```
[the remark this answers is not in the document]
tomdavies73: I can also see Playwright is passing so, I must be missing something

src/components/action-popover/action-popover.component.tsx
```diff
@@ -221,7 +224,7 @@ export const ActionPopover = ({
         "data-element": "action-popover-button",
         ariaAttributes: {
           "aria-haspopup": "true",
-          "aria-label": l.actionPopover.ariaLabel(),
+          "aria-label": ariaLabel || l.actionPopover.ariaLabel(),
```
tomdavies73, line 227: **comment/question** (non-blocking): I can see here you're using an OR operator, which will default to the en-gb locale. Which by default is just `"Action"`, I was just wondering if we shouldn't change this within the PR to be a tad more descriptive, especially as we're leaving the translation in for the Button which opens the popover. Maybe something like `"Open Popover"` etc.

src/components/action-popover/action-popover.pw.tsx
```diff
@@ -820,6 +820,14 @@ test.describe("check props for ActionPopover component", () => {
       await expect(itemChevron).toHaveAttribute("type", chevronType);
     });
   });
+
+  test("should render with aria-label prop", async ({ mount, page }) => {
```
tomdavies73, line 824: **comment**: Happy to have this left in, however I feel like we can just cover this with a unit test

src/components/action-popover/action-popover.pw.tsx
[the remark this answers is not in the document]
robinzigmond: You are, but no worries, it's really easy to miss - iirc I first made the role `"button"` but then the test failed. The reason is the Download "item" is actually a link - if you look at the story, the `ActionPopoverItem` has an `href`, which makes  `ActionPopoverItem` render an anchor instead of a button.

src/components/action-popover/action-popover.component.tsx
robinzigmond: While I dislike the "actions" default, I don't feel comfortable changing this as it will change the accessible name for any current implementation where they don't currently providfe their own. Come to think of it, it's probably really not a big deal, but I'm also not sure I want a debate now on what to change it to, as none of us really has a clue what it should be changed to. (It really needs a content designer and we don't have one on the team.) Happy to be persuaded though.

DipperTheDan (approved):

tomdavies73 (approved):

DipperTheDan (dismissed):

tomdavies73 (approved):

harpalsingh:
Did a demo with @tempertemper and update is good.

ZhuoyuJin (approved):

tomdavies73 (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6559, opened 2024-01-31
fix(textarea, textbox): ensure enterkeyhint attribute can be set

### LINKED ISSUE #6525: Support `enterKeyHint` on inputs
### Desired behaviour

We'd like to be able to set the `enterKeyHint` property on inputs so that we can have more control over what devices with a virtual keyboard (tablets mainly) display on the enter key.

Usually such a device will say "Next" on the enter key, pressing this tabs to the next input, but then changes to "Go" or "Submit" when it reaches the last input. This is undesirable for us as there are inputs beyond this, they just aren't enabled yet.

The spec for this property is at https://html.spec.whatwg.org/dev/interaction.html#input-modalities:-the-enterkeyhint-attribute.

### Current behaviour

The property does not seem to be passed through to the input in the dom.

### Suggested Solution

Document and pass thr value down to the input in the dom.

_No response_

_No response_

### Confidentiality

tomdavies73:
fix #6525

### Proposed behaviour

The enterkeyhint attribute can be passed to the the internal input component. A known issue where the prop name is filtered by a downstream dependency on Styled Components has been avoided by setting the attribute directly with the` setAttribute()` method. There is also a further capacity to add more attributes directly in the future.

### Current behaviour

Currently, due to a known bug which we can not avoid in `emotion-js is-prop-valid` the enterkeyhint attribute cannot be set as the attribute is incorrectly filtered as invalid

### Checklist

#### QA

### Testing instructions


src/__internal__/input/input.spec.tsx
```diff
@@ -66,6 +66,26 @@ describe("Input", () => {
...
+    "enter",
+    "done",
+    "go",
+    "next",
+    "previous",
+    "search",
+    "send",
+  ] as EnterKeyHintTypes[])(
+    "'enterKeyHint' is correctly passed to the input when prop value is %s",
+    (keyHints) => {
```
edleeks87, line 79: non-blocking:

```suggestion
    (keyHint) => {
```

edleeks87 (approved):

DipperTheDan (approved):

DipperTheDan (approved):

sianford (approved):

edleeks87 (approved):

PR #6564, opened 2024-02-01
fix(popover-container): ensure styling is correct when rendered as a child of menu

### LINKED ISSUE #6538: Broken styling in popover container within menu item
### Current behaviour

When updating to V125.0.0 https://github.com/Sage/carbon/releases/tag/v125.0.0, the popover container used within a menu item breaks, but it doesn't seem to be related to any of the release notes for this version.

I have a codesandbox example here https://codesandbox.io/p/sandbox/quizzical-gould-n4gzj6 which will hopefully illustrate the issue. If you run it as is (on version 124.6.1) you'll see the popover container working correctly but if you updated to 125.0.0 you'll see the broken behaviour

https://github.com/Sage/carbon/assets/110925864/d735dc8f-d354-4566-a20b-a80ceb9e832b

### Expected behaviour

https://github.com/Sage/carbon/assets/110925864/452187b6-87b7-469d-a685-21ceac5e7bb4

https://codesandbox.io/p/sandbox/quizzical-gould-n4gzj6

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

125.0.0

### Design Tokens Version

_No response_

Chrome

Windows

_No response_

### Confidentiality

edleeks87:
fix #6538

### Proposed behaviour

Re-introduces Portal to component so that is is not a direct descendant of any potential component
that may cascade styles targeting the container's children elements.

Adds `useFocusPortalContent` hook to handle overriding the tab ordering

image

### Current behaviour

Container is not rendered in Portal meaning any children may receive style overrides form any parent elements

image

### Checklist

#### QA

### Testing instructions

**For QA:** I've added a new test story to check that the issue raised is fixed as part of the proposed changes


src/hooks/__internal__/useFocusPortalContent/useFocusPortalContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+            document.querySelectorAll(defaultFocusableSelectors) ||
+              /* istanbul ignore next */ []
+          ) as HTMLElement[];
+          const filteredElements = allFocusableElements.filter(
+            (el) => el === target.current || Number(el.tabIndex) !== -1
+          );
+          const nextIndex =
+            filteredElements.indexOf(target.current as HTMLElement) + 1;
+
+          focusCallback?.((ev as unknown) as React.KeyboardEvent<HTMLElement>);
```
edleeks87, line 57: Not a huge fan of this but I couldn't think of a better way to coerce/ get the types to line up here

src/__internal__/popover/popover.component.tsx
```diff
@@ -35,6 +35,8 @@ export interface PopoverProps {
   animationFrame?: boolean;
   // Optional strategy to use for positioning the floating element. Defaults to "absolute".
   popoverStrategy?: "absolute" | "fixed";
+  // Allows child ref to be set via a prop instead of dynamically finding it via children iteration
+  childRefOverride?: MutableRefObject<HTMLDivElement | null>;
```
Parsium, line 39: **question**: would this functionality still work if `Popover` was a forward ref?

src/components/popover-container/popover-container.pw.tsx
```diff
@@ -526,6 +527,40 @@ test.describe("Event tests", () => {
 
     await expect(callbackCount).toBe(0);
   });
+
+  test("should focus the next element after the open button when user tabs and last element in the container is focused", async ({
```
Parsium, line 531: **praise**: nice one for also considering the focus behaviour 👍🏼

src/components/popover-container/popover-container.component.tsx
```diff
@@ -44,20 +45,24 @@ export const renderOpen = ({
...
+    <PopoverContainerOpenIcon
+      tabIndex={tabIndex}
+      onClick={onClick}
+      data-element={dataElement}
+      ref={ref}
+      aria-label={ariaLabel}
+      aria-haspopup={ariaHasPopup}
+      aria-expanded={ariaExpanded}
+      id={id}
+      {...rest}
```
Parsium, line 60: **suggestion**: what do you think about replacing spreading the remaining props with passing `isOpen` directly?

```suggestion
      isOpen={isOpen}
```

With the open icon being a styled `button`, spreading `rest` would technically allow consumers to pass any valid html attributes to it.

src/hooks/__internal__/useFocusPortalContent/useFocusPortalContent.tsx
Parsium, line 1: **issue**: it seems the focus isn't being trapped when `shouldCoverButton` is true:

https://github.com/Sage/carbon/assets/18368713/57cdbedf-5797-44cc-bd8b-c66428a057a3

I assume its because this hook sets the focus outside the trap?

src/hooks/__internal__/useFocusPortalContent/useFocusPortalContent.tsx
edleeks87: I actually realised that the `onClose` callback accepts `Event` as a type for the argument so I can just go with 

```
focusCallback: (ev: KeyboardEvent) => void;

src/__internal__/popover/popover.component.tsx
edleeks87: I'm not really sure what you mean here, Popover doesn't actually render an element it either returns the content/ children or wraps them in a portal and returns that.

src/hooks/__internal__/useFocusPortalContent/useFocusPortalContent.tsx
edleeks87: Yep good spot I changed the implementation to just add a listener but forgot to add a short circuit if the FocusTrap is composed

src/components/popover-container/popover-container.component.tsx
edleeks87: I don't think they can here as it's just the default that renders if they don't set the prop etc. However, I agree it's probably better to explicitly destructure here

DipperTheDan (dismissed):

Parsium (approved):
LGTM @edleeks87 👍🏼 It might be an idea for us to collate our focus management code in the future, just so we don't have focus utils and hooks located in several different places.

edleeks87:
> LGTM @edleeks87 👍🏼 It might be an idea for us to collate our focus management code in the future, just so we don't have focus utils and hooks located in several different places.

Great point @Parsium, tbh there's a lot of housekeeping we could do with our internal folder structure, I'll see if there's a ticket and raise if not 👍

DipperTheDan (approved):

Parsium (approved):

divyajindel (approved):

PR #6576, opened 2024-02-07
fix(message): reduce left padding of content for `transparent` variant

### LINKED ISSUE #6552: `Message` component `transparent` mode displays extra paddings
### Current behaviour

As shown in this storybook link, 
https://carbon.sage.com/?path=/docs/message--default#transparent
The text in the `Message` component has excessive padding, resulting in a wide gap between the icon and the text itself.

Screenshot 2024-01-30 at 08 42 37

### Expected behaviour

Screenshot 2024-01-30 at 08 45 45

https://carbon.sage.com/?path=/docs/message--default#transparent

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

124.6.0

### Design Tokens Version

4.28.0

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
The message conent always has a left padding of 20px. This works fine for most variants with a solid background to the icon, but looks excessively large when the `transparent` variant is used. The padding has therefore been reduced when the `transparent` prop is true.

fix #6552

### Proposed behaviour

There should be less left-padding on the message content for the `transparent` variant, compared to others: 
image

### Current behaviour

There is the same amount of left-padding for the `transparent` variant as for others, which looks excessively large: 
image

### Checklist

#### QA


tomdavies73 (commented):
**suggestion** (non-blocking):LGTM, I was just wondering would it be worth adding a unit test for this too? I know we have full coverage here + a snapshot has been updated but a unit test to just catch any breaks down the line could be a good shout.

I'm happy to approve as it is though.

nineteen88:
> > **suggestion** (non-blocking):LGTM, I was just wondering would it be worth adding a unit test for this too? I know we have full coverage here + a snapshot has been updated but a unit test to just catch any breaks down the line could be a good shout.
> > I'm happy to approve as it is though.
> 
> Thanks. Personally I think this is the kind of thing we shouldn't use unit tests for, and Chromatic is perfect for it. But I'll wait to see what a second reviewer says - more than happy to add a unit test for this if it's thought to be needed.

Yeah I don't believe a unit test is required for this personally.

nineteen88:
I thought when we discussed this as a team we were going to just allow a consumer to override the padding themselves for this issue. But as it's been done this way, I'll approve it and @harpalsingh can feedback if it's fine with the variant solution. It is more consistent with how we would typically do it in fairness..

nineteen88 (approved):

tomdavies73 (approved):

divyajindel (approved):

PR #6577, opened 2024-02-08
feat(textarea): add new borderRadius prop - FE-6363

DipperTheDan:
### Proposed behaviour

Allow a custom border radius to be specified on Textarea component

Screenshot 2024-02-12 at 15 49 37

### Current behaviour

No option to apply a custom border radius to Textarea component

### Checklist

#### QA

### Additional context

N/A
### Testing instructions

- New story has been created called `with borderRadius`. This story should render has expected.
- `Controls` for `borderRadius` have been added to the `test` example of Textarea.
- There should be no visual regressions with Textarea.

### DIFF
--- a/src/__internal__/input/input-presentation.component.tsx
+++ b/src/__internal__/input/input-presentation.component.tsx
@@ -1,4 +1,5 @@
 import React, { useContext } from "react";
+import { BorderRadiusType } from "../../components/box/box.component";
 import InputPresentationStyle, {
   StyledInputPresentationContainer,
 } from "./input-presentation.style";
@@ -29,6 +30,8 @@ export interface CommonInputPresentationProps extends ValidationProps {
   size?: Sizes;
   /** If true, the component has an icon rendered inside */
   hasIcon?: boolean;
+  /** Specify a custom border radius. Any valid border-radius design token, or an array of border-radius design tokens. */
+  borderRadius?: BorderRadiusType | BorderRadiusType[];
 }
 
 export interface InputPresentationProps extends CommonInputPresentationProps {
@@ -37,19 +40,20 @@ export interface InputPresentationProps extends CommonInputPresentationProps {
 }
 
 const InputPresentation = ({
+  align,
+  borderRadius = "borderRadius050",
   children,
-  positionedChildren,
-  prefix,
+  disabled,
+  error,
+  hasIcon,
+  info,
   inputWidth,
   maxWidth,
-  align,
-  disabled,
+  positionedChildren,
+  prefix,
   readOnly,
   size = "medium",
-  error,
   warning,
-  info,
-  hasIcon,
 }: InputPresentationProps): JSX.Element => {
   const { hasFocus, onMouseDown, onMouseEnter, onMouseLeave } = useContext(
     InputContext
@@ -93,6 +97,7 @@ const InputPresentation = ({
         info={info}
         validationRedesignOptIn={validationRedesignOptIn}
         hasIcon={hasIcon}
+        borderRadius={borderRadius}
       >
         {children}
       </InputPresentationStyle>
--- a/src/__internal__/input/input-presentation.style.ts
+++ b/src/__internal__/input/input-presentation.style.ts
@@ -69,19 +69,29 @@ const InputPresentationStyle = styled.div<
       | "warning"
       | "info"
       | "hasIcon"
+      | "borderRadius"
     > &
     Pick<CarbonProviderProps, "validationRedesignOptIn">
 >`
   align-items: stretch;
   background: var(--colorsUtilityYang100);
   border: 1px solid var(--colorsUtilityMajor300);
-  border-radius: var(--borderRadius050);
   box-sizing: border-box;
   cursor: text;
   display: flex;
   flex-wrap: wrap;
   width: 100%;
   margin: 0;
+
+  ${({ borderRadius }) => {
+    if (Array.isArray(borderRadius)) {
+      return `border-radius: ${borderRadius
+        .map((value) => `var(--${value})`)
+        .join(" ")};`;
+    }
+    return `border-radius: var(--${borderRadius});`;
+  }}
+
   ${({ size, hasIcon, align }) =>
     size &&
     css`
--- a/src/__internal__/input/input.component.tsx
+++ b/src/__internal__/input/input.component.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useContext, useRef, useCallback } from "react";
 import StyledInput from "./input.style";
 import { InputContext, InputGroupContext } from "../input-behaviour";
+import { BorderRadiusType } from "../../components/box/box.component";
 
 export interface CommonInputProps
   extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "type"> {
@@ -16,6 +17,8 @@ export interface CommonInputProps
   disabled?: boolean;
   /** HTML id attribute of the input */
   id?: string;
+  /** Specify a custom border radius for the input. Any valid border-radius design token, or an array of border-radius design tokens. */
+  inputBorderRadius?: BorderRadiusType | BorderRadiusType[];
   /** A callback to retrieve the input reference */
   inputRef?: (
     input: React.RefObject<HTMLInputElement | HTMLTextAreaElement>
@@ -101,6 +104,7 @@ const Input = React.forwardRef<
       id,
       name,
       validationIconId,
+      inputBorderRadius = "borderRadius050",
       ...rest
     }: InputProps,
     ref
@@ -235,6 +239,7 @@ const Input = React.forwardRef<
         onBlur={handleBlur}
         onClick={handleClick}
         onChange={handleChange}
+        inputBorderRadius={inputBorderRadius}
       />
     );
   }
--- a/src/__internal__/input/input.style.ts
+++ b/src/__internal__/input/input.style.ts
@@ -2,11 +2,13 @@ import styled, { css } from "styled-components";
 import { CommonInputProps } from "./input.component";
 
 const StyledInput = styled.input<
-  Pick<CommonInputProps, "align" | "disabled" | "readOnly">
+  Pick<
+    CommonInputProps,
+    "align" | "disabled" | "readOnly" | "inputBorderRadius"
+  >
 >`
   background: transparent;
   border: none;
-  border-radius: var(--borderRadius050);
   color: var(--colorsUtilityYin090);
   flex-grow: 1;
   font-size: var(--fontSizes100);
@@ -19,6 +21,15 @@ const StyledInput = styled.input<
     background-clip: text;
     -webkit-background-clip: text;
   }
+
+  ${({ inputBorderRadius }) => {
+    if (Array.isArray(inputBorderRadius)) {
+      return `border-radius: ${inputBorderRadius
+        .map((value) => `var(--${value})`)
+        .join(" ")};`;
+    }
+    return `border-radius: var(--${inputBorderRadius});`;
+  }}
 
   ${({ align }) =>
     align &&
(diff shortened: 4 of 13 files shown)


src/__internal__/input/input-presentation.style.ts
```diff
@@ -69,19 +69,23 @@ const InputPresentationStyle = styled.div<
...
-  border-radius: var(--borderRadius050);
   box-sizing: border-box;
   cursor: text;
   display: flex;
   flex-wrap: wrap;
   width: 100%;
   margin: 0;
+
+  ${({ borderRadius }) =>
+    `border-radius: ${borderRadius || "var(--borderRadius050)"};`}
```
robinzigmond, line 87: **suggestion** (non-blocking): this default `--borderRadius050` could be set as a default prop to `InputPresentation`, which would remove the need to have it here with the `||`. I think it would also make it more obvious to readers of the code what the default is (it should also make the default show up in the props table on Storybook, although I just realised that this is an internal component so that's not relevant here!)

src/__internal__/input/input.style.ts
```diff
@@ -20,6 +22,9 @@ const StyledInput = styled.input<
     -webkit-background-clip: text;
   }
 
+  ${({ inputBorderRadius }) =>
+    `border-radius: ${inputBorderRadius || "var(--borderRadius050)"};`}
```
robinzigmond, line 26: **suggestion** (non-blocking): as above, this default could be handled better (imo) with a default to `inputBorderRadius` in the `Input` component itself

src/components/textarea/textarea-test.stories.tsx
```diff
@@ -92,6 +92,11 @@ export default {
         type: "text",
       },
     },
+    borderRadius: {
```
robinzigmond, line 95: **praise**: nice work updating the `controls` for the test stories - it's something I always forget, but I've confirmed this works nicely 👍

src/components/textarea/textarea.component.tsx
```diff
@@ -328,6 +332,7 @@ export const Textarea = React.forwardRef(
           as="textarea"
           inputRef={inputRef}
           validationIconId={validationRedesignOptIn ? undefined : validationId}
+          inputBorderRadius={borderRadius}
```
robinzigmond, line 335: **question** (non-blocking): I was puzzled that we need to set a `borderRadius` on both `InputPresentation` and on `Input` itself. Experimenting in devtools, I see that we do need it on the wrapping div (that's the `InputPresentation`) - but removing it from the `textarea` element itself (which is what `Input` renders here) seems to have no adverse (or other) effect, the corners are still rounded.

I may have missed something having only played with this for 10 seconds, but if I haven't, we can not only get rid of this prop being passsed here, but all the code you've added/changed in the `Input` component to make this work, keeping only the changes to `InputPresentation`.

src/components/textarea/textarea.spec.tsx
```diff
@@ -711,12 +711,34 @@ describe("componentWillUnmount", () => {
...
+      {
+        borderRadius: "16px 8px 24px 36px",
+      },
+      mount(<Textarea borderRadius="16px 8px 24px 36px" />).find(
+        StyledInputPresentation
+      )
+    );
+  });
+
+  it("renders with the expected custom border radius styling using a design-token value", () => {
```
robinzigmond, line 734: **comment** (non-blocking): I'm not convinced we need 2 separate tests here, since it's a simple string prop that just gets passed straight down to the styles with no messing about. But I don't object to explicitly checking either, it's always better to have too many tests than too few.

src/__internal__/input/input-presentation.style.ts
DipperTheDan: Good suggestion. That didn't occur to me to do it that way. I'm OK with making that change as it's good to make it obvious.

src/components/textarea/textarea-test.stories.tsx
DipperTheDan: Cheers. I asked @sianford if having it there would be of benefit to QA (since the issues with CSB cropped up) and she agreed it would help the QA lot out.

src/components/textarea/textarea.spec.tsx
DipperTheDan: My thoughts here were that if we upgrade the `design-tokens` package at any point, we at least have a unit test to test if any issues were to appear before it gets to the Chromatic stage. I'm easy though. I can remove the test if you think I'm being overly cautious.

src/components/textarea/textarea.component.tsx
DipperTheDan: It's needed for the `Input` component too, I believe. If you remove the border radius from the `Input` it looks this screenshot. 

Screenshot 2024-02-09 at 09 36 43

robinzigmond: fair enough, let's keep it then. I did say I'd probably missed something (I hadn't actually tried typing anything in 😅 )

src/components/textarea/textarea.spec.tsx
robinzigmond: I'm not sure upgrading `design-tokens` would cause any issues that this test would catch (as the token value wouldn't actually change, just the styles that got rendered by it) - but still, let's keep it as there's no such thing really as "overly cautious" when it comes to tests.

robinzigmond (dismissed):

src/__internal__/input/input-presentation.component.tsx
```diff
@@ -29,6 +29,8 @@ export interface CommonInputPresentationProps extends ValidationProps {
   size?: Sizes;
   /** If true, the component has an icon rendered inside */
   hasIcon?: boolean;
+  /** Specify a custom border radius. Any valid CSS string */
+  borderRadius?: string;
```
edleeks87, line 33: Comment: I wonder if we're better off just copying how this was done in the `Box` component for now rather than offering tis level of granular control, maybe a question for @harpalsingh though

https://github.com/Sage/carbon/blob/129cebf1b0d82ea29334d5d769fdac1421c9803c/src/components/box/box.component.tsx#L27
https://github.com/Sage/carbon/blob/129cebf1b0d82ea29334d5d769fdac1421c9803c/src/components/box/box.component.tsx#L61

edleeks87 (commented):
Overall happy with the approach just want the one comment clarifying so we're not having to tighten the prop at a later date etc, good work @DipperTheDan 👍

nineteen88 (dismissed):

src/__internal__/input/input-presentation.component.tsx
DipperTheDan: @harpalsingh has requested we limit to just tokens. I'll get this implemented now 👍🏻

nineteen88 (approved):

src/__internal__/input/input.component.tsx
```diff
@@ -16,6 +17,8 @@ export interface CommonInputProps
   disabled?: boolean;
   /** HTML id attribute of the input */
   id?: string;
+  /** Specify a custom border radius for the input. Any valid CSS string */
```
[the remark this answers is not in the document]
DipperTheDan: Even though it's internal, it's best to keep the descriptions correct so I'll get it updated.

src/components/textarea/textarea.component.tsx
```diff
@@ -117,6 +118,8 @@ export interface TextareaProps
   Pass string to display icon, tooltip and orange border
   Pass true boolean to only display orange border */
   warning?: boolean | string;
+  /** Specify a custom border radius for the component. Any valid CSS string. */
```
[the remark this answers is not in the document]
DipperTheDan: Yep. I forgot to update the descriptions 😆

robinzigmond (approved):

nineteen88 (dismissed):

src/__internal__/input/input-presentation.component.tsx
```diff
@@ -30,8 +30,8 @@ export interface CommonInputPresentationProps extends ValidationProps {
   size?: Sizes;
   /** If true, the component has an icon rendered inside */
   hasIcon?: boolean;
-  /** Specify a custom border radius. Any valid border-radius design token name. */
-  borderRadius?: BorderRadiusType;
+  /** Specify a custom border radius. Any valid string or array of border-radius design tokens. */
```
robinzigmond, line 33: **suggestion** (non-blocking): I'm not sure I like the "any valid string" phrasing when it can only be a design token. It's not actually wrong, since I'm sure you'd argue that in this context "valid" means "equal to a design token name" 😆 But I'd personally prefer to phrase this something like: `any valid border-radius design token, or an array of such tokens`.

src/__internal__/input/input.component.tsx
```diff
@@ -17,8 +17,8 @@ export interface CommonInputProps
   disabled?: boolean;
   /** HTML id attribute of the input */
   id?: string;
-  /** Specify a custom border radius for the input. Any valid border-radius design token name. */
-  inputBorderRadius?: BorderRadiusType;
+  /** Specify a custom border radius for the input. Any valid string or array of border-radius design token name . */
```
robinzigmond, line 20: **comment** (non-blocking): as above for the description

src/components/textarea/textarea.component.tsx
```diff
@@ -118,8 +118,8 @@ export interface TextareaProps
   Pass string to display icon, tooltip and orange border
   Pass true boolean to only display orange border */
   warning?: boolean | string;
-  /** Specify a custom border radius for the component. Any valid border-radius design token name. */
-  borderRadius?: BorderRadiusType;
+  /** Specify a custom border radius for the component. Any valid string or array of border-radius design token. */
```
robinzigmond, line 121: **comment** (non-blocking): as above for the description

src/components/textarea/textarea.stories.mdx
```diff
@@ -159,6 +159,10 @@ You can use the `required` prop to indicate if the field is mandatory.
 
 #### Customer Border Radius
 
+The `borderRadius` prop accepts a single string that will applied to each corner or an array of values that will be appied in the same order as the CSS border-radius values.
```
robinzigmond, line 162: **suggestion** (non-blocking): I think these docs could be a bit clearer imo - because 2 or 3 values are also valid, not just an array of 4. This also doesn't mention that the values must all be design tokens.

Maybe something like this?

```
The `borderRadius` prop accepts either a single design token, or an array of them, that will be applied as the CSS `border-radius` property. See MDN docs (add link here) for examples of how this works with multiple values.
```

src/components/textarea/textarea.stories.mdx
```diff
@@ -159,6 +159,10 @@ You can use the `required` prop to indicate if the field is mandatory.
 
 #### Customer Border Radius
 
+The `borderRadius` prop accepts a single string that will applied to each corner or an array of values that will be appied in the same order as the CSS border-radius values. 
+
+The array will be applied in the order of: border-top-left-radius, border-top-right-radius, border-bottom-right-radius, border-bottom-left-radius.
+
 <Canvas>
```
robinzigmond, line 166: **suggestion** (non-blocking): I think some examples of custom border-radius values with 2/3/4 individual values (certainly I would include 4) will help make things clearer here (users can inspect the DOM and see what CSS is actually applied). [I see you've changed the one example we have so it uses 3 values rather than 1 - that's good but I'd personally keep a 1-value example as that's probably the most common use-case, and show a 4-value one as well, and maybe a 2-value one too. Up to you though, the main point is I think it needs more than one example.]

DipperTheDan: I'll do an example for each. That means we'll have a chromatic snapshot for each should anything go wrong in future.

robinzigmond (approved):

edleeks87 (approved):

stephenogorman (approved):

PR #6579, opened 2024-02-08
feat(button): add new `gradient-white` and `gradient-grey` buttonTypes

edleeks87:
### Proposed behaviour

Adds new `buttonType` options for `gradient-white` and `gradient-grey` buttons

NB: I've avoided adding much unit test past what's required for coverage as I prefer chromatic to capture the visual stuff here

image
image
image
image
image

### Current behaviour

No support for gradient buttons

### Checklist

#### QA

### DIFF
--- a/src/components/button/button-test.stories.tsx
+++ b/src/components/button/button-test.stories.tsx
@@ -95,14 +95,14 @@ export const ButtonAsASiblingStory = ({
   ...args
 }: Partial<ButtonProps>) => {
   return (
-    <div>
+    <Box>
       <Button subtext={subtext} {...args} onClick={action("click")}>
         {children}
       </Button>
       <Button subtext={subtext} {...args} onClick={action("click")} ml={2}>
         {children}
       </Button>
-    </div>
+    </Box>
   );
 };
 
@@ -202,7 +202,7 @@ export const generateButtons = (
 export const generateButtonsNoWrapStory = (buttonType: ButtonTypes) => {
   return BUTTON_SIZES.map((size) => {
     return (
-      <div key={`${buttonType}-${size}`}>
+      <Box key={`${buttonType}-${size}`}>
         <Box width="100px">
           <Button buttonType={buttonType} noWrap size={size}>
             Long button text
@@ -236,7 +236,7 @@ export const generateButtonsNoWrapStory = (buttonType: ButtonTypes) => {
             Long button text
           </Button>
         </Box>
-      </div>
+      </Box>
     );
   });
 };
@@ -259,6 +259,18 @@ export const NoWrapTertiaryButtonsStory = () => {
 
 NoWrapTertiaryButtonsStory.storyName = "noWrap tertiary";
 
+export const NoWrapGradientWhiteButtonsStory = () => {
+  return <Box>{generateButtonsNoWrapStory("gradient-white")}</Box>;
+};
+
+NoWrapGradientWhiteButtonsStory.storyName = "noWrap gradient-white";
+
+export const NoWrapGradientGreyButtonsStory = () => {
+  return <Box>{generateButtonsNoWrapStory("gradient-grey")}</Box>;
+};
+
+NoWrapGradientGreyButtonsStory.storyName = "noWrap gradient-grey";
+
 export const NoWrapDashedButtonsStory = () => {
   return <Box>{generateButtonsNoWrapStory("dashed")}</Box>;
 };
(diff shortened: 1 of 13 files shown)


src/components/button/button-types.style.ts
```diff
@@ -13,6 +13,30 @@ function makeColors(color: string) {
   `;
 }
 
+const disbaledImageStyle = `
```
robinzigmond, line 16:
```suggestion
const disabledImageStyle = `
```

**suggestion**: typo in const name

src/components/button/button.component.tsx
```diff
@@ -115,12 +120,21 @@ function renderChildren({
   tooltipTarget,
 }: /* eslint-enable */
 RenderChildrenProps) {
-  const iconColorMap = {
-    primary: "--colorsActionMajorYang100",
-    secondary: "--colorsActionMajor500",
-    tertiary: "--colorsActionMajor500",
-    darkBackground: "--colorsActionMajor500",
-    dashed: "--colorsActionMajor500",
+  const iconColor = () => {
```
robinzigmond, line 123: **comment** (non-blocking): I actually preferred the previous approach of using an object/map to a `switch` statement. But it's obviously something you cared about (since there was no need to change it) and I don't care enough to block the PR over it.

src/components/button/button.stories.mdx
```diff
@@ -214,6 +214,68 @@ Dark Background buttons can be set to `noWrap` mode to prevent text wrapping.
   <Story name="darkBackground/noWrap" story={stories.DarkBackgroundButtonNoWrap} />
 </Canvas>
 
+### Gradient Buttons
```
robinzigmond, line 217: **comment** (non-blocking): I don't think there was a need to add so many examples of both the new "gradient" types. Personally I think a single example of each, or perhaps 2 (one with and one without an icon) was sufficient - the rest (`disabled` etc) are just combinations of props that consumers would expect to work anyway, we don't need to have a story for every possible combination of props imo!

If this was done for Chromatic then we could move some of these to test stories?

But not a blocker, I don't object to leaving all these in if you tihnk it's best.

src/components/button/button-types.style.ts
edleeks87: 😆 cheers

src/components/button/button.component.tsx
edleeks87: Tbh it was just to avoid having to declare the same colour for both gradient variants, I'm happy to revert it back if the second reviewer also prefers the old approach

src/components/button/button.stories.mdx
edleeks87: yeah I just matched the other variants tbh, my thinking is that once we decide on how we want to structure stories going forward they would all match. I also wanted to make it clear that it supports these but not `destructive` but I think I can achieve that by adding some more text to the main stories. Happy to move these to the test file if the other reviewer agrees

src/components/button/button.component.tsx
robinzigmond: oh I hadn't even noticed the repetition 🤦 That's definitely a small argument in favour of your change - overall I still slightly prefer the object but as I said I can easily live with it either way, happy to leave it up to you and/or a second reviewer.

src/components/button/button-test.stories.tsx
```diff
@@ -594,6 +594,681 @@ export const ButtonDifferentTypes = (props: Partial<ButtonProps>) => {
...
+        Gradient white
+      </Button>
+      <Button buttonType="gradient-grey" {...props}>
+        Gradient Grey
+      </Button>
+    </Box>
   );
 };
+
+export const DisabledButtonsWithImageChildren = () => (
```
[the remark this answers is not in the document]
edleeks87: https://github.com/Sage/carbon/blob/d36baf023839693dfe2c4bb74feb774043e1bc9b/src/components/button/button-test.stories.tsx#L16

It's not being excluded

src/components/button/button.component.tsx
edleeks87: No worries Tom has pointed out that there's other repetition there as well 🤦, I'll update it based on his suggestion but likely go with `if` statements as it can be shorthanded a bit more that way

```
const iconColor = () => {
    if (buttonType === "primary") {
      return "--colorsActionMajorYang100";
    }

    if (buttonType.includes("gradient")) {
      return "--colorsActionMinorYin090";
    }

    return "--colorsActionMajor500";
  };

src/components/button/button-test.stories.tsx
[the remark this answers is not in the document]
edleeks87: You're probably right, that story likely has no value at all but I'm reluctant to remove it as part of this ticket as I don't have the time to verify if it's used anywhere by QA etc

tomdavies73 (approved):

tomdavies73 (dismissed):

robinzigmond (approved):

tomdavies73 (approved):

robinzigmond (approved):

divyajindel (approved):

PR #6580, opened 2024-02-09
feat(loader): add gradient color to component

tomdavies73:
### Proposed behaviour

adds a `color` prop, with the option for a `gradient` type . This new variant will give each loader a different color (green, blue, purple) to achieve a gradient appearance.

Screenshot 2024-02-09 at 14 50 37

### Current behaviour

Currently, there is no way to achieve a gradient appearance on the `Loader` component, as each loader has the same set colour.

### Checklist

#### QA

### DIFF
--- a/src/components/loader/loader-square.spec.tsx
+++ b/src/components/loader/loader-square.spec.tsx
@@ -5,6 +5,11 @@ import StyledLoaderSquare, {
 } from "./loader-square.style";
 import { assertStyleMatch } from "../../__spec_helper__/test-utils";
 import CarbonProvider from "../carbon-provider/carbon-provider.component";
+import Loader from ".";
+
+jest.mock("../../hooks/useMediaQuery", () => {
+  return jest.fn(() => true);
+});
 
 function render(props: StyledLoaderSquareProps = {}) {
   return mount(<StyledLoaderSquare {...props} />);
@@ -12,18 +17,54 @@ function render(props: StyledLoaderSquareProps = {}) {
 
 describe("Loader square", () => {
   let wrapper;
-  it("renders as expected", () => {
-    wrapper = render();
+
+  it.each([0, 1, 2])("each loader square renders as expected", (index) => {
+    wrapper = mount(<Loader />);
     assertStyleMatch(
       {
         backgroundColor: "var(--colorsActionMajor500)",
-        height: "12px",
-        width: "12px",
-        marginRight: "6px",
+        height: "16px",
+        width: "16px",
+        marginRight: "8px",
         borderRadius: "var(--borderRadiusCircle)",
       },
-      wrapper
+      wrapper.find(StyledLoaderSquare).at(index)
     );
+  });
+
+  describe("when the variant prop is set to 'gradient'", () => {
+    it("first loader square has Salem (#13A038) background color", () => {
+      wrapper = mount(<Loader variant="gradient" />);
+
+      assertStyleMatch(
+        {
+          backgroundColor: "#13A038",
+        },
+        wrapper.find(StyledLoaderSquare).at(0)
+      );
+    });
+
+    it("second loader square has Cerulean (#0092DB) background color", () => {
+      wrapper = mount(<Loader variant="gradient" />);
+
+      assertStyleMatch(
+        {
+          backgroundColor: "#0092DB",
+        },
+        wrapper.find(StyledLoaderSquare).at(1)
+      );
+    });
+
+    it("third loader square has Electric Violet (#8F49FE) background color", () => {
+      wrapper = mount(<Loader variant="gradient" />);
+
+      assertStyleMatch(
+        {
+          backgroundColor: "#8F49FE",
+        },
+        wrapper.find(StyledLoaderSquare).at(2)
+      );
+    });
   });
 
   describe("when inside button", () => {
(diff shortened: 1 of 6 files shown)


src/components/loader/loader-square.spec.tsx
```diff
@@ -26,6 +26,41 @@ describe("Loader square", () => {
     );
   });
 
+  describe("when color prop is set to 'gradient'", () => {
+    it("first loader square has Salem (#13A038) background color", () => {
+      wrapper = render({ loaderNumber: 0, color: "gradient" });
+
+      console.log(wrapper);
```
edleeks87, line 33: comment: random console.log

src/components/loader/loader-square.style.ts
```diff
@@ -52,10 +55,16 @@ const getDimentions = (
   `;
 };
 
+// TO-DO: Replace hard coded hex values with tokens if added in the future.
```
edleeks87, line 58: comment: I think it would be better to raise tickets for these rather than add TODOs but I also don't mind if you leave them but include the ticket number etc

src/components/loader/loader.component.tsx
```diff
@@ -12,14 +12,15 @@ import StyledLoaderSquare, {
...
-  extends StyledLoaderSquareProps,
+  extends Omit<StyledLoaderSquareProps, "loaderNumber">,
     MarginProps,
     TagProps {
   /** Specify a custom accessible name for the Loader component */
   "aria-label"?: string;
 }
 
 export const Loader = ({
+  color = "default",
```
edleeks87, line 23: Comment: I think we should use `variant` here so we're consistent with how we've done it in most of our other newer components

src/components/loader/loader.component.tsx
```diff
@@ -43,21 +51,13 @@ export const Loader = ({
...
-            isInsideButton={isInsideButton}
-            isActive={isActive}
-            size={size}
-          />
-          <StyledLoaderSquare
-            isInsideButton={isInsideButton}
-            isActive={isActive}
-            size={size}
-          />
+          {[0, 1, 2].map((index) => (
```
edleeks87, line 54: Suggestion (non-blocking): you could slightly tweak this so you're only passing a single prop to the `StyledLoaderSquare`s instead of `color` and `loaderNumber` etc

```
const StyledLoaderSquare = styled.div<StyledLoaderSquareProps>`
  ${({ size, isInsideButton, isActive, theme, backgroundColor = "var(--colorsActionMajor500)" }) => css`
...
    background-color: ${backgroundColor};
`};

// in component file
...
const isGradient = varient === "gradient";
...

 <StyledLoaderSquare
            isInsideButton={isInsideButton}
            isActive={isActive}
            size={size}
           backgroundColor={isGradient ? "#13A038" : undefined}
          />
          <StyledLoaderSquare
            isInsideButton={isInsideButton}
            isActive={isActive}
            size={size}
           backgroundColor={isGradient ? "#0092DB" : undefined} 
          />
          <StyledLoaderSquare
            isInsideButton={isInsideButton}
            isActive={isActive}
            size={size}
           backgroundColor={isGradient ? "#8F49FE" : undefined} 
          />

File: src/components/loader/loader-square.style.ts [hunk 5]
```diff
@@ -8,6 +8,9 @@ export interface StyledLoaderSquareProps {
   isInsideButton?: boolean;
   /** Applies slate color. Available only when isInsideButton is true. */
   isActive?: boolean;
+  /** Toggle between the default variant and gradient variant */
+  color?: "default" | "gradient";
+  loaderNumber?: number;
```
edleeks87, line 13: comment: If you do keep the same approach here I think it's worth adding a comment even if the new prop isn't external facing

edleeks87 (commented):
Good job @tomdavies73, just some minor comments from me but functionally seems to deliver the requirements

src/components/loader/loader.component.tsx
tomdavies73: Cool, I'll stick with mapping but use this approach

Parsium (commented):
Nice work on this @tomdavies73 👍🏼 Nothing to add than what @edleeks87 has mentioned already.

src/components/loader/loader.component.tsx
Parsium: **suggestion**: what do you think about combining the `gradientLoaderColors` with your map to simplify the code?:

```tsx
{["#13A038", "#0092DB", "#8F49FE"].map((color) => (
    <StyledLoaderSquare
      key={color}
      backgroundColor={
        variant === "gradient"
          ? color
          : "var(--colorsActionMajor500)"
      }
      {...loaderSquareProps}
    />
))}
```

src/components/loader/loader-square.style.ts
```diff
@@ -8,6 +8,10 @@ export interface StyledLoaderSquareProps {
   isInsideButton?: boolean;
   /** Applies slate color. Available only when isInsideButton is true. */
   isActive?: boolean;
+  /** Toggle between the default variant and gradient variant */
+  variant?: string;
```
Parsium, line 12: **suggestion**: because you are conditionally setting the background colour in `loader.component.tsx` depending on `variant` , we could remove `variant` as a prop for this styled component:

```suggestion
```

tomdavies73: We extend `StyledLoaderSquareProps` in the component file, so I just left it in there, but i'm happy to move it back into the component file, however I thought it wouldn't make much difference

edleeks87: Agree with @Parsium let's move that as it's irrelevant to the styled component now

Parsium (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6589, opened 2024-02-15
docs(step-sequence): deprecate component

tomdavies73:
### Proposed behaviour

Added a deprecation warning for component.

### Current behaviour

No deprecation warning present for the component.

### Checklist

#### QA


src/components/step-sequence/step-sequence.spec.tsx
```diff
@@ -1,11 +1,50 @@
...
+import { mount } from "enzyme";
 import TestRenderer from "react-test-renderer";
+import Logger from "../../__internal__/utils/logger";
 
 import StepSequence, { StepSequenceProps } from "./step-sequence.component";
 import StepSequenceItem from "./step-sequence-item/step-sequence-item.component";
 import { testStyledSystemSpacing } from "../../__spec_helper__/test-utils";
 
 describe("StepSequence", () => {
+  describe("Deprecation warning", () => {
```
nineteen88, line 11: comment(non-blocking): I don't think the Deprecation Warning describe block really offers much extra and could probably be removed

src/components/step-sequence/step-sequence.spec.tsx
```diff
@@ -1,11 +1,50 @@
...
+    it("when user uses the component, a deprecation warning is raised only once in the console", () => {
```
nineteen88, line 12: suggestion: `it` blocks should be contextually readable, i.e. `it should display a deprecation warning when component is used for the first time`. 

Otherwise just use `test()`

src/components/step-sequence/step-sequence.stories.mdx
```diff
@@ -7,6 +8,12 @@ import * as stories from "./step-sequence.stories.tsx";
 
 # StepSequence
 
+<DeprecationWarning>
+We've deprecated StepSequence. It's also deprecated in the Design System and we recommend that you do not use it in new products.
```
nineteen88, line 12: suggestion(non-blocking): I wonder if we should just be more to the point here and remove `we`

Very minor and feel free to ignore but maybe something like this:

```suggestion
StepSequence has been deprecated in Carbon and the Design System.
<br />
It is recommended to use the StepFlow component - which is an accessible alternative.
```

nineteen88 (approved):

nuria1110 (approved):

PR #6592, opened 2024-02-16
fix(sort): add a prop and translation support to provide an accessible name

tomdavies73:
### Proposed behaviour

adds the `accessibleName` prop as well as translation support for a new visually hidden span which provides the component with an accessible name

### Current behaviour

The current accessible name uses string interpolation with the sort type and children props, which may result in English-only announcements for users of another language. The accessible name is also contained in a span with the `hidden` attribute applied to it, making it in theory inaccessible to screen readers

### Checklist

#### QA

### DIFF
--- a/src/components/flat-table/flat-table.mdx
+++ b/src/components/flat-table/flat-table.mdx
@@ -10,6 +10,7 @@ import * as FlatTableRowHeaderStories from "./flat-table-row-header/flat-table-r
 import * as FlatTableCheckboxStories from "./flat-table-checkbox/flat-table-checkbox.stories";
 import * as SortStories from "./sort/sort.stories";
 import * as FlatTableStories from "./flat-table.stories";
+import TranslationKeysTable from "../../../.storybook/utils/translation-keys-table";
 
 <Meta title="Flat Table" of={FlatTableStories} />
 
@@ -287,3 +288,21 @@ The `FlatTableRow` relies on specific child composition to identify sticky colum
 ### FlatTableCheckbox
 
 <ArgTypes of={FlatTableCheckboxStories} />
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object to the
+[i18nProvider](../?path=/docs/documentation-i18n--docs).
+
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "sort.accessibleName",
+      description: "The accessible name that is announced on interaction with the component."+
+      " It is recommended that the current sort type as well as the current content of the component"+
+      " are included in this string. Both can be accessed via the `sortContent` and `sortType` parameters.",
+      type: "func",
+      returnType: "string",
+    }
+  ]}
+/>
--- a/src/components/flat-table/flat-table.mdx
+++ b/src/components/flat-table/flat-table.mdx
@@ -178,6 +178,14 @@ By using the `hasMaxHeight` prop you can automatically apply a max-height of 100
 
 <Canvas of={FlatTableStories.WithSortingHeaders} />
 
+### With sorting headers and custom accessible name 
+
+A custom accessible name can be provided to the `<Sort>` component with the `accessibleName` prop. It is recommended 
+that text content of the `<Sort>` component as well as the current `sortType` are included
+in this accessible name.
+
+<Canvas of={FlatTableStories.WithSortingHeadersAndCustomAccessibleName} />
+
 ### With a cell spanning the whole row
 
 <Canvas of={FlatTableStories.WithColspan} />
--- a/src/components/flat-table/sort/sort.component.tsx
+++ b/src/components/flat-table/sort/sort.component.tsx
@@ -1,20 +1,24 @@
 import React, { useRef, useContext } from "react";
 import Event from "../../../__internal__/utils/helpers/events";
+import Typography from "../../typography";
 import { StyledSort, StyledSpaceHolder, StyledSortIcon } from "./sort.style";
 import guid from "../../../__internal__/utils/helpers/guid";
+import useLocale from "../../../hooks/__internal__/useLocale";
 import { FlatTableThemeContext } from "../flat-table.component";
 
 export interface SortProps {
   /** if `asc` it will show `sort_up` icon, if `desc` it will show `sort_down` */
-  sortType?: "ascending" | "descending" | false;
-  /** Callback fired when the `FlatTableSortHeader` is clicked */
+  sortType?: "ascending" | "descending";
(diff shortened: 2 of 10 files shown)


src/components/flat-table/sort/sort.component.tsx
```diff
@@ -11,10 +13,20 @@ export interface SortProps {
...
+   */
+  accessibleName?: string;
 }
 
-const Sort = ({ children, onClick, sortType }: SortProps) => {
+const Sort = ({ children, onClick, sortType, accessibleName }: SortProps) => {
   const id = useRef(guid());
+  const locale = useLocale();
+
+  const verifiedChild = typeof children === "string" ? children : undefined;
```
edleeks87, line 28: Comment: I wonder if we should just bite the bullet here and make a breaking change so only string values can be passed to `children`. It might be worth doing a bit of an investigation and seeing if anyone has passed a node here instead, I can't see why or how it would have worked if they did but it would at least allow us to get a measure of the impact if we do change it etc

src/locales/en-gb.ts
```diff
@@ -153,6 +153,14 @@ const enGB: Locale = {
       close: () => "Close",
     },
   },
+  sort: {
+    accessibleName: (children, sortType) =>
```
edleeks87, line 157: Suggestion: we could probably go with a more intuitive name here rather than `children`

src/locales/en-gb.ts
```diff
@@ -153,6 +153,14 @@ const enGB: Locale = {
...
+      `Sort all ${children ? `${children}s` : "contents"}${
+        sortType
+          ? ` in an ${sortType} order.`
+          : " in an ascending or descending order."
```
edleeks87, line 161: Praise: good idea to handle for this as well 👍

src/components/flat-table/sort/sort.component.tsx
```diff
@@ -11,10 +13,20 @@ export interface SortProps {
   onClick?: () => void;
   /** Sets the content of `FlatTableSortHeader` */
   children?: React.ReactNode;
+  /**
+   * Creates an accessible name for `FlatTableSortHeader`, it is recommended that the
+   * text content of the `FlatTableSortHeader` as well as the current `sortType` are included
+   * in this accessible name
```
Parsium, line 19: **suggestion**: it might be helpful to add an example to our docs demonstrating how to use the `accessibleName` prop. What do you think about using a Storybook story to describe how it should be used instead?

Parsium (commented):
Nice attention to detail regarding all the possible edge cases @tomdavies73 :+1:

To assist consumers with taking in translation key updates, what do you think about splitting the current commmit in two, one for the `accessibleName` prop and one for the translation key? That way the addition of the `sort.accessibleName` key will appear as a clear, single entry in the changelog.

src/components/flat-table/flat-table.stories.mdx
```diff
@@ -359,6 +357,17 @@ The `FlatTableRow` relies on specific child composition to identify sticky colum
...
+<TranslationKeysTable
+  translationData={[
+    {
+      name: "sort.accessibleName",
+      description: "The accessible name which is announced on interaction with the component.",
+      type: "func",
+      returnType: "string",
+    }
+  ]}
+/>
```
Parsium, line 369: **nitpick**: would you be happy to relocate the translation key table below the props section? This is to prevent it from being accidentially overlooked while browsing the page.

```mdx
# FlatTable

## Contents

## Quick Start

## Examples

## Props

## Translation keys

The following keys are available to override the translations for this component by passing in a custom locale object to the
[i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).

<TranslationKeysTable translationData={[ /* ... */>]}>
```

src/components/flat-table/flat-table.stories.mdx
```diff
@@ -233,6 +221,16 @@ By using the `hasMaxHeight` prop you can automatically apply a max-height of 100
   <Story name="with sorting headers" story={stories.WithSortingHeaders} />
 </Canvas>
 
+### With sorting headers and custom accessible name 
+
+A custom accessible name can be provided to the `<Sort>` component with the `accessibleName` prop. It is recomended
```
Parsium, line 226: **typo**: 

```suggestion
A custom accessible name can be provided to the `<Sort>` component with the `accessibleName` prop. It is recommended 
```

Parsium (commented):
Thanks for the changes @tomdavies73 👍🏼 

I noticed we had multiple references of a `FlatTableSortHeader` component, even though such a component doesn't appear to exist. If `Sort` was intended here instead, could you please update the references accordingly?

src/components/flat-table/sort/sort.component.tsx
```diff
@@ -26,10 +30,9 @@ const Sort = ({ children, onClick, sortType }: SortProps) => {
 
   return (
     <>
-      <span hidden id={id.current}>
-        {children}
-        {sortType ? `, sort type ${sortType}` : ", sortable"}
-      </span>
+      <Typography screenReaderOnly id={id.current}>
+        {locale.sort.accessibleName(children, sortType)}
```
edleeks87: I thought the intention was to change `children` to be `string`? Currently we're potentially passing a node here and are having to include it in the translation type

https://github.com/Sage/carbon/pull/6592#discussion_r1497743459

tomdavies73: The type has been changed from a `node` -> `string` in the second commit. Commit includes a breaking change too, to notify consumers

Parsium (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

src/components/flat-table/flat-table.mdx
```diff
@@ -287,3 +296,19 @@ The `FlatTableRow` relies on specific child composition to identify sticky colum
 ### FlatTableCheckbox
 
 <ArgTypes of={FlatTableCheckboxStories} />
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object to the
+[i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
```
Parsium, line 303: **issue**: this URL won't work unfortunately due to a change in Storybook v7 with link generation. Could we change this to the following please?: 

```suggestion
[i18nProvider](../?path=/docs/documentation-i18n--docs).
```

Parsium (approved):

ZhuoyuJin (approved):

PR #6598, opened 2024-02-20
feat(search): search button is rendered at all times when `searchButton` prop is set FE-6078

### LINKED ISSUE #6581: `Submenu with search` storybook is broken, padding changed
### Description

Updating `carbon` from `123.9.0` to `125.12.1` - has revealed the styling of the nested `Search` element has changed.
Appears this issue was introduced in`125.0.1`
Also visible in Carbon storybook

Screenshot 2024-02-12 at 16 12 03

### Reproduction

https://carbon.sage.com/iframe.html?viewMode=docs&id=menu--default-story&args=#submenu-with-search

### Steps to reproduce

_No response_

SBS-84420

### Suggested solution

_No response_

### Carbon version

124.0.0

### Design tokens version

4.34.0

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

This is currently blocking a `Template UI` update to `125.12.1`

### Confidentiality

### LINKED ISSUE #6597: MenuItems do not maintain width when focused in MenuFullscreen
### Description

MenuItems do not have the full page width when focused in MenuFullscreen

image

### Reproduction

www.carbon.sage.com/?path=/story/menu--fullscreen-view-story

### Steps to reproduce

- reduce window size until fullscreen layout renders
- focus menu item
- decide whether to laugh or cry

_No response_

### Suggested solution

_No response_

### Carbon version

latest

### Design tokens version

4.29.0

### Relevant browsers

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

edleeks87:
fix #6581
fix #6597

BREAKING CHANGE: The `threshold` prop has been removed as it no longer does anything in the
redesigned component

BREAKING CHANGE: The `inputRef` is no more

### Proposed behaviour

Implements new designs for `Search` component. The button is now rendered at all times when the
`searchButton` is set; expands the type of this prop to also accept string values which will support
more granular control over the text rendered in the button. Support for overriding the button text
via the locale provider has also been added. Updates have also been made to `Link` and `Menu` style
files to ensure the styling is correct when `Search` is rendered in a `Menu`. Styling targeting the
search input in menus has also been removed. These changes have also addressed the issue with
padding of `MenuItem` components that render `Search`.

Search redesign:
image
image
image
image
image

image
image
image
image

Fixes to menu:
image
image
image

Deprecated the `inputRef` prop, consumers should use `ref` in order to get a ref to the input
element.

### Current behaviour

Designs support search Button appearing on focus or when there is a value in the input but disappearing when the user blurs, this creates issues for users tabbing to navigate as the DOM tries to focus the button that is no longer there.

Style regression with Search in Menu
image

Style regression with menu items and Search in MenuFullscreen

image
image

### Checklist

#### QA

### DIFF
--- a/src/components/search/search.component.tsx
+++ b/src/components/search/search.component.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from "react";
+import React, { useState, useRef, useImperativeHandle } from "react";
 import invariant from "invariant";
 import { MarginProps } from "styled-system";
 import { filterStyledSystemMarginProps } from "../../style/utils";
@@ -70,9 +70,14 @@ export interface SearchProps extends ValidationProps, MarginProps {
   tooltipPosition?: "top" | "bottom" | "left" | "right";
 }
 
+export type SearchHandle = {
+  /** Programmatically focus on root container of Dialog. */
+  focus: () => void;
+} | null;
+
 let deprecateUncontrolledWarnTriggered = false;
 
-export const Search = React.forwardRef(
+export const Search = React.forwardRef<SearchHandle, SearchProps>(
   (
     {
       defaultValue,
@@ -97,13 +102,24 @@ export const Search = React.forwardRef(
       info,
       tooltipPosition,
       ...rest
-    }: SearchProps,
-    ref: React.ForwardedRef<HTMLInputElement>
+    },
+    ref
   ) => {
     const isControlled = value !== undefined;
     const initialValue = isControlled ? value : defaultValue;
     const locale = useLocale();
     const searchRef = useRef<HTMLDivElement>(null);
+    const inputRef = useRef<HTMLInputElement>(null);
+
+    useImperativeHandle<SearchHandle, SearchHandle>(
+      ref,
+      () => ({
+        focus() {
+          inputRef.current?.focus();
+        },
+      }),
+      []
+    );
 
     if (!deprecateUncontrolledWarnTriggered && !isControlled) {
       deprecateUncontrolledWarnTriggered = true;
@@ -170,8 +186,7 @@ export const Search = React.forwardRef(
         });
       }
 
-      const input = searchRef.current?.querySelector("input");
-      input?.focus();
+      inputRef.current?.focus();
     };
 
     const handleMouseDown = (event: React.MouseEvent<HTMLElement>) => {
@@ -233,7 +248,7 @@ export const Search = React.forwardRef(
           onBlur={handleBlur}
           onChange={handleChange}
           onKeyDown={handleKeyDown}
-          ref={ref}
+          ref={inputRef}
           tabIndex={tabIndex}
           error={error}
           warning={warning}
--- a/src/components/search/search.component.tsx
+++ b/src/components/search/search.component.tsx
@@ -29,8 +29,6 @@ export interface SearchProps extends ValidationProps, MarginProps {
   defaultValue?: string;
   /** Prop for `id` */
   id?: string;
-  /** A callback to retrieve the input reference */
-  inputRef?: React.MutableRefObject<HTMLInputElement | null>;
   /** Prop for `name` */
   name?: string;
   /** Prop for `onBlur` events */
@@ -72,7 +70,6 @@ export interface SearchProps extends ValidationProps, MarginProps {
   tooltipPosition?: "top" | "bottom" | "left" | "right";
 }
 
-let deprecateInputRefWarnTriggered = false;
 let deprecateUncontrolledWarnTriggered = false;
 
 export const Search = React.forwardRef(
@@ -94,7 +91,6 @@ export const Search = React.forwardRef(
       placeholder,
       variant = "default",
       "aria-label": ariaLabel = "search",
-      inputRef,
       tabIndex,
       error,
       warning,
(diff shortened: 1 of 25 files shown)


src/components/search/search.style.ts
```diff
@@ -148,38 +164,27 @@ const StyledSearch = styled.div<StyledSearchProps>`
...
+
       ${StyledIcon} {
-        ${darkVariant &&
-        css`
-          ${iconColor &&
+        :not([data-element="search"]) {
+          ${darkVariant &&
           css`
-            color: var(--colorsUtilityMajor400);
+            color: var(--colorsUtilityYang080);
```
DipperTheDan, line 172: **a sudden realisation:** It might actually be this causing the issue that I pointed out in a previous comment. 
Screenshot 2024-02-21 at 14 24 21

src/components/search/search.style.ts
```diff
@@ -24,62 +23,47 @@ const StyledSearch = styled.div<StyledSearchProps>`
...
-    searchIsActive,
     searchHasValue,
     showSearchButton,
     theme,
     variant,
   }) => {
     const darkVariant = variant === "dark";
     const variantColor = darkVariant
-      ? "var(--colorsUtilityMajor200)"
+      ? "var(--colorsUtilityYang080)"
```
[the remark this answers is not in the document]
edleeks87: good spot, will get that addressed thanks

src/components/search/search.style.ts
edleeks87: yeah I haven't handled for the searchButton in dark background basically should just be an additional check to ensure no styling changes when `searchButton` is truthy

src/components/search/search.spec.tsx
```diff
@@ -312,25 +287,79 @@ describe("Search", () => {
...
+      value: "",
+      searchButton: true,
+      id: "Search",
+      name: "Search",
+    });
+    const searchButton = wrapper.find(Button);
+    const icon = wrapper
+      .find(Icon)
+      .findWhere((n) => n.props().type === "search")
+      .hostNodes();
```
Parsium, line 301: **suggestion**: so this test doesn't return a false positive by accidentially finding the search button icon, what do you think of the following addition?:


```suggestion
    const inputIcon = wrapper
      .find(Input)
      .find(Icon)
      .findWhere((n) => n.props().type === "search")
      .hostNodes();
```

then you can do an existence assertion:

```ts
expect(inputIcon.exists()).toBeFalsy();
```

src/components/search/search.pw.tsx
```diff
@@ -568,6 +577,24 @@ test.describe("Functionality tests for Search component", () => {
...
+
+  test("should clear a Search input after enter key pressed and cross icon is focused", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<SearchComponent />);
+
+    const searchDefaultInputElement = searchDefaultInput(page);
+    await searchDefaultInputElement.clear();
+    await searchDefaultInputElement.type(testDataStandard);
```
Parsium, line 591: **suggestion**: what do you think about replacing the `type` action with `fill`? It seems [in Playwright v1.38 the `type` action has been deprecated](https://playwright.dev/docs/release-notes#version-138).

```suggestion
    await searchDefaultInputElement.fill(testDataStandard);
```

src/components/search/search.component.tsx
```diff
@@ -224,6 +169,9 @@ export const Search = React.forwardRef(
           },
         });
       }
+
+      const input = searchRef.current?.querySelector("input");
+      input?.focus();
```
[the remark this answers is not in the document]
edleeks87: I'm not fully convinced that complexity is needed here, this is internal so adding a handle feels like overkill (we'd need to change `Textbox` and `Input` for example) and we'd also need to both handle for the ref not being passed and coercing it into something that allows access to `current`. I'm not sure DOM querying is a problem as long as we are aware of any effects being run and re-fire the query in an effect or as part of the function needing the list: it's the approach we agreed on to remove the extensive use of cloning and iterating children etc

src/components/search/search.style.ts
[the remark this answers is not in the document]
edleeks87: Agreed there's a lot of styles that need overriding in a less than ideal way, it something I'd like to do before we start any work to replace styled-components

src/components/search/search.spec.tsx
edleeks87: Good idea, it'll need to be `.find(Input)` -> `.find(Textbox)` but it can definitely be scoped better

edleeks87: also not sure asserting the props().value is correct here as it will always be undefined 🤦

src/components/search/search.component.tsx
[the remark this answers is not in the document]
edleeks87: I think on reflection this is a worthwhile change to make as it's consistent with how other components handle this and I'll be making a breaking change anyway so would help to batch all the ones relating to Search in one PR

Parsium (approved):

DipperTheDan (approved):

DipperTheDan (approved):

Parsium (approved):

ZhuoyuJin (approved):

DipperTheDan (approved):

PR #6601, opened 2024-02-21
feat(button, button-minor, button-toggle-group): add support for aria-labelledby and aria-describedby (FE-6364)

### LINKED ISSUE #6567: Add `aria-descibedby` support to buttons
### Desired behaviour

When using a screen reader buttons with identical text (in the case of a product catalog) require additional context, such as heading text. Adding support for `aria-describedby` will allow us to add context without causing the accessible name to deviate from the button text.

### Current behaviour

Buttons do not support a `aria-describedby` prop.

### Suggested Solution

_No response_

_No response_

_No response_

### Confidentiality

Parsium:
fixes #6567

### Proposed behaviour

- Add support for `aria-labelledby` and `aria-describedby` for `Button` and `ButtonMinor`
- Add support for `aria-describedby` for `ButtonToggleGroup` - accessible description is set on each toggle button inside the group ([as recommended by this source](https://devdoc.net/web/developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role.html)).
- Simplify `Button` tests
- Update copyright dates

### Current behaviour

- Accessible description cannot be set on `Button` or `ButtonMinor`.[^1]
- Accessible description cannot be set on `ButtonToggleGroup`
- Copyright dates in README and licence are outdated

### Checklist

#### QA

### Testing instructions

Test with the following screen readers: **macOS VoiceOver (with Safari)** and **NVDA (with Chrome and Firefox)**.

- For `Button`:
  1. Open the test story "With External Labels".
  2. With a screen reader active, focus on the button.
  3. Verify if the accessible name "Add Product Addon" and description "You will not be charged until you checkout" are both read out.

- For `ButtonMinor`:
  1. Open the test story "With External Labels".
  2. With a screen reader active, focus on the button.
  3. Verify if the accessible name "Add Product Addon" and description "You will not be charged until you checkout" are both read out.
 
- For `ButtonToggle`:
  1. Open the story "with inputHint".
  2. With a screen reader active, focus on any of the three toggle buttons.
  3. Verify if the accessible name "inputHint example" and description "Hint text" are both read out.

[^1]: Possible to pass any HTML button attribute as a prop, but TypeScript throws an error as they're not officially part of either component's interfaces. https://github.com/Sage/carbon/blob/d0524b57701c790429dfa7e50361897e3841f552/src/components/button/button.component.tsx#L320


src/components/button/button.spec.tsx
```diff
@@ -1,109 +1,102 @@
...
+      expect(loggerSpy).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  testStyledSystemSpacing((props) => <Button {...props}>Test</Button>, {
+    px: "24px",
+    m: "0",
   });
 
+  describe.each([
```
edleeks87, line 81:
```suggestion
  it.each([
```

src/components/button/button.spec.tsx
```diff
@@ -121,234 +114,148 @@ describe("Button", () => {
...
-      wrapper = render({ children: "foo" }, TestRenderer.create).toJSON();
+    consoleSpy.mockRestore();
+  });
+
+  it.each([
+    [
+      "primary",
+      "var(--colorsActionDisabled500)",
+      "transparent",
+      "var(--colorsActionMajorYin030)",
```
edleeks87, line 206: Suggestion: as every `color` is the same value we could probably simplify this array by removing it and just asserting `color: var(--colorsActionMajorYin030)` in the test

src/components/button/button.spec.tsx
```diff
@@ -460,441 +362,361 @@ describe("Button", () => {
...
-              wrapper
-            );
-          });
-        }
+  it.each([
+    [
+      "primary",
+      "var(--colorsActionDisabled500)",
+      "transparent",
+      "var(--colorsActionMajorYin030)",
```
edleeks87, line 405: Same suggestion as above wrt `color`

src/components/button/button.spec.tsx
edleeks87, line 1: praise: good work removing `react-test-renderer` and flattening out some of these tests 👏

src/components/button/button.spec.tsx
```diff
@@ -199,39 +199,13 @@ describe("Button", () => {
...
-      "transparent",
-      "transparent",
-      "var(--colorsActionMajorYin030)",
-    ],
-    [
-      "dashed",
-      "transparent",
-      "var(--colorsActionDisabled500)",
-      "var(--colorsActionMinorYin030)",
-    ],
```
Parsium: Decided to remove `dashed` cases, given the variant is now deprecated.

edleeks87 (dismissed):

src/components/button-minor/button-minor-test.stories.tsx
```diff
@@ -10,7 +10,7 @@ import {
 
 export default {
   title: "Button Minor/Test",
-  includeStories: ["DefaultStory"],
+  excludeStories: ["ButtonMinorCustom", "ButtonMinorDifferentTypes"],
```
tomdavies73, line 13: **question**: what was the reason for excluding these stories? The `ButtonMinorDifferentTypes` story especially seems useful to keep

Parsium: Good point 👍🏼 These stories appear to be old artifacts for our Cypress/Playwright tests.

Before we migrated all the Cypress tests to TypeScript, we decided to temporarily place the mock components within the `test-stories.tsx` files so we could type-check them. We also have these mock components defined in the `components.test-pw.tsx` file.

If having these mock components as Storybook test stories would be valuable, I'm happy to include them again 😄

Parsium:
Updated changes to `ButtonToggleGroup` after rebasing with `master`. `aria-describedby` is now placed on the group container, rather than passed to each toggle button.

edleeks87 (approved):

tomdavies73 (dismissed):

tomdavies73 (dismissed):

edleeks87 (approved):

divyajindel (approved):

PR #6602, opened 2024-02-22
chore: tree-shake date-fns package

JinCoreana:
### Proposed behaviour
Tree-shake `dates-fns` and reduce bundle size to **252KB**

## AFTER
Screenshot 2024-02-23 at 09 09 05

Screenshot 2024-02-23 at 09 09 17

### Current behaviour

`dates-fns` includes methods that are not in use to the bundle increasing size to **1.23MB**
Open issue: https://github.com/date-fns/date-fns/issues/3004

## BEFORE
Screenshot 2024-02-22 at 11 15 16

Screenshot 2024-02-22 at 13 13 10

### Checklist

#### QA


src/locales/date-fns-locales/index.ts
edleeks87, line 1: Comment: Carbon is open source, we can't limit these locales here imo, if we did it would need to be a breaking change as well. How much would reverting this file affect the reduction in bundle size out of interest?

JinCoreana: Thank you for the comment.
Now I will check.

- solely limiting methods import? 
- downgrading or upgrading `date-fns`?
- lazy-loading `date-fns`?
- consumer side webpack configuration?

JinCoreana: Hi @edleeks87 
I've looked into your comment about the date locales logic. It seems that Carbon only requires `en-GB` as the default, which is set in the `I18nProvider`. If users need different date locales, they'll have to install `date-fns` and pass is as `dateFnsLocale` of `locale` prop to the `I18nProvider`.

Screenshot 2024-02-23 at 03 54 04

I checked if there is any logic for users to simply pass a locale code, and Carbon automatically uses the related date locale which then would involve bundling all date locales, but I couldn't find anything like that. I understand the failing part of the Playwright tests is for testing the locale that consumer passes works correctly so I can import the whole `date-fns` in testing files as it does not seem to increase the bundle size. Could you confirm if I'm missing something?

edleeks87: Hey @JinCoreana thanks for the detailed investigation. We export the locales via Carbon so that projects using it don't also require having date-fns installed. The easiest win here might be to move this out of carbon, add `date-fns` as a peer dependency and then require consuming projects to import and use the locales they want. We would need to do it as a breaking change so my preference is to leave this file for now and discuss it with the Carbon team as there may well be other changes we'd like to make and it would be better for all that to be bundled together as one new major version imo

AlexKryzh: Thanks @edleeks87 , this was a Miro's team spike ticket of 2-3 days to investigate the best solution. 
Can you confirm that Carbon Team prefer to do a refactor? 
Just for extra info: our next step was to investigate to lazy load this locations.

AlexKryzh: @edleeks87 will discuss this changes with the Carbon Team, probably it will be part of breaking change Carbon update.

edleeks87, line 1: comment: I'm happy with this change. We'll need to do it as a breaking change though as anyone who's imported from this file will obviously need to update and install `date-fns`

edleeks87 (commented):
If we're no longer going to re-export the `date-fns` locales we should add it as a [peerDependency](https://github.com/Sage/carbon/blob/8a06ef5cbd5663ba27e5f03a64e89b427c2acdc8/package.json#L54) so consumers have it installed in their own node_modules. I think the commit message should look something like this as well:

```
fix: ensure that tree-shaking works with date-fns package 

Updates the import paths for `Date` utils to allow tree-shaking to work and a reduction in bundle size.
    
BREAKING CHANGE: Carbon no longer re-exports the `date-fns` locales, consuming projects will need to install the package and use them directly from there.

closes: SBS-84395

nineteen88:
I am content with the changes proposed. I will approve it once we've addressed Ed's comments as I think it's important to address.

JinCoreana:
Hello, I am back and caught up with @AlexKryzh
According to his conversation with @edleeks8, this PR is closed as Carbon team will take it over and address this breaking changes. Thank you!
cc: @nineteen88

nineteen88:
@JinCoreana This has now been merged in the following PR: https://github.com/Sage/carbon/pull/6713

PR #6605, opened 2024-02-26
fix(tile): ensure borderVariant overrides border color associated to variant prop

robinzigmond:
When a consumer provides both `variant` and `borderVariant` to Tile, the `borderVariant` should determine what color the border is. At present the `variant` overrides this for the `active` and `grey` variants which have their own default border color.

### Proposed behaviour

When a `borderVariant` prop is specified, that should determine the border-color of the tile, no matter which `variant` is used for it.

### Current behaviour

When the `variant` is "active" or "grey" and a `borderVariant` prop is specified, the borderVariant` is ignored and the variant's default border color is used instead..

### Checklist

#### QA

### Testing instructions

For manual testing, use the "default" test story - the controls have been updated to allow any variant including "grey" to be selected, and for the `borderVariant` to also be specified.


src/components/tile/tile.pw.tsx
```diff
@@ -277,6 +277,28 @@ test.describe("Tile component", () => {
     });
   });
 
+  borderVariants
```
Parsium, line 280: **suggestion**: given that we are validating the styling when the `borderVariant`and `variant` props are set (in tests on lines 90 and 266), I'm thinking about the additional value brought in by this parameterised test, apart from ensuring the previous override behaviour doesn't reoccur.

As an alternative, what do you think about modifying the test on line 90 to remove checking the border colour? This way, we could avoid redundant verification of the border colour.

src/components/tile/tile-test.stories.tsx
```diff
@@ -27,6 +31,12 @@ export default {
         type: "select",
       },
     },
+    borderVariant: {
```
DipperTheDan, line 34: **praise(non-blocking):** Good idea adding this for our lovely QA's.

Parsium (approved):

DipperTheDan (approved):

stephenogorman (approved):

Issue #4452, opened 2021-10-04
Missing landmark on select options


maspychaj-sage:
### Current behaviour

After opening select and checking with axe plugin it fails `All page content should be contained by landmarks` rule:

![image](https://user-images.githubusercontent.com/75613789/135826805-44bcf22f-23db-4bf4-8ebb-19af7b5dc596.png)

### Expected behaviour

After opening select it should not fail `All page content should be contained by landmarks` rule.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/festive-panka-oi2mw?file=/src/index.js

### JIRA Ticket (Sage Only)

RIV-527

### Carbon Version

87.0.1

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

MacOS, Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

jamime:
@tempertemper do you have any resources or advice on what landmarks to add to this example?

The issue in this example is because we use a `Portal` which is rendered outside of the application stacking context, so it will be outside of the `role=main` landmark.

Thanks,
Jamie

tempertemper:
Probably not the answer you'll be looking for, but I don't think this should be given a landmark…

Giving the dropdown it its own landmark wouldn't quite make sense because it's not an individual, identifiable page landmark like a header, footer, call to action, etc. – it's part of a form control.

Also, it's content that you can reach with a screen reader without removing focus from the input, but it's after the second 'Hello Carbon' button in the source order, which is a bit disorienting, especially when there's a lot more content between the form control and the bottom of the page.

So the ideal fix for this would be to insert it alongside the text `<input>` that triggers it in the DOM. That way it:

- would be contained by the same `main` landmark, or whatever sub-landmark the whole form's a part of
- will appear in context for screen reader users who manage to read 'into' the dropdown itself, leaving focus on the input

msinga:
This was raised as part of https://jira.sage.com/browse/RIV-436 - ? should we be fixing this issue @maspychaj-sage  before we closed the RIV-436 ticket cc @chriswaymansage 

Newbie not sure on the process of closing JIRA stories with outstanding issues 😉

ClaireF-Sage:
Corresponding RIV-527 ticket

mkrds:
[...]

PR #6609, opened 2024-02-27
fix(select): render select list in DOM order rather than in a portal

### LINKED ISSUE #5171: VO key not moving directly to multi-select menu
### Current behaviour

I've given this another quick look and there's another issue with the DOM-disconnect between the trigger and the menu. When using VoiceOver, you move through the content with the VoiceOver key ('VO') (usually Control + Option) and the right/left arrow keys. When doing this, you have to move through the page content below the trigger box to get to the list of options. I added a bit more content to the page and [took a quick video](https://user-images.githubusercontent.com/2418781/168255619-3bb37e34-d915-4548-abcc-f8f010a87bd7.mov) so you can see it in action. Here's what I did:

Opened the menu by pressing the a key on the Select
Moved to the next bit of content with VO+→
Read through the content I'd added under the Select (hidden by the dropdown menu)
Got to the menu once I reached it in the DOM
Moving the menu HTML to directly under the Select trigger would fix this behaviour, as well as the landmark issue. Might be removing the portal functionality may be necessary from a screen reader usability perspective.

### Expected behaviour

VoiceOver key should navigate to the open select menu

https://carbon.sage.com/?path=/docs/select-multiselect--default-story

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

106.6.5

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

robinzigmond:
The select list is no longer rendered inside a portal, but after the select input in DOM order. This ensures that screenreaders will move directly through the list options when focus is on the input and the user steps through the following content.

BREAKING CHANGE: The `disablePortal` prop has been removed from all 3 Select variants. It is no longer needed as the component now always behaves as it did before with this prop added.

fix #5171

### Proposed behaviour

When the Select input is focused with the select list open, and a screenreader is told to move to the "next" element (Control+Option+RightArrow is the default on VoiceOver), it should read through the option list, no matter what content may occur below/after the select.

### Current behaviour

Using the above-mentioned key combination moves to any content after the select, only coming to the select list and its options after reaching the bottom of the page.

### Checklist

#### QA

### Additional context

This fix should also remove the Axe error about content not being contained in landmarks (see #4452). To see this, you need to run Axe with "best practices" turned ON. Note that on Storybook Select examples this will still bring up errors as we do not put the select inside an element with a landmark role, but if a `Select` is rendered inside eg a `main` HTML element, previously this error would still come up when the select list was open, and now it will not.

See @tempertemper 's comments on the just-linked issue, which suggested the fix that has been implemented in this PR.

### Testing instructions

Can be tested with screenreaders on the storybook docs for each of the 3 selects - test on the "docs" tab rather than "canvas" to ensure there is content between the Select being tested and the bottom of the page.

Test with NVDA on Windows as well as VoiceOver on Mac.

Test Select (all 3 varieties) for any regressions.


src/components/select/filterable-select/filterable-select.spec.tsx
```diff
@@ -1137,9 +1132,9 @@ describe("FilterableSelect", () => {
       });
 
       it("should not reopen the SelectList when a user selects and Option by clicking", () => {
-        const container = document.createElement("div");
```
Parsium, line 1140: **suggestion(non-blocking)**: what do you think about removing `enzymeContainer` attachment code here, given we have delated this to the setup/teardown hooks on lines 54 and 60?

robinzigmond: good spot that this isn't needed any more. I've removed it and cleaned it up, along with a few other instances of the same in the tests for all 3 select components - all tests still pass 👍

edleeks87 (approved):

harpalsingh:
Spoke with Martin and confirmed how it now renders should solve the issue and screen readers appear to work correctly.

sianford (approved):

PR #6611, opened 2024-02-28
feat(loader-spinner): add new component

tomdavies73:
### Proposed behaviour

Adds a new component for consumers to demonstrate a loading state to end users. Can display a predictable or unpredictable wait time.

### Current behaviour

Currently, there is only the `Loader` and `LoaderBar` components, neither of which provide means to display a spinner loading pattern.

### Checklist

#### QA


src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+   * however there will still be sufficient contrast between them
+   */
+  color?: LoaderSpinnerColors;
+  /**
+   * The `labelColor` prop can be used to change the font color of the
+   * rendered label
+   */
+  labelColor?: "dark" | "light";
+  /** If set to `false` all motion will be suspended */
+  motion?: boolean;
```
edleeks87, line 47: comment (non-blocking): we could name this a bit more intuitively for example `hasMotion` etc

src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+import {
+  StyledSpinnerWrapper,
+  StyledLabel,
+  StyledSpinnerCircleSvg,
+} from "./loader-spinner.style";
+import useLocale from "../../hooks/__internal__/useLocale";
+import useMediaQuery from "../../hooks/useMediaQuery";
+import Typography from "../typography";
+
+export type LoaderSpinnerSizes = "XS" | "S" | "M" | "L" | "XL";
```
edleeks87, line 15: comment: I see why you've gone with the short-hand here as we have a few components like it in the codebase. However, I think we should go with the long form values here as we've adopted that more often in other components including [loader](https://github.com/Sage/carbon/blob/ab116604a60104ca15241750558a211f4c326d0b/src/components/loader/loader-square.style.ts). I think we should raise a ticket to align the others that have this approach currently etc

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -0,0 +1,244 @@
...
+    it("the spinner label renders", () => {
+      const { container } = render(<LoaderSpinner />);
+      expect(
+        container.querySelector('[data-element="spinner-label"]')
+      ).toHaveTextContent("Loading...");
+    });
+  });
+
+  describe("when custom props are passed", () => {
+    it.each(["XS", "S", "M", "L", "XL"] as LoaderSpinnerSizes[])(
```
edleeks87, line 39: suggestion (non-blocking): you could store this in a typed array and then re-use it rather than casting to the type etc

src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+   * The color prop can be used to change the color of the component.
+   * Typically both the outer and inner spinner will change color,
+   * however there will still be sufficient contrast between them
+   */
+  color?: LoaderSpinnerColors;
+  /**
+   * The `labelColor` prop can be used to change the font color of the
+   * rendered label
+   */
+  labelColor?: "dark" | "light";
```
edleeks87, line 45: question: is this something that we definitely want to act independent of color/ variant? The designs seem to suggest it's only `light` when in the `inverse` variant?

src/components/loader-spinner/loader-spinner.pw.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+      await expect(loaderSpinnerVisibleLabel(page)).toHaveCSS(
+        "color",
+        labelColorTokenValues
+      );
+    });
+  });
+});
+
+test.describe("Accessibility tests for StepFlow component", () => {
```
edleeks87, line 157:
```suggestion
test.describe("Accessibility tests for LoaderSpinner component", () => {
```

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+<Canvas>
+  <Story name="colors" story={stories.Colors}/>
+</Canvas>
+
+
+### Label color
+
+Please find an example of a spinner below when the `labelColor` prop is set to `"light"`. This will ensure proper contrast is 
+achieved on the label, when used in a dark background.
```
edleeks87, line 79:
```suggestion
achieved on the label, when used on a dark background.
```

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+<Canvas>
+  <Story name="animation time" story={stories.AnimationTime}/>
+</Canvas>
+
+## Props
+
+### Content
+
+<StyledSystemProps of={LoaderSpinner} margin noHeader />
```
edleeks87, line 116: Comment: we'll need to include the `TranslationKeys` table as well here

edleeks87 (commented):
Great work @tomdavies73, I know this wasn't the easiest component to get working, I've left some comments but they're all minor improvements, in terms of the functionality it looks to me like you've nailed it 👏

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,176 @@
...
+              : css`
+                  animation: ${tracked
+                      ? "trackedAnimation"
+                      : "untrackedAnimation"}
+                    ${animationTime}s cubic-bezier(0.2, 0.1, 0.8, 1) infinite;
+                `}
+          `
+        }
+      }
+    `}
```
Parsium, line 175: **suggestion**: separating the arc colour and movement animations as separate sequences should resolve the "jolt" issue when the animation is slowed down:

https://github.com/Sage/carbon/assets/18368713/033ca92c-a5b5-4147-97e2-85f082d30b7e

```suggestion
        ${
          motion &&
          css`
            animation-name: ${tracked
                ? "trackedAnimation"
                : "untrackedAnimation"},
              ${color === "gradient-grey" ||
              (color === "gradient-white" && "gradientAnimation")
                ? "gradientAnimation"
                : "none"};
            animation-duration: ${animationTime}s, 3s;
            animation-timing-function: cubic-bezier(0.2, 0.1, 0.8, 1);
            animation-iteration-count: infinite;
          `
        }
```

I'd suggest making the colour animation duration >=3 seconds as well, to reduce the number of flashes per second.

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,176 @@
...
+            transform: rotate(280deg);
+            stroke-dasharray: 100;
+          }
+
+          100% {
+            transform: rotate(360deg);
+            stroke-dasharray: 100;
+          }
+        }
+
```
Parsium, line 157: **suggestion**: what do you think about reducing the number of keyframes for each of the three animations, in order to improve performance slightly?:

```suggestion
        @keyframes gradientAnimation {
          0% {
            stroke: #00D639;
          }

          33% { 
            stroke: #11AFFF;
          }

          66% {
            stroke: #8F49FE;
          }
        };

      @keyframes trackedAnimation {
        from {
          stroke-dasharray: 100;
          stroke-dashoffset: 100;
        }

        to {
          stroke-dasharray: 100;
          stroke-dashoffset: 20;
        }
        };

        @keyframes untrackedAnimation {
          0% {
            transform: rotate(0deg);
            stroke-dasharray: 100;
          }

          40% {
            stroke-dasharray: 80;
          }

          80% {
            stroke-dasharray: 100;
          }

          100% {
            transform: rotate(360deg);
          }
        };
```

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,176 @@
...
+        stroke-linecap: round;
+
+        stroke-dasharray: 100px;
+        stroke-dashoffset: 80px;
+
+        transform-origin: 12px 12px 0px;
+
+        cx: 12px;
+        cy: 12px;
+        r: 10px;
```
Parsium, line 82: **suggestion**: making both the spinner circle and animated arc SVGs should resolve the issue of the arc not fully covering the spinner, as previously discussed externally: 

https://github.com/Sage/carbon/assets/18368713/b2370c89-e59d-4abc-a5ae-300da7f89371


```suggestion
      circle[data-element="outer-arc"] {
        fill: transparent;
        stroke-width: ${LOADER_SPINNER_SIZE_PARAMS[size].strokeWidth}px;
        stroke: ${calculateColors(false, color)};
        stroke-opacity: 0.5;
        cx: 12px;
        cy: 12px;
        r: 10px;
      }

      circle[data-element='spinner-circle'] {
        fill: transparent;
        stroke-width: ${LOADER_SPINNER_SIZE_PARAMS[size].strokeWidth + 0.1}px;
        stroke: ${calculateColors(true, color)}
        stroke-linecap: round;

        stroke-dasharray: 100px;
        stroke-dashoffset: 80px;

        transform-origin: 12px 12px 0px;
        cx: 12px;
        cy: 12px;
        r: 10px;
```

and in `loader-spinner.component.tsx`:

```tsx
<StyledSpinnerCircleSvg role="presentation" viewBox="0 0 24 24"> {...spinnerProps}>
  <circle data-element="outer-arc" />
  <circle data-element="spinner-circle" />
</StyledSpinnerCircleSvg>
```

src/components/loader-spinner/loader-spinner.style.ts
tomdavies73: Yeah, the jankiness is slightly more noticeable on the larger sizes, i'll use this approach to make it more seamless.

In terms of the default time - i'll see about maybe 1.5-2 seconds as we'd still like the animation to be fairly quick as its doubtful many loading times will reach 3 seconds - and we'd like end users to be able to see the full animation if possible.

src/components/loader-spinner/loader-spinner.pw.tsx
```diff
@@ -0,0 +1,223 @@
...
+      await mount(<LoaderSpinnerComponent size={spinnerSizes} />);
+
+      const wrapperShadow = await getStyle(
+        loaderSpinnerWrapper(page),
+        "box-shadow"
+      );
+
+      expect(wrapperShadow).toContain(
+        `${LOADER_SPINNER_SIZE_PARAMS[spinnerSizes].boxShadow}px`
+      );
```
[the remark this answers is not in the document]
tomdavies73: The `box-shadow` css value value includes more than just this, `toHaveCss` would fail as what we're testing against is not an exact match.

That's why I've went for the `toContain` approach here 👍

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -0,0 +1,244 @@
...
+      "when the 'size' prop is passed as `%s` the wrapper has the correct box shadow",
+      (sizes) => {
+        const { getByRole } = render(<LoaderSpinner size={sizes} />);
+        const wrapperElement = getByRole("status");
+        const computedStyle = window.getComputedStyle(wrapperElement);
+
+        expect(computedStyle.boxShadow).toContain(
+          `${LOADER_SPINNER_SIZE_PARAMS[sizes].boxShadow}px`
+        );
```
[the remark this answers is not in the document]
tomdavies73: Again, I don't believe what we're testing here is all of the css value - so we're using `toContain` over `toHaveStyle` - as `toContain` would fail

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+<Canvas>
+  <Story name="motion" story={stories.Motion}/>
+</Canvas>
+
+
+### Tracked
+
+When the `tracked` prop is set to `true`, the spinner will become tracked, for specific use cases where wait/loading times
+are predictable. The `tracked` prop cannot be true when the `gradient-grey` or `gradient-white` colors are set.
```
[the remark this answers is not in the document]
tomdavies73: Yeah, there aren't solid designs using a typical tracked animation type + @harpalsingh raised concerns about any sort of spinning gradient colour - as this may not be performative and technically feasible without the use of third party plug ins

src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <StyledSpinnerWrapper
+      size={size}
+      color={color}
+      role="status"
+      reduceMotion={reduceMotion}
+      {...tagComponent("loader-spinner", rest)}
+      {...rest}
+    >
```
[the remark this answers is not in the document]
tomdavies73: Yeah good shout

src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+      {reduceMotion ? (
+        spinnerLabel
+      ) : (
+        <>
+          <StyledSpinnerCircleSvg {...spinnerProps} viewBox="0 0 24 24">
```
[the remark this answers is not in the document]
tomdavies73: I think seeing as `spinnerProps` isn't public facing and only used internally in this file from our destructured props, this should be fine no?

Consumers shouldn't be able to pass any other props here

src/components/loader-spinner/loader-spinner.stories.mdx
tomdavies73: Yeah good spot

src/components/loader-spinner/loader-spinner.pw.tsx
tomdavies73: Guess you guys now know which components PW file structure I copied now :0

src/components/loader-spinner/loader-spinner.component.tsx
tomdavies73: It's a good point, however with the `gradient-white` variant, white font might also be needed. 

I'm happy to just inverse the label colour based on these two variants to be fair, as it would remove the need for another prop

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,176 @@
...
+  extends Pick<LoaderSpinnerProps, "size" | "color"> {
+  reduceMotion?: boolean;
+}
+
+export const StyledSpinnerWrapper = styled.div<StyledSpinnerWrapperProps>`
+  ${margin}
+  ${({ size, reduceMotion, color }) =>
+    size &&
+    css`
+      width: ${LOADER_SPINNER_SIZE_PARAMS[size].wrapperDimensions}px;
```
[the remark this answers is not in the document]
tomdavies73: Yeah this looks a lot better

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        transform-origin: 12px 12px 0px;
+
+        cx: 12px;
+        cy: 12px;
+        r: 10px;
+
+        transform: rotate(270deg);
+
+        @keyframes gradientAnimation {
```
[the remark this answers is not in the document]
tomdavies73: Yeah agreed, i'll do this based on this suggestion and @Parsium 's suggestion below

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+Please find an examples of all spinner sizes below. Ranging from sizes `XS` to `XL`.
+
+<Canvas>
+  <Story name="sizes" story={stories.Sizes}/>
+</Canvas>
+
+### Hide Spinner label
+
+When the `showSpinnerLabel` prop is set to `false` no visual label will be rendered underneath the spinner.
+A visually hidden label will still be available for assistive technologies. However, even with `hideSpinnerLabel`
```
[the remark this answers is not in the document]
tomdavies73: Yeah correct @Parsium - `hideSpinnerLabel` was the original prop name but i switched it around, and must have forgot to update a portion of this

src/components/loader-spinner/loader-spinner.component.tsx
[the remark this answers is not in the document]
Parsium: I was thinking more for us internally, just to catch any potential mistakes we may make 🙂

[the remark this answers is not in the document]
tomdavies73: Fairs mate, as it's only being passed to one component I'll pass the props individually

src/components/loader-spinner/loader-spinner.style.ts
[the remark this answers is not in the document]
tomdavies73: I had some issues with this as it introduced some type errors, specifically around using size, attrs on a div. 

`Property 'size' does not exist on type 'Omit<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "ref"> & { ref?: ((instance: HTMLDivElement | null) => void) | RefObject<...> | null | undefined; } & ThemeProps<...>'.`

As this comment is non blocking i'm just going to stick with the original approach

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -0,0 +1,290 @@
...
+    });
+
+    it("when 'showSpinnerLabel' is `false` the visible label is not rendered in the live region container", () => {
+      render(<LoaderSpinner showSpinnerLabel={false} />);
+      const wrapperElement = screen.getByRole("status");
+      const visibleLabelElement = screen.queryByTestId("visible-label");
+
+      expect(wrapperElement).toHaveTextContent("Loading...");
+      expect(visibleLabelElement).not.toBeInTheDocument();
+    });
```
Parsium, line 128: **comment**: thanks for amendment here @tomdavies73 👍🏼 

I just wanted to suggest an alternative approach for us to explore in future PRs. Instead of using a test-id approach, we could assert that the label inside the live region is present in the document but visually hidden. This ensures the intended behaviour is directly tested while making the test resistant to markup changes. What do you think?

```tsx
import { render, screen, within } from "@testing-library/react";

it("when 'showSpinnerLabel' is `false`, loading label has visually hidden styling", () => {
  render(<LoaderSpinner showSpinnerLabel={false} />);
  const wrapperElement = screen.getByRole("status");
  const label = within(wrapperElement).getByText("Loading..."); // query throws error if label is not in the document
  
  expect(label).toHaveStyle({
    border: "0",
    height: "1px",
    margin: "-1px",
    overflow: "hidden",
    padding: "0",
    position: "absolute",
    width: "1px",
    "white-space": "nowrap",
  });
});
```

src/components/loader-spinner/loader-spinner.spec.tsx
Parsium, line 1: **praise**: thorough tests as always @tomdavies73 👍🏼

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+loading label will be displayed.
+
+<Canvas>
+  <Story name="default" story={stories.DefaultStory}/>
+</Canvas>
+
+### Sizes
+
+Please find an examples of all spinner sizes below. Ranging from sizes `extra-small` to `extra-large`.
+
```
Parsium, line 52: **typo**: 

```suggestion
Please find examples of all spinner sizes below. Ranging from sizes `extra-small` to `extra-large`.
```

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+<Canvas>
+  <Story name="show spinner label" story={stories.ShowSpinnerLabel}/>
+</Canvas>
+
+### Variants
+
+Please find an examples of all spinner variants below. Shown against the ligher background are the `action` and `neutral` variants.
+Against the darker background are the `inverse`, `gradient-grey` and `gradient-white` variants.
+
```
Parsium, line 71: **typo**:

```suggestion
Please find examples of all spinner variants below. Shown against the lighter background are the `action` and `neutral` variants.
Against the darker background are the `inverse`, `gradient-grey` and `gradient-white` variants.
```

src/components/loader-spinner/loader-spinner.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+  translationData={[
+    {
+      name: "loaderSpinner.loading",
+      description: "The spinner label which is shown on the screen underneath the spinner, solely displayed when users prefer reduced motion and inside"+
+      " the visually hidden label if rendered.",
+      type: "func",
+      returnType: "string",
+    }
+  ]}
+/>
```
Parsium, line 131: **suggestion**: what do you think about tweaking the page structure slightly, so the translations table is placed in its own section?:

```mdx
## Contents

- [Quick Start](#quick-start)
- [Examples](#examples)
- [Props](#props)
- [Translation keys](#translation-keys)

## Quick Start

## Examples

## Props

## Translation keys
The following keys are available to override the translations for this component by passing in a custom locale object to the i18nProvider.

<TranslationKeysTable />
```

Parsium (commented):
Thanks for the update @tomdavies73 👍🏼 Just a couple of minor comments from me

src/components/loader-spinner/loader-spinner.style.ts
```diff
@@ -0,0 +1,152 @@
...
+};
+
+interface StyledSpinnerWrapperProps
+  extends Pick<LoaderSpinnerProps, "size" | "variant"> {
+  reduceMotion?: boolean;
+  size: LoaderSpinnerSizes;
+}
+
+export const StyledSpinnerWrapper = styled.div<
+  Pick<StyledSpinnerWrapperProps, "size">
```
edleeks87, line 36: comment: it seems a bit odd to define an interface for this and then only pick part of it, I think `reduceMotion` isn't used anywhere anymore so it can just be removed and you can use the full interface etc

src/components/loader-spinner/loader-spinner.style.ts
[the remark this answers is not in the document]
edleeks87: Yeah sorry I missed the type for `size`, the below would work (again non-blocking)
```
export const StyledSpinnerWrapper = styled.div.attrs(({size}: Pick<StyledSpinnerWrapperProps, "size">) => ({
  dimensions: `${LOADER_SPINNER_SIZE_PARAMS[size].wrapperDimensions}px`,
}))<StyledSpinnerWrapperProps>`
  ${({dimensions}) => css`
     width: ${dimensions};
      height: ${dimensions};
      min-width: ${dimensions};
      min-height: ${dimensions};
  `}
`;

File: src/components/loader-spinner/loader-spinner.config.ts [hunk 24]
```diff
@@ -0,0 +1,52 @@
+type LoaderSpinnerSizes =
```
edleeks87, line 1: suggestion (non-blocking): you can derive the types from the array here if you want

```
const LOADER_SPINNER_VARIANTS = [
  "action",
  "neutral",
  "inverse",
  "gradient-grey",
  "gradient-white",
] as const;

type LoaderSpinnerVariants = typeof LOADER_SPINNER_VARIANTS[number];

const LOADER_SPINNER_SIZES = [
  "extra-small",
  "small",
  "medium",
  "large",
  "extra-large",
] as const;

type LoaderSpinnerSizes = typeof LOADER_SPINNER_SIZES[number];

File: src/components/typography/typography.component.tsx [hunk 25]
```diff
@@ -85,6 +85,8 @@ export interface TypographyProps extends SpaceProps, TagProps {
   isDisabled?: boolean;
   /** @private @ignore Set whether the component should be recognised by assistive technologies */
   "aria-hidden"?: "true" | "false";
+  /** @private @ignore Attribute aids in targeting and selecting specific elements within the component during testing */
+  "data-testid"?: string;
```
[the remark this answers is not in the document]
edleeks87: suggestion: does this not work without it? `data-` props should be implicitly accepted so maybe we make a small amendment to `tagComponent` to add support this (also would work with explicit if here if preferred):

```
interface RestProps {
  [restKeys: string]: any;
}

export interface TagProps {
  /** @private @ignore Identifier used for testing purposes, applied to the root element of the component. */
  "data-component"?: string;
  /** Identifier used for testing purposes, applied to the root element of the component. */
  "data-element"?: string;
  /** Identifier used for testing purposes, applied to the root element of the component. */
  "data-role"?: string;
  /** @private @ignore  Identifier used for testing purposes, applied to the root element of the component. */
  "data-testid"?: string;
}

/**
 * Builds props object containing top level data tags
 */
function tagComponent(
  componentName: string | undefined,
  props: TagProps & RestProps
): TagProps {
  const tagProps: TagProps = {
    "data-component": componentName,
    "data-element": undefined,
    "data-role": undefined,
    "data-testid": undefined,
  };

  Object.entries(props).forEach(([key, value]) => {
    if (Object.keys(tagProps).includes(key)) {
      const tagPropName = key as keyof typeof tagProps;
      tagProps[tagPropName] = value as string;
    }
  })

  return tagProps;
}

export default tagComponent;

File: src/components/loader-spinner/loader-spinner.spec.tsx (hunk 19 above)
Comment (tomdavies73) [reply src/components/loader-spinner/loader-spinner.spec.tsx:128]: Yeah looks good, I don't mind using this approach here either but happy to keep the current approach too

File: src/components/loader-spinner/loader-spinner.style.ts (hunk 23 above)
Comment (tomdavies73) [reply src/components/loader-spinner/loader-spinner.style.ts:36]: Yeah good point, `reduceMotion` was originally added to change styling on the wrapper. As we're not applying any major styles to the wrapper and just the circle svgs we can remove it, and just pick straight from `LoaderspinnerProps` opposed to making a whole new interface

**Review by Parsium — APPROVED** [review]

**Review by edleeks87 — APPROVED** [review]

File: src/components/loader-spinner/loader-spinner.mdx [hunk 26]
```diff
@@ -0,0 +1,117 @@
...
+## Props
+
+### LoaderSpinner
+
+<ArgTypes of={LoaderSpinnerStories} />
+
+## Translation keys
+
+The following keys are available to override the translations for this component by passing in a custom locale object to the
+[i18nProvider](https://carbon.sage.com/?path=/story/documentation-i18n--page).
```
Parsium, line 105: **suggestion**: unfortunately this link won't work after upgrading to Storybook v7.0. This should do the trick though:

```suggestion
[i18nProvider](../?path=/docs/documentation-i18n--docs).
```

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -0,0 +1,341 @@
...
+      });
+    });
+
+    it.each([
+      "small",
+      "medium",
+      "large",
+      "extra-large",
+    ] as LoaderSpinnerSizes[])(
+      "when the 'size' prop is passed as `%s` the label has the correct width has the correct width",
```
Parsium, line 152: **typo**:
```suggestion
      "when the 'size' prop is passed as `%s` the label has the correct width",
```

src/components/loader-spinner/loader-spinner.pw.tsx
Parsium, line 1: **suggestion(non-blocking)**: there may be some benefit in combining some of the separate tests into one, just so the context of the current design requirements is a bit clearer to any future contributors, making maintaining the tests easier. What do you think?

```ts
test("when the 'size' prop is 'extra-small', spinner and label are both displayed on the same row", () => {
    // check flexbox positioning of elements, margins, etc.
});

["small", "medium", "large", "extra-large" as const].forEach((size) => {
   test(`when the 'size' prop is '${size}', spinner, spinner and label are vertically stacked`, () => {
       // check flexbox positioning of elements, margins, etc.
   });   
});
```

src/components/loader-spinner/loader-spinner.pw.tsx
```diff
@@ -0,0 +1,286 @@
...
+      await expect(loaderSpinnerSvg(page)).toHaveCSS(
+        "height",
+        `${LOADER_SPINNER_SIZE_PARAMS[spinnerSizes].wrapperDimensions}px`
+      );
+      await expect(loaderSpinnerSvg(page)).toHaveCSS(
+        "width",
+        `${LOADER_SPINNER_SIZE_PARAMS[spinnerSizes].wrapperDimensions}px`
+      );
+    });
+  });
```
Parsium, line 61: **suggestion(non-blocking)**: given we are asserting the `min-height` and `min-width` in another test, we probably don't need to assert the `height` and `width` as well. What do you think?

```suggestion
```

src/components/loader-spinner/loader-spinner.mdx
tomdavies73: Last time I'll miss this, I promise 😅

src/components/loader-spinner/loader-spinner.pw.tsx
tomdavies73: I get what you mean, but I think we may encounter issues down the line where specific tests fail, and it may take slightly more effort to debug opposed to one specific test failing where the test description clearly describes what is being tested etc.

Parsium (approved):

edleeks87 (approved):

PR #6623, opened 2024-03-07
fix(select): objects as value fix

### LINKED ISSUE #6600: Select (With Object): First open, the value is not selected
### Description

The first time that we open the Select, we cannot see the option selected.
By the way, do we really need to use a useRef in this case ?

### Reproduction

https://carbon.sage.com/?path=/docs/select--with-object-as-value

### Steps to reproduce

After loading the page:
- Open the select
- No value are loaded (Normally Green need to be selected)

_No response_

### Suggested solution

_No response_

### Carbon version

126.5.0

### Design tokens version

4.23.0

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #6600

### Proposed behaviour

When values for options in the `Select` component are set to objects, `select-list` is not able to compare object values and does not highlight the selected option as it does not find it in the children values. Added a shallow equality check to be able to compare property values instead of object references. 

### Current behaviour

'Object as value' story for `Select` components does not highlight the preselected value. 

### Checklist

#### QA

### DIFF
--- a/src/components/select/filterable-select/filterable-select.pw.tsx
+++ b/src/components/select/filterable-select/filterable-select.pw.tsx
@@ -750,7 +750,24 @@ test.describe("FilterableSelect component", () => {
     await expect(selectListWrapper(page)).toBeVisible();
   });
 
-  test("should open correct list and select one when an object is already set as a value", async ({
+  test("should have correct option highlighted when select list is opened and value is an object", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<FilterableSelectObjectAsValueComponent />);
+
+    const inputElement = getDataElementByValue(page, "input");
+    await expect(inputElement).toHaveValue("Green");
+    await dropdownButton(page).click();
+    const optionElement = selectOptionByText(page, "Green");
+
+    await expect(optionElement).toHaveCSS(
+      "background-color",
+      "rgb(153, 173, 183)"
+    );
+  });
+
+  test("should select another option when one is already selected and the value is an object", async ({
     mount,
     page,
   }) => {
--- a/src/components/select/select-list/select-list.component.tsx
+++ b/src/components/select/select-list/select-list.component.tsx
@@ -187,12 +187,29 @@ const SelectList = React.forwardRef(
       children,
     ]) as React.ReactElement<OptionProps>[];
 
+    // check if object values are equal
+    function shallowEqual(
+      objA: Record<string, unknown>,
+      objB: Record<string, unknown>
+    ) {
+      const keysA = Object.keys(objA);
+
+      return keysA.every((key) => objA[key] === objB[key]);
+    }
+
     const getIndexOfMatch = useCallback(
       (valueToMatch) => {
-        return childrenList.findIndex(
-          (child) =>
+        return childrenList.findIndex((child) => {
+          if (child.props.value && typeof valueToMatch === "object") {
+            return shallowEqual(
+              child.props.value as Record<string, unknown>,
+              valueToMatch
+            );
+          }
+          return (
             React.isValidElement(child) && child.props.value === valueToMatch
-        );
+          );
+        });
       },
       [childrenList]
     );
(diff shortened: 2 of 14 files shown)


src/components/select/select-list/select-list.component.tsx
```diff
@@ -189,12 +189,33 @@ const SelectList = React.forwardRef(
...
+    ) {
+      const keysA = Object.keys(objA);
+
+      for (const key of keysA) {
+        if (objA[key] !== objB[key]) {
+          return false;
+        }
+      }
+      return true;
+    }
```
DipperTheDan, line 204: **suggestion(non-blocking):** I _think_ you can do this with the `.every()` JS method. I tried it locally and it seemed to still work and all the tests (unit and PW) passed. I'm happy if you want to leave it the way you've done it as it works and it's easy to read. This is what I did though:

```
    function shallowEqual(
      objA: Record<string, unknown>,
      objB: Record<string, unknown>
    ): boolean {
      const keysA = Object.keys(objA);

      return keysA.every((key) => objA[key] === objB[key]);
    }
```

src/components/select/select-list/select-list.spec.tsx
```diff
@@ -721,7 +780,7 @@ describe("SelectList", () => {
       );
     });
 
-    it("and is in multiColum mode, then a Loader Component should be rendered", () => {
+    it("and is in multiColumn mode, then a Loader Component should be rendered", () => {
```
DipperTheDan, line 783: **praise:** Well spotted.

src/components/select/simple-select/simple-select.pw.tsx
```diff
@@ -586,7 +586,23 @@ test.describe("SimpleSelect component", () => {
...
+    mount,
+    page,
+  }) => {
+    await mount(<SimpleSelectObjectAsValueComponent />);
+
+    const inputElement = getDataElementByValue(page, "input");
+    await expect(inputElement).toHaveValue("Green");
+    await selectText(page).click();
+    const optionElement = selectOptionByText(page, "Green");
+    await expect(optionElement).toHaveCSS(
```
DipperTheDan, line 599: **nitpick:** We've been putting a line break between the body of the test and the `expect()`.

```suggestion
    const optionElement = selectOptionByText(page, "Green");
    
    await expect(optionElement).toHaveCSS(
```

src/components/select/filterable-select/filterable-select.pw.tsx
```diff
@@ -720,7 +720,23 @@ test.describe("FilterableSelect component", () => {
...
+    mount,
+    page,
+  }) => {
+    await mount(<FilterableSelectObjectAsValueComponent />);
+
+    const inputElement = getDataElementByValue(page, "input");
+    await expect(inputElement).toHaveValue("Green");
+    await dropdownButton(page).click();
+    const optionElement = selectOptionByText(page, "Green");
+    await expect(optionElement).toHaveCSS(
```
DipperTheDan, line 733: **nitpick:** We've been putting a line break between the body of the test and the `expect()`.

```suggestion
    const optionElement = selectOptionByText(page, "Green");
    
    await expect(optionElement).toHaveCSS(
```

src/components/select/select-list/select-list.component.tsx
edleeks87: you can short circuit it faster if you invert the logic and use `some`:

```
 function shallowEqual(
      objA: Record<string, unknown>,
      objB: Record<string, unknown>
    ): boolean {
      const keysA = Object.keys(objA);

      return !keysA.some((key) => objA[key] !== objB[key]);
    }

edleeks87 (dismissed):

DipperTheDan (approved):

ZhuoyuJin (dismissed):

edleeks87 (approved):

DipperTheDan (approved):

ZhuoyuJin (approved):

PR #6624, opened 2024-03-08
fix(submenu): ensure last child in segment title has correct styling - FE-6308

### LINKED ISSUE #6488: MenuSegmentTitle
### Current behaviour

`MenuSegmentTitle` applies rounding to the last child, which is not correct when the `MenuSegmentTitle` component is not the last item in the menu.

### Expected behaviour

No rounding to occur where other elements follow below `MenuSegmentTitle`.

https://codesandbox.io/p/sandbox/menusegmenttitle-border-radius-4tqhzp

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

124.3.0

### Design Tokens Version

_No response_

Chrome

MacOS

_No response_

### Confidentiality

DipperTheDan:
The last child in the segment title should not have rounded corners.

fixes #6488

### Proposed behaviour

Screenshot 2024-03-08 at 11 01 57

Screenshot 2024-03-08 at 11 03 47

### Current behaviour

border-radius is applied to all last children of the segment title. 

### Checklist

#### QA

### Additional context

I have opted to do the styling test in Playwright alone instead of doubling up with unit and automation. The reason for this is we already have a humongous amount of unit tests for `Menu` and it's associated components. Doing the tests in Playwright is also much simpler and easier to maintain. With a component as complex as `Menu`, I felt this way benefited future development work but I am happy to add unit tests if we feel strongly that they should be included. 

### Testing instructions

- No visual regressions in the other menu examples
- New test story `MenuWithTwoSegments` should have the correct styling applied to the last menu items in the submenus.

### DIFF
--- a/src/components/menu/__internal__/submenu/submenu.component.tsx
+++ b/src/components/menu/__internal__/submenu/submenu.component.tsx
@@ -98,7 +98,9 @@ const Submenu = React.forwardRef<
     }: SubmenuProps,
     ref
   ) => {
-    const submenuRef = useRef<HTMLUListElement & HTMLDivElement>(null);
+    const [submenuRef, setSubmenuRef] = useState<
+      HTMLUListElement | HTMLDivElement | null
+    >(null);
     const submenuId = useRef(guid());
     const menuContext = useContext(MenuContext);
     const {
@@ -111,8 +113,10 @@ const Submenu = React.forwardRef<
     const [submenuFocusId, setSubmenuFocusId] = useState<string | null>(null);
     const [submenuItemIds, setSubmenuItemIds] = useState<(string | null)[]>([]);
     const [characterString, setCharacterString] = useState("");
+    const [applyFocusRadius, setApplyFocusRadius] = useState<boolean>(false);
     const shiftTabPressed = useRef(false);
     const focusFirstMenuItemOnOpen = useRef(false);
+
     const numberOfChildren = submenuItemIds.length;
 
     const { submenuMaxHeight } = useContext(FixedNavigationBarContext);
@@ -120,12 +124,12 @@ const Submenu = React.forwardRef<
     const onSubmenuOpen = useStableCallback(onSubmenuOpenProp);
 
     const blockIndex = useMemo(() => {
-      const items = submenuRef.current?.querySelectorAll(BLOCK_INDEX_SELECTOR);
+      const items = submenuRef?.querySelectorAll(BLOCK_INDEX_SELECTOR);
 
       if (items && submenuOpen && numberOfChildren) {
         const childrenArray = Array.from(items);
 
-        const scrollableBlock = submenuRef.current?.querySelector(
+        const scrollableBlock = submenuRef?.querySelector(
           `[data-component='${SCROLLABLE_BLOCK}']`
         );
 
@@ -141,7 +145,7 @@ const Submenu = React.forwardRef<
       }
 
       return -1;
-    }, [submenuOpen, numberOfChildren]);
+    }, [submenuOpen, numberOfChildren, submenuRef]);
 
     const characterTimer = useRef<NodeJS.Timeout | null>(null);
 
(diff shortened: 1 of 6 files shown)


src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -393,6 +396,46 @@ const Submenu = React.forwardRef<
       }
     }, [characterString, submenuItemIds]);
 
+    useEffect(() => {
```
edleeks87, line 399: comment: we could probably add some short circuiting here, for example if `segmentBlocks` is falsy there's no need to make any further queries as state will be set to false no matter what.

src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -111,6 +111,9 @@ const Submenu = React.forwardRef<
     const [submenuFocusId, setSubmenuFocusId] = useState<string | null>(null);
     const [submenuItemIds, setSubmenuItemIds] = useState<(string | null)[]>([]);
     const [characterString, setCharacterString] = useState("");
+    const [applyFocusRadius, setApplyFocusRadius] = useState<
```
edleeks87, line 114: suggestion: I would just make this a boolean type and coerce the value with `!!` when it's set etc

src/components/menu/__internal__/submenu/submenu.style.ts
```diff
@@ -23,6 +24,7 @@ interface StyledSubmenuProps
     Pick<SubmenuProps, "variant"> {
   submenuDirection?: string;
   maxHeight?: string;
+  applyFocusRadiusStyling?: boolean;
```
edleeks87, line 27:
```suggestion
  applyFocusRadiusStyling: boolean;
```

If we always coerce the value to a bool it doesn't need to be optional here

src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -393,6 +396,46 @@ const Submenu = React.forwardRef<
...
+          menuItemElementsInSegmentBlock.length - 1
+        ];
+
+      // Check if the last item in the segment block is the same as the last MenuItem in the submenu
+      const isLastItemInSubmenuTheSameLastItemInSegmentBlock =
+        lastMenuItemInSegmentBlock &&
+        lastMenuItem &&
+        lastMenuItemInSegmentBlock === lastMenuItem;
+
+      setApplyFocusRadius(isLastItemInSubmenuTheSameLastItemInSegmentBlock);
```
edleeks87, line 436:
```suggestion
      setApplyFocusRadius(!!isLastItemInSubmenuTheSameLastItemInSegmentBlock);
```

personal opinion is that this variable name is a bit verbose but I'm fine with leaving it if the additional intuitiveness is preferred here, maybe something like `isLastItemInSegmentBlock`

edleeks87 (commented):
This looks like it does address the issue when using the composition described in the issue (first pic). However, here's still a style regression I think we need to address here when a scrollable block is used and there's other menu items rendered after (second pic). I think if possible it's better to fix it all in this PR but if it proves quite complex maybe worth discussing with @harpalsingh and moving it to a new ticket.

image


image

src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -168,70 +170,48 @@ const Submenu = React.forwardRef<
...
+        // Finds all ul elements that are not submenus
+        const ulElements = Array.from(
+          submenuRef?.querySelectorAll("ul:not([data-component='submenu'])") ||
+            /* istanbul ignore next */ []
+        );
 
-      // Get the last segment block
-      const lastSegmentBlock = ulElements.pop();
+        // Terminate early if there are no ul elements as we have nothing we need to apply styles to.
+        if (ulElements.length === 0) return;
```
edleeks87, line 182: Praise: good idea 👏

edleeks87 (commented):
Good work @DipperTheDan, I'm not in love with how much JS we're needing to write to apply some CSS styling here but I can't think of a better way due to all the compositional combinations we support here

edleeks87 (approved):

nuria1110 (approved):

PR #6628, opened 2024-03-12
feat(dialog): allow subtitle prop to accept a node

### LINKED ISSUE #5844: enable multiline subtitles for Dialog
### Desired behaviour

Allow Dialog, subtitle prop to change, not to reveice exclusively a string parameter, but considering using an array of strings. render all iff present or not render the wrapper. This way i could show more than one subtitle, separated by line spaces or rows.

### Current behaviour

Dialog subtitle prop allows only string, this means im limited to one line and one message, as string this cannot be splitted or rearranged to be in multiple lines.

### Suggested Solution

Hello team,
There is a property from Dialog component to set subtitle, its a string so only allows one line.
A new design is in the need of having multilines for the subtitles. The way I found to make this happen is to use title property which allows to use a Component, this will do it for now, but the native prop for subtitle should also work similar than title on this regard.

https://codesandbox.io/s/pensive-minsky-v1kupv

Also note that using a component at title prop is not showing the subtitle, even if present

### Confidentiality

mihai-albu-sage:
enable consumers to pass a node as the subtitle prop

fix #5844

### Proposed behaviour

Allow the consumers to pass a node as the subtitle prop

### Current behaviour

Currently the type of the subtitle prop can only be  `string`

### Checklist

#### QA

### DIFF
--- a/src/components/dialog-full-screen/dialog-full-screen-test.stories.tsx
+++ b/src/components/dialog-full-screen/dialog-full-screen-test.stories.tsx
@@ -1,13 +1,16 @@
-import React, { useState } from "react";
+import React, { useState, useRef } from "react";
+import { StoryObj } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 import DialogFullScreen, { DialogFullScreenProps } from ".";
 import Dialog from "../dialog";
 import Button from "../button";
 import Form from "../form";
+import Icon from "../icon";
+import Textbox from "../textbox";
 
 export default {
   title: "Dialog Full Screen/Test",
-  includeStories: ["DefaultStory"],
+  includeStories: ["DefaultStory", "WithTwoDifferentNodes"],
   parameters: {
     info: { disable: true },
     chromatic: {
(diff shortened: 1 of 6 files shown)


edleeks87 (dismissed):

edleeks87 (commented):
@mihai-albu-sage I've just realised that [dialog-full-screen] has the same prop, I think we should do the same change in that component as well

mihai-albu-sage:
> @mihai-albu-sage I've just realised that [dialog-full-screen] has the same prop, I think we should do the same change in that component as well

I've added this change to the [dialog-full-screen] as well.

edleeks87 (dismissed):

Parsium (approved):

sianford (dismissed):

edleeks87 (dismissed):

sianford (dismissed):

src/components/dialog-full-screen/dialog-full-screen-test.stories.tsx
Parsium, line 1: **issue**: happy to use chromatic to visually test the new `WithTwoDifferentNodes` example, though it seems we have disabled chromatic for the file on line 16 so this will need changing.

src/components/dialog-full-screen/dialog-full-screen-test.stories.tsx
```diff
@@ -128,3 +130,37 @@ export const Nested = () => {
...
+        <Textbox label="Textbox1" value="Textbox1" />
+        <Textbox label="Textbox2" value="Textbox2" />
+        <Textbox label="Textbox3" value="Textbox3" />
+        <Form>{children}</Form>
+      </DialogFullScreen>
+    </>
+  );
+};
+
+WithTwoDifferentNodes.storyName = "with two different nodes in subtitle";
```
Parsium, line 166: **suggestion**: could we enable Chromatic to snapshot this story?

As we are opting to have a visual test for this story instead of writing a Playwright test, this will make sure any visual changes to this story are caught and assessed. 

```suggestion
WithTwoDifferentNodes.storyName = "with two different nodes in subtitle";
WithTwoDifferentNodes.parameters = {
   chromatic: { disableSnapshot: false },
   themeProvider: { chromatic: { theme: "sage" } }
};
```

sianford (approved):

src/components/dialog-full-screen/dialog-full-screen-test.stories.tsx
edleeks87: I'm going to approve as once the themes have been removed it's less important to set the chromatic theme imo

edleeks87 (approved):

PR #6634, opened 2024-03-14
build: remove proptypes from production bundles

robinzigmond:
Carbon components are released with `propTypes` definitions, automatically generated from the Typescript types, which are in some cases very large (eg Box has over 3500 lines of proptypes in the npm package). These can be useful in development for teams not using typescript, but are not needed in production (and proptypes warnings are turned off by React in production builds anyway). To reduce the size of consumers' builds, this change wraps the generated proptype definitions in a conditional check on `process.env.NODE_ENV`, so that they will still be there in development but will be stripped out of production builds by the bundler.

BREAKING CHANGE: Carbon components will no longer expose `propTypes` in production builds - any functionality that relies on reading the `propTypes` of a component will break

### Proposed behaviour

`propTypes` of Carbon components should only exist in development, and the definitions not be added to production builds.

### Current behaviour

Carbon builds include all `propTypes` definitions unconditionally, which will then be bundled into the build output of any consuming app, including in production.

### Checklist

#### QA

### Additional context

To see the problem being addressed here, look at the code in our npm builds: go to https://www.npmjs.com/package/carbon-react?activeTab=code and drill down to `lib/components/box/box.component.js`. The file has over 3800 lines, but everything from line 59 to the end (barring the final 2 lines) is prop-types.

This is made worse by the fact that some other components - eg `flex-tile-cell` inside `Tile` - extend the props of Box, and end up bundling the same huge proptypes declaration all over again (they don't share a single definition - changing that would require changes in the third-party typescript-to-proptypes package we use).

And while these are extreme examples, even just the styled-system margin props end up adding around 300 lines of unneeded propTypes to each component definition that uses them (many).

Here's an illustration of the difference this can make - with a simple app (I used Create React App but this of course isn't required), and the following minimal UI that happens to use two of the worst-offending components:

```
function App() {
  return (
    <>
    <Box>This is a box</Box>
    <FlexTileCell>This is a flex tile cell</FlexTileCell>
    </>
  );
}
```

the file-size of the production build of the compiled JS is 475kb, and clearly includes those huge propTypes definitions. 
Screenshot 2024-03-14 at 11 50 59 This goes down to just 238kb when run against a Carbon build using the changes in this PR.

### DIFF
--- a/plugins/generate-proptypes/inject-prop-types.mjs
+++ b/plugins/generate-proptypes/inject-prop-types.mjs
@@ -3,7 +3,10 @@ import * as ttp from "typescript-to-proptypes";
 
 /**
  * Create `Component.propTypes = {...}` source and inject it into the
- * AST of the current file
+ * AST of the current file.
+ * Wraps in an `if (process.env.NODE_ENV !== "production")` to ensure proptypes are not generated
+ * in production builds - as these are not needed in production and for some components (eg Box) lead
+ * to a big increase in bundle size.
  * @param t Babel types object
  * @param options path, AST of props, and name of node to inject proptypes for
  */
@@ -13,6 +16,8 @@ function injectPropTypes(t, options) {
   const propTypesSource = ttp.generate(props, {
     importedName: "PropTypes",
   });
+
+  const conditionalPropTypes = `if (process.env.NODE_ENV !== "production") {${propTypesSource}}`;
 
   if (propTypesSource.length === 0) {
     return;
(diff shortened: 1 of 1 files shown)


edleeks87 (approved):

Parsium (approved):
LGTM @robinzigmond 👍🏼 

Cannot find any PropType references in a production code of a built Carbon app:

Screenshot 2024-03-14 at 16 00 54

Issue #6403, opened 2023-10-31
ActionPopover Accessibility Issue


IrynaD26:
### Current behaviour

`ActionPopover` causes the following accessibility issue: "All page content should be contained by landmarks". More info about the issue: https://dequeuniversity.com/rules/axe/4.8/region?application=AxeChrome

This probably happens because `ActionPopover` appends `div` with the menu to `body`. So the menu and its items are not within landmarks.  

### To Reproduce

1. Click the menu button
3. Run Axe Tools (**sensure best practice is switched on for Axe**)
4. Observe the following error: "All page content should be contained by landmarks" (Axe Tools show errors for the entire page in the sandbox.  So you need to click arrows in Axe Tools to get to the `div` which wraps `ActionPopover` )

NOTE:  This accessibility issue occurs even if `ActionPopover`  is within `main` or some other landmark. That's because the action popover menu is always in the `div` which is appended to `body`. 

https://github.com/Sage/carbon/assets/148762678/d2b8f4ed-2168-4103-9853-ab2082e5d760

 

### Expected behaviour

No Axe issue "All page content should be contained by landmarks"

### CodeSandbox or Storybook URL

https://codesandbox.io/s/compassionate-currying-xqh4vy?file=/src/App.js:175-603

### JIRA Ticket (Sage Only)

SBS-74832

### Suggested Solution

Possible solutions:
1. not to render the action popover menu in `body`(maybe add some prop for that)
2. use `role="region"` and appropriate aria-label for the wrapper `div` of the action popover menu (not sure if it's a good solution though)

### Carbon Version

120.6.1

### Design Tokens Version

4.25.0

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

MacOS

### Anything else we should know?

Axe Tools may not show this error if there are other accessibility issues. So if you are trying to reproduce this issue in some other place (not sandbox) and can't reproduce it, try fixing other accessibility issues.

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

Parsium:
FE-6268

PR #6639, opened 2024-03-20
fix(popover): update portal landmark role to 'region'

ibsenjg:
fix (#6403)

### Proposed behaviour
for Dashboard MFE, inside Tasks tab there is one ActionPopover implemented [here](https://github.com/Sage/sbc.common.project.ui/blob/master/src/tasks/Tile/taskTile.tsx)  for Actions button. This having the issue with "Portal" landmark, so one option to solve we can either use DisablePortal prop inside the carbon component https://github.com/Sage/carbon/blob/master/src/components/action-popover/action-popover.component.tsx#L279 or to change the landmark when it comes to a Portal.

### Current behaviour

Current Popover component as renders at top DOM level it asks for main Landmarks to wrap it and their content. 
Renders a simple div and it should use role="region", this also affects or add extra A11y rules as making these landmarks unique and by doing it you should describe them either by aria-label or aria-labelledby, this also affects some other E2E components that uses Popover somehow. 

This issue can be skipped by using disablePortal prop but lets say this developmnet will allow both situations to live together.

### Checklist

#### QA

### DIFF
--- a/src/__internal__/popover/popover.component.tsx
+++ b/src/__internal__/popover/popover.component.tsx
@@ -37,6 +37,8 @@ export interface PopoverProps {
   popoverStrategy?: "absolute" | "fixed";
   // Allows child ref to be set via a prop instead of dynamically finding it via children iteration
   childRefOverride?: MutableRefObject<HTMLDivElement | null>;
+  /** Prop to specify the aria-labelledby attribute of the input */
+  ariaLabelledBy?: string;
 }
 
 const defaultMiddleware = [
@@ -56,6 +58,7 @@ const Popover = ({
   animationFrame,
   popoverStrategy = "absolute",
   childRefOverride,
+  ariaLabelledBy,
 }: PopoverProps) => {
   const elementDOM = useRef<HTMLDivElement | null>(null);
   const { isInModal } = useContext<ModalContextProps>(ModalContext);
@@ -79,6 +82,10 @@ const Popover = ({
     content = children;
   } else {
     content = React.cloneElement(children, { ref: floatingReference });
+  }
+  if (elementDOM.current && !disablePortal && ariaLabelledBy) {
+    elementDOM.current.setAttribute("role", "region");
+    elementDOM.current.setAttribute("aria-labelledby", ariaLabelledBy);
   }
 
   useFloating({
--- a/src/components/form/components.test-pw.tsx
+++ b/src/components/form/components.test-pw.tsx
@@ -637,9 +637,9 @@ export const WithLabelsInline = () => (
       labelWidth={30}
       labelId="inline-inputs"
     >
-      <Textbox aria-labelledby="inline-inputs" />
-      <Textbox aria-labelledby="inline-inputs" />
-      <Select aria-labelledby="inline-inputs">
+      <Textbox aria-label="inline input A" />
+      <Textbox aria-label="inline input B" />
+      <Select aria-label="inline select A">
         <Option value="1" text="option 1" key="1" />
         <Option value="2" text="option 2" key="1" />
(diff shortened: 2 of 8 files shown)


edleeks87 (changes requested):
Hi @ibsenjg thanks for raising this. I don't think `aside` is the right element to use here, it's [for content that's tangentially associated with the main content on a page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside). I chatted with @tempertemper yesterday and his suggestion was that we could try keeping it as a `div` and setting the `role="region"` or using something more generic like `section`, we'll need to add an `aria-label` to stay compliant with https://www.w3.org/TR/WCAG/#name-role-value. Alternatively we can continue to ignore this as it's fairly trivial and I believe we'll eventually be removing `Portal`s from our `Popover` component.

```
elementDOM.current = document.createElement("div");
elementDOM.current.setAttribute("role", "region");
elementDOM.current.setAttribute("aria-label", ariaLabel)

File: src/__internal__/popover/popover.component.tsx [hunk 1]
```diff
@@ -64,6 +64,8 @@ const Popover = ({
 
   if (!elementDOM.current && !disablePortal) {
     elementDOM.current = document.createElement("div");
+    elementDOM.current.setAttribute("role", "region");
+    elementDOM.current.setAttribute("aria-label", "ariaLabel-example");
```
JinCoreana, line 68: Can you assign a unique aira-label or set aria-labelledby for it to follow a unique Id?

src/components/form/components.test-pw.tsx
```diff
@@ -653,9 +653,9 @@ export const WithLabelsInline = () => (
       labelWidth={30}
       labelId="inline-inputs-second"
     >
-      <Textbox aria-labelledby="inline-inputs-second" />
-      <Textbox aria-labelledby="inline-inputs-second" />
-      <Select aria-labelledby="inline-inputs-second">
+      <Textbox aria-label="inline input second A" />
+      <Textbox aria-label="inline input second B" />
+      <Select aria-label="inline select B">
```
JinCoreana, line 658: How this changes affect `Form` component?

edleeks87: I would also like to understand why this change was made, I don't think there's an issue updating here but just wondering if there was an issue with using labelledby here?

ibsenjg: You think its correct to have one label for every of these, Is not better to label them each?

src/components/select/select-list/select-list.component.tsx
edleeks87, line 1: I think when this [PR](https://github.com/Sage/carbon/pull/6609/files#diff-55faed352d08ca12908f546812ccf331e07f38b23a295bab3ba193b7703565f4) is merged we won't need this change anymore

src/components/form/components.test-pw.tsx
ibsenjg: One of the reasons of my change is to ensure these role="region" elements should have unique labels is being affected by these, as they use the same labels.. they end having its attribute into the same one.
Captura de pantalla 2024-03-28 a las 12 35 33

src/components/select/select-list/select-list.component.tsx
ibsenjg: Yes, if you arent using a rendered element outside this is not needed anymore. Anyways I'll update this version for dashboard before

ibsenjg:
This PR became unneeded after [#6609](https://github.com/Sage/carbon/pull/6609/files#diff-55faed352d08ca12908f546812ccf331e07f38b23a295bab3ba193b7703565f4)

PR #6642, opened 2024-03-22
chore(playwright): configure `getByTestId` locator to find `data-role` tags

Parsium:
### Proposed behaviour

- Configure Playwright's `getByTestId` locator to find `data-role` tags, given Carbon uses data-role extensively as a test identifier and to match RTL.
- Update testing guide's sections on Playwright and Jest

### Checklist

#### QA

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 __spec__.js
 __definition__.js
+*.md
--- a/contributing/testing-guide.md
+++ b/contributing/testing-guide.md
@@ -2,32 +2,31 @@
 
 ## Contents
 
-- [Testing guide](#testing-guide)
-  - [Contents](#contents)
-  - [Introduction](#introduction)
-  - [Component testing](#component-testing)
-    - [Custom utilities](#custom-utilities)
-    - [Use of Snapshot tests](#use-of-snapshot-tests)
-    - [Continuous Integration (CI)](#continuous-integration-ci)
-  - [Functional Browser Testing](#functional-browser-testing)
-    - [Playwright File Structure](#playwright-file-structure)
-    - [Locators](#locators)
-  - [Visual Testing](#visual-testing)
-    - [Adding new visual tests](#adding-new-visual-tests)
+- [Component tests](#component-tests)
+  - [Enzyme utilities](#enzyme-utilities)
+  - [React Testing Library test ID tests](#react-testing-library-test-id-tests)
+  - [Use of Snapshot tests](#use-of-snapshot-tests)
+  - [Continuous Integration (CI)](#continuous-integration-ci)
+- [Browser-based component tests](#browser-based-component-tests)
+  - [Playwright File Structure](#playwright-file-structure)
+  - [Locators](#locators)
+- [Visual Testing](#visual-testing)
 
 ## Introduction
 
-This guide details Carbon's testing setup, common conventions and utilities available to you.
+This guide details Carbon's testing setup, common conventions, and utilities available to you.
 
-## Component testing
+## Component tests
 
-We use the [Jest](https://facebook.github.io/jest/) framework and [Enzyme](https://enzymejs.github.io/enzyme/) library for our component tests. Tests should describe the **behaviour of the components** rather than describe the implementation to keep the tests clean and reliable. All props, branches and paths and each of their conditions need to be tested to meet Carbon's 100% coverage policy.
+> **NOTE**: From March 2024, we're transitioning from Enzyme to [React Testing Library (RTL)](https://testing-library.com/docs/react-testing-library/intro/). Please use RTL where possible when testing new functionality.
 
-### Custom utilities
+We use the [Jest](https://jestjs.io/) framework and [Enzyme](https://enzymejs.github.io/enzyme/) library for our component tests. Tests should describe the **behaviour of the components** rather than describe the implementation to keep the tests clean and reliable. All props, branches, and paths and each of their conditions need to be tested to meet Carbon's 100% coverage policy.
 
-To help with common testing scenarios, we have a number of custom utilities available in `/src/__spec_helper__` that can be imported directly into your tests.
+### Enzyme utilities
 
-For example the `assertStyleMatch` method which asserts if all the expected CSS properties have been applied to a DOM element or React component:
+To help with common testing scenarios, we have several custom utilities available in `/src/__spec_helper__` that can be imported directly into your tests.
+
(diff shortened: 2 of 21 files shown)


edleeks87 (commented):
Worth double checking if the playwright tests have failed legitimately or have flaked. Let me know what you think about my comment happy to approve now if you'd prefer push that to an new ticket etc

src/components/icon/icon.component.tsx
```diff
@@ -19,7 +21,10 @@ export type LegacyIconTypes =
   | "success"
   | "messages";
 
-export interface IconProps extends Omit<StyledIconProps, "type">, MarginProps {
+export interface IconProps
+  extends Omit<StyledIconProps, "type">,
+    MarginProps,
+    TagProps {
```
edleeks87, line 27: suggestion (non-blocking): `data-component` is private/ ignored in the TagProps interface, it would not complain if a value is passed but wouldn't do anything so maybe it's worth just omitting it to avoid this

```suggestion
    Omit<TagProps, "data-component"> {
```

Parsium: Happy to add this in 👍🏼 It may be necessary though to set "icon" as the default value for `data-role` within `Icon` itself, due to scenarios where it is deeply nested in the React DOM tree and we cannot directly set the attribute's value:

```tsx
// in file-upload-status.component.tsx...

const fileLink = linkProps ? (
    <Link {...linkProps}>{filename}</Link> // 👈 Cannot set data-role of Icon directly
  ) : (
    /* falsy case */
  );
```

Parsium: Might be worth discussing if we need both the `data-element` and `data-role` tags going forward, as they seem to be used interchangeably by the tests

edleeks87: I don't mind this approach, the alternative is to set it in the Link component like below but I'm not overly bothered which approach we take

```
 <Icon
          type={icon}
          disabled={disabled}
          ariaLabel={removeAriaLabelOnIcon ? undefined : ariaLabel}
          tooltipMessage={tooltipMessage}
          tooltipPosition={tooltipPosition}
          data-role={`${icon}-link-icon`}
        />

**Review by edleeks87 — COMMENTED** [review]
I'm happy to approve this in it's current form but will hold off until there's a second reviewer as I'm on the fence about [this](https://github.com/Sage/carbon/pull/6642/files#r1537750065) so would like them to help break the tie

**Review by DipperTheDan — APPROVED** [review]

**Review by edleeks87 — APPROVED** [review]

File: contributing/testing-guide.md [hunk 2]
```diff
@@ -163,4 +208,4 @@ Chromatic is set-up to check for regressions in all component stories. Typically
           └── [component-name]-test.stories.* (private stories)
 ```
 
-To introduce a new visual test, create a story in the relevant stories file that demonstrates the behaviour and it will automatically be detected by Chromatic.
+To introduce a new visual test, create a story in the relevant stories file that demonstrates the behaviour, and it will automatically be detected by Chromatic.
```
sianford, line 211: Should we specify "once enabled" here? As in:
"To introduce a new visual test, create a story in the relevant stories file that demonstrates the behaviour, and, once enabled, it will automatically be detected by Chromatic."

playwright/README.md
```diff
@@ -9,23 +9,28 @@
```
sianford, line 9: In line 8 above, "Playwright" should have an upper case "P". I know it was there already, but I thought that as the file is being amended anyway then it could be corrected :)

playwright/README.md
```diff
@@ -9,23 +9,28 @@
 
 ## Installation
 
-We are planning to use [Playwright](https://playwright.dev) testing framework for functional and regression testing. Playwright is already installed in the Carbon project. Clone and install the carbon repository to apply the installation:
+Ensure Carbon is first installed on your machine by doing:
```
sianford, line 12: I think it would read better if this ended with "the following:", i.e. "Ensure Carbon is first installed on your machine by doing the following:".

contributing/testing-guide.md
```diff
@@ -45,34 +44,49 @@ describe("FlatTableRow", () => {
...
 1. `npm format` - run prettier to format code under `/src`.
 2. `npm run lint` - run linter on code under `/src`.
 3. `npm run type-check` - run TypeScript compiler to check for type errors.
 4. `npm test` - runs unit tests.
 
-## Functional Browser Testing
+## Browser-based component testing
 
-> **NOTE** - As of 3rd August 2023, we are in the progress of switching test framework for our functional browser tests from [Cypress](https://www.cypress.io/) to [Playwright](https://playwright.dev). The following section refers to our upcoming Playwright setup, but further details regarding our current Cypress setup can be found in [our Cypress docs](../cypress/README.md).
+We use [Playwright](https://playwright.dev) for browser-based component testing. Jest-tested functionality doesn't necessarily need retesting in Playwright unless it adds additional insight into user behaviour.
```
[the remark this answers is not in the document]
Parsium: I assume you mean for tests that check event handling such as clicking, typing, focusing, etc.? I guess the decision of which framework to use, RTL or Playwright, would be on a test-by-test basis, as we'd need to decide if the benefits of using a real browser environment would outweigh using RTL for speed.

RTL, especially when compared to Enzyme, does encourage you to write tests in a way that mimics user interactions more closely, but RTL tests are still run in a simulated browser.

Perhaps this is a topic for us to discuss more as a team first, so I'm happy to remove this docs change for now.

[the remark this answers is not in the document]
Parsium: @edleeks87 @DipperTheDan what do you think?

DipperTheDan (dismissed):

.eslintignore
```diff
@@ -1,2 +1,3 @@
 __spec__.js
 __definition__.js
+*.md
```
Parsium, line 3: ESlint is currently incorrectly parsing `*.md` files resulting in parsing errors

DipperTheDan (approved):

edleeks87 (approved):

contributing/testing-guide.md
```diff
@@ -161,4 +208,40 @@ Chromatic is set-up to check for regressions in all component stories. Typically
           └── [component-name]-test.stories.* (private stories)
 ```
 
-To introduce a new visual test, create a story in the relevant stories file that demonstrates the behaviour and it will automatically be detected by Chromatic.
+To introduce a new visual test, create a story in the relevant stories file that demonstrates the behaviour.
```
sianford, line 211: I really like this amendment with the additional information. It's well explained, with examples. Nice!

PR #6665, opened 2024-04-09
Added woff2 format for Icon font

harpalsingh:
### Proposed behaviour

There is no update to the icons themselves, but adding the woff2 format of the font, which when required we can updated to.

### Current behaviour

Only provide the woff format of the icon font.

### Checklist

#### QA


chrisbarber86:
I think we should update this CSS to reference the both woff2 and woff files:

https://github.com/Sage/carbon/blob/master/src/style/fonts.css#L4

Equally we should probably update this to preload the woff2 by default like the other fonts:

https://github.com/Sage/carbon/blob/master/.storybook/preview-head.html#L2-L6

chrisbarber86 (approved):

edleeks87 (approved):

PR #6682, opened 2024-04-16
fix(filterable-select): ensure no runtime error is triggered when component has a single option - FE-6413

DipperTheDan:
### Proposed behaviour

Filterable-Select: No runtime error should be thrown when the component has just one option.

https://github.com/Sage/carbon/assets/56251247/a3b930d5-21ef-4252-921b-10f837f20273

Multi-Select: No console error should be thrown when the component has just one option.

https://github.com/Sage/carbon/assets/56251247/e9ffb572-c002-452e-a8bd-c89afa3769d0

### Current behaviour

Filterable-Select: If the component has a single option, a runtime error is thrown when the select textbox is typed in.

https://github.com/Sage/carbon/assets/56251247/c7ae402a-aa5a-4856-a67e-0eebcb406ea0

Multi-Select: If the component has a single option, a console error is thrown when the select textbox is typed in.

### Checklist

#### QA

### Additional context

N/A 

### Testing instructions

Filterable-Select: 
- Visit new `Single Option` story in the `Test` folder for Filterable-Select. 
- Type in the Select Textbox. 
- The component should not throw a runtime error. 

Multi-Select: 
- Visit new `Single Option` story in the `Test` folder for Multi-Select. 
- Type in the Select Textbox. 
- The component should not throw a console error. 

It might also be worth doing a regression test on the other Filterable-Select and Multi-Select examples just as a sanity check.

### DIFF
--- a/src/components/select/filterable-select/filterable-select.component.tsx
+++ b/src/components/select/filterable-select/filterable-select.component.tsx
@@ -202,8 +202,8 @@ export const FilterableSelect = React.forwardRef(
       textToMatch: string,
       list: React.ReactNode
     ) {
-      return (list as React.ReactElement[]).find((child) => {
-        const { text } = child.props;
+      return React.Children.toArray(list).find((child) => {
+        const { text } = (child as React.ReactElement).props;
 
         return text?.toLowerCase().indexOf(textToMatch?.toLowerCase()) !== -1;
       });
@@ -213,7 +213,10 @@ export const FilterableSelect = React.forwardRef(
       (newFilterText: string, isDeleteEvent: boolean) => {
         setSelectedValue((previousValue) => {
           const trimmed = newFilterText.trimStart();
-          const match = findElementWithMatchingText(trimmed, children);
+          const match = findElementWithMatchingText(
+            trimmed,
+            children
+          ) as React.ReactElement;
           const isFilterCleared = isDeleteEvent && !newFilterText.length;
 
           if (!match || isFilterCleared || match.props.disabled) {
--- a/src/components/select/multi-select/multi-select-test.stories.tsx
+++ b/src/components/select/multi-select/multi-select-test.stories.tsx
@@ -537,3 +537,31 @@ export const MultiSelectWithDisabledOption = () => {
     </MultiSelect>
   );
 };
+
+export const SingleOption = () => {
+  const [selectedPills, setSelectedPills] = useState([] as string[]);
+  const handleActivityChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSelectedPills((event.target.value as unknown) as string[]);
+    partialAction("onChange")();
+  };
+  return (
+    <MultiSelect
+      name="testing"
+      value={selectedPills}
+      onChange={handleActivityChange}
+      onOpen={partialAction("onOpen")}
+      onClick={partialAction("onClick")}
+      onFilterChange={partialAction("onFilterChange")}
+      onFocus={partialAction("onFocus")}
+      onBlur={partialAction("onBlur")}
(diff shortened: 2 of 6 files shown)


src/components/select/multi-select/multi-select-test.stories.tsx
```diff
@@ -537,3 +537,34 @@ export const MultiSelectWithDisabledOption = () => {
     </MultiSelect>
   );
 };
+
+export const SingleOption = () => {
+  const MAX_SELECTIONS_ALLOWED = 2;
```
nuria1110, line 542: **nitpick(non-blocking):** I think this could be removed from the story since there is only one option

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -211,8 +211,8 @@ export const FilterableSelect = React.forwardRef(
       textToMatch: string,
       list: React.ReactNode
     ) {
-      return (list as React.ReactElement[]).find((child) => {
-        const { text } = child.props;
+      return React.Children.toArray(list).find((child) => {
+        const { text } = (child as React.ReactElement).props;
 
         return text?.toLowerCase().indexOf(textToMatch?.toLowerCase()) !== -1;
```
Parsium, line 217: **suggestion**: what do you think about amending the typing like this, to avoid having to use type assertions and mask potential type errors?:

```suggestion
      return React.Children.toArray(list).find(
        (child): child is React.ReactElement<OptionProps> => {
          if (!React.isValidElement<OptionProps>) return;
          
          const { text } = child.props;

          return text.toLowerCase().indexOf(textToMatch?.toLowerCase()) !== -1;
```

src/components/select/filterable-select/filterable-select.component.tsx
```diff
@@ -222,7 +222,10 @@ export const FilterableSelect = React.forwardRef(
       (newFilterText: string, isDeleteEvent: boolean) => {
         setSelectedValue((previousValue) => {
           const trimmed = newFilterText.trimStart();
-          const match = findElementWithMatchingText(trimmed, children);
+          const match = findElementWithMatchingText(
+            trimmed,
+            children
+          ) as React.ReactElement;
```
Parsium, line 228: **comment**: using the previous comment's typing should mean `match` is automatically typed as either an `Option` or `undefined`:

```suggestion
          const match = findElementWithMatchingText(
            trimmed,
            children
          );
```

src/components/select/multi-select/multi-select.component.tsx
```diff
@@ -240,17 +240,20 @@ export const MultiSelect = React.forwardRef(
       textToMatch: string,
       list: React.ReactNode
     ) {
-      return (list as React.ReactElement[]).find((child) => {
-        const { text } = child.props;
+      return React.Children.toArray(list).find((child) => {
```
Parsium, line 243: **suggestion**: likewise what do you think of using a type predicate and `React.isValidElement` here, instead of the type assertions, so no potential type errors are masked?

src/components/select/filterable-select/filterable-select.component.tsx
DipperTheDan: It should work if I pass the `child` as an argument to `isValidElement`.

src/components/select/multi-select/multi-select.component.tsx
DipperTheDan: I can but as a side effect of this I'd probably need to also manipulate the type passed in the `handleTextboxChange` to something like:

```
        if (match) {
          setHighlightedValue(String(match.props.value));
        }
```

Parsium (approved):

stephenogorman (approved):

PR #6696, opened 2024-04-25
feat(loader-spinner): add spinnerLabel prop to override the default loading label with custom text

### LINKED ISSUE #6694: Loader Spinner : custom text
### Description

Add the possibility to customise the text for each Loader Spinner

### Suggested solution

Add a label prop or the possibility to have a child

### Demo URL

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Confidentiality

tomdavies73:
fix #6694

### Proposed behaviour

Consumers can now use the `spinnerLabel` prop to override the default label of `"Loading..."` and provide their own custom string instead

### Current behaviour

Currently, consumers can only override the default label of `Loading...` by using an `<I18nProvider/ >`

### Checklist

#### QA

### DIFF
--- a/src/components/loader-spinner/loader-spinner.spec.tsx
+++ b/src/components/loader-spinner/loader-spinner.spec.tsx
@@ -56,6 +56,27 @@ describe("LoaderSpinner", () => {
   });
 
   describe("when custom props are passed", () => {
+    it("should be visible when the 'spinnerLabel' prop is passed a custom string value", () => {
+      render(<LoaderSpinner spinnerLabel="foo" />);
+      const visibleLabelElement = screen.getByText("foo");
+
+      expect(visibleLabelElement).toBeVisible();
+    });
+
+    it("should override the visible label text when the 'spinnerLabel' prop is passed a custom string value", () => {
+      render(<LoaderSpinner spinnerLabel="foo" />);
+      const wrapperElement = screen.getByRole("status");
+
+      expect(wrapperElement).toHaveTextContent("foo");
+    });
+
+    it("should override the visually hidden label text when the 'spinnerLabel' prop is passed a custom string value", () => {
+      render(<LoaderSpinner spinnerLabel="bar" showSpinnerLabel={false} />);
+      const hiddenLabelElement = screen.getByTestId("hidden-label");
+
+      expect(hiddenLabelElement).toHaveTextContent("bar");
+    });
+
     it("when the 'size' prop is passed as 'extra-small' the component wrapper has a flex-direction of row", () => {
       render(<LoaderSpinner size="extra-small" />);
       const wrapperElement = screen.getByRole("status");
--- a/src/components/loader-spinner/loader-spinner-test.stories.tsx
+++ b/src/components/loader-spinner/loader-spinner-test.stories.tsx
@@ -16,6 +16,11 @@ export default {
     },
   },
   argTypes: {
+    spinnerLabel: {
+      control: {
+        type: "text",
+      },
+    },
     size: {
       options: LOADER_SPINNER_SIZES,
       control: {
--- a/src/components/loader-spinner/loader-spinner.component.tsx
+++ b/src/components/loader-spinner/loader-spinner.component.tsx
@@ -19,6 +19,11 @@ import Typography from "../typography";
 
 export interface LoaderSpinnerProps extends MarginProps, TagProps {
   /**
+   * Use the spinnerLabel prop to override the default `"Loading..."` label with
+   * any custom string
+   */
+  spinnerLabel?: string;
+  /**
    * The size prop allows a specific size to be set, ranging from
    * `extra-small` to `extra-large`
    */
@@ -49,6 +54,7 @@ export interface LoaderSpinnerProps extends MarginProps, TagProps {
 }
 
 export const LoaderSpinner = ({
+  spinnerLabel,
   size = "medium",
   showSpinnerLabel = true,
   variant = "action",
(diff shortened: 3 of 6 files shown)


src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -56,6 +56,20 @@ describe("LoaderSpinner", () => {
   });
 
   describe("when custom props are passed", () => {
+    it("when the 'spinnerLabel' prop is passed a custom string value it overrides the default visible label", () => {
+      render(<LoaderSpinner spinnerLabel="foo" />);
+      const visibleLabelElement = screen.getByTestId("visible-label");
+
+      expect(visibleLabelElement).toHaveTextContent("foo");
```
Parsium, line 63: **suggestion(non-blocking)**: what do you think about querying if the label text is visible instead of using test ids? That way, it would mimick how a visible user would see the label:

```suggestion
      const visiblelabel = screen.getByText("foo");

      expect(visibleLabel).toBeVisible();
```

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -56,6 +56,20 @@ describe("LoaderSpinner", () => {
...
+    });
+
+    it("when the 'spinnerLabel' prop is passed a custom string value it overrides the visually hidden label", () => {
+      render(<LoaderSpinner spinnerLabel="bar" showSpinnerLabel={false} />);
+      const hiddenLabelElement = screen.getByTestId("hidden-label");
+
+      expect(hiddenLabelElement).toHaveTextContent("bar");
```
Parsium, line 70: **suggestion(non-blocking)**: what do you think about checking the label is visually hidden on the page instead of using test ids? That way, it would mimick how an AT user would receive the "bar" message:  

```suggestion
      // import { within } from "@testing-library/react";

      const liveRegion = screen.getByRole("status");
      const hiddenLabel = within(hiddenLabel).getByText("bar");

      expect(hiddenLabel).toHaveStyle({
        border: "0",
        height: "1px",
        margin: "-1px",
        overflow: "hidden",
        padding: "0",
        position: "absolute",
        width: "1px",
        "white-space": "nowrap",
      });
```

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -56,6 +56,20 @@ describe("LoaderSpinner", () => {
   });
 
   describe("when custom props are passed", () => {
+    it("when the 'spinnerLabel' prop is passed a custom string value it overrides the default visible label", () => {
```
DipperTheDan, line 59: **nitpick(non-blocking):** Test descriptions should make sense as a sentence. 

Something like:
```suggestion
    it("should allow a custom string value to be passed to the 'spinnerLabel' prop to override the default visible label", () => {
```
Feel free to make it something better than my suggestion if you want to 😄

src/components/loader-spinner/loader-spinner.spec.tsx
```diff
@@ -56,6 +56,20 @@ describe("LoaderSpinner", () => {
...
+      render(<LoaderSpinner spinnerLabel="foo" />);
+      const visibleLabelElement = screen.getByTestId("visible-label");
+
+      expect(visibleLabelElement).toHaveTextContent("foo");
+    });
+
+    it("when the 'spinnerLabel' prop is passed a custom string value it overrides the visually hidden label", () => {
```
DipperTheDan, line 66: nitpick(non-blocking): Same point here too. 

Something like: 
```suggestion
    it("should override the visually hidden label when the 'spinnerLabel' prop is passed a custom string value", () => {
```
or yet again something similar/better 😄

src/components/loader-spinner/loader-spinner.spec.tsx
tomdavies73: I wouldn't be opposed to this necessarily, but the previous approach dictates that we're actually testing the text is contained visually hidden labels DOM element, whereas in theory this is just asserting any text exists in the DOM - potentially anywhere

src/components/loader-spinner/loader-spinner.spec.tsx
tomdavies73: I don't mind this approach either, but won't this test become fairly brittle as it will fail every time any of these CSS attributes are slightly changed going forward?

src/components/loader-spinner/loader-spinner.spec.tsx
Parsium: I think this would be intentional here, as a visually-abled user won't care about which DOM element inside of the `LoaderSpinner` the text is placed in, but rather that the loading text is visible to them.

src/components/loader-spinner/loader-spinner.spec.tsx
Parsium: Good point 👍🏼 Maybe the `toBeInTheDocument` assertion would a better fit in that case?

src/components/loader-spinner/loader-spinner.spec.tsx
tomdavies73: I get what you mean mate, but this visible label is still announced to screen-reader users as it's rendered within the implicit aria-live region. The visually alternative hidden label is only rendered in its stead and announced when the `showSpinnerLabel` prop is `true`

src/components/loader-spinner/loader-spinner.spec.tsx
tomdavies73: I think it probably would, but we still have the dilemma that `getByText` will just assert that that specific text is found within the live region, and does not indicate whether the text found is visually hidden or not. Whereas the test-id approach does, as it directly points to the specific DOM element

DipperTheDan (approved):

Parsium (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #6701, opened 2024-04-29
fix(typography): textOverflow prop has no effect when the truncate prop is set

### LINKED ISSUE #6098: `textOverflow` prop on `Typography` can't be used
### Current behaviour

In plain CSS to to use the `text-overflow` property, you need to set `overflow: hidden` and `white-space: nowrap` which in Carbon can be done by setting the truncate prop, but setting this prop explicity sets `text-overflow` to `ellipsis`, so the value of textOverflow never has any effect.

### Expected behaviour

To be able to specify `textOverflow` alongside the `truncate` prop.

https://codesandbox.io/s/crimson-flower-f0o3go?file=/src/App.js

### JIRA Ticket (Sage Only)

_No response_

### Suggested Solution

_No response_

### Carbon Version

116.2.0

### Design Tokens Version

_No response_

Firefox, Chrome

Windows

_No response_

### Confidentiality

nuria1110:
fix #6098

### Proposed behaviour

When the `truncate` prop is set, `textOverflow` can be overridden. 

### Current behaviour

The `textOverflow` prop has no effect and cannot be overridden when the `truncate` prop is set. 

### Checklist

#### QA

### Testing instructions

 [Stackblitz with issue](https://stackblitz.com/edit/parsium-carbon-starter-hd4jq1?file=src%2FApp.tsx,package.json,src%2Fmain.tsx)

Go to "Truncated" story in storybook for fix.


src/components/typography/typography.style.ts
```diff
@@ -178,7 +178,7 @@ const StyledTypography = styled.span.attrs(
     white-space: ${truncate ? "nowrap" : whiteSpace};
     word-wrap: ${wordWrap};
     text-align: ${textAlign};
-    text-overflow: ${truncate ? "ellipsis" : textOverflow};
+    text-overflow: ${truncate && !textOverflow ? "ellipsis" : textOverflow};
```
tomdavies73, line 181: **suggestion**: here instead could we do `text-overflow: ${textOverflow || truncate && "ellipsis"}`. Might be slightly cleaner?

src/components/typography/typography.stories.tsx
```diff
@@ -122,8 +122,7 @@ export const TruncateStory: Story = () => (
   <>
     <Box height={80} width={80} backgroundColor="yellow">
       <Typography truncate>
-        The is an example of using the truncate prop. This is an example of some
-        text with applied.
+        The is an example of using the truncate prop with a block element.
```
tomdavies73, line 125: **praise**: good change

src/components/typography/typography.spec.tsx
```diff
@@ -606,16 +606,20 @@ describe("Typography", () => {
     );
 
     it.each(["clip", "ellipsis", "string", "initial", "inherit"])(
-      "applies text-overflow of %s",
+      "applies text-overflow of %s when text is truncated",
```
tomdavies73, line 609: **suggestion** (non-blocking): could it be worth testing this in a separate test, as we might want to test the attribute is still correctly applied before `truncate` is passed and after

src/components/typography/typography.mdx
```diff
@@ -35,7 +35,9 @@ It's also possible to create ordered and unordered lists using the `List`, `List
 
 ## Example of truncate
 
-The `truncate` prop is a boolean prop that automatically applies the styles required to truncate a piece of text.
+The `truncate` prop is a boolean prop that automatically applies  `overflow: hidden` and `white-space: nowrap`, which are the styles required to truncate a piece of text.
+
+By default, this prop will truncate the text with an ellipsis, but you can also pass the `textOverflow` prop to override this behaviour.
```
tomdavies73, line 40: **praise**: Nice, I like this a lot more, and it's a lot clearer. Lets consumers know that the `truncate` prop is not useless when `textOverflow` is applied

src/components/typography/typography.stories.tsx
```diff
@@ -133,6 +132,11 @@ export const TruncateStory: Story = () => (
         truncate.
       </Typography>
     </Box>
+    <Box height={80} width={80} backgroundColor="lightblue">
```
tomdavies73, line 135: **suggestion**(non-blocking): it could be worth seeing if the width of these box's could be extended to kind of give a better idea of the content, even though I know its meant to be truncated haha

tomdavies73 (commented):
Nice work @nuria1110 as always, as the `text-overflow` attribute is set with `overflow` and `white-space` could we look into making sure the `textOverflow` prop has an effect when used independently? If it has no effect when applied by itself, we may have to conditionally apply the other attributes too, similar too how we do it with `truncate`

src/components/typography/typography.spec.tsx
nuria1110: "white-space: nowrap" and "overflow" hidden" have to be set in order for the "text-overflow" property to have any effect and these can be set by the `truncate` prop, so they will always have to be used together. The properties could be set individually, but that is currently not possible as we do not have a prop to set the "overflow" property.

nuria1110: Yeah I didn't think it was worth adding more complexity to the styles if they could be just used together if needed

DipperTheDan (approved):

tomdavies73 (approved):

PR #6709, opened 2024-05-02
chore: set-up linting plugins for jest-dom and React Testing Library

Parsium:
### Proposed behaviour

- Install linting plugins for [`jest-dom`](https://github.com/testing-library/jest-dom#tohavestyle) and [React Testing Library (RTL)](https://testing-library.com/docs/react-testing-library/intro)
- Configure to use all the recommended rules from both these plugins
- Move all RTL tests to files with `*.test.*` extension
- Address all linting errors that have arisen in the existing RTL tests

### Current behaviour

- Enzyme and RTL tests are stored in the same files with `*.spec.*` extension
- Not all the existing RTL tests follow the best practices recommended by RTL's maintainers 

### Checklist

   - **Contains unverified commit to assist reviewers with reading changes, to be squashed**


src/hooks/__internal__/useMenuKeyboardNavigation/useMenuKeyboardNavigation.test.tsx
```diff
@@ -31,6 +31,8 @@ const MockComponent = ({ hideCb }: MockComponentProps) => {
 
   const getButtonChildren = useCallback(
     () =>
+      // DOM query is not used by a test to target elements
```
Parsium, line 34: Although not ideal, this query selector is used within a tested component instead of the test code itself.

src/components/time/time.test.tsx
```diff
@@ -131,24 +133,22 @@ describe("Time component", () => {
       />
     );
 
-    const user = userEvent.setup();
-    await act(async () => {
-      await user.tab();
+    const user = userEvent.setup({
+      advanceTimers: jest.advanceTimersByTime,
+      delay: null,
```
Parsium, line 138: We should be cautious about using the `delay` option.

By default, RTL adds a small delay between subsequent inputs, like key presses, to simulate how a real user would provide input. Setting `delay` to `null` removes this delay entirely, so it could potentially mask potential issues related to timeouts or promises in our code.

nineteen88:
I see some of the previous spec files that were more or less already RTL have had their filename changed so that RTL linting will occur on them but we should also make a note of these files that have been converted so they don't get missed when we sweep them for refactor. There's still quite a lot of bad practices in those files that we want to get on top of.

src/hooks/__internal__/useMenuKeyboardNavigation/useMenuKeyboardNavigation.test.tsx
nineteen88: question: it took a few reads of this comment to understand what was meant

Could we be a little more verbose with the comment so it's clearer what is happening?

Parsium:
Adding DO NOT MERGE label - since this PR depends on the release of #6715 first

Parsium:
> I see some of the previous spec files that were more or less already RTL have had their filename changed so that RTL linting will occur on them but we should also make a note of these files that have been converted so they don't get missed when we sweep them for refactor. There's still quite a lot of bad practices in those files that we want to get on top of.

Raised FE-6549 for this 👍🏼

src/components/file-input/file-input.component.tsx
```diff
@@ -25,7 +25,7 @@ import useLocale from "../../hooks/__internal__/useLocale";
 export interface FileInputProps
   extends Pick<ValidationProps, "error">,
     Pick<InputProps, "id" | "name" | "required">,
-    TagProps,
+    Omit<TagProps, "data-component">,
```
edleeks87, line 28: Praise: good spot 👏

src/components/flat-table/flat-table-body-draggable/flat-table-body-draggable.test.tsx
```diff
@@ -22,51 +127,34 @@ describe("FlatTableBodyDraggable", () => {
...
-            <FlatTableRow key="2" id={2}>
-              <FlatTableCell>Row three</FlatTableCell>
-            </FlatTableRow>
-          </FlatTableBodyDraggable>
-        </FlatTable>
-      );
-    });
-
-    test("on initial render the rows are in the correct order", () => {
+    it("on initial render the rows are in the correct order", () => {
```
edleeks87, line 148: suggestion: this doesn't really read quite right to me, maybe something like the below would work

```
it("should have roes in correct order on initial render")

File: src/components/flat-table/flat-table-body-draggable/flat-table-body-draggable.test.tsx [hunk 5]
```diff
@@ -150,43 +235,38 @@ describe("FlatTableBodyDraggable", () => {
...
-            <FlatTableRow expandable subRows={subRows} key="1" id={1}>
-              <FlatTableCell>Row two</FlatTableCell>
-            </FlatTableRow>
-            <FlatTableRow expandable subRows={subRows} key="2" id={2}>
-              <FlatTableCell>Row three</FlatTableCell>
-            </FlatTableRow>
-          </FlatTableBodyDraggable>
-        </FlatTable>
-      );
+    it("still allow a dynamically added row to be dragged and dropped", async () => {
```
edleeks87, line 239:
```
it("should allow a dynamically added row to be dragged and dropped"

File: src/components/menu/__internal__/keyboard-navigation/index.test.tsx
Comment (edleeks87) [inline src/components/menu/__internal__/keyboard-navigation/index.test.tsx:1]: praise: I really like the overall structure of this file, whilst there's nested `describes` here they logically split out the two branches this code supports. I do think we should avoid nesting as much as possible but if we do this is the level I'd like us to not go beyond

File: src/components/menu/__internal__/keyboard-navigation/index.test.tsx [hunk 6]
```diff
@@ -19,139 +18,123 @@ const getMockEvent = (key: string, which?: number) => {
...
-      )
-    );
-
-  const getItem = (index: number) =>
-    Array.from(document.querySelectorAll(`[data-component='${MENU_ITEM}']`))[
-      index
-    ];
+  afterAll(() => loggerSpy.mockRestore());
 
   describe("characterNavigation", () => {
```
edleeks87, line 44: suggestion (non-blocking): maybe something like this will read out better

```
describe("via character keys", () => {

File: src/components/menu/__internal__/keyboard-navigation/index.test.tsx [hunk 7]
```diff
@@ -19,139 +18,123 @@ const getMockEvent = (key: string, which?: number) => {
...
-      });
+    it("returns the correct element when the starting letter of a submenu is passed in", () => {
+      render(<MockMenu />);
+      const focusableItems = screen.getAllByRole("listitem");
+      const result = characterNavigation("s", focusableItems);
+      expect(result).toEqual(screen.getByTestId("submenu"));
     });
   });
 
   describe("menuKeyboardNavigation", () => {
```
edleeks87, line 89: suggestion (non-blocking): maybe this would be a bit more intuitive 

```
describe("via navigation keys", () => {

File: src/components/navigation-bar/fixed-navigation-bar-context.test.tsx [hunk 8]
```diff
@@ -35,7 +35,7 @@ describe("FixedNavigationBarContextProvider", () => {
           <MockComponent position={position} orientation="top" offset="20px" />
         );
         const result = screen.getByTestId("output");
-        expect(result.textContent).toBe("");
+        expect(result).toHaveTextContent("");
```
edleeks87, line 38: Praise: nice 👏

src/components/step-flow/step-flow.test.tsx
```diff
@@ -1,279 +1,333 @@
...
 
+  const stepsBefore = currentStep - 1;
+  const stepsAfter = totalSteps - currentStep;
+
+  return [stepsBefore, stepsAfter];
+}
+
+describe("Step Flow component", () => {
   describe("prop checks", () => {
     it("when the 'category' prop is passed, the correct element and text renders", () => {
```
edleeks87, line 52: suggestion (non-blocking):

```
it("should render the correct element and text when the `category` prop is passed"

Comment (edleeks87) [reply src/components/step-flow/step-flow.test.tsx:52]: most of the `it` blocks could do with updating here imo but I don't mind pushing that to a new ticket, only commenting as you've updated these elsewhere

File: src/components/tooltip/tooltip.test.tsx [hunk 10]
```diff
@@ -16,49 +11,43 @@ const mockedGuid = "guid-12345";
...
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function render(props: Partial<TooltipProps> = {}) {
-  const children = <div data-role="tooltip-target">foo</div>;
-  const message = "foo";
-
-  return renderRTL(
-    <Tooltip {...props} message={message || props.message}>
-      {props.children || children}
+function renderTooltip(props: Partial<TooltipProps> = {}) {
```
edleeks87, line 14: comment (non-blocking): same comment about abstracting the render, happy to sort all this later though 👍

src/components/tooltip/tooltip.test.tsx
```diff
@@ -174,41 +163,44 @@ describe("Tooltip", () => {
...
-              isVisible: true,
-              position,
-              isPartOfInput: true,
-              inputSize: size,
-            });
-
-            let middleware;
-            if (useFloatingSpy.mock.calls[0][0]?.middleware?.[0]) {
-              middleware = useFloatingSpy.mock.calls[0][0]?.middleware?.[0];
+      describe.each(["top", "bottom", "left", "right"] as const)(
```
edleeks87, line 166: comment (non-blocking): this is the prime example of a test that we need to refactor 😆. Again happy for it to be part of future work

src/hooks/__internal__/useModalAria/useModalAria.test.tsx
edleeks87, line 1: comment (non-blocking): again the `it` block text doesn't quite read right here but happy to leave as is for now

edleeks87 (commented):
Great work @Parsium some comments from me but mainly non-blocking

src/__internal__/focus-trap/focus-trap.test.tsx
```diff
@@ -238,723 +178,727 @@ describe("FocusTrap", () => {
...
+        expect(screen.getByRole("dialog")).toHaveFocus();
+        await user.tab({ shift: true });
+        expect(screen.getByRole("dialog")).toHaveFocus();
       });
     });
 
     describe("and some children elements are disabled", () => {
-      beforeEach(() => {
-        render(
+      it("only focuses those that are not", async () => {
```
nineteen88, line 382: suggestion(non-blocking): be more verbose with language

that are not what? This is more an issue with the describe block which we obviously need to revisit with these files but still think since we're adding an `it` it would make sense to be more clear

src/__internal__/focus-trap/focus-trap.test.tsx
```diff
@@ -238,723 +178,727 @@ describe("FocusTrap", () => {
...
+      </MockComponent>
+    );
 
-    describe("when focus on first radio button shift-tab pressed", () => {
-      it("should loop focus to the last focusable element", async () => {
-        focusElement(screen.getByLabelText(RADIO_LABEL_ONE));
-        await shiftTabPress();
-        expect(screen.getByText(BUTTON_TWO)).toHaveFocus();
-      });
+    it("loops focus to the last focusable element, when focus on first radio button shift-tab pressed", async () => {
```
nineteen88, line 420: praise: much improved language/description of the test

Applies to all of those changes

src/components/file-input/file-input.test.tsx
```diff
@@ -1,95 +1,80 @@
...
 
   it("accepts an inputHint prop", () => {
     render(<FileInput inputHint="help" onChange={() => {}} />);
     const hintText = screen.getByText("help");
-    expect(hintText).toBeInTheDocument();
     expect(hintText).toBeVisible();
   });
 
-  it("accepts data tag props", () => {
+  it("renders root container with correct data tag props", () => {
```
nineteen88, line 50: praise: much improved test description

Applies for the others you've made as well

src/components/file-input/file-input.test.tsx
```diff
@@ -98,29 +83,36 @@ describe("rendering with no file uploaded", () => {
     render(
       <FileInput label="file input" name="input-name" onChange={() => {}} />
     );
-    const hiddenInput = screen.queryByLabelText("file input");
+    const hiddenInput = screen.getByLabelText("file input");
     expect(hiddenInput).toHaveAttribute("name", "input-name");
   });
 
-  it("accepts a required prop", () => {
+  it("renders asterisk next to label when required prop is passed", () => {
```
nineteen88, line 90: suggestion(non-blocking): "renders **an** asterisk... "

src/components/file-input/file-input.test.tsx
```diff
@@ -98,29 +83,36 @@ describe("rendering with no file uploaded", () => {
...
-    render(<FileInput label="file input" isOptional onChange={() => {}} />);
-    const label = screen.getByText("file input");
-    assertStyleMatch({ content: '"(optional)"' }, label.parentElement, {
+    expect(screen.getByText("file input")).toHaveStyleRule("content", '"*"', {
       modifier: "::after",
     });
   });
 
-  it("accepts an isVertical prop", () => {
+  it("renders label with optional suffix when isOptional prop is passed", () => {
```
nineteen88, line 98: suggestion(non-blocking): "renders **a** label..."

src/components/file-input/file-input.test.tsx
```diff
@@ -98,29 +83,36 @@ describe("rendering with no file uploaded", () => {
...
+    expect(screen.getByTestId("label-container")).toHaveStyleRule(
+      "content",
+      '"(optional)"',
+      {
+        modifier: "::after",
+      }
+    );
+  });
+
+  it("renders root container with flex direction column when isVertical prop is true", () => {
```
nineteen88, line 110: suggestion(non-blocking): "renders **a** root container..."

src/components/flat-table/flat-table-body-draggable/flat-table-body-draggable.test.tsx
```diff
@@ -150,43 +235,38 @@ describe("FlatTableBodyDraggable", () => {
...
+
+      expect(
+        screen.getAllByRole("row").map((cell) => cell.textContent)
+      ).toEqual(["Row three", "Row one", "Row two"]);
     });
+  });
 
-    test("on initial render the rows are in the correct order", () => {
+  describe("with sub rows", () => {
+    it("on initial render the rows are in the correct order", () => {
```
nineteen88, line 260: suggestion: should remain as a `test()` unless the wording is updated

src/components/step-flow/step-flow.test.tsx
```diff
@@ -1,279 +1,333 @@
...
+    currentStep = totalSteps;
   }
 
+  const stepsBefore = currentStep - 1;
+  const stepsAfter = totalSteps - currentStep;
+
+  return [stepsBefore, stepsAfter];
+}
+
+describe("Step Flow component", () => {
```
nineteen88, line 50: suggestion(non-blocking): Describe isn't required

I know you've just moved it and we will revisit this file but it was flagging as a change so thought I'd mention. Not the end of the world for now as long as we tidy it up later

nineteen88 (commented):
Good work James. Can see a lot of very good improvements overall. A few things still don't sit right with me as previously mentioned/discussed but we'll get there!

src/__internal__/focus-trap/focus-trap.test.tsx
```diff
@@ -186,41 +134,33 @@ const WithAdditionalWrapperRefs = () => {
...
-          <button type="button">{BUTTON_TWO}</button>
-          <button type="button">{BUTTON_THREE}</button>
-        </div>
-      </FocusTrap>
-      <button type="button">{BUTTON_FOUR}</button>
-      <button type="button">{BUTTON_FIVE}</button>
-      <button type="button">{BUTTON_SIX}</button>
-    </>
-  );
+const setup = (ui: React.ReactElement) => {
```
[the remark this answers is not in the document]
Parsium: Good point, in hindsight this abstraction isn't really needed here

src/components/menu/__internal__/keyboard-navigation/index.test.tsx
Parsium: Thanks @edleeks87! Ideally, there is more refining I would like to do with this file, but it might be worth doing this as part of ticket FE-6549 or FE-6427

src/components/tooltip/tooltip.test.tsx
Parsium: Yeah, this file needs a good review due to the nested structure and styling rules, so I'll leave this for now.

src/hooks/__internal__/useMenuKeyboardNavigation/useMenuKeyboardNavigation.test.tsx
Parsium: I've made some adjustments to the comment. It might be a good idea for us to reconsider the `useMenuKeyboardNavigation` hook in a broader context. I feel the functionality it offers would be more effectively integrated into a single, versatile modal component or hook that encompasses all menu functionalities.

src/components/step-flow/step-flow.test.tsx
```diff
@@ -48,211 +48,209 @@ function calculateStepStateIndexes(
...
+        totalSteps={6}
+        category="bar"
+        ref={() => {}}
+      />
+    );
+
+    expect(screen.getByText("bar")).toBeVisible();
+  });
+
+  it("when the 'title' prop is passed, the correct element and text renders", () => {
```
nineteen88, line 65: suggestion(non-blocking): language still wrong on this test

Might be worth adding to the re-visit ticket rather than dealing with it now unless there were other changes to be made

src/components/step-flow/step-flow.test.tsx
```diff
@@ -48,211 +48,209 @@ function calculateStepStateIndexes(
...
+        overflow: "hidden",
+        padding: "0",
+        position: "absolute",
+        width: "1px",
+        "white-space": "nowrap",
+      });
+    }
+  );
+
+  it("when the 'showCloseIcon' prop is true, the correct element renders", () => {
```
nineteen88, line 242: suggestion(non-blocking): language still wrong on this test

Might be worth adding to the re-visit ticket rather than dealing with it now unless there were other changes to be made

nineteen88 (approved):

edleeks87 (approved):

PR #6713, opened 2024-05-02
fix: ensure tree-shaking works with date-fns package

nineteen88:
### Proposed behaviour

Update the import paths for `Date` utils to allow tree-shaking to work and reduce bundle size.

BREAKING CHANGE: Carbon no longer re-exports the `date-fns` locales. Consuming projects need to install the package and use locales directly from there.

closes: SBS-84395

### Current behaviour

date-fns tree-shaking is not working

### Checklist

#### QA

N/A


edleeks87 (commented):
Most of it LGTM @nineteen88 thanks for picking this up again! I think we should tweak the commit message slightly as I don't think they'll need to install date-fns as it will be in the node_modules already

edleeks87 (approved):

sianford (approved):

PR #6733, opened 2024-05-14
chore(badge): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert `Badge` tests to RTL.

### Current behaviour

`Badge` tests are Enzyme.

### Checklist

#### QA


src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    renderComponent({ counter: 101 });
+
+    expect(screen.getByText("99")).toBeVisible();
+  });
+
+  it("should not render badge when counter is `0`", async () => {
+    renderComponent({ counter: 0, onClick: () => {} });
+
+    expect(screen.queryByText("0")).not.toBeInTheDocument();
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
```
edleeks87, line 37: comment: I don't think you need both assertions here, the first one suffices imo

src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import Badge from "./badge.component";
+
+const renderComponent = (props = {}) =>
+  render(<Badge {...props}>Button</Badge>);
```
edleeks87, line 6:
```suggestion
  render(<Badge {...props}>Foo</Badge>);
```

non-blocking just looks a bit weird having that as the child text imo

src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+describe("Badge", () => {
+  it("should render correctly if counter is a number", async () => {
```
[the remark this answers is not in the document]
edleeks87: Also most of these don't need to be `async` tests as we're using synchronous `getBy` queries instead of `findBy` which is asynchronous and needs awaiting

src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  it("should not render badge when counter is not set", async () => {
+    renderComponent({ onClick: () => {} });
+
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+  });
+
+  it("should render as a button element when onClick is set", async () => {
+    renderComponent({ counter: 9, onClick: () => {} });
+
+    // investigate why toBeVisible() fails to find "9"
```
edleeks87, line 49: I wonder if the fact the close icon is also rendered here and includes a pseudo that it's appearing hidden to RTL 🤷.  If we end up needing to do an investigation separate from this ticket then we should raise one and include it in this comment as a FIXMe

src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  it("should not render as a button if onClick is not set", async () => {
+    renderComponent({ counter: 9 });
+
+    expect(screen.getByText("9")).toBeVisible();
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+  });
+
+  it("should have the relevant aria-label when aria-label is specified", async () => {
```
edleeks87, line 61: question: should we also test that it's not set when onClick is passed?

src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    renderComponent({ counter: 9, "aria-label": "Generic aria message" });
+
+    expect(screen.getByLabelText("Generic aria message")).toBeVisible();
+  });
+
+  describe("Styles", () => {
```
edleeks87, line 67: suggestion: I think we should avoid nesting the `describe` here, the `it` blocks have enough of a description imo

src/components/badge/badge.test.tsx
Parsium: After some investigation, removing `display: none;` CSS rule on Line 75 in `badge.style.ts` causes the test to pass. Sounds to me like either the pseudoelement @edleeks87 mentioned is causing unintended effects, or JSDOM, or even both. Worth us looking into in the future for sure :+1:

@nuria1110 would you be happy to raise a ticket for this?

src/components/badge/badge.test.tsx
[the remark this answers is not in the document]
nuria1110: Yeah I'll remove both of those tests, they don't seem to affect coverage

src/components/badge/badge.test.tsx
Parsium, line 6: **comment**: just wanted to mention, this simple abstraction is alright for now, but let's keep an eye out for similar ones in future PRs, especially when we're dealing with more complex test files.

It's important to ensure that any helper functions we add genuinely enhance clarity and don't make the file structure more complicated in the name of staying DRY.

Parsium:
Not much to add to @edleeks87's comments - nice work @nuria1110 👍🏼

src/components/badge/badge.test.tsx
nuria1110: Raised FE-6575 to investigate this.

Parsium (approved):
LGTM @nuria1110 - once @edleeks87 agrees let's squash the changes 👍🏼

edleeks87 (approved):

edleeks87:
Looks good @nuria1110, yeah lets squash the commits before we merge but no rush 👍

ZhuoyuJin (approved):

PR #6737, opened 2024-05-16
chore(button-bar): convert enzyme test to rtl

tomdavies73:
### Proposed behaviour

Converts enzyme tests to rtl.

### Current behaviour

Current tests are written in enzyme.

### Checklist

#### QA


src/components/button-bar/button-bar.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+      </ButtonBar>
+    );
+
+    const buttonChildren = screen.getAllByRole("button", {
+      name: "button-in-bar",
+    });
+
+    buttonChildren.forEach((button) => {
+      expect(button).toHaveStyle("font-size: var(--fontSizes100)");
+      expect(button).toHaveStyle("min-height: 40px");
```
DipperTheDan, line 65: **suggestion(non-blocking):** `toHaveStyle` accepts an object so you could put these into one assertion. I'm happy with either way. It's just whichever way we decide we want to do it going forward.

```suggestion
      expect(button).toHaveStyle({
        fontSize: "var(--fontSizes100)",
        minHeight: "40px",
      });
```

src/components/button-bar/button-bar.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    });
+  });
+
+  it("should render the Buttons with the 'iconPosition' prop set to 'before'", () => {
```
Parsium, line 69: **suggestion(non-blocking)**: how about we amend a few of test titles to make the intended behavior clearer? This way, if something breaks and a test fails, it will be easier for us to understand what the component is supposed to do.

For example:
```tsx
it("renders icon with right margin when iconPosition prop is 'before'");
it("does not render Buttons with full width when fullWidth prop is false");
```

I appreciate that most of the tests are checking trivial CSS rules though, so I'll leave this up to you. What are your thoughts?

DipperTheDan (approved):

Parsium (approved):

PR #6738, opened 2024-05-17
chore(confirm): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert `Confirm` tests to RTL.

### Current behaviour

`Confirm` tests are in enzyme. 

### Checklist

#### QA


src/components/confirm/confirm.test.tsx
```diff
@@ -0,0 +1,217 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import Confirm from "../confirm";
+
+jest.mock("../../__internal__/utils/helpers/guid", () => () => "guid-12345");
+
+describe("Confirm", () => {
```
nineteen88, line 8: suggestion: remove unnecessary describe block

src/components/confirm/confirm.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+  });
+
+  it("should call onCancel when cancel button is clicked", async () => {
+    const onCancel = jest.fn();
+    const user = userEvent.setup();
+    render(<Confirm open onConfirm={() => {}} onCancel={onCancel} />);
+
+    await user.click(screen.getByRole("button", { name: "No" }));
+    expect(onCancel).toHaveBeenCalledTimes(1);
+  });
```
nineteen88, line 46: suggestion(non-blocking): these four tests should be able to be condensed into two tests

Believe that should be possible anyway, if we lose coverage then obviously not. 

We could also add a more descriptive test title for them as well. e.g.

`it("should render a confirm button when onConfirm prop is set and will execute the function when clicked")`

src/components/confirm/confirm.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+          "data-element": "bar",
+        }}
+        closeButtonDataProps={{
+          "data-role": "close-button",
+          "data-element": "baz",
+        }}
+      />
+    );
+
+    expect(screen.queryByTestId("confirm-button")).toHaveAttribute(
```
edleeks87, line 175:
```suggestion
    expect(screen.getByRole("button", {name: "Yes")).toHaveAttribute(
    expect(screen.getByRole("button", {name: "No")).toHaveAttribute(

```
(Non-blocking) I think the recommended approach is to use getByRole wherever possible

src/components/confirm/confirm.test.tsx
nineteen88: Should also add that if tests aren't within a describe block then they should be `test()` as opposed to `it()` based on Kent C Dodd's recommendations

See: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing#inline-it

src/components/confirm/confirm.test.tsx
nuria1110: I went with this approach to check that the button was accessible by the new "data-role" being set, but I can change this to use `getByRole` and then assert the data-role attribute is set also correctly if thats preferred.

edleeks87 (approved):

nineteen88 (approved):

PR #6756, opened 2024-05-30
chore(box): convert to tests to rtl (FE-6557)

nineteen88:
### Proposed behaviour

- convert tests to RTL
- remove unnecessary style tests

### Current behaviour

- tests are currently written in Enzyme

### Checklist


src/components/box/box.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  testStyledSystemLayout,
+  testStyledSystemFlexBox,
+  testStyledSystemGrid,
+  testStyledSystemPosition,
+} from "../../__spec_helper__/test-utils";
+import Box from "./box.component";
+import Logger from "../../__internal__/utils/logger";
+
+// mock Logger.deprecate so that no console warnings occur while running the tests
+const loggerSpy = jest.spyOn(Logger, "deprecate");
```
Parsium, line 17: **suggestion**: what do you think about removing this mock, so any legitimate warnings that the tests throw that need addressing aren't masked?

edleeks87: I think it's worth trying to mock it for the specific test we want to ignore the console for, yes there's a small chance that specific one will have another legitimate console warning it's not likely a huge issue and removing the mock altogether means we'll have a lot of superfluous console logging polluting the output when we run our tests

edleeks87: Actually, is there any deprecation warnings we need to mock here?

nineteen88: I guess that's actually hangover from when we had deprecation warnings on Box. So I'm happy to remove it.

edleeks87 (approved):

PR #6766, opened 2024-06-06
fix(flat-table): sticky column with button - focus border partially covered in expanded rows

### LINKED ISSUE #6561: `FlatTableRowHeader` with `Button` causes extra paddings
### Current behaviour

The default settings cause extra paddings within the cell,

```
      <FlatTableRowHeader stickyAlignment="right">
        <Button>Default</Button>
      </FlatTableRowHeader>
```

Screenshot 2024-02-01 at 11 27 07

With p={0} attribute, focus outline gets partially hidden
```
      <FlatTableRowHeader p={0} stickyAlignment="right">
        <Button>Padding 0</Button>
      </FlatTableRowHeader>
```
Screenshot 2024-02-01 at 11 28 58

### Expected behaviour

When a `Button` is placed within a sticky cell, it should not result in additional padding, and the focus outline should be fully visible.
Screenshot 2024-02-01 at 11 29 45

https://codesandbox.io/p/sandbox/table-padding-focus-kzm5nr

### JIRA Ticket (Sage Only)

SBS-81079

### Suggested Solution

_No response_

### Carbon Version

124.6.0

### Design Tokens Version

4.28.0

Chrome

MacOS

_No response_

### Confidentiality

mihai-albu-sage:
fixes #6561

- The component `FlatTableRowHeader` uses `position: sticky;` and also recevies `p={0}`. 
- If the component `FlatTableRowHeader` receives the prop `stickyAlignment` with the value `left` any `FlatTableCell` to the left of the `FlatTableRowHeader` will also receive `position: sticky;`. 
- If the component `FlatTableRowHeader` receives the prop `stickyAlignment` with the value `right` any `FlatTableCell` to the right of the `FlatTableRowHeader` will also receive `position: sticky;`.
 
For each `FlatTableRowHeader` or `FlatTableCell` that are using `position: sticky;` together with z-index a new stacking context is created.

Because all the created stacking contexts have the same z-index value, they will render in the order in which they have been added into the hierarchy which means the element that appears later in the HTML will be rendered on top of the element that appears earlier, and we get the following scenario:

The `th` element is under the `td` to the right and the `th` below it.

In order to fix this, we need to increase the z-index of the cell containing the button to be higher than the rest of the cells that are sticky.

This behavior can be observed in the test story `FlatRowHeaderWithNoPaddingAndButtons`


src/components/flat-table/flat-table-row-header/flat-table-row-header.component.tsx
```diff
@@ -74,6 +75,10 @@ export const FlatTableRowHeader = ({
     [isExpandableCell, onClick]
   );
 
+  const handleOnFocus = useCallback((ev: React.FocusEvent<HTMLElement>) => {
```
edleeks87, line 78: Same question as above

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -28,7 +28,7 @@ const firstColumnOldFocusStyling = `
 
 const newFocusStyling = `
   ${addFocusStyling(true)}
-  z-index: 1000;
+  z-index: 1005;
```
edleeks87, line 31: suggestion (non-blocking): it might be better to use a more consistent approach like you have below in `flat-table.style`

```
z-index: ${({ theme }) => theme.zIndex.overlay + 5};

File: src/components/flat-table/flat-table-cell/flat-table-cell.component.tsx [hunk 3]
```diff
@@ -59,8 +59,13 @@ export const FlatTableCell = ({
     isInHighlightedRow,
     isInSelectedRow,
     tabIndex,
+    bringToFront,
   } = useTableCell(internalId.current);
 
+  const handleOnFocus = useCallback((ev: React.FocusEvent<HTMLElement>) => {
```
edleeks87, line 65: question: Is there any performance enhancement from `useCallback` with an empty array here? As we're not passing in the `bringToFront` as a dependency do we need to memoize at all or is the cost in memory overkill here?

https://kentcdodds.com/blog/usememo-and-usecallback

src/components/flat-table/flat-table-row-header/flat-table-row-header.spec.tsx
```diff
@@ -312,15 +318,92 @@ describe("FlatTableRowHeader", () => {
...
+
+        const mockEvent = {
+          nativeEvent: {
+            composedPath: composedPathMockFunction,
+          },
+        };
+
+        component.simulate("focus", mockEvent);
+
+        expect(composedPathMockFunction).toHaveBeenCalled();
```
Parsium, line 340: **question**: given we are testing the z-index increase behaviour in the succeeding test, is asserting the internals of the focus event necessary here?

If this assertion is required for coverage reasons, moving this into a dedicated test would help with readability.

src/components/flat-table/flat-table-cell/flat-table-cell.component.tsx
mihai-albu-sage: This has been refactored and the unnecessary `useCallback` has been removed in order to increase performance.

src/components/flat-table/flat-table-row-header/flat-table-row-header.component.tsx
mihai-albu-sage: This has been refactored and the unnecessary `useCallback` has been removed in order to increase performance.

src/components/flat-table/flat-table-row/flat-table-row.style.ts
mihai-albu-sage: This has been refactored to use the same approach as in `flat-table.style`

src/components/flat-table/flat-table-row-header/flat-table-row-header.spec.tsx
mihai-albu-sage: This has been removed from the original test and a new test has been added with a clear description in order to improve readability.

edleeks87 (approved):

Parsium (approved):

stephenogorman (approved):

PR #6769, opened 2024-06-08
Select: ensure virtualised Select list container always has at least 1 pixel of height

damienrobson-sage:
### Proposed behaviour

`virtualizer.getTotalSize()` returns the total size in pixels for the virtualized items. If nothing is loaded, then the list height is set to 0 (line 21 of src/components/select/select-list/select-list.style.ts) as the method always returns a number and never undefined, it seems. [A recent change in the TanStack/virtual repo](https://github.com/TanStack/virtual/commit/e43e03e500588b2da36f184ecaba2b8e5a506596) suggests that items are never returned regardless if the container height is set to 0, so the proposed change ensures that a container of at least 1px is always available to render into.

### Current behaviour

An edge-case but game-breaking bug can occur whereby the select list can initially be opened and closed but then never re-opens properly. 

The recording [see jira ticket] shows the fix in action via manual means (this is how it was discovered after reading the above commit to the TannStack library)

### Checklist

#### QA


edleeks87 (approved):
I'm happy to provisionally approve this but I think we need to make sure this is rigorously manually tested to make sure none of our stories are broken as a result of the change. @damienrobson-sage whoever QAs this may also want to test this in your app as we aren't able to repro it in storybook or stackblitz so they may message at some point fyi

Parsium (approved):
Thanks for this @damienrobson-sage 👍🏼 

@edleeks87 @robinzigmond, considering the linked TanStack issue, it might be beneficial for us to upgrade `@tanstack/react-virtual` in the future if the bug has been addressed, especially since we are currently on a beta version.

robinzigmond:
> A recent change in the TanStack/virtual repo

I just checked - and we've already got this fix, in fact although sadly I don't remember the details, I think we needed precisely this for a previous fix, and I directly referenced this feature (returning zero items if the height is 0) in the (long) comment on [this commit](https://github.com/Sage/carbon/commit/0ce2847c42bfe79ba34192725af77d8ed6a31c6b).

This is what made me a little nervous about this fix - but since all tests are passing, including those that I added for that fix mentioned above, I think it's probably OK (provided we do some careful manual testing as well).

But you are correct that the package is now out of beta, so I agree that we should take a "real" version as soon as we can, assuming the updates don't cause major versions.

sianford (approved):
Tested all Select stories in Chrome, Safari, Firefox and Edge. Observed the fix in he application. Testing passed.

PR #6770, opened 2024-06-10
Refactor Date unit tests to RTL FE-6566

edleeks87:
### Proposed behaviour

Converts unit tests to RTL

### Current behaviour

Unit tests are enzyme

### Checklist

#### QA


src/components/date/__internal__/navbar/navbar.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  render(<Navbar onPreviousClick={onPreviousClick} onNextClick={jest.fn()} />);
+  const prevButton = screen.getByRole("button", { name: "Previous month" });
+  prevButton.focus();
+  await user.keyboard("{Enter}");
+
+  expect(onPreviousClick).toHaveBeenCalled();
+});
+
+test("should change the current month when ' ' (Space) key is pressed and previous button is focused", async () => {
```
nineteen88, line 162: suggestion(minor, non-blocking): just put space rather than ' '

The enter tests just say enter, so I don't see a reason why we couldn't just say space. Very minor, barely worth the effort of changing at all though.

src/components/date/__internal__/utils.test.ts
```diff
@@ -0,0 +1,448 @@
...
+    "01/01/2022",
+  ]);
+
+  expect(additionalYears("yyyy/MM/dd", "2022/01/01")).toEqual([
+    "yyyy/MM/dd",
+    "2022/01/01",
+  ]);
+});
+
+describe.each(yearValuesLessThan69)(
```
nineteen88, line 234: comment(non-blocking): this seems very wasteful and excessive

Looping through so many dates and making a new describe block for each is a little too much in my opinion. Maybe something we can chat about after this is in and make a ticket to address.

src/components/date/__internal__/utils.test.ts
```diff
@@ -0,0 +1,448 @@
...
+    ])(
+      "returns expected output when format is `%s` and reference date is `%s`",
+      (format, referenceDate, expected) => {
+        expect(additionalYears(format, referenceDate)).toEqual(expected);
+      }
+    );
+  }
+);
+
+describe.each(yearValuesGreaterThan69)(
```
nineteen88, line 273: comment(non-blocking): same as before

Excessive looping through so many dates (more this time)

src/components/date/__internal__/utils.test.ts
nineteen88, line 1: comment(non-blocking): there's several situations where we are over testing the same thing

I think this file could be streamlined a lot but there's already a lot of work involved with this PR so we should probably just create a new ticket to revisit at a future date

src/components/date/date.test.tsx
```diff
@@ -0,0 +1,1419 @@
...
+      await user.click(input);
+      await user.type(input, inputString);
+      await user.tab();
+
+      expect(input).toHaveValue("04/04/2019");
+    }
+  );
+});
+
+test("should update the input value and call `onChange` with expected parameters when the user types a strig with a valid leap year and the input is blurred", async () => {
```
DipperTheDan, line 1165:
```suggestion
test("should update the input value and call `onChange` with expected parameters when the user types a string with a valid leap year and the input is blurred", async () => {
```

src/components/date/__internal__/utils.test.ts
edleeks87: Yeah the describe could be removed but I think we still need to loop or at least agree on a subset of dates to assert as there's specific logic in the component around how it treats abbreviated years. I think it's worth pushing the discussion and potential new ticket outside this PR as I suspect there's other components that involve user input that will have similar approaches to testing etc

src/components/date/__internal__/utils.test.ts
edleeks87: yep there's a lot of redundancy and likely repetition here, it was a choice I made at the time as the component's accuracy is fairly critical in accounting applications. Looking at it again I'm inclined to agree that we can strip some of this back and maybe agree on a set of values to assert that allow us to be comfortable that the component hasn't regressed. I'll raise a ticket to review tests like this, as mentioned above there'll be other inputs that have adopted the same approach (Decimal etc) so we can include that in the discussion and ticket

DipperTheDan (approved):

nineteen88 (dismissed):

DipperTheDan (approved):

PR #6784, opened 2024-06-17
chore(link): convert enzyme to RTL - FE-6605

DipperTheDan:
### Proposed behaviour

Unit tests for Link use RTL 

### Current behaviour

Unit tests currently use Enzyme

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/link/link.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+
+  expect(buttonElement).toBeInTheDocument();
+});
+
+test("when rendered as an `<a>` element, it should set the expected aria attributes", () => {
+  render(<Link aria-label="test" />);
+
+  const linkElement = screen.getByTestId("link-anchor");
+
+  expect(linkElement).toHaveAttribute("aria-label", "test");
```
nuria1110, line 232: **suggestion**: We could use an accessible name assertion here
```suggestion
  expect(linkElement).toHaveAccessibleName("test");
```

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+});
+
+test("when rendered as a `button` element, it should set the aria attributes on the button", () => {
+  render(<Link onClick={() => {}} aria-label="test" />);
+
+  const buttonElement = screen.getByRole("button");
+
+  expect(buttonElement).toHaveAttribute("aria-label", "test");
```
nuria1110, line 240: Same as suggestion above.

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+  render(
+    <Link disabled onClick={spy}>
+      Test Content
+    </Link>
+  );
+
+  const linkElement = screen.getByText("Test Content");
+  await userEvent.click(linkElement);
+
+  expect(spy).not.toHaveBeenCalled();
```
nuria1110, line 58: **suggestion(non-blocking)**: I think you could also assert that it is disabled: `expect(linkElement).toBeDisabled()`

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+  );
+
+  const linkElement = screen.getByTestId("link");
+  const iconElement = screen.getByTestId("icon");
+
+  user.hover(linkElement);
+
+  expect(linkElement).toHaveStyle("color: var(--colorsActionMajor600)");
+  expect(iconElement).toHaveStyle("color: var(--colorsActionMajor600)");
+});
```
nuria1110, line 297: **comment**: I've seen you have done this in previous tests, I would also make this test `async` and use `await` for userEvents as it might cause some unexpected behaviour.  Same for the next few tests.

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,570 @@
...
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import Link from "./link.component";
+import MenuContext from "../menu/menu.context";
+import { baseTheme } from "../../style/themes";
+
+describe("If `isSkipLink` provided", () => {
+  it("should render `Skip to main content` text inside of Link", () => {
```
Parsium, line 13: **comment(non-blocking)**: I assume we need the below style tests for coverage reasons too. If not, could we utilise Playwright or Chromatic? This would offer the benefit of increased confidence in our style tests, as the component would be rendered in an actual browser rather than with JSDOM.

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,570 @@
...
+    </Link>
+  );
+
+  const iconElement = screen.getByTestId("icon");
+
+  expect(iconElement).toHaveStyle("margin-left: var(--spacing100)");
+});
+
+describe("when the `onKeyDown` event is triggered", () => {
+  it("should trigger an `onKeyDown` event", async () => {
```
Parsium, line 180: **suggestion**: might be worth re-wording this slightly so its a bit clearer?

```suggestion
  it("should call onKeyDown callback, when provided and link is triggered via the Enter key", async () => {
```

src/components/link/link.test.tsx
```diff
@@ -0,0 +1,570 @@
+// We've chosen to ignore this for the whole file as this warning fires 15 times.
+// It warns that `... anchors are primarily expected to navigate. Use the button element instead.`
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
Parsium, line 3: **comment(non-blocking)**: we have a ticket on our backlog to remove polymorphism from `Link` as an accessibility improvement. What do you think about referencing this here?

```suggestion
/* TODO: FE-6579 To re-enable once button-related props are removed from Link */
/* eslint-disable jsx-a11y/anchor-is-valid */
```

DipperTheDan: Yeah, I had no idea we already had a ticket would resolve this issue. I can change the comment to that which you have provided.

src/components/link/link.test.tsx
DipperTheDan: Yes we need these for coverage. Something to consider once we have the shared coverage completed though.

src/components/link/link.test.tsx
DipperTheDan: I'll get this amended to what you have provided here. Thanks 😄

Parsium (dismissed):

nuria1110 (approved):

ZhuoyuJin (approved):

Parsium (approved):

ZhuoyuJin (approved):

PR #6795, opened 2024-06-24
feat(step-flow): allow component title to be compositional

### LINKED ISSUE #6732: StepFlow - change title prop type
### Description

Dialog’s title prop can be an Element and we use it to display an icon next to the title when the action is destructive.
We are now introducing StepFlow in Dialogs but StepFlow’s title is only a string.

### Suggested solution

Make StepFlow's title prop accept an Element

### Demo URL

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Confidentiality

tomdavies73:
fix #6732 

### Proposed behaviour

The `StepFlow` component will support a more compositional approach, allowing consumers to pass a node to the `title` prop. This means consumers can render any relevant elements within the node passed to the `title` prop, providing greater flexibility to meet specific design requirements. 

In order to ensure the component still functions, and our accessibility solutions still work, consumers need to render the `StepFlowTitle` sub component. This new sub component will not be noticeable for current consumers just wishing to pass a string to the `title` prop, but can be used and rendered within a custom node to ensure the component functions correctly.

Screenshot 2024-06-28 at 12 30 20

Screenshot 2024-06-28 at 12 30 12

Currently, consumers are restricted to just passing a title via the `title` prop which is limited to being a string. This means if consumers wanted to add additional context to this title via a node, they are unable to do so.

### Checklist

#### QA


src/components/step-flow/step-flow-title/step-flow-title.component.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <Typography
+          fontWeight="900"
+          fontSize="var(--fontSizes600)"
+          lineHeight="var(--sizing375)"
+          variant="span"
+          aria-hidden="true"
+          data-element="visible-title-text"
+        >
+          {titleString}
+        </Typography>
```
DipperTheDan, line 73: **suggestion(non-blocking):** You could move this into it's own variable seeing as you're kind of using the same thing twice. It means in future if we need to change the styles, we do it in one place and not in two. I'm OK with either way though.

```
const TitleTypography = ({ variant, dataElement, children }) => (
  <Typography
    fontWeight="900"
    fontSize="var(--fontSizes600)"
    lineHeight="var(--sizing375)"
    variant={variant}
    aria-hidden="true"
    data-element={dataElement}
  >
    {children}
  </Typography>
);
```
and then use it like:
```
      <TitleTypography variant="span" dataElement="visible-title-text">
        {titleString}
      </TitleTypography>
```

src/components/step-flow/step-flow.mdx
```diff
@@ -46,6 +46,29 @@ An example of how this can be achieved is provided in the **Example implementati
 
 <Canvas of={StepFlowStories.DefaultStory} />
 
+### Step Flow Title node
+
+To allow a compositional approach, the `title` prop accepts a node. This can be used to compose custom content to be passed as the title of the `StepFlow` component.
+However, it is important to note that the `StepFlowTitle` sub component must be a descendant of the passed node, to ensure the component is accessible to screen reader users.
```
DipperTheDan, line 52: **nitpick(non-blocking):** I _think_ subcomponent is all one word. If it's not then I'm demonstrating my ignorance to the world...

```suggestion
However, it is important to note that the `StepFlowTitle` subcomponent must be a descendant of the passed node, to ensure the component is accessible to screen reader users.
```

src/components/step-flow/step-flow-title/step-flow-title.component.tsx
```diff
@@ -0,0 +1,120 @@
...
+  titleVariant?: "h1" | "h2";
+  /** Override the screen reader only title with any additional context or information.
+   * If not provided, only the `titleString` prop will be used.
+   */
+  screenReaderOnlyTitle?: string;
+}
+
+let noContextWarnTriggered = false;
+
+export const StepFlowTitle = ({
```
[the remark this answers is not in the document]
edleeks87: You'll need to update some tests as the above change will mean the hidden text isn't rendered unless the current and total steps are set via the context. I think this is correct as currently this span renders initially empty until the context initialises anyway so it should just not render until the screen reader string can be computed fully etc

src/components/step-flow/step-flow.component.tsx
```diff
@@ -77,6 +81,22 @@ export const StepFlow = forwardRef<StepFlowHandle, StepFlowProps>(
 
     let currentStepWarnTriggered = false;
     let noRefWarnTriggered = false;
+    let noStepFlowTitleWarnTriggered = false;
+
+    const hasStepFlowTitle = (children: React.ReactNode): boolean => {
```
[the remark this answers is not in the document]
tomdavies73: `title` is required, but as we're allowing a node there is no guarantee that consumers will pass `StepFlowTitle`, which they need to in order for the component to be accessible

src/components/step-flow/step-flow.component.tsx
```diff
@@ -146,37 +178,29 @@ export const StepFlow = forwardRef<StepFlowHandle, StepFlowProps>(
...
-            aria-hidden="true"
-            data-element="visible-title-text"
-          >
-            {title}
-          </Typography>
-          <Typography
-            variant="span"
-            data-element="visually-hidden-title-text"
-            screenReaderOnly
+        {typeof title === "string" ? (
```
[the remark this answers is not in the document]
tomdavies73: the `StepFlowTitle` has its own styled focus wrapper to ensure focus is only pulled to the title, and not the entire node. This stops a screen reader announcing everything, instead it just focuses on the wrapper and announces the contents of the screen reader only span

src/components/step-flow/step-flow-title/step-flow-title.component.tsx
DipperTheDan: Even better.

DipperTheDan (dismissed):

edleeks87 (approved):

src/components/step-flow/step-flow.stories.tsx
```diff
@@ -167,6 +198,81 @@ export const ExampleImplementation: Story = () => {
 };
 ExampleImplementation.storyName = "Example Implementation";
 
+export const ExampleImplementationWithTitleNode: Story = () => {
+  const lowestStep = 1;
+  const highestStep = 3;
+
+  const [isOpen, setIsOpen] = useState(false);
```
edleeks87, line 205: comment: sorry @tomdavies73 I've missed this when reviewing and only noticed when I've checked the chromatic build. If we want this to be captured by chromatic we'll need to use [isChromatic](https://github.com/Sage/carbon/blob/ae5079cf2afaff11554a5562e8f0659bb8bcbab0/.storybook/isChromatic.js).

DipperTheDan (approved):

PR #6801, opened 2024-06-26
refactor(password): convert enzyme test to rtl

tomdavies73:
### Proposed behaviour

Converts enzyme tests to rtl.

### Current behaviour

Tests are currently written in enzyme.

### Checklist

#### QA


src/components/password/password.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+import userEvent from "@testing-library/user-event";
+import Password from "./password.component";
+import guid from "../../__internal__/utils/helpers/guid";
+
+const mockedGuid = "guid-12345";
+jest.mock("../../__internal__/utils/helpers/guid");
+
+(guid as jest.MockedFunction<typeof guid>).mockImplementation(() => mockedGuid);
+
+test("should render the `Textbox` component as a child on render", () => {
```
Parsium, line 12: **suggestion(non-blocking)**: since the `Textbox` component being a child is an implementation detail, can we assert that the input box and the 'Show Password' button are visible instead?

I think checking these two elements only should be sufficient since they are what the user will actually notice.

src/components/password/password.test.tsx
Parsium, line 1: **nitpick**: for the tests involving the visibility toggle button, what do you think about amending its name from `ButtonMinor` so its function is clearer?

```tsx
// test("should render the `ButtonMinor` component as a child on render", () => {});
// becomes:
test("should render the 'Show Password' button", () => {});

// test("should change the `Textbox` input type from 'password' to 'text' when the `ButtonMinor` is clicked", async () => {});
// becomes:
test("should change the `Textbox` input type from 'password' to 'text' when 'Show Password' button is clicked", async () => {});
```

src/components/password/password.test.tsx
tomdavies73: Yeah, i'm happy with that. I can change the text descriptions referencing `Textbox` to just reference the input instead

src/components/password/password.test.tsx
tomdavies73: I'll change all mention of `ButtonMinor` to be the `Show Password` or `Hide Password` buttons in the test descriptions

Parsium (dismissed):

src/components/password/password.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+import userEvent from "@testing-library/user-event";
+import Password from "./password.component";
+import guid from "../../__internal__/utils/helpers/guid";
+
+const mockedGuid = "guid-12345";
+jest.mock("../../__internal__/utils/helpers/guid");
+
+(guid as jest.MockedFunction<typeof guid>).mockImplementation(() => mockedGuid);
+
+test("should render a text input on render with type 'password' on render", () => {
```
nuria1110, line 12: **nit-pick(non-blocking):** doubled up on the 'on render'
```suggestion
test("should render a text input with type 'password' on render", () => {
```

nuria1110 (approved):
Looks good to me, minor nit-pick but approved anyways

src/components/password/password.test.tsx
tomdavies73: oops, kept them both to just mention one render now 👍

nuria1110 (approved):

Parsium (approved):

PR #6805, opened 2024-07-01
refactor(anchor-navigation): convert enzyme tests to rtl

robinzigmond:
### Proposed behaviour

AnchorNavigation component should have unit tests written using react-testing-library

### Current behaviour

AnchorNavigation tests use Enzyme

### Checklist

#### QA


src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+import { render, screen, act } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import Textbox from "../textbox";
+import {
+  AnchorNavigation,
+  AnchorNavigationItem,
+  AnchorSectionDivider,
+} from ".";
+import { AnchorNavigationProps } from "./anchor-navigation.component";
```
tomdavies73, line 11: **suggestion**: we can import `AnchorNavigationProps` from the index above, reducing the need for an additional import

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+const SECTION_VISIBILITY_OFFSET = 200;
```
tomdavies73, line 13: **suggestion**: as this is only used once, we can probably just use the numerical value on line `248` reducing the need for a const being defined here

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+const SELECTED_ITEM_STYLES = [
```
tomdavies73, line 15: **suggestion**: similar to @Parsium's comment, I feel we can remove this here and just assert the `backgroundColor` and `borderLeftColor` where appropriate

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+  renderComponent();
+
+  expect(screen.getAllByRole("listitem")[1]).toHaveStyleRule(
+    "background-color",
+    "var(--colorsActionMinor100)",
+    { modifier: "& a:hover" }
+  );
+});
+
+test("renders selected navigation item with proper background when hovered", async () => {
```
tomdavies73, line 336: **question**: are these, and other styling related tests included to meet coverage?

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+    refs.forEach((ref, index) => {
+      if (ref.current) {
+        jest
+          .spyOn(ref.current, "getBoundingClientRect")
+          .mockImplementation(
+            () => ({ top: topEdgeOffsets[index] - scrollPosition } as DOMRect)
+          );
+      }
+    });
```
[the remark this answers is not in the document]
tomdavies73: If we're finding it hard to locate these elements using RTL queries, I wouldn't be opposed to `data-role`'s being added in a previous commit to help find any elements which can't be located otherwise.

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+    "anchor-sticky-navigation"
+  );
+
+  const anchorNavigationLinks = screen.getAllByRole("link");
+  anchorNavigationLinks.forEach((anchor) => {
+    expect(anchor).toHaveAttribute("data-element", "anchor-navigation-item");
+  });
+});
+
+test.each([0, 1, 2, 3, 4])(
```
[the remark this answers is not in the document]
robinzigmond: it's a fair question. I did it this way mainly because that's how we were doing it in the Enzyme tests, and I'd always rather have too many tests than too few if unsure. But I think you're right - I'm happy to change it to just testing one (not the first one as that's pre-selected so the test might still pass even if this got broken).

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+  ["border-left-color", "var(--colorsActionMajor500)"],
+];
+
+interface ContentProps {
+  title?: string;
+  noTextbox?: boolean;
+  tabIndex?: number;
+  ["data-role"]?: string;
+}
+const Content = React.forwardRef<HTMLDivElement, ContentProps>(
```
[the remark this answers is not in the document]
robinzigmond: again, I mostly just kept this from before without thinking about it too much - as it doesn't appear to be doing nothing. Looking again, I agree that it's adding complexity that we'd be better off without - I can certainly simplify some of it (eg the `noTextbox` prop doesn't seem to be used at all), and I'll see if it's possibly to remove it entirely.

src/components/anchor-navigation/anchor-navigation.test.tsx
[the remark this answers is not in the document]
robinzigmond: fair points - I'll see what I can do

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+    // toHaveStyle still passes even with incorrect styles when used with CSS variables
+    // (see https://github.com/testing-library/jest-dom/issues/461 - the variables seem to be
+    // treated by js-dom as "invalid values" as explained in the first reply) - so using toHaveStyleRule instead
+    SELECTED_ITEM_STYLES.forEach(([property, value]) => {
+      expect(screen.getAllByRole("listitem")[i]).toHaveStyleRule(
+        property,
+        value,
+        { modifier: "& a" }
+      );
+    });
```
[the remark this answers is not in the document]
robinzigmond: Thanks for the suggestion, which I agree with. It ended up this way because I was having a lot of trouble with these style tests and didn't want to have to rewrite a bunch of tests each time I changed it - but yes as it ended up there's no real gain to having this `forEach` rather than just having the 2 explicit assertions each time.

src/components/anchor-navigation/anchor-navigation.test.tsx
robinzigmond: yeah, I missed this as I didn't even look at the imports (other than the bits I had to change of course)

src/components/anchor-navigation/anchor-navigation.test.tsx
robinzigmond: not so sure I agree with this one - "magic numbers" are considered a bad thing, the const is better even if it's only used once, as it's easier for readers of this code to see what's going on (rather than seeing a hard-coded 200 in a test and wondering what it means), and also easier for future maintenance to know which value to update if UI changes ever mean this has to be updated, rather than hunting for a particular number in a particular test which needs updating.

Happy to hear counter-arguments if I've missed something though!

src/components/anchor-navigation/anchor-navigation.test.tsx
robinzigmond: Yeah I agree with this one (as already noted in reply to James's comment mentioning this), as contrary to the `SECTION_VISIBILITY_OFFSET` it should be fairly clear what these styles mean when they're used in a test

src/components/anchor-navigation/anchor-navigation.test.tsx
robinzigmond: An equivalent test was there in the Enzyme tests, and I hadn't thought too much about it as it wasn't too hard to convert this one to RTL. I don't know if it's needed for coverage - I doubt it to be honest - but I quickly checked the Playwright tests and they don't appear to test anything to do with hovering, so we should have this in imo. It might be better to add this in playwright, but given that this was in the unit tests before and the time pressure we're under I'd prefer to leave these in, we can always decide to change tests later after a thorough review of what should be tested in each place. (Particularly if we can get combined coverage working.)

src/components/anchor-navigation/anchor-navigation.test.tsx
robinzigmond: I suppose the const definition could be moved to be just inside that one test, since it's only used once - I'd be perfectly happy with that.

Parsium (approved):

tomdavies73 (approved):

PR #6811, opened 2024-07-02
feat(progress-tracker): remove aria attribute and showDefaultLabel props

### LINKED ISSUE #6755: Screenreader isn't reading ProgressTracker description
### Description

I am using the progress tracker component as part of a custom component (see ticket SBS-89576)

The existing aria-labels that are by default on progress tracker shouldn't be there according to external discussion with 
the Accessibility team.

Screenshot below shows issue - screenreader is not reading the text but is instead reading the progress bar itself, which is undesirable

image

Carbon example
image

### Reproduction

https://carbon.sage.com/?path=/story/progress-tracker--size-small

### Steps to reproduce

(On carbon) Open Voiceover, select progress tracker component, then CTRL + OPTION + SHIFT + Right arrow key (Mac)

SBS-89576

### Suggested solution

Get rid of default-set aria-labels on component, set component to auto add description text to aria-label, but have that overridden by any custom aria-label set as a prop
### Carbon version

133.1.0 but works on latest storybook

### Design tokens version

4.34.0

### Relevant browsers

Chrome, Safari

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #6755 

### Proposed behaviour

- "aria-label", "aria-describedby", "aria-valuenow", "aria-valuemin", "aria-valuemax" and "aria-valuetext" props have been removed. 
- `showDefaultLabels` prop has been removed, meaning `ProgressTracker` can no longer be used without labels. 
- Screen reader announces visible label. 
</img>
- Updated docs and storybook examples. 
- Width of the inner bar is now the percentage value of `progress` and will scale depending on the length of the bar.
- Instead of using the `ProgressTracker` component, `FileUploadStatus` subcomponent now only uses the visual progress bar. 
- Move `Message` subcomponents into an `__internal__` directory.

### Current behaviour

- "aria-label", "aria-describedby", "aria-valuenow", "aria-valuemin", "aria-valuemax" and "aria-valuetext" props are deprecated.
- The component can be used without label. 
- Screen reader does not announce visible label when set. 
</img>
- Docs and storybook examples are out-of-date or inaccurate. 
- Inner bar width set by a `calc()` function depending on `length` and `progress` values.
- `Message` subcomponents are in a public directory.

### Checklist

#### QA

### Testing instructions

- Should have no significant visual regressions, some small changes in inner-bar width to be expected. 
- Storybook examples should match what is described.
- Check visible label is announced by screen reader. 
- Visual progress bar should be ignored by screen reader.
- Default label should always be displayed unless a valid `currentProgressLabel` is provided. 
- Check no regressions in file input with progress tracker.


src/components/progress-tracker/progress-tracker.stories.tsx
```diff
@@ -77,9 +77,30 @@ ColorVariants.storyName = "Color Variants";
 export const DefaultLabelValue: Story = () => {
   return (
     <Box display="flex" flexDirection="column" alignItems="center">
-      <ProgressTracker mt={2} progress={15} currentProgressLabel="15%" />
-      <ProgressTracker mt={2} progress={50} currentProgressLabel="50%" />
-      <ProgressTracker mt={2} progress={100} currentProgressLabel="100%" />
+      <ProgressTracker mt={2} progress={15} showDefaultLabels />
```
edleeks87, line 80: question: if we're having to set `showDefaultLabels` in most of these stories is it worth setting it so it's true by default? @harpalsingh may be a question for you tbh

edleeks87: If it should remain false by default I think it's wroth making sure we have a story that is capture by chromatic that has no labels at all

src/components/progress-tracker/progress-tracker.style.ts
```diff
@@ -111,16 +111,16 @@ const StyledValuesLabel = styled.span<
 `;
 
 const InnerBar = styled.span<
-  Pick<ProgressTrackerProps, "progress" | "size" | "length" | "error">
+  Pick<ProgressTrackerProps, "progress" | "size" | "error">
```
edleeks87, line 114: Comment (non-blocking): by extending the set of the `ProgressTrackerProps` interface here you're actually saying `progress` can potentially be undefined. It might be worth making it `Required` in this interface as it will never not be a number given it's defaulted in the component prop destructuring and updating this type would mean it would complain if anyone removed that default inadvertently as in that scenario it could end being `undefined%` in the styles. The same is true for `size` and `error` as they have values by the time they are passed to the styled-component we could do the following:
```
Required<Pick<ProgressTrackerProps, "progress" | "size" | "error">>

File: src/components/progress-tracker/progress-tracker.stories.tsx (hunk 1 above)
Comment (nuria1110) [reply src/components/progress-tracker/progress-tracker.stories.tsx:80]: I added this prop to most stories so that we didn't encourage using this component without labels, and I do think having this as true by default would be ideal but will let you/Harpal decide on that. 

I think that the `screenReaderOnlyLabel` story I added should be capturing the component without labels anyways.

File: src/components/progress-tracker/progress-tracker.component.tsx [hunk 3]
```diff
@@ -75,40 +54,12 @@ const ProgressTracker = ({
...
-  }, []);
-
-  useLayoutEffect(() => {
-    updateBarLength();
-  }, [updateBarLength]);
-
-  useResizeObserver(barRef, () => {
-    updateBarLength();
-  });
-
```
Parsium, line 111: **praise**: the new CSS approach of ensuring the inner bar length is proportional to the progress value is much simplier. Good spot!

src/components/progress-tracker/progress-tracker.mdx
```diff
@@ -57,61 +56,48 @@ This is an example of a large `ProgressTracker` component.
 
 ### Custom bar length
 
-The length of the bar can be set to any valid css property. This sets the width of a horizontal bar or the height of a vertical bar. The default is "256px".
+By default, the length of the component is "256px". However, you can use the `length` prop to override this with any valid css string.
```
Parsium, line 59: **praise**: the wording of the docs is a lot more cohesive :+1:

Parsium (commented):
Nice work on this @nuria1110 👏🏼 Nothing else to add other than what @edleeks87 has mentioned 👍🏼

src/components/progress-tracker/progress-tracker.style.ts
```diff
@@ -111,16 +111,16 @@ const StyledValuesLabel = styled.span<
 `;
 
 const InnerBar = styled.span<
-  Pick<ProgressTrackerProps, "progress" | "size" | "length" | "error">
+  Required<Pick<ProgressTrackerProps, "progress" | "size" | "error">>
 >`
-  ${({ progress, size = "medium", length, error, theme }) => css`
+  ${({ progress, size = "medium", error, theme }) => css`
```
edleeks87, line 116: Apologies I wasn't clear here, but we don't need to default the `size` prop here anymore (it's defaulted in the component) now we've removed the potential for it to be undefined

edleeks87 (approved):
Approving for now but let's wait for @harpalsingh to steer on the prop value before we merge etc

Parsium (approved):

harpalsingh:
@edleeks87 Looking at the design I feel it should always have a label and if you see the DS guidance it also says that it should always have a label and never without one. https://zeroheight.com/2ccf2b601/p/00f4d0-progress-tracker

src/components/progress-tracker/progress-tracker.mdx
```diff
@@ -57,61 +55,37 @@ This is an example of a large `ProgressTracker` component.
...
-This is an example of a `ProgressTracker` component with `currentProgressLabel` and `maxProgressLabel` props.
+To set your own labels, you can set the `currentProgressLabel` and `maxProgressLabel` props. 
 
-<Canvas of={ProgressTrackerStories.CustomLabelValues} />
-
-It is also possible to render the component with a mix of default and custom labels by only setting one of the `currentProgressLabel`
-and `maxProgressLabel` props and optionally setting the `showDefaultLabels` prop.
-
-<Canvas of={ProgressTrackerStories.DefaultAndCustomLabelValues} />
+Note that the preposition between the current and max labels will be 'of' by default, but you can override this by setting the `customValuePreposition` prop and making use of the `of` translation key.
```
Parsium, line 73: **suggestion(non-breaking)**: this might make it clearer that consumers don't need to do both :)

```suggestion
Note that the preposition between the current and max labels will be 'of' by default, but you can override this by setting the `customValuePreposition` prop or making use of the `of` translation key.
```

src/components/progress-tracker/progress-tracker.component.tsx
```diff
@@ -45,48 +47,44 @@ const ProgressTracker = ({
...
-      return showDefaultLabels ? defaultValue : undefined;
-    };
-
-    const displayedCurrentProgressLabel = label(
-      currentProgressLabel,
-      `${progress}%`
-    );
-
-    const displayedMaxProgressLabel = label(maxProgressLabel, "100%");
+    if (currentProgressLabel) {
```
edleeks87, line 71: Comment: I think it's worth adding at least one story that demos the permutations we can now have as a result of this code (apologies if we already have it I may have missed it), it will also be easier for UX to review etc

nuria1110: There is an example of these in the custom label story, and then the default label is present in the rest 👍

Parsium (approved):

edleeks87 (approved):

Parsium (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6812, opened 2024-07-03
refactor(date-range): convert unit tests to RTL

edleeks87:
### Proposed behaviour

Convert unit tests to RTL

### Current behaviour

Unit tests are enzyme
### Checklist

#### QA


src/components/date-range/date-range.test.tsx
```diff
@@ -0,0 +1,1125 @@
...
+      onChange={() => {}}
+      endLabel="end"
+      value={["10/10/2016", "11/11/2016"]}
+      startRef={ref}
+    />
+  );
+  unmount();
+
+  expect(ref.current).toBeNull();
+});
```
Parsium, line 924: **question**: are these tests asserting a ref is cleared required for coverage? If not, I'm wondering if asserting this is necessary given its a core React feature?

src/components/date-range/date-range.test.tsx
```diff
@@ -0,0 +1,1125 @@
...
+    </CarbonProvider>
+  );
+
+  const start = screen.getByRole("textbox", { name: "start" });
+  await user.hover(start);
+  const tooltips = screen.getAllByRole("tooltip");
+
+  expect(tooltips).toHaveLength(1);
+  expect(tooltips[0]).toBeVisible();
+  expect(tooltips[0]).toHaveTextContent("start error");
```
Parsium, line 995: **suggestion(non-breaking)**: what do you think about querying the tooltip directly? This would help improve readability 

```suggestion
  const tooltip = screen.getByRole("tooltip", { name: "start error" });

  expect(tooltip).toBeVisible();
```

edleeks87: Yep good point not sure why I've deviated from this approach here tbh

src/components/date-range/date-range.test.tsx
edleeks87: I'm not sure tbh, they are just part of the set of test @robinzigmond added when he worked on adding support to our components so I've just taken the approach of replicating them whenever I've refactored the tests

robinzigmond: it's a fair point. I don't even remember adding these tbh 😅 . While I think the value of similar tests for refs in other components is maybe questionable, I think here it looks like there's something worth testing as `startRef` is one of our own props, not something built-in to React, so it seems fair to test that the ref ends up being added to the right element. When in doubt I'd rather have too many tests than too few.

But definitely not a disaster if this is removed and isn't needed for coverage, so I won't insist on it.

src/components/date-range/date-range.test.tsx
```diff
@@ -0,0 +1,1125 @@
...
+  render(
+    <DateRange
+      startLabel="start"
+      endLabel="end"
+      value={["2016-10-10", "2016-11-11"]}
+      onChange={() => {}}
+    />
+  );
+
+  expect(screen.getByRole("textbox", { name: "start" })).toHaveValue(
```
robinzigmond, line 27:
```suggestion
  expect(screen.getByLabelText("start")).toHaveValue(
```

**suggestion** (non-blocking): this is a more concise way to get the same thing - which obvious applies to more or less every time you're querying for specific date inputs. Not a big deal though.

robinzigmond (commented):
LGTM, I've left one comment but it's not major. Just address @Parsium's comment(s) and I'll happily approve

src/components/date-range/date-range.test.tsx
edleeks87: It's a difficult one to judge tbh, as the way the [docs](https://testing-library.com/docs/queries/about/#priority) are written it suggests that `getByRole` takes priority over the other queries. It does say that `getByLabelText` is ideal for testing form fields but as we're wrapping the inputs in some additional features and functionality I'm leaning towards using `getByRole` to confirm the component is part of the accessibility tree and to remain consistent with the approach we've adopted in other test files.

robinzigmond: fair enough, it's not a huge thing and on reflection you're probably right anyway 👍

robinzigmond (approved):

Parsium (approved):

PR #6813, opened 2024-07-03
refactor(textbox): convert enzyme tests to react-testing-library

robinzigmond:
### Proposed behaviour

Textbox unit tests should use react-testing-library

### Current behaviour

Textbox tests use Enzyme

### Checklist

#### QA


src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+
+testStyledSystemMargin(
+  (props) => <Textbox {...props} />,
+  undefined,
+  (component) => component.find(FormFieldStyle),
+  { modifier: "&&&" }
+);
+
+describe(`when the characterLimit prop is passed`, () => {
+  test.each([2, 3, 4])("renders a character counter", (characterLimit) => {
```
nineteen88, line 55: suggestion(non-blocking): if a child of a describe `it` should be used instead of `test`

Describe blocks combined with it blocks should read as continuous prose to describe the test. Converting the above (and the other `test` blocks) to an `it` would work

src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+
+    expect(
+      screen.getByText("1 character too many", {
+        selector: '[aria-hidden="true"]',
+      })
+    ).toHaveStyleRule("color", "var(--colorsSemanticNegative500)");
+  });
+});
+
+describe("refs", () => {
```
nineteen88, line 106: suggestion(non-blocking): this describe block isn't adding much value

We should probably either be more verbose with the description or drop the describe block

src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+      >
+        normal children
+      </Textbox>
+    );
+    await user.click(screen.getByTestId("input-icon-toggle"));
+    expect(onMouseDown).not.toHaveBeenCalled();
+  }
+);
+
+describe("validation icon", () => {
```
nineteen88, line 288: suggestion(non-blocking): this describe block isn't adding much value

We should probably either be more verbose with the description or drop the describe block

src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+  test("when the align prop is 'right' then 'flex-direction' should be 'row'", () => {
+    const prefixValue = "bar";
+    render(<Textbox value="foo" prefix={prefixValue} align="right" />);
+    expect(screen.getByRole("presentation")).toHaveStyle({
+      flexDirection: "row",
+    });
+  });
+});
+
+describe("required", () => {
```
nineteen88, line 362: suggestion(non-blocking): this describe block isn't adding much value

We should probably either be more verbose with the description or drop the describe block

src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+  const Component = () => <div>positionedChildren content</div>;
+  render(<Textbox positionedChildren={<Component />} />);
+  const positionChildren = screen.getByText("positionedChildren content");
+  const input = screen.getByRole("textbox");
+  expect(positionChildren.compareDocumentPosition(input)).toEqual(
+    Node.DOCUMENT_POSITION_FOLLOWING
+  );
+});
+
+describe("aria attributes", () => {
```
nineteen88, line 386: suggestion(non-blocking): this describe block isn't adding much value

We should probably either be more verbose with the description or drop the describe block.

src/components/textbox/textbox.test.tsx
```diff
@@ -0,0 +1,607 @@
...
+      <CarbonProvider validationRedesignOptIn>
+        <Textbox labelHelp="labelHelp" inputHint="inputHint" error="foo" />
+      </CarbonProvider>
+    );
+    expect(screen.getByText("inputHint")).toBeInTheDocument();
+    expect(screen.queryByText("labelHelp")).not.toBeInTheDocument();
+  });
+});
+
+describe("new validations", () => {
```
nineteen88, line 550: suggestion(non-blocking): this describe block isn't adding much value

We should probably either be more verbose with the description or drop the describe block

src/components/textbox/textbox.test.tsx
robinzigmond: ah OK, thanks for this - I'd looked at other PRs and thought that a decision had been reached to change all `it`'s to `test` so had just been blindly doing this everywhere!

src/components/textbox/textbox.test.tsx
robinzigmond: fair comment on this, and the other similar ones - I've already removed/collapsed a lot of `describe`s and left what I thought was a sensible minimum, but happy to look again and get rid of more of them.

edleeks87: Agree massive improvement on what we had before but I think we can flatten these one's further and just use `test` here

src/components/textbox/textbox.test.tsx
edleeks87: I get what @nineteen88 is saying, I'm not overly fussed with what we use inside the describes tbh, it makes more sense to follow the `describe -> it` pattern but I don't think the tests are unreadable or the same meaning can't be gathered from them using `describe ->  test` so happy to defer to you on whether you update

edleeks87 (approved):

nineteen88 (approved):

PR #6815, opened 2024-07-04
refactor(note): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert tests to RTL.

### Current behaviour

Tests are in enzyme.

### Checklist

#### QA


src/components/menu/menu.component.tsx
```diff
@@ -80,5 +80,4 @@ export const Menu = ({ menuType = "light", children, ...rest }: MenuProps) => {
   );
 };
 
-export { MenuContext };
```
edleeks87, line 83: comment: good spot, annoyingly I missed this when I moved all the contexts to internal folders. Unfortunately this will need to be a breaking change as it's exported as part of the public facing components, so lets not remove it from this file at this time and batch it with other trivial breaking changes. We can leave the path updates that you've made and then remove this when we have some more housekeeping to do

src/components/note/__internal__/status-icon/status.spec.tsx
edleeks87, line 1: comment (non-blocking): we might need to add some PW tests for this if there aren't any already as we have no way of capturing any regressions with chromatic yet. Happy for this to just be a new ticket for now and added in a new PR

src/components/note/note.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+        timeStamp: "23 May 2020, 12:10 PM",
+      }}
+    />
+  );
+
+  const statusText = screen.getByText("Edited");
+  expect(statusText).toBeVisible();
+  await user.hover(statusText);
+
+  const tooltip = await screen.findByText("23 May 2020, 12:10 PM");
```
edleeks87, line 62: suggestion:

```
  const tooltip = await screen.findByRole("tooltip", {name: "23 May 2020, 12:10 PM"});

File: src/components/note/__internal__/status-icon/status.spec.tsx
Comment (nuria1110) [reply src/components/note/__internal__/status-icon/status.spec.tsx:1]: Happy to add a couple PW tests in this PR to cover this

File: src/components/note/note.test.tsx [hunk 3]
```diff
@@ -0,0 +1,162 @@
...
+test("should render with provided `title` prop", () => {
+  render(
+    <Note
+      createdDate="23 May 2020, 12:08 PM"
+      noteContent={EditorState.createEmpty()}
+      title="Title"
+    />
+  );
+
+  expect(screen.getByRole("banner")).toHaveTextContent("Title");
```
Parsium, line 30: **praise**: nice job for looking into the implicit roles of elements :+1:

src/components/note/__internal__/status-icon/status.spec.tsx
nuria1110: Having had a better look at this, they seem to be styles applied on the status.text so shouldn't these styles be captured in the 'With Status' story?

src/components/note/note.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+      noteContent={EditorState.createEmpty()}
+      previews={previews}
+    />
+  );
+
+  const links = screen.getAllByRole("link");
+
+  expect(links).toHaveLength(2);
+  expect(links[0]).toHaveAttribute("href", "https://example.com");
+});
```
Parsium, line 83: **suggestion**: since React handles fragments and arrays identically in this case, we could remove this test to simplify the code. 

```suggestion
```

src/components/note/note.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+  );
+
+  await user.click(screen.getByRole("button", { name: "actions" }));
+
+  expect(screen.getByRole("button", { name: "Copy" })).toBeVisible();
+  expect(screen.getByRole("button", { name: "Edit" })).toBeVisible();
+});
+
+test("should throw when `inlineControls` is not an instance of `ActionPopover`", () => {
+  const spy = jest.spyOn(console, "error").mockImplementation(() => {});
```
Parsium, line 129: **praise**: keeping this localised where its needed will make sure we don't accidentially mask any legitimate console errors :+1:

src/components/note/note.test.tsx
nuria1110: For some reason the component handles different preview types differently so both tests are needed for coverage [here](https://github.com/Sage/carbon/blob/e08800c88a352396118a500763048b982017e80e/src/components/note/note.component.tsx#L108)

Parsium: Ah missed we were using child iteration in the code 👍🏼

Parsium (approved):

src/components/note/__internal__/status-icon/status.spec.tsx
edleeks87: Yep you're right sorry I misread it as testing the styles on the tooltip which won't be shown in chromatic but looking again it's on the text element so we can leave it 👍

edleeks87 (approved):

PR #6816, opened 2024-07-04
refactor(sidebar): convert enzyme tests to RTL

Parsium:
### Proposed behaviour

- Rewrite jest tests to use RTL

### Current behaviour

- All jest tests are written with Enzyme

### Checklist

#### QA


src/components/sidebar/sidebar.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+import {
+  testStyledSystemPadding,
+  testStyledSystemWidth,
+} from "../../__spec_helper__/__internal__/test-utils";
+import CarbonProvider from "../carbon-provider";
+import Form from "../form";
+
+import Sidebar, { SidebarProps } from ".";
+
+const AllTheProviders = ({ children }: { children: React.ReactNode }) => (
```
robinzigmond, line 18: **comment**: I really don't like the name `AllTheProviders` when there is only *one* provider involved here!

Further, it doesn't apepar to me that this particular wrapping is needed for any tests other than those that use the new validation pattern (which I'm not sure any of them do, having looked through all the tests in this file - none appear to use any input components at all). So it doesn't seem right to me that all the tests use this, if it's not necessary.

Finally, even if this is needed, and is renamed, since it's not particularly verbose I'd prefer to just use it explicitly in the `render` function: `render(<CarbonProvider validationRedesignOptIn><Sidebar {{...props}}/>{children}</Sidebar></CarbonProvider>`. I realise that is just a matter of opinion but I find it more readable, rather than having to go to the docs to figure out what `wrapper` is doing.

EDIT: I only just noticed this is also used in the `Dialog` tests, so perhaps it's better for consistency. I'm still quite unconvinced it's necessary or helpful though. Maybe there was a discussion I missed at the time of `Dialog` being converted to RTL?

src/components/sidebar/sidebar.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+  <CarbonProvider validationRedesignOptIn>{children}</CarbonProvider>
+);
+
+beforeEach(() => {
+  jest.useFakeTimers();
+});
+
+afterEach(() => {
+  jest.runOnlyPendingTimers();
```
robinzigmond, line 27: **question**: is it necessary to run the pending timers after each test? I was curious about this, finding no use of `setTimeout` or `setInterval` in any related component other than in `FocusTrap`, where it's used in the `setElementFocus` method, where it wouldn't feel right to wait for this *after* a test rather than during it.

Because of this, I tried running all the `sidebar` tests with this line commented out, expecting at least some of them to fail - but they all passed. I'd rather not include this line unless we know it's necessary.

src/components/dialog/dialog.test.tsx
```diff
@@ -15,6 +15,12 @@ const AllTheProviders = ({ children }: { children: React.ReactNode }) => (
   <CarbonProvider validationRedesignOptIn>{children}</CarbonProvider>
 );
 
+beforeEach(() => jest.useFakeTimers());
+afterEach(() => {
+  jest.runOnlyPendingTimers();
```
robinzigmond, line 20: **question**: same question here as I asked about `Sidebar`.

src/components/sidebar/__internal__/sidebar-header/sidebar-header.spec.tsx
robinzigmond, line 1: **question** (non-blocking): I'm sure it's probably OK, but just to note that I'm not comfortable with removing all these tests unless we're 100% sure they're being covered and properly tested elsewhere. Is Chromatic definitely catching these, or are there Playwright tests for these styles on the SidebarHeader? (I haven't checked in detail, you'll know better than me!)

src/components/sidebar/sidebar.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+  jest.useRealTimers();
+});
+
+test("sidebar element has necessary dialog properties", () => {
```
robinzigmond, line 31:
```suggestion
test("sidebar element has 'aria-modal' attribute set to 'true'", () => {
```

**suggestion** (non-blocking): I prefer this more explicit wording since we're only asserting against that single property. I guess you left it generic in case we later add more, but I'd prefer to only change the description at that point while testing for more properties (or alternatively to add a new, equally explicit, test). I think it's more obvious what the test is doing, both when reading the code and when seeing the test description if it happens to fail, when it's explicit.

It would also make it more clear what the difference is between this and the following test, where setting `enableBackgroundUI` makes `aria-modal` have the value "false", and where you've been explicit about what's happening in that one.

src/components/sidebar/sidebar.test.tsx
Parsium: I'm happy to move the context provider into the render function of each test, especially since we are actively improving test reliability 👍🏼 

We do need the provider here though, mainly for supplying the [`TopModalContext` that `Sidebar` needs](https://github.com/Sage/carbon/blob/master/src/components/sidebar/sidebar.component.tsx#L141):

```tsx
// line 141
// hook consumes `TopModalContext` provided by `CarbonProvider`
const isTopModal = useModalAria(sidebarRef); 
```

Though the `validationRedesignOptIn` flag isn't necessarily required here.

src/components/sidebar/sidebar.test.tsx
Parsium: I included the timer flushing because we use React Transition Group for handling CSS transitions, which relies on timers for state updates. 

Given that it’s a third-party package, I felt it was a best practice to flush any potential running timers. This precaution helps prevent any unexpected behavior from scheduled tasks leaking into other tests.

src/components/sidebar/sidebar.test.tsx
robinzigmond: Ah yes, thanks - I should have realised that, since I was the one who added the `TopModalContext` into `CarbonProvider` 🤦‍♂️ 

In that case yes the provider is needed, but I agree to drop the `validationRedesignOptIn` and maybe (optionally, but I think it could be helpful to future readers) add a comment that this provider has been added specifically for the `TopModalContext`

src/components/sidebar/sidebar.test.tsx
robinzigmond: That's a good point that I hadn't thought of. Happy to do this if it's a general best-practice.

src/components/sidebar/__internal__/sidebar-header/sidebar-header.spec.tsx
Parsium: I decided to remove this since the styles are covered by a Chromatic snapshot. We have a [story where the close icon is rendered](https://www.chromatic.com/component?appId=5ecf782fe724630022d27d7d&csfId=sidebar&buildNumber=12383&k=668d5ebba0574e1e74880855-desktop-snapshot-true&h=6&b=-2), and a [story where the header has custom padding](https://www.chromatic.com/component?appId=5ecf782fe724630022d27d7d&csfId=sidebar--with-header-and-footer-padding&buildNumber=12383&k=668d5ebba0574e1e7488085c-desktop-snapshot-true&h=7&b=-3).

robinzigmond (dismissed):

src/components/sidebar/sidebar.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+});
+
+test("calls a forwarded ref callback on render with sidebar element as argument", () => {
+  const ref = jest.fn();
+  render(<Sidebar open header="My sidebar" ref={ref} />);
+
+  expect(ref).toHaveBeenNthCalledWith(1, screen.getByRole("dialog"));
+});
+
+test("ensures overflowing content is scrollable", () => {
```
edleeks87, line 287: Question (non-blocking): is this needed for coverage, this feels like something we can capture in chromatic?

src/components/sidebar/sidebar.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+  render(<Sidebar open />);
+
+  const sidebarContent = screen.getByTestId("sidebar-content");
+  expect(sidebarContent).toHaveStyle("overflow: auto");
+});
+
+test("does not control styling of overflowing content when there is a child Form with a sticky footer", () => {
```
edleeks87, line 294: same question as above

src/components/sidebar/sidebar.test.tsx
Parsium: Yeah, that's right. It's mainly because we have some logic to delegate the responsibility of handling overflow to a child `Form`.

robinzigmond (approved):

src/components/sidebar/sidebar.test.tsx
edleeks87: This test could probably be written in a more RTL way by adding a few elements and then asserting some aren't visible. We could rely on chromatic to capture the scrollbar and styling. I'm happy for this to be done in a new ticket as we have plenty of tests like this across the code base etc

edleeks87 (approved):

src/components/sidebar/sidebar.test.tsx
robinzigmond: While I haven't tried it, I'd be surprised if this approach works in a simple way - I don't think RTL's `toBeVisible` takes into account whether things are in or out of the viewport. Even with Cypress and Playwright we've had to write custom utils for testing this, and that will be harder in js-dom as I'm pretty sure it makes no attempt to make things like `getBoundingClientRect` work properly.

I could be wrong though, maybe RTL has some magic baked in to do viewport checking.

PR #6818, opened 2024-07-05
refactor(button-minor): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert tests to RTL.

### Current behaviour

Tests are in enzyme.

### Checklist

#### QA


src/components/button-minor/button-minor.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  expect(screen.getByRole("button", { name: "foo" })).toBeVisible();
+});
+
+test("should render disabled button when `disabled` is true", () => {
+  render(<ButtonMinor disabled>foo</ButtonMinor>);
+
+  expect(screen.getByRole("button", { name: "foo" })).toBeDisabled();
+});
+
+// coverage
```
robinzigmond, line 19: **comment (non-blocking)**: I'm not a fan of putting a `coverage` comment on all these tests. While I'm sure they are needed for coverage, they're also testing the styles of various button types etc, and that's not without value. I think the tests are fine and I'd just remove these comments.

src/components/button-minor/button-minor.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  expect(button).toHaveStyleRule("background", "var(--colorsActionMinor500)");
+  expect(button).toHaveStyleRule(
+    "border-color",
+    "var(--colorsActionMinorTransparent)"
+  );
+  expect(button).toHaveStyleRule("color", "var(--colorsActionMinorYang100)");
+});
+
+// coverage
+test("should render with expected styles when `buttonType` is 'tertiary'", () => {
```
robinzigmond, line 41: **comment (non-blocking)**: I'm puzzled that we're testing `primary` and `tertiary`, but not `secondary`. Especially when `secondary` is the default. Could we not add another identical test but against the styles for a secondary button, without the `buttonType` prop being passed. (And optionally do the same but with the prop explicitly set to secondary - that way we know we're testing all values even if the default ever changes.)

src/components/button/button.test.tsx
```diff
@@ -159,6 +159,30 @@ test("sets the correct height when 'services' is passed to the 'buttonType' prop
   expect(icon).toHaveStyle("height: 6px");
 });
 
+/* Styling test for coverage */
```
robinzigmond, line 162: **comment** (non-blocking): I assume these are needed now because they were previously being covered by tests for `ButtonMinor` that you've removed.

I'm happy for these tests to be added to `Button` (unless they're already covered by playwright tests - which they might be), but I would prefer to also:
1) add similar tests explicitly to `ButtonMinor` so that we know for sure that the `size` and `iconType` props from ButtonMinor is getting passed down to Button
2) I'd personally remove the "for coverage" comments as I don't think those are adding very much

src/components/button-minor/button-minor.test.tsx
nuria1110: I think we started adding these to let reviewers know that these are needed for coverage as we had to clarify during reviews, specially with these style tests. These might also be useful once we have combined coverage with PW to know which tests can be removed/should be done in PW/chromatic.

src/components/button-minor/button-minor.test.tsx
nuria1110: As these are mainly for coverage and styles are being captured in chromatic, I didn't think that adding a test `secondary` variant would add much value, same with the `medium` size. Happy to add the tests if the second reviewer also thinks they should be included.

src/components/button/button.test.tsx
nuria1110: Same as above, as these are mainly for coverage and styles are captured chromatic so I didn't think it was worth doubling up.

src/components/button-minor/button-minor.test.tsx
nuria1110: Though this is something that could be discussed with the team as at the moment I think this is mostly preference.

Parsium (commented):
Nice work on this, @nuria1110! 👍🏼

Regarding @robinzigmond's point about the style tests, I understand his concerns. There is a risk we might inadvertently approve Chromatic snapshot changes that could reduce coverage for certain props and styles. For now, let's follow Robin's suggestions and plan a team discussion to revisit the pros and cons.

Parsium (approved):

robinzigmond (approved):

PR #6821, opened 2024-07-08
feat(portrait, profile): deprecate the gravatar pop

tomdavies73:
### Proposed behaviour

Deprecates the `gravatar` prop

### Current behaviour

Currently no deprecation warning exists for the `gravatar` prop

### Checklist

#### QA


src/components/portrait/portrait.component.tsx
```diff
@@ -82,9 +85,17 @@ const Portrait = ({
...
   const { roundedCornersOptOut } = useContext(RoundedCornersOptOutContext);
   const defaultShape = roundedCornersOptOut ? "square" : "circle";
 
+  if (isGravatarLoaded && !deprecatedGravatarWarnTriggered) {
+    deprecatedGravatarWarnTriggered = true;
+    Logger.deprecate(
+      "The `gravatar` prop has been deprecated and will soon be removed."
+    );
+  }
+
```
Parsium, line 98: **suggestion**: since [`Logger.deprecate` will be excluded from a consumer's production build](https://github.com/Sage/carbon/blob/master/src/__internal__/utils/logger/index.ts#L3), this new state won't control anything, so I'd suggest we replace the state with the prop to reduce the number of re-renders.

```suggestion
  if (gravatar && !deprecatedGravatarWarnTriggered) {
    deprecatedGravatarWarnTriggered = true;
    Logger.deprecate(
      "The `gravatar` prop has been deprecated and will soon be removed."
    );
  }

```

tomdavies73: The reason why I've introduced a new state variable is because we only want to fire the warning when the `gravatar` prop is used. 

For example, `Profile` renders `Portrait`, if you pass the `email` prop this is passed down to `Portrait` as the `gravatar` prop. Which means consumers may get warnings for the use of the `gravatar` prop even without using it intentionally. This way the warning will only fire if the `gravatar` prop is passed, and a gravatar image is actually loaded 👍

nuria1110 (commented):
**comment:** imo I would just completely remove Gravatar from `Profile` seeing as you have removed any mention of it in the docs and it doesn't (really) affect consumers using this component, the avatar just won't be loaded. I see you are trying to not be too disruptive but I don't think its worth having to add a new state as we're going to remove all of this soon anyways. (thats just my pov but happy to be overruled)

nuria1110 (approved):

Parsium (approved):

PR #6822, opened 2024-07-09
refactor(textarea): convert enzyme tests to rtl

robinzigmond:
### Proposed behaviour

Textarea component should have unit tests using react-testing-library

### Current behaviour

Unit tests are written using Enzyme

### Checklist

#### QA


src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+  expect(screen.getByRole("textbox")).toHaveAttribute("placeholder", "");
+});
+
+testStyledSystemMargin((props) => <Textarea {...props} />);
+
+// not sure what this is really testing - I assume it's the expandTextArea stuff, but can't get it to fail!
+// would rather remove this but leave it, commented out, to see if it causes coverage problems. Would like
+// to rewrite if it does, otherwise happy to remove it!
+test("the height of the textarea should remain unchanged when the window is resized", () => {
+  render(<Textarea />);
```
edleeks87, line 72: comment: yes this is to test that without the `expandable` prop set the textarea height doesn't change when the window resizes and the content wraps, it adds a scrollbar instead etc. When `expandable` is true the height adapts as the content wraps. It might be worth adding text to the input for this test 🤷

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+        expect.objectContaining({ placement: "bottom" })
+      );
+
+      useFloatingSpy.mockRestore();
+    });
+  }
+);
+
+test.each(validationTypes)(
+  "when %s validation prop is true boolean, does not render any validation iconn",
```
edleeks87, line 210:
```suggestion
  "when %s validation prop is true boolean, does not render any validation icon",
```

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+});
+
+test.each(["info", "warning", "error"])(
+  "with id prop provided, %s prop set as a string and the textarea element focused, the id of the validation tooltip should be added to 'aria-describedby' in the textarea element",
+  (validationType) => {
+    render(<Textarea id="foo" label="bar" {...{ [validationType]: "test" }} />);
+    const textarea = screen.getByRole("textbox");
+    textarea.focus();
+
+    expect(textarea).toHaveAttribute("aria-describedby", "foo-validation");
```
edleeks87, line 239: suggestion: you could use [toHaveAccessibleDescription](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveaccessibledescription) here

```
expect(textarea).toHaveAccessibleDescription("test")

File: src/components/textarea/textarea.test.tsx [hunk 4]
```diff
@@ -0,0 +1,719 @@
...
+  render(<Textarea label="foo" labelHelp="bar" helpAriaLabel="baz" />);
+
+  expect(screen.getByRole("button")).toHaveAttribute("aria-label", "baz");
+});
+
+test("should set the Help component's text content to the value of the labelHelp prop", () => {
+  render(<Textarea label="foo" labelHelp="bar" helpAriaLabel="baz" />);
+
+  screen.getByRole("button").focus();
+  expect(screen.getByTestId("tooltip")).toHaveTextContent("bar");
```
edleeks87, line 229:
```suggestion
  expect(screen.getByRole("tooltip", {name: "bar"})).toBeVisible();
```

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+);
+
+test.each(["info", "warning", "error"])(
+  "with id prop not provided, %s prop set as a string and the textarea element focused, the id of the validation tooltip should be added to 'aria-describedby' in the textarea element",
+  (validationType) => {
+    render(<Textarea label="bar" {...{ [validationType]: "test" }} />);
+    const textarea = screen.getByRole("textbox");
+    textarea.focus();
+
+    expect(textarea).toHaveAttribute(
```
edleeks87, line 250: suggestion: you could use [toHaveAccessibleDescription](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveaccessibledescription) here

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+
+test("renders the hint when inputHint prop is provided", () => {
+  render(<Textarea inputHint="foo" />);
+  expect(screen.getByText("foo")).toBeInTheDocument();
+});
+
+test("assigns the input hint a guid as id and references it in the aria-describedby of the input", () => {
+  render(<Textarea inputHint="bar" />);
+  expect(screen.getByText("bar")).toHaveAttribute("id", mockedGuid);
+  expect(screen.getByRole("textbox")).toHaveAttribute(
```
edleeks87, line 265: suggestion: you could use [toHaveAccessibleDescription](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveaccessibledescription) here

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+  );
+  expect(screen.getByText("inputHint")).toBeInTheDocument();
+  expect(screen.queryByText("labelHelp")).not.toBeInTheDocument();
+});
+
+test("when id and fieldHelp are both present, the id of the field help is added to 'aria-describedby' in the input", () => {
+  render(<Textarea id="foo" label="bar" fieldHelp="baz" />);
+
+  expect(screen.getByText("baz")).toHaveAttribute("id", "foo-field-help");
+  expect(screen.getByRole("textbox")).toHaveAttribute(
```
edleeks87, line 285: same suggestion/comment as above

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+});
+
+test("fieldHelp is present and id is not present, the id of the field help is added to 'aria-describedby' in the input", () => {
+  render(<Textarea label="bar" fieldHelp="baz" />);
+
+  expect(screen.getByText("baz")).toHaveAttribute(
+    "id",
+    `${mockedGuid}-field-help`
+  );
+  expect(screen.getByRole("textbox")).toHaveAttribute(
```
edleeks87, line 298: same suggestion/comment as above

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+
+  removeEventListenerSpy.mockRestore();
+});
+
+test("the required prop is passed to the input", () => {
+  render(<Textarea required label="required" />);
+  expect(screen.getByRole("textbox")).toBeRequired();
+});
+
+test("when the required prop is set, the label includes the 'required' asterisk", () => {
```
edleeks87, line 550: Comment: unless this and the `isOptional` one are needed for coverage I'd be happy to remove these as chromatic is capturing them etc

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+test("sets ref to empty after unmount", () => {
+  const ref = { current: null };
+  const { unmount } = render(<Textarea ref={ref} />);
+
+  unmount();
+
+  expect(ref.current).toBe(null);
+});
+
+test("renders with the expected default border radius styling", () => {
```
edleeks87, line 644: comment (non-blocking): I think we should remove any style tests that aren't needed for coverage that are already captured by chromatic/playwright. I'm happy for that to be done in a new ticket as these pass and aren't really hurting anything right now if we leave it etc

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+      ]}
+    />
+  );
+  expect(screen.getByRole("textbox")).toHaveStyleRule(
+    "border-radius",
+    "var(--borderRadius050) var(--borderRadius100) var(--borderRadius200) var(--borderRadius400)"
+  );
+});
+
+test("fires a console warning if more than four border-radius values are passed", () => {
```
edleeks87, line 677: comment: if you do end up removing style tests I think this one is worth keeping as we're obviously testing that the warning fires correctly etc

src/components/textarea/textarea.test.tsx
edleeks87: In fact lets leave this for now and do it in a second pass as we'll have other test files that are doing the same etc

src/components/textarea/textarea.test.tsx
robinzigmond: oops, that was a comment I put in to myself while converting the tests, and forgot to remove it 🤦‍♂️ 

I'll get the comment removed or fixed to reflect actual reality!

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+    );
+  }
+);
+
+test("renders the hint when inputHint prop is provided", () => {
+  render(<Textarea inputHint="foo" />);
+  expect(screen.getByText("foo")).toBeInTheDocument();
+});
+
+test("assigns the input hint a guid as id and references it in the aria-describedby of the input", () => {
```
Parsium, line 262: **suggestion(non-breaking)**: for the accessible name and description tests, what do you think about refining the test descriptions to focus on the intended outcomes rather than the implementation details? This can help make the purpose of each test clearer.

```tsx
// test("assigns the input hint a guid as id and references it in the aria-describedby of the input", () => {
test("textarea uses input hint text as its accessible description when inputHint prop is provided", () => {

// test("when id and fieldHelp are both present, the id of the field help is added to 'aria-describedby' in the input", () => {
test("textarea uses field help content as its accessible description when fieldHelp prop is provided", () => {

// test("with id prop provided, error prop set as a string and the textarea element focused, the id of the validation tooltip should be added to 'aria-describedby' in the textarea element", () => {
test("adds validation tooltip content to textarea's accessible description, when error prop is a string and textarea is focused", () => {
```

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+    expect(screen.getByRole("textbox")).toHaveAttribute(
+      "enterkeyhint",
+      keyHint
+    );
+  }
+);
+
+const validationTypes = ["error", "warning", "info"];
+
+describe.each(validationTypes)(
```
Parsium, line 141: **suggestion**: to minimise the amount of nested tests we have here, what do you think about amending the structure to the following?:

```tsx
it.each(["error", "warning", "info"] as const)("renders validation icon within textarea when %s prop is a string", () => {});
it.each(["error", "warning", "info"] as const)("renders validation icon on the label when %s prop is a string and validationOnLabel is true", () => {});
it("position of validation icon tooltip is overridden, when tooltipPosition prop is set", () => {});
```

This would approach would align with our other test files, ensuring consistency and improving readability.

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+          limitMinusValue
+            ? `${characterLimit - valueString.length} ${underCharacters} left`
+            : `${
+                valueString.length - characterLimit
+              } ${overCharacters} too many`
+        }`,
+        { selector: '[aria-hidden="true"]' }
+      )
+    ).toBeInTheDocument();
+  });
```
Parsium, line 368: **question**: is it necessary to test all three scenarios here? Splitting this test to handle the singular and plural use cases separately might simplify things.

Additionally, removing the conditional logic in this test could enhance readability. What do you think?

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+    ).toHaveAttribute("id", mockedGuid);
+  });
+
+  it("references the visually hidden hint id in the input's aria-described by", () => {
+    render(<Textarea value="foo" characterLimit={73} />);
+    expect(screen.getByRole("textbox")).toHaveAttribute(
+      "aria-describedby",
+      mockedGuid
+    );
+  });
```
Parsium, line 385: **suggestion**: what do you think about amending the test description to focus on the intended test outcome, rather than the implementation details. This might help the purpose of the tested code clearer to contributors.

```tsx
it("renders character limit information in a visually hidden container", () => {});
it("character limit information forms part of the textarea's accessible description", () => {});
```

src/components/textarea/textarea.test.tsx
```diff
@@ -0,0 +1,719 @@
...
+  "styles the textarea appropriately when an icon is present inside",
+  (props) => {
+    render(<Textarea value="test string" data-role="icon-test" {...props} />);
+    expect(screen.getByTestId("icon-test")).toHaveStyleRule(
+      "padding-right",
+      "var(--spacing500)",
+      { modifier: `& ${StyledInput}` }
+    );
+  }
+);
```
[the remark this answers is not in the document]
robinzigmond: Agreed. I've just been cautious throughougt and kept tests in where they were streaightforward to convert to RTL, rather than spend time on checking if style tests are already covered by Playwright/chromatic. I thought that we were going to come back and review these styling tests later - but if Chromatic definitely captures this and it's not needed for coverage (which seems unlikely in this case) then I'm happy to remove 👍

src/components/textarea/textarea.test.tsx
robinzigmond: I agree - if I change the test implementation in the way @edleeks87 has suggested, then this is a better way to word the test description.

src/components/textarea/textarea.test.tsx
robinzigmond: OK, it turns out this test is fine, and is providing value in that I can get it to fail if I comment out the appropriate `if (expandable)` condition (I think I was trying with the wrong one previously 🤦 ). So I'll leave this in, remove the confusing comment, and move the test so that it's next to the related ones where the `expandable` prop is true, rather than miles before either of them.

While I'm at it I'm also replacing `window.dispatchEvent("resize")` with `fireEvent.resize(window)` - `userEvent` doesn't seem to have anything for resize, and given that I think `fireEvent` is fine to use and is a bit more readable (I think it just does a `dispatchEvent` itself, with some sensible default properties).

src/components/textarea/textarea.test.tsx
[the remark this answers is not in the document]
robinzigmond: I just checked and there's certainly no story on Textarea for the `inputIcon` prop (this prop is not even mentioned in the docs except for being included in the prop table). So to get Chromatic to capture this I'd have to add such a story - which I'm not sure is wise as it probably doesn't add very much from a documentation point of view.

Based on this I've decided to leave this test in place for now, pending a later repo-wide discussion of what to do about styling tests.

src/components/textarea/textarea.test.tsx
```diff
@@ -17,28 +17,18 @@ jest.mock("../../__internal__/utils/helpers/guid");
...
+
   render(<Textarea name="my-textarea" defaultValue="test" />);
 
   expect(loggerSpy).toHaveBeenCalledWith(
     "Uncontrolled behaviour in `Textarea` is deprecated and support will soon be removed. Please make sure all your inputs are controlled."
   );
 
   expect(loggerSpy).toHaveBeenCalledTimes(1);
+
+  loggerSpy.mockClear();
```
Parsium, line 31: **suggestion**: what do you think about doing this, so the original implementation is restored?

```suggestion
  loggerSpy.mockRestore();
```

src/components/textarea/textarea.test.tsx
```diff
@@ -396,7 +339,8 @@ describe('when the "expandable" prop is set to "true"', () => {
       .spyOn(textarea, "scrollHeight", "get")
       .mockImplementation(() => expectedScrollHeight);
 
-    window.dispatchEvent(new Event("resize"));
+    fireEvent.resize(window);
```
Parsium, line 342: **praise**: nice spot that this can be done with `fireEvent`!

src/components/textarea/textarea.test.tsx
robinzigmond: oh yeah, not sure how/why I ended up with the wrong method here when I think it was correct before I cleaned it up!

Parsium (approved):

edleeks87 (approved):

PR #6824, opened 2024-07-10
refactor(select): convert tests for internal utils to RTL 

Parsium:
### Proposed behaviour

- Migrate internal utils for the `Select` components to RTL

### Current behaviour

- Internal util tests are written with Enzyme

### Checklist

#### QA


src/components/select/__internal__/utils/highlight-part-of-text.tsx
```diff
@@ -14,9 +14,12 @@ function highlightPartOfText(
     return text;
   }
 
-  const precedingText = text.substr(0, indexOfFirstMatch);
-  const matchingText = text.substr(indexOfFirstMatch, partToHighlight.length);
-  const followingText = text.substr(
```
Parsium, line 19: `substr` is a deprecated method for strings, so replaced with `slice`

src/components/select/__internal__/utils/highlight-part-of-text.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  expect(screen.getByText(/abc/i)).toBeVisible();
+  expect(screen.getByText(/def/i)).toBeVisible();
+
+  const matchedText = screen
+    .getAllByText(/foo/i)
+    .filter(
+      (element) =>
+        window.getComputedStyle(element).fontWeight === "700" &&
+        window.getComputedStyle(element).textDecoration === "underline"
+    );
```
Parsium, line 46: Open to any other alternative ways to test this

src/components/select/__internal__/utils/highlight-part-of-text.tsx
edleeks87: Comment (non-blocking): I wonder if using [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr) here would be a little more intuitive, not a huge deal either way though as long as we can guarantee the slight differences between [substring and slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) aren't an issue for us

edleeks87 (commented):
Really good stuff, I like how you've separated this work into smaller pars; the tests are much clearer now and you've removed some uses of deprecated functions 👏

src/components/select/__internal__/utils/highlight-part-of-text.tsx
```diff
@@ -29,7 +36,9 @@ function highlightPartOfText(
 
   const newValue = [
     <span key="preceding">{precedingText}</span>,
-    <MatchingText key="match">{matchingText}</MatchingText>,
+    <span key="match" data-role="matching-text" style={highlightStyling}>
```
robinzigmond, line 39: **comment** (non-blocking): I'm not sure why you've replaced the styled component here with inline styles. I'm confident it will have the same effect but I'd prefer to avoid inline styles unless we have a really good reason, and while I may be missing something, I don't see any reason here to make this change. (I know you've added the `data-role` as I requested, but as a `data-` attribute it ought to be fine for any HTML element, and even if TS does complain for some reason you could just have explicitly added `data-role` to the props interface for `MatchingText`.)

robinzigmond (approved):

edleeks87 (approved):

PR #6825, opened 2024-07-11
FE-6641 - Convert Hooks Tests to RTL

nineteen88:
### Proposed behaviour

All Hooks tests should use RTL and follow new best practices

### Current behaviour

Many Hooks tests are still using Enzyme and several of them need updating to follow new best practices

### Checklist

#### QA


src/hooks/__internal__/useFormSpacing/add-form-spacing/add-form-spacing.test.ts
edleeks87, line 1: Comment: the commit for this scopes the changes to `useFormSpacing` instead of `addFormSpacing` fyi

edleeks87 (commented):
comment (non-blocking): not a huge deal as they're not going to end up in the release notes etc but I wonder if it's a bit more legible to hyphen separate the scope name in the commit like so:

```
chore(use-media-query): convert tests to use RTL

File: src/hooks/__internal__/useCharacterCount/useCharacterCount.test.tsx [hunk 1]
```diff
@@ -0,0 +1,94 @@
+import * as React from "react";
+import { render, screen } from "@testing-library/react";
+import { act } from "react-dom/test-utils";
```
Parsium, line 3: **suggestion(non-breaking)**: what do you think of using RTL's wrapper for `act`? [The maintainers recommend using it mainly for consistency reasons](https://github.com/testing-library/testing-library-docs/issues/524#issuecomment-658681951)

src/hooks/__internal__/useCharacterCount/useCharacterCount.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const TestComponent = ({ value, characterLimit }: TestComponentProps) => {
+  const [characterCount] = useCharacterCount(value, characterLimit);
+
+  return <div>{characterCount}</div>;
+};
+
+const MOCK_VALUE = "test string";
+
+test.each([5, 10, 15])(
```
Parsium, line 19: **suggestion**: what do you think of splitting this test, so we assert being over and under the limit separately? That way we can avoid using conditionals in our tests.

src/hooks/__internal__/useFloating/useFloating.test.tsx
```diff
@@ -41,120 +41,100 @@ const MockComponent = ({
...
-          elements.floating.style.width = `${rects.reference.width}px`;
-        },
-      }),
-    ];
-
-    const { rerender } = render(
-      <MockComponent isOpen middleware={middleWare} />
-    );
+test("saves floating element original styles and restores them after closing", async () => {
+  const { rerender } = await render(<MockComponent isOpen />);
```
Parsium, line 82: **nitpick**: it looks like the await on `render` might be unnecessary. Removing it could simplify the code a bit :)

src/hooks/__internal__/useResizeObserver/useResizeObserver.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  it("should observe the element on component mount when enabled", () => {
+    render(<TestComponent callback={callbackProp} disabled={false} />);
+
+    expect(observeMock).toHaveBeenCalledWith(
+      screen.getByTestId("observed-node")
+    );
+  });
+
+  it("should unobserve the element and disconnect the observer on component unmount when enabled", () => {
```
Parsium, line 54: **question**: do we need to `unmount()` the component in this test?

src/hooks/__internal__/useScrollBlock/useScrollBlock.spec.tsx
```diff
@@ -1,124 +0,0 @@
...
-      .spyOn(document.documentElement, "clientWidth", "get")
-      .mockImplementation(() => 180);
-  });
-
-  afterEach(() => {
-    window.innerWidth = 1024;
-    clientWidthSpy.mockReset();
-  });
-
-  it("sets proper styles on the body element on mount and resets them on unmount", () => {
```
Parsium, line 35: **question**: would it be worth us keeping this test for the hook?

src/hooks/__internal__/useCharacterCount/useCharacterCount.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const MOCK_VALUE = "test string";
+
+test("shows characters left when input is shorter than the character limit", () => {
+  const characterLimit = 10;
+  const valueShorterThanLimit = "Test";
+  const expectedCharactersLeft = characterLimit - valueShorterThanLimit.length;
+  const expectedText = `${expectedCharactersLeft} ${
+    expectedCharactersLeft === 1 ? "character" : "characters"
+  } left`;
```
Parsium, line 24: **suggestion**: What do you think about making the expected text static? This would have the benefit of avoiding conditional logic within the test:

```suggestion
  const characterLimit = 10;
  const valueShorterThanLimit = "Test";
  const expectedText = "6 characters left";
```

src/hooks/__internal__/useFloating/useFloating.test.tsx
```diff
@@ -41,120 +41,100 @@ const MockComponent = ({
...
-      strategy: "fixed",
-      placement: "top",
-    });
+  render(<MockComponent isOpen strategy="fixed" placement="top" />);
+  expect(computePositionSpy.mock.calls[0][0]).toBe(
+    screen.getByTestId("reference-element")
+  );
+  expect(computePositionSpy.mock.calls[0][1]).toBe(
+    screen.getByTestId("floating-element")
+  );
```
Parsium, line 53: **suggestion(non-blocking)**: what do you think of this change to make the assertions more readable? Not essential though.

```suggestion
  expect(computePositionSpy).toHaveBeenNthCalledWith(1,
    screen.getByTestId("reference-element")
  );
  expect(computePositionSpy).toHaveBeenNthCalledWith(2,
    screen.getByTestId("floating-element")
  );
```

nineteen88: For the record I'd tried to do it like this originally but it was erroring so I had to change it to the current version. Unless I'm missing something?

Parsium (approved):

edleeks87 (approved):

PR #6831, opened 2024-07-12
fix(link): ensure focus styles are removed when underlying element loses focus

### LINKED ISSUE #6688: Pager component retains focus styling highlight on pagination buttons
### Description

Screenshot 2024-04-18 at 16 25 23

going from "next" to "previous" then back to "next" will replicate.
the focus from the previous render of the button will retain the black focus bar under the text

### Reproduction

https://carbon.sage.com/?path=/docs/pager--docs

### Steps to reproduce

going from "next" to "previous" then back to "next" will replicate.
the focus from the previous render of the button will retain the black focus bar under the text

There is a recording attached to the JIRA ticket for further detail calrification

SBS-93920

### Suggested solution

_No response_

### Carbon version

130.0.0

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

### LINKED ISSUE #6808: Bad style on current crumb when navigating insde
### Description

I am navigating in the breadcrumb, i expect to not have the big black bold underligne below my current page.
If i navigate a lot, i can have all my crumbs underligned with bold black underlignes
image

For me case, i can do a workaround to refresh my whole component to clean the style (using a key on my component)

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-angxru?file=src%2FApp.tsx

### Steps to reproduce

Click on Path 2 for exemple
Click on Add crumb 

_No response_

### Suggested solution

_No response_

### Carbon version

140.1.0

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

edleeks87:
fix #6808
fix #6688

### Proposed behaviour

Ensures that styling updates when focus is lost but no blur is not triggered. This means that when a `Crumb` is clicked and becomes current the focus styling is not persisted.

### Current behaviour

Part of the Link focus styling is persisted when the link loses focus but no blur event isn't triggered. See linked issue for current behaviour

### Checklist

#### QA

### Testing instructions

Test story added for `Breadcrumb` component

### DIFF
--- a/src/components/link/link.component.tsx
+++ b/src/components/link/link.component.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo, useState } from "react";
+import React, { useContext, useEffect, useMemo, useState } from "react";
 
 import Icon, { IconType } from "../icon";
 import MenuContext from "../menu/__internal__/menu.context";
@@ -161,6 +161,12 @@ export const Link = React.forwardRef<
         </>
       );
     };
+
+    useEffect(() => {
+      if (disabled || !(href || onClick)) {
+        setHasFocus(false);
+      }
+    }, [disabled, href, onClick]);
 
     return (
       <StyledLink
--- a/src/components/breadcrumbs/breadcrumbs-test.stories.tsx
+++ b/src/components/breadcrumbs/breadcrumbs-test.stories.tsx
@@ -4,7 +4,7 @@ import { Crumb, CrumbProps } from "./crumb";
 
 export default {
   title: "Breadcrumbs/Test",
-  includeStories: ["DefaultCrumb"],
+  includeStories: ["DefaultCrumb", "WhenFocusedCrumbBecomesCurrent"],
   parameters: {
     info: { disable: true },
     chromatic: {
@@ -48,5 +48,22 @@ export const DefaultCrumb = (props: Partial<CrumbProps>) => {
   );
 };
 
+export const WhenFocusedCrumbBecomesCurrent = () => {
+  const [current, setCurrent] = React.useState(false);
+
+  return (
+    <>
+      <Breadcrumbs>
+        <Crumb href="#bar" onClick={() => setCurrent(true)} isCurrent={current}>
+          Crumb{current ? "" : " not"} current
+        </Crumb>
+      </Breadcrumbs>
+
+      <div id="bar">Container</div>
+    </>
+  );
+};
+
 Default.storyName = "default";
 DefaultCrumb.storyName = "single crumb";
+WhenFocusedCrumbBecomesCurrent.storyName = "when focused crumb becomes current";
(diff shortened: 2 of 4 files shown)


DipperTheDan (approved):

src/components/link/link.component.tsx
```diff
@@ -84,6 +84,12 @@ export const Link = React.forwardRef<
     const [hasFocus, setHasFocus] = useState(false);
     const l = useLocale();
     const { inMenu } = useContext(MenuContext);
+    const internalRef = useRef<HTMLLinkElement | HTMLButtonElement>(null);
```
Parsium, line 87: **suggestion**: the bug seems to surface when the focused `Link` is either disabled or has its hyperlink removed immediately after and the state is not updated.

What do you think about using an effect to handle this? This way we wouldn't need to use refs:

```tsx
useEffect(() => {
      if ((disabled || !href) && hasFocus) setHasFocus(false);
}, [disabled, hasFocus, href]);
```

We could avoid an extra re-render by adding logic to `Pager`, `Breadcrumbs`, etc. But I expect it could become unnecessarily complex to handle all possible scenarios where focused `Link` could be disabled immediately after.

edleeks87: yep good point, I'd missed that `isCurrent` negated the href and onClick. I'm not keen on adding them in the parent components as we'd need to handle it every time we add a new one that composes Link etc so fixing it as low down as possible is the better approach.

I think the following will suffice as we don't really care about the current state and if we do we can handle it in a callback within the state setter

```
useEffect(() => {
      if (disabled || !(href && onClick)) {
        setHasFocus(false);
      }
    }, [disabled, href, onClick]);

File: src/components/link/link.component.tsx [hunk 2]
```diff
@@ -162,6 +162,12 @@ export const Link = React.forwardRef<
       );
     };
 
+    useEffect(() => {
+      if (disabled || !(href && onClick)) {
```
Parsium, line 166: **issue**: we have a case with `Pager` where `Link` rendered without `onClick` or `href`, so the state will be set to false on every render:

https://github.com/user-attachments/assets/d845e634-6f4b-4be1-be13-8ab21a641cd3

```suggestion
      if (disabled || !(href || onClick)) {
```

Parsium (approved):

Parsium (approved):

src/components/link/link.component.tsx
edleeks87: Yep you're right I rushed it and forgot to change the operator to OR as originally had it as `!href && !onClick`. The fact there's an `onClick` and no `href `is being negated from false to true, I don't think there's a point where Pager links don't have an onClick, the types say they're optional but they're always passed as we wrap the prop callbacks in useCallback then pass it etc but it's irrelevant here

DipperTheDan (approved):

stephenogorman (approved):

PR #6834, opened 2024-07-15
chore(modal): convert tests to RTL - FE-6610

DipperTheDan:
### Proposed behaviour

Convert tests to RTL

### Current behaviour

Tests are currently using Enzyme.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/modal/modal.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+  onCancelFnTwo = jest.fn();
+
+  render(
+    <>
+      <Modal data-role="first-modal" open onCancel={onCancelFn} />
+      <Modal data-role="second-modal" open onCancel={onCancelFnTwo} />
+    </>
+  );
+
+  fireEvent.keyUp(document, { key: "Escape", code: "Escape" });
```
Parsium, line 97: **suggestion**: what do you think of using the user-event library here instead? The library would handle events in a manner that more closely resembles how events are handled in the browser. 

```suggestion
  await user.keyboard("{Escape}");
```

You may need to use Jest's fake timers if you agree, since `Modal` uses a third-party code (React Transition Group) which could schedule asynchronous tasks that are outside our control:

```ts
test("closes top modal when Escape key is pressed", async () => {
   jest.useFakeTimers();
   
   const firstOnCancel = jest.fn();
   const secondOnCancel = jest.fn();
   const user = userEvent.setup({ advanceTimers: jest.advanceByTime });

   // test code...

   jest.runOnlyPendingTimers(); // flush any potential scheduled tasks
   jest.useRealTimers();
});
```
For more detail on this, you can refer to[ RTL's advice on Fake Timers](https://testing-library.com/docs/using-fake-timers/).

src/components/modal/__internal__/modal-manager.test.tsx
```diff
@@ -0,0 +1,325 @@
...
+  otherModal: HTMLElement | null;
+
+test("when the addModal method has been called, then the element passed in an attribute should be the topmost element", () => {
+  cb1 = jest.fn();
+  cb2 = jest.fn();
+
+  mockModal1 = document.createElement("div");
+  mockModal2 = document.createElement("div");
+
+  ModalManager.clearList();
```
robinzigmond, line 19: **comment**: I see that we're calling this `clearList` method in the setup for many of these tests, although not all of them. I assume this is done to clean up the `ModalManager` between tests to stop them interfering with each other - but in this case I would prefer that this is called in an `afterEach` so that it is clear that this is cleanup that happens after each test.

src/components/modal/modal.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+    addEventListenerSpy = jest.spyOn(document, "addEventListener");
+    removeEventListenerSpy = jest.spyOn(document, "removeEventListener");
+  });
+
+  afterEach(() => {
+    addEventListenerSpy.mockRestore();
+    removeEventListenerSpy.mockRestore();
+  });
+
+  it("binds the key event listener to the document on mount", () => {
```
robinzigmond, line 36: **suggestion** (non-blocking): making this comment non-blocking as I'm not sure how easy this will be, but I think it should be possible - all these tests for adding/removing event listeners seem too focused on implementation. The important thing is that, once the modal is mounted, it responds to key presses, and then doesn't once unmounted. I would have thought that we can test for that by simulating actual user interactions, rather than the implementation details of when we call `addEventListener` and `removeEventListener`.

src/components/modal/__internal__/modal-manager.test.tsx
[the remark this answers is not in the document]
DipperTheDan: For lack of a better solution I've gone with your `splice` way. We can always refactor it in future if we can think if a better/cleaner way to do it.

src/components/modal/__internal__/modal-manager.test.tsx
```diff
@@ -0,0 +1,325 @@
...
+  ModalManager.removeModal(mockModal1);
+
+  expect(window.__CARBON_INTERNALS_MODAL_LIST?.length).toEqual(0);
+
+  cb1?.mockClear();
+  cb2?.mockClear();
+});
+
+// This test case is required to hit coverage. It covers the case where the modal to be removed is not in the `modalList` array.
+test("when the removeModal method has been called and the param does not match a Modal in the list, then nothing happens", () => {
```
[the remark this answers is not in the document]
DipperTheDan: I've removed the comment. I had this here because I struggled for a few hours to get the code coverage for the particular lines of code this covered. I put this comment here so someone in the future wouldn't have the same issue I did 🤣

src/components/modal/__internal__/modal-manager.test.tsx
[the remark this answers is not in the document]
DipperTheDan: I kind of went with what we had before. These tests are quite implementation-y so I tried to go with descriptions that say exactly what the implementation is doing. I can try to tweak them a bit though.

src/components/modal/modal.test.tsx
DipperTheDan: I did try to do it this way originally but I didn't do the `advanceTimers` so the tests either didn't pass or weren't reliable, hence the use of `fireEvent`. I've tried it your way and this works so I'll go with this.

src/components/modal/__internal__/modal-manager.test.tsx
DipperTheDan: I'll get this `afterEach` suggestion implemented 👍🏻

Parsium (approved):

robinzigmond (approved):

PR #6838, opened 2024-07-17
refactor(heading): refactor enzyme tests to RTL

tomdavies73:
### Proposed behaviour

Converts all enzyme tests to RTL.

### Current behaviour

Tests are currently written in enzyme.

### Checklist

#### QA


src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  render(
+    <Heading
+      title="foo"
+      data-role="heading"
+      data-component="heading"
+      data-element="heading"
+    />
+  );
+
+  const headingwWrapper = screen.getByTestId("heading");
```
nuria1110, line 18: **nit-pick:** typo
```suggestion
  const headingWrapper = screen.getByTestId("heading");
```

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  "renders the correct heading element when the `headingType` prop is passed as %s",
+  (headingType, level) => {
+    render(<Heading title="foo" headingType={headingType} />);
+
+    const heading = screen.getByRole("heading", { level });
+    expect(heading).toBeVisible();
+  }
+);
+
+test("renders wih a custom help text, via the `help` prop", async () => {
```
nuria1110, line 84:
```suggestion
test("renders with a custom help text, via the `help` prop", async () => {
```

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+    name: "help-aria-label",
+  });
+  await user.hover(helpIconButton);
+  const helpText = await screen.findByText("help text");
+
+  expect(helpIconButton).toHaveAttribute("data-element", "help");
+  expect(helpText).toBeVisible();
+});
+
+test("renders wih a custom help text link, via the `helpLink` prop", () => {
```
nuria1110, line 100:
```suggestion
test("renders with a custom help text link, via the `helpLink` prop", () => {
```

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const backLink = screen.getByRole("button", { name: "Back" });
+  expect(backLink).toBeVisible();
+});
+
+test("focuses the back link, when `backLink` prop is a function", () => {
+  const backLinkFunction = jest.fn();
+  render(<Heading title="foo" backLink={backLinkFunction} />);
+
+  const backLink = screen.getByRole("button", { name: "Back" });
+  fireEvent.mouseDown(backLink);
```
nuria1110, line 167: **question:** is there a reason why we aren't using userEvent here? Unless we are specifically testing on mouse down, I think something like `user.tab()` would be closer to a real user's actions.

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  render(<Heading title="foo" pills={pillsNode} />);
+
+  const pillsContainer = screen.getByTestId("heading-pills-container");
+  const pills = screen.getAllByTestId("pill");
+
+  expect(pillsContainer).toHaveAttribute("data-element", "pills");
+  pills.forEach((pill) => {
+    expect(pill).toBeVisible();
+  });
+});
```
nuria1110, line 239: **suggestion:** By doing this we have no way of knowing if they've both been rendered so instead of iterating through the pills, to make sure both are visible I would give them a different data-roles and assert they are both visible individually. Or alternatively, making sure `pills` has a length of two for e.g. 

The same thing could be said about the previous test for the divider/separator.

nuria1110 (commented):
Looking good 👍, just a few minor comments and suggestions

src/components/heading/heading.test.tsx
tomdavies73: I could add another check to see if 2 children are rendered etc. but not too sure using two data-roles is the way to go on this one. I'll see what the other reviewer says

src/components/heading/heading.test.tsx
tomdavies73: We'd have to be specifically testing mouse down behaviour here, maybe user.Click?

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  const backLink = screen.getByRole("link", { name: "Back" });
+  fireEvent.mouseDown(backLink);
+
+  expect(backLink).toBeFocused();
+});
+
+test("renders a back button, when `backLink` prop is a function", () => {
+  const backLinkFunction = jest.fn();
+  render(<Heading title="foo" backLink={backLinkFunction} />);
```
robinzigmond, line 156:
```suggestion
  render(<Heading title="foo" backLink={() => {}} />);
```

**suggestion** (non-blocking): since we're not asserting that the function has been called or anything, I don't see any need for a mock, just provide a "dummy function" like this (which is essentially what `jest.fn()` gives you apart from being able to track calls to it, which as I said we're not doing).

Same in the next test too.

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  const backLink = screen.getByRole("button", { name: "Back" });
+  fireEvent.mouseDown(backLink);
+
+  expect(backLink).toBeFocused();
+});
+
+/* styling test for coverage */
```
robinzigmond, line 172: **suggestion** (non-blocking): maybe move the "just for coverage" tests to the very end, so it's clear when you get to them that there's nothing else important, and it's easier to remove them later if we can figure out how to get combined coverage.

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  render(<Heading title="foo" divider />);
+
+  const separator = screen.getByRole("separator");
+  expect(separator).toBeVisible();
+});
+
+test("renders a separator when the `separator` prop is true", () => {
+  render(<Heading title="foo" divider separator />);
+
+  const separators = screen.getAllByRole("separator");
```
robinzigmond, line 215: **comment/suggestion**: I have to admit, I'm not really a fan of this test.

For one thing, all we're saying is that all the `hr` elements (they seem to be the only things to do with the Heading component that will have a role of "separator") that are in the DOM are visible. Literally the only way that could fail is if we had one or more `<hr>` elements in the DOM but hid them with CSS. We could not render any of them and this test will still pass (without actually making a single assertion). [EDIT: actually that's not true, I just looked up the RTL docs and apparently `getAllBy` throws an error if it does't find any, which is what we'd want here. However the other problems I'm going on to describe would remain.] More likely, and just as serious, we could fail to render the separator `<hr>` but render the divider one, and this would still pass (I've just tested this btw - delete the line of the component file that renders the <StyledSeparator>, and all tests still pass).

If we really have to go down this route, at the very least we should assert that the length of `separators` is 2. But that feels too implementation-dependent to me (if we ever get rid of the divider - yes I know that's not likely - we'd have to change this without having actually affected how `separator` behaved, and without any comments it's really not clear what the "magic number" 2 would be doing there).

Since the `StyledSeparator` has a `data-role` I actually think the best thing to do here is just `getByTestId("heading-separator")` and assert that that's visible.

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+    "3px solid var(--colorsSemanticFocus500)",
+    { modifier: "button:focus" }
+  );
+});
+
+test("renders a divider when the `divider` prop is true", () => {
+  render(<Heading title="foo" divider />);
+
+  const separator = screen.getByRole("separator");
+  expect(separator).toBeVisible();
```
robinzigmond, line 209: **comment**: while this test is better than the following one for the separator (I'd recommend reading my long comment on that first for context), I'm still not really a fan, since this test will pass even if we accidentally stop the divider being rendered, if there's any `<hr>` being rendered elsewhere (including if we accidentally always render the separator).

This isn't as easy to fix because the unlike the `StyledDivider` doesn't have a `data-role`. I don't see why we couldn't add one though, to make this test more robust.

src/components/heading/heading.test.tsx
robinzigmond: I agree with @nuria1110 here, and see that she pointed out the same problem that I've just pointed out with the divider/separator tests (but I've just typed those out so I'll leave them in now! 😅 ).

I think just asserting that the length is 2 is the right way here, rather than adding separate data-roles. But since these data-roles would be added just in the test rather than the component itself, there's no particular harm - and you're already adding the same data-role to each so it would just mean changing them to 2 different ones. (But overall a simple check on length is perfectly fine by me, as well as the visibility checks.)

nuria1110: Yeah sorry I probably wasn't clear enough in my initial comment about the data-roles, basically what Robin said just doing this is what I meant, but yeah either or is fine my me
```
      <Pill data-role="pill1">Pill one</Pill>
      <Pill data-role="pill2">Pill two</Pill>
```

src/components/heading/heading.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+  );
+
+  const backLink = screen.getByRole("link", { name: "Back" });
+  expect(backLink).toHaveAttribute(
+    "href",
+    "https://www.warnerbros.com/movies/heat"
+  );
+});
+
+test("focuses the back link, when `backLink` is a href on mouse down", () => {
```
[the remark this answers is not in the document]
tomdavies73: If a string is passed, the back link renders as a link. If a function is passed it renders as a button, maybe just specifying string instead would be better

robinzigmond (approved):

nuria1110 (approved):

PR #6839, opened 2024-07-17
refactor(number): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert tests to RTL.

### Current behaviour

Tests are in enzyme.

### Checklist

#### QA


src/components/number/number.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  expect(screen.getByRole("textbox")).toHaveValue("123");
+});
+
+test("calls onChange when valid characters are provided", async () => {
+  const user = userEvent.setup();
+  const onChange = jest.fn();
+  render(<Number onChange={onChange} />);
+
+  await user.tab();
+  await user.keyboard("123");
```
robinzigmond, line 35: **suggestion** (non-blocking): user-event provides the `type` utility specifically for typing into input elements, which I feel is more appropriate to use for this component that `user.keyboard` (and also removes the need to manually tab into or otherwise focus the input) https://testing-library.com/docs/user-event/utility#type

Same applies to all uses of `user.keyboard` in here where we're just entering text into the input

src/components/number/number.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  render(<Number disabled />);
+
+  expect(screen.getByRole("textbox")).toBeDisabled();
+});
+
+test("renders with `ref` when passed as an object", () => {
+  const ref = { current: null };
+  render(<Number ref={ref}>foo</Number>);
+
+  expect(ref.current).not.toBeNull();
```
robinzigmond, line 133: **suggestion** (non-blocking): it might be worth checking it's equal (strict reference equality with `toBe` should work here as it's the same DOM element) to the input itself, rather than any random non-null value: ie `expect(ref.current).toBe(screen.getByRole("textbox"))`

src/components/number/number.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+});
+
+test("renders with `ref` when passed as a callback", () => {
+  const ref = jest.fn();
+  render(<Number ref={ref}>foo</Number>);
+
+  expect(ref).toHaveBeenCalled();
```
robinzigmond, line 140: **suggestion** (non-blocking): same as above, it's probably worth checking what it's been called with (it should be the DOM element corresponding to the input)

src/components/number/number.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+});
+
+test("sets `ref` to empty after unmount", () => {
+  const ref = { current: null };
+  const { unmount } = render(<Number ref={ref}>foo</Number>);
+
```
DipperTheDan, line 146: suggestion(non-blocking): just a suggestion so that we assert what the value should be before it's unmounted/ returned back to null. I'm happy either way though if you think it's overkill. 

```suggestion
expect(ref.current).toBe(screen.getByRole("textbox"))
```

DipperTheDan (approved):

PR #6842, opened 2024-07-18
fix(batch-selection): ensure all button children are disabled when inside a disabled BatchSelection

tomdavies73:
### Proposed behaviour

Uses context to ensure all supported children are disabled, when the `disabled` prop is passed to `BatchSelection`. This ensures every button cannot be interacted with.

Screenshot 2024-07-18 at 13 51 42

### Current behaviour

Currently, when the `disabled` prop is passed to `BatchSelection` only some disabled styling is passed to button children but they are not actually disabled. Which results in some buttons appearing disabled and not possible to be interacted with via mouse, but accessible via keyboard.

Screenshot 2024-07-18 at 13 54 01

### Checklist

#### QA


src/components/batch-selection/__internal__/batch-selection.context.ts
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export type BatchSelectionContextType = {
+  disabled?: boolean;
```
edleeks87, line 4: comment: we don't need to make this prop optional as it has a default value below

edleeks87: suggestion (non-blocking): We could also rename the prop so that we don't have to rename it when we destructure.

```
const { disabled: batchSelectionDisabled } = useContext(
      BatchSelectionContext
    );
```

could be

```
const { batchSelectionDisabled } = useContext(
      BatchSelectionContext
    );
```

src/components/batch-selection/batch-selection.mdx
```diff
@@ -47,6 +50,8 @@ import IconButton from "carbon-react/lib/components/icon-button";
 
 ### Disabled
 
+All button children will be automatically disabled when the `disabled` prop is true.
```
edleeks87, line 53:
```suggestion
All Carbon components passed as children that can render as a button will be automatically disabled when the `disabled` prop is true.
```

src/components/button-minor/button-minor.component.tsx
```diff
@@ -15,7 +16,7 @@ export const ButtonMinor = React.forwardRef<
   (
     {
       buttonType: buttonTypeProp = "secondary",
-      disabled = false,
+      disabled,
```
edleeks87, line 19:
```suggestion
      disabled = false,
```

src/components/button-minor/button-minor.component.tsx
```diff
@@ -30,11 +31,15 @@ export const ButtonMinor = React.forwardRef<
...
     } = useContext(ButtonBarContext);
+    const { disabled: batchSelectionDisabled } = useContext(
+      BatchSelectionContext
+    );
 
     const buttonType = buttonTypeContext || buttonTypeProp;
     const size = sizeContext || sizeProp;
     const iconPosition = iconPositionContext || iconPositionProp;
     const fullWidth = fullWidthContext || fullWidthProp;
+    const isDisabled = disabled || batchSelectionDisabled || false;
```
edleeks87, line 42:
```suggestion
    const isDisabled = disabled || batchSelectionDisabled;
```

If we keep both the prop and context default values we don't need the last part here

src/components/button/button.component.tsx
```diff
@@ -194,7 +195,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       buttonType: buttonTypeProp = "secondary",
       children,
       destructive = false,
-      disabled = false,
+      disabled,
```
edleeks87, line 198: same as above wrt default values

src/components/icon-button/icon-button.component.tsx
```diff
@@ -39,6 +40,10 @@ const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
     }: IconButtonProps,
     ref
   ) => {
+    const { disabled: batchSelectionDisabled } = useContext(
+      BatchSelectionContext
+    );
+    const isDisabled = disabled || batchSelectionDisabled || false;
```
edleeks87, line 46: comment (non-blocking): It might be worth adopting the same approach as the other components that consume this context wrt to using default values so it's consistent etc

src/components/link/link.component.tsx
```diff
@@ -84,14 +85,19 @@ export const Link = React.forwardRef<
     const [hasFocus, setHasFocus] = useState(false);
     const l = useLocale();
     const { inMenu } = useContext(MenuContext);
+    const isAButton = onClick && !href;
```
edleeks87, line 88: comment: I think it might be overkill to add these additional checks here. We could probably just add the same as in the other components and not really care if it's a button etc

edleeks87 (commented):
Good stuff @tomdavies73 overall approach LGTM just some minor comments 👍

DipperTheDan (commented):
Nothing more to add from me 😄

src/components/batch-selection/batch-selection.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const minorButton = screen.getByRole("button", { name: "edit" });
+
+  expect(minorButton).toBeDisabled();
+});
+
+test("`Link` children should be automatically disabled via context", () => {
+  render(
+    <BatchSelection selectedCount={0} disabled>
+      <Link>Link as a button</Link>
```
edleeks87, line 79:
```suggestion
      <Link>Link as an anchor</Link>
```

src/components/batch-selection/batch-selection.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+    </BatchSelection>
+  );
+
+  const link = screen.getByTestId("link-anchor");
+  expect(link).toHaveStyle("cursor: not-allowed");
```
edleeks87, line 84: question: can we not just assert `toBeDisabled` here?

tomdavies73: Yeah, I tried this originally, but playwright doesn't support the `.toBeBisabled` assertion on anchor elements, due to the attribute not being natively supported. So instead I just asserted against the behaviour we apply to link to make it "disabled"

https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-disabled

DipperTheDan, line 84: nitpick of formatting:

```suggestion
  const link = screen.getByTestId("link-anchor");
  
  expect(link).toHaveStyle("cursor: not-allowed");
```

src/components/batch-selection/batch-selection.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  link.focus();
+  expect(link).not.toHaveFocus();
```
DipperTheDan, line 86: nitpick of formatting:

```suggestion
  link.focus();
  
  expect(link).not.toHaveFocus();
```

src/components/batch-selection/batch-selection.test.tsx
edleeks87: Yep good point, every few months I need reminding that a disable link isn't a thing and passing the href or not is how you control that

DipperTheDan (approved):

edleeks87 (approved):

ZhuoyuJin (approved):

PR #6844, opened 2024-07-18
refactor(action-popover): convert enzyme tests to rtl

robinzigmond:
### Proposed behaviour

ActionPopover should have unit tests written using react-testing-library

### Current behaviour

unit tests use Enzyme

### Checklist

#### QA


src/components/action-popover/action-popover.style.ts
robinzigmond, line 1: Only just realised I'd left the comments on this file. I purely added them for my own benefit when trying to understand what the tests for this function were really doing, and meant to remove them before pushing and raising the PR 🤦 

Still, I'll leave them for now and only remove when addressing comments - especially as there's an argument that these might be useful for others reading the code and therefore worth keeping?

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+
+  render(
+    <ActionPopover>
+      <ActionPopoverItem icon="email">Email Invoice</ActionPopoverItem>
+      <ActionPopoverItem icon="print">Print Invoice</ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
+  await screen.getByRole("button", { name: "Print Invoice" }).focus();
```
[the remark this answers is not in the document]
edleeks87: There's await -> focus calls on line 295, 338, 375 and 412 as well if you do decide to remove etc

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+      <ActionPopoverItem icon="print" disabled>
+        Print Invoice
+      </ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
+  await user.click(screen.getByText("Print Invoice"));
+
+  expect(screen.getByRole("list")).toBeInTheDocument();
```
edleeks87, line 359: suggestion: it might be better to assert that it's also visible here if possible rather than it's just in the DOM etc

```suggestion
  expect(screen.getByRole("list")).toBeVisible();
```

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+        Print Invoice
+      </ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
+  await screen.getByRole("button", { name: "Print Invoice" }).focus();
+  await user.keyboard("{Enter}");
+
+  expect(screen.getByRole("list")).toBeInTheDocument();
```
edleeks87, line 378:
```suggestion
  expect(screen.getByRole("list")).toBeVisible();
```

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+  const onClose = jest.fn();
+
+  render(
+    <ActionPopover onClose={onClose}>
+      <ActionPopoverItem>example item</ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
+  expect(screen.getByRole("list")).toBeInTheDocument();
```
edleeks87, line 480:
```suggestion
  expect(screen.getByRole("list")).toBeVisible();
```

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+
+  const onClose = jest.fn();
+
+  render(
+    <ActionPopover onClose={onClose}>
+      <ActionPopoverItem>example item</ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
```
edleeks87, line 479: comment (non-blocking): I assume this and the button located below are the same element so we could store this in a variable etc

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+      <ActionPopoverItem disabled>disabled item</ActionPopoverItem>
+      <ActionPopoverItem>example item</ActionPopoverItem>
+    </ActionPopover>
+  );
+
+  await user.click(screen.getByRole("button"));
+
+  await user.click(screen.getByRole("button", { name: "disabled item" }));
+
+  expect(screen.getByRole("list")).toBeInTheDocument();
```
edleeks87, line 501:
```suggestion
  expect(screen.getByRole("list")).toBeVisible();
```

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+        <ActionPopoverItem>example item 1</ActionPopoverItem>
+        <ActionPopoverItem>example item 2</ActionPopoverItem>
+      </ActionPopover>
+    );
+
+    screen.getByRole("button").focus();
+    const userEventKeyCode = key === "Space" ? " " : `{${key}}`;
+    await user.keyboard(userEventKeyCode);
+
+    expect(screen.getByRole("list")).toBeInTheDocument();
```
edleeks87, line 538:
```suggestion
    expect(screen.getByRole("list")).toBeVisible();
```

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+    );
+
+    await user.click(screen.getByRole("button"));
+
+    expect(
+      screen.queryByRole("button", { name: "submenu item 1" })
+    ).not.toBeInTheDocument();
+
+    // move mouse over the item with the submenu
+    await user.pointer({
```
edleeks87, line 900: question: out of interest is there a reason why we're using `pointer` instead of `hover`?

edleeks87 (commented):
Really good work @robinzigmond, I've left a few comments but they're all very trivial and in many cases the same comment repeated etc/ I do notice when I run these locally that there's quite a few act warnings relating to this [function](https://github.com/Sage/carbon/blob/a0055e9b3d7cd49d764946ce9f8c6daeff8d43d4/src/components/action-popover/action-popover.component.tsx#L130), might be worth investigating if we can easily remove them.

robinzigmond:
> Really good work @robinzigmond, I've left a few comments but they're all very trivial and in many cases the same comment repeated etc/ I do notice when I run these locally that there's quite a few act warnings relating to this [function](https://github.com/Sage/carbon/blob/a0055e9b3d7cd49d764946ce9f8c6daeff8d43d4/src/components/action-popover/action-popover.component.tsx#L130), might be worth investigating if we can easily remove them.

no problem, thanks for the review @edleeks87 👍 I'll take a look at these tomorrow.

I hadn't noticed any `act` warnings when running locally, which is a surprise as I'm usually aware of them and try to make sure I get rid of them. I can only guess some crept in at a late stage and I never noticed, I'll definitely take a look at this as well as addressing all your other comments.

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+      <ActionPopover>
+        <ActionPopoverItem href="#" download>
+          test download
+        </ActionPopoverItem>
+      </ActionPopover>
+    );
+
+    await user.click(screen.getByRole("button"));
+
+    await screen.getByRole("link").focus();
```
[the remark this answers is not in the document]
robinzigmond: yeah you're right, thanks - I think this was a copy-paste error where I changed something and forgot that it made the `await` unnecessary

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+test("renders with the menu closed by default", () => {
+  render(
+    <ActionPopover>
+      <ActionPopoverItem>example item</ActionPopoverItem>
+    </ActionPopover>
+  );
+  expect(screen.queryByRole("list")).not.toBeInTheDocument();
+});
+
+test.each<["top" | "bottom", boolean, string]>([
```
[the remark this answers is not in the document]
robinzigmond: It wasn't clear to me that this is tested in Playwright. I've just looked again and there are some similar tests, but I'm not sure they're doing exactly the same thing. In particular, I can't see any tests of the `top-start` case.

Obviously though this is a pretty janky test that is hard to read and only tests implementation details, so I'd be more than happy to get rid of it - which on balance I think you're right that I can (unless needed for coverage - I doubt it is but will check).

src/components/action-popover/action-popover.test.tsx
robinzigmond: yeah probably. There were a few tests where I can't do that because the `getByRole` call does a new query each time, so if the DOM has changed then it's needed (eg to assert the element is there to start with and then not there after an action), but looking at this one I think you're right that it should be OK to store the resulting element in a variable and reuse it.

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+  expect(divider).toHaveAttribute("data-element", "action-popover-divider");
+});
+
+test("has a default aria-label", () => {
+  render(
+    <ActionPopover>
+      <ActionPopoverItem>example item</ActionPopoverItem>
+    </ActionPopover>
+  );
+  expect(screen.getByRole("button")).toHaveAttribute("aria-label", "actions");
```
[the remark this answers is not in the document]
robinzigmond: I take the point, and am happy to change this - just thought I'd make the point that since we're explicitly testing a prop called `aria-label`, there's an argument for making sure it actually becomes the `aria-label` of the input, rather than something else going on (eg with `aria-labeledby`) to ensure the input has the correct accessible name. In one sense that's just an implementation detail, but as I said if we're providing a prop called `aria-label` I think it would be confusing if it didn't end up as the actual HTML `aria-label` attribute.

That's mostly be just playing devil's advocate though, so having made the point I'm happy to make the change you've suggested 👍

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+});
+
+test("uses the aria-label prop if provided", () => {
```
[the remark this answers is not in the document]
robinzigmond: I take the point, and am happy to change this - just thought I'd make the point that since we're explicitly testing a prop called `aria-label`, there's an argument for making sure it actually becomes the `aria-label` of the input, rather than something else going on (eg with `aria-labeledby`) to ensure the input has the correct accessible name. In one sense that's just an implementation detail, but as I said if we're providing a prop called `aria-label` I think it would be confusing if it didn't end up as the actual HTML `aria-label` attribute.

That's mostly be just playing devil's advocate though, so having made the point I'm happy to make the change you've suggested 👍

src/components/action-popover/action-popover.test.tsx
robinzigmond: yes - that I didn't even think of it 😂 🤦 

I'm actually not 100% sure if this is what `hover` is really for, although it probably is - I'll give it a go.

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+    });
+
+    expect(
+      screen.getByRole("button", { name: "submenu item 1" })
+    ).toBeInTheDocument();
+  });
+
+  // test needed for coverage of the clearTimeout call - it's not clear if the clearTimeout is actually needed
+  // and if it has any real observable consequences. We should investigate that and remove the call, and this
+  // test, if it is not needed.
```
nineteen88, line 915: comment: let's make a ticket a for this task so it doesn't get lost

src/components/action-popover/action-popover.test.tsx
```diff
@@ -0,0 +1,2303 @@
...
+    });
+
+    expect(
+      screen.queryByRole("button", { name: "submenu item 1" })
+    ).not.toBeInTheDocument();
+  });
+
+  // test needed for coverage of the clearTimeout call - it's not clear if the clearTimeout is actually needed
+  // and if it has any real observable consequences. We should investigate that and remove the call, and this
+  // test, if it is not needed.
```
nineteen88, line 1003: comment: let's make a ticket a for this task so it doesn't get lost

src/components/action-popover/action-popover.test.tsx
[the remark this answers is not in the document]
robinzigmond: I just looked again - there's definitely no Playwright tests for at least some of the prop combinations tested here. In any case, even the PW tests aren't that good in the sense they're still testing implementation details (`data-floating-placement` attributes, which I believe are added by `floatingui` rather than us anyway) so I don't have much more confidence in those than these janky unit tests. So I'd prefer to leave them for now, although if we later review our tests and decide to move these to Playwright that would be fine.

src/components/action-popover/action-popover.test.tsx
robinzigmond: now created as FE-6722

src/components/action-popover/action-popover.test.tsx
robinzigmond: now created as FE-6722 (same as above as obviously this isn't worth 2 separate tickets)

edleeks87 (approved):

nineteen88 (approved):

ZhuoyuJin (approved):

PR #6846, opened 2024-07-19
refactor(numeral-date): convert unit tests to RTL

edleeks87:
### Proposed behaviour

Converts `NumeralDate` unit tests to RTL
### Current behaviour

Unit tests are enzyme

### Checklist

#### QA


src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -317,6 +317,7 @@ export const NumeralDate = <DateType extends NumeralDateObject = FullDate>({
     const isValidKey =
       Events.isNumberKey(event) ||
       Events.isTabKey(event) ||
+      Events.isEnterKey(event) ||
```
edleeks87, line 320: matches change added in #6837

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+          onChange={() => {}}
+          error="error"
+        />
+      </CarbonProvider>
+    );
+
+    const dayInput = screen.getByRole("textbox", { name: "Day" });
+    await user.hover(dayInput);
+
+    expect(screen.getByRole("tooltip", { name: "error" })).toBeVisible();
```
robinzigmond, line 126: **suggestion** (non-blocking): since this is the string case, it might be worth also checking that the tooltip has the expected text content (with `.toHaveTextContent("error")`

(Similarly for the other tooltip tests with a string validation prop, but I won't repeat myself for this one.)

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+        error="error"
+      />
+    </CarbonProvider>
+  );
+
+  const dayInput = screen.getByRole("textbox", { name: "Day" });
+  const monthInput = screen.getByRole("textbox", { name: "Month" });
+  const yearInput = screen.getByRole("textbox", { name: "Year" });
+
+  expect(dayInput).toBeInvalid();
```
robinzigmond, line 561: **suggestion** (non-blocking): I'm wondering if you should make these "invalid" checks into separate tests for each of these cases, just as you did for the case where `validationRedesignOptIn` wasn't set. I don't actually mind much which way we do it (in theory I slightly prefer separate tests, but we already have a LOT of validation tests for this component...), but the inconsistency bothers me.

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+    const monthInput = screen.getByRole("textbox", { name: "Month" });
+    await user.type(monthInput, "02");
+    await user.tab();
+
+    expect(
+      screen.queryByText("Day in February should be a number within 1-28.")
+    ).not.toBeInTheDocument();
+  });
+
+  it("should use the current year and not render the validation message when the user types '02' in the`Month` field and tabs when the `Day` field has a valid value", async () => {
```
robinzigmond, line 882: **comment** (non-blocking): I'm not sure what "use the current year" has to do with this test to be honest. It's testing February 28th which of course is a valid date in any year. So I would change the description to just be something like "valid dates in February should not render the validation message", although arguably that's not even needed in addition to the other tests we have.

If we do want to test the functionality of taking the current date, we'd have to mock native date functions (`Date.prototype.getFullYear` I think, having looked at the component code) - and well I'll leave it up to you to judge if that's worth it 😆

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+
+    expect(dayInput).toBeInvalid();
+    expect(monthInput).toBeInvalid();
+    expect(yearInput).toBeInvalid();
+    expect(
+      screen.getByText("Year should be a number within a 1800-2200 range.")
+    ).toBeVisible();
+  });
+
+  it("should render the appropriate validation message when the `Day` and `Month` fields are blurred and have values greater than the valid limits", async () => {
```
robinzigmond, line 1007: **comment** (non-blocking): I don't personally see the need for checking every combination of 2 of 3 fields with errors for the specific combined error message. I think just one test, ideally with errors in all 3, should give us enough confidence that all the individual errors are rendered. Maybe you need to do it this way for coverage though (I haven't checked)?

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+          "Month should be a number within a 1-12 range. " +
+          "Year should be a number within a 1800-2200 range."
+      )
+    ).toBeVisible();
+  });
+});
+
+test("should submit the form when enter key is pressed", async () => {
+  const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
+  const onSubmit = jest.fn();
```
robinzigmond, line 1541:
```suggestion
  const onSubmit = jest.fn((e) => { e.preventDefault(); });
```

**suggestion** (non-blocking): it probably doesn't matter is the test is obviously passing and I have no idea how jsdom handles form submission, but I'd feel a little more comfortably if we were explicitly preventing default so that we know there's definitely going to be no attempt at reloading the page etc.

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+    </form>
+  );
+
+  const dayInput = screen.getByRole("textbox", { name: "Day" });
+  await user.type(dayInput, "{enter}");
+
+  expect(onSubmit).toHaveBeenCalled();
+});
+
+test("should update the input values when the delete key is pressed", async () => {
```
robinzigmond, line 1558: **comment** (non-blocking): I'm a bit surprised at this test. I've not seen in any other input components that we're testing that pressing delete clears the input - this is standard browser behaviour anyway, so I don't see what value it adds. As always I might be missing something specific to this component which gives it value.

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  await user.type(dayInput, "{delete}");
+  await user.type(monthInput, "{delete}");
+  await user.type(yearInput, "{delete}");
+
+  expect(dayInput).toHaveValue("");
+  expect(monthInput).toHaveValue("");
+  expect(yearInput).toHaveValue("");
+});
+
+test("should update the input values when the backspace key is pressed", async () => {
```
robinzigmond, line 1579: **comment** (non-blocking): same comment as on previous test

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  const dayInput = screen.getByRole("textbox", { name: "Day" });
+  const monthInput = screen.getByRole("textbox", { name: "Month" });
+  const yearInput = screen.getByRole("textbox", { name: "Year" });
+
+  expect(dayInput).toBeRequired();
+  expect(monthInput).toBeRequired();
+  expect(yearInput).toBeRequired();
+});
+
+test("should render the expected inputs the `dateFormat` is set as 'mmddyyyy'", () => {
```
robinzigmond, line 1636:
```suggestion
test("should render the expected inputs when the `dateFormat` is set as 'mmddyyyy'", () => {
```

**suggestion**: typo

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  );
+
+  const inputText = screen.getAllByTestId("numeral-date-input-text");
+
+  expect(inputText[0]).toHaveTextContent("Month");
+  expect(inputText[1]).toHaveTextContent("Day");
+  expect(inputText[2]).toHaveTextContent("Year");
+});
+
+test("should render the expected inputs the `dateFormat` is set as 'yyyymmdd'", () => {
```
robinzigmond, line 1652:
```suggestion
test("should render the expected inputs when the `dateFormat` is set as 'yyyymmdd'", () => {
```

**suggestion**: typo

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  );
+
+  const inputText = screen.getAllByTestId("numeral-date-input-text");
+
+  expect(inputText[0]).toHaveTextContent("Year");
+  expect(inputText[1]).toHaveTextContent("Month");
+  expect(inputText[2]).toHaveTextContent("Day");
+});
+
+test("should render the expected inputs the `dateFormat` is set as 'ddmm'", () => {
```
robinzigmond, line 1668:
```suggestion
test("should render the expected inputs when the `dateFormat` is set as 'ddmm'", () => {
```

**suggestion**: typo

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  );
+
+  const inputText = screen.getAllByTestId("numeral-date-input-text");
+
+  expect(inputText).toHaveLength(2);
+  expect(inputText[0]).toHaveTextContent("Day");
+  expect(inputText[1]).toHaveTextContent("Month");
+});
+
+test("should render the expected inputs the `dateFormat` is set as 'mmdd'", () => {
```
robinzigmond, line 1684:
```suggestion
test("should render the expected inputs when the `dateFormat` is set as 'mmdd'", () => {
```

**suggestion**: typo

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  );
+
+  const inputText = screen.getAllByTestId("numeral-date-input-text");
+
+  expect(inputText).toHaveLength(2);
+  expect(inputText[0]).toHaveTextContent("Month");
+  expect(inputText[1]).toHaveTextContent("Day");
+});
+
+test("should render the expected inputs the `dateFormat` is set as 'mmyyyy'", () => {
```
robinzigmond, line 1700:
```suggestion
test("should render the expected inputs when the `dateFormat` is set as 'mmyyyy'", () => {
```

**suggestion**: typo

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+      value={{ dd: "", mm: "", yyyy: "" }}
+      onChange={() => {}}
+      data-component="numeral-date-root"
+      data-element="numeral-date-element"
+      data-role="numeral-date-role"
+    />
+  );
+  const rootElement = screen.getByTestId("numeral-date-role");
+
+  expect(rootElement).toHaveAttribute("data-component", "numeral-date-root");
```
robinzigmond, line 1944: **comment** (non-blocking): rather surprised that we allow `data-component` to  be overr-idden, and not even as a private prop. But apparently we do! 😱 (In which case the test is fair enough.)

src/components/numeral-date/numeral-date.component.tsx
edleeks87: yep it will be rebased this is just so I can add the tests in as it wasn't merged when this was raised

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+        onChange={() => {}}
+        labelHelp="labelHelp"
+      />
+    </CarbonProvider>
+  );
+
+  const labelHelp = screen.getByRole("button", { name: "help" });
+  await user.hover(labelHelp);
+
+  expect(screen.getByRole("tooltip", { name: "labelHelp" })).toBeVisible();
```
edleeks87, line 667: I'll update these to use `await find...` after I have a second reviewer

src/components/numeral-date/numeral-date.test.tsx
edleeks87: This is mainly for coverage, it's intentionally testing that when the year field is empty it uses the current year to validate, we don't need to mock it as it's always valid.

src/components/numeral-date/numeral-date.test.tsx
edleeks87: I agree it's odd, but we have code that implements it so imo it's better to test it now and remove if we refactor 

https://github.com/Sage/carbon/blob/3b62762ff577617f971f7fd36d9f84bf5ca468e1/src/components/numeral-date/numeral-date.component.tsx#L321

src/components/numeral-date/numeral-date.test.tsx
edleeks87: yep agree, I've chosen to avoid making such a trivial breaking change for now as I think we should batch ti with other ones etc, I'll raise something in jira for it once I've reviewed if we have other components doing the same (we likely do 😞 )

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+    const yearInput = screen.getByRole("textbox", { name: "Year" });
+    await user.hover(yearInput);
+
+    expect(screen.queryByRole("tooltip")).not.toBeInTheDocument();
+  });
+});
+
+test("should render the validation message when the `error` prop is passed a string value and `validationRedesignOptIn` is set", () => {
+  render(
+    <CarbonProvider validationRedesignOptIn>
```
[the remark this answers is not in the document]
robinzigmond: oh yeah, got copy/paste happy and hadn't realised this one was testing the other case 🤦

src/components/numeral-date/numeral-date.test.tsx
robinzigmond: I agree there's no need to mock anything for a test on February 28th as that is valid in any year, but:
- there's likely no point in a test for "february 28th is a valid date, at least if you leave out the year"
- even if we decide there is, the description of the test doesn't match what it's actually doing

I get that we need to do a case where there's no year given to hit coverage, but I think there is actually value in testing that - and I think the best way to do that is to test with February 29th and mock `getFullYear` to return say 2024 in one test (where it should be valid) and say 2023 in another (where it shouldn't be). If you disagree that that's worth a proper test, then fair enough, but at least then let's put in a "dummy test" with a comment saying it's only for coverage, not have a description that looks like it's testing the "use current year" functionality but isn't really (except in the coverage sense).

DipperTheDan, line 882: nittiest of nitpicks:

```suggestion
  it("should use the current year and not render the validation message when the user types '02' in the `Month` field and tabs when the `Day` field has a valid value", async () => {
```

DipperTheDan (commented):
Nothing more to add really. @robinzigmond seems to have been very thorough as usual with the review.

src/components/numeral-date/numeral-date.component.tsx
edleeks87: Rebased now fyi

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+
+testStyledSystemMargin(
+  (props) => <NumeralDate {...props} />,
+  undefined,
+  (component) => component.find(FormFieldStyle),
+  { modifier: "&&&" }
+);
+
+beforeEach(() => {
+  jest.useFakeTimers();
```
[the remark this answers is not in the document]
edleeks87: Yeah I think I'll look to reduce the use of these to the user interaction at the most (hopefully less than that) but the reason I added it was that it sped the spec's overall run time but I confess I didn't attempt to be more specific etc so I'll see how that goes

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+jest.mock("../../__internal__/utils/logger");
+
+testStyledSystemMargin(
+  (props) => <NumeralDate {...props} />,
+  undefined,
+  (component) => component.find(FormFieldStyle),
+  { modifier: "&&&" }
+);
+
+beforeEach(() => {
```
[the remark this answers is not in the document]
edleeks87: see [below](https://github.com/Sage/carbon/pull/6846#discussion_r1704173639)

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+  jest.useFakeTimers();
+});
+
+afterEach(() => {
+  jest.runOnlyPendingTimers();
```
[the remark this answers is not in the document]
edleeks87: The reply [here](https://github.com/Sage/carbon/pull/6846#discussion_r1704173639) explains why I've added them, I suspect I can be more specific with where it's applied as the only timeout is in the blur events

src/components/numeral-date/numeral-date.test.tsx
edleeks87: As it's running in a node environment, there's browser behaviour so there's nothing to prevent when a form submission occurs

src/components/numeral-date/numeral-date.test.tsx
edleeks87: I believe this is the approach we agreed as preferred, using the `name` property is essentially selecting it via the text and then asserting it as visible etc. I'd prefer to leave these as is for now as overall I think we've adopted this same approach in other files (we do have the approach you mention as well though) so if we decide to pivot I'd prefer to do them all at once

src/components/numeral-date/numeral-date.test.tsx
edleeks87: I'm not sure I fully agree, we have tests for leap years, this test is purely to assert that it gets the full year for the current date so the fact it's not testing leap years isn't critical for me. I'd prefer to do this than have to mock the required functionality so as not to have failures when that updates every year etc

src/components/numeral-date/numeral-date.test.tsx
edleeks87: There is some coverage concerns here but I could definitely reduce the set I have here. In the long run I'd prefer to move these to chromatic and use play functions but in the interim I'd prefer to be a little overly cautious in testing them as each input works independently and we want to guarantee the order of the messages when in the old validation pattern etc

src/components/numeral-date/numeral-date.test.tsx
```diff
@@ -0,0 +1,1995 @@
...
+        dateFormat={["xx"]}
+        value={{ dd: "", mm: "", yyyy: "" }}
+        onChange={() => {}}
+      />
+    )
+  ).toThrow(expected);
+  consoleSpy.mockReset();
+});
+
+test("should not throw an error if no `dateFormat` is passed", () => {
```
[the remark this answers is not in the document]
edleeks87: No `dateFormat` being passed is a condition within the invariant so it makes sense to cover it imo

src/components/numeral-date/numeral-date.test.tsx
edleeks87: I'm not convinced the minor discrepancy is a problem here tbh

DipperTheDan (approved):

nineteen88 (approved):

PR #6848, opened 2024-07-22
refactor(radio-button, radio-button-group): convert enzyme tests to RTL

nuria1110:
### Proposed behaviour

Convert tests to RTL

### Current behaviour

Tests are in enzyme

### Checklist

#### QA


src/components/radio-button/radio-button-group/radio-button-group.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    <RadioButtonGroup
+      name="group"
+      legend="Radio Group Legend"
+      onChange={() => {}}
+    >
+      <RadioButton value="radio1" label="Radio Button 1" />
+    </RadioButtonGroup>
+  );
+
+  const fieldset = screen.getByRole("group", { name: "Radio Group Legend" });
```
Parsium, line 67: **praise**: great job checking the ARIA roles! Your method of targeting elements mirrors how an assistive technology user would inspect the page 👍🏼

src/components/radio-button/radio-button-group/radio-button-group.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    <RadioButtonGroup name="group">
+      <RadioButton value="radio1" label="Radio Button 1" />
+    </RadioButtonGroup>
+  );
+
+  expect(loggerSpy).toHaveBeenCalledWith(
+    "Uncontrolled behaviour in `Radio Button` is deprecated and support will soon be removed. Please make sure all your inputs are controlled."
+  );
+  expect(loggerSpy).toHaveBeenCalledTimes(1);
+});
```
Parsium, line 23: **suggestion**: restoring the logger's original behavior should prevent the console from being accidentally masked in other tests:

```suggestion
    loggerSpy.mockRestore();
});
```

src/components/radio-button/radio-button-group/radio-button-group.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+      <RadioButton value="radio1" label="Radio Button 1" />
+    </RadioButtonGroup>
+  );
+
+  const radioButton = screen.getByRole("radio", { name: "Radio Button 1" });
+
+  expect(radioButton).toBeRequired();
+});
+
+test("renders with inline legend when screen is larger than `adaptiveLegendBreakpoint`", () => {
```
robinzigmond, line 130: **suggestion**: I'd feel more comfortable if there was also a test for the screen being smaller than `adaptiveLegendBreakpoint`. I think this might also avoid the need for some of the tests below that you've marked as being only for coverage (although I'm not sure about this)

src/components/radio-button/radio-button-group/radio-button-group.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+      <RadioButton value="radio1" label="Radio Button 1" />
+    </RadioButtonGroup>
+  );
+
+  const legend = screen.getByTestId("legend");
+
+  expect(legend).toHaveStyle({ float: "left" });
+});
+
+test("removes provided margin-left when screen is smaller than `adaptiveSpacingBreakpoint`", () => {
```
robinzigmond, line 148: **suggestion**: as above, I'd like to cover the other case (it took me a while to realise there were two different breakpoint props being tested in these 2 tests - I thought at first they were testing the 2 different cases of a single breakpoint prop)

src/components/radio-button/radio-button.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+});
+
+test("calls `onFocus` when provided and the radio button is focused", async () => {
+  const user = userEvent.setup();
+  const onFocus = jest.fn();
+  render(
+    <RadioButton value="radio1" label="Radio Button 1" onFocus={onFocus} />
+  );
+
+  await user.tab();
```
robinzigmond, line 51: **suggestion** (non-blocking): I'd prefer in these focus/blur tests to explicitly fetch a particular element to call `.focus()` (and then tab/click as necessary for the rest of the test) - this ensures we're focusing/blurring the correct thing and avoids any future test flakiness if we later introduce other tab stops etc

robinzigmond (commented):
Really good job - just some very minor comments to add to those of @Parsium

src/components/radio-button/radio-button.test.tsx
nuria1110: The focus method seems to call the actual HTML focus() method but this isn't getting flagged by the linter (issue raised [here](https://github.com/testing-library/eslint-plugin-testing-library/issues/752)). I do agree that using tab() might be a bit flakey, though I think using click() should be fine.

robinzigmond: hmm, I wasn't aware there was any recommendation against using `.focus()` on an HTML element, I thought this was what you needed to do to ensure a particular element was focused at the start of a test (I agree that once into the actual sequence of steps you're testing you should do "user things" like click/tab). I've done it myself in plenty of RTL tests recently and never been suggested not to 😂

src/components/radio-button/radio-button-group/radio-button-group.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    </RadioButtonGroup>
+  );
+
+  expect(loggerSpy).toHaveBeenCalledWith(
+    "Uncontrolled behaviour in `Radio Button` is deprecated and support will soon be removed. Please make sure all your inputs are controlled."
+  );
+  expect(loggerSpy).toHaveBeenCalledTimes(1);
+});
+
+test("renders with `RadioButton` children", () => {
```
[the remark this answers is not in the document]
nuria1110: Yeah this was just a basic check for the children but I agree this one may be a bit overkill, I was mainly testing the svg and label were visible separately as the actual radio input is hidden. However I'm already checking this in the `RadioButton` tests so I think may remove some assertions and try to make the test descriptions a bit more clear.

Parsium (approved):

robinzigmond (approved):

PR #6878, opened 2024-08-02
fix(simple-color-picker, advanced-color-picker): ensure blur blocking works when switching colors - FE-6707

### LINKED ISSUE #6850: [AdvancedColorPicker] - onBlur not being triggered the first time a new color is clicked
### Description

When changing the input value in AdvancedColorPicker by clicking a non-selected color, the onBlur callback should be triggered. But this does not happen the first time a change happens (it does on subsequent ones)

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-bigb4p?file=src%2FApp.tsx

### Steps to reproduce

Open the dialog with the picker in, click a non-red color, and note that there is no output in the console.
Subsequently clicking a new color fires the console message as expected.

_No response_

### Suggested solution

_No response_

### Carbon version

141.0.6

### Design tokens version

4.29.0

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

Initial investigation suggests the problem is to do with the setting of `isBlurBlocked` in the `handleMouseDown` function (https://stackblitz.com/edit/parsium-carbon-starter-bigb4p?file=src%2FApp.tsx), with the value incorrectly being `true` when the call fails. I do not know why this is happening specifically the first time, or why we are even blocking blur at all.

### Confidentiality

DipperTheDan:
Blur should be blocked when switching between colors and fired when the component itself is blurred.

fixes #6850

### Proposed behaviour

Blur event should not be fired when switching between colours.
Add deprecation warning to `isBlurBlocked` prop.

### Current behaviour

Blur event is triggered every time a colour is changed.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

New test story called `OnBlurExample` has been added to `AdvancedColorPickers` test stories. This example allows you to test when the `onBlur` event is fired as it will be logged in the console. 

Blur should only be fired if the component itself is blurred or the component is closed using the `X` button.

### DIFF
--- a/src/components/simple-color-picker/simple-color-picker.component.tsx
+++ b/src/components/simple-color-picker/simple-color-picker.component.tsx
@@ -23,6 +23,7 @@ import { ValidationProps } from "../../__internal__/validations";
 import Logger from "../../__internal__/utils/logger";
 
 let deprecateUncontrolledWarnTriggered = false;
+const isBlurBlockedDeprecateWarnTriggered = false;
 
 export interface SimpleColorPickerProps extends ValidationProps, MarginProps {
   /** The SimpleColor components to be rendered in the group */
@@ -260,9 +261,15 @@ export const SimpleColorPicker = React.forwardRef<
   const handleOnBlur = (ev: React.FocusEvent<HTMLInputElement>) => {
     ev.preventDefault();
 
-    if (!blurBlocked && onBlur) {
-      onBlur(ev);
-    }
+    setTimeout(() => {
+      const hasBlurred = !gridItemRefs?.current?.find(
+        (colorRef) => colorRef === document.activeElement
+      );
+      /* istanbul ignore else */
+      if (onBlur && hasBlurred && !blurBlocked) {
+        onBlur(ev);
+      }
+    }, 5);
   };
 
   const handleOnMouseDown = (ev: React.MouseEvent<HTMLElement>) => {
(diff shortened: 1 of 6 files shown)


nineteen88 (approved):

src/components/simple-color-picker/simple-color-picker.component.tsx
```diff
@@ -260,9 +260,15 @@ export const SimpleColorPicker = React.forwardRef<
...
 
-    if (!blurBlocked && onBlur) {
-      onBlur(ev);
-    }
+    setTimeout(() => {
+      const hasBlurred = !gridItemRefs?.current?.find(
+        (colorRef) => colorRef === document.activeElement
+      );
+      /* istanbul ignore else */
+      if (onBlur && hasBlurred && !blurBlocked) {
```
edleeks87, line 268: comment (non-blocking): I'm not convinced we need the `blurBlocked` flag anymore as the new find function should handle it as far as I can tell. If that's the case, maybe we should remove the internal state and add a deprecation warning for it? It doesn't seem to cause any issues leaving it in for now (I do think getting the dep warning in would be good) so happy to push this to a new ticket if preferred.

DipperTheDan: Yeah that is a good point. I did originally get rid of all the `blurBlocked` stuff (state/prop) and it worked with the amended blur handling code. I chose to add it back in otherwise it would have been a breaking change. I'll gladly put a deprecation warning on the prop and we can add it to the next bundle of breaking changes we do.

src/components/simple-color-picker/simple-color-picker.spec.tsx
```diff
@@ -778,3 +778,35 @@ describe("SimpleColorPicker", () => {
...
+        name="test"
+        isBlurBlocked
+      />
+    );
+
+    expect(loggerSpy).toHaveBeenCalledWith(
+      "The 'isBlurBlocked' prop in SimpleColorPicker is deprecated and support will soon be removed."
+    );
+
+    expect(loggerSpy).toHaveBeenCalledTimes(1);
```
edleeks87, line 810: comment (non-blocking): this is kind of pointless as we're only rendering one component etc but it's not a huge deal

edleeks87 (dismissed):

nineteen88 (dismissed):

paulrobinson88 (approved):

edleeks87 (approved):

nineteen88 (approved):

paulrobinson88 (approved):

PR #6879, opened 2024-08-02
fix(heading): ensure back link wrapper does not have additional link focus style

tomdavies73:
### Proposed behaviour

Applies `box-shadow: none` to the back link wrapper. Also adds `display: block` to the anchor element to make it a block level element.

Screenshot 2024-08-02 at 14 53 59

### Current behaviour

Currently, `box-shadow` is applied to the anchor element and the link wrapper, resulting in incorrect focus styling. Also, the back link anchor is treated as an inline element, which results in the element being too small on focus.

Screenshot 2024-08-02 at 14 53 45

### Checklist

#### QA


src/components/heading/heading.pw.tsx
```diff
@@ -121,6 +121,28 @@ test.describe("Heading component", () => {
     });
   });
 
+  testData.forEach((backLink) => {
```
nineteen88, line 124: comment: I'm not sure what the value of this .forEach() is

I know you're just basing it off whatever was already there, but it doesn't actually seem to loop through anything so I think it would be better not to be a .forEach()

mihai-albu-sage (approved):

nineteen88 (approved):

mihai-albu-sage (approved):

paulrobinson88 (approved):

PR #6881, opened 2024-08-05
chore(loader): convert enzyme to RTL - FE-6607

DipperTheDan:
### Proposed behaviour

- Refactor unit tests to RTL
- Add Playwright test to assert the correct CSS values are applied to the `gradient` version of Loader.

### Current behaviour

Tests currently use Enzyme.

### Checklist

#### QA

### Additional context

I have added comments where tests are required for coverage. 

### Testing instructions

N/A


src/components/loader/loader.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
```
edleeks87, line 3: comment: I don't think we need to import this in either file, if we do it should be done in the jest config (I think it is already but could be wrong)

src/components/loader/loader.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import useMediaQuery from "../../hooks/useMediaQuery";
+
+jest.mock("../../hooks/useMediaQuery", () => ({
+  __esModule: true,
+  default: jest.fn().mockReturnValue(false),
+}));
+
+testStyledSystemMargin((props) => <Loader {...props} />);
+
+test("if the user disallows animations or their preference cannot be determined, alternative loading text is rendered", () => {
```
edleeks87, line 15:
```suggestion
test("when the user disallows animations or their preference cannot be determined, alternative loading text is rendered", () => {
```

non-blocking

src/components/loader/loader.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  render(<Loader />);
+
+  expect(screen.getByText("Loading")).toBeInTheDocument();
+});
+
+describe("if the user allows animations", () => {
```
edleeks87, line 21:
```suggestion
describe("when the user allows animations", () => {
```

non-blocking

src/components/loader/loader-square.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+  });
+});
+
+// These styling tests required for coverage
+test("when inside button and `isActive` prop is false, the expected background colour is applied", () => {
+  render(<Loader isInsideButton isActive={false} />);
+
+  const loaderSquares = screen.getAllByTestId("loader-square");
+
+  expect(loaderSquares[0]).toHaveStyle({
```
edleeks87, line 88: comment: I think we need to use toHaveStyleRule here as the jest DOM matcher returns false positives when testing token values

edleeks87 (dismissed):

src/components/loader/loader.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    render(<Loader />);
+
+    expect(screen.getByLabelText("Loading")).toBeInTheDocument();
+  });
+
+  test("when custom `aria-label` is passed, set accessible name to its value", () => {
+    render(<Loader aria-label="Still loading" />);
+
+    expect(screen.getByLabelText("Still loading")).toBeInTheDocument();
+  });
```
Parsium, line 46: **suggestion(non-blocking)**: what do you think of using the `toHaveAccessibleName` assertion, so the error messaging aligns more closely to each test's objective?:

```suggestion
    expect(screen.getByRole("progressbar")).toHaveAccessibleName("Loading");
  });

  test("when custom `aria-label` is passed, set accessible name to its value", () => {
    render(<Loader aria-label="Still loading" />);

    expect(screen.getByRole("progressbar")).toHaveAccessibleName("Still loading");
  });
```

DipperTheDan: Yeah, I'm happy to do that. I'll get on it now ✈️

Parsium (approved):

edleeks87 (approved):

stephenogorman (approved):

PR #6884, opened 2024-08-07
fix(radio-button, checkbox): prevent labelHelp and fieldHelp from rendering with new validation

nuria1110:
### Proposed behaviour

RadioButton with validationRedesignOptIn:
- Does not render labelHelp tooltip. 
- Does not render fieldHelp.
- Does not render with validation (error/warning) tooltip. 
- Required stars do not appear in a required group's children. 

Checkbox within a group with validationRedesignOptIn:
- Does not render labelHelp tooltip. 
- Does not render fieldHelp.

### Current behaviour

RadioButton with validationRedesignOptIn:
- Renders labelHelp tooltip. 
- Renders fieldHelp.
- Renders with validation (error/warning) tooltip. 
- Required stars also appear in a required group's children. 

Checkbox within a group with validationRedesignOptIn:
- Renders labelHelp tooltip. 
- Renders fieldHelp.

image

### Checklist

#### QA

### Testing instructions

- Note that individual Checkboxes that aren't in a group can still render labelHelp & fieldHelp with new validation. 
- See this [demo](https://stackblitz.com/edit/parsium-carbon-starter-i4wbci?file=src%2FApp.tsx) for issues.
- See new test stories, `WithNewValidation`, `WithNewValidationGroup` for both components.

### DIFF
--- a/src/components/checkbox/checkbox.mdx
+++ b/src/components/checkbox/checkbox.mdx
@@ -63,16 +63,6 @@ import { Checkbox, CheckboxGroup } from "carbon-react/lib/components/checkbox";
 
 <Canvas of={CheckboxStories.Reversed} />
 
-### With fieldHelp
-
-<Canvas of={CheckboxStories.WithFieldHelp} />
-
-### With labelHelp
-
-**Note:** The `legendHelp` tooltip will be rendered as hint text if the `validationRedesignOptIn` flag on the `CarbonProvider` is true.
-
-<Canvas of={CheckboxStories.WithLabelHelp} />
-
 ### With custom labelWidth
 
 <Canvas of={CheckboxStories.WithCustomLabelWidth} />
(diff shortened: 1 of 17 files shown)


src/components/radio-button/radio-button.component.tsx
Parsium, line 1: **comment**: I noticed that the old labeling props aren't currently flagged as legacy in the docs for `Checkbox` and `RadioButton`.

What do you think about refactoring the validation flag logic from `Checkbox` and `RadioButton` into their common ancestor, `CheckableInput`, and marking the `labelHelp` and `fieldHelp` props as legacy there? This approach would reduce code duplication and ensure that these changes are automatically applied to all descendents of `CheckableInput`.

src/components/radio-button/radio-button.mdx
```diff
@@ -52,6 +52,8 @@ import {
 
 A legend can be set for the group, and each radio button can have a label.
 
+**Note**: The `labelHelp` prop is not supported if the `validationRedesignOptIn` flag on the `CarbonProvider` is true.
+
 <Canvas of={RadioButtonStories.WithLegendAndLabels} />
```
Parsium, line 57: **suggestion(non-blocking)**: what do you about updating the original story to use the new validation designs by default? This could reduce confusion for consumers who might overlook the statement above, ensuring they clearly see what the latest behaviour in Carbon should be.

src/components/radio-button/radio-button.mdx
```diff
@@ -107,6 +109,8 @@ The radio buttons and labels can be disabled using the `disabled` prop on each r
 
 Help text for each input can be added with the `fieldHelp` prop on `RadioButton`. This can be made inline with the `fieldHelpInline` prop.
 
+**Note**: The `fieldHelp` prop is not supported if the `validationRedesignOptIn` flag on the `CarbonProvider` is true.
```
Parsium, line 112: **suggestion(non-blocking)**: what do you think about removing this section on the fieldHelp prop, since it is now part of a legacy design?

src/components/checkbox/checkbox.mdx
```diff
@@ -69,8 +69,6 @@ import { Checkbox, CheckboxGroup } from "carbon-react/lib/components/checkbox";
 
 ### With labelHelp
```
Parsium, line 70: **suggestion(non-blocking)**: similar to `RadioButton`, what do you think about removing this section from the public facing docs, to reduce potential confusion over current behaviour?

src/components/radio-button/radio-button.mdx
edleeks87: I don't think it hurts to have it further documented somewhere (we've marked it legacy but we've still not had any steer wrt when it actually gets removed etc) but I do agree with the comment elsewhere that we should display the nee designs first and the legacy ones secondary to that.

src/components/radio-button/radio-button.mdx
```diff
@@ -54,29 +54,12 @@ A legend can be set for the group, and each radio button can have a label.
 
 <Canvas of={RadioButtonStories.WithLegendAndLabels} />
 
-### With inline legend
```
edleeks87, line 57: We still have the WithInlineLegend and EnableAdaptiveBehaviour stories in the tsx file, if we're satisfied this behaviour is adequately covered elsewhere and they aren't imported in another file we should remove them. We do need to ensure that any regressions are captured so it will need to either be unit/playwright/chromatic that does it as until the old layout is deprecated we still need to ensure it works etc

edleeks87: Same needs confirming in all the stories we've updated, I agree the first thing consumers see should be the latest layout designs etc but we still have some who work on legacy projects so we can't not have something in the docs relating to that etc

src/components/radio-button/radio-button-test.stories.tsx
```diff
@@ -138,52 +183,52 @@ export const WithTooltipPositionOnRadioGroup: StoryFn<
 WithTooltipPositionOnRadioGroup.storyName =
   "with tooltip position on RadioGroup";
 
-const radioContainerWidth = 400;
-
-export const RadioButtonComponent = (props: Partial<RadioButtonProps>) => {
-  const [isChecked, setIsChecked] = React.useState(false);
+export const WithNewValidation = (props: Partial<RadioButtonProps>) => {
```
Parsium, line 186: **suggestion(non-blocking)**: what do you think about removing the `WithNewValidation` and `WithNewValidationGroup` test stories, in favour of the consumer-facing stories `NewValidationDefault` and `NewValidationDefaultGroup`? It appears the latter two will be captured by a Chromatic snapshot, whereas the former two won't (disabled on line 23).

Parsium (approved):

paulrobinson88 (approved):

Issue #6860, opened 2024-07-25
Allow configuration of minHeight and padding for TextArea component


harryharford:
### Description

Currently the TextArea component has hardcoded minHeight and padding values https://github.com/Sage/carbon/blob/master/src/components/textarea/textarea.style.ts. 

It is therefore not possible to use this component with only 1 row height, as it does not fit cleanly on a page that also includes TextBox components. 

### Suggested solution

Make the minHeight and padding values configurable and set the current values to defaults. 

### Alternatives

Using TextBox in this case is not suitable as the field needs to expand and needs to support carriage returns. 

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
Discussed with @ljemmo on slack, agreed that min-height can be configurable with the current set as default. There are plans to add some padding presets in but after discussing with @harryharford I don't think they need that right now so we can probably just update the min-height

nicktitchmarsh:
FE-6726

PR #6887, opened 2024-08-07
fix(textarea): add minheight prop to Textarea

mihai-albu-sage:
fixes: #6860

### Proposed behaviour
Optional `minHeight?: number;` prop added to Textarea. If no value is passed it will use the same default value as it currently uses which is `64px`.

### Current behaviour
The Textarea component does not allow for the `min-height` to be configurable and uses a default value of `64px`.

### Checklist

#### QA


src/components/textarea/textarea.style.ts
```diff
@@ -6,13 +6,15 @@ import { StyledLabelContainer } from "../../__internal__/label/label.style";
...
-export const MIN_HEIGHT = 64;
+export const DEFAULT_MIN_HEIGHT = 64;
 
 export interface StyledTextAreaProps {
   /** When true, label is placed in line an input */
   labelInline?: boolean;
   /** When true, adjusts padding for icon */
   hasIcon?: boolean;
+  /** Specify the mimimum height */
+  minHeight?: number;
```
nineteen88, line 17: suggestion: for consistency we should probably keep the type as string

maxWidth is a string and we shouldn't really limit the user to using px when they might want relative heights

src/components/textarea/textarea.component.tsx
```diff
@@ -129,6 +129,8 @@ export interface TextareaProps
   borderRadius?: BorderRadiusType | BorderRadiusType[];
   /** Hides the borders for the component. Please note that validation and focus styling will still be applied */
   hideBorders?: boolean;
+  /** Specify the minimum height */
+  minHeight?: number;
```
nineteen88, line 133: suggestion: for consistency we should probably keep the type as string

maxWidth is a string and we shouldn't really limit the user to using px when they might want relative heights

src/components/textarea/textarea.style.ts
nineteen88, line 17: suggestion: rather than setting the prop type in two locations it would be better to extend the base Textarea with a Pick<>

You can see this being doing in other components, for example Card

mihai-albu-sage: Refactor the code to include the above suggestion.

mihai-albu-sage: The `minHeight` is used in a comparison against the `scrollHeight`, which is of type number which represent a pixel value.

In order to allow the `minHeight` to be a string and accept relative height units, we would need a helper function that allows us to dynamically convert the CSS relative units into pixels in a way that is accurate within the current rendering context of the document so that we can continue to make the comparison mentioned above which is necessary for the correct behavior of the component.

An example of how that function could look like is the following:
```
function convertHeightToPixels(heightStr, contextElement = document.body) {
    // Create a temporary element
    const tempElement = document.createElement('div');
 
    // Set the height of the temporary element to the value we want to convert
    tempElement.style.height = heightStr;
 
    // Append the element to the provided context element (default is body)
    contextElement.appendChild(tempElement);
 
    // Get the computed height in pixels
    const pixels = parseFloat(getComputedStyle(tempElement).height);
 
    // Remove the temporary element
    contextElement.removeChild(tempElement);
 
    // Return the height in pixels
    return pixels;
}
 
// Example usage:
const heightInPercent = convertHeightToPixels('10%', <Textarea parent element>);
console.log(heightInPercent);  // Output will be 10% of the height of the parent element in pixels.
```
Another thing we have to consider is that we will have to trigger the above function in case of browser resize events, since the `minHeight` representation in pixels will change as well.

While passing the `minHeight` as a string increases the flexibility of the component, I believe it also ads a layer of complexity to this component that maybe is not required? Please let me know your thoughts on this.

nineteen88: The level of complexity required due to the Scroll Height needing to be calculated as a fixed number does seem cumbersome. I'll open a ticket in order to do some further investigation on this later. It's frustrating to not be able to support relative heights out of the gate but it is what it is for now

nineteen88 (dismissed):

src/components/textarea/textarea.style.ts
DipperTheDan: We do have the `resizeObserver` hook in Carbon that could be used for the browser resize stuff, so you wouldn't need to write you own from scratch. The above function is like cutting down a whole tree and setting it all on fire just to light a cigar though. I'm OK with what Will has suggested with regards to opening a separate ticket for an investigation though 👍🏻

DipperTheDan (dismissed):

DipperTheDan (approved):

stephenogorman (approved):

PR #6893, opened 2024-08-13
refactor(card): convert tests to RTL

nineteen88:
### Proposed behaviour

Tests for Card should be simplified and converted to RTL

### Current behaviour

There are too many tests that shouldn't be in unit tests and they are currently written in Enzyme

### Checklist

#### QA


src/components/card/card-column/card-column.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { render, screen } from "@testing-library/react";
+import CardColumn from "./card-column.component";
+
+test("renders children within CardColumn as expected", () => {
+  const content = <div>content</div>;
+  render(<CardColumn>{content}</CardColumn>);
+  expect(screen.getByText("content")).toBeInTheDocument();
+});
+
+test("has the expected data attributes when they are passed in", () => {
```
edleeks87, line 11: question: Do we need to have two renders here? The below would cover it I think:

```
test("has the expected data attributes when they are passed in", () => {
render(
    <CardColumn data-element="foo" data-role="bar">
      <div>content</div>
    </CardColumn>
  );
  
const cardColumnElementWithAttributes = screen.getByTestId("bar"); // data-element is implicitly asserted but we can explicitly if needed
  expect(cardColumnElementWithAttributes).toHaveAttribute(
    "data-component",
    "card-column"
  );
  expect(cardColumnElementWithAttributes).toHaveAttribute(
    "data-element",
    "foo"
);
});

File: src/components/card/card-footer/card-footer.test.tsx [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+          roundness,
+          spacing: "medium",
+        }}
+      >
+        <CardFooter data-role="card-footer">foo</CardFooter>
+      </CardContext.Provider>
+    );
+
+    const cardFooterElement = screen.getByTestId("card-footer");
+    expect(cardFooterElement).toHaveStyle({
```
edleeks87, line 34: comment: I think toHaveStyle returns true no matter what when token values are asserted so we might need to use toHaveStyleRule in the interim

src/components/card/card-footer/card-footer.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+        roundness === "default" ? "1" : "2"
+      }00)`,
+      borderBottomRightRadius: `var(--borderRadius${
+        roundness === "default" ? "1" : "2"
+      }00)`,
+    });
+  }
+);
+
+test("has the expected data attributes when they are passed in", () => {
```
edleeks87, line 45: comment: again not sure if we need to renders here but there's defaults so not a huge deal to leave as is

src/components/card/card-row/card-row.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    render(
+      <CardContext.Provider value={{ spacing }}>
+        <CardRow data-role="card-row">
+          <div />
+        </CardRow>
+      </CardContext.Provider>
+    );
+
+    const cardRowElement = screen.getByTestId("card-row");
+    expect(cardRowElement).toHaveStyle({
```
edleeks87, line 33: comment: same comment as above

src/components/card/card.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <Card roundness={roundness} data-role="card">
+        Content
+      </Card>
+    );
+    const borderRadiusValue = roundness === "default" ? "100" : "200";
+    const radius = `var(--borderRadius${borderRadiusValue})`;
+
+    const cardElement = screen.getByTestId("card");
+
+    expect(cardElement).toHaveStyle(`border-radius: ${radius}`);
```
edleeks87, line 45: comment: same as above, worth investigating etc

DipperTheDan (approved):

DipperTheDan (approved):

edleeks87 (approved):

PR #6897, opened 2024-08-15
refactor(grouped-character): convert tests to RTL - FE-6596

DipperTheDan:
### Proposed behaviour

Tests should use RTL

### Current behaviour

Tests currently use Enzyme

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  await user.clear(input);
+  await user.type(input, "1222");
+  await input.setSelectionRange(6, 6);
+  await user.tab();
+
+  expect(input.selectionStart).toBe(7);
+  expect(input.selectionEnd).toBe(7);
+});
+
+test("pressing backspace after a separator moves the cursor backwards to positions", async () => {
```
robinzigmond, line 372:
```suggestion
test("pressing backspace after a separator moves the cursor backwards two positions", async () => {
```

**suggestion** - I *think* this is what you mean, and it's a simple typo, but I'm far from certain. What you actually had certainly doesn't make sense though, so please correct it appropriately if I've misinterpreted

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+      />
+    );
+  };
+
+  render(<MockComponent />);
+
+  input = screen.getByRole("textbox");
+
+  await user.click(input);
+  await user.keyboard("{arrowLeft}");
```
robinzigmond, line 399: **comment/praise**: I think these arrowleft/arrowright combinations are a better way of getting the cursor to a specific position than using `setSelectionRange`: it's a much closer fit to what a user might do.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  expect(input.selectionEnd).toBe(5);
+
+  await user.keyboard("1");
+  jest.runAllTimers();
+
+  expect(input.selectionStart).toBe(7);
+  expect(input.selectionEnd).toBe(7);
+});
+
+test("when no character is present, cursor position goes to 0", async () => {
```
robinzigmond, line 461: **comment** (non-blocking): I'm not sure this test is really adding much value. Perhaps it's for coverage, in which case I guess it's fine

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  input = screen.getByRole("textbox");
+
+  await user.click(input);
+  jest.runAllTimers();
+
+  expect(input.selectionStart).toBe(0);
+  expect(input.selectionEnd).toBe(0);
+});
+
+test("pressing backspace when the cursor is after the middle character in the final group, should move the cursor one position backwards", async () => {
```
robinzigmond, line 494: **comment** (non-blocking): this test description is quite confusing, mostly because it seems weirdly specific. Is it that important that it's specifically the middle character in the final group (that only makes sense if the final group is of size 3, or I guess 5,7,etc)? I suspect this is really just testing that if you press backspace and you're not just after a separator, then it only goes back 1 space - but I'm far from certain. If that is the case I would try to simplify the test description so that it isn't mentioning a very specific case which isn't relevant.

src/components/grouped-character/grouped-character.component.tsx
```diff
@@ -90,7 +90,7 @@ export const GroupedCharacter = React.forwardRef(
       const { target } = ev;
       const { selectionEnd } = target as HTMLInputElement;
 
-      let newCursorPos = selectionEnd ?? 0;
+      let newCursorPos = selectionEnd ?? /* istanbul ignore next */ 0;
```
robinzigmond, line 93: **comment** (non-blocking): I'm a little uneasy at the `istanbul ignore` here. There are a lot of tests involving selection positions so I'm surprised this hasn't ended up being covered, if it presumably was before. I'd prefer to have something covering it but can live with the ignore if the test would really end up being not very relevant. I'm just curious if we know which Enzyme test was covering it that has now been removed, and if there's a sensible way to convert it.

src/components/grouped-character/grounded-character.test.tsx
DipperTheDan: Yep, spelling mistake on my part 😄

src/components/grouped-character/grounded-character.test.tsx
DipperTheDan: Yes it is for coverage. I think your comment has raised a valid point that I should probably put a comment of sorts here saying it's for coverage as it does seem pointless to have.

src/components/grouped-character/grounded-character.test.tsx
DipperTheDan: Yeah it's to hit a `else if` condition in the `lastPosition` block in the `handleChange` function. I might change it to include the words that roughly mean `isAtOneBeyondSeparator` as that is where this test covers in the code.

Code block I am referring to: 
https://github.com/Sage/carbon/blob/master/src/components/grouped-character/grouped-character.component.tsx#L109-L112

src/components/grouped-character/grouped-character-test.stories.tsx
```diff
@@ -78,3 +83,39 @@ NewValidation.args = {
   separator: "-",
   ...getCommonTextboxArgs(),
 };
+
+export const IsOptionalExample = () => {
```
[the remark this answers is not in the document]
DipperTheDan: Yep, thank you for pointing this out. I did this without checking we actually had Chromatic taking snapshots here.

src/components/grouped-character/grouped-character.component.tsx
DipperTheDan: I'll have another look through the tests we had in Enzyme but I converted all the selectionRange tests. We did do the tests a bit different in Enzyme with simulating certain the change event and mocking the selectionRange. I'll see if I can manually trigger this line of code and write a test doing what I did manually.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+  render(<MockComponent />);
+
+  input = screen.getByRole("textbox");
+
+  await user.click(input);
+  await user.clear(input);
+  await user.type(input, "1222");
+  await input.setSelectionRange(6, 6);
+  await user.tab();
```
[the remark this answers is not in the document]
DipperTheDan: Looking at it again, I've given it the wrong test description. I've gone for `placing cursor at the point where a separator should appear moves cursor forward one position`

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  input = screen.getByRole("textbox");
+
+  await user.click(input);
+  await user.clear(input);
+  await user.type(input, "12-34-");
+  await input.setSelectionRange(6, 6);
+  await user.keyboard("{backspace}");
+  jest.runAllTimers();
+
+  expect(input.selectionStart).toBe(4);
```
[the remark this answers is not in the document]
DipperTheDan: The two separate test cases were done this way in Enzyme and it's to test the different groupings. I just left as is and converted as I figured they were tested this way for a valid reason. 

The selection range puzzle I believe is due to the fact that a 6 digit value, you will also have two separators `-` to consider when the value is formatted. Where we're pressing backspace here will also remove the separator which the `handleChange` in the component handles. I'm happy to change these tests to more inline with what you did in Decimal if it inspires more confidence in the tests?

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+  const lastOnBlurCall = onBlur.mock.calls[onBlur.mock.calls.length - 1][0];
+
+  expect(lastOnBlurCall).toMatchObject({
+    formattedValue: "12-34-56",
+    rawValue: "123456",
+  });
+});
+
+test("does nothing if onBlur is not provided", async () => {
```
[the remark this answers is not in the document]
DipperTheDan: Yeah I agree. I just converted it as it was being tested so I assumed it was originally being tested for a reason. I cannot see any particular code in the component that would mean we need to test this specific case. I'll remove it.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+  expect(screen.getByRole("textbox")).toHaveValue("12-34-5678");
+});
+
+test("emits a formatted string on change event", async () => {
+  onChange = jest.fn();
+  onBlur = jest.fn();
+  const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
+
+  const MockComponent = () => {
```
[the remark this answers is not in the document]
DipperTheDan: I could do that and I did think of it but my assumption was we wanted everything to be declared in the test itself as much as possible. We'll see what a second reviewer says about it and we'll go with the majority.

src/components/grouped-character/grounded-character.test.tsx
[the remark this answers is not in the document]
DipperTheDan: Turns out I do need it for coverage so I've had to add it back in. I'll add a comment stating that it is purely for coverage too.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+    <GroupedCharacter
+      separator="-"
+      defaultValue="aabbcccc"
+      groups={[2, 2, 4]}
+      onChange={onChange}
+    />
+  );
+
+  await user.type(screen.getByRole("textbox"), "cc-aa-aabb");
+  const lastOnChangeCall =
```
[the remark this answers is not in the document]
DipperTheDan: Thanks for this. I've sorted all these.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  const input = screen.getByRole("textbox");
+
+  await user.click(input);
+  await user.clear(input);
+  await user.type(input, "123456789");
+
+  expect(input).toHaveValue("12-34-5678");
+});
+
+let input: HTMLInputElement;
```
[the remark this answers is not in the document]
DipperTheDan: Moved this in to the tests. I think I just left it here from the Enzyme file as it declared at the start of the `describe` block.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+test("pressing backspace after a separating character moves cursor backwards if the cursor positioned at the end of input", async () => {
```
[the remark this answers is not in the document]
DipperTheDan: Ah yeah, I merged the `describe` block title with the `it` block and it made this title. I'll get them reworded as they are a bit difficult to read.

src/components/grouped-character/grounded-character.test.tsx
DipperTheDan, line 1: I realise the name of this file is incorrect. I have amended this to it's proper name

src/components/grouped-character/grouped-character.component.tsx
DipperTheDan: After some investigation we've found it's safe to ignore this. It was covered by [this test case](https://github.com/Sage/carbon/blob/master/src/components/grouped-character/grouped-character.spec.tsx#L91-L99) which is only testing the onChange handler, not specifically the null side of the nullish coalescing operator. It seems that it was covered by accident and was done by simulating change rather the RTL method of testing actual user behaviours.

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  await user.keyboard("{arrowLeft}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  await user.keyboard("{arrowRight}");
+  jest.runAllTimers();
```
Parsium, line 530: **suggestion**: it seems we don’t need to exhaust the fake timers after a userEvent action, since the action is already wrapped with `act` to keep React in sync with the UI. What do you think about removing those cases?

```suggestion
```

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+
+import userEvent from "@testing-library/user-event";
+import GroupedCharacter from "./grouped-character.component";
+import { testStyledSystemMargin } from "../../__spec_helper__/__internal__/test-utils";
+import FormFieldStyle from "../../__internal__/form-field/form-field.style";
+import Logger from "../../__internal__/utils/logger";
+
+jest.mock("../../__internal__/utils/logger");
+
+jest.useFakeTimers();
```
Parsium, line 12: **suggestion(non-blocking)**: what do you think about amending the fake timer setup like so, to ensure timer code from one test doesn’t affect the next?   

```suggestion
beforeEach(() => {
  jest.useFakeTimers();
})

afterEac(() => {
  jest.runOnlyPendingTimers();
  jest.useRealTimers();
})
```

src/components/grouped-character/grounded-character.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+import { render, screen } from "@testing-library/react";
+
+import userEvent from "@testing-library/user-event";
+import GroupedCharacter from "./grouped-character.component";
+import { testStyledSystemMargin } from "../../__spec_helper__/__internal__/test-utils";
+import FormFieldStyle from "../../__internal__/form-field/form-field.style";
+import Logger from "../../__internal__/utils/logger";
+
+jest.mock("../../__internal__/utils/logger");
+
```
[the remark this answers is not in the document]
DipperTheDan: Messaged privately about this but just so it's visible on the PR. We've agreed to leave this line in as the tests testing the deprecation warning fail without it.

src/components/grouped-character/grouped-character.test.tsx
```diff
@@ -0,0 +1,451 @@
...
+  const input = screen.getByRole("textbox");
+
+  await user.click(input);
+  await user.clear(input);
+  await user.type(input, "123456789");
+
+  expect(input).toHaveValue("12-34-5678");
+});
+
+test("pressing backspace after a separating character moves cursor the backwards, if the cursor is positioned at the end of input", async () => {
```
robinzigmond, line 259:
```suggestion
test("pressing backspace after a separating character moves the cursor backwards, if the cursor is positioned at the end of input", async () => {
```

**suggestion** - seems to have been a typo left in

src/components/grouped-character/grouped-character.test.tsx
```diff
@@ -0,0 +1,451 @@
...
+      formattedValue: "12-34-56",
+      rawValue: "123456",
+    })
+  );
+});
+
+// coverage
+// This test scenario may seem obvious but it is required for coverage
+test("does nothing if onBlur is not provided", async () => {
+  const onChange = jest.fn();
```
robinzigmond, line 213: **suggestion** (non-blocking): I wasn't going to bother to leave this comment, since the test is a useless one purely for coverage, but since I've ended up leaving another comment I thought I might as well...

this is another `onChange` that doesn't really have anything to do with the behaviour being tested. The "relevant" thing here is the `onBlur` (and its lack of being passed to the component), whether `onChange` is passed or not seems irrelevant.

Unless of course that turns out to be needed for coverage, in which case no bother - but I'd be very surprised.

src/components/grouped-character/grouped-character.test.tsx
```diff
@@ -0,0 +1,451 @@
...
+  await user.clear(input);
+  await user.type(input, "12-");
+  input.setSelectionRange(3, 3);
+  await user.keyboard("{backspace}");
+
+  expect(input.selectionStart).toBe(1);
+  expect(input.selectionEnd).toBe(1);
+});
+
+test("placing cursor at the point where a separator should appear, it moves the cursor forward one position", async () => {
```
robinzigmond, line 285: **suggestion** (non-blocking): it's really not clear what "it" means here without reading the test. (Reading the test, "it" appears to be "pressing tab", which confuses me as I mentioned [in a comment on my last review](https://github.com/Sage/carbon/pull/6897#discussion_r1723386541) - but so be it I guess. The original description was "pressing a character", which to me means a letter or number, I'd rather change the description to something like that if we can change the test so that it's doing that rather than tab.)

src/components/grouped-character/grounded-character.test.tsx
robinzigmond: thanks, the description reads much better now 👍

src/components/grouped-character/grouped-character.test.tsx
```diff
@@ -0,0 +1,451 @@
...
+
+import userEvent from "@testing-library/user-event";
+import GroupedCharacter, { CustomEvent } from "./grouped-character.component";
+import { testStyledSystemMargin } from "../../__spec_helper__/__internal__/test-utils";
+import FormFieldStyle from "../../__internal__/form-field/form-field.style";
+import Logger from "../../__internal__/utils/logger";
+
+jest.mock("../../__internal__/utils/logger");
+
+const MockComponent = ({
```
Parsium, line 12: **suggestion(non-blocking)**: what do you think about renaming this to `ControlledGroupedCharacter`, so its more descriptive? Not essential though.

Parsium (approved):

robinzigmond (approved):

robinzigmond (approved):

PR #6905, opened 2024-08-20
refactor(decimal): convert enzyme tests to react-testing-library

robinzigmond:
### Proposed behaviour

tests for Decimal component should use react-testing-library

### Current behaviour

tests use Enzyme

### Checklist

#### QA


src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+      });
+
+      expect(decimalInput).toHaveValue(result);
+      expect(screen.getByTestId("hidden-input")).toHaveValue(result);
+    }
+  );
+
+  // insertions at start and end have to be handled separately from the cases above, as setting the
+  // initialSelectionStart to 0 or 4 (the default input length) seems to result in the entire value
+  // being selected. So we explicitly use arrow-key manipulation to handle these cases correctly.
```
DipperTheDan, line 458: **praise:** Good idea warning people in the future why we had to do this 😄

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+      expect(screen.getByTestId("hidden-input")).toHaveValue(sanitisedValue);
+    }
+  );
+
+  // as above, a simple cursor placement (no selection) at the start or end need special handling
+  it("doesn't delete anything when backspace is pressed with the cursor at the start of the input", async () => {
+    const user = userEvent.setup();
+    render(<Decimal defaultValue="1234.56" />);
+
+    const decimalInput = screen.getByRole("textbox") as HTMLInputElement;
```
DipperTheDan, line 552: **suggestion:** I don't think you need the `as HTMLInputElement` here. I don't think you're doing anything that requires the input element type.

```suggestion
    const decimalInput = screen.getByRole("textbox");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+
+    expect(decimalInput).toHaveValue("1,234.56");
+    expect(screen.getByTestId("hidden-input")).toHaveValue("1234.56");
+  });
+
+  it("allows deletion of the final character using the backspace key with the cursor at the end of the input", async () => {
+    const user = userEvent.setup();
+    render(<Decimal defaultValue="1234.56" />);
+
+    const decimalInput = screen.getByRole("textbox") as HTMLInputElement;
```
DipperTheDan, line 565: **suggestion:** same thing might apply here too.

```suggestion
    const decimalInput = screen.getByRole("textbox");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+      decimalInput.setSelectionRange(selectionStart, selectionEnd);
+      await user.keyboard("{Delete}");
+
+      expect(decimalInput).toHaveValue(rawValue);
+      expect(screen.getByTestId("hidden-input")).toHaveValue(sanitisedValue);
+    }
+  );
+
+  // as above, a simple cursor placement (no selection) at the start or end need special handling
+  it("deletes the fisrt character when delete is pressed with the cursor at the start of the input", async () => {
```
DipperTheDan, line 635: **nitpick:** Spelling mistake.
```suggestion
  it("deletes the first character when delete is pressed with the cursor at the start of the input", async () => {
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+    const user = userEvent.setup();
+    render(<Decimal defaultValue="1234.56" />);
+
+    const decimalInput = screen.getByRole("textbox") as HTMLInputElement;
```
DipperTheDan, line 639: **suggestion:** Same thing here too.

```suggestion
    const decimalInput = screen.getByRole("textbox");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+
+    expect(decimalInput).toHaveValue(",234.56");
+    expect(screen.getByTestId("hidden-input")).toHaveValue(",234.56");
+  });
+
+  it("doesn't delete anything when delete is pressed with the cursor at the end of the input", async () => {
+    const user = userEvent.setup();
+    render(<Decimal defaultValue="1234.56" />);
+
+    const decimalInput = screen.getByRole("textbox") as HTMLInputElement;
```
DipperTheDan, line 652: **suggestion:** and here again too.
```suggestion
    const decimalInput = screen.getByRole("textbox");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+    );
+  });
+
+  it("adds the expected HTML attributes to the hidden input", async () => {
+    const user = userEvent.setup();
+    render(<Decimal name="example" />);
+
+    await user.type(screen.getByRole("textbox"), "1.23");
+    const hiddenInput = screen.getByTestId("hidden-input");
+    expect(hiddenInput).toHaveAttribute("data-component", "hidden-input");
```
DipperTheDan, line 704: **nitpick:** Just a little format tweak to match the others here.

```suggestion
    const hiddenInput = screen.getByTestId("hidden-input");
    
    expect(hiddenInput).toHaveAttribute("data-component", "hidden-input");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+  });
+});
+
+test("the `required` prop is not passed to the hidden input", () => {
+  render(<Decimal label="required" required />);
+
+  expect(screen.getByTestId("hidden-input")).not.toBeRequired();
+});
+
+test("the `isOptional` prop adds an '(optional)' suffix to the label", () => {
```
DipperTheDan, line 1398: **suggestion:** Unless this is for coverage and a second reviewer agrees, I think you could move this test to be a Chromatic example as it's testing styling. `Decimal` does have the snapshots disabled in the test stories though.

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+    (precisionValue, decimalValue) => {
+      render(
+        <Decimal defaultValue={decimalValue} precision={precisionValue} />
+      );
+      expect(screen.getByRole("textbox")).toHaveValue(decimalValue);
+      expect(screen.getByTestId("hidden-input")).toHaveValue(decimalValue);
+    }
+  );
+
+  it("pads the value with 0s if the precision is greater than the nunber of decimal places", () => {
```
edleeks87, line 197:
```suggestion
  it("pads the value with 0s if the precision is greater than the number of decimal places", () => {
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+    );
+    expect(loggerSpy).toHaveBeenCalledTimes(1);
+
+    loggerSpy.mockRestore();
+  });
+
+  it("has a default value of 0.00 when no defaultValue prop is provided", () => {
+    render(<Decimal />);
+    expect(screen.getByRole("textbox")).toHaveValue("0.00");
+    expect(screen.getByTestId("hidden-input")).toHaveValue("0.00");
```
edleeks87, line 51:
```suggestion
    expect(screen.getByRole("textbox", { hidden: true })).toHaveValue("0.00");
```

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+      const input = char.repeat(4);
+      await user.type(screen.getByRole("textbox"), input);
+      await user.tab();
+
+      expect(screen.getByRole("textbox")).toHaveValue(input);
+      expect(screen.getByTestId("hidden-input")).toHaveValue(input);
+    }
+  );
+
+  it("calls the onKeyPress callback when a key is pressed", async () => {
```
edleeks87, line 676: comment: good to see us using a deprecated keyboard event 🤦, I'll see if we have a ticket to update this and raise one if not

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+      "decimal"
+    );
+  });
+
+  it("adds the expected HTML attributes to the hidden input", async () => {
+    const user = userEvent.setup();
+    render(<Decimal name="example" />);
+
+    await user.type(screen.getByRole("textbox"), "1.23");
+    const hiddenInput = screen.getByTestId("hidden-input");
```
edleeks87, line 703: comment: if we use `getByRole("textbox", { hidden: true })` we would get some of the below assertions implicitly

src/components/decimal/decimal.test.tsx
```diff
@@ -0,0 +1,1406 @@
...
+  expect(consoleSpy).not.toHaveBeenCalled();
+
+  consoleSpy.mockRestore();
+});
+
+test("the `required` prop is passed to the visible input and adds an asterisk to the label", () => {
+  render(<Decimal label="required" required />);
+
+  expect(screen.getByRole("textbox")).toBeRequired();
+  expect(screen.getByText("required")).toHaveStyleRule("content", '"*"', {
```
edleeks87, line 1387: comment (non-blocking): as we have chromatic here I don't think the style assertion is needed in this test

src/components/decimal/decimal.test.tsx
edleeks87: FE-6766 raised

src/components/decimal/decimal.test.tsx
robinzigmond: as discussed privately, I don't like having a blank line between the `const hiddenInput = ...` and the subsequent assertions, as that `const` is part of the "assertion step" of the test, it just helps avoid repetition in those assertions.

I do agree though that there's a blank line missing in here, but think it should go *above* the `const` (after the `await user.type(...)`). I'll get that changed.

src/components/decimal/decimal.test.tsx
robinzigmond: you're right - I checked Playwright tests and couldn't find one for Optional, but it should be covered by Chromatic as the [relevant story](https://github.com/Sage/carbon/blob/master/src/components/decimal/decimal.stories.tsx#L174) doesn't seem to have it disabled. I'll therefore simply remove this as suggested 👍

src/components/decimal/decimal.test.tsx
robinzigmond: as discussed privately, this change doesn't work - it still picks up the visible input not the hidden one (the `{hidden: true}` just modifies the query so it can also pick up any text inputs that are hidden from the accessibility tree through eg `aria-hidden` - but an `<input type="hidden" />` doesn't even have the textbox role so the query doesn'g find it). As a result, although the change works on this particular test it only does so because both inputs have the same value (and it's just repeating the assertion on the line above), it fails on many other tests.

src/components/decimal/decimal.test.tsx
robinzigmond: [as noted above](https://github.com/Sage/carbon/pull/6905#discussion_r1724844170), this selector won't work

DipperTheDan (approved):

edleeks87 (approved):

PR #6909, opened 2024-08-21
add flat-table enhancements

tomdavies73:
#### `FlatTable`

- Add `hasOuterVerticalBorders` prop  - toggles visibility of table's outer left and right borders
- Add `bottomBorderRadius` prop - for customising the table's bottom border radius. Accepts the following tokens:
        - "borderRadius000"
        - "borderRadius010"
        - "borderRadius025"
        - "borderRadius050"
        - "borderRadius100"
        - "borderRadius200"
        - "borderRadius400"

After these additions a minimal design can now be achieved:

Screenshot 2024-08-27 at 16 20 01

Screenshot 2024-08-27 at 16 20 09

### Current behaviour

Currently, there is no support for horizontal-only borders between cells, and the bottom border radius of the table cannot be customised.

### Checklist

#### QA

### Testing instructions

- Validate the functionality of the new “Minimal Design” story. Ensure it meets expectations and behaves as intended.
- Review the new `hasOuterVerticalBorders` and `bottomBorderRadius` props added to `FlatTable`. Assess if they function correctly and identify any potential issues that may arise from introducing the props. Both props can be experimented with using the controls for the "Default" `FlatTable` story.

### DIFF
--- a/src/components/flat-table/flat-table-row/flat-table-row.style.ts
+++ b/src/components/flat-table/flat-table-row/flat-table-row.style.ts
@@ -237,7 +237,14 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
       ${StyledFlatTableHeader} {
         border-bottom: 1px solid ${borderColor(colorTheme)};
 
+        ${horizontalBorderSize !== "small" &&
+        css`
+          border-bottom: ${horizontalBorderSizes[horizontalBorderSize]} solid
+            var(--colorsUtilityMajor100);
+        `}
+
         ${!isInSidebar &&
+        !colorTheme?.includes("transparent") &&
         `
           :first-child {
             border-left: 1px solid ${borderColor(colorTheme)};
--- a/src/components/flat-table/flat-table-test.stories.tsx
+++ b/src/components/flat-table/flat-table-test.stories.tsx
@@ -36,6 +36,7 @@ export default {
     "FlatTableSizeFocus",
     "FlatTableInsideDrawer",
     "FlatRowHeaderWithNoPaddingAndButtons",
+    "MinimalDesign",
   ],
   parameters: {
     info: { disable: true },
--- a/src/components/flat-table/flat-table.component.tsx
+++ b/src/components/flat-table/flat-table.component.tsx
@@ -1,5 +1,6 @@
 import React, { useLayoutEffect, useRef, useState, useContext } from "react";
 import { MarginProps } from "styled-system";
+import * as DesignTokens from "@sage/design-tokens/js/base/common";
 
 import {
   StyledFlatTableWrapper,
@@ -36,6 +37,13 @@ export interface FlatTableProps extends MarginProps {
   size?: "compact" | "small" | "medium" | "large" | "extraLarge";
   /** Applies max-height of 100% to FlatTable if true */
   hasMaxHeight?: boolean;
+  /** Toggles the visibility of the table's outer left and right borders. When false, the left border of the first column and the right border of the last column are hidden.  */
+  hasOuterVerticalBorders?: boolean;
+  /** Sets the border radius of the first and last cells in the last row. */
+  bottomBorderRadius?: Extract<
+    keyof typeof DesignTokens,
+    `borderRadius${number}`
+  >;
   /** Set the overflow X of the table wrapper. Any valid CSS string */
   overflowX?: string;
   /** Width of the table. Any valid CSS string */
@@ -56,6 +64,8 @@ export const FlatTable = ({
   isZebra,
   size = "medium",
   hasMaxHeight = false,
+  hasOuterVerticalBorders = true,
+  bottomBorderRadius = "borderRadius100",
   ariaDescribedby,
   minHeight,
   overflowX,
@@ -219,6 +229,7 @@ export const FlatTable = ({
     <StyledFlatTableWrapper
       ref={wrapperRef}
       data-component="flat-table-wrapper"
+      data-role="flat-table-wrapper"
       isInSidebar={isInSidebar}
       hasStickyHead={hasStickyHead}
       colorTheme={colorTheme}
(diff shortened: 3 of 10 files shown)


src/components/flat-table/flat-table-row/flat-table-row.component.tsx
```diff
@@ -23,6 +24,12 @@ import SubRowProvider, { SubRowContext } from "./__internal__/sub-row-provider";
 import { buildPositionMap } from "../__internal__";
 import FlatTableHeadContext from "../flat-table-head/__internal__/flat-table-head.context";
 
+type DesignTokensType = keyof typeof DesignTokens;
+export type BorderRadiusType = Extract<
```
edleeks87, line 28: Comment: we seem to be re-using this type from the `Box` component so imo it's worth doing the same here. We should probably review whether we want this type to be publicly exposed or not, if we do need to move it internal having them all imported in a consistent way will limit the likelihood we miss one etc

edleeks87: I also think we should limit this a bit more, we don't want to support `borderRadiusCirlce` for example. I'm also not convinced we want to support this on every row, so it might be better moving the prop to `FlatTable` and restricting it to the first and last rows etc

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -237,7 +261,15 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
...
         border-bottom: 1px solid ${borderColor(colorTheme)};
 
+        ${horizontalBorderSize !== "small" &&
+        css`
+          border-bottom: ${horizontalBorderSizes[horizontalBorderSize]} solid
+            var(--colorsUtilityMajor100);
+        `}
+
         ${!isInSidebar &&
+        colorTheme !== "transparent-white" &&
```
edleeks87, line 271: suggestion (non-blocking):
```suggestion
        !colorTheme.includes("transparent") && `
```

src/components/flat-table/flat-table-test.stories.tsx
```diff
@@ -789,3 +790,49 @@ export const FlatRowHeaderWithNoPaddingAndButtons = () => {
     </FlatTable>
   );
 };
+
+export const MinimalDesign = () => {
+  return (
+    <FlatTable colorTheme="transparent-white" hasNoHorizontalBorders={false}>
```
edleeks87, line 796: comment: this isn't defined on the `FlatTable` interface, it's `horizontalBorders` that I think needs using here

edleeks87: I'm a bit confused by this implementation overall though, the `horizontalBorders` says it will make them transparent but below we also have `horizontalBorderColor="--colorsUtilityMajor050"` etc set on all the body rows. I think a name like `hasVerticalBorders` would be the most intuitive name but `verticalBorders` would also work

src/components/flat-table/flat-table.pw.tsx
```diff
@@ -117,6 +117,11 @@ const tableSorting: ["first" | "second", string, number, number][] = [
   ["second", "asc", 2, 1],
 ];
 
+const basicColorThemes: [FlatTableProps["colorTheme"], string, string][] = [
```
edleeks87, line 120: suggestion (non-blocking): looks like you could potentially `slice` this sub array off the `colorThemes` or you could do something like:

```
const basicColorThemes: [FlatTableProps["colorTheme"], string, string][] = [
  ["dark", "rgb(51, 91, 112)", "rgb(102, 132, 148)"],
  ["light", "rgb(204, 214, 219)", "rgb(179, 194, 201)"],
];

const colorThemes: [FlatTableProps["colorTheme"], string, string][] = [
  ...basicColorThemes,
["transparent-base", "rgb(242, 245, 246)", "rgb(242, 245, 246)"],
  ["transparent-white", "rgb(255, 255, 255)", "rgb(255, 255, 255)"],
  ];

File: src/components/flat-table/flat-table-row/flat-table-row.component.tsx (hunk 1 above)
Comment (edleeks87) [reply src/components/flat-table/flat-table-row/flat-table-row.component.tsx:28]: @Parsium I still think it's worth restricting this a bit more, if we don't then do it later it's a breaking change etc

File: src/components/flat-table/flat-table.component.tsx [hunk 5]
```diff
@@ -36,8 +36,8 @@ export interface FlatTableProps extends MarginProps {
   size?: "compact" | "small" | "medium" | "large" | "extraLarge";
   /** Applies max-height of 100% to FlatTable if true */
   hasMaxHeight?: boolean;
-  /** If `false` the border color of the first and last `FlatTableCell`'s in every `FlatTableRow` will be made transparent */
-  horizontalBorders?: boolean;
+  /** Controls whether to render table cells with visible vertical borders. */
```
edleeks87, line 39: suggestion: it might be worth adding something about how when false it will hide the outer borders of the table (border left for first column and border right for last column etc)

src/components/flat-table/flat-table.mdx
```diff
@@ -171,11 +171,11 @@ By using the `hasMaxHeight` prop you can automatically apply a max-height of 100
 
 <Canvas of={FlatTableStories.WithHasMaxHeight} />
 
-### With horizontalBorders
+### Removing vertical borders
 
-When the `horizontalBorders` prop is `false` the border color of the first and last `FlatTableCell`'s in every `FlatTableRow` will be made transparent.
+Setting the `hasVerticalBorders` prop to `false` will remove all vertical borders from the table cells.
```
edleeks87, line 176: comment: I think it's only supposed to remove/hide the left and right borders

edleeks87 (dismissed):

nineteen88 (dismissed):

nineteen88 (dismissed):

edleeks87 (dismissed):

edleeks87 (approved):

edleeks87:
LGTM @Parsium I'll add a new test story as part of my RTL refactor to capture the alt background functionality on the headers

stephenogorman (approved):

PR #6914, opened 2024-08-27
fix(carbon-provider): add aria-hidden to token provider

mihai-albu-sage:
### Proposed behaviour
Added an optional prop `ariaHidden?: "true" | "false"` to the `CarbonProvider` which is applied to the inner `CarbonScopedTokensProvider`

### Current behaviour
Currently there is no way to control the aria-hidden attribute for the `CarbonProvider` component.

### Checklist

#### QA

### DIFF
--- a/src/components/carbon-provider/carbon-provider.component.tsx
+++ b/src/components/carbon-provider/carbon-provider.component.tsx
@@ -15,6 +15,7 @@ export interface CarbonProviderProps extends NewValidationContextProps {
   children: React.ReactNode;
   /** Theme which specifies styles to apply to all child components. Set to `sageTheme` by default. */
   theme?: Partial<ThemeObject>;
+  ariaHidden?: "true" | "false";
 }
 
 export const CarbonProvider = ({
@@ -23,6 +24,7 @@ export const CarbonProvider = ({
   validationRedesignOptIn = false,
   roundedCornersOptOut = false,
   focusRedesignOptOut = false,
+  ariaHidden = "false",
 }: CarbonProviderProps) => {
   const {
     roundedCornersOptOut: existingRoundedCornersOptOut,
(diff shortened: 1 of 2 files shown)


src/components/carbon-provider/carbon-provider.component.tsx
```diff
@@ -15,6 +15,7 @@ export interface CarbonProviderProps extends NewValidationContextProps {
   children: React.ReactNode;
   /** Theme which specifies styles to apply to all child components. Set to `sageTheme` by default. */
   theme?: Partial<ThemeObject>;
+  ariaHidden?: "true" | "false";
```
edleeks87, line 18: comment: I don't think we need this as a configurable prop, we could probably just add the attribute as I don't think. If it does need to be a prop we should match the attribute case:

```

  "aria-hidden"?: "true" | "false";
  
  ...
  "aria-hidden": ariaHidden = "false",

File: src/components/carbon-provider/carbon-provider.test.tsx [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+        <Button buttonType="primary">Button</Button>
+      </CarbonProvider>
+    );
+
+    const button = screen.getByRole("button", {
+      name: "Button",
+    });
+
+    // eslint-disable-next-line testing-library/no-node-access
+    const parent = button.closest("div");
```
edleeks87, line 19: comment: I'd prefer to add a `data-role` to the div and locate it with `getByTestId`

edleeks87 (commented):
I think it's fine to add the `aria-hidden` attribute but I don't believe it's going to fully address the issue laid out in the ticket:

```
Issue in Copilot project where CarbonProvider is wrapping a menuItem due to it being loaded from elsewhere. Extra div around an <li> element is causing issues with screen readers.
```
Voiceover may now be addressed but it will still render with incorrect HTML semantics so we may need to either conditionally render the `ScopedTokensProvider` or use the `as` prop to morph it into a different/valid element (not sure it's possible). If we did conditionally render the provider we will need to confirm that it does not remove the tokens etc

PR #6915, opened 2024-08-27
refactor(dialog-full-screen): convert tests to react-testing-library

robinzigmond:
### Proposed behaviour

DialogFullScreen should have unit tests written using react-testing-library

### Current behaviour

unit tests use Enzyme

### Checklist

#### QA


src/components/dialog-full-screen/dialog-full-screen.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+import StyledContent from "./content.style";
+import StyledIconButton from "../icon-button/icon-button.style";
+import { StyledHeader, StyledHeading } from "../heading/heading.style";
+import Form from "../form";
+import { StyledFormContent } from "../form/form.style";
+import CarbonProvider from "../carbon-provider";
+import Logger from "../../__internal__/utils/logger";
+
+// mock Logger.deprecate so that no console warnings occur while running the tests
+const loggerSpy = jest.spyOn(Logger, "deprecate");
```
Parsium, line 15: **suggestion**: it seems no warnings are raised if the spy is removed. To avoid potentially masking legitimate deprecation warnings in the future, what do you think about removing this?

```suggestion
```

src/components/dialog-full-screen/dialog-full-screen.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+  render(
+    <DialogFullScreen open>
+      <input type="text" />
+    </DialogFullScreen>
+  );
+
+  jest.runAllTimers();
+
+  expect(screen.getByRole("dialog")).toHaveFocus();
+});
```
Parsium, line 88: **suggestion**: for the focus-related tests in this file, what do you think about replacing the use of `jest.runAllTimers` with `waitFor`? This would be more closely aligned to RTL's recommendations on handling async behaviour:

```suggestion
test("the dialog container should be focused when the dialog opens", async () => {
  render(
    <DialogFullScreen open>
      <input type="text" />
    </DialogFullScreen>
  );

  await waitFor(() => {
    expect(screen.getByRole("dialog")).toHaveFocus();
  });
});
```

src/components/dialog-full-screen/dialog-full-screen.test.tsx
robinzigmond: Thanks, I'd just realised in the tests I'm working on now, in a different component, that the `loggerSpy` can be removed without warnings appearing in the console - which surprised me, but that probably means it's unneeded here too 👍

Note that I don't agree with the point about "masking legitimate deprecation warnings in the future" - such deprecation warnings still shouldn't appear in the console output when running the tests, as they're not useful there and just add noise to the output (which was why these mocks were added in the first place). But I agree it's better not to mock if it's not needed, so I'll remove it.

src/components/dialog-full-screen/dialog-full-screen.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+
+// these tests should be rewritten to use toHaveStyle, to avoid passing when they shouldn't. The first should be:
+// - toHaveStyle({ "overflow-y": "auto" })
+// and the second
+// - toHaveStyle({ "overflow-y": "inherit" })
+// These don't work at the moment due to :has not being implemented in jsdom before v23.2.0, which would require a
+// jest upgrade.
+// toHaveStyleRule is more forgiving as it passes when the rule is passed to the styled component even if it's
+// overridden by another rule on the same component
```
Parsium, line 312: **suggestion(non-blocking)**: how about raising a ticket for this and rewording the comment to make it more actionable? Here is a proposed suggestion: 

```suggestion
/* TODO: <ticket number> Update these tests to use toHaveStyle to avoid false positives:
   - toHaveStyle({ "overflow-y": "auto" })
   - toHaveStyle({ "overflow-y": "inherit" })
   This requires upgrading jsdom, as :has() is not supported in jsdom before v23.2.0.
   toHaveStyleRule is currently used, but it is less strict and passes even if styles are overridden. */
```

Additionally, [once Carbon supports Jest v29](https://github.com/Sage/carbon/pull/6916), we could consider upgrading JSDOM independently to Jest in the near future if it doesn't cause issues with Jest.

src/components/dialog-full-screen/dialog-full-screen.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+});
+
+test("the onCancel callback should be called when the close button is clicked", async () => {
+  const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
+  const onCancel = jest.fn();
+  render(<ControlledDialog onCancel={onCancel} />);
+
+  await user.click(screen.getByRole("button", { name: "Close" }));
+  jest.runAllTimers();
+  expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
```
Parsium, line 406: **suggestion**: for these `onCancel` tests, what do you think about replacing `jest.runAllTimers` with [RTL's `waitForElementToBeRemoved`](https://testing-library.com/docs/dom-testing-library/api-async#waitforelementtoberemoved)? This would be more closely aligned to RTL's recommendations on waiting for an element's disappearance:

```suggestion
  await waitForElementToBeRemoved(() => screen.queryByRole("dialog"));
```

src/components/dialog-full-screen/dialog-full-screen.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+  const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
+  const onCancel = jest.fn();
+  render(<ControlledDialog onCancel={onCancel} />);
+
+  screen.getByRole("button", { name: "Close" }).focus();
+  await user.keyboard("a");
+  jest.runAllTimers();
+  expect(screen.getByRole("dialog")).toBeVisible();
+  expect(onCancel).not.toHaveBeenCalled();
+});
```
Parsium, line 432: **suggestion**: similarly what do you think about using `waitFor` here instead?: 

```suggestion

  expect(screen.getByRole("dialog")).toBeVisible();
  await waitFor(() => {
    expect(onCancel).not.toHaveBeenCalled();
  });
```

src/components/dialog-full-screen/dialog-full-screen.test.tsx
robinzigmond: happy to do this for this test, but note that for the next one where we're testing the `disableAutoFocus` prop, we shouldn't use `waitFor` as that won't fail if the prop's functionality ever gets broken - the dialog is always initially unfocused (it only becomes so after a delay) so "waiting" for it not to be focused will always pass. That one only fails in the broken case by using `runAllTimers`. I'll add a clarifying comment to that test to explain it.

src/components/dialog-full-screen/dialog-full-screen.test.tsx
robinzigmond: I'm nervous about the `waitFor` approach here, for the same reason as I explained in [this comment](https://github.com/Sage/carbon/pull/6915#discussion_r1740869722) - waiting for something to be visible that's already visible isn't going to fail even if we were to accidentally cause the dialog to close (after a small delay) in response to a non-Enter press.

I realise your comment is about the `not.toHaveBeenCalled` assertion rathre than the `toBeVisible` one - but in that case it seems strange to me to use `waitFor`, as you can't wait for something *not* to be have been called, as once the function has been called that status is never going to change, no matter how long you wait!

So I think I'll leave this test as it is, with an explanatory comment as I explained in the reply I just linked to

src/components/dialog-full-screen/dialog-full-screen.test.tsx
Parsium: I see your point about ensuring the `disableAutoFocus` functionality is tested. However, I believe using `waitFor` would be more appropriate and idiomatic with RTL in this case.

The goal is to verify that the dialog remains unfocused and does not receive focus after any potential UI updates or asynchronous behaviour completes, rather than simply checking if the `disableAutoFocus` code path is executed.

That said, I’m not strongly opposed to the current approach if you feel it fits this specific scenario here. My only concern is that, in general, we avoid tying our tests too closely to implementation details.

robinzigmond: > The goal is to verify that the dialog remains unfocused and does not receive focus after any potential UI updates or asynchronous behaviour completes, rather than simply checking if the disableAutoFocus code path is executed.

Exactly. This isn't about coverage at all, and I'm not sure where I mis-spoke to give you that impression. It's about ensuring the test works.

Using `waitFor(() => expect(screen.getByRole("dialog")).not.toBeFocused()` doesn't do what you rightly say the goal is, because it is saying "keep checking for a short period and if the dialog is focused for that whole time, fail the test". But, by design, the dialog is not focused at the point the `waitFor` starts, so this won't ever fail.

What we want is to say "keep checking for a short period and if the dialog ever moves from being unfocused to being focused, then fail the test". Unfortunately I don't know of a good way to do that - `runAllTimers` isn't great because it's very implementation-specific (because we happen to use `setTimeout` I think deep inside the code that does the focusing), so I'd love to have a better way, but `waitFor` isn't designed for these situations. That's for "I know this isn't true right now, but it should be in a little bit", which is ideal for tests like this specific one you've commented on where we want to check the dialog gets focused. It doesn't help with "this is actually true right now, but I want to make sure it becomes false in a little bit", which is what we're testing in the `disableAutoFocus` case.

Hope this makes more sense now, but please let me know if not!

Parsium: Ah thanks for clarifying this @robinzigmond! I now see what you mean about the assertion being satisfied immediately due to the dialog initially having focus. Happy to stick with this 👍🏼

Parsium (approved):

edleeks87 (approved):

PR #6916, opened 2024-08-27
chore: consume JSDOM fix to enhance performance of `window.getComputedStyle`

Parsium:
### Proposed behaviour

- Upgrade jsdom to version 21.1.x - to consume a performance enhancement to `window.getComputedStyle`
- Upgrade jest and jest-environment-jsdom to v29.x
- Fix CI workflow to parallelise jest tests
- Upgrade @testing-library/jest-dom to v6.x

### Current behaviour

- Tests involving the rendering of floating elements, such as tooltips, inconsistently fail
- On v27.x of jest and jest-environment-jsdom
- Jest tests are not parallelised in CI - due to jest being on v27.x

### Checklist

#### QA

### Additional context

This change replaces the default version of JSDOM that comes with jest-environment-jsdom to enhance the performance of `getComputedStyle`.

The @floating-ui/dom library, heavily used in Carbon, frequently calls `getComputedStyle`. Due to its slow performance in JSDOM, these calls could cause tests to fail inconsistently if they timed out before `getComputedStyle` completed.

[Link to JSDOM fix](https://github.com/jsdom/jsdom/pull/3482)


jest.config.json
```diff
@@ -12,7 +12,6 @@
   "testMatch": ["**/?(*.)+(spec|test).[jt]s?(x)"],
   "testPathIgnorePatterns": ["node_modules", "lib", "esm"],
   "moduleDirectories": ["src", "node_modules"],
-  "collectCoverage": true,
```
edleeks87, line 15: Comment: I understand the need to not collect the coverage report when running in CI mode but I wonder if we can add something to conditionally opt-in/out when in CI mode as we can set an `env` in the workflows. I think it's a better approach than adding a new `coverage` script but I'm okay with it if the above isn't possible

```
const isCI = process.env.CI === 'true';

...

collectCoverage: !isCI

edleeks87 (commented):
Great work @Parsium 👏 . Just the one comment about coverage from me. I do also wonder if we should have the `revert` commit later in the order of commits so if we ever have to rollback we have that temp fix there instead of falling back on the flakey implementation we saw before but it's a non-blocker for me and I don't think it's worth making that change now that you've done  a lot of work on top etc

edleeks87 (approved):

nineteen88 (approved):

PR #6919, opened 2024-08-28
feat(preview): add `shape` and `disableAnimation` props

nuria1110:
### Proposed behaviour

- Adds `shape` prop - "text", "rectangle", "rectangle-round" and "circle". 
image
- Adds the `disableAnimation` prop to disable the loading animation, this can be set manually or will automatically be true when prefer reduce-motion is enabled.
- Updates background colour to a gradient.

### Current behaviour

- Preview does not have a `shape` prop.
- Preview does not handle prefer reduce-motion. 

### Checklist

#### QA

### Testing instructions

- Shape variations should support changes to dimensions using `width` and `height` props. 
- The "circle" shape uses `height` as its diameter, ignoring the `width`. 
- The `lines` prop should render specified number of Preview shapes.
- When shape is "text", if more than one line, the last line should be 80% of the width.  
- Prefer reduce-motion should remove animation.


src/components/preview/preview.component.tsx
```diff
@@ -1,40 +1,63 @@
...
+  };
+
+  const reduceMotion = !useMediaQuery(
+    "screen and (prefers-reduced-motion: no-preference)"
+  );
+
   if (hasPlaceholder) {
     const placeholders = [];
 
     for (let i = 1; i <= lines; i++) {
```
edleeks87, line 50: suggestion (non-blocking): might be worth seeing if we can iterate from 0 here so it's a but more intuitive etc

src/components/preview/preview.component.tsx
```diff
@@ -1,40 +1,63 @@
...
 }
 
 export const Preview = ({
   children,
   loading,
   lines = 1,
+  height,
+  width,
+  shape = "text",
+  isStatic,
```
edleeks87, line 33: comment: I notice we have a similar prop [here](https://github.com/Sage/carbon/blob/2783aceed1b13b5bdaee9cc7ec87ab8fea118b94/src/components/popover-container/popover-container.component.tsx#L183), I think it might be worth using the same name in this component for consistency

src/components/preview/preview.component.tsx
```diff
@@ -1,40 +1,63 @@
...
   ...props
 }: PreviewProps) => {
   const marginProps = filterStyledSystemMarginProps(props);
   const hasPlaceholder = loading === undefined ? !children : loading;
```
edleeks87, line 37: suggestion (non-blocking): feel free to ignore this as not part of the changes you've made, but we could use [nullish coalescence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) here to clean the code up a bit imo

```suggestion
  const hasPlaceholder = loading ?? !children;
```

src/components/preview/preview.style.ts
```diff
@@ -1,14 +1,106 @@
...
+    case "rectangle-round":
+      return "40px";
+    case "circle":
+      return "56px";
+    default:
+      return "14px";
+  }
+}
+
+function getWidth(shape: Shapes) {
```
edleeks87, line 54: suggestion (non-blocking): as there's essentially only 2 values being returned here you could use an if/else to remove one check here but it's not critical etc

```
function getWidth(shape: Shapes) {
  if (shape.includes("rectangle")) {
    return "120px";
  }

  return "100%";
}

File: src/components/preview/preview.style.ts [hunk 5]
```diff
@@ -1,14 +1,106 @@
...
+    case "rectangle-round":
+      return "--borderRadius400";
+    case "circle":
+      return "--borderRadiusCircle";
+    default:
+      return "--borderRadius100";
+  }
+}
+
+function getHeight(shape: Shapes) {
```
edleeks87, line 41: suggestion (non-blocking): you could consolidate the returns fro the rectangle shapes here but it's not critical what you have here is more than fine

```
function getHeight(shape: Shapes) {
  if (shape.includes("rectangle")) {
    return "40px";
  }

  switch (shape) {
    case "circle":
      return "56px";
    default:
      return "14px";
  }
}

File: src/components/preview/preview.style.ts [hunk 6]
```diff
@@ -1,14 +1,106 @@
...
+  switch (shape) {
+    case "rectangle":
+      return "40px";
```
edleeks87, line 44: suggestion: might be worth using the `--sizing` tokens for these values

image

src/components/preview/preview-test.stories.tsx
```diff
@@ -11,17 +13,15 @@ export default {
...
-    height: "",
-    lines: 1,
-    loading: true,
-    width: "",
-  },
+Default.args = {
+  children: "Text rendered as children component.",
+  lines: 1,
+  loading: true,
 };
```
Parsium, line 27: **suggestion(non-blocking)**: how about writing the story in [CSF 3.0](https://storybook.js.org/blog/storybook-csf3-is-here/)? Not essential, but it could align well with Storybook's latest standards.

```suggestion
export const Default: Story = {
  args:  {
    children: "Text rendered as children component.",
    lines: 1,
    loading: true,
  }
};
```

edleeks87 (approved):

Parsium (dismissed):

dariusbercea-sage (dismissed):

Parsium (approved):

edleeks87 (approved):

dariusbercea-sage (approved):

PR #6946, opened 2024-09-11
refactor(popover): rewrite jest tests to use RTL

Parsium:
### Proposed behaviour

- Rewrite tests for `Popover` using React Testing Library (RTL)

### Current behaviour

- Tests for `Popover` are currently use Enzyme

### Checklist

#### QA


src/__internal__/popover/popover.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    <main>
+      <PopoverWithButton disableBackgroundUI />
+      <p>Lorem ipsum dolor sit amet...</p>
+    </main>
+  );
+
+  const backdrop = screen.getByTestId("popup-backdrop");
+  expect(backdrop).toHaveStyle("background: transparent");
+  expect(backdrop).toHaveStyle(`z-index: ${baseTheme.zIndex.popover}`);
+});
```
Parsium, line 148: I recognise that this test doesn’t verify the actual scroll-blocking behaviour. Ideally, that would be the preferred approach. However, I encountered difficulties implementing a solution that works within JSDOM. If anyone has suggestions or alternative approaches, I’d appreciate the input.

src/__internal__/popover/popover.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+            <div>I float!</div>
+          </Popover>
+        )}
+      </dialog>
+    );
+  };
+  const user = userEvent.setup();
+  render(<DialogWithPopover />);
+
+  // mount the popover
```
robinzigmond, line 100: **question** (non-blocking): not a big deal, but I'm curious why in this test we're using a user action (button click) to trigger the popover opening, when in the previous tests we are simply making assertions about an always-open popover. This test is only about where in the DOM it is rendered, which shouldn't be affected by precisely when the component was mounted.

src/__internal__/popover/popover.test.tsx
robinzigmond: I think what you've done is fine, as I can't think of a better way either. Maybe a Playwright test would be better - I don't think we need to test these styles in Jest for coverage, although I could be wrong - but I see there isn't currently a Playwright test file for this component, and it's probably not worth introducing one just for this.

I do have some small suggestions though to improve this test (imo) as it stands:
1) I would also test for the `position: fixed` style, with the positioning style properties (`top` etc) all set to 0, as that is what keeps it fixed to the entire viewport, which in combination with the high z-index is what prevents the background from being scrollable. (At least I think so - I haven't experimente.)
2) `toHaveStyle` [can also take an object as an argument](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavestyle) with all the styles that you want to test. This feels less awkward to me than testing strings of CSS, and also allows all the styles to be tested in a single assertion rather than having to repeat the assertion for each style being tested.

robinzigmond (commented):
**praise**: well done for completely rewriting the existing test suite - it's now doing a much better job of testing the component's behaviour rather than implementation details 👏

src/__internal__/conditional-wrapper/conditional-wrapper.component.tsx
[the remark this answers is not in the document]
Parsium: For context, I created this as I noticed in several places where we have follow a similar conditional wrapper pattern. However, I appreciate refactoring part of `Popover` is not the sole reason for this PR, so I'm happy to revert it if the second reviewer agrees 👍🏼

src/__internal__/popover/popover.test.tsx
Parsium: Conditional rendering of the `Popover` is necessary here because without it, the button ref passed to reference remains uninitialised, causing the test to fail. While I agree that refactoring how the component handles React refs would be a cleaner long-term solution, I opted for this approach to maintain consistency with the previous implementation and ensure the test passes for now.

Parsium: Yeah I too feel this is a bug, I'll raise a ticket for this 👍🏼

robinzigmond (dismissed):

src/components/pages/pages.test.tsx
```diff
@@ -152,6 +152,33 @@ test("navigating to the next page should render the first page when currently on
   expect(screen.getByRole("heading")).toHaveTextContent("My First Page");
 });
 
+test("attempting to navigate pages when there is only one page should not change the rendered content", async () => {
```
DipperTheDan, line 155: **suggestion(non-blocking):** It took me a few attempts to read and understand it. Maybe something like this to make the description more readable. 

```suggestion
test("when attempting to navigate pages and there is only one page, it should not change the rendered content", async () => {
```

DipperTheDan (approved):

robinzigmond (approved):

src/__internal__/popover/popover.test.tsx
Parsium: @robinzigmond Raised as **FE-6819**

PR #6960, opened 2024-09-17
fix(navigation-bar): move context provider to not wrap parent

robinzigmond:
The FixedNavigationBarContextProvider, which uses a ResizeObserver to respond to size changes on the navbar element, currently wraps that same element. This can result in unexpected ResizeObserver errors on initial render, when a media query is being used to determine layout. The context provider only needs to wrap the navbar's children, and by moving it down the React tree to only wrap the children and not the parent, the error goes away.

### Proposed behaviour

No ResizeObserver errors should be raised when a NavigationBar component is rendered

### Current behaviour

ResizeObserver errors are raised on initial render when padding is set by a media query

### Checklist

#### QA

### Testing instructions

A Test story has been added to demonstrate the bug scenario and allow manual testing - when run without the code change that fixes the issue, this results in an error message related to ResizeObserver being rendered on the screen. (In Safari, this can also be seen on the Javascript console.) With the fix in place, this error no longer appears (either on screen or in the console, in any browser).

When regression testing, please pay particular attention to the NavigationBarWithSubmenuAndChangingHeight test story, as this is the one most likely to have been affected by this change (but I do not believe it will have been).


nineteen88 (approved):
The change on the surface seems fine, just not sure whether it will have any knock of effects. If there's any issues though, it should flag up in QA so happy to approve

edleeks87 (approved):

dariusbercea-sage (approved):

PR #6967, opened 2024-09-18
refactor(accordion): convert test to react-testing-library

robinzigmond:
### Proposed behaviour

Accordion component unit tests should use React Testing Library

### Current behaviour

Tests use Enzyme

### Checklist

#### QA


src/components/accordion/accordion.test.tsx
```diff
@@ -0,0 +1,456 @@
...
+});
+
+describe("Accordion", () => {
+  it("should display deprecation warning for `scheme` prop once", () => {
+    render(<Accordion scheme="transparent" title="Title" />);
+
+    expect(loggerSpy).toHaveBeenCalledWith(
+      "The `scheme` prop for `Accordion` component is deprecated and will soon be removed."
+    );
+    expect(loggerSpy).toHaveBeenCalledTimes(1);
```
DipperTheDan, line 36: **question(non-blocking):** If we have only one Accordion component won't the loggerSpy only be called once anyway?

src/components/accordion/accordion.test.tsx
```diff
@@ -0,0 +1,456 @@
...
+    render(
+      <Accordion expanded title="Title">
+        child content
+      </Accordion>
+    );
+
+    expect(screen.getByText("child content")).toBeVisible();
+  });
+
+  it("mounts collapsed when `expanded prop` is passed as false", () => {
```
DipperTheDan, line 68: **nitpick:**
```suggestion
  it("mounts collapsed when `expanded` prop is passed as false", () => {
```

src/components/accordion/accordion.test.tsx
```diff
@@ -0,0 +1,456 @@
...
+    });
+  });
+
+  it("adds a subtitle when `subTitle` prop is set and `size` is large (default)", () => {
+    render(<Accordion title="Title" subTitle="a subtitle" />);
+
+    expect(screen.getByText("a subtitle")).toBeVisible();
+  });
+
+  it("does not adds a subtitle when `subTitle` prop is set and `size` is small", () => {
```
DipperTheDan, line 209: **nitpick:**
```suggestion
  it("does not add a subtitle when `subTitle` prop is set and `size` is small", () => {
```

src/components/accordion/accordion.test.tsx
robinzigmond: 🤷 I put this in because the description includes the word "once". I'll want to remove that word if I remove that assertion - I don't think it makes much of a difference so happy to do it either way.

DipperTheDan: It's fine. I have had a look elsewhere and we've done this same exact assertion with the same kind of wording. Keep it as is which means it'll be consistent.

nuria1110 (approved):

DipperTheDan (approved):

PR #6972, opened 2024-09-19
refactor: convert utils & validations from enzyme to RTL 

tomdavies73:
### Proposed behaviour

1. Renames and re-writes all test and test descriptions of test files which do not contain any JSX.
2. Fully re-writes and converts all test files which do contain JSX from enzyme -> RTL.

### Current behaviour

All test files are either written in enzyme and or use poor practices.

### Checklist

#### QA


src/__internal__/utils/helpers/events/events.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    div.appendChild(ul);
+
+    const path = [div, ul, li];
+    const event = ({ target: li } as unknown) as CustomEvent;
+    const result = Events.composedPath(event);
+
+    expect(result).toEqual(path);
+  });
+
+  test("Builds the path if it is not available on the element from a ReactWrapper", () => {
```
Parsium, line 312: **comment**: I wonder if we will still need our custom version of `composedPath` after Enzyme is removed. It might be worth us investigating this in the future.

What do you think about raising a ticket for this and noting its number in this test?

```suggestion
  /* TODO: <ticket-number> Investigate if `composedPath` is still required post removal of Enzyme.  */
  test("Builds the path if it is not available on the element from a ReactWrapper", () => {
```

src/__internal__/utils/logger/logger.test.ts
```diff
@@ -0,0 +1,22 @@
...
+});
+
+test("should output a warning to the console with a deprecation prefix when logging is enabled", () => {
+  Logger.setEnabledState(true);
+  const assert = expectConsoleOutput(
+    "[Deprecation] This is a deprecation message",
+    "warn"
+  );
+  Logger.deprecate("This is a deprecation message");
+  assert();
```
Parsium, line 21: **suggestion**: what do you think of mocking the console and checking its been called directly?

This would avoid the linter raising the "no assertions" warning.

```suggestion
test("should output a warning to the console with a deprecation prefix when logging is enabled", () => {
  Logger.setEnabledState(true);
  const consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation(() => {});

  Logger.deprecate("This is a deprecation message");
  
  expect(console.warn).toHaveBeenCalledWith("[Deprecation] This is a deprecation message");
  consoleWarnSpy.mockReset();
```

src/__internal__/validations/validation-icon.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+    .spyOn(global.console, "error")
+    .mockImplementation(() => undefined);
+  const errorMessage = `The tooltipFlipOverrides prop supplied to ValidationIcon must be an array containing some or all of ["top", "bottom", "left", "right"].`;
+
+  expect(() => {
+    render(
+      <ValidationIcon
+        isPartOfInput
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore:next-line
```
Parsium, line 186: **suggestion**: what do you think of using [`@ts-expect-error` directive](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check) here? This way we can inform the compiler of the expected type error without having to disable the linter: 

```suggestion
       /* @ts-expect-error testing invalid prop value */
```

src/__internal__/validations/validation-icon.test.tsx
```diff
@@ -0,0 +1,215 @@
...
```
Parsium, line 207: **suggestion**: likewise here :)

```suggestion
       /* @ts-expect-error testing invalid prop value */
```

nuria1110 (commented):
Agree with James' comments, nothing else from me so happy to approve once they've been addressed.

src/__internal__/utils/logger/logger.test.ts
```diff
@@ -0,0 +1,24 @@
...
+});
+
+test("should output a warning to the console with a deprecation prefix when logging is enabled", () => {
+  Logger.setEnabledState(true);
+  const consoleWarnSpy = jest
+    .spyOn(console, "warn")
+    .mockImplementation(() => {});
+  Logger.deprecate("This is a deprecation message");
+
+  expect(console.warn).toHaveBeenCalledWith(
```
Parsium, line 20: **suggestion**: apologies @tomdavies73, I overlooked in my previous suggestion that this would cause a lint warning. What do you think of amending it to this to get around it?

```suggestion
  expect(consoleWarnSpy).toHaveBeenCalledWith(
```

Parsium (approved):

nuria1110 (approved):

PR #6989, opened 2024-09-25
feat(storybook): upgrade storybook to 8.3

nineteen88:
### Proposed behaviour

- update version to 8.3.3

### Current behaviour

- Storybook is currently on 8.2.6

### Checklist

#### QA

### Testing instructions

General smoke and regression testing across Storybook and Chromatic. Shouldn't be as impactful as previous Storybook upgrades with the regressions.


nineteen88:
Added Blocked label as there is some issues with an update to Rollup whereby it now panics when there are unicode characters. We don't use Rollup ourselves, but Playwright utilises it and therefore is crashing when trying to run our playwright tests.

The Rollup update itself is to address an XSS (High) security vulnerability that is currently present in the Master branch of Carbon.

.storybook/preview.ts
```diff
@@ -138,6 +138,7 @@ const preview: Preview = {
   decorators,
   globalTypes,
   loaders,
+  tags: ["autodocs"],
```
Parsium, line 141: **suggestion**: as auto-generated documentation is enabled for all stories by default, what do you think about disabling it for the Welcome Page stories?:

```tsx
// in .storybook/welcome-page/welcome.stories.js
export default {
  title: "Welcome",
  component: Welcome,
  parameters: {
    chromatic: {
      disableSnapshot: true,
    },
    layout: "fullscreen",
  },
  tags: ["!autodocs"], // 👈 disable autogen of docs page
};
```

This would prevent visitors landing on the generated docs page when they visit carbon.sage.com:

Screenshot 2024-09-27 at 17 05 14

src/components/select/simple-select/simple-select.spec.tsx
```diff
@@ -366,10 +366,10 @@ describe("SimpleSelect", () => {
         });
 
         describe("and the focus triggered by mouseDown", () => {
-          it("then that prop should not be called", () => {
+          it("then that prop should not be called more than once", () => {
             simulateSelectTextboxEvent(wrapper, "mousedown");
             simulateSelectTextboxEvent(wrapper, "focus");
-            expect(onOpenFn).not.toHaveBeenCalled();
+            expect(onOpenFn).toHaveBeenCalledTimes(1);
```
DipperTheDan, line 372: **question:** The test description refers to `... not be called` but the assertion has been changed to `toHaveBeenCalled`. Is this correct?

src/components/select/simple-select/simple-select.spec.tsx
```diff
@@ -366,10 +366,10 @@ describe("SimpleSelect", () => {
         });
 
         describe("and the focus triggered by mouseDown", () => {
-          it("then that prop should not be called", () => {
+          it("then that prop should not be called more than once", () => {
```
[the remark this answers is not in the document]
nineteen88: Yes, I'm not sure why it passed in the first place. @edleeks87 and me sat down and had a look at this one together but the test should have been failing as far as we could tell because that wasn't the behaviour that was actually in place. So the test title was updated and the test itself now checks for it be called once

src/components/select/simple-select/simple-select.spec.tsx
nineteen88: Yeah, see the response to James: https://github.com/Sage/carbon/pull/6989/files#r1787414429

.storybook/preview.ts
nineteen88: Sure, good suggestion. I'll do that

package.json
```diff
@@ -67,6 +67,7 @@
     "@babel/preset-typescript": "^7.23.3",
     "@babel/template": "^7.22.15",
     "@babel/types": "^7.23.4",
+    "@chromatic-com/storybook": "^1.9.0",
```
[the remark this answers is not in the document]
nineteen88: This was the version that Storybook included as part of the upgrade to 8.3. I can take a look into if there is any issue with upgrading to 2.x though

Parsium (approved):

DipperTheDan (approved):

paulrobinson88 (approved):

Parsium (approved):

PR #6991, opened 2024-09-26
Select list width

### LINKED ISSUE #6861: Allow the width of a select dropdown (e.g. Filterable) to be wider than the original component.
### Description

When implementing a select component, in this case Filterable - the length of the text for the options in the dropdown often far exceed the width of the control itself. We need to be able to set the dropdown width to be larger than that of the main select component. 

### Suggested solution

Allow the dropdown width to be configurable and unlock this so that the width can exceed that of the parent control. 

### Demo URL

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Confidentiality

edleeks87:
fix #6861

### Proposed behaviour

Removes `left` and `right` options from `listPlacement` type, only `top` and `bottom` remain supported

BREAKING CHANGE: `listPlacement` prop only supports `top` and `bottom` now

Adds support for overriding the list width via `listWidth` prop. Adds support for placing
list at `top-start`, `top-end`, `bottom-start` and `botton-end`. When `top` or `bottom`
is passed to `listPlacement` it will internally append `-end` to it.

image

image

### Current behaviour

`listPlacement` supports `left` and `right`
no support for `listWidth`

### Checklist

#### QA


DipperTheDan (dismissed):

src/components/select/simple-select/simple-select.pw.tsx
```diff
@@ -996,6 +997,93 @@ test.describe("SimpleSelect component", () => {
...
+      page,
+    }) => {
+      await mount(
+        <ListWidth margin={margin} listPlacement={position} listWidth={350} />
+      );
+
+      const positionValue =
+        !position.includes("end") && !position.includes("start")
+          ? `${position}-end`
+          : position;
```
Parsium, line 1024: **suggestion(non-blocking)**: what do you think about defining this value as a parameter for the tests? This would avoid the need for conditional logic within the test:

```tsx
([
  ["top", "top-end", "100px 300px", 150, 2],
  ["bottom", "bottom-end", "0px 300px", 150, 68],
] as const).forEach(([position, expectedPosition, margin, left, top] => {
   // test code...
});
```

src/components/select/simple-select/simple-select.pw.tsx
```diff
@@ -996,6 +997,93 @@ test.describe("SimpleSelect component", () => {
...
+      );
+      await expect(listElement).toBeVisible();
+      await expect(listElement).toHaveCSS("width", "350px");
+      expect(inputWidth).toBe("200px");
+
+      // sub-pixel rendering means this is not always exact
+      expect(listLeftPosition).toBeLessThanOrEqual(left);
+      expect(listTopPosition).toBeLessThanOrEqual(top);
+      expect(listLeftPosition).toBeGreaterThanOrEqual(left - 1);
+      expect(listTopPosition).toBeGreaterThanOrEqual(top - 1);
```
Parsium, line 1051: **suggestion(non-blocking)**: another option we could consider here is using the `toBeCloseTo` assertion, which is designed for handling this kind of comparison. Not essential though - just something that might be helpful:

```suggestion
      expect(listLeftPosition).toBeCloseTo(left);
      expect(listTopPosition).toBeCloseTo(top);
      expect(listLeftPosition).toBeCloseTo(left);
      expect(listTopPosition).toBeCloseTo(top);
```

Parsium (approved):

src/components/select/simple-select/simple-select.pw.tsx
edleeks87: I think I'll leave this for now as there is other examples of the same approach in this file so I'd prefer it be tidied up by the code owners when they come to review the suite as a whole etc

src/components/select/simple-select/simple-select.pw.tsx
edleeks87: Good point I'll update this before I merge

Parsium (dismissed):

Parsium (dismissed):

DipperTheDan (dismissed):

DipperTheDan (approved):

Parsium (approved):

dariusbercea-sage (approved):

PR #7001, opened 2024-10-04
Select virtualiser tweaks

### LINKED ISSUE #6939: Select component does not expand to show all of singular-item multi-line content
### Description

We have multiple locations in our code where a multiple-line message is given: or just that in one language it becomes longer and multi-line. If a Box is used to provide a singular React node component child: the Select option is only the height of a singular line and does not expand to fit the content given: requiring a scrollbar to be used.

When these components were first written (years ago): this worked as expected.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-4wxr6b?file=src%2FApp.tsx

### Steps to reproduce

_No response_

SBS-108085

### Suggested solution

_No response_

### Carbon version

"latest" in stackblitz"

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

### LINKED ISSUE #6978: Select : Unexpected undefined -  Virtualizer - tanstack/react-virtual
### Description

I'm facing this issue randomly when I try to reach our Select lazy loading component (Which use FilterableSelect select component
It’s a known issue, and I believe it was fixed in https://github.com/TanStack/virtual/releases/tag/v3.6.0  
(https://github.com/tailwindlabs/headlessui/pull/3356)

### Reproduction

This issue is reproduced randomly.

### Steps to reproduce

_No response_

_No response_

### Suggested solution

Upgrade  @tanstack/react-virtual package 

### Carbon version

Latest

### Design tokens version

120.1.0

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

Capture d’écran 2024-09-23 à 17 34 30

Capture d’écran 2024-09-23 à 17 34 54

### Confidentiality

edleeks87:
fix #6978
fix #6939

### Proposed behaviour

Upgrades `@tanstack/react-virtual` to `3.10.8`

Updates `Option` interface to allow `value` and `text` to be optional to allow users to compose `children` for complex layouts. This ensures that `SelectList` height is set relatively when a custom option is passed as child so that there is no overflow (scrollbars). When no value is passed the option is treated as non-interactive (can't be clicked or selected) and is styled as such.

### Current behaviour

`@tanstack/react-virtual` is `3.0.0-beta.68`
Option requires `value` and `text` to be set

### Checklist

#### QA

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@floating-ui/react-dom": "~1.3.0",
         "@octokit/rest": "^18.12.0",
         "@styled-system/prop-types": "^5.1.5",
-        "@tanstack/react-virtual": "^3.0.0-beta.68",
+        "@tanstack/react-virtual": "^3.6.0",
         "@types/styled-system": "^5.1.22",
         "chalk": "^4.1.2",
         "ci-info": "^3.8.0",
@@ -6982,11 +6982,11 @@
       }
     },
     "node_modules/@tanstack/react-virtual": {
-      "version": "3.2.0",
-      "resolved": "https://registry.npmjs.org/@tanstack/react-virtual/-/react-virtual-3.2.0.tgz",
-      "integrity": "sha512-OEdMByf2hEfDa6XDbGlZN8qO6bTjlNKqjM3im9JG+u3mCL8jALy0T/67oDI001raUUPh1Bdmfn4ZvPOV5knpcg==",
+      "version": "3.10.8",
+      "resolved": "https://registry.npmjs.org/@tanstack/react-virtual/-/react-virtual-3.10.8.tgz",
+      "integrity": "sha512-VbzbVGSsZlQktyLrP5nxE+vE1ZR+U0NFAWPbJLoG2+DKPwd2D7dVICTVIIaYlJqX1ZCEnYDbaOpmMwbsyhBoIA==",
       "dependencies": {
-        "@tanstack/virtual-core": "3.2.0"
+        "@tanstack/virtual-core": "3.10.8"
       },
       "funding": {
         "type": "github",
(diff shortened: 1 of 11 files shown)


src/components/select/simple-select/simple-select.stories.tsx
```diff
@@ -854,3 +854,52 @@ export const SelectWithDynamicallyAddedOption: Story = () => {
...
+                    Something went wrong
+                  </Typography>
+                </Box>
+                <Typography variant="p" color="errorRed" mb={0}>
+                  We couldn't load the data. Please try again later.
+                </Typography>
+              </Box>
+            </Box>
+          </Box>
+        </Option>
```
Parsium, line 900: **comment**: I have a few thoughts regarding the idea of making `Option` a composable slot - particularly regarding the composition of non-selectable content.

While it’s true that this content wouldn’t be selectable, due to the [custom logic we have in `SelectList`](https://github.com/Sage/carbon/blob/master/src/components/select/__internal__/select-list/select-list.component.tsx#L340-L368), the way it’s rendered inside a `<li>` tag within a `<ul>` list isn't semantically correct:

```html
<!-- Error message is technically an option -->
<ul>
  <li aria-setsize="1" aria-posinset="1">
    <div>Error message</div>
  </li>
</ul>
```

Secondly, even though this content isn’t selectable, the fact that it can’t be navigated to via keyboard could pose accessibility challenges for screen readers, as they won’t be able to access this information.

Thirdly, if we allow non-selectable content to be rendered in the dropdown popup but outside the list, like this:

```html
<!-- Correct rendering for non-selectable content -->
<ul>
</ul>
<div>Error message</div>
```

We would need to make sure that the content can still be accessed by keyboard users, without the list closely prematurely.

To ensure clarity, it might be useful to document that Option isn’t intended for non-option content, while also encouraging consumers to explore alternative UX patterns for handling errors in dynamic Select lists.

edleeks87: I think I'll move the story to a test story for QA and then remove it from public docs but add some text around  how not setting a value on the Option means it's content is non-interactive and that in general it's not recommended etc

src/components/select/option/option.spec.tsx
```diff
@@ -142,3 +176,13 @@ describe("Option", () => {
     });
   });
 });
+
+test("should not render with curesor style pointer when no value or text is passed", () => {
```
nineteen88, line 180: suggestion: typo

```suggestion
test("should not render with cursor style pointer when no value or text is passed", () => {
```

src/components/select/simple-select/simple-select.stories.tsx
Parsium: Yeah, I appreciate we don't want to make too many breaking changes at this stage. I’m happy with making this a test story for QA, and adding a disclaimer note about its intended use for now 👍🏼

Parsium (dismissed):

nineteen88 (dismissed):

Parsium (approved):

Parsium (dismissed):

dariusbercea-sage (dismissed):

dariusbercea-sage (approved):

Parsium (approved):

nineteen88 (approved):

PR #7003, opened 2024-10-08
fix(action-popover): ensure that opening using the up arrow focuses last focusable element - FE-6793

### LINKED ISSUE #6826: ActionPopover - opening with Up Arrow focuses first element rather than last
### Description

From the code in the ActionPopover component, it is apparent that pressing the up arrow key is supposed to open the menu with the last focusable item focused. We have:

- a unit test for this behaviour (which passes, although it does not when attempting to convert it to RTL - I am not sure why the test passes) https://github.com/Sage/carbon/blob/master/src/components/action-popover/action-popover.spec.tsx#L837
- a Playwright test (currently skipped due to presumably-unrelated focus flakiness) https://github.com/Sage/carbon/blob/master/src/components/action-popover/action-popover.pw.tsx#L164
- code within the component itself that is clearly designed to accomplish this (note the `else if` clause that this is inside) https://github.com/Sage/carbon/blob/master/src/components/action-popover/action-popover.component.tsx#L177

Yet not only does an RTL unit test for this behaviour fail, it fails when tested in the browser in our Storybook docs.

### Reproduction

https://carbon.sage.com/?path=/docs/action-popover--docs

### Steps to reproduce

Tab to menu button and open by pressing the up arrow. The first item is focused, rather than the last.

_No response_

### Suggested solution

Confirm with UX/accessibility whether this behaviour (focusing the last element) is actually wanted or not (there is a comment in the code referring to ARIA best practices and referencing https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions/, but that example focuses the first element, not the last, when opened with the up arrow key).

If focus-last-element is correct, then we should fix the bug, and uncomment the commented-out RTL test that I am adding for this.

If focus-first-element is correct, then the existing code and tests which claim to implement this should be removed, and tests added to ensure that the first element is focused instead.

### Carbon version

latest

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

DipperTheDan:
Currently, we have a bug whereby when a user presses the up arrow key, the first focusable item in the menu is focused. This appears to be a bug as we have code that specifically handles this behaviour but is overridden elsewhere in the code. This fix ensures that the last focusable item is focused.

fixes #6826

### Proposed behaviour

- Pressing the `Up Arrow` key should open the menu and focus the last focusable element.

https://github.com/user-attachments/assets/3505ce95-fa05-4d0b-960a-d61fd529141c

### Current behaviour

- Pressing the `Up Arrow` key opens the menu and focuses the first focusable element.

https://github.com/user-attachments/assets/d9fa3e32-da09-44ef-bc81-f9e7f9234d70

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- Ensure that the `Up Arrow` key focuses on the last focusable element in the menu. 
- No other functionality-related regressions regarding keyboard navigation.

### DIFF
--- a/src/components/action-popover/__internal__/action-popover-utils.tsx
+++ b/src/components/action-popover/__internal__/action-popover-utils.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { ActionPopoverItem } from "../action-popover-item/action-popover-item.component";
+
+// Reusable type alias for item types
+type ReactItem = React.ReactChild | React.ReactFragment | React.ReactPortal;
+
+export const getItems = (
+  children: React.ReactNode | React.ReactNode[]
+): ReactItem[] =>
+  React.Children.toArray(children).filter(
+    (child) => React.isValidElement(child) && child.type === ActionPopoverItem
+  );
+
+export const isItemDisabled = (item: ReactItem) =>
+  React.isValidElement(item) && !!item.props?.disabled;
+
+export const findFirstFocusableItem = (items: ReactItem[]): number =>
+  items.findIndex((_, index) => !isItemDisabled(items[index]));
+
+// FIX-ME: FE-6248
+// Once we no longer support Node 16, this function can be removed and `findLastIndex()` can be used in its place.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex
+export const findLastFocusableItem = (items: ReactItem[]): number => {
+  for (let i = items.length - 1; i >= 0; i--) {
+    if (!isItemDisabled(items[i])) {
+      return i;
+    }
+  }
+  return -1;
+};
--- a/src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx
+++ b/src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx
@@ -1,10 +1,4 @@
-import React, {
-  useCallback,
-  useMemo,
-  useContext,
-  useState,
-  useEffect,
-} from "react";
+import React, { useCallback, useMemo, useContext, useState } from "react";
 import invariant from "invariant";
 
 import { Menu } from "../action-popover.style";
@@ -16,6 +10,12 @@ import ActionPopoverDivider from "../action-popover-divider/action-popover-divid
 import ActionPopoverContext, {
   Alignment,
 } from "../__internal__/action-popover.context";
+import {
+  findFirstFocusableItem,
+  findLastFocusableItem,
+  getItems,
+  isItemDisabled,
+} from "../__internal__/action-popover-utils";
 
 export interface ActionPopoverMenuBaseProps {
   /** Children for the menu */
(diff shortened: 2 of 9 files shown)


src/components/action-popover/__internal__/action-popover-utils.tsx
```diff
@@ -0,0 +1,39 @@
...
+type ReactItem = React.ReactChild | React.ReactFragment | React.ReactPortal;
+
+export const getItems = (
+  children: React.ReactNode | React.ReactNode[]
+): ReactItem[] =>
+  React.Children.toArray(children).filter(
+    (child) => React.isValidElement(child) && child.type === ActionPopoverItem
+  );
+
+export const isItemDisabled = (items: ReactItem[]) => (
```
edleeks87, line 14: suggestion: we could simplify this and pass the item we want here instead of relying on currying to pass the items and then an index

```
const isItemDisabled = (item: ReactItem) => React.isValidElement(item) && !!item.props?.disabled

File: src/components/action-popover/action-popover-menu/action-popover-menu.component.tsx [hunk 2]
```diff
@@ -179,7 +149,7 @@ const ActionPopoverMenu = React.forwardRef<
           let indexValue = focusIndex - 1;
           while (
             indexValue >= firstFocusableItem &&
-            isItemDisabled(indexValue)
+            checkItemDisabled(indexValue)
```
edleeks87, line 152: suggestion: if you make the change suggested here you can just call `isItemDisabled(items[indexValue)` or if you prefer to keep the memoisation above you could do:

```
  const checkItemDisabled = useCallback(
      (value) => isItemDisabled(items[value]),
      [items]
    );

File: src/components/action-popover/__internal__/action-popover-utils.tsx [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+  value: number
+): boolean => {
+  const item = items[value];
+  return React.isValidElement(item) && !!item.props?.disabled;
+};
+
+export const findFirstFocusableItem = (
```
edleeks87, line 21: Suggestion: If you make the change above you can remove the second parameter here and have something like:

```
const findFirstFocusableItem = (items: ReactItem[]): number => items.findIndex((_, index) => !isItemDisabled(items[index));

const findLastFocusableItem = (
  items: ReactItem[],
): number => {
  for (let i = items.length - 1; i >= 0; i--) {
    if (!isItemDisabled(items[i])) {
      return i;
    }
  }
  return -1;
};

File: src/components/action-popover/__internal__/action-popover-utils.tsx [hunk 4]
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { ActionPopoverItem } from "../action-popover-item/action-popover-item.component";
+
+// Reusable type alias for item types
+type ReactItem = React.ReactChild | React.ReactFragment | React.ReactPortal;
+
+export const getItems = (
```
edleeks87, line 7: comment (non-blocking): I wonder if this is the chance to decouple the implementation from specific react features like iterating over children and using DOM querying etc. If not at least this will isolate the parts that will need refactoring

DipperTheDan: I'll certainly give it a shot. If it's straight forward, it's a nice thing to get done.

src/components/action-popover/action-popover.test.tsx
```diff
@@ -22,9 +22,8 @@ jest.mock("../../__internal__/utils/helpers/guid");
   () => "guid-12345"
 );
 
-beforeAll(() => {
+beforeEach(() => {
```
edleeks87, line 25: comment: we can just call this beforeAll and reinitialise real timers afterAll etc

src/components/action-popover/action-popover.test.tsx
```diff
@@ -448,6 +447,7 @@ test("clicking the menu button focuses the first focusable element", async () =>
   );
 
   await user.click(screen.getByRole("button"));
+  jest.runAllTimers();
```
edleeks87, line 450: comment: might be worth using pending timers here as I don't think we need to run both macro and micro tasks

edleeks87 (approved):

mihai-albu-sage (approved):

dariusbercea-sage (approved):

PR #7014, opened 2024-10-10
Option comps rtl

edleeks87:
### Proposed behaviour

Converts Option, OptionGroupHeader and OptionRow unit tests to RTL

### Current behaviour

 
Option, OptionGroupHeader and OptionRow unit tests are enzyme

### Checklist

#### QA


src/components/select/option-group-header/option-group-header.test.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import OptionGroupHeader from ".";
+import guid from "../../../__internal__/utils/helpers/guid";
+
+const mockedGuid = "guid-12345";
+jest.mock("../../../__internal__/utils/helpers/guid");
+(guid as jest.MockedFunction<typeof guid>).mockImplementation(() => mockedGuid);
```
Parsium, line 8: **question**: it doesn't appear we are asserting the id set on an underlying element in any of the tests. Is this mock required for something else?

src/components/select/option/option.test.tsx
```diff
@@ -0,0 +1,209 @@
...
+      <ul>
+        <Option text="foo" value="1" />
+      </ul>
+    );
+    const option = screen.getByRole("option", { name: "foo" });
+
+    expect(option).not.toHaveAttribute("aria-disabled");
+  });
+
+  it("should call `onSelect` or `onClick` when the user clicks on the element", async () => {
```
Parsium, line 129: **suggestion(non-blocking)**: what do you think of this so it reflects calling both event handlers?

```suggestion
  it("should call `onSelect` and `onClick` when the user clicks on the element", async () => {
```

edleeks87: I think it works either way here tbh as in the scenario it would call both if it wasn't disabled etc

src/components/select/option-group-header/option-group-header.test.tsx
edleeks87: I'll look into it, just left them as they were in the original files

src/components/select/option/option.test.tsx
```diff
@@ -0,0 +1,209 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import Option from ".";
+import SelectListContext from "../__internal__/select-list/select-list.context";
+import guid from "../../../__internal__/utils/helpers/guid";
+
+const mockedGuid = "guid-12345";
+jest.mock("../../../__internal__/utils/helpers/guid");
+(guid as jest.MockedFunction<typeof guid>).mockImplementation(() => mockedGuid);
```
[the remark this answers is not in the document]
edleeks87: [we are using it here](https://github.com/Sage/carbon/blob/ef7e9d2a2f7287110395ba8cca290a7242dd861c/src/components/select/option/option.test.tsx#L159)

tomdavies73 (approved):

src/components/select/option/option.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+    await user.click(screen.getByRole("option", { name: "foo" }));
+
+    expect(onSelect).toHaveBeenCalledWith({
+      text: "foo",
+      value: "bar",
+      id: mockedGuid,
+    });
+  });
+
+  it("should call `onSelect` when the user clicks on the element and `onClick` is not set but custom `id` is passed", async () => {
```
Parsium, line 189: **suggestion(non-blocking)**: similarly, what do you think about this tweak for clarity?:

```suggestion
  it("calls `onSelect` when clicked, without `onClick` set, but with `id` provided.", async () => {
```

tomdavies73 (approved):

Parsium (approved):

paulrobinson88 (approved):

PR #7037, opened 2024-10-24
refactor(multi-select): rewrite all enzyme tests with React Testing Library (FE-6693)

Parsium:
### Proposed behaviour

- Rewrite all Enzyme tests with React Testing Library (RTL)

### Current behaviour

- The majority of the jest tests for `MultiSelect` are written with Enzyme, which is a legacy package

### Checklist

#### QA


src/components/select/multi-select/multi-select.component.tsx
```diff
@@ -149,27 +149,24 @@ export const MultiSelect = React.forwardRef<HTMLInputElement, MultiSelectProps>(
     const accessibilityLabelId = useRef(guid());
     const containerRef = useRef<HTMLDivElement>(null);
     const listboxRef = useRef<HTMLDivElement>(null);
-    const isInputFocused = useRef(false);
     const isClickTriggeredBySelect = useRef(false);
     const isMouseDownReported = useRef(false);
     const isMouseDownOnInput = useRef(false);
-    const isOpenedByFocus = useRef(false);
+    const [isOpenedByFocus, setIsOpenedByFocus] = useState(false);
```
DipperTheDan, line 155: **question(non-blocking):** What has been the thinking in moving this from `ref` to `state`? Did you discover an issue or something else when writing the new tests?

edleeks87: I think the changes have improved the code, but I'm not convinced we needed to do this at this time as we are just trying to get this refactored to RTL and all the Selects are being rewritten anyway

src/components/select/multi-select/multi-select.test.tsx
```diff
@@ -1,24 +1,45 @@
...
+  // Mock non-zero dimensions for the scrollable container in dropdown list. To ensure react-virtual renders options in the dropdown list correctly.
+  mockDOMRect(40, 100, "select-list-scrollable-container");
+});
 
-test("renders combobox without text overlay", () => {
+afterEach(() => {
+  jest.restoreAllMocks();
+});
+
+test("displays input field", () => {
```
edleeks87, line 29: comment: I'm not sure we get much value from this test as we can imply it's presence/visibility from the other tests that query/assert against it

src/components/select/multi-select/multi-select.test.tsx
```diff
@@ -46,60 +67,1111 @@ test("initially renders combobox with custom placeholder when placeholder prop i
...
+
+    await user.click(screen.getByRole("combobox"));
+    await user.click(screen.getByText("Outside content"));
+
+    await waitFor(() =>
+      expect(screen.queryByRole("listbox")).not.toBeInTheDocument()
+    );
+  });
+
+  it("closes when input is clicked twice", async () => {
```
edleeks87, line 512: suggestion (non-blocking): might be worth doing the same test for the input icon

src/components/select/multi-select/multi-select.test.tsx
```diff
@@ -46,60 +67,1111 @@ test("initially renders combobox with custom placeholder when placeholder prop i
...
+    </>
+  );
+
+  await user.type(screen.getByRole("combobox"), "a");
+  await user.click(screen.getByText("Outside content"));
+
+  expect(screen.getByRole("combobox")).toHaveValue("");
+});
+
+test("clears the input after an option is selected", async () => {
```
edleeks87, line 807: question: won't the input always have an empty value as nothing has been typed here?

src/components/select/multi-select/multi-select.component.tsx
Parsium: I updated this to state after identifying it as a reactive value. It's generally recommended to [avoid writing to refs during rendering](https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs) as it can make the component's behaviour more difficult to reason.

While this change isn't strictly necessary for the RTL refactor, I chose to make it during refactoring as it also helped simplify the component's code, which I decided to do to achieve 100% test coverage. Although modifying a component just to pass tests doesn't seem ideal, I ran into challenges with RTL's user-centric approach. Without these changes, I was concerned about introducing excessive bloat to the tests. Given that a full component rewrite is unlikely soon, I opted to make small, incremental improvements where possible.

edleeks87 (approved):

edleeks87 (approved):

paulrobinson88 (approved):

PR #7080, opened 2024-11-18
feat(date-input, date-range): upgrade react-day-picker to v9

damienrobson-sage:
Upgrades react-day-picker to remove the dependency on React 18

### Checklist

#### QA

### DIFF
--- a/src/components/date/__internal__/date-picker/date-picker.component.tsx
+++ b/src/components/date/__internal__/date-picker/date-picker.component.tsx
@@ -1,37 +1,41 @@
-import React, { useCallback, useEffect, useMemo, useRef } from "react";
-import DayPicker, {
-  DayPickerProps,
-  DayModifiers,
-  Modifier,
-  LocaleUtils,
-} from "react-day-picker";
 import { flip, offset } from "@floating-ui/dom";
+import React, {
+  useCallback,
+  KeyboardEvent,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import {
+  DayPicker,
+  DayPickerProps,
+  defaultLocale,
+  Modifiers,
+} from "react-day-picker";
 
-import { getDisabledDays } from "../utils";
-import Popover from "../../../../__internal__/popover";
 import useLocale from "../../../../hooks/__internal__/useLocale";
+import Popover from "../../../../__internal__/popover";
 import Navbar from "../navbar";
 import Weekday from "../weekday";
-import StyledDayPicker from "./day-picker.style";
-import Events from "../../../../__internal__/utils/helpers/events";
+import { getDisabledDays } from "../utils";
 import { defaultFocusableSelectors } from "../../../../__internal__/focus-trap/focus-trap-utils";
+import Events from "../../../../__internal__/utils/helpers/events";
+
+import StyledDayPicker from "./day-picker.style";
 
 type CustomRefObject<T> = {
   current?: T | null;
 };
 
-/** there is an issue with typescript-to-proptypes package that means we need to override these types */
-interface Modifiers {
-  today: NonNullable<Modifier> | NonNullable<Modifier>[];
-  outside: NonNullable<Modifier> | NonNullable<Modifier>[];
-  [other: string]: NonNullable<Modifier> | NonNullable<Modifier>[];
-}
-
 export interface PickerProps
-  extends Omit<DayPickerProps, "disabledDays" | "modifiers" | "selectedDays"> {
-  disabledDays?: NonNullable<Modifier> | NonNullable<Modifier>[] | undefined[];
+  extends Omit<
+    DayPickerProps,
+    "mode" | "disabledDays" | "modifiers" | "selectedDays"
+  > {
+  disabledDays?: NonNullable<Date> | NonNullable<Date>[] | undefined[];
   modifiers?: Partial<Modifiers>;
-  selectedDays?: NonNullable<Modifier> | NonNullable<Modifier>[] | undefined[];
+  selectedDays?: NonNullable<Date> | NonNullable<Date>[] | undefined[];
 }
 
 export interface DatePickerProps {
@@ -46,7 +50,7 @@ export interface DatePickerProps {
   /** Element that the DatePicker will be displayed under */
   inputElement: CustomRefObject<HTMLElement>;
   /** Currently selected date */
-  selectedDays?: Date;
+  selectedDays?: Date | undefined;
   /** Callback to handle mousedown event on picker container */
   pickerMouseDown?: () => void;
   /** Sets whether the picker should be displayed */
@@ -82,6 +86,9 @@ export const DatePicker = ({
   pickerTabGuardId,
   onPickerClose,
 }: DatePickerProps) => {
+  const [focusedMonth, setFocusedMonth] = useState<Date | undefined>(
+    selectedDays || new Date(),
+  );
   const locale = useLocale();
   const { localize, options } = locale.date.dateFnsLocale();
   const { weekStartsOn } = options || /* istanbul ignore next */ {};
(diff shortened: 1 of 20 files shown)


src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -1,37 +1,44 @@
...
 import Weekday from "../weekday";
-import StyledDayPicker from "./day-picker.style";
-import Events from "../../../../__internal__/utils/helpers/events";
+import { getDisabledDays } from "../utils";
 import { defaultFocusableSelectors } from "../../../../__internal__/focus-trap/focus-trap-utils";
+import Events from "../../../../__internal__/utils/helpers/events";
+
+import StyledDayPicker from "./day-picker.style";
+
+import "react-day-picker/style.css";
```
edleeks87, line 27: question (non-blocking): do we get anything that we need from this? I tried removing it and couldn't see any regressions but I could be missing something. Happy for it to be non-blocking as it's not a big deal including it and the jest config update etc

jest.config.ts
```diff
@@ -30,6 +31,9 @@ const config: Config = {
     "^.+\\.(js|mjs|jsx|ts|tsx)$": "babel-jest",
     "^.+\\.svg$": "<rootDir>/svgTransform.mjs",
   },
+  moduleNameMapper: {
```
edleeks87, line 34: comment: if we don't need to import the day picker style sheet this can be removed

src/components/date/date.component.tsx
```diff
@@ -125,7 +126,7 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
       onClick,
       onFocus,
       onKeyDown,
-      pickerProps = {},
+      pickerProps = { mode: "single" },
```
edleeks87, line 129: comment: if setting this to anything but `single` will break the component layout etc, I think rather than setting this as a default we could apply it at the point we pass the props to the DayPicker in a way that prevents it being overridden etc (we should `Omit` it from the pickerProps type as well imo)

in date-picker.component:
```
<DayPicker
   ...
  {...pickerProps}
  mode="single"

File: src/components/date/date.component.tsx [hunk 4]
```diff
@@ -155,11 +156,19 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
     );
     const { inputRefMap, setInputRefMap } = useContext(DateRangeContext);
     const [open, setOpen] = useState(false);
-    const [selectedDays, setSelectedDays] = useState(
-      checkISOFormatAndLength(value)
+    const [selectedDays, setSelectedDays] = useState(() => {
+      const parsedDate = parse(value, "P", new Date(), {
```
edleeks87, line 160: Comment: this potentially confusing as we already have a util called `parseDate` imported in the file. Maybe it would be worth using, or updating if needed, that function so we're consistent. We should also be able to use `isDateValid` from the same util file

src/components/date/date.stories.tsx
```diff
@@ -124,6 +124,26 @@ export const Empty: Story = () => {
 };
 Empty.storyName = "Empty";
 
+export const DisabledDates: Story = () => {
```
edleeks87, line 127: Praise: a nice addition 👏

src/components/tabs/__internal__/tab-title/tab-title.style.ts
edleeks87, line 1: comment: I think this file has been included by mistake

edleeks87 (commented):
Really good work getting this update done so fast @damienrobson-sage

src/components/date/__internal__/date-picker/day-picker.style.ts
```diff
@@ -1,190 +1,322 @@
...
 
-  .DayPicker-NavButton {
-    position: absolute;
-    width: 1.5rem;
-    height: 1.5rem;
-    background-repeat: no-repeat;
-    background-position: center;
-    background-size: contain;
+  .rdp-day {
+    width: var(--rdp-day-width);
```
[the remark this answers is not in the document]
edleeks87: The styling on lines 97 and 98 also need updating with these tokens as well I think

damienrobson-sage:
Playwright tests pass locally but seem a little flaky when running in the pipeline.
Screenshot 2024-11-20 at 08 25 18

src/components/date/date.pw.tsx
```diff
@@ -130,15 +131,14 @@ test.describe("Functionality tests", () => {
...
 
-    const dayPicker = page
-      .getByRole("row")
-      .locator(`div[aria-label="${TODAY}"]`);
-    await expect(dayPicker).toHaveAttribute("aria-label", TODAY);
-    await expect(dayPicker).toHaveClass(dayClass);
+    const dayPicker = page.locator(`td[data-today="true"]`).getByRole("button");
+    await expect(dayPicker).toHaveAttribute("aria-label", `Today, ${TODAY}`);
+    const inputParent = page.locator(`td[data-today="true"]`);
+    await expect(inputParent).toHaveClass(dayClass);
```
Parsium, line 141: **suggestion(non-blocking)**: I noticed the React Day Picker's docs doesn't seem to cover the set `data-*` attributes. To make our tests more robust and less reilant on undocumented attributes, what do you think about using an approach like this?:

```suggestion
    const todayButton = page.getByRole("button", { name: `Today, ${TODAY}`});
    const todayCell = page.getByRole("gridcell").filter({
      has: todayButton
    });
    
    await expect(todayButton).toBeVisible();
    await expect(todayCell).toHaveClass(dayClass);
```

This approach avoids directly depending on the `data-*` tags, instead locating elements in a way similar to how assitive tech would interpret the page.

edleeks87 (approved):

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -223,66 +191,108 @@ export const DatePicker = ({
...
+              localize: {
+                ...defaultLocale.localize,
+                months: monthsLong,
+                weekdaysLong,
+                weekdaysShort,
+              },
+            }}
+            selected={focusedMonth}
+            month={focusedMonth}
+            onDayClick={(e) => {
```
Parsium, line 252: **suggestion**: I noticed when the date is cleared in the input box, the picker doesn't navigate to the current month as expected. What do you think about adding a default value for `month` to handle this case?:

```suggestion
            month={focusedMonth || new Date()}
```

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -82,6 +87,13 @@ export const DatePicker = ({
   pickerTabGuardId,
   onPickerClose,
 }: DatePickerProps) => {
+  Logger.warn(
+    "Both minDate and maxDate must be provided to DatePicker component in order to set disabled dates.",
+  );
```
Parsium, line 92: **suggestion**: for the warning message, what do you think about this tweak? By introducing a flag like `minMaxDateWarningLogged`, the warning would only appear once when the condition is met.

This could make the logging more consumer-friendly and prevent repeated warnings, especially in case the component frequently re-renders:

```tsx
let minMaxDateWarningLogged = false;

const DatePicker = () => {
  if (!!minDate !== !!maxDate) {
    Logger.warn(
      "Both minDate and maxDate must be provided to DatePicker component in order to set disabled dates."
    );
    minMaxDateWarningLogged = true;
  }

  // component code...
};
```

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -82,6 +86,15 @@ export const DatePicker = ({
   pickerTabGuardId,
   onPickerClose,
 }: DatePickerProps) => {
+  if ((minDate && !maxDate) || (!minDate && maxDate)) {
+    console.warn(
+      "Both minDate and maxDate must be provided to DatePicker component in order to set disabled dates.",
+    );
+  }
```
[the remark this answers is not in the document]
edleeks87: If we add an invariant it's a breaking change, I'm not really sure why we need it as both work independently of each other like they did before the upgrade

Parsium (approved):

edleeks87 (approved):

Parsium (dismissed):

Parsium (approved):

edleeks87 (approved):

harpalsingh:
while there is some color/spacing adjustments, looks fine to me and I know there are designs are date input grades in the future anyway, so this is good.

dariusbercea-sage (approved):

Parsium (approved):

edleeks87 (approved):

dariusbercea-sage (approved):

PR #7083, opened 2024-11-20
Remove cross-env and replace cross platform environment reading with dotenv

edleeks87:
### Proposed behaviour

Removes `cross-env`
Moves `dotenv` to devDep and upgrades to latest
Adds `.env` file, adds environment variables and adds script files for each script previously relying on `cross-env` and those variables.
Updates packages that rely on `cross-spawn` that has a security vulnerability flagged against it, can't fully address due to `semantic-release`

### Current behaviour

Relies on `cross-env` to handle support for passing cross-platform environment variables

### Checklist

#### QA

### Testing instructions

Worth testing in windows (has been tested already but better to be safe)

- pull branch down
- `npm install` to update `node-modules`
- Run the following npm scripts to test they work in windows:
  - `npm run build-storybook`
  - `npm run start:debug-theme`
  - `npm run start-storybook`


.storybook/withThemeProvider.js
```diff
@@ -12,15 +12,15 @@ const themes = [noTheme, sageTheme].reduce((themesObject, theme) => {
   return themesObject;
 }, {});
 
-if (process.env.STORYBOOK_DEBUG_THEME === "true") {
+if (!isChromatic() && process.env.STORYBOOK_DEBUG_THEME === "true") {
```
edleeks87, line 15: Due to placing the variables in `.env` I've had to add this to prevent it including the debug theme in the comparisons in chromatic. I assume it's because it uses it's own `build-storybook` script instead of ours and pulls in both variables in `.env`

scripts/build-storybook.js
```diff
@@ -0,0 +1,14 @@
+const dotenv = require("dotenv");
+const fs = require("fs");
+const { execSync } = require("child_process");
+
+const envConfig = dotenv.parse(fs.readFileSync(".env"));
```
damienrobson-sage, line 5: **suggestion**: I don't think you need `fs` here; you can just do `require('dotenv').config()` to load everything in from the `.env` file ([docs](https://www.npmjs.com/package/dotenv#%EF%B8%8F-usage))

scripts/debug-storybook.js
```diff
@@ -0,0 +1,14 @@
+const dotenv = require("dotenv");
+const fs = require("fs");
+const { execSync } = require("child_process");
+
+const envConfig = dotenv.parse(fs.readFileSync(".env"));
```
damienrobson-sage, line 5: **suggestion**: as above

scripts/start-storybook.js
```diff
@@ -0,0 +1,14 @@
+const dotenv = require("dotenv");
+const fs = require("fs");
+const { execSync } = require("child_process");
+
+const envConfig = dotenv.parse(fs.readFileSync(".env"));
```
damienrobson-sage, line 5: **suggestion**: as above

scripts/build-storybook.js
edleeks87: We don't want to load everything from the env file though

damienrobson-sage (approved):

scripts/build-storybook.js
Parsium, line 1: **suggestion**: to simplify handling variables for different scenarios, what do you think about using multiple environment files?

This approach would also centralise the configuration settings, eliminating the need for embedding config logic in scripts and keeping the settings organised in one place:
```yaml
# .env.production
NODE_ENV=production
STORYBOOK_BUILD=true
STORYBOOK_DEBUG_THEME=false

#.env.development
NODE_ENV=development
STORYBOOK_BUILD=false
STORYBOOK_DEBUG_THEME=false

# .env.storybook-debug-theme
# Separate for easier removal in the future
NODE_ENV=development
STORYBOOK_BUILD=false
STORYBOOK_DEBUG_THEME=true
```

You could then preload these environment variables in the commands. For easier management of multiple environments, the [creator of `dotenv` recommends their newer package, `dotenvx`](https://www.npmjs.com/package/dotenv#preload). Here’s how the scripts might look:

```json
{
  "scripts": {
    "build-storybook": "dotenvx run --env-file=.env.production -- storybook build -c .storybook",
    "start-storybook": "dotenvx run --env-file=.env.production -- storybook dev -c .storybook",
    "start:debug-theme": "dotenvx run --env-file=.env.storybook-debug-theme -- npm run start"
  }
}
```

Parsium (commented):
Thanks for

Parsium:
**suggestion**: I noticed that we set environment variables in multiple places, such as in our GitHub Actions workflows and the build script (`scripts/build.js`). If we’re aiming to centralise our configuration in .env files, it might be worth including the variables from these scripts for consistency and completeness. What are your thoughts?

edleeks87:
> **suggestion**: I noticed that we set environment variables in other places, such as in our GitHub Actions workflows ([linked example](https://github.com/Sage/carbon/blob/master/.github/workflows/semantic-release.yml#L59)) and the build script ([`scripts/build.js`](https://github.com/Sage/carbon/blob/master/scripts/build.js#L8-L11)).
> 
> If we’re aiming to centralise our configuration in .env files, it might be a good idea to include the variables from these scripts as well, for consistency and completeness. What are your thoughts?

Cheers James, I've raised FE-7004 so we can take our time investigating/implementing your suggestion

Parsium (approved):

dariusbercea-sage (approved):

PR #7091, opened 2024-11-25
feat(labels): updates alignment prop usage so it works regardless of the inline prop

robkilby:
### Proposed behaviour

Allows labels to be aligned regardless of whether the inline prop is set or not

### Current behaviour

Labels currently require the inline prop to be enabled (which is deprecated and doesn't work with the new validation system) for the align prop to take effect.

### Checklist

#### QA


src/__internal__/label/label.component.tsx
```diff
@@ -62,7 +62,7 @@ const tooltipPosition = ({
 };
 
 export const Label = ({
-  align = "right",
+  align = "left",
```
edleeks87, line 65: comment: this is causing some of the baselines to update in [chromatic](https://www.chromatic.com/test?appId=5ecf782fe724630022d27d7d&id=67448128300e429aa2d30e59) so I think we need to be a little more nuanced here. We may need to calculate the value of the prop inside the component as opposed to in the prop destructuring

```
// label.component
export const Label = ({
  align,
  ...
}) => {
  const alignment = inline && !align ? "right" : align;
 ...
 <StyledLabelContainer
      data-role="label-container"
      align={alignment}
      inline={inline}
      width={width}
      optional={optional}
      pr={pr}
      pl={pl}
      className={className}
    >

```

```
// label.style
${({ align }) => css`
   justify-content: ${align !== "right" ? "flex-start" : "flex-end"};
`}

Comment (edleeks87) [reply src/__internal__/label/label.component.tsx:65]: Some of the baselines are still going to flag changes, for example I notice in numeral-date we are setting labelAlign="right", we should delete them for the `Required` and `isOptional` stories as it doesn't make sense that they're being set

File: src/__internal__/label/label.style.ts [hunk 2]
```diff
@@ -56,7 +56,11 @@ export const StyledLabelContainer = styled.div<StyledLabelContainerProps>`
   align-items: center;
   margin-bottom: 8px;
 
-  ${({ align, inline, pr, pl, width }) =>
+  ${({ align }) => css`
+    justify-content: ${align === "right" ? "flex-end" : "flex-start"};
```
edleeks87, line 60: question: @harpalsingh I notice when we set `flex-end` it flexes the whole width of the window/parent container, should we ensure it doesn't go past the width of the input or are you happy for us to push responsibility for that to implementing projects?

image

src/__internal__/label/label.component.tsx
robkilby: Ok, sounds good!

src/__internal__/label/label.style.ts
harpalsingh: Hey, for me it shouldn't go past the width of the input, otherwise the alignment would be off anyway? If it can be done inside the component update then it ensures people can accidentally make that error? If we push that responsibility, how would they handle that? if its simple enough then maybe provide instruction/demo in storybook for that.

src/components/textbox/textbox.component.tsx
```diff
@@ -82,7 +82,7 @@ export interface CommonTextboxProps
   leftChildren?: React.ReactNode;
   /** Label content */
   label?: string;
-  /** [Legacy] Inline label alignment */
+  /** Label alignment */
```
tomdavies73, line 85: **suggestion**: If these changes are made, they should also be applied to `Textarea` to ensure consistency

src/__internal__/label/label.component.tsx
tomdavies73: Yeah, I agree here, this should reduce the changeset + the changes you've made to our pw tests

src/__internal__/label/label.style.ts
tomdavies73: Consumers can control the width of the `StyledLabelContainer` with the `labelWidth` prop, so we could document its use to ensure the proper behaviour is achieved like so.

Screenshot 2024-11-26 at 11 17 26

However we would be left with an issue where if not explicity set we could see some incorrect behavioiur like in the screenshot posted by @edleeks87. I suggest we ensure the width of the label does not exceed that of the below inputs.

A pure css solution would be preferable 👍, but if not possible we may have to find a js solution.

edleeks87: Thanks @harpalsingh, I think if we set `max-width` on the `FormField` we can fix the issue , but we'll need to do it conditionally based on the label props passed to the input components (there also may be a better way)
[example of changes](https://github.com/Sage/carbon/tree/demo-max-width-change)

src/components/textbox/textbox.component.tsx
robkilby: Fixed in 6af9fda57cb95af3f8d40efc06fba858f9550015

edleeks87 (approved):
LGTM 👍 

@harpalsingh fyi I've discussed with Rob and we're going to push the work to add this to the `NumeralDate` and `Time` components into a new ticket; they both use fieldset wrappers and legends instead of labels so will need slightly different approach than the Textbox based ones. As there not needed as urgently they can be addressed at a later date etc.

FE-7029 has been raised for that work

tomdavies73 (approved):

harpalsingh (approved):

stephenogorman (approved):

PR #7094, opened 2024-11-26
chore: build script no longer uses cross-env and args can be passed t…

edleeks87:
…o build-storybook script

### Proposed behaviour

Removes `cross-env` call in `build` script as environment variables are being set inside the function's scope so should work avoid cross-platform issues without the need for cross-env or dot-env etc.

Ensures `build-storybook` scrip can receive `args`.
 
### Current behaviour

The `build` script still calls `cross-env`.
The `build-storybook` cannot receive any args passed when it is called.

### Checklist

#### QA

### Testing instructions

Run `npm run build` on a windows machine to test it works cross-platform after `cross-env` removed

### DIFF
--- a/scripts/build-storybook.js
+++ b/scripts/build-storybook.js
@@ -2,11 +2,41 @@ const dotenv = require("dotenv");
 const fs = require("fs");
 const { execSync } = require("child_process");
 
+const whitelistArgs = [
+  "-h",
+  "--help",
+  "-V",
+  "--version",
+  "-o",
+  "--output-dir",
+  "-c",
+  "--config-dir",
+  "--loglevel",
+  "--quiet",
+  "--debug",
+  "--debug-webpack",
+  "--stats-json",
+  "--docs",
+  "--test",
+  "--ci",
+  "--smoke-test",
+  "--preview-url",
+  "--force-build-preview",
+  "--disable-telemetry",
+  "--enable-crash-reports",
+  "--webpack-stats-json",
+];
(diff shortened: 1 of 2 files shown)


damienrobson-sage (dismissed):
LGTM 👍

edleeks87:
Discussed with @Parsium on slack about whitelisting the args we want to pass so I'm going to add that now

edleeks87:
Scripts have been tested by @stephenogorman on windows so I'm removing the QA flag

scripts/build-storybook.js
```diff
@@ -2,11 +2,41 @@ const dotenv = require("dotenv");
...
+  "--force-build-preview",
+  "--disable-telemetry",
+  "--enable-crash-reports",
+  "--webpack-stats-json",
+];
 const envConfig = dotenv.parse(fs.readFileSync(".env"));
 process.env.STORYBOOK_BUILD = envConfig.STORYBOOK_BUILD;
+const args = process.argv.slice(2);
+const storybookArgs = args
+  .filter((arg) => whitelistArgs.includes(arg.split("=")[0]))
```
Parsium, line 33: **suggestion**: what do you think about this tweak? This would help ensure that non key-value arguments like `--ci` are not filtered out:

```suggestion
  .filter((arg) => {
      const [key] = arg.split("=");
      return whiteListArgs.includes(key) || whiteListArgs.includes(arg)
  })
```

Parsium (approved):

PR #7110, opened 2024-12-03
fix(box): css style properties passed as HTML attributes

mihai-albu-sage:
### Proposed behaviour

The `height`, `width`, `opacity` and `color` are not being passed directly to the `StyledBox` component but instead are placed in the `cssProps` object, and then they are used inside the `StyledBox`. This prevents the props from being applied to the HTML node directly.

Playwright tests that were checking for the attribute being present on the node have been amended since the attribute is no longer present. 

after

### Current behaviour

The `height`, `width`, `opacity` and `color`  props are passed directly to the `StyledBox` component which applies them directly to the HTML node.

before

### Checklist

#### QA

### DIFF
--- a/src/components/box/box.test.tsx
+++ b/src/components/box/box.test.tsx
@@ -76,6 +76,46 @@ it("applies the boxShadow styling correctly when a design token is passed in", (
   expect(box).toHaveStyle(`box-shadow: var(--boxShadow100)`);
 });
 
+it("applies the correct styling from the cssProps", () => {
+  render(
+    <Box
+      width="100px"
+      height="100px"
+      opacity="25%"
+      color="yellow"
+      data-role="box"
+    />,
+  );
+
+  const box = screen.getByTestId("box");
+  expect(box).toHaveStyle({
+    width: "100px",
+    height: "100px",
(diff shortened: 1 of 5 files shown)


damienrobson-sage (approved):

src/components/box/box.test.tsx
```diff
@@ -76,6 +76,40 @@ it("applies the boxShadow styling correctly when a design token is passed in", (
...
+      width="100px"
+      height="100px"
+      opacity="25%"
+      color="yellow"
+      data-role="box"
+    />,
+  );
+
+  const box = screen.getByTestId("box");
+  expect(box).toHaveStyle(`width: 100px`);
```
DipperTheDan, line 91: **suggestion(non-blocking):** `toHaveStyle` accepts an object with multiple values, so you could pass these in as one assertion. I'm OK with the way you've done it, though. Feel free to ignore this comment if you prefer your way 😄

DipperTheDan (approved):

damienrobson-sage (approved):

dariusbercea-sage (approved):

PR #7113, opened 2024-12-04
feat(classname prop): remove className from public props

damienrobson-sage:
The following properties have been removed from the listed components. Ensure any components
in the list below do not contain the associated properties.

BREAKING CHANGE:
- The `className` property has been removed from all components.
- Accordion: `scheme` and `buttonHeading`
- Box: `tabIndex`
- ButtonToggle: `grouped`
- Decimal: `onKeyPress`
- Icon: 'extra-small' variant
- Portrait: `gravatar`
- SimpleColorPicker: `isBlurBlocked`
- ListItem: `variant` (now inherits from parent List component)

### Current behaviour

Some of our components still support className as a prop, which would allow consumers to override the component's CSS styling, which is strongly discouraged for maintenance reasons.

### Checklist

#### QA

### Testing instructions

Ensure that all documentation is free of reference to `className`, ensure that passing custom class names to affected components does not change their styling.

### DIFF
--- a/src/components/accordion/accordion.style.ts
+++ b/src/components/accordion/accordion.style.ts
@@ -1,4 +1,5 @@
 import styled, { css } from "styled-components";
+
 import { space, margin } from "styled-system";
 
 import Icon from "../icon";
@@ -13,10 +14,6 @@ const StyledAccordionGroup = styled.div`
 export interface StyledAccordionContainerProps {
   /** Toggles left and right borders, set to none when variant is subtle */
   borders?: "default" | "full" | "none";
-  /** (Deprecated) Renders the accordion heading in the style of a tertiary button */
-  buttonHeading?: boolean;
-  /** (Deprecated) Sets background as white or transparent, set to transparent when variant is subtle */
-  scheme?: "white" | "transparent";
   /** Sets accordion width */
   width?: string;
   /** Sets accordion variant */
@@ -26,15 +23,18 @@ export interface StyledAccordionContainerProps {
 const StyledAccordionContainer = styled.div<StyledAccordionContainerProps>`
   ${space}
   display: flex;
-  align-items: ${({ buttonHeading, variant }) =>
-    (!buttonHeading || variant !== "subtle") && "stretch"};
+  ${({ variant }) =>
+    variant &&
+    css`
+      align-items: stretch;
+    `};
   justify-content: center;
   flex-direction: column;
   box-sizing: border-box;
   width: ${({ width }) => width || "100%"};
   color: var(--colorsUtilityYin090);
-  background-color: ${({ scheme, variant }) =>
-    scheme === "white" && variant !== "subtle"
+  background-color: ${({ variant }) =>
+    variant !== "subtle"
       ? "var(--colorsUtilityYang100)"
       : "var(--colorsUtilityMajorTransparent)"};
   border: 1px solid var(--colorsUtilityMajor100);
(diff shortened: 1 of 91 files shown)


edleeks87 (commented):
This commit will need to be a breaking change as we are removing the prop from the public interface of these components etc

src/components/button-toggle/button-toggle-group/button-toggle-group.component.tsx
```diff
@@ -87,6 +91,13 @@ const ButtonToggleGroup = ({
   className,
   ...props
 }: ButtonToggleGroupProps) => {
+  if (!deprecatedClassNameWarningShown && className) {
```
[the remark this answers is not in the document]
damienrobson-sage: Happy to implement this approach if you'd prefer it - totally agree that it gives users more warning about the impending change.

damienrobson-sage:
To be merged on 31st January 2025

src/components/inline-inputs/inline-inputs.pw.tsx
```diff
@@ -11,16 +11,12 @@ import {
 import { Default as InlineInputComponent } from "./inline-inputs-test.stories";
 import {
   inlineInputContainer,
-  inlineInputsPreview,
```
nuria1110, line 14: **suggestion (non-blocking):** If this is no longer needed we could also remove it from the playwright `locator.ts` file if it is not being used elsewhere

src/components/tile-select/tile-select.pw.tsx
```diff
@@ -26,7 +23,6 @@ import {
   WithCustomSpacing,
 } from "./components.test-pw";
 import {
-  tileSelectDataComponent,
```
nuria1110, line 29: Same here

src/components/vertical-menu/vertical-menu-item/vertical-menu-item.test.tsx
```diff
@@ -7,10 +7,6 @@ import sageTheme from "../../../style/themes/sage";
 import { testStyledSystemPadding } from "../../../__spec_helper__/__internal__/test-utils";
 import Icon from "../../icon";
 import { VerticalMenuItem, VerticalMenuFullScreen } from "..";
-import Logger from "../../../__internal__/utils/logger";
-
-// mock Logger.deprecate so that no console warnings occur while running the tests
-const loggerSpy = jest.spyOn(Logger, "deprecate");
```
nuria1110, line 13: **praise:** Nice tidying up 👏

nuria1110 (commented):
Some really nice cleaning up of the codebase, glad to be removing some of these older props!

src/components/inline-inputs/inline-inputs.pw.tsx
damienrobson-sage: Seems it is removed elsewhere so I've left it alone for now 😄

src/components/tile-select/tile-select.pw.tsx
damienrobson-sage: Same with this one, used in a few places beyond this test

src/components/accordion/accordion.style.ts
```diff
@@ -26,15 +23,15 @@ export interface StyledAccordionContainerProps {
 const StyledAccordionContainer = styled.div<StyledAccordionContainerProps>`
   ${space}
   display: flex;
-  align-items: ${({ buttonHeading, variant }) =>
-    (!buttonHeading || variant !== "subtle") && "stretch"};
+  align-items: ${({ variant }) =>
```
tomdavies73, line 26: **suggestion** (non-blocking): instead of a ternary here, you could do a truthy assertion against variant, if variant is true then add `align-items: stretch`. This might be better than applying the css attribute regardless

nuria1110 (dismissed):

tomdavies73 (approved):

nuria1110 (approved):

paulrobinson88 (approved):

tomdavies73 (approved):

nuria1110 (approved):

PR #7124, opened 2024-12-11
feat(typography): allow listitem to inherit variant from parent list

### LINKED ISSUE #5413: ListItem to inherit font styles from List
### Desired behaviour

`ListItem` inherits font styles from `List`. To apply a font variant to all list items, the variant prop may be set on `List` alone.

```javascript
<List variant="big">
	<ListItem>item 1</ListItem>
	<ListItem>item 2</ListItem>
	<ListItem>item 3</ListItem>
</List>
```

### Current behaviour

Currently, `ListItem` does not inherit font styles from `List`. To apply a font variant to all list items, each list item must have the variant set.

```javascript
<List>
	<ListItem variant="big">item 1</ListItem>
	<ListItem variant="big">item 2</ListItem>
	<ListItem variant="big">item 3</ListItem>
</List>
```

### Suggested Solution

_No response_

https://codesandbox.io/s/list-9tee68

_No response_

### Confidentiality

damienrobson-sage:
In order to reduce replication of code, ListItem now receives information on the variant of the parent List item via context; setting variant on List will pass it to child ListItem components.

Resolves #5413

### Proposed behaviour

`ListItem` now inherits `variant` directly from it's parent `List` component

### Current behaviour

Customers have to manually define the `variant` on each `ListItem`

### Checklist

#### QA

### Testing instructions

Use Storybook to view the new approach/component structure. Additionally, verify that a deprecation warning is thrown if `variant` is added to any `ListItem`


src/components/typography/typography.mdx
```diff
@@ -29,15 +29,27 @@ It's possible to change the text colour and background, using the color props. S
 
 It's possible to change the text position, using the spacing props. See the prop tables below for more information.
 
-It's also possible to create ordered and unordered lists using the `List`, `ListItem` helpers. These components are syntactic sugar around the `Typography` component and accept the same props.
-
 <Canvas of={TypographyStories.VariantsStory} />
 
+## Lists
```
Parsium, line 34: **praise**: great idea to make the list functionality more explicit in the docs 👍🏼

src/components/typography/list.component.tsx
Parsium, line 1: **suggestion(non-blocking)**: for greater visibility of the deprecation notice, what do you think about marking `ListItem`'s `variant` prop as deprecation in its description too? This would allow the notice to be visible in our docs and via IntelliSense

```tsx
export interface ListItemProps extends TypographyProps {
  /** (Deprecated) The visual style to apply to the component */
  variant?: VariantTypes;
}
```

damienrobson-sage: Good shout 👍

Parsium (approved):

nuria1110 (commented):
Looks good to me, my only comment would be that I don't believe this is a breaking change just yet if we are only adding a deprecation warning to `ListItem` and the prop is still available.

nuria1110 (approved):

paulrobinson88 (approved):

PR #7132, opened 2024-12-16
React 18

Parsium:
### Proposed behaviour

- Introduce compatibility with React v18
- Upgrade `react-dnd` and `react-dnd-backend-html5` to v16 - this is required for React v18 support
- Upgrade Storybook to v8.4.7

### Current behaviour

- Carbon only supports React v17

### Checklist

#### QA


quinnturner:
**This is exciting 🥳**

Question: is the hope that the `peerDependency` will be set to `^18 || ^19`? There are a lot of goodies in React 19 (I am most excited about the compiler!), and the lift seems to be much less significant. Would it be too much also to support `^19`? 😄

Parsium:
> **This is exciting 🥳**
> 
> Question: is the hope that the `peerDependency` will be set to `^17 || ^18 || ^19`? ~I am surprised that this PR suggests that `^17` is still supported, given that peer dependencies like DnD are only compatible with specific react versions.~ There are a lot of goodies in React 19 (I am most excited about the compiler!), and the lift seems to be much less significant from 18 -> 19 than 17 -> 18. Would it be too much also to support `^19` in a follow-up release? 😄
> 
> EDIT: I think I am wrong. `react-dnd` is not listed as a peer dependency currently (maybe it should be?), and the consumers of this package can specify the required version they want. Carbon can probably technically support both, which is great news.

Hi @quinnturner 👋🏼

At the moment, we're focused on React 18 support due to team capacity constraints. We've reviewed some of the breaking changes introduced in React 19 and identified areas of tech debt that need attention before we can fully support it. Key challenges include:
- **Prop type support** - Carbon generates prop types for legacy projects that don't utilise TypeScript. We'll need to evaluate the impact of removing this or consider an interim solution for storing type information.
- **Removal of `defaultProps` in all functional components** - This change will require migrating away from styled-components v4, which is already part of our roadmap.

That said, we're eager to support React 19 soon, especially since we're currently two major versions behind. As part of that effort, we're currently experimenting with a replacement for `react-dnd`. After releasing support for React 18, we'll be focusing on major refactoring efforts to tackle our remaining tech debt, including migrating away from `styled-components@v4`.

If you have any follow-up questions, feel free to reach out to us in our #carbon slack channel 😄

Issue #7064, opened 2024-11-08
Filterable Select with object values


gusch:
### Description

I have linked a FilterableSelect to react form.
The particularity here is that the value of the options is an object.
Filtering is working but ...
If you select “Black”, then put the cursor at the end of the text, you won’t be able to remove the letters with backspace.
If values are string, it is working fine.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-aa9gdv?file=package.json,src%2FApp.tsx

### Carbon version

144.2.2

### Design tokens version

4.22.0

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

We have the same problem in CM using Formik. So it seems not to be a react form issue.

What i saw is that for each backspace in the “string version”, i have a onChange then a onFitlerChange.
But i only have the onChange for the “object version”.

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

nuria1110:
FE-6901

edleeks87:
We're keen to address this as part of the RFC and rewrite work we have planned for the Select components. We suspect it will require some rewriting of code which seems redundant when the whole component is soon to be rewritten etc. In the mean time we think a [workaround is available](https://stackblitz.com/edit/parsium-carbon-starter-uwzx8a?file=package.json,src%2FApp.tsx), if this is not the case for you please let us know and we'll increase the priority of this.

@nicktitchmarsh fyi

PR #7136, opened 2024-12-19
fix(filterable-select): cannot backspace with object as values

mihai-albu-sage:
fixes: #7064

### Proposed behaviour
When using objects as values, updating the state through the function received on the `onChange` prop triggers a re-render, which subsequently runs the `useEffect` containing `setMatchText`, even when it shouldn't be run. This issue does not occur when using strings as values.

The difference lies in how React handles state updates. Strings are primitive values, and React compares them by value. If the new string value is the same as the previous one, React skips the re-render. However, objects are reference types, and React compares them by reference. Even if the properties of two objects are identical, React sees them as different if their references are not the same, triggering a re-render.

To prevent unnecessary re-renders, I implemented a fix that checks whether the value object's properties remain the same between renders. If they do, it indicates that the object hasn't changed, so `setMatchText` is not called.

Users can easily also avoid this issue by:
1. Using `useRef` to store the option values array, ensuring consistent references across renders.
2. Leveraging the `selectionConfirmed` boolean sent in the custom event passed to the `onChange` handler, and updating the state inside the handler only when `selectionConfirmed` is `true`.

FilterableSelect with string values

https://github.com/user-attachments/assets/824176d8-699e-4189-96ee-2e622d0362cd

FilterableSelect with object values

https://github.com/user-attachments/assets/a2c4e195-8f93-425a-9850-dc1b0d149064

### Current behaviour

Currently the handler received on the `onChange` prop sets the value in the state even if that value has not changed. This in turn, triggers a re-render which in turn causes the `setMatchText` to be run, when it shouldn't preventing us from deleting the last character.

This is only happening for the object values not string values.

FilterableSelect with string values (working ok)

https://github.com/user-attachments/assets/ef26674e-a68c-4a33-a164-20b590a2cc20

FilterableSelect with object values (not working)

https://github.com/user-attachments/assets/5f27f296-b209-4961-8bda-857c6c514392

### Checklist

#### QA


src/components/select/filterable-select/filterable-select.test.tsx
```diff
@@ -699,6 +729,25 @@ describe("when the input is focused", () => {
     expect(screen.getByRole("combobox")).toHaveValue("red");
   });
 
+  it("if the values are objects, deleting the last charcter with backspace works as expected and the correct match text is provided", async () => {
```
DipperTheDan, line 732: **nitpick:** Just a minor grammatical/typo improvement suggestion.

```suggestion
  it("should delete the last character with backspace and the correct match text is provided, if the values are objects", async () => {
```

damienrobson-sage (approved):

src/components/select/filterable-select/filterable-select.test.tsx
mihai-albu-sage: Refactored the test description, as per the suggested improvement.

damienrobson-sage (dismissed):

DipperTheDan (dismissed):

damienrobson-sage (approved):

DipperTheDan (approved):

paulrobinson88 (approved):

Issue #7004, opened 2024-10-08
Update date-fns to v4


quinnturner:
### Description

The version of `date-fns` used is quite old. I'd like our bundle to start using newer versions of `date-fns`. Note that v3 stopped supporting IE, but I am pretty sure IE isn't supported right now.

https://blog.date-fns.org/v40-with-time-zone-support/

### Suggested solution

Update `date-fns` to latest 😄 You can also move off of `date-fns-tz` in favour of the latest `date-fns` timezone implementation.

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

tomdavies73:
FE-6858

quinnturner:
Hi, I made progress on this ticket here: https://github.com/Sage/carbon/pull/7156

PR #7148, opened 2025-01-08
feat: introduce support for react v18 (FE-6015)

Parsium:
addresses #6517

### Proposed behaviour

:warning: BREAKING CHANGES

* Dependencies `react-dnd` and `react-dnd-html5-backend` are now ESM-only packages.
* Temporary incompatibility with Strict Mode in React 18

Features

* Upgrade `react-dnd` and `react-dnd-html5-backend` to version ^16.0.0
* Introduce support for React ^18.0.0.
* Upgrade `@tanstack/react-virtual` dependency to version ^3.11.2
* `Portrait` - Add `data-element` and `data-role` props for testing purposes
* `Menu` - let browser handle tabbing to prevent focus on non-focusable items
* `Menu` - skip non-focusable submenu items when moving focus with Up, Down, Home and End keys
* `Menu` - prevent showing hover styling for non-focusable menu items

Fixes

* `Menu` - ensure focus state remains consistent between tabbing and Home/End key presses
* `Menu` - prevent root element from being focused when containing focusable content
* `Select` - debounce scroll handler on option list to reduce frequency of onListScrollBottom calls

### Current behaviour

- Carbon only supports React v17

### Checklist

- [x] Commits follow our style guide
- [x] Related issues linked in commit messages if required
- [x] Screenshots are included in the PR if useful
- [x] All themes are supported if required
- [x] Unit tests added or updated if required
- [x] Playwright automation tests added or updated if required
- [x] Storybook added or updated if required
- [x] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [x] Typescript `d.ts` file added or updated if required
- [x] Related docs have been updated if required

#### QA

- [ ] Tested in provided StackBlitz sandbox/Storybook
- [ ] Add new Playwright test coverage if required
- [ ] Carbon implementation matches Design System/designs
- [ ] UI Tests GitHub check reviewed if required

### Testing instructions

Using the Storybook build, verify that the following remain unchanged for all Carbon components:
- Visuals (appearance, layout, spacing, animations, etc.)
- Behaviour (following interactions such as with the mouse or keyboard, screen reader behaviour)
- Props (the set of available props and their default values)

Compare these attributes against the latest production build of Storybook, using the latest versions of Chrome, Safari and Firefox.

#### Exceptions
The following stories have been updated in the PR and are expected to show changes. These should not be flagged for regressions:
- `Dialog`: "Editable" story (removed)
- `RadioButton`: "With Legends and Labels" story (small change to component usage)
[...]

PR #7156, opened 2025-01-15
FE-6858 - update date-fns from ^2.30.0 to ^4.1.0

quinnturner:
Closes: https://github.com/Sage/carbon/issues/7004

### Proposed behaviour

Updates date-fns

### Checklist

#### QA

### Additional context

The main concern with my current implementation is that `<DayPicker` has a property locale that accepts a `date-fns` localize type.

`weekdaysLong` and `weekdaysShort` are now not accepted properties.

I did not know what to do in this scenario, so I removed the code entirely. However, I suspect that's not the correct behaviour.

I would appreciate guidance on how to proceed.


damienrobson-sage (approved):

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -245,9 +238,6 @@ export const DatePicker = ({
             locale={{
               localize: {
                 ...defaultLocale.localize,
-                months: monthsLong,
```
quinnturner, line 248: I really am not sure what to do here. There's a good chance this isn't the best behaviour. Looking for advice/opinions.

Parsium: The visible labels and ARIA labels look consistent with production, so this should be fine. We have passing tests that check the weekday labels are two characters long in German and Polish, so this edge case is covered :)

Parsium (approved):
Thanks for contributing @quinnturner! This should eliminate another piece of tech debt for us. I'll mark this as ready for QA.

Before we release, since this involves a major version upgrade of a dependency, we should mark it as a breaking change. Could you update the commit message to reflect this please? The message will be used in our release notes.

Here's a suggestion:

```suggestion
chore: upgrade `date-fns` to version ^4.1.0


BREAKING CHANGE: `date-fns` is now ESM-first. While CommonJS
remains supported, this change may cause issues in some environments.
If you encounter issues, you may need to update your bundler
configuration to support ESM code.
```

quinnturner:
Thanks, @Parsium; I've amended the commit message to match yours verbatim.

edleeks87:
Do not merge until https://github.com/Sage/carbon/pull/7148 has been released

Issue #5460, opened 2022-09-08
Page component with empty string title border issue


StylianosKoidakisSage:
### Current behaviour

When we render a Page component with the title as an empty string the border line of the title is present. 

### Expected behaviour

If the title is a falsy value we should not see the borderline

### CodeSandbox or Storybook URL

https://codesandbox.io/s/serene-butterfly-bqq2yk?file=/src/index.js

### Carbon Version

110.4.1

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

MacOS

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

DipperTheDan:
FE-5367

tomdavies73:
This issue has been addressed by [this PR ](https://github.com/Sage/carbon/pull/7165)

PR #7165, opened 2025-01-20
fix(heading): maintain layout when title prop is set to null

DobroTora:
### Proposed behaviour
Screenshot 2025-01-30 at 10 47 00
Screenshot 2025-01-30 at 10 47 13

Screenshot 2025-01-30 at 10 47 32

### Current behaviour

Screenshot 2025-01-30 at 10 39 50

https://github.com/Sage/carbon/issues/5460

### Checklist

#### QA

### Testing instructions

https://github.com/user-attachments/assets/fe345453-4408-4289-a363-556272b5249e

### DIFF
--- a/src/components/dialog-full-screen/dialog-full-screen.component.tsx
+++ b/src/components/dialog-full-screen/dialog-full-screen.component.tsx
@@ -185,7 +185,7 @@ export const DialogFullScreen = ({
           pagesStyling={pagesStyling}
           role={role}
         >
-          {dialogTitle()}
+          {title || headerChildren ? dialogTitle() : null}
           {closeIcon()}
           <StyledContent
             hasHeader={title !== undefined}
--- a/src/components/dialog-full-screen/dialog-full-screen.style.ts
+++ b/src/components/dialog-full-screen/dialog-full-screen.style.ts
@@ -47,7 +47,6 @@ const StyledDialogFullScreen = styled.div<{ pagesStyling?: boolean }>`
     css`
       ${StyledContent} {
         padding: 0;
-        margin-top: -25px;
       }
 
       ${StyledIconButton} {
--- a/src/components/pages/page/page.component.tsx
+++ b/src/components/pages/page/page.component.tsx
@@ -9,7 +9,7 @@ import { filterStyledSystemPaddingProps } from "../../../style/utils";
 
 export interface PageProps extends PaddingProps {
   /** The title for the page, normally a Heading component. */
-  title: React.ReactNode;
+  title?: React.ReactNode;
   /** This component supports children. */
   children: React.ReactNode;
   /** The ARIA role to be applied to the component */
@@ -38,7 +38,9 @@ const Page = ({ role, title, children, ...rest }: PageProps) => {
         ref={styledPageNodeRef}
         role={role}
       >
-        <FullScreenHeading hasContent>{title}</FullScreenHeading>
+        {title ? (
+          <FullScreenHeading hasContent>{title}</FullScreenHeading>
+        ) : null}
         <StyledPageContent
           data-element="carbon-page-content"
           data-role="page-content"
--- a/src/components/pages/pages-test.stories.tsx
+++ b/src/components/pages/pages-test.stories.tsx
@@ -10,7 +10,7 @@ import Box from "../box";
 
 export default {
   title: "Pages/Test",
-  includeStories: ["DefaultStory", "DifferentPageHeights"],
+  includeStories: ["DefaultStory", "DifferentPageHeights", "WithoutTitle"],
   parameters: {
     info: { disable: true },
     chromatic: {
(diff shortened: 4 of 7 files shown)


src/components/heading/heading.style.ts
```diff
@@ -12,6 +12,7 @@ import addFocusStyling from "../../style/utils/add-focus-styling";
 
 const StyledHeading = styled.div`
   width: 100%;
+  min-height: 53px;
```
DipperTheDan, line 15: **issue:** I am not so sure this fix works. We may need to speak with UX to see if we should render the white heading section if no title is provided. I don't think we have designs that specify what should happen if an empty `title` prop is passed. 
 
Screenshot 2025-01-21 at 16 22 25

DobroTora: Hi Dan, we were trying to minimise the changes for chromatic. As you see it was not the best solution.

edleeks87: Agree with @DipperTheDan, I chatted with @harpalsingh about it and we shouldn't be rendering the empty white space if there's no title text. The current interface requires `title` we should make that optional and then conditionally render the heading section in `Page`. This will mean that if no title is rendered the whole `Page` is the grey background, if a title is passed we render the white section and then the grey part is for the content etc. We should also do the same in the fullscreen dialog as I notice we are rendering two heading sections (see below). 

image

DobroTora: Ok, I get it, but as I was reading original thread Stylianos Koidakis mentioned that they needed a page without the title.

Parsium (commented):
**issue**: it looks like we some negative margins are being applied, which is affecting the positioning of the `Page` content:

Screenshot 2025-02-03 at 09 39 45

It appears that a negative top margin is being applied within `DialogFullscreen` ([link to source](https://github.com/Sage/carbon/blob/FE-5367/src/components/dialog-full-screen/content.style.ts)). Would you be able to take a look and see if adjusting this could help?

Parsium:
**suggestion**: I noticed there a few new inline comments that describe code that already seems self-explanatory. To keep the code clean, we might want to remove any that aren't adding significant value.

Parsium:
Thanks for the updates @DobroTora 👍🏼 Just a few comments from me

src/components/heading/heading.style.ts
DipperTheDan, line 15: **suggestion:** I don't think we need this line of code anymore, so it can be deleted. It's from when you were trying to get the Chromatic snapshots to look correct when you initially picked up the ticket.

src/components/pages/page/page.component.tsx
```diff
@@ -38,7 +38,10 @@ const Page = ({ role, title, children, ...rest }: PageProps) => {
         ref={styledPageNodeRef}
         role={role}
       >
-        <FullScreenHeading hasContent>{title}</FullScreenHeading>
+        {/* Conditionally rendered the heading */}
```
DipperTheDan, line 41: **suggestion:** This comment is not required as the code plainly and clearly describes that the heading is conditionally rendered.

src/components/pages/pages-test.stories.tsx
```diff
@@ -10,7 +10,7 @@ import Box from "../box";
 
 export default {
   title: "Pages/Test",
-  includeStories: ["DefaultStory", "DifferentPageHeights"],
+  includeStories: ["DefaultStory", "DifferentPageHeights", "TEST_TEST"],
```
DipperTheDan, line 13:
```suggestion
  includeStories: ["DefaultStory", "DifferentPageHeights", "WithoutTitle"],
```

src/components/pages/pages-test.stories.tsx
```diff
@@ -525,3 +525,111 @@ export const DifferentPageHeights = () => {
...
+              />
+            }
+          >
+            Third Page
+          </Page>
+        </Pages>
+      </DialogFullScreen>
+    </div>
+  );
+};
```
DipperTheDan, line 613: **suggestion:** This whole example can be deleted as @Parsium commented previously regarding the inclusion of the `WithoutTitle` example, which has superseded the `TEST_TEST` example, as that was just to check that removing the header would be possible.

src/components/dialog-full-screen/dialog-full-screen.style.ts
```diff
@@ -58,8 +58,10 @@ const StyledDialogFullScreen = styled.div<{ pagesStyling?: boolean }>`
         z-index: 1;
       }
 
+      /* If the component has a title, we need to keep the same styles. If it does not have a title,
+       we need the padding to compensate for loss of 32px height */
```
DipperTheDan, line 62: **suggestion: ** This comment can be deleted as it refers to the code we had previously and no longer provides the correct context to the styling code below it.

src/components/dialog-full-screen/dialog-full-screen.component.tsx
```diff
@@ -185,7 +185,8 @@ export const DialogFullScreen = ({
           pagesStyling={pagesStyling}
           role={role}
         >
-          {dialogTitle()}
+          {/* Conditionally render the dialog title */}
```
DipperTheDan, line 188: **suggestion:** This comment is not required as the code plainly and clearly describes that the `dialogTitle` is conditionally rendered.

src/components/dialog-full-screen/dialog-full-screen.style.ts
```diff
@@ -59,7 +58,7 @@ const StyledDialogFullScreen = styled.div<{ pagesStyling?: boolean }>`
       }
 
       ${StyledFullScreenHeading} {
-        padding: 32px 32px 0;
+        padding: 64px 32px 0;
```
Parsium, line 61: **suggestion**: what do you think about restoring the top padding value? This would keep the rendered element's total height the same as in production, minimising the change in positioning.

```suggestion
        padding: 32px 32px 0;
```

DobroTora: I cannot agree more @Parsium

Parsium (approved):
LGTM @DobroTora ✅

DipperTheDan (approved):

paulrobinson88 (approved):

PR #7176, opened 2025-01-23
feat(loader): remove progressbar role and add loaderLabel prop

### LINKED ISSUE #6956: Incorrect role given to loader
### Description

Loaders have a role of progressbar which fails WCAG 1.3.1 Info and relationships.

### Reproduction

https://carbon.sage.com/?path=/docs/loader--docs

### Steps to reproduce

_No response_

SBS-68239

### Suggested solution

Role should be removed, instead use visually hidden text of 'loading' for screen reader user.

### Carbon version

142.9.1

### Design tokens version

_No response_

### Relevant browsers

Safari

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #6956

### Proposed behaviour

- Remove role of "progressbar". 
- Deprecate `aria-label` prop.  
- Add `loaderLabel` prop to set a visually hidden label for the component instead of an aria-label. 

### Current behaviour

`Loader` has role of "progressbar". 

### Checklist

#### QA


src/components/loader/loader.mdx
```diff
@@ -67,11 +67,7 @@ This is an example of the large Loader component. The larger size is only used w
...
 please remember to pass the `isInsideButton` prop to the `Loader` component.
 
-This example also demonstrates how to ensure that the `Loader` component is accessible when used inside of the `Button` component. 
-It is important to wrap the `Button` component in a `span` (use when elements are inline) or `div` (use when elements are in a block) tag and pass the `aria-live` attribute with the value of `polite`. This will ensure that the screen reader will reliabily announce the updates of the button to the user.
-
-Please note that ARIA live regions provide a means for conveying dynamic updates to users who rely on assistive technologies like screen readers. 
-These updates in this case include changes to the button's content. By incorporating live regions permanently into the DOM, we can ensure that users are consistently informed of relevant changes as they occur.
+Note that `Loader` has a role of "status", meaning it has an implicit `aria-live="polite"` and will be announced by screen readers.
```
edleeks87, line 70: 🔥

src/components/loader/loader.stories.tsx
```diff
@@ -54,16 +54,14 @@ export const InsideButton: Story = () => {
...
-        aria-label={ariaContent}
-        onClick={handleButtonClick}
-      >
-        {buttonContent}
-      </Button>
-    </span>
+    <Button
+      m={2}
+      buttonType="primary"
+      aria-label={ariaContent}
```
edleeks87, line 60: Question (non-blocking): Does this need an aria-label at all when isLoading is false as there's text content at that point?

```
  const ariaContent = isLoading ? "Loading" : undefined;

**Review by edleeks87 — DISMISSED** [review]

File: src/components/loader/loader.stories.tsx [hunk 3]
```diff
@@ -54,16 +54,14 @@ export const InsideButton: Story = () => {
   const buttonContent = isLoading ? <Loader isInsideButton /> : "Click me";
   const ariaContent = isLoading ? "Loading" : "Click me";
   return (
-    <span aria-live="polite">
```
Parsium, line 57: **issue**: this change impacts screen reader behaviour. While `Loader` is marked as an ARIA live region, screen readers may not announce its content because it's conditionally rendered:

https://github.com/user-attachments/assets/4dece5e9-96ec-447d-bf88-24552c0321d9

For the live region to work as expected, it needs to be present in the DOM before any updates are made to its content.

Given most consumers will likely render `Loader` conditionally, they would need to wrap it in an ARIA live region themselves and ensure the wrapper remains in the DOM both before and during the display of the loading visuals or text.

To make the component work more seamlessly out of the box, we might need to explore alternative approaches, as creating a live region within the component may not work well when it is conditionally rendered.

src/components/loader/loader.test.tsx
```diff
@@ -36,17 +70,27 @@ describe("when the user allows animations", () => {
     expect(squares).toHaveLength(3);
   });
 
-  test("root element has accessible name", () => {
-    render(<Loader />);
+  test("root element has default accessible name", () => {
```
Parsium, line 73: **suggestion(non-blocking)**: how about refining this slightly for clarity? This would make it clearer now we're no longer setting an actual `aria-label` on the root element:

```suggestion
  test("visually hidden label is set to "Loading" when neither `aria-label` nor `loaderLabel` are passed", () => {
```

Parsium (dismissed):

edleeks87 (approved):

harpalsingh:
Reviewed this and spoke with Martin about it as well. We may need to document how users can add aria-polite around a container where it might be needed on a dynamic usage of loader but as updated the component itself does not need it out the box.

dariusbercea-sage (approved):

edleeks87 (approved):

PR #7177, opened 2025-01-24
feat(portrait, profile): allow custom background/foregound colours for the Portrait and Profile components

damienrobson-sage:
### Proposed behaviour

- Add a prop to `Portrait` that allows developers to specify their own custom background colour via hex code;
- Add utilities to the codebase to take the given hex code and generate an appropriate, accessible foreground colour to ensure that any colour can be used without e.g. contrast or luminance issues.
- Allow users to override the foreground colour with their own (only if `backgroundColor` is specified), to support particular branding requirements (e.g. using company colours for initials)

### Current behaviour

Customers have a choice of two colour modes for the `Portrait` component (and by association, the `Profile` component) which effectively boils down to "light" and "dark" variants.

### Checklist

#### QA

### Additional context

The functions that calculate the contrast and luminance values were provided by Co-Pilot and then modified to be re-usable. An extensive test suite has been added to ensure that combinations of colours are appropriately accessible.

### Testing instructions

A story has been added to the `Portrait` documentation to demonstrate the capabilities of the new prop. The `Profile` documentation has also been updated to include an example which demonstrates usage via that particular component.

### DIFF
--- a/src/components/portrait/__internal__/utils.test.ts
+++ b/src/components/portrait/__internal__/utils.test.ts
@@ -1,0 +1,118 @@
+import getColoursForPortrait from "./utils";
+
+test("returns the default string if no arguments are passed", () => {
+  const result = getColoursForPortrait(undefined);
+  expect(result).toBe(
+    `background-color: var(--colorsUtilityReadOnly400); color: var(--colorsUtilityYin090);`,
+  );
+});
+
+test("returns a fixed string if only the `backgroundColor` argument is set to true", () => {
+  const result = getColoursForPortrait("#FF0000");
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a fixed string if the `darkBackground` argument is set to true", () => {
+  const result = getColoursForPortrait(undefined, true);
+  expect(result).toBe(
+    "background-color: var(--colorsUtilityYin090); color: var(--colorsUtilityReadOnly600);",
+  );
+});
+
+test("returns a fixed string if neither `darkBackground` nor `backgroundColor` argument are defined", () => {
+  const result = getColoursForPortrait(undefined, false);
+  expect(result).toBe(
+    `background-color: var(--colorsUtilityReadOnly400); color: var(--colorsUtilityYin090);`,
+  );
+});
+
+test("returns a string with the custom background color if the `backgroundColor` argument is defined", () => {
+  const result = getColoursForPortrait("#FF0000", false);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a string with the custom background color if only the `backgroundColor` argument is defined and all others are false", () => {
+  const result = getColoursForPortrait("#FF0000", false, false, false);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a string with the custom background color if the `backgroundColor` argument is defined and `largeText` argument is true", () => {
+  const result = getColoursForPortrait("#FF0000", false, true);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a string with the custom background color if the `backgroundColor` and `largeText` arguments are defined and `strict` argument is false", () => {
+  const result = getColoursForPortrait("#FF0000", false, true, false);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a string with the custom background color if the `backgroundColor` and `largeText` arguments are defined and `strict` argument is true", () => {
+  const result = getColoursForPortrait("#FF0000", false, true, true);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+describe("Contrast ratio tests", () => {
+  it("uses a white foreground colour if the white contrast ratio meets the minimum contrast threshold and is higher than the black contrast ratio", () => {
+    const result = getColoursForPortrait("#0000FF");
+    expect(result).toBe("background-color: #0000FF; color: #FFFFFF;");
+  });
+
+  it("uses a black foreground colour if the  black contrast ratio meets the minimum contrast threshold", () => {
+    const result = getColoursForPortrait("#FFFF00");
+    expect(result).toBe(
+      "background-color: #FFFF00; color: var(--colorsUtilityYin090);",
+    );
+  });
+});
+
+test("returns a string with the custom background color and light text if the `backgroundColor` argument is set to a colour with poor contrast ratios (higher white contrast)", () => {
+  const result = getColoursForPortrait("#0000FF");
+  expect(result).toBe("background-color: #0000FF; color: #FFFFFF;");
+});
+
+test("returns a string with the custom colors if the `backgroundColor` and `foregroundColor` arguments are provided and all others are false", () => {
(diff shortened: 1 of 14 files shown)


src/components/portrait/portrait.mdx
```diff
@@ -34,53 +34,76 @@ import Portrait from "carbon-react/lib/components/portrait";
...
+table at the bottom of this page.
 
 [Vist Props Table](#props)
 
 <Canvas of={PortraitStories.WithMargin} />
 
+### Custom colors
+
+The `Portrait` component provides a set of props that allow for custom foreground and background colors
+to be applied. These props are `backgroundColor` and `foregroundColor`, and both accept a HEX color code.
```
DipperTheDan, line 95: **suggestion:** It might also be nice to mention here that this will also accept a design token value.

src/components/portrait/portrait.style.tsx
```diff
@@ -56,14 +63,14 @@ export const StyledIcon = styled(Icon)<Pick<StyledPortraitProps, "size">>`
...
-  background-color: ${({ darkBackground }) =>
-    darkBackground
-      ? "var(--colorsUtilityYin090)"
-      : "var(--colorsUtilityReadOnly400)"};
+  ${({ darkBackground, backgroundColor, size, foregroundColor }) =>
+    getColoursForPortrait(
+      backgroundColor,
+      darkBackground,
+      !["XS", "S"].includes(size),
+      true,
```
tomdavies73, line 71: **question**: It appears this `strict` param is always true, this util is internal so is this correct? Is there any instances where it can be false?

src/components/portrait/portrait.test.tsx
```diff
@@ -199,3 +216,125 @@ test("allows a custom onClick function to be passed via the `onClick` prop", asy
...
+    const icon = screen.getByTestId("icon");
+    expect(icon).toBeVisible();
+    expect(icon).toHaveAttribute("type", "individual");
+
+    const container = screen.getByTestId("portrait");
+    expect(container).toHaveStyle("background-color: #000000");
+    expect(container).toHaveStyleRule("color", "#A0FFAF");
+  });
+
+  it("renders with the correct colours when the `darkBackground` prop is set to true", () => {
```
tomdavies73, line 322: **suggestion** maybe write this as overides the colours set by `darkBackground`etc. As this may be a little bit clearer

src/components/portrait/portrait.test.tsx
```diff
@@ -199,3 +216,125 @@ test("allows a custom onClick function to be passed via the `onClick` prop", asy
...
+      expect(icon).toBeVisible();
+      expect(icon).toHaveAttribute("type", "individual");
+
+      const container = screen.getByTestId("portrait");
+      expect(container).toHaveStyle(`color: ${value}`);
+    });
+  });
+
+  ["XS", "S", "M", "ML", "L", "XL", "XXL"].forEach((size) => {
+    it(`renders with the correct colours when the size is set to ${size}`, () => {
```
tomdavies73, line 290: **question**: is this needed for coverage? If not we could probably remove this, as i'm not sure how valuable it is

src/components/portrait/__internal__/utils.test.ts
```diff
@@ -0,0 +1,104 @@
...
+});
+
+test("returns a string with the custom background color if the `backgroundColor` argument is truthy", () => {
+  const result = getColoursForPortrait("#FF0000", false);
+  expect(result).toBe(
+    "background-color: #FF0000; color: var(--colorsUtilityYin090);",
+  );
+});
+
+test("returns a string with the custom background color if only the `backgroundColor` argument is truthy and all others are falsey", () => {
```
tomdavies73, line 38: **sugestion**: i'm not sure about `"truthy"` being used in instances where a specific non-boolean param is used. I would just used defined, or something along those lines

playwright/index.tsx
```diff
@@ -25,10 +25,7 @@ const mountedTheme = (theme: string) => {
 
 // Setup required providers on mounted component before running test. See https://playwright.dev/docs/test-components#hooks
 beforeMount<HooksConfig>(async ({ App, hooksConfig }) => {
-  const {
-    theme = "sage",
-    validationRedesignOptIn,
-  } = hooksConfig || {};
+  const { theme = "sage", validationRedesignOptIn } = hooksConfig || {};
```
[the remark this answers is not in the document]
damienrobson-sage: This keeps happening and Im not sure why - I've removed it from previous commits but it keeps being picked up whenever I run Prettier 😄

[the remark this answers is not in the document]
damienrobson-sage: Just checked and it's already in [master](https://github.com/Sage/carbon/blob/master/playwright/index.tsx#L28) so I don't know why this is being flagged 🤷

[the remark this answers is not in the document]
damienrobson-sage: As per reply to Dan above, this is already in `master` so I don't know why it's being flagged here.

src/components/portrait/portrait.style.tsx
damienrobson-sage: Currently no, but there may be a time when we need it to be available.

WCAG 2.0 level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text; AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text. Whilst our target is AA compliance, we should support AAA too, even though currently its unlikely to be required. 

I've flipped the default to false (so that we're using AA settings) and updated the argument's description for clarity.

src/components/portrait/portrait.mdx
damienrobson-sage: It won't actually - the current approach only works with hex codes to keep in line with the scope of the story. Would be worth investigating if people need that ability, but I wouldn't expect so, at least not out of the door.

src/components/switch/switch.component.tsx
```diff
@@ -269,11 +269,7 @@ export const Switch = React.forwardRef(
             flexDirection={!reverse ? reverseDirection : direction}
             width={labelInline ? "100%" : "auto"}
           >
-            <Box
-              data-role="label-wrapper"
-              // width={labelInline ? rest.labelWidth : 30}
-              // minWidth={label && labelInline ? "32px" : 0}
```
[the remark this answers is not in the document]
DipperTheDan: No need to undo it, just move it into a new separate commit on the same PR.

src/components/portrait/portrait.mdx
damienrobson-sage: Never mind, it's working now

DipperTheDan (approved):

src/components/portrait/portrait.test.tsx
```diff
@@ -121,3 +134,131 @@ test("allows a custom onClick function to be passed via the `onClick` prop", asy
...
+    const icon = screen.getByTestId("icon");
+    expect(icon).toBeVisible();
+    expect(icon).toHaveAttribute("type", "individual");
+
+    const container = screen.getByTestId("portrait");
+    expect(container).toHaveStyle("background-color: #000000");
+    expect(container).toHaveStyleRule("color", "#A0FFAF");
+  });
+
+  it("supports design tokens being used for the `backgroundColor` and `forefroundColor` props", () => {
```
tomdavies73, line 244: **suggestion**: I agree with the addition here, but I think a pw test which actually tests the tokens are resulting in the correct rgb/hex values would be beneficial too

tomdavies73 (commented):
@damienrobson-sage good stuff, I still do think profile should be added to the commits scope though, due to there being changes within the component code 👍

tomdavies73 (dismissed):

DipperTheDan (dismissed):

DipperTheDan (approved):

tomdavies73 (approved):

paulrobinson88 (approved):

Issue #7179, opened 2025-01-27
Margin left spacing in Switch component with reverse = true prop, the label text wasn't centered


EstebanGoldMuros:
### Description

I'm using the Switch component with the reverse prop activated the margin left spacing has 24px, if I use ml="0" then it continue with this margin.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-peqjdaze?file=src%2FApp.tsx

### Carbon version

146.5.3

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Confidentiality

- [x] I confirm there is no confidential or commercially sensitive information included.

damienrobson-sage:
@ljemmo @harpalsingh regarding the ON/OFF labels, can you advise as to whether this is something we want to consider, please?

damienrobson-sage:
FE-7066

Parsium:
> [@ljemmo](https://github.com/ljemmo) [@harpalsingh](https://github.com/harpalsingh) regarding the ON/OFF labels, can you advise as to whether this is something we want to consider, please?

A separate feature request has been raised for this: #7185

ljemmo:
Can you confirm what you mean by 'on off labels'? This issue seems to be about margin placement?

damienrobson-sage:
@ljemmo see the link provided by James above for clarification 😄

ljemmo:
@damienrobson-sage so i understand that he's wanting the margin bug to be fixed when the reverse prop is set to true (BTW DS no longer support this layout so wouldn't advise this usage where possible). 

However, i don't understand the comment about On / Off labels. I missed @Parsium 's comment so will review the data there.

PR #7198, opened 2025-02-10
fix(switch): margins and alignment

mihai-albu-sage:
fixes: #7179

### Proposed behaviour
When the `Switch` component has `labelInline` set to `true` and `reverse` set to `false`, the `input-wrapper` container inside the `Switch` will have `margin-left` set to `0`.

afrter1

When the `Switch` component has `labelInline` set to `true` and `reverse` set to `true`, the `input-wrapper` container inside the `Switch` will have `margin-right` set to `0`.

after2

Also, the label has been aligned with the Switch when it is displayed inline.

### Current behaviour

When the `Switch` component has `labelInline` set to `true` and `reverse` set to `false`, the `input-wrapper` container inside the `Switch` receives a left margin of value `3`, dictated by `defaultInputWrapperMargin`. This margin can not be changed.

labelinline-true-reverse-false-before

labelInline-true-reverse-true-before

Also, the label is not aligned with the switch.

### Checklist

#### QA

### Testing instructions

1. Go to the story http://localhost:9001/?path=/story/switch-test--with-margin
2. Set `reverse` to `false`
3. Set `ml` to `0`
4. Inspect the `input-wrapper` which should have `margin-left: 0px`


src/components/switch/switch.component.tsx
```diff
@@ -252,7 +252,25 @@ export const Switch = React.forwardRef(
...
+        }
+      } else if (marginProps.mr === 0 || marginProps.marginRight === 0) {
+        defaultInputWrapperMargin = 0;
+      } else {
+        defaultInputWrapperMargin = 3;
+      }
+    } else {
+      defaultInputWrapperMargin = 0;
+    }
+
```
nuria1110, line 273: Unless I am missing something, it looks like the incorrect margins described in the issue are set internally on the `input-wrapper` and we could just set them depending on whether the `reverse` prop is true or false on lines 320/321, as I don't think they should be set by default:

```
ml={reverse ? errorMargin : undefined}
mr={!reverse ? errorMargin : undefined}
```
image
image


I don't believe we should be setting these depending on the margin props consumers pass as those are applied to the component wrapper and should not affect the margins we set internally.

mihai-albu-sage: Yes, I believe it makes more sense to have it like this. The code has been updated.

src/components/switch/switch-test.stories.tsx
```diff
@@ -159,3 +165,37 @@ WithLongTextStrings.parameters = {
...
+      </Box>
+    </CarbonProvider>
+  );
+};
+
+WithMargin.storyName = "With Margin";
+WithMargin.parameters = {
+  chromatic: {
+    disableSnapshot: false,
+  },
```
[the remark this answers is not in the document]
mihai-albu-sage: Done, added the `sage` theme.

nuria1110 (approved):

DipperTheDan (approved):

dariusbercea-sage (approved):

PR #7199, opened 2025-02-10
fix(time): ensure that ARIA attributes conform to valid values - FE-7060

DipperTheDan:
### Proposed behaviour

Fix axe warning regarding `ARIA attributes must conform to valid values`. This warning is flagged because the `inputHint` is passing an ID to the `combinedAriaDescribedby` value when `inputHint` is not rendered in the DOM. This fix ensures that an ID is not passed to the ariaDescribedby prop unless `inputHint` exists.

### Current behaviour

`ARIA attributes must conform to valid values` axe warning is present in Storybook.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- Ensure that there are no new axe violations with `Time` component.
- Ensure that screenreader behaviour is not affected by this change.


tomdavies73 (commented):
LGTM @DipperTheDan but I think it would be worth adding a unit test to check the value of the `aria-describedby` attribute with the presence and without the presence of the `inputHint` prop would be beneficial to establish a baseline

src/components/time/time.test.tsx
```diff
@@ -998,3 +998,26 @@ test("should apply the custom `data-` attributes on the toggle component wrapper
...
+      value={{ hours: "", minutes: "" }}
+      onChange={() => {}}
+      inputHint="hint"
+    />,
+  );
+
+  const fieldset = screen.getByRole("group");
+  const hint = screen.getByText("hint");
+
+  expect(fieldset).toHaveAttribute("aria-describedby", hint.id);
```
nuria1110, line 1014:
```suggestion
  expect(fieldset).toHaveAccessibleDescription("hint");
```
**nit-pick(non-blocking):** We could use this assertion instead to make sure that the computed accessible description is set correctly and not just the attribute being present.

nuria1110 (approved):

tomdavies73 (approved):

paulrobinson88 (approved):

PR #7204, opened 2025-02-13
feat(adaptive-sidebar): Implements adaptive sidebar

damienrobson-sage:
implements a sidebar component to be used for Help and Chat purposes

### Proposed behaviour

Implement an adaptive sidebar, which can be used as a sibling component to display toggled elements alongside existing layouts

### Current behaviour

Customers are restricted to the modal Sidebar component, which removes interaction from the screen and hides information, or the Drawer component which would require changes to keep it working as is currently implemented whilst also supporting the new styles, etc. 

### Checklist

#### QA

### DIFF
--- a/src/components/adaptive-sidebar/adaptive-sidebar-test.stories.tsx
+++ b/src/components/adaptive-sidebar/adaptive-sidebar-test.stories.tsx
@@ -1,0 +1,458 @@
+import React, { useState } from "react";
+
+import { StoryObj } from "@storybook/react/*";
+import AdaptiveSidebar from ".";
+import Box from "../box";
+import Button from "../button";
+import { Checkbox } from "../checkbox";
+import DateInput, { DateChangeEvent } from "../date";
+import Form from "../form";
+import Hr from "../hr";
+import Icon from "../icon";
+import IconButton from "../icon-button";
+import Search from "../search";
+import { Select, Option } from "../select";
+import Textbox from "../textbox";
+import { Tile } from "../tile";
+import Typography from "../typography";
+
+import isChromatic from "../../../.storybook/isChromatic";
+
+const defaultOpenState = isChromatic();
+
+export default {
+  title: "Adaptive Sidebar/Test",
+  parameters: {
+    chromatic: { disableSnapshot: true },
+    layout: "fullscreen",
+    info: { disable: true },
+  },
+};
+
+type StoryType = StoryObj<typeof AdaptiveSidebar>;
+
+export const Default: StoryType = ({
+  backgroundColor,
+  childText,
+  renderAsModal,
+}: {
+  backgroundColor: "white" | "black" | "app";
+  childText: string;
+  renderAsModal: boolean;
+}) => {
+  const [adaptiveSidebarOpen, setAdaptiveSidebarOpen] =
+    useState(defaultOpenState);
+  const bodyText =
+    "Praesent vulputate accumsan quam eget pellentesque. Sed condimentum, diam ut dapibus ornare, enim sem gravida dolor, vel molestie arcu nulla eget orci. Mauris vulputate in massa eu ornare. Aenean condimentum ligula a quam mattis, sit amet cursus augue consectetur. Mauris mollis tristique ex, vitae dapibus sapien bibendum at. Fusce blandit, tellus sit amet commodo vehicula, velit ipsum viverra urna, at feugiat magna ante non est. Nullam malesuada urna et varius rhoncus. Fusce ac blandit leo.";
+
+  return (
+    <Box display="flex" flexDirection="row">
+      <Box>
+        <Button
+          onClick={() => setAdaptiveSidebarOpen(!adaptiveSidebarOpen)}
+          mb={2}
+        >
+          {adaptiveSidebarOpen ? "Close" : "Open"}
+        </Button>
+        <Typography variant="p">{childText}</Typography>
+      </Box>
+      <AdaptiveSidebar
+        backgroundColor={backgroundColor}
+        open={adaptiveSidebarOpen}
+        renderAsModal={renderAsModal}
+        width="300px"
+      >
+        <Box p={2} display="flex" flexDirection="column">
+          <Button onClick={() => setAdaptiveSidebarOpen(false)} mb={2}>
+            Close
+          </Button>
+          {bodyText}
+        </Box>
+      </AdaptiveSidebar>
+    </Box>
+  );
+};
+
+Default.storyName = "Default";
+Default.argTypes = {
+  backgroundColor: {
+    control: {
+      type: "select",
+    },
+    options: ["app", "white", "black"],
+  },
+  renderAsModal: {
+    control: "boolean",
+    description: "Render as a modal",
+  },
+};
+
+Default.args = {
+  backgroundColor: "white",
+  renderAsModal: false,
+};
+
+Default.decorators = [
+  (Story) => (
+    <Box height="100vh" width="100vw">
+      <Story />
+    </Box>
+  ),
+];
+
+Default.parameters = {
+  chromatic: { disableSnapshot: false, viewports: [1200, 900] },
+  layout: "fullscreen",
+  themeProvider: {
+    chromatic: { theme: "sage" },
+  },
+};
+
+export const WithForm: StoryObj = () => {
+  const [adaptiveSidebarOpen, setAdaptiveSidebarOpen] =
+    useState(defaultOpenState);
+  const [date, setDate] = useState("01/06/2020");
+  const selectOptions = [
+    {
+      id: "1",
+      name: "Orange",
+    },
+    {
+      id: "2",
+      name: "Blue",
+    },
+    {
+      id: "3",
+      name: "Green",
+    },
+    {
+      id: "4",
+      name: "Black",
+    },
+    {
+      id: "5",
+      name: "Yellow",
+    },
+    {
+      id: "6",
+      name: "White",
+    },
+    {
+      id: "7",
+      name: "Magenta",
+    },
+    {
+      id: "8",
+      name: "Cyan",
+    },
+    {
+      id: "9",
+      name: "Red",
+    },
+    {
+      id: "10",
+      name: "Grey",
+    },
+    {
+      id: "11",
+      name: "Purple",
+    },
+  ];
+
+  return (
+    <Box display="flex" flexDirection="row" justifyContent="space-between">
+      <Box>
+        <Button
+          onClick={() => setAdaptiveSidebarOpen(!adaptiveSidebarOpen)}
+          mb={2}
+        >
+          {adaptiveSidebarOpen ? "Close" : "Open"}
+        </Button>
+        <Typography variant="p">
+          This demonstrates using complex components within the adaptive sidebar
+        </Typography>
+      </Box>
+      <AdaptiveSidebar open={adaptiveSidebarOpen} width="500px">
+        <Form
+          m={2}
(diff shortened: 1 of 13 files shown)


src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
```diff
@@ -0,0 +1,485 @@
...
+          <Box
+            display="flex"
+            flexDirection="row"
+            justifyContent="space-between"
+            p={1}
+          >
+            <Typography variant="h3">Content</Typography>
+            <Button onClick={() => setAdaptiveSidebarOpen(false)}>Close</Button>
+          </Box>
+          <Hr my={0} />
```
DipperTheDan, line 124: **issue:** You've used `Hr` here, but it doesn't appear in the sidebar. I can see it in the DOM when I inspect it, it just seems not to be visible. 

Screenshot 2025-02-14 at 13 25 05

src/components/adaptive-sidebar/adaptive-sidebar-test.stories.tsx
```diff
@@ -0,0 +1,408 @@
...
+
+Default.args = {
+  backgroundColor: "white",
+  childText:
+    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi at odio ultricies, luctus dolor at, fringilla elit. Nulla non nunc eu sapien tempus porta. Nullam sodales nisi ut orci efficitur, nec ullamcorper nunc pulvinar. Integer eleifend a augue ac accumsan. Fusce ultrices auctor aliquam. Sed eu metus sit amet est tempor ullamcorper. Praesent eu elit eget lacus fermentum porta at ut dui.",
+  defaultState: false,
+  renderAsModal: false,
+};
+
+export const WithForm = () => {
```
DipperTheDan, line 100: **nitpick/suggestion:** It might be worth adding a margin to this example just so the focus isn't cut off. Not the end of the world as it's a test story though. 

Screenshot 2025-02-14 at 13 28 05

src/components/adaptive-sidebar/adaptive-sidebar-test.stories.tsx
```diff
@@ -0,0 +1,408 @@
...
+          />
+          <Checkbox name="checkbox" label="Do you like my Dog" />
+          <div>This is an example of an adaptive with a Form as content</div>
+        </Form>
+      </AdaptiveSidebar>
+    </Box>
+  );
+};
+
+export const ExampleImplementation = () => {
```
DipperTheDan, line 207: **praise:** I like this idea.

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -0,0 +1,120 @@
...
+  children,
+  contentMargin = "var(--spacing200)",
+  height = "100%",
+  open,
+  renderAsModal = false,
+  width = "320px",
+  ...props
+}: AdaptiveSidebarProps) => {
+  const largeScreen = useIsAboveBreakpoint(adaptiveBreakpoint);
+  const adapativeSidebarRef = useRef<HTMLDivElement>(null);
```
DipperTheDan, line 49: **nitpick:** Spelling mistake

```suggestion
  const adaptiveSidebarRef = useRef<HTMLDivElement>(null);
```

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -0,0 +1,120 @@
...
+          {children}
+        </Box>
+      </StyledSidebar>
+    );
+  }
+
+  return (
+    <>
+      <CSSTransition
+        nodeRef={adapativeSidebarRef}
```
DipperTheDan, line 95: To bring inline with the spelling mistake spotted earlier in the file.

```suggestion
        nodeRef={adaptiveSidebarRef}
```

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -0,0 +1,120 @@
...
+          {...colours}
+        >
+          {children}
+        </Box>
+      </StyledSidebar>
+    );
+  }
+
+  return (
+    <>
```
DipperTheDan, line 93: **suggestion:** I don't _think_ you need to wrap this all in a fragment.

src/components/adaptive-sidebar/adaptive-sidebar.pw.tsx
DipperTheDan, line 1: **suggestion:** If you want to add tests to cover the behaviour on smaller screens, you can do this in Playwright like this:

`await page.setViewportSize({ width: 600, height: 600 });`

I'll leave that up to you and the other reviewer if there is any value in that.

src/components/adaptive-sidebar/adaptive-sidebar.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  jest.clearAllMocks();
+});
+
+[
+  { adaptiveBreakpoint: true, reduceMotion: true },
+  { adaptiveBreakpoint: true, reduceMotion: false },
+  { adaptiveBreakpoint: false, reduceMotion: true },
+  { adaptiveBreakpoint: false, reduceMotion: false },
+].forEach(({ adaptiveBreakpoint, reduceMotion }) => {
+  test(`should render correctly if the user has set 'reduced motion' in their OS settings to ${reduceMotion} and the adaptive breakpoint ${adaptiveBreakpoint ? "has" : "has not"} been reached`, async () => {
```
DipperTheDan, line 70: **nitpick:** formatting on this test and the one below is different to rest in the file.

src/components/adaptive-sidebar/adaptive-sidebar.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const user = userEvent.setup();
+    render(<MockApp />);
+    expect(screen.queryByText("My Content")).not.toBeInTheDocument();
+    const openButton = screen.getByTestId("adaptive-sidebar-control-button");
+    await user.click(openButton);
+    expect(screen.getByText("My Content")).toBeInTheDocument();
+  });
+});
+
+it("should handle changes in screen size and motion preferences", async () => {
```
DipperTheDan, line 82: **nitpick:** formatting on this test and the one above is different to rest in the file.

DipperTheDan (commented):
Good work overall on this, @damienrobson-sage. Just several comments from me and most of them are not too serious or real issues.

src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
edleeks87, line 1: comment: If we're intending for all of these to be captured by chromatic we'll need to use the `isChromatic` util so that that the component is open when the snapshot is taken

src/components/adaptive-sidebar/adaptive-sidebar.test.tsx
edleeks87: I agree with Dan these are a little harder to read than the other tests so might benefit from some more spacing between the renders/querying/asserting etc like you have in other tests

edleeks87 (commented):
Really good work @damienrobson-sage I've left some minor comments/suggestions/questions but overall happy with how the component looks when playing around with it locally 👍

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }, [backgroundColor]);
+
+  if (renderAsModal || !largeScreen) {
+    return (
+      <StyledSidebar open={open} p={0} backgroundColor={backgroundColor}>
+        <Box
+          height="100%"
+          data-role="adaptive-sidebar-content-wrapper"
+          {...colours}
```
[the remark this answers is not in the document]
damienrobson-sage: No, they're just handled differently. I took this approach so as to not change the underlying `Sidebar` component - background colours are handled in "normal" mode via the `backgroundColor` prop of the `AdaptiveSidebar`

DipperTheDan (dismissed):

edleeks87 (approved):

src/components/adaptive-sidebar/adaptive-sidebar.style.ts
```diff
@@ -0,0 +1,70 @@
+import styled, { css } from "styled-components";
+
+import { margin, MarginProps, padding, PaddingProps } from "styled-system";
+
+import Box from "../box";
+import Sidebar, { SidebarProps } from "../sidebar";
+
+import { AdaptiveSidebarProps } from "./adaptive-sidebar.component";
+
+const getColors = (
```
edleeks87, line 10: Suggestion (non-blocking): this is pretty similar to the switch case in the component file, it would be better if we could unify that somehow as it limits the places we need to update if we have additional background colours to add etc. You could move it to a util and do something like:

```
export const getColors = (
  backgroundColor: AdaptiveSidebarProps["backgroundColor"],
) => {
  switch (backgroundColor) {
    case "app":
      return {
        "background-color": "var(--colorsUtilityMajor025)",
        color: "var(--colorsUtilityYin090)"
      };
    case "black":
      return {
        "background-color": "var(--colorsUtilityYin100)",
        color: "var(--colorsUtilityYang100)"
      };
    case "white":
    default:
      return {
        "background-color": "var(--colorsUtilityYang100)"
        color: "var(--colorsUtilityYin090)"
      };
  }
};

export const kebabToCamelCase = (str: string) => {
  return str.replace(/-./g, match => match.charAt(1).toUpperCase());
};

```
Then in the component file you would do 
```
import { getColors, kebabToCamelCase } from "./__internal__/utils"

...

  if (renderAsModal || !largeScreen) {
   const colors = Object.entries(getColors(backgroundColor)).reduce((acc, [key, value]) => {
    acc[kebabToCamelCase(key)] = value;
    return acc;
  }, {} as Record<string, string>);;
   
    return (
      <StyledSidebar open={open} p={0} backgroundColor={backgroundColor}>
        <Box
          height="100%"
          data-role="adaptive-sidebar-content-wrapper"
          {...colours}

File: src/components/adaptive-sidebar/adaptive-sidebar.style.ts [hunk 11]
```diff
@@ -0,0 +1,70 @@
...
+  "backgroundColor" | "borderColor" | "height" | "width"
+> &
+  MarginProps &
+  PaddingProps;
+
+const StyledAdaptiveSidebar = styled(Box)<StyledAdaptiveSidebarProps>`
+  ${({ backgroundColor, borderColor, height, width }) => css`
+    ${getColors(backgroundColor)}
+    ${borderColor &&
+    borderColor !== "none" &&
```
edleeks87, line 44: question: are we supporting any colour here or are should we be limiting it to token values? If it's tokens only we could go with:

```
${borderColor && `border-left: 1px solid var(${borderColor});`}
```

The above assumes that if `none` is passed we don't pass the borderColor down so it would be undefined, it's more than fine to keep the check you have though

edleeks87 (approved):

DipperTheDan (approved):

edleeks87 (approved):

DipperTheDan (approved):

harpalsingh:
Passed a11y review after feedback from Hannah and updates implemented.

DipperTheDan (approved):

edleeks87 (approved):

paulrobinson88 (approved):

PR #7225, opened 2025-02-24
docs: add guidance on importing compound components

Parsium:
### Proposed behaviour

- Add guidance on importing Carbon components that follow the Compound Component pattern, like `Menu` and `FlatTable`. This is to advice projects using micro-frontends to ensure the parent and its subcomponents are imported from the same installed copy of Carbon, to prevent issues with shared context and state.

### Checklist

#### QA


damienrobson-sage (dismissed):

mihai-albu-sage (dismissed):

docs/extending-styles-using-styled-components.mdx
```diff
@@ -6,13 +6,18 @@ import * as ExtendingStylesStories from "./extending-styles-using-styled-compone
...
 
-There may be a rare use case where you need to extend the styles of a Carbon component. This can be done by creating a custom component using the `styled-components` library and extending the component you would like the additional styling applied to.
+> ⚠️ **WARNING:**
+>
+> **[We strongly discourage extending the styles of Carbon components](./?path=/docs/documentation-recommended-practices--docs).**
+>
+> **This should be used as a last resort.**
 
-**We would always recommend this as a last resort as Carbon should already cater to a lot of your use cases**.
+While it is rare, there may be cases where you need to extend the styles of a Carbon component. If you need to, please proceed with caution. We cannot guarantee our components' internal HTML/CSS will remain unchanged in future updates, therefore extending styles could make your application more fragile. If you decide to go ahead, you can use `styled-components` to create a custom component and apply the necessary styles.
```
DobroTora, line 15: I think that if we say that,  it is not a good practice to extend Carbon components with custom styles, it would be maybe better to suggest the channel to go to if there is such case, especially if it is rare, so maybe the team could suggest for example different component that could be the answer to the issue ?

paulrobinson88 (approved):

docs/recommended-practices.mdx
```diff
@@ -0,0 +1,55 @@
...
+  // some additional logic
+  // ...
+  return <Component {...rest} />;
+};
+```
+
+It is also possible to achieve the same by using hooks if you prefer:
+
+```javascript
+const useComponentHook = (customProp, rest) => {
```
DobroTora, line 50: Should we also add a note to be cautious when using certain hooks like useMemo or useCallback as spreading a new object reference ({...props}) every render can cause unnecessary re-renders ?

DobroTora (approved):

DobroTora (approved):

mihai-albu-sage (approved):

paulrobinson88 (approved):

PR #7226, opened 2025-02-24
feat: support server-side rendering

quinnturner:
This PR supports applications to use SSR with React 18 by checking for `window`, `document`, and `navigator` object availability in places that aren't supported during SSR.

I needed these changes to make our admin portal + react-router@7's new routing API work.

Additionally, I added an ESLint to prevent accidental usage of global objects during React rendering. I went with that instead of the restricted syntax lint rule because that lint rule would require checking for `window` existence when we know it exists (in `useEffect`s, etc.).

FE-7103

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -12,7 +12,7 @@ const esmOnlyPackages = [
 
 const isCI = process.env.CI === "true";
 
-const config: Config = {
+const clientConfig: Config = {
   notify: false,
   setupFiles: ["raf/polyfill"],
   testEnvironment: "jsdom",
@@ -20,7 +20,10 @@ const config: Config = {
     "<rootDir>/src/__spec_helper__/__internal__/index.ts",
     "jest-canvas-mock",
   ],
-  testMatch: ["**/?(*.)+(spec|test).[jt]s?(x)"],
+  testMatch: [
+    "**/?(*.)+(spec|test).[jt]s?(x)",
+    "!**/*.server.(spec|test).[jt]s?(x)",
+  ],
   testPathIgnorePatterns: ["node_modules", "lib", "esm"],
   moduleDirectories: ["src", "node_modules"],
   collectCoverage: true,
@@ -45,6 +48,17 @@ const config: Config = {
   moduleNameMapper: {
     "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js",
   },
+};
+
+const serverConfig: Config = {
+  ...clientConfig,
+  testMatch: ["**/*.server.(spec|test).[jt]s?(x)"],
+  testPathIgnorePatterns: ["node_modules", "lib", "esm"],
+  testEnvironment: "node",
+};
+
+const config: Config = {
+  projects: [clientConfig, serverConfig],
 };
 
 export default config;
--- a/src/__internal__/dom/globals.server.test.ts
+++ b/src/__internal__/dom/globals.server.test.ts
@@ -1,0 +1,28 @@
+import { getWindow, getDocument, getNavigator } from "./globals";
+
+describe("getWindow on server", () => {
+  it("should return undefined if window does not exist", () => {
+    expect(typeof window).toBe("undefined");
+    expect(getWindow()).toBeUndefined();
+  });
+});
+
+describe("getDocument on server", () => {
+  it("should return undefined if document does not exist", () => {
+    expect(typeof document).toBe("undefined");
+    expect(getDocument()).toBeUndefined();
+  });
+});
+
+describe("getNavigator on server", () => {
+  it("should return whether the navigator exists", () => {
+    // Node.js 21 introduced support for navigator, so it may or may not exist depending on the version of Node.js
+    if (typeof navigator === "undefined") {
+      // eslint-disable-next-line jest/no-conditional-expect
+      expect(getNavigator()).toBeUndefined();
+    } else {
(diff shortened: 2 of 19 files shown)


src/components/modal/__internal__/modal-manager.ts
```diff
@@ -10,6 +10,10 @@ class ModalManagerInstance {
   private modalList: ModalList;
 
   constructor() {
+    if (typeof window === "undefined") {
```
quinnturner, line 13: `typeof window === "undefined"` is the safe way to check for the existence of `window` in Node.js environments.

jest.config.ts
```diff
@@ -12,15 +12,18 @@ const esmOnlyPackages = [
...
   notify: false,
   setupFiles: ["raf/polyfill"],
   testEnvironment: "jsdom",
   setupFilesAfterEnv: [
     "<rootDir>/src/__spec_helper__/__internal__/index.ts",
     "jest-canvas-mock",
   ],
-  testMatch: ["**/?(*.)+(spec|test).[jt]s?(x)"],
+  testMatch: [
+    "**/?(*.)+(client.spec|client.test|spec|test).[jt]s?(x)",
```
quinnturner, line 24: This might be simplifiable to remove the `client.*`

src/components/modal/__internal__/modal-manager.server.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+});
+
+test("when the addModal method has been called, then the element passed in an attribute should be the topmost element", () => {
+  const cb1 = jest.fn();
+
+  const mockModal1 = "div";
+
+  ModalManager.addModal(mockModal1 as unknown as HTMLElement, cb1);
+
+  expect(ModalManager.isTopmost(mockModal1 as unknown as HTMLElement)).toBe(
```
quinnturner, line 14: I wanted to avoid `null` since that's not a useful test case but the server doesn't have access to `HTMLElement`, so I just passed a `string`.

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -278,7 +280,9 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
           }
 
           /* Styling for safari. Position relative does not work on tr elements on Safari  */
-          ${isSafari(navigator) &&
+          // FIXME: this can cause hydration mismatches during SSR.
+          ${nav &&
+          isSafari(nav) &&
```
DobroTora, line 285: How about using `typeof navigator !== "undefined" && isSafari(navigator) &&` ?

src/__internal__/dom/globals.server.test.ts
```diff
@@ -0,0 +1,28 @@
+import { getWindow, getDocument, getNavigator } from "./globals";
+
+describe("getWindow on server", () => {
```
nineteen88, line 3: comment(non-blocking): for test consistency we have moved away from single test describe blocks

We will likely just tidy this up for consistency once it's in though, so don't worry about it just flagging for future reference

src/components/flat-table/flat-table-row/flat-table-row.style.ts
```diff
@@ -278,7 +280,9 @@ const StyledFlatTableRow = styled.tr<StyledFlatTableRowProps>`
           }
 
           /* Styling for safari. Position relative does not work on tr elements on Safari  */
-          ${isSafari(navigator) &&
+          // FIXME: this can cause hydration mismatches during SSR.
```
nineteen88, line 283: comment: we should probably add a JIRA ticket or Github Issue to this so that it doesn't get missed

src/components/icon/icon.style.ts
```diff
@@ -164,16 +167,18 @@ const StyledIcon = styled.span<StyledIconProps & StyledIconInternalProps>`
           font-size: ${iconConfig.iconSize[fontSize]};
           line-height: ${iconConfig.iconSize[fontSize]};
         `}
-
-        ${type === "services" &&
-        browserTypeCheck(window) &&
+        // FIXME: this can cause hydration mismatches during SSR.
```
nineteen88, line 170: comment: we should probably add a JIRA ticket or Github Issue to this so that it doesn't get missed

src/hooks/__internal__/useScrollBlock/scroll-block-manager.server.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  const id1 = "id1";
+  scrollBlockManager.registerComponent(id1);
+
+  it("should set isBlocked to true after registering at least one component", () => {
+    expect(scrollBlockManager.isBlocked()).toBe(true);
+  });
+});
+
+describe("ScrollBlockManager behavior with registerComponent and unregisterComponent", () => {
+  it("isBlocked should still return true after unregisterComponent is called with only one of the registered IDs ('id1')", () => {
```
nineteen88, line 14: comment(non-blocking): for consistency we have tried to ensure describe/it tests are written with continuous prose

Again, we will likely just pick this up once it's in and fix for consistency

nineteen88 (approved):

edleeks87 (approved):

quinnturner:
I've rebased onto the latest `master` and squashed all commits into one following semantic release.

Issue #5094, opened 2022-04-27
FlatTable with StickyHead & stickyFooter : ActionPopover detached  when scrolling


hagad:
### Current behaviour

Build FlatTable with hasStickyHead ,hasStickyFooter and ActionPopover 

The ActionPopover component is detached from the list when scrolling

### Expected behaviour

 the ActionPopover should remain stick to the row

### CodeSandbox or Storybook URL

https://codesandbox.io/s/elegant-banzai-qnxvoh?file=/src/index.js:2089-2102

### Carbon Version

106.0.2

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

samtjo:
Hi @hagad - is this the right codesandbox for this issue? We are unable to replicate the issue using this sandbox, and it looks like it's demonstrating a different issue with the pager.

hagad:
@samtjo  I added more information in the sandbox :https://codesandbox.io/s/elegant-banzai-qnxvoh?file=/src/index.js
You can find also one screenshot. 
![image](https://user-images.githubusercontent.com/17086956/168081854-4927537d-4592-4409-a6bb-e3a744d56199.png)
slack discussion : https://sageone.slack.com/archives/C0D6DD3N2/p1651074785462089

Parsium:
FE-5176

harpalsingh:
FYI looking at the sandbox link, its attached to the row, but that in itself causes an issue. So @hagad would you like a chat around this issue?

Arkalex:
Still happening

PR #7227, opened 2025-02-25
fix(flat-table): action-popover detached when scrolling flat table with sticky head and footing

mihai-albu-sage:
fixes: #5094

### Proposed behaviour

Scrolling inside the `FlatTable` which has `hasStickyHead` and `hasStickyFooter` has been disabled if the components `ActionPopover`, `MultiActionButton`, `Date`, `DateRange` and  `SplitButton` are inside the table and are opened.

### Current behaviour

If we have an `ActionPopover` component opened inside the `FlatTable` scrolling is enabled and the popover-container is detached.

https://github.com/user-attachments/assets/fc5627e7-3501-4e6e-a806-c83a3b832c74

### Checklist

#### QA

### Testing instructions
1. Go to the story: flat-table-test--flat-table-with-sticky-head-and-footer
2. Scroll to the middle of the table
3. Open one of the components ActionPopover, Date, etc.
4. Try to scroll
5. Scrolling while these components are opened is prevented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [151.2.2](https://github.com/Sage/carbon/compare/v151.2.1...v151.2.2) (2025-03-06)
+
+### Bug Fixes
+
+* **date-input:** update documentation ([df5a8ba](https://github.com/Sage/carbon/commit/df5a8bab0b6e34b9f7e2183d51e32328cee6f695))
+
 ## [151.2.1](https://github.com/Sage/carbon/compare/v151.2.0...v151.2.1) (2025-03-04)
 
 ### Bug Fixes
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "carbon-react",
-  "version": "151.2.1",
+  "version": "151.2.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "carbon-react",
-      "version": "151.2.1",
+      "version": "151.2.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@floating-ui/dom": "~1.2.9",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "carbon-react",
-  "version": "151.2.1",
+  "version": "151.2.2",
   "description": "A library of reusable React components for easily building user interfaces.",
   "files": [
     "lib",


src/components/action-popover/action-popover.component.tsx
```diff
@@ -108,6 +112,8 @@ export const ActionPopover = forwardRef<
     const [guid] = useState(createGuid());
     const buttonRef = useRef<HTMLDivElement>(null);
     const menu = useRef<HTMLUListElement>(null);
+    const { isInFlatTable } =
```
DobroTora, line 115: Shall we consider the scenario, where isInFlatTable is undefined ? If we do then maybe we should ensure that the context always provides a defined value. When creating a context in FlatTableContext we can specify a default value for isInFlatTable, so even if the context is not specifically set by a parent component, it has a default.

src/components/flat-table/flat-table.test.tsx
```diff
@@ -1444,3 +1451,88 @@ test("should set the title correctly", () => {
     "this is a title",
   );
 });
+
+test("when an ActionPopover is opened inside the FlatTable, it will have the background disabled to prevent scrolling in the table", async () => {
```
DobroTora, line 1455: Shall we also include a test for the actual Popover behavior, like whether it responds correctly, when there are clicks outside for example ?

src/components/action-popover/action-popover.component.tsx
mihai-albu-sage: Refactored to provide value `false` to `isInFlatTable` as a default value.

src/components/flat-table/flat-table.test.tsx
mihai-albu-sage: I've updated the test to make sure the `ActionPopoverItem` is visible when we first click the button and if clicking elsewhere in the document the `ActionPopover` closes and the backdrops is no longer in the document.

DobroTora (dismissed):
As long as we are not mocking FlatTableContext to control isInFlatTable value for a testing purposes I am happy to approve it.

src/components/flat-table/flat-table-test.stories.tsx
```diff
@@ -870,3 +879,188 @@ FlatTableThemesWithStickyHead.parameters = {
...
+                </FlatTableRow>
+              );
+            })}
+        </FlatTableBody>
+      </FlatTable>
+    </Box>
+  );
+};
+
+FlatTableWithStickyHeadAndFooter.parameters = {
```
edleeks87, line 1063: comment: I don't think there's any value in capturing this if none of the popovers are open when it renders as it's just capturing the static table etc

edleeks87 (commented):
question: would it be worth doing the same thing for `PopoverContainer`?

DobroTora (approved):
After addresing @edleeks87 comment and changes I am happy to approve.

src/components/flat-table/flat-table-test.stories.tsx
mihai-albu-sage: Removed the capturing.

mihai-albu-sage:
> question: would it be worth doing the same thing for `PopoverContainer`?

Add the same logic into the `PopoverContainer` as well.

DobroTora (approved):

edleeks87 (approved):

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -195,6 +199,17 @@ export const DatePicker = ({
     }, 0);
   };
 
+  const { isInFlatTable, setHasOpenDatePicker } =
+    useContext<FlatTableContextProps>(FlatTableContext);
+
+  useEffect(() => {
+    // setDatePickerActive is optional but is initialized as an empty function within the context
```
edleeks87, line 206: comment: I think you missed updating this comment from `setDatePickerActive` to `setHasOpenDatePicker`

src/components/date/__internal__/date-picker/date-picker.component.tsx
```diff
@@ -195,6 +199,17 @@ export const DatePicker = ({
...
+    /* istanbul ignore else */
+    if (setHasOpenDatePicker) {
```
edleeks87, line 208: comment (non-blocking): if it has a default value then is there a need for the guard or the ignore comment here?

I think it might be better to not have a default value for it and use nullish coalescence to avoid the need for the ignore comment. We probably don't need to include `isInFlatTable` in the  argument as we know if setHasOpenDatePicker exists and can be called the date component is inside a flat table

```
useEffect(() => {
   setHasOpenDatePicker?.(open);
 }, [open, setHasOpenDatePicker]);

mihai-albu-sage: Thanks for the suggestion, it makes much more sense to have it like this. Refactored.

src/components/date/__internal__/date-picker/date-picker.component.tsx
mihai-albu-sage: Comment has been removed completely since it's no longer needed.

edleeks87 (dismissed):

DobroTora (dismissed):

DobroTora (dismissed):

edleeks87 (approved):

DobroTora (approved):

paulrobinson88 (approved):

PR #7259, opened 2025-03-14
feat(use-character-count): make hook public

### LINKED ISSUE #7242: Textarea characters remaining (CRLF option)
### Description

On Sage 200 UKI we are a product which has always run on Windows and are porting all of our forms to the web.

When users enter text in a text area, we convert line endings to CRLF in our back end just before saving to the database. This is required so that our Windows Forms based desktop client is able to display new lines correctly.

In our web front end we have yup validations which convert \n to \r\n before doing max length checks to ensure that user input will fit in our database without errors.

As web browsers work with just `\n`, the characters remaining displayed on the Carbon textarea ends up telling our users that they have characters remaining, while our yup validations which convert to CRLF tell the user they have exceeded the limit.

To make these consistent, we'd like to be able to either be able to optionally tell the Textarea component to do a conversion to CRLF when calculating characters remaining, or to be able to pass a custom function to calculate characters remaining if you think that would be better.

### Suggested solution

Either;

- A prop to optionally tell the Carbon Textarea to convert to CRLF before calculating remaining characters
- A prop that allows us to pass a custom function to calculate characters remaining

### Demo URL

_No response_

### Alternatives

Our alternative would be not showing characters remaining but this would be poor from a UX perspective.

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #7242

### Proposed behaviour

- `useCharacterCount` is a public hook. 
-  Add `aria-describedby` prop to text based input props, which gets added to the hint and error message description. 
- Deprecate `ariaDescribedBy` prop in text based components. 

### Current behaviour

- `useCharacterCount` is an internal hook. 
- `ariaDescribedBy` prop is available in text based components. 

### Checklist

#### QA


src/__internal__/input/input.component.tsx
```diff
@@ -17,7 +17,7 @@ export interface CommonInputProps
   extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "type"> {
   /* The default value alignment on the input */
   align?: "right" | "left";
-  /** The id of the element that describe the input. */
+  /** Appends value to the internal aria-describedby array. */
```
nuria1110, line 20: I have appended the aria-describedby prop value to the current internal array `combinedAriaDescribedBy` as I believe the hint and validation messages should still be included in the accessible description (if present) even if consumers supply an additional one.

I'm happy to discuss this if any reviewers disagree or prefer an alternative method. Also open to suggestions for the prop description as I'm not 100% the current wording is suitable.

src/hooks/useCharacterCount/use-character-count.mdx
```diff
@@ -0,0 +1,39 @@
+import { ArgTypes, Meta, Canvas } from "@storybook/addon-docs";
+
+import * as useCharacterCountStories from "./use-character-count.stories";
+
+<Meta title="Documentation/Hooks/useCharacterCount"/>
+
+# useCharacterCount
+
+`useCharacterCount` is a custom React hook that provides a character counter for input fields.
```
edleeks87, line 9:
```suggestion
`useCharacterCount` is a custom React hook that provides a character counter for text based input fields. 
```

non-blocking fyi

src/components/textarea/textarea.component.tsx
```diff
@@ -137,6 +137,7 @@ export const Textarea = React.forwardRef(
   (
     {
       "aria-labelledby": ariaLabelledBy,
+      ariaDescribedBy: ariaDescribedByProp,
```
edleeks87, line 140: Comment: I realise this is coming from the internal Input interface, but I think as we're documenting it as matching the casing for the HTML element interface (in the new stories below) we should ensure the prop added to Textbox and Textarea use it as well rather than camel case. Obviously this will require updating the Input interface to match but if we don't do it now we'll have to make a breaking change down the line to update it as we're exposing the prop now. However, if this interface is already exposed elsewhere then lets leave it for now as looking through the code base we've been wonderfully inconsistent with the naming etc so we'll need to make a breaking change at some point to consolidate anyway

src/hooks/useCharacterCount/use-character-count.mdx
edleeks87: Might also be worth adding that this is not supported with TextEditor atm (again non-blocking)

src/__internal__/input/input.component.tsx
edleeks87: I think it makes sense to do this I think we could remove the part about an array as it's very implementation specific and this will appear in the prop table: you could probably leave it as it was or adds something about how it's combined with the error and input hint text etc

src/components/textarea/textarea.component.tsx
nuria1110: The `ariaDescribedBy` prop was unintentionally being extended from the Input props so it is already present in the interface/props table though it was not getting passed anywhere, so to avoid a possible breaking change I've "fixed" it by now actually passing it. 
I agree we should try to be consistent and use the HTML format though its whether we think removing the existing prop is a BC or not - happy to see what the other reviewer thinks.

src/components/textarea/textarea.component.tsx
```diff
@@ -358,7 +360,7 @@ export const Textarea = React.forwardRef(
         <Input
           aria-invalid={!!error}
           aria-labelledby={ariaLabelledBy}
-          ariaDescribedBy={combinedAriaDescribedBy}
+          aria-describedby={combinedAriaDescribedBy}
```
nuria1110, line 363: I've used the new "aria-describedby" prop here since its what should be used from now on, the old `ariaDescribedBy` will still be spread through rest to input, so its behaviour hasn't changed from this fix.

src/components/textarea/textarea.component.tsx
```diff
@@ -360,7 +360,7 @@ export const Textarea = React.forwardRef(
...
```
nuria1110, line 363: I've used the new "aria-describedby" prop here since its what should be used from now on, the old ariaDescribedBy will still be spread through rest to input, so its behaviour hasn't changed from this fix.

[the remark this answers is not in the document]
Parsium: **question**: since consumers might be using `ariaDescribedBy`, without realising it overrides any descriptions we set internally, would it be worth us including the prop in the `combinedAriaDescribedBy` instead of spreading it?

src/components/textbox/textbox.component.tsx
```diff
@@ -286,7 +288,7 @@ export const Textbox = React.forwardRef(
           align={align}
           aria-invalid={!!error}
           aria-labelledby={ariaLabelledBy}
-          ariaDescribedBy={combinedAriaDescribedBy}
+          aria-describedby={combinedAriaDescribedBy}
```
Parsium, line 291: **question**: similarly to `Textarea`, should we consider adding the `ariaDescribedBy` to `combinedAriaDescribedBy`, if teams are using this prop extensively at the moment?

nuria1110: I'm okay with adding it into the array, though I wonder if its worth doing this if it is being deprecated and we want to discourage it's use?

src/__internal__/input/input.component.tsx
```diff
@@ -17,8 +17,13 @@ export interface CommonInputProps
   extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "type"> {
   /* The default value alignment on the input */
   align?: "right" | "left";
-  /** The id of the element that describe the input. */
+  /**
```
edleeks87, line 20: comment: I think it's worth following the same pattern as we do with other deprecations and using `Logger.deprecate` here with a flag to prevent it called twice and a test etc

edleeks87: It might need to be in Textbox and Textarea if we go with @Parsium's suggestion about passing it into the combined array thinking about it

Parsium (dismissed):

edleeks87 (dismissed):

stephenogorman (dismissed):

edleeks87 (approved):

Parsium (approved):

paulrobinson88 (approved):

Issue #7251, opened 2025-03-10
Time component's onChange event not giving the correct values for hours/minutes


danielthomas2:
### Description

The Time component's onChange event handler is giving the wrong values when we programmatically set the "value" prop.

I would like the Time component's hour and minutes value to be zero padded. To achieve this, I am using the formattedHours and formattedMinutes we get from the onBlur event, and setting the value on the state. However, if I change the hour value, the onChange event is giving the wrong value for the minutes value, even though it was only the hours value that has changed.

The onChange event of the Time component seems to return one of the values from the internal state (**inputValues**), rather than from the actual input fields for hours and minutes. The **inputValues** state holds the internal values, but this state is only updated inside the onChange event handler and not when the **value** prop is changed. This causes issues when the time component's value is updating programmatically.

I had raised an earlier issue (https://github.com/Sage/carbon/issues/7183) but it seems the problem still persists.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-2lvhtuns?file=package.json,src%2FApp.tsx,package-lock.json

### Steps to reproduce

Consider the following scenario. 

- The user changes the time to **"1:3"**. The displayed value becomes **1:3** and **inputValues = [1, 3].**
- If the **value** prop for the Time component is updated programmatically to **{ hours: "01", minutes: "03" }**, the displayed value updates to **01:03** as expected. However, **inputValues** still contains the old values, i.e., **[1, 3]**.
- The next time the user modifies the time **value**, e.g., changing the hour from **01** to **2**, the onChange event will incorrectly return the following:
  - Correct hour: **2** (from the textbox input change)
  - Incorrect minutes: **3** (from the outdated **inputValues** state, rather than **03** as displayed).

I have included another example sandbox, with the same issue where we are still unable to display the zero padded times correctly. 

There is also an example where you set the hours/minutes programmatically by setting the values in the two below textboxes and clicking the button. While the value is displayed correctly on the time component, changing one of the hours and minutes will show that the onChange is giving the wrong old values from **inputValues** which haven't been updated when the **value** was updated.

### Carbon version

150.1.0

### React version

v18

### Relevant browsers

Chrome

### Relevant OSs

Windows

### Confidentiality

- [x] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
You have control of the input you can use a callback in the setState to conditionally update the input that requires updating whilst persisting the other

[...]

PR #7260, opened 2025-03-15
fix(time): onChange event not giving the correct values for hours/minutes

mihai-albu-sage:
fixes: #7251

### Proposed behaviour

A `useEffect` has been added to keep `inputValues` and `formattedInputValues` in sync with the received `value` when this is changed. This `useEffect` will run only when the component is controlled and the `value` is updated. This ensures that the `onChange` event contains the correct values. The `TimeInputEvent` has also been added to the exported types. 

https://github.com/user-attachments/assets/9c24d419-0109-48c7-af94-fed9b8160896

### Current behaviour

In the current behavior, when the initial `value` is changed, the `inputValues` and `formattedInputValues` are not updated causing them to return wrong values for the `onChange` event.

https://github.com/user-attachments/assets/ba01c7fc-f615-4e09-bc19-28d936cce6c4

### Checklist

#### QA


src/components/time/time.component.tsx
```diff
@@ -170,6 +171,35 @@ const Time = React.forwardRef<TimeHandle, TimeProps>(
...
+        const formattedHours = hourValue.length
+          ? hourValue.padStart(2, "0")
+          : hourValue;
+        const formattedMinutes = minuteValue.length
+          ? minuteValue.padStart(2, "0")
+          : minuteValue;
+
+        setFormattedInputValues([formattedHours, formattedMinutes]);
+      }
+      // This useEffect ensures inputValues and formattedInputValues stay in sync with the received value (hourValue and minuteValue).
```
edleeks87, line 197: comment: both `value` and `onChange` are required props so I think we're safe to assume it won't be used uncontrolled

src/components/time/time-test.stories.tsx
```diff
@@ -100,3 +104,83 @@ export const HintTextAlignment: Story = ({ ...args }) => {
   );
 };
 HintTextAlignment.storyName = "Hint Text Alignment";
+
+export const WithValueModifiers: Story = () => {
```
edleeks87, line 108: Comment: Good idea adding this test story, I don't think there's any value in capturing it in chromatic at this point as we currently don't support play functions so there's nothing that isn't already captured by other stories etc

src/components/time/time.component.tsx
mihai-albu-sage: `inputValues` has been added to the dependency array.

src/components/time/time-test.stories.tsx
mihai-albu-sage: Removed capturing in chromatic.

edleeks87 (approved):

DipperTheDan (approved):

dariusbercea-sage (approved):

PR #7268, opened 2025-03-21
fix:  ensure there is no flash of reduce motion text in Loader components when page is reloaded

### LINKED ISSUE #7240: Loading text flashes on screen before <Loader /> is rendered
### Description

When I refresh the page, the 'Loading' text flashes on the screen before the <Loader /> component is rendered

### Reproduction

Messaged with video privately

### Steps to reproduce

_No response_

SBS-124044

### Suggested solution

_No response_

### Carbon version

144.9.11

### React version

v17

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

DobroTora:
fixes #7240

### Proposed behaviour

https://github.com/user-attachments/assets/73d222ff-f8be-45b2-a88d-0fbce17cdcf2

https://github.com/user-attachments/assets/91364182-0fc0-4ae9-96e0-14542e4acb34

Initialises `useMediaQuery` as `undefined` to prevent reduced motion `loading` text from rendering when browser motion preferences are yet to be detected when there is latency on the network or a slow page reload

### Current behaviour

When there is network latency they reduced motion `loading` text flashes on screen as the browser motion preference is yet to be determined and `useMediaQuery` has an initial value of false.

### Checklist

#### QA

### Additional context

In useMediaQuery hook useLayoutEffect has been used to make use the matchMedia value is set synchronously before the browser painting. 

### Testing instructions

Pick any Loader story and set the browser network to 4G or worse and refresh the page, in master you will see the flash of the `loading` text. In this PR you shouldn't see it

### DIFF
--- a/src/components/loader-bar/loader-bar.component.tsx
+++ b/src/components/loader-bar/loader-bar.component.tsx
@@ -21,9 +21,13 @@ export interface LoaderBarProps
 export const LoaderBar = ({ size = "medium", ...rest }: LoaderBarProps) => {
   const l = useLocale();
 
-  const reduceMotion = !useMediaQuery(
+  const allowMotion = useMediaQuery(
     "screen and (prefers-reduced-motion: no-preference)",
   );
+
+  if (allowMotion === undefined) {
+    return null;
+  }
 
   return (
     <StyledLoader
@@ -32,12 +36,12 @@ export const LoaderBar = ({ size = "medium", ...rest }: LoaderBarProps) => {
       {...tagComponent("loader-bar", rest)}
       {...filterStyledSystemMarginProps(rest)}
     >
-      {reduceMotion ? (
-        l.loader.loading()
-      ) : (
+      {allowMotion ? (
         <StyledLoaderBar data-role="outer-bar" size={size}>
           <InnerBar data-role="inner-bar" size={size} />
         </StyledLoaderBar>
+      ) : (
+        l.loader.loading()
       )}
     </StyledLoader>
   );
--- a/src/components/loader-bar/loader-bar.test.tsx
+++ b/src/components/loader-bar/loader-bar.test.tsx
@@ -3,43 +3,59 @@ import { render, screen } from "@testing-library/react";
 import LoaderBar from "./loader-bar.component";
 import useMediaQuery from "../../hooks/useMediaQuery";
 
-jest.mock("../../hooks/useMediaQuery", () => {
-  return {
-    __esModule: true,
-    default: jest.fn().mockReturnValue(true),
-  };
+jest.mock("../../hooks/useMediaQuery", () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
+  typeof useMediaQuery
+>;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockUseMediaQuery.mockReturnValue(true);
+});
+
+afterAll(() => {
+  jest.restoreAllMocks();
 });
 
 test("renders", () => {
   render(<LoaderBar />);
-
   const loaderBar = screen.getByRole("progressbar", { name: "Loading" });
   expect(loaderBar).toBeVisible();
 });
 
+test("uses default size when size is undefined", async () => {
+  render(<LoaderBar />);
+
+  const outerBar = await screen.findByTestId("outer-bar");
+  const innerBar = await screen.findByTestId("inner-bar");
+
+  expect(outerBar).toHaveStyle("height: 8px");
+  expect(innerBar).toHaveStyle("height: 8px");
+});
+
 test("renders with custom data role and element", () => {
   render(<LoaderBar data-role="foo" data-component="bar" data-element="baz" />);
-
   const loaderBar = screen.getByTestId("foo");
   expect(loaderBar).toHaveAttribute("data-role", "foo");
   expect(loaderBar).toHaveAttribute("data-element", "baz");
 });
 
 test("renders alternative loading text if user prefers reduced motion", () => {
-  const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
-    typeof useMediaQuery
-  >;
   mockUseMediaQuery.mockReturnValueOnce(false);
-
   render(<LoaderBar />);
-
   const loaderBar = screen.getByRole("progressbar", { name: "Loading" });
   expect(loaderBar).toHaveTextContent("Loading");
 });
 
 /* Styling test for coverage */
-it.each([
+test.each([
+  [undefined, "8px"],
   ["small", "4px"],
+  ["medium", "8px"],
   ["large", "16px"],
 ] as const)(
   "when size is %s, both the inner and outer bars have the correct height",
--- a/src/components/loader-spinner/loader-spinner.component.tsx
+++ b/src/components/loader-spinner/loader-spinner.component.tsx
@@ -65,9 +65,13 @@ export const LoaderSpinner = ({
 }: LoaderSpinnerProps) => {
   const locale = useLocale();
 
-  const reduceMotion = !useMediaQuery(
+  const allowMotion = useMediaQuery(
     "screen and (prefers-reduced-motion: no-preference)",
   );
+
+  if (allowMotion === undefined) {
+    return null;
+  }
 
   const isLabelDark = variant !== "inverse" && variant !== "gradient-white";
 
@@ -100,7 +104,6 @@ export const LoaderSpinner = ({
     if (animationTime) {
       return animationTime;
     }
-
     return isGradientVariant ? 2 : 1;
   };
 
@@ -111,9 +114,7 @@ export const LoaderSpinner = ({
       {...tagComponent("loader-spinner", rest)}
       {...filterStyledSystemMarginProps(rest)}
     >
-      {reduceMotion ? (
-        renderSpinnerLabel
-      ) : (
+      {allowMotion ? (
         <>
           <StyledSpinnerCircleSvg
             role="presentation"
@@ -128,6 +129,7 @@ export const LoaderSpinner = ({
             <circle data-role="outer-arc" />
             <circle data-role="inner-arc" />
           </StyledSpinnerCircleSvg>
+
           {showSpinnerLabel ? (
             renderSpinnerLabel
           ) : (
@@ -140,6 +142,8 @@ export const LoaderSpinner = ({
             </Typography>
           )}
         </>
+      ) : (
+        renderSpinnerLabel
       )}
     </StyledSpinnerWrapper>
   );
--- a/src/components/loader-spinner/loader-spinner.test.tsx
+++ b/src/components/loader-spinner/loader-spinner.test.tsx
@@ -1,25 +1,58 @@
 import React from "react";
-import { render, screen } from "@testing-library/react";
+import { render, screen, within } from "@testing-library/react";
 import { LoaderSpinner } from ".";
 import useMediaQuery from "../../hooks/useMediaQuery";
 import {
   LOADER_SPINNER_SIZE_PARAMS,
   LoaderSpinnerSizes,
   LOADER_SPINNER_SIZES as sizes,
+  LOADER_SPINNER_VARIANTS,
 } from "./loader-spinner.config";
 
-jest.mock("../../hooks/useMediaQuery", () => {
-  return {
-    __esModule: true,
-    default: jest.fn().mockReturnValue(true),
-  };
+jest.mock("../../hooks/useMediaQuery", () => ({
+  __esModule: true,
+  default: jest.fn(),
+}));
+
+const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
+  typeof useMediaQuery
+>;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockUseMediaQuery.mockReturnValue(true);
 });
+
+afterAll(() => {
+  jest.restoreAllMocks();
+});
+
+const outerArcStrokeValues: Record<
+  (typeof LOADER_SPINNER_VARIANTS)[number],
+  string
+> = {
+  action: "var(--colorsActionMajor150)",
+  neutral: "var(--colorsSemanticNeutral200)",
+  inverse: "var(--colorsActionMajorYang100)",
+  "gradient-grey": "#0000001A",
+  "gradient-white": "var(--colorsActionMajorYang100)",
+};
 
 test("the component wrapper renders with a role of status", () => {
   render(<LoaderSpinner />);
   const wrapperElement = screen.getByRole("status");
 
   expect(wrapperElement).toBeVisible();
+});
+
+test("should return null and render nothing when useMediaQuery returns undefined", () => {
+  mockUseMediaQuery.mockReturnValueOnce(undefined);
+
+  render(<LoaderSpinner />);
+
+  const wrapperElement = screen.queryByRole("status");
+
+  expect(wrapperElement).not.toBeInTheDocument();
 });
 
 test("the svg spinner circle wrapper renders", () => {
@@ -69,9 +102,13 @@ describe("when custom props are passed", () => {
 
   it("should override the visually hidden label text when the 'spinnerLabel' prop is passed a custom string value", () => {
     render(<LoaderSpinner spinnerLabel="bar" showSpinnerLabel={false} />);
-    const hiddenLabelElement = screen.getByTestId("hidden-label");
 
-    expect(hiddenLabelElement).toHaveTextContent("bar");
+    const status = screen.getByRole("status");
+    const hiddenLabelElement = within(status).getByText("bar");
+    const visibleLabelElement = screen.queryByTestId("visible-label");
+
+    expect(hiddenLabelElement).toBeInTheDocument();
+    expect(visibleLabelElement).not.toBeInTheDocument();
   });
 
   it("when the 'size' prop is passed as 'extra-small' the component wrapper has a flex-direction of row", () => {
@@ -177,28 +214,27 @@ describe("when custom props are passed", () => {
     expect(visibleLabelElement).not.toBeInTheDocument();
   });
 
-  it("when 'showSpinnerLabel' is `false` a visually hidden alternative label is rendered in the live region container", () => {
+  it("when 'showSpinnerLabel' is `false`, a visually hidden alternative label is rendered in the live region container", () => {
     render(<LoaderSpinner showSpinnerLabel={false} />);
+
     const wrapperElement = screen.getByRole("status");
     const hiddenLabelElement = screen.getByTestId("hidden-label");
+    const visibleLabelElement = screen.queryByTestId("visible-label");
 
     expect(wrapperElement).toHaveTextContent("Loading...");
     expect(hiddenLabelElement).toBeInTheDocument();
+    expect(visibleLabelElement).not.toBeInTheDocument();
   });
 
-  it.each([
-    ["action", "var(--colorsActionMajor150)"],
-    ["neutral", "var(--colorsSemanticNeutral200)"],
-    ["inverse", "var(--colorsActionMajorYang100)"],
-    ["gradient-grey", "#0000001A"],
-    ["gradient-white", "var(--colorsActionMajorYang100)"],
-  ] as const)(
-    "when the 'variant' prop is passed as `%s` the correct outer arc stroke (color) is rendered",
-    (variants, strokeValues) => {
-      render(<LoaderSpinner variant={variants} />);
+  it.each(LOADER_SPINNER_VARIANTS)(
+    "when the 'variant' prop is passed as '%s' the correct outer arc stroke (color) is rendered",
+    (variant) => {
+      render(<LoaderSpinner variant={variant} />);
       const outerArcSvgElement = screen.getByTestId("outer-arc");
 
-      expect(outerArcSvgElement).toHaveStyle(`stroke: ${strokeValues}`);
+      expect(outerArcSvgElement).toHaveStyle(
+        `stroke: ${outerArcStrokeValues[variant]}`,
+      );
     },
   );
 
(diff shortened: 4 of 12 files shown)


src/components/loader-bar/loader-bar.test.tsx
```diff
@@ -3,51 +3,57 @@ import { render, screen } from "@testing-library/react";
...
-  >;
   mockUseMediaQuery.mockReturnValueOnce(false);
-
   render(<LoaderBar />);
-
   const loaderBar = screen.getByRole("progressbar", { name: "Loading" });
   expect(loaderBar).toHaveTextContent("Loading");
 });
 
-/* Styling test for coverage */
```
nineteen88, line 40: suggestion(non-blocking): keep the comment

It's useful for discerning which tests we've had to put in purely to meet coverage but should be removed at a later date once we've done the work to merge the coverage reports

src/components/loader-bar/loader-bar.test.tsx
```diff
@@ -3,51 +3,57 @@ import { render, screen } from "@testing-library/react";
...
 it.each([
```
edleeks87, line 44: comment: this should be `test` instead of `it` as there is no parent `describe` etc

edleeks87: I'm not sure it needs to be async either as you are mocking the value of the `useMediaQuery` hook

src/components/loader-spinner/loader-spinner.test.tsx
```diff
@@ -302,9 +326,10 @@ describe("when custom props are passed", () => {
 
   it.each([1, 2, 3, 5, 10])(
     "when 'animationTime' is passed as `%s` the animation duration is the correct time",
-    (animationTimes) => {
+    async (animationTimes) => {
```
edleeks87, line 329: comment: I could be wrong on this but I don't think we need the tests to be async as we're mocking the hook etc. It's not critical but it will make the test suite run slower as it `awaits` so if we don't need them to be async I think it's better to not have them be etc

src/components/loader/loader.test.tsx
```diff
@@ -30,13 +44,21 @@ test("throws a deprecation warning if the 'aria-label' prop is set", () => {
...
+});
+
+test("when the user disallows animations or their preference cannot be determinded, alternative loading text is rendered", () => {
   render(<Loader />);
 
   expect(screen.getByText("Loading")).toBeVisible();
 });
 
-test("when the user disallows animations or their preference cannot be determined, the provided `aria-label` is rendered", () => {
+test("when the user disallows animations or their preference cannot be determinded, the provided `aria-label` is rendered", () => {
```
edleeks87, line 61:
```suggestion
test("when the user disallows animations or their preference cannot be determined, the provided `aria-label` is rendered", () => {
```

src/components/loader/loader.test.tsx
```diff
@@ -56,16 +78,12 @@ test("when the user disallows animations or their preference cannot be determine
...
   beforeEach(() => {
-    const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
-      typeof useMediaQuery
-    >;
-    mockUseMediaQuery.mockReturnValueOnce(true);
+    mockUseMediaQuery.mockReturnValue(true);
   });
-
-  test("renders three square animation", () => {
+  test("renders three square animation", async () => {
```
edleeks87, line 83: comment: same one about whether this needs to be async

src/hooks/useMediaQuery/useMediaQuery.ts
```diff
@@ -1,9 +1,9 @@
 import React from "react";
 
-export default function useMediaQuery(queryInput: string): boolean {
+export default function useMediaQuery(queryInput: string): boolean | undefined {
   const query = queryInput.replace(/^@media( ?)/m, "");
 
-  const [match, setMatch] = React.useState(() => false);
+  const [match, setMatch] = React.useState<boolean | undefined>(undefined);
```
edleeks87, line 6: comment (non-blocking): I wonder if `null` is more appropriate here. Happy to defer to the other reviewer I'm leaning towards what you have being fine but want to canvas opinion etc

edleeks87 (commented):
Just some minor comments from me @DobroTora, overall really good 👍 . Wit respect to the commit message the following will be added to release notes and isn't very clear to someone reading them what change has occurred etc:

```
fix(loader, bar-loader, loader-spinner, loader-star): text hidden when loader is displayed
```
 
I think something like this might be more appropriate:

```
fix: ensure there is no flash of reduce motion text in Loader components when page is reloaded

File: src/components/loader/loader.test.tsx (hunk 4 above)
Comment (nineteen88) [inline src/components/loader/loader.test.tsx:61]: suggestion: fix typo of determined

```suggestion
test("when the user disallows animations or their preference cannot be determined, the provided `aria-label` is rendered", () => {
```

src/components/loader-bar/loader-bar.component.tsx
```diff
@@ -21,10 +21,16 @@ export interface LoaderBarProps
 export const LoaderBar = ({ size = "medium", ...rest }: LoaderBarProps) => {
   const l = useLocale();
 
-  const reduceMotion = !useMediaQuery(
+  const noPreference = useMediaQuery(
```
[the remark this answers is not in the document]
DobroTora: @edleeks87  I thought   `const reduceMotion = !noPreference;` is guard close to prevent rendering until the media query result is known ?

src/hooks/useMediaQuery/useMediaQuery.ts
DobroTora: @edleeks87 that was my first idea to be honest. As `matchMedia` will eventually return boolean value `undefined` seemed to be a best "placeholder" in this context. Also it is idiomatic for TS and React for "not set yet", which exactly what is happening here.

src/components/loader-bar/loader-bar.component.tsx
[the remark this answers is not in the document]
edleeks87: The guard is still there, the rest of the logic is just flipped to prevent the need for storing the negated value returned by the hook in a second variable

edleeks87 (approved):

src/components/loader/loader.test.tsx
```diff
@@ -30,6 +44,34 @@ test("throws a deprecation warning if the 'aria-label' prop is set", () => {
...
+
+test("does not render anything when `reduceMotion` is undefined", () => {
+  mockUseMediaQuery.mockReturnValueOnce(undefined);
+
+  render(<Loader />);
+
+  expect(screen.queryByTestId("loader")).not.toBeInTheDocument();
+});
+
 test("when the user disallows animations or their preference cannot be determined, alternative loading text is rendered", () => {
```
nineteen88, line 75: comment: although this test hasn't been changed, it's always passing

Following the updates I would expect this test to fail when the user does not have reduced motion enabled. It seems that it's finding the visually hidden component with the `getByText("Loading")` instead of the "Loading" text.

We will need to update this test so that it's more specific to what it should be checking. The following should resolve the issues:

```
test("when the user disallows animations, alternative loading text is rendered", () => {
  mockUseMediaQuery.mockReturnValueOnce(false);
  render(<Loader />);
  
  expect(screen.queryByTestId("hidden-label")).not.toBeInTheDocument();
  expect(screen.getByText("Loading")).toBeVisible();
});
```

src/components/loader/loader.test.tsx
```diff
@@ -7,12 +7,26 @@ import Logger from "../../__internal__/utils/logger";
...
+
+afterAll(() => {
+  jest.restoreAllMocks();
+});
+
 testStyledSystemMargin(
   (props) => <Loader data-role="loader" {...props} />,
-  () => screen.getByTestId("loader"),
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  () => screen.queryByTestId("loader")!,
```
nineteen88, line 29: suggestion(non-blocking): this could stay as getByTestID("loader")

That would mean we don't have to lint ignore anymore

nineteen88:
suggestion: could we update the title of the PR as well

Now that the commit messages has been updated, it should match, ideally.

src/components/loader-spinner/loader-spinner.test.tsx
```diff
@@ -301,26 +343,22 @@ describe("when custom props are passed", () => {
...
 
 describe("when the component is rendered and the end user prefers reduced motion", () => {
-  beforeEach(() => {
-    const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
-      typeof useMediaQuery
-    >;
-    mockUseMediaQuery.mockReturnValueOnce(false);
-  });
-
   it("the spinner label renders", () => {
```
edleeks87, line 362: comment: this is missing a mock, it will always pass as the label will be rendered no matter if the hook is true (and showLabel) or false so we should add another assertion to catch any regressions

```
it("the spinner label renders", () => {
    mockUseMediaQuery.mockReturnValueOnce(false);

    render(<LoaderSpinner />);
    const wrapperElement = screen.getByRole("status");
    const visibleLabelElement = screen.getByTestId("visible-label");
    const hiddenLabelElement = screen.getByTestId("hidden-label");

    expect(wrapperElement).toHaveTextContent("Loading...");
    expect(visibleLabelElement).toBeVisible();
    expect(hiddenLabelElement).not.toBeInTheDocument();
  });

File: src/components/loader-spinner/loader-spinner.test.tsx [hunk 11]
```diff
@@ -46,7 +46,7 @@ test("the component wrapper renders with a role of status", () => {
 });
 
 test("should return null and render nothing when useMediaQuery returns undefined", () => {
-  mockUseMediaQuery.mockReturnValueOnce(undefined);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 49: comment: this should be `mockUseMediaQuery.mockReturnValueOnce(undefined);`

src/components/loader-spinner/loader-spinner.test.tsx
```diff
@@ -223,7 +219,7 @@ describe("when custom props are passed", () => {
     const hiddenLabelElement = screen.getByTestId("hidden-label");
 
     expect(wrapperElement).toHaveTextContent("Loading...");
-    expect(hiddenLabelElement).toBeInTheDocument();
+    expect(hiddenLabelElement).not.toBeVisible();
```
edleeks87, line 222: comment: I think it's worth keeping the assertion as before but adding the new one etc. That way we're asserting that the label is there but not visible etc
```
expect(hiddenLabelElement).not.toBeVisible();
expect(hiddenLabelElement).toBeInTheDocument();

File: src/components/loader-spinner/loader-spinner.test.tsx [hunk 13]
```diff
@@ -104,13 +101,12 @@ describe("when custom props are passed", () => {
...
 
   it("should override the visually hidden label text when the 'spinnerLabel' prop is passed a custom string value", () => {
-    mockUseMediaQuery.mockReturnValueOnce(false);
     render(<LoaderSpinner spinnerLabel="bar" showSpinnerLabel={false} />);
 
     const status = screen.getByRole("status");
     const hiddenLabelElement = within(status).getByText("bar");
 
-    expect(hiddenLabelElement).toBeInTheDocument();
+    expect(hiddenLabelElement).not.toBeVisible();
```
edleeks87, line 109: comment: I think it's worth keeping the assertion as before but adding the new one etc. That way we're asserting that the label is there but not visible etc
```
expect(hiddenLabelElement).not.toBeVisible();
expect(hiddenLabelElement).toBeInTheDocument();

File: src/components/loader-star/loader-star.test.tsx [hunk 14]
```diff
@@ -31,7 +31,7 @@ test("should render with the expected data- attributes", () => {
 });
 
 test("should return null and render nothing when useMediaQuery returns undefined", () => {
-  mockUseMediaQuery.mockReturnValueOnce(undefined);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 34: comment: this should be `mockUseMediaQuery.mockReturnValueOnce(undefined);`

src/components/loader-star/loader-star.test.tsx
```diff
@@ -40,7 +40,7 @@ test("should return null and render nothing when useMediaQuery returns undefined
 });
 
 test("component should have a visually hidden label for assistive technologies by default", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 43: comment: the hidden label will only render when motion is allowed so this should be `true`

src/components/loader-star/loader-star.test.tsx
```diff
@@ -49,7 +49,7 @@ test("component should have a visually hidden label for assistive technologies b
 });
 
 test("renders a custom visually hidden label for assistive technologies to use, when loaderStarLabel prop is provided", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 52: comment: the hidden label will only render when motion is allowed so this should be `true`

src/components/loader/loader-square.test.tsx
```diff
@@ -22,7 +22,7 @@ afterAll(() => {
 });
 
 it.each([0, 1, 2])("each loader square renders as expected", (index) => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 25: comment: the should be `true`, but as there is a `beforeEach` at the top of the file already you don't need to have any mock value here

src/components/loader/loader-square.test.tsx
```diff
@@ -32,7 +32,7 @@ it.each([0, 1, 2])("each loader square renders as expected", (index) => {
 
 // These styling tests required for coverage
 test("when `size` prop is set to large, the expected width, height, and margin styles are applied", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 35: comment: the should be `true`, but as there is a `beforeEach` at the top of the file already you don't need to have any mock value here

src/components/loader/loader-square.test.tsx
```diff
@@ -57,7 +57,7 @@ test("when `size` prop is set to large, the expected width, height, and margin s
 
 // These styling tests required for coverage
 test("when `size` prop is set to small, the expected width, height, and margin styles are applied", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 60: comment: the should be `true`, but as there is a `beforeEach` at the top of the file already you don't need to have any mock value here

src/components/loader/loader-square.test.tsx
```diff
@@ -79,7 +79,7 @@ test("when `size` prop is set to small, the expected width, height, and margin s
 
 // These styling tests required for coverage
 test("when inside button, the expected white background colour is applied", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 82: comment: the should be `true`, but as there is a `beforeEach` at the top of the file already you don't need to have any mock value here

src/components/loader/loader-square.test.tsx
```diff
@@ -99,7 +99,7 @@ test("when inside button, the expected white background colour is applied", () =
 
 // These styling tests required for coverage
 test("when inside button and `isActive` prop is false, the expected background colour is applied", () => {
-  mockUseMediaQuery.mockReturnValueOnce(true);
+  mockUseMediaQuery.mockReturnValueOnce(false);
```
edleeks87, line 102: comment: the should be `true`, but as there is a `beforeEach` at the top of the file already you don't need to have any mock value here

src/components/loader-spinner/loader-spinner.pw.tsx
```diff
@@ -168,8 +168,6 @@ test.describe("Prop checks for Loader Spinner component", () => {
     mount,
     page,
   }) => {
-    await page.emulateMedia({ reducedMotion: "no-preference" });
```
edleeks87: praise: I like that you have removed this as we should rely on the default browser setting which is to allow motion etc

src/components/loader-spinner/loader-spinner.test.tsx
```diff
@@ -68,10 +104,13 @@ describe("when custom props are passed", () => {
...
   it("should override the visually hidden label text when the 'spinnerLabel' prop is passed a custom string value", () => {
+    mockUseMediaQuery.mockReturnValueOnce(false);
     render(<LoaderSpinner spinnerLabel="bar" showSpinnerLabel={false} />);
-    const hiddenLabelElement = screen.getByTestId("hidden-label");
 
-    expect(hiddenLabelElement).toHaveTextContent("bar");
+    const status = screen.getByRole("status");
+    const hiddenLabelElement = within(status).getByText("bar");
+
+    expect(hiddenLabelElement).toBeInTheDocument();
```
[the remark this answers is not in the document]
edleeks87: I may have this wrong as the styling for the label means it might still be considered visible by RTL, maybe asserting the visble label isn't in the doc is a better approach here

```
   const visibleLabelElement = screen.queryByTestId("visible-label");
   expect(visibleLabelElement).not.toBeInTheDocument();

File: src/components/loader-spinner/loader-spinner.test.tsx (hunk 13 above)
Comment (edleeks87) [reply src/components/loader-spinner/loader-spinner.test.tsx:110]: as mentioned above it looks like the hidden label will be visible to RTL so lets go with instead
```
const visibleLabelElement = screen.queryByTestId("visible-label"); expect(visibleLabelElement).not.toBeInTheDocument();

src/components/loader-spinner/loader-spinner.test.tsx
edleeks87: same comment about asserting the visible label is not in the document instead of the hidden label not being visible

src/components/loader-spinner/loader-spinner.component.tsx
```diff
@@ -142,7 +143,9 @@ export const LoaderSpinner = ({
           )}
         </>
       ) : (
-        renderSpinnerLabel
+        <Typography data-role="hidden-label" variant="span" screenReaderOnly>
```
edleeks87, line 146: https://github.com/Sage/carbon/pull/7268#discussion_r2026543009

edleeks87 (commented):
Hey @DobroTora good work getting the build to pass. It looks like you have two comments from me that haven't been actioned, I'm happy to approve once they have 

https://github.com/Sage/carbon/pull/7268/files#r2031416421
https://github.com/Sage/carbon/pull/7268/files#r2031411623

edleeks87 (dismissed):

nineteen88 (dismissed):

src/components/loader/loader-square.style.ts
```diff
@@ -48,29 +48,32 @@ const getDimensions = (size: StyledLoaderSquareProps["size"]) => {
   `;
 };
 
+export const StyledLoaderPlaceholder = styled.div<
+  Pick<StyledLoaderSquareProps, "size">
+>`
+  display: inline-block;
+  ${({ size }) => getDimensions(size)}
```
edleeks87, line 55: comment: I had this wrong it needs to have a min-width of 64px to account for the `squares` when they actually render

https://gist.github.com/edleeks87/d899b4701a65f3157f020e37689f8ad1

src/components/loader/loader.component.tsx
edleeks87, line 1: https://gist.github.com/edleeks87/d899b4701a65f3157f020e37689f8ad1

suggested change

src/hooks/useMediaQuery/useMediaQuery.ts
```diff
@@ -1,17 +1,24 @@
-import React from "react";
+import { useLayoutEffect, useState } from "react";
+import { getWindow } from "../../__internal__/dom/globals";
 
-export default function useMediaQuery(queryInput: string): boolean {
-  const query = queryInput.replace(/^@media( ?)/m, "");
+export default function useMediaQuery(queryInput: string): boolean | undefined {
+  const query = queryInput.replace(/^@media\s*/i, "");
```
edleeks87, line 5: comment: this regex pattern will match differently as it's now not case-sensitive for example

src/hooks/useMediaQuery/useMediaQuery.ts
```diff
@@ -1,17 +1,24 @@
...
 
-  const [match, setMatch] = React.useState(() => false);
+  const [match, setMatch] = useState<boolean | undefined>(undefined);
+
+  useLayoutEffect(() => {
+    const browserWindow = getWindow();
+
+    if (!browserWindow) {
```
edleeks87, line 12: suggestion: as we have tested the behaviour elsewhere lets just ignore the if here as we don't need the coverage etc

```
import { useLayoutEffect, useState } from "react";
import { getWindow } from "../../__internal__/dom/globals";

export default function useMediaQuery(queryInput: string): boolean | undefined {
  const query = queryInput.replace(/^@media( ?)/m, "");

  const [match, setMatch] = useState<boolean | undefined>(undefined);

  useLayoutEffect(() => {
    const browserWindow = getWindow();
    
    /* istanbul ignore if */
    if (!browserWindow) {
      return undefined;
    }
    
    const queryList = browserWindow.matchMedia(query);
    const updateMatch = () => setMatch(queryList.matches);
    updateMatch();
    queryList.addEventListener("change", updateMatch);
    return () => {
      queryList.removeEventListener("change", updateMatch);
    };
  }, [query]);

  return match;
}

src/hooks/useMediaQuery/useMediaQuery.test.tsx
edleeks87, line 1: comment: this can all be reverted as we don't really care about the coverage of the new `if` that was added (see other comment)

DobroTora: that was the result of coverage report.

edleeks87 (approved):

damienrobson-sage (approved):

PR #7269, opened 2025-03-21
feat(message): add ai variant

tomdavies73:
### Proposed behaviour

Adds the new ai variant to `Message`

Screenshot 2025-03-21 at 18 23 00

Refines our approach to `Message`, altering some padding values to align with DS designs, removing internal components, reducing code complexity and using preferred CSS instead of absolutely positioning the close button on `Message`.

### Current behaviour

Currently, there is no ai variant, and `Message` utilises unnecessary absolute styling, internal components and inefficient component code. 

### Checklist

#### QA


src/components/message/message.component.tsx
```diff
@@ -64,49 +69,94 @@ export const Message = React.forwardRef<HTMLDivElement, MessageProps>(
...
+                  fill="#00D639"
+                  d="M17.172 5.655a2.827 2.827 0 1 0 0-5.655 2.827 2.827 0 0 0 0 5.655Z"
+                />
+              </g>
+              <defs>
+                <clipPath id="a">
+                  <path fill="#fff" d="M0 0h20v20H0z" />
+                </clipPath>
+              </defs>
+            </svg>
```
damienrobson-sage, line 128: **question (non-blocking)**: is there any chance of this SVG needing to be re-used in other places in future? Just wondering if it would be better off placed in it's own file in e.g. `__internals__`

src/components/message/message.style.ts
```diff
@@ -41,20 +39,50 @@ const MessageStyle = styled.div<MessageStyleProps & MarginProps>`
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 16px;
+
   ${StyledIconButton} {
-    position: absolute;
-    right: 20px;
-    top: 50%;
-    transform: translateY(-50%);
+    margin-left: 24px;
```
damienrobson-sage, line 54: **suggestion (non-blocking)**: swap these out for the DS `spacing` tokens to future-proof.

src/components/message/message.test.tsx
Parsium, line 1: **suggestion(non-blocking)**: for these tests, what do you think about adding a `data-role` to the close button's icon and category icon?

For example:

```ts
    const categoryIcon = screen.getByTestId("category-icon");
    
    expect(categoryIcon).toBeVisible();
    expect(categoryIcon).toHaveAttribute("type", "info");
```

This could simplify the query and make it easier to understand.

src/components/message/message.style.ts
Parsium: We might actually be able to remove this margin rule altogether. With flexbox parent `MessageContent` justifying the flex items with `space-between`, the close icon will be positioned far to right as possible:

Screenshot 2025-03-24 at 09 45 37

src/components/message/message.component.tsx
```diff
@@ -64,49 +69,94 @@ export const Message = React.forwardRef<HTMLDivElement, MessageProps>(
...
+          ) : (
+            <Icon type={VARIANT_ICON_MAP[variant]} />
+          )}
+        </TypeIconStyle>
+
+        <Typography screenReaderOnly>
+          {locale.message?.[variant]?.()}
```
Parsium, line 135: **question**: I noticed the new `locale.message.ai` translation key is optional—just curious if there's a specific reason for this since most of the keys in the `Locale` interface aren't.

src/components/message/message-test.stories.tsx
```diff
@@ -5,7 +5,7 @@ import Message, { MessageProps } from "./message.component";
 
 export default {
   title: "Message/Test",
-  includeStories: ["Default", "Transparent"],
+  includeStories: ["Default", "Transparent", "TitleWithLongText", "Ai"],
```
Parsium, line 8: **suggestion(non-breaking)**: since this covers all the stories in this file, this option seems a bit redundant now. What do you think about removing it?

```suggestion
```

Parsium (commented):
**praise**: nice job simplifying the styling @tomdavies73! The flexbox approach looks much simpler  👍🏼

src/components/message/message.component.tsx
tomdavies73: @Parsium I've just made it optional for now whilst we wait for other translations to be added, but some other files do use a Partial<Locale> type, so the use of an optional type here may be redundant, good point. I'll check to make sure

src/components/message/message.style.ts
Parsium: Ah nice spot 👍🏼 Could we potentially add a `gap` to `MessageContent` instead, given its a flex container? That way, we could avoid setting a override on the `IconButton`

tomdavies73: Good idea mate, I'll double-check to make sure it doesn't have an adverse effect on any other configurations. But a gap could be the better solution here 👍

src/components/message/message.component.tsx
tomdavies73: After checking, this needs to remain optional until we have appropriate translations for our other locale docs

src/components/message/message.test.tsx
```diff
@@ -139,7 +194,7 @@ test("renders with expected styles when `transparent` is true", () => {
     background: "transparent",
   });
   expect(screen.getByTestId("message-content")).toHaveStyle({
-    padding: "15px 50px 15px 10px",
+    padding: "--spacing200",
```
Parsium, line 197: **issue**: we might need to use the `toHaveStyleRule` assertion instead of toHaveStyle for this test.

Due to a bug in jsdom with resolving CSS custom properties, `toHaveStyle` can produce a false positive when both the expected and actual style values are custom properties but don't match.

Parsium (dismissed):

damienrobson-sage (approved):

stephenogorman (approved):

damienrobson-sage (approved):

Parsium (approved):

Issue #6182, opened 2023-07-10
`disableClose` prop has no effect on `DialogFullscreen`


clheppell:
### Current behaviour

The `disableClose` prop listed in the storybook docs and typescript types has no impact on a `DialogFullscreen` and that you can still click on the X or Escape to close the dialog. I'm told by our QAs that it works on a normal `Dialog`.

### Expected behaviour

Either the prop to work or not be documented.

### CodeSandbox or Storybook URL

https://codesandbox.io/s/dazzling-wilson-y625xz?file=/src/App.js

### Carbon Version

118.2.1

### What browsers are you seeing the problem on?

Firefox, Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

robinzigmond:
FE-6053

Issue #6166, opened 2023-06-30
`timeout` prop has no effect on Dialog


clheppell:
### Current behaviour

A `timeout` prop is documented on Dialog and DialogFullscreen but has no effect.

(Might exist on the Sidebar TS interface too, but doesn't show in the storybook).

### Expected behaviour

Either timeout to work or the prop not be included

### CodeSandbox or Storybook URL

https://codesandbox.io/s/headless-field-jhpqvt?file=/src/App.js

### Carbon Version

118.2.1

### Design Tokens Version

2.65.0

### What browsers are you seeing the problem on?

Firefox, Chrome

### What Operating System are you seeing the problem on?

Windows

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
FE-6044

PR #7284, opened 2025-03-31
fix(dialog, dialog-full-screen): deprecate disableClose and timeout props

mihai-albu-sage:
fixes: [#6166](https://github.com/Sage/carbon/issues/6166)
fixes: [#6182](https://github.com/Sage/carbon/issues/6182)

### Proposed behaviour

Added a deprecation warning using the Logger  for the `timeout` prop when it is used inside the `Dialog` component and for the `timeout` and `disableClose` prop when it is used inside the `DialogFullScreen` component.

### Current behaviour

Currently these two props are inherited from the `Modal`. The `timeout` prop is not used inside the `Dialog` component and the `timeout` and `disableClose` props have no effect on the `DialogFullScreen` component.

### Checklist

#### QA


src/components/dialog/dialog.component.tsx
```diff
@@ -152,6 +153,15 @@ export const Dialog = forwardRef<DialogHandle, DialogProps>(
 
     const isTopModal = useModalAria(containerRef);
 
+    let deprecatedTimeoutTrigger = false;
```
Parsium, line 156: **issue**: if we would like to ensure the warning is logged only once, we might need to move this flag outside the component's scope.

Since it's currently within the component, it would reset to false on every re-render.

src/components/dialog-full-screen/dialog-full-screen.component.tsx
```diff
@@ -112,6 +113,23 @@ export const DialogFullScreen = ({
 
   const isTopModal = useModalAria(dialogRef);
 
+  let deprecatedDisableCloseTrigger = false;
+  let deprecatedTimeoutTrigger = false;
```
Parsium, line 117: **issue**: if we would like to ensure the warnings are logged only once, we might need to move these flags outside the component's scope.

Since they're currently within the component, they would reset to false on every re-render.

src/components/dialog-full-screen/dialog-full-screen.component.tsx
Parsium, line 1: **suggestion**: what do you think about about tweaking `DialogFullscreenProps` like this to mark the `disableClose` prop as deprecated in its description?:

```tsx
export interface DialogFullscreenProps extends Omit<ModalProps, "disableClose"> {
    /** @deprecated Determines if the Dialog can be closed */
    disableClose?: boolean; 
}
```

This way, the prop will be marked as deprecated in the Storybook docs and IDE code completion tooltips.

src/components/dialog/dialog.component.tsx
Parsium, line 1: **suggestion**: what do you think about about tweaking `ModalProps` like this to mark the `timeout` prop as deprecated in its description?:

```tsx
export interface ModalProps {
     /** @deprecated Transition time */
     timeout?: number;
}
```

This way, any components using props from `Modal` will see the prop marked as deprecated in the Storybook docs and IDE code completion tooltips.

Parsium (commented):
Just a few comments from me @mihai-albu-sage, let me know what you think! 😄

Parsium:
**suggestion(non-blocking)**: what do you think about separating the deprecations into two commits, one per component?

This would allow semantic release to generate separate release notes, which might be helpful since `Dialog` will continue to support `disableClose`

src/components/dialog/dialog.component.tsx
mihai-albu-sage: Moved the flag outside the component's scope.

src/components/dialog-full-screen/dialog-full-screen.component.tsx
mihai-albu-sage: Moved the flags outside the component's scope.

src/components/dialog-full-screen/dialog-full-screen.component.tsx
mihai-albu-sage: Updated the interface.

src/components/dialog/dialog.component.tsx
mihai-albu-sage: Done.

mihai-albu-sage:
> **suggestion(non-blocking)**: what do you think about separating the deprecations into two commits, one per component?
> 
> This would allow Semantic Release to generate separate release notes for each component, which might be helpful since `Dialog` will continue to support `disableClose`

Created two commits, one for each component.

Parsium (approved):

nuria1110 (approved):

dariusbercea-sage (approved):

PR #7305, opened 2025-04-15
feat: subcomponents log console warnings when used outside their parents (FE-7167)

Parsium:
### Proposed behaviour

For subcomponents that rely on their parent Carbon components to function, ensure a console warning is logged when they are not nested within their required parent.

This includes subcomponents of the following components:
- `Breadcrumbs`
- `ButtonToggleGroup`
- `Card`
- `Dl` (Definition list)
- `FlatTable`
- `StepFlow`
- `StepSequence`
- `Tile`
- `VerticalMenu`

### Current behaviour

Many subcomponents must be nested within their parent components in order to function correctly. No error handling occurs if this isn't the case. This means developers using these components incorrectly won't receive any warning that they're following an anti-pattern.

### Checklist

#### QA

#### Why log errors over throwing them?

For now, `createStrictContext` logs an error to the browser console and returns a fallback default value, as we are aware this anti-pattern has been used in practice, and we want to avoid breaking apps at this stage. This behaviour should be replaced with throwing an error in the future. 

#### Why don't `Accordion`, `Tabs`, etc. log errors?

Currently these components use the legacy `React.Children` API to modify the props of its children. This will be changed in the future in favour of context. When this occurs, similar error handling will need to be done.

### Testing instructions

Ensure a “Context not found” warning is logged in the browser console when any of the following components are not wrapped in their required parent:

- `Crumb` → `Breadcrumbs`
- `ButtonToggle` → `ButtonToggleGroup`
- `CardFooter`, `CardColumn` → `Card`
- `Dt`, `Dd` → `Dl`
- `FlatTableRow`, `FlatTableHeader`, `FlatTableRowHeader`, `FlatTableCell`, `Sort` → `FlatTable`
- `StepFlowTitle` → `StepFlow`
- `StepSequenceItem` → `StepSequence`
- `TileContent` → `Tile`
- `VerticalMenuItem` → `VerticalMenu` or `VerticalMenuFullScreen`


src/components/flat-table/__internal__/flat-table.context.ts
```diff
@@ -1,14 +1,10 @@
...
   isInFlatTable?: boolean;
-  setHasOpenDatePicker?: Dispatch<SetStateAction<boolean>>;
+  setHasOpenDatePicker?: (value: boolean) => void;
 }
 
-export default React.createContext<FlatTableContextProps>({
-  getTabStopElementId: () => "",
-  isInFlatTable: false,
-});
+const FlatTableContext = createContext<FlatTableContextType | null>(null);
```
damienrobson-sage, line 8: **query**: a number of consuming instances seem to default to `{}` if null is returned from the context, e.g.

```javascript
const { isInFlatTable } = useContext(FlatTableContext) || {};
```

Could we do something here to default the return to `{}` if there's no context available?

damienrobson-sage (commented):
I know `createStrictContext` is a pre-existing context but I'm wondering if there's any merit to writing a unit test or two for it, given how many places it's used in (alternatively/currently, we rely on these components failing to inform us that there is a problem with that specific context)

src/components/flat-table/__internal__/flat-table.context.ts
Parsium: I agree, it would be simpler to give `FlatTableContext` a default value here 👍🏼

damienrobson-sage (dismissed):

edleeks87 (approved):

paulrobinson88 (approved):

edleeks87 (approved):

damienrobson-sage (approved):

Issue #7295, opened 2025-04-10
Text Editor, add onBlur


allan-sage:
### Description

https://stackblitz.com/edit/parsium-carbon-starter-rgpn6cbk?file=src%2FApp.tsx

Use case:
When a Text Editor, with required property and is empty, loses focus; I want to fire/display error, that input is required in the Text Editor

### Suggested solution

add onBlur?

### Confidentiality

- [x] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
FE-7205

PR #7312, opened 2025-04-22
feat(text-editor): adds onBlur and onFocus support to Text Editor

damienrobson-sage:
BREAKING CHANGE: The `onChange` handler is no longer debounced internally. Consumers should implement their own debounce handlers when using `TextEditor` to ensure that e.g. performance is not affected.

Addresses #7295 

### Proposed behaviour

Support added for `onBlur` and `onFocus` props

### Current behaviour

There is currently no way to perform actions or monitor events when the Text Editor gains or loses focus

### Checklist

#### QA

### Testing instructions

Use the provided `onBlur Handler` and `onFocu Handler` Storybook examples


src/components/text-editor/text-editor.component.tsx
```diff
@@ -187,6 +196,36 @@ export const TextEditor = ({
...
+          );
+          // Get JSON representation
+          jsonData = (editorRef.current as LexicalEditor).toJSON();
+        });
+      }
+
+      onBlur?.(currentTextContent, { htmlString, json: jsonData });
+    },
+    [onBlur],
+  );
```
damienrobson-sage, line 227: Originally this used the `SerialiseLexical` helper function but that uses `editor.update` and didn't always fire successfully, if at all. I've broken it out into the direct code here as it's unlikely to ever be needed anywhere beyond here.

src/components/text-editor/text-editor.component.tsx
```diff
@@ -187,6 +196,36 @@ export const TextEditor = ({
...
+  const handleBlur = useCallback(
+    (newState: EditorState) => {
+      const currentTextContent = newState.read(() =>
+        $getRoot().getTextContent(),
+      );
+
+      let htmlString;
+      let jsonData;
+      /* istanbul ignore else */
+      if (editorRef.current) {
```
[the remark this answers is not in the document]
tomdavies73: Also, as we're ignoring else here, it may be worth using optional chaining during the definition, something like this 

`const editor = editorRef.current?.editor as LexicalEditor;`

src/components/text-editor/text-editor.component.tsx
```diff
@@ -208,6 +204,7 @@ export const TextEditor = ({
 
       let htmlString;
       let jsonData;
+
       /* istanbul ignore else */
       if (editorRef.current) {
```
tomdavies73, line 209: **suggestion**: I'd still recommend a local instance of this `editorRef.current` here with the `LexicalEditor` assertion, doesn't have to be file wide but even just in this `handleBlur` would be sufficient

tomdavies73 (dismissed):

src/components/text-editor/text-editor.component.tsx
```diff
@@ -72,10 +75,14 @@ export interface TextEditorProps extends MarginProps, TagProps {
   labelText: string;
   /** The identifier for the Text Editor. This allows for the using of multiple Text Editors on a screen */
   namespace?: string;
+  /** The callback to fire when the editor loses focus */
+  onBlur?: (value: string, formattedValues: EditorFormattedValues) => void;
```
Parsium, line 79: **question**: just wondering if we need to specify these value parameters in the `onBlur` callback, or if consumers could rely on `onChange` to access them? This might help simplify the design by separating concerns.

For instance, if a consumer is storing the raw and formatted values in state, they should have access to these when the input is blurred:

```tsx
export const App = () => {
  const [value, setValue] = React.useState<string>("");
  const [warning, setWarning] = React.useState<string>("");

  const handleBlur = () => {
    if (valueString.length === 0) {
      setWarning("Please enter some text.");
    } else {
      setWarning("");
    }
  };

  return (
    <TextEditor
      namespace="my-text-editor"
      labelText="Text Editor"
      onChange={(newValue) => setValue(newValue)}
      onBlur={handleBlur}
      warning={warning}
    />
  );
};
```

src/components/text-editor/text-editor.component.tsx
Parsium, line 1: **suggestion**: would it make sense to provide access to the original DOM event objects in `onFocus` and `onBlur` since consumers may expect it?

```ts
interface TextEditorProps {
  onFocus?: (ev: React.FocusEvent) => void;
  onBlur?: (ev: React.FocusEvent) => void;
}
```

This could give consumers extra flexibility, for instance if they want to handle specific focus behaviours.

damienrobson-sage:
Discussed with James and agreed to keep things simple and consistent with the rest of the components. I've removed the content handler functionality from `onBlur` so that it (and `onFocus`) now just return the DOM focus/blur event. If consumers need the data they can use `onChange` etc.

damienrobson-sage:
`onChange` handler is no longer debounced internally. Consumers should implement their own debounce handlers when using `TextEditor` to ensure that e.g. performance is not affected.

src/components/text-editor/text-editor.mdx
```diff
@@ -99,14 +96,39 @@ the JSON and HTML content of the editor. The `onCancel` function callback will b
...
+
 ### onChange Handler
 
 To handle changes to the content of the editor, you can set the `onChange` property to a function. The function will be called whenever the
-editor state changes, and provides a string representation of the editor content. This function is debounced by 500ms to prevent excessive
-calls to the callback function. The second argument passed to the `onChange` callback contains the `htmlString` and `JSON` content of the editor.
+editor state changes, and provides a string representation of the editor content. The second argument passed to the `onChange` callback contains
+the `htmlString` and `JSON` content of the editor.
+
+It is recommended that you use `onChange` in conjunction with a debounced function to avoid performance issues.
```
Parsium, line 122: **suggestion(non-breaking)**: given consumers should only need consider debouncing if their use case requires it, what do you think about removing this paragraph and the debounce function in the example?

This would simplify the code example, and avoid unintentionally communicating that passing a debounced change handler is a necessity.

Parsium (dismissed):

Parsium:
LGTM @damienrobson-sage 👍🏼 Since this is a breaking change, it might be worth updating the commit message so its released as one :)

tomdavies73 (dismissed):

src/components/text-editor/text-editor.component.tsx
Parsium, line 1: **suggestion**: Apologies for the late input, but as an alternative to the lexical plugin approach, would it make sense to attach event handlers directly onto `StyledWrapper`, the common ancestor of the toolbar and editor?

This might simplify the solution while ensuring that `onFocus` triggers only when focus enters the component, and `onBlur` only when it fully leaves, without having to check for cases like the toolbar being focused/blurred.  

```tsx
<StyledWrapper
  onBlur={(ev) => {
    if (!ev.currentTarget.contains(ev.relatedTarget)) {
      onBlur?.(ev);
    }
  }}
  onFocus={(ev) => {
    if (!ev.currentTarget.contains(ev.relatedTarget)) {
      onFocus?.(ev);
    }
  }}
>
```
To support this, we'd also need to allow focus events from the `Toolbar` to propagate:

```tsx
// on line 110 in toolbar.component.tsx:
<StyledToolbar
  onFocus={(e) => e.stopPropagation()} // remove to re-allow propagation
>
```

Parsium (approved):

tomdavies73 (approved):

paulrobinson88 (approved):

PR #7314, opened 2025-04-25
Deprecate "optional" prop

damienrobson-sage:
### Proposed behaviour

Add deprecation warnings to all impacted components to indicate that the property will be removed in a future update.

### Current behaviour

We currently offer an "optional" property on components.

### Checklist

#### QA

### DIFF
--- a/src/__internal__/checkable-input/checkable-input.component.tsx
+++ b/src/__internal__/checkable-input/checkable-input.component.tsx
@@ -41,7 +41,10 @@ export interface CommonCheckableInputProps
   labelWidth?: number;
   /** Flag to configure component as mandatory */
   required?: boolean;
-  /** Flag to configure component as optional. */
+  /**
+   * [Legacy] Flag to configure component as optional.
+   * @deprecated If the value of this component is not required, use the `required` prop and set it to false instead.
+   */
   isOptional?: boolean;
   /** If true the label switches position with the input */
   reverse?: boolean;
--- a/src/components/select/__internal__/select-textbox/select-textbox.component.tsx
+++ b/src/components/select/__internal__/select-textbox/select-textbox.component.tsx
@@ -67,7 +67,10 @@ export interface FormInputPropTypes
    *
    */
   labelId?: string;
-  /** Flag to configure component as optional in Form */
+  /**
+   * [Legacy] Flag to configure component as optional.
+   * @deprecated If the value of this component is not required, use the `required` prop and set it to false instead.
+   */
   isOptional?: boolean;
 }
 
--- a/src/__internal__/fieldset/fieldset.component.tsx
+++ b/src/__internal__/fieldset/fieldset.component.tsx
@@ -51,7 +51,10 @@ export interface FieldsetProps extends MarginProps {
   legendMargin?: Pick<MarginProps, "mb">;
   /** Any valid CSS string to set the component's width */
   width?: string;
-  /** Flag to configure component as optional in Form */
+  /**
+   * [Legacy] Flag to configure component as optional.
+   * @deprecated If the value of this component is not required, use the `isRequired` prop and set it to false instead.
+   */
   isOptional?: boolean;
   /** Apply disabled styling to the legend content */
   isDisabled?: boolean;
--- a/src/__internal__/fieldset/fieldset.style.ts
+++ b/src/__internal__/fieldset/fieldset.style.ts
@@ -23,6 +23,10 @@ StyledFieldset.defaultProps = {
 
 type StyledLegendContentProps = {
   isRequired?: boolean;
+  /**
+   * [Legacy] Flag to configure component as optional.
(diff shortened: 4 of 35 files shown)


src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -99,6 +105,13 @@ const CheckableInput = React.forwardRef(
     }: CheckableInputProps,
     ref: React.ForwardedRef<HTMLInputElement>,
   ) => {
+    if (!deprecateOptionalWarnTriggered && !!isOptional) {
```
tomdavies73, line 108: **suggestion**: I don't think a double negation is necessary here, we can just use the original `isOptional` value, as that itself is a boolean

tomdavies73: This approach appears to have been used for every instance of `isOptional`, I'll just leave this comment here instead of commenting everywhere it's used 👍

src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -99,6 +105,13 @@ const CheckableInput = React.forwardRef(
...
+      deprecateOptionalWarnTriggered = true;
+      Logger.deprecate(
+        "`isOptional` is deprecated in CheckableInput and support will soon be removed. Please use `required` instead.",
```
tomdavies73, line 111: **suggestion**: it might be worth fleshing this out a bit more, instead of `Please use required instead`, we could explain that required should be used for required fields, and therefore the absence of required indicates a field is optional.

I know this is still implied with your deprecation message, but it might be worth being extra explicit

src/__internal__/checkable-input/checkable-input.component.tsx
```diff
@@ -41,7 +42,10 @@ export interface CommonCheckableInputProps
   labelWidth?: number;
   /** Flag to configure component as mandatory */
   required?: boolean;
-  /** Flag to configure component as optional. */
+  /**
+   * [Legacy] Flag to configure component as optional.
+   * @deprecated Please use `required` instead.
```
tomdavies73, line 47: **suggestion**: it might be worth fleshing this out a bit more, instead of `Please use required instead`, we could explain that required should be used for required fields, and therefore the absence of required indicates a field is optional.

I know this is still implied with your deprecation message, but it might be worth being extra explicit

tomdavies73: This approach appears to have been used in every prop description, I'll just leave this comment here instead of commenting everywhere its used 👍

src/__internal__/checkable-input/checkable-input.component.tsx
tomdavies73: This approach appears to have been used in every deprecation message, I'll just leave this comment here instead of commenting everywhere it's used 👍

src/components/select/__internal__/select-textbox/select-textbox.component.tsx
```diff
@@ -120,6 +126,12 @@ const SelectTextbox = React.forwardRef(
     }: SelectTextboxProps,
     ref: React.ForwardedRef<HTMLInputElement>,
   ) => {
+    if (!deprecateOptionalWarnTriggered && !!isOptional) {
```
tomdavies73, line 129: **issue**: we don't need a warning here as this is an internal component, it will fire every time `SimpleSelect`, `MultiSelect` or `FilterableSelect` is rendered resulting in a double deprecation warning

src/__internal__/checkable-input/checkable-input.test.tsx
```diff
@@ -2,6 +2,23 @@ import React from "react";
...
 import CheckableInput from ".";
 
+import Logger from "../../__internal__/utils/logger";
+
+jest.mock("../../__internal__/utils/logger");
+
+test("should display deprecation warning once when rendered as optional", () => {
+  const loggerSpy = jest.spyOn(Logger, "deprecate");
+
+  render(<CheckableInput type="checkbox" label="label" id="foo" isOptional />);
```
tomdavies73, line 12: **suggestion**: as we're using file-level decleration for the `..warnTriggered` `let`s this means it will not be triggered more than once, as it persists accross multiple renders. I would use a fragment and render two `CheckableInput`s, and still assert the logger is still only called once

```
<>
<CheckableInput type="checkbox" label="label" id="foo" isOptional />
<CheckableInput type="checkbox" label="label" id="foo" isOptional />
</>
```

tomdavies73: This approach appears to have been used in every test, I'll just leave this comment here instead of commenting everywhere its used 👍

DipperTheDan (dismissed):

tomdavies73 (approved):

DipperTheDan (approved):

dariusbercea-sage (approved):

PR #7315, opened 2025-04-28
fix(button): ensure button focuses on click for Safari compatibility

DobroTora:
### Proposed behaviour

handleClick function inside Button component was introduced to:
    * First, manually call internalRef.focus().
    * Then, invoke either the SplitButton's onChildButtonClick handler (if the button is inside a SplitButton) or the normal onClick prop.

The Button no longer directly uses the provided onClick inside the JSX; it wraps it with the handleClick.

Buttons properly receive focus on mouse clicks in Safari, which does not automatically focus buttons after mouse interactions by default.

Focus behaviour via keyboard (Tab navigation) remains unchanged.

SplitButton children, which rely on onChildButtonClick, are also automatically covered because the new handleClick preserves the correct event forwarding.

Components like ButtonBar component, which renders child Buttons, and the Button component now ensures correct focus behaviour on click, ButtonBar automatically inherits the improved focus behaviour without needing any internal modification.

### Current behaviour

The Button and related components that are passing button as a child (apart from Split Button) are not displaying focus border, when clicked on Safari browser. 

### Checklist

#### QA

### DIFF
--- a/src/components/button/button.component.tsx
+++ b/src/components/button/button.component.tsx
@@ -273,6 +273,18 @@ const Button = React.forwardRef<
       }
     };
 
+    const handleClick = (
+      event: React.MouseEvent<HTMLButtonElement | HTMLAnchorElement>,
+    ) => {
+      internalRef?.focus();
+
+      if (inSplitButton) {
+        onChildButtonClick?.(onClick)?.(event);
+      } else if (onClick) {
+        onClick(event);
+      }
+    };
+
     switch (size) {
       case "small":
         paddingX = 2;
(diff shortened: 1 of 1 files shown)


src/components/button/button.component.tsx
```diff
@@ -273,6 +273,18 @@ const Button = React.forwardRef<
...
+    const handleClick = (
+      event: React.MouseEvent<HTMLButtonElement | HTMLAnchorElement>,
+    ) => {
+      internalRef?.focus();
+
+      if (inSplitButton) {
+        onChildButtonClick?.(onClick)?.(event);
+      } else if (onClick) {
+        onClick(event);
+      }
```
damienrobson-sage, line 285: **suggestion (non-blocking)**: you could re-write this to be cleaner, but it's fine as it is (the `else` isn't really needed given there's only a single line of code within it):

```javascript
if (inSplitButton) {
  onChildButtonClick?.(onClick)?.(event);
}
onClick?.(event);
```

edleeks87 (approved):

Parsium (approved):

stephenogorman (approved):

PR #7334, opened 2025-05-13
fix(decimal): ensure invariant message is not thrown when using strict mode - FE-7198

DipperTheDan:
### Proposed behaviour

Decimal should render as expected when using React's strict mode.

### Current behaviour

Invariant warning of `...switch from uncontrolled to controlled (or vice versa)...` is displayed when Decimal is rendered using React's strict mode.

### Checklist

#### QA

### Additional context

We decided that we would upgrade RTL during this work so that we could use the `reactStrictMode` render option. Prior to this the solution was the change the global config value of `reactStrictMode: false` to `true` for the test. In v16.3.0 of RTL the `reactStrictMode` render option was released and it is something we can utilise in other unit tests concerning strict mode behaviour. 

### Testing instructions

To boot up Storybook locally in `Strict Mode`, please run the cmd `npm run start:strict-mode`. Decimal should also be checked without strict mode too. 

- There should be no functional regressions for Decimal.
- No invariant message should fire when component is used in strict mode.

### DIFF
--- a/src/components/decimal/decimal.component.tsx
+++ b/src/components/decimal/decimal.component.tsx
@@ -278,7 +278,7 @@ export const Decimal = React.forwardRef(
     };
 
     const isControlled = value !== undefined;
-    const prevControlledRef = useRef<boolean>();
+    const prevControlledRef = useRef<boolean | null>();
 
     useEffect(() => {
       const message =
@@ -288,6 +288,10 @@ export const Decimal = React.forwardRef(
       invariant(prevControlledRef.current !== isControlled, message);
 
       prevControlledRef.current = isControlled;
+
+      return () => {
+        prevControlledRef.current = null;
+      };
     }, [isControlled]);
 
     const prevValue = usePrevious(value);
--- a/src/components/decimal/decimal.test.tsx
+++ b/src/components/decimal/decimal.test.tsx
@@ -1449,3 +1449,23 @@ test("the `required` prop is not passed to the hidden input", () => {
 
   expect(screen.getByTestId("hidden-input")).not.toBeRequired();
 });
+
+test("component should render without invariant firing in strict mode", () => {
+  const consoleErrorSpy = jest
+    .spyOn(console, "error")
+    .mockImplementation(() => {});
+
+  render(<ControlledDecimal startingValue="123" />, { reactStrictMode: true });
+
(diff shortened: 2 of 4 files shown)


src/components/decimal/decimal.component.tsx
```diff
@@ -279,8 +279,14 @@ export const Decimal = React.forwardRef(
...
     const isControlled = value !== undefined;
     const prevControlledRef = useRef<boolean>();
+    const isFirstRender = useRef(true);
 
     useEffect(() => {
+      if (isFirstRender.current) {
+        isFirstRender.current = false;
+        return;
+      }
+
```
Parsium, line 289: **suggestion**: I noticed that while your fix addresses the error, the value of `prevControlledRef` will still persist in memory after the component unmounts. What do you think about using a cleanup function like this?: 

```tsx
const prevControlledRef = useRef<boolean | null>(null);

useEffect(() => {
  // ...declare message and invariant...

  prevControlledRef.current = isControlled;

  return () => {
    prevControlledRef.current = null;
  };
}, [isControlled]);
```

This will ensure `prevControlledRef` is properly cleaned up when the component unmounts.

[The React team suggest this pattern over preventing effects from firing](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development), since it can prevent memory leaks from a previous mounts for a variety of scenarios.

DipperTheDan: I shall get this implemented and I wasn't aware of that section in the React docs. Thank you for sharing it 😄

Parsium (dismissed):

src/components/decimal/decimal.test.tsx
```diff
@@ -1449,3 +1449,25 @@ test("the `required` prop is not passed to the hidden input", () => {
 
   expect(screen.getByTestId("hidden-input")).not.toBeRequired();
 });
+
+test("component should render without invariant firing in strict mode", () => {
+  configure({ reactStrictMode: true });
```
edleeks87, line 1454: comment: As this is setting a global configuration I think we should call `configure({ reactStrictMode: false })` at the end of the `test`.

edleeks87 (approved):

Parsium (approved):

edleeks87 (approved):

dariusbercea-sage (approved):

PR #7337, opened 2025-05-16
test(playwright): output incomplete accessiblity findings to console

damienrobson-sage:
### Proposed behaviour

Log incomplete tests to the console, which will appear in both pipeline runs and when ran locally.

### Current behaviour

Playwright currently does not flag accessibility findings that are marked as `incomplete` by `axe-core`, which has the potential to be allowing a number of false negatives through the check process.

### Checklist

#### QA

### Additional context

This is a much simpler approach than would be preferred however it does expedite the process of finding potential issues without requiring changes to the pipeline or other CI processes.

### Testing instructions

Run the following test:

```
src/components/button/button.pw.tsx > should pass accessibility tests for SecondaryButtonWhite example
``` 

Note that it passes. Open `src/components/button/components.test-pw.tsx` and set the `backgroundColor` of the wrapping `Box` to `white` in the `SecondaryButtonWhite` component:

```javascript
export const SecondaryButtonWhite = () => {
  return (
    <Box
      display="flex"
      flexDirection="row"
      gap={2}
      backgroundColor="white"
    >
      <Button size="small" isWhite>
        Small
      </Button>
      <Button ml={2} isWhite>
        Medium
      </Button>
      <Button size="large" isWhite>
        Large
      </Button>
    </Box>
  );
};
```

Re-run the test, and note the output. The test should still pass, but a warning should be logged in the console:
```
Running 1 test using 1 worker
[chromium] › src/components/button/button.pw.tsx:418:7 › Accessibility tests for Button › should pass accessibility tests for SecondaryButtonWhite example

ACCESSIBILITY SCAN INCOMPLETE. Incomplete rules: 
        - color-contrast: this is a serious accessibility issue. Ensure the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds

  1 passed (7.2s)
```

### DIFF
--- a/playwright/support/helper.ts
+++ b/playwright/support/helper.ts
@@ -61,6 +61,27 @@ export const checkAccessibility = async (
     .disableRules([...preDisabledRules, ...disableRules])
     .analyze();
 
+  const isCI = process.env.CI === "true";
+
+  if (accessibilityScanResults.incomplete.length > 0 && !isCI) {
+    // Capture the calling stack
+    const { stack } = new Error();
+
+    // Parse stack to find the test file name
+    const testFileMatch = stack?.match(/at .*?(\/[^\s]+\.pw\.tsx):(\d+):(\d+)/);
+    const componentName = testFileMatch
+      ? testFileMatch[1].split("/").slice(-2, -1)[0]
+      : "Unknown component";
+
+    // eslint-disable-next-line no-console
+    console.warn(
+      `\nACCESSIBILITY SCAN INCOMPLETE. Incomplete rules: ${accessibilityScanResults.incomplete.map(
+        (rule) =>
+          `\n\t- ${rule.id}: this is a ${rule.impact} accessibility issue. ${rule.description}`,
+      )}\nPlease check and ensure that the "${componentName}" component meets accessibility criteria manually.\n`,
+    );
+  }
+
   expect(accessibilityScanResults.violations).toEqual([]);
 };
 


playwright/support/helper.ts
```diff
@@ -61,6 +61,15 @@ export const checkAccessibility = async (
     .disableRules([...preDisabledRules, ...disableRules])
     .analyze();
 
+  if (accessibilityScanResults.incomplete.length > 0) {
```
edleeks87, line 64: comment: I know we have plenty of lint warnings in master but I think it's worth adding a lint ignore here so we're not adding more etc

playwright/support/helper.ts
```diff
@@ -61,6 +61,15 @@ export const checkAccessibility = async (
...
+    console.warn(
+      `\nACCESSIBILITY SCAN INCOMPLETE. Incomplete rules: ${accessibilityScanResults.incomplete.map(
```
edleeks87, line 66: suggestion: I think it might be worth adding the component that the incomplete result has come from right now it looks like below so doesn't have much context.

image

edleeks87: suggestion (non-blocking): It might also be worth adding a check for `process.env.CI` and only adding the warn locally rather than polluting the workflow runs. I think if we want this information gathering during CI runs we can generate a report or something but happy for that to be a longer term goal. Maybe we can just put a single small message in the CI saying the following components have incomplete accessibility checks or some sort of count

edleeks87 (approved):

DipperTheDan (approved):

stephenogorman (approved):

PR #7341, opened 2025-05-19
refactor(link): split into Link and LinkBase for logic and presentation

DobroTora:
### Checklist

#### QA


src/components/link/base-link.component.tsx
edleeks87, line 1: Comment: this should be in an `__internal__` directory. I think we should split the tests up as well so there is a `base-link.test.tsx` added which tests all the functional behaviour of the component.

```
 link
  |__ __internal__
  |       |__ base-link.component.tsx
  |       |__ base-link.test.tsx
  |       |__ index.ts
  |
  |__ link.component.tsx
  |__ link.style.ts
  |__ link.stories.tsx
  |__ link-test.stories.tsx
  |__ link.pw.tsx
  |__ link.config.ts
  |__ link.mdx
  |__ link.test.tsx
  |__ components.test-pw.tsx

**Review by edleeks87 — COMMENTED** [review]
I haven't done a full review as the build is failing, I've left one comment about how the component's directory has been organised and I'll do a full review when the build is passing

File: src/components/link/__internal__/base-link.component.tsx [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+      ref?: Ref<HTMLButtonElement>;
+    } & ButtonHTMLAttributes<HTMLButtonElement>);
+
+const BaseLink = React.forwardRef<
+  HTMLAnchorElement | HTMLButtonElement,
+  BaseLinkProps
+>(({ href, children, $styles, ...rest }, ref) => {
+  return (
+    <StyledBaseLinkWrapper $styles={$styles}>
+      {href ? (
```
edleeks87: Comment: this will render a `button` if no `href` is passed which isn't the correct semantics

```
const setAnchorRef = React.useCallback(
      (reference: HTMLAnchorElement | null) => {
        if (!ref) return;
        if (typeof ref === "object" && ref !== null)
          (
            ref as React.MutableRefObject<
              HTMLAnchorElement | HTMLButtonElement | null
            >
          ).current = reference;
          ref(reference);
        }
      },
      [ref],
    );

    const setButtonRef = React.useCallback(
      (reference: HTMLButtonElement | null) => {
        if (!ref) return;
        if (typeof ref === "object" && ref !== null)
          (
            ref as React.MutableRefObject<
              HTMLAnchorElement | HTMLButtonElement | null
            >
          ).current = reference;
          ref(reference);
        }
      },
      [ref],
    );

 {onClick && !href ? (
          <button ref={setButtonRef} type="button" {...componentProps}>
            {children}
          </button>
        ) : (
          <a ref={setAnchorRef} {...componentProps} data-role="link-anchor">
            {children}
          </a>
        )}

File: src/components/link/__internal__/base-link.component.tsx [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+import type {
+  AnchorHTMLAttributes,
+  ButtonHTMLAttributes,
+  Ref,
+  AriaAttributes,
+} from "react";
+import type { FlattenSimpleInterpolation } from "styled-components";
+import StyledBaseLinkWrapper from "./base-link.style";
+
+interface SharedProps extends AriaAttributes {
```
edleeks87: comment: this feels like it's too complex and might create issue higher up n the component hierarchy as TS tries to evaluate types etc. Something like below should work:

```
export interface BaseLinkProps extends React.AriaAttributes {
  /** An href for an anchor tag. */
  href?: string;
  /** Function called when the mouse is clicked. */
  onClick?: (
    ev:
      | React.MouseEvent<HTMLAnchorElement>
      | React.MouseEvent<HTMLButtonElement>
      | React.KeyboardEvent<HTMLAnchorElement>
      | React.KeyboardEvent<HTMLButtonElement>,
  ) => void;
  /** Function called when a key is pressed. */
  onKeyDown?: (
    ev:
      | React.KeyboardEvent<HTMLAnchorElement>
      | React.KeyboardEvent<HTMLButtonElement>,
  ) => void;
  /** Function called when a mouse down event triggers. */
  onMouseDown?: (
    ev:
      | React.MouseEvent<HTMLAnchorElement>
      | React.MouseEvent<HTMLButtonElement>,
  ) => void;
  /** Function called when focus is received. */
  onFocus?: (
    ev:
      | React.FocusEvent<HTMLAnchorElement>
      | React.FocusEvent<HTMLButtonElement>,
  ) => void;
  /** Function called when focus is lost. */
  onBlur?: (
    ev:
      | React.FocusEvent<HTMLAnchorElement>
      | React.FocusEvent<HTMLButtonElement>,
  ) => void;
  /** Set the component to disabled */
  disabled?: boolean;
  /** Child content to render in the link. */
  children?: React.ReactNode;
  /** Target property in which link should open ie: _blank, _self, _parent, _top */
  target?: string;
  /** Aria label for accessibility purposes */
  ariaLabel?: string;
  /** allows to set rel property in <a> tag */
  rel?: string;
  /**
   * @private
   * @internal
   * @ignore
   * Sets className for component. INTERNAL USE ONLY. */
  className?: string;
  /**
   * @private
   * @internal
   * @ignore
   */
  "data-component"?: string;
  /**
   * @private
   * @internal
   * @ignore
   */
  "data-element"?: string;
  /**
   * @private
   * @internal
   * @ignore
   */
  "data-role"?: string;
  styles?: FlattenSimpleInterpolation;
}

File: src/components/link/__internal__/base-link.component.tsx [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+  $styles?: FlattenSimpleInterpolation;
+  disabled?: boolean;
+}
+
+type BaseLinkProps =
+  | (SharedProps & {
+      href: string;
+      target?: string;
+      rel?: string;
+      ref?: Ref<HTMLAnchorElement>;
```
edleeks87: comment: you don't need a `ref` prop, BaseLink should just support `forwardRef`

```
export const BaseLink = React.forwardRef<
  HTMLAnchorElement | HTMLButtonElement,
  BaseLinkProps
>(

File: src/components/link/__internal__/base-link.style.ts [hunk 4]
```diff
@@ -0,0 +1,11 @@
+import styled, { FlattenSimpleInterpolation } from "styled-components";
```
damienrobson-sage: **question**: can you explain why you're using `FlattenSimpleInterpolation` here as opposed to `SimpleInterpolation`. The only difference I can see from the documentation is that the `FlattenSimpleInterpolation` variant is a read-only variant of `SimpleInterpolation`.

src/components/link/__internal__/base-link.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  it("applies inline style", () => {
+    render(
+      <BaseLink href="#" style={{ color: "red" }}>
+        Styled
+      </BaseLink>
+    );
+    const anchor = screen.getByRole("link");
+    expect(anchor).toHaveStyle({ color: "red" });
+  });
```
damienrobson-sage: **comment**: I don't think we should be testing this. The styling should occur within the child component - `BaseLink` is just a wrapper

src/components/link/link.component.tsx
```diff
@@ -1,206 +1,149 @@
...
-  /** [Legacy] A message to display as a tooltip to the link. */
+  /** The name of the icon to display alongside the link content. */
+  icon?: string;
+  /** Which side of the link to render the icon. */
+  iconAlign?: "left" | "right";
+  /** Tooltip text displayed when hovering over the link. */
   tooltipMessage?: string;
-  /** [Legacy] Positions the tooltip with the link. */
+  /** The position of the tooltip relative to the link. */
   tooltipPosition?: "bottom" | "left" | "right" | "top";
```
damienrobson-sage: **suggestion**: As tooltips are being removed soon, this legacy label should remain.

src/components/link/link.component.tsx
```diff
@@ -1,206 +1,149 @@
...
+      hasFocus,
+      isDarkBackground,
+    });
 
-    useEffect(() => {
-      if (disabled || !(href || onClick)) {
-        setHasFocus(false);
-      }
-    }, [disabled, href, onClick]);
+    const fallbackLabel = ariaLabel || "Back";
```
damienrobson-sage: **comment**: I don't think we should use a fall-back here, as you can't guarantee that using `Back` for every non-aria link is appropriate. Users should be passing child elements to the `Link` component as per [the docs](https://carbon.sage.com/?path=/docs/link--docs)

src/components/link/link.style.types.ts
```diff
@@ -0,0 +1,22 @@
+export type Variants = "default" | "negative" | "neutral";
```
damienrobson-sage: **suggestion**: Use `VARIANTS` instead to maintain consistency with other components e.g `src/components/link/link.config.ts`

src/components/link/__internal__/base-link.test.tsx
DobroTora: this test case has been removed

src/components/link/__internal__/base-link.style.ts
DobroTora: so I am using it in a return type for StyledLinkStyles as I am using css from styled-components for example. As a result css tagged template returns flatten usable block of styles. I changed as I am not going to argue over that one.

src/components/link/link.style.ts
```diff
@@ -53,7 +22,7 @@ const colorMap: ColorMap = {
       disabledColor: "var(--colorsActionMajorYin030)",
     };
   },
-  dark: (variant) => {
+  dark: (variant?: StyledLinkProps["variant"]) => {
```
edleeks87:
```suggestion
  dark: (variant?: Variants) => {
```

src/components/link/link.style.ts
```diff
@@ -72,173 +41,109 @@ const colorMap: ColorMap = {
   },
 };
 
-const StyledLink = styled.span.attrs(applyBaseTheme)<
-  StyledLinkProps & PrivateStyledLinkProps
->`
-  ${({
+const StyledLinkStyles = (props: StyledLinkProps): SimpleInterpolation => {
```
edleeks87:
```suggestion
const linkStyles = (props: StyledLinkProps): SimpleInterpolation => {
```

I think we'd be better served following the function naming convention of camelCase here, it will hopefully make it more intuitive for anyone reading it etc

edleeks87: I think it would be better to avoid wholesale changes to styles for now, it's a bit out of scope of this ticket so if we do want to make some improvements a new ticket would be better. It also makes the review more complex and we run the risk of introducing style regressions that we may not catch. For example the focus styling now has a border radius when previously we would have set that to 0 etc

src/components/link/link.style.ts
```diff
@@ -1,42 +1,11 @@
...
-interface ColorMap {
-  light: ColorMapProp;
-  dark: ColorMapProp;
-}
-
-const colorMap: ColorMap = {
-  light: (variant) => {
+import type { StyledLinkProps } from "./link.style.types";
+
+export const StyledContent = styled.span``;
```
edleeks87: comment: I think we can just use `span` directly in the component rather than creating an additional empty styled-component

src/components/link/link.style.ts
```diff
@@ -1,42 +1,11 @@
...
+
+const colorMap = {
+  light: (variant?: StyledLinkProps["variant"]) => {
```
edleeks87:
```suggestion
  light: (variant?: Variants) => {
```

src/components/link/link.style.types.ts
damienrobson-sage, line 1: **comment**: I'd leave this declaration in the `link.style.ts` file, to remain consistent with how we approach them in other components

src/components/link/__internal__/base-link.component.tsx
edleeks87: This is still falling back to button

PR #7344, opened 2025-05-20
feat(text-editor): add programmatic focus support for the text editor

### LINKED ISSUE #7320: Extending TextEditor to call `onChange` on style changes and expose `ref`
### Description

After `<TextEditor>` was migrated to using `lexical`, the following behaviours are no longer supported:
- `onChange` is no longer triggered after style changes (bold, italic, ...etc.). This is a crucial feature because not every consumer of `<TextEditor>` uses the Save button and thus not logging the latest styles when the user changes the style without typing anything new
- `ref` is no longer exposed. This is needed when integrating with form libraries like React Hook Form, for use cases like field focus on error

### Suggested solution

These behaviours should be restored to facilitate migration to using the latest `<TextEditor>`.

### Demo URL

_No response_

### Alternatives

No alternatives possible

### Additional context

_No response_

### Confidentiality

tomdavies73:
fixes #7320 

### Proposed behaviour

`TextEditor` now accepts a `forwardRef`, and an imperative handle has been introduced which provides consumers with access to a focus method only.

Tests have been added to establish a baseline for `onChange` behaviour being triggered by toolbar buttons.

### Current behaviour

`TextEditor` no longer supports a `forwardRef` since its conversion to Lexical, but some consumers rely on passing a ref for programmatic focus control.

Also clicks on toolbar buttons result in `onChange` being triggered, due  previous changes, however there are no tests added to test against this behaviour.

### Checklist

#### QA

### Testing instructions

The new programmatic focus story on `TextEditor` should be sufficient enough to test that the new imperative handle does indeed result in the TextEditor being focused

### DIFF
--- a/src/components/text-editor/__internal__/plugins/Toolbar/toolbar.test.tsx
+++ b/src/components/text-editor/__internal__/plugins/Toolbar/toolbar.test.tsx
@@ -11,6 +11,7 @@ import React from "react";
 
 import userEvent from "@testing-library/user-event";
 import { ToolbarPlugin } from "..";
+import TextEditor from "../../../text-editor.component";
 
 function headlessEditor() {
   const editor = createHeadlessEditor({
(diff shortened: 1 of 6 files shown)


damienrobson-sage (approved):
LGTM 👍

paulrobinson88 (approved):

PR #7349, opened 2025-05-23
refactor: replace polished with custom utils

nuria1110:
### Proposed behaviour

- Remove `polished` dependency. 
- Replace `shade` with CSS. 
- Replace `meetsConstrastGuidelines` with our own custom utils. 
- When using custom colours on Pill, text colour should pass colour contrast checks

### Current behaviour

- Icon and Pill components use `shade` from polished to determine colours on hover when a custom one is provided. 
- Pill uses `meetsContrastGuidelines` from polished to determine the text and close icon colour when a custom `borderColor` is provided. 
- Some content (text) colours calculated by polished cause colour contrast fails:

image
image

### Checklist

#### QA

### DIFF
--- a/src/components/pill/pill.style.ts
+++ b/src/components/pill/pill.style.ts
@@ -1,5 +1,4 @@
 import styled, { css } from "styled-components";
-import { shade, meetsContrastGuidelines } from "polished";
 import { margin, MarginProps } from "styled-system";
 
 import styleConfig from "./pill.style.config";
@@ -9,6 +8,8 @@ import StyledIcon from "../icon/icon.style";
 import StyledIconButton from "../icon-button/icon-button.style";
 import { toColor } from "../../style/utils/color";
 import getColorValue from "../../style/utils/get-color-value";
+import getHexValue from "../../style/utils/get-hex-value";
+import getAccessibleForegroundColor from "../../style/utils/get-accessible-foreground-color";
 
 export interface StyledPillProps extends MarginProps {
   /** Override color variant, provide any color from palette or any valid css color value. */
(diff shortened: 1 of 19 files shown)


src/style/utils/get-accessible-foreground-color.ts
```diff
@@ -27,6 +27,13 @@ const calculateLuminance = (hexColor: string): number => {
   return luminance;
 };
 
+/**
```
Parsium, line 30: **praise**: nice job for repurposing this function, improving its documentation and test coverage :+1:

src/style/utils/get-hex-value.ts
```diff
@@ -0,0 +1,298 @@
...
+    const alpha = Math.round(a * 255)
+      .toString(16)
+      .padStart(2, "0");
+    hex += alpha;
+  }
+
+  return hex;
+}
+
+const COLOR_NAMES: Record<string, string> = {
```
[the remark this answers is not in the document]
Parsium: >suggestion: rather than having to maintain a list of colours, we could use something like the follows:

it might be worth verifying if this doesn't cause issues for apps using Server Side Rendering.

As it's likely a React component would be relying on this function during rendering, there's a risk an error may be thrown since `document` wouldn't exist in a server environment ([link to example](https://www.npmjs.com/package/eslint-plugin-ssr-friendly#no-dom-globals-in-react-fc))

src/style/utils/get-hex-value.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  const namedHex = namedColorToHex(c);
+  if (namedHex) {
+    return namedHex;
+  }
+
+  // eslint-disable-next-line no-console
+  console.error(
+    `Could not parse "${color}", please provide a valid hex, rgb, rgba, hsl, hsla or named color.`,
+  );
```
Parsium: **issue**: there's a chance the error message could appear in production if we use `console` directly.

Alternatively, what do you think about using `Logger` for this? This should achieve the same result, but messages will only be logged in development environments: 

```suggestion
  Logger.error(
    `Could not parse "${color}", please provide a valid hex, rgb, rgba, hsl, hsla or named color.`,
  );
```

src/components/pill/pill.style.ts
```diff
@@ -56,30 +57,24 @@ const StyledPill = styled.span<AllStyledPillProps>`
     let buttonFocusColor;
     let contentColor;
```
Parsium: **suggestion**: what do you think of typing these variables, so we can check that they are defined before they are used in the style rules below?:


```suggestion
    let pillColor: string;
    let buttonFocusColor: string;
    let contentColor: string;
```

src/components/pill/pill.style.ts
```diff
@@ -56,30 +57,24 @@ const StyledPill = styled.span<AllStyledPillProps>`
...
+    if (borderColor) {
+      pillColor = toColor(theme, borderColor);
+
+      // get token value in rgb
+      const colorVal = getColorValue(pillColor);
+      contentColor = getAccessibleForegroundColor(
+        getHexValue(colorVal),
+        false,
+        true,
       );
```
Parsium, line 69: **comment**: replacing our old use of `polished` with the util fixes a couple of colour contrast violations in a `Pill` story, great job! :+1:

As there may be differences in how `polished` calculated its contrast checks compared to our util, I wonder if it would be worth making the commit a fix. As there may be some text colours that differ from the current Carbon version, it might help the QAs if they check that no visual regressions exist. What do you think?

src/style/utils/get-hex-value.ts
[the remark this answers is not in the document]
nuria1110: Even though listing the colours may not be ideal, I was unsure of the issues that could pop up with SSR/testing by using the canvas so I decided to manually convert to hex instead, happy to look into further into this method if its preferred

src/components/pill/pill.style.ts
```diff
@@ -178,18 +172,45 @@ const StyledPill = styled.span<AllStyledPillProps>`
             ${StyledIcon} {
               color: ${contentColor};
             }
+
+            ${borderColor
```
Parsium, line 176: **question**: did you notice any issues with setting the background colour directly on the icon button? I thought I would ask, as it seems simpler than using a CSS pseudo-selector.

nuria1110: Yes, unfortunately the CSS filter is applied on the whole element, meaning the cross Icon and the focus outline would also have a dark filter over them on hover/focus instead of only the background, so I used a pseudo-selector so that I could make sure it was only getting applied to the background :)

Parsium: Ah makes sense, thanks for explaining this 👍🏼

paulrobinson88 (approved):

Parsium (approved):

damienrobson-sage (approved):

PR #7356, opened 2025-05-30
feat(useDraggable): add public facing hook

tomdavies73:
**FYI:** All commits will be squashed into one final commit, with any relevant issues also linked.

### Proposed behaviour

This PR introduces a comprehensive drag-and-drop system built on top of Atlassian's Pragmatic Drag & Drop library. 

For information about the Pragmatic Drag & Drop library, [see here](https://atlassian.design/components/pragmatic-drag-and-drop/about).

The implementation provides a flexible, reusable solution for creating draggable lists with automatic reordering capabilities.

Consumers can now use the `useDraggable` hook, which will be externally available to make any array of React Elements draggable.

#### Key Components Added:

**Core Architecture:**
- `DraggableContainer` - Main container component that manages drag operations and list state
- `DraggableItem` - Individual draggable item wrapper with drop target capabilities  
- `useDraggable` - Hook that simplifies usage by wrapping items automatically
- Context providers for sharing state between container and items

Screenshot 2025-05-30 at 16 56 37

**Key Features:**

**Dual Drag Modes:**
   - `continuous` - Items reorder in real-time as user drags
   - `onDrop` - Items only reorder when dropped (better performance for large lists)

**Smart State Management:**
   - Automatic list synchronisation when items are added/removed
   - Rollback to previous state on invalid drops (outside valid targets)
   - Tracks original order for reset functionality

**Robust ID Handling:**
   - Automatic unique ID generation for items
   - Duplicate ID detection with console warnings
   - Parent ID resolution for nested elements

**Imperative API:**
   - Exposed `reOrder(itemId, toIndex)` method for programmatic moves
   - Supports both direct item IDs and child element IDs

**Edge Detection:**
   - Visual feedback for drop zones (top/bottom edges)
   - Prevents dropping items onto themselves
   - Container boundary validation

**Technical Implementation:**

- Built on `@atlaskit/pragmatic-drag-and-drop` for reliable cross-browser support
- Uses React Context for efficient state sharing
- Implements `useImperativeHandle` for external control

**Benefits:**
- **Developer Experience:** Simple hook-based API with sensible defaults
- **Flexibility:** Supports custom container/item elements and props
- **Reliability:** Comprehensive error handling and state recovery
- **Performance:** Efficient updates with minimal re-renders
- **Optomisation:** Consumers will no longer have to install third-party draggable libraries and can instead use the `useDraggable` hook directly if needed. There has also been a major net improvement in bundle size within Carbon and Carbon components. Our current draggable library `react-dnd` has a minified size of `33.9kb` whereas `pragmatic-drag-and-drop` has a minified size of `176b` which is a x200 reduction in size.

Within our components, the biggest win is in `FlatTableBodyDraggable` with a with 52% reduction (~3KB saved). Overall, there has been a net saving across affected components of 3.7KB. The public facing hook`useDraggable` has a rendered size of `815B` making it an attractive alternative for consumers.
- **Accessibility:** Built on a proven drag-and-drop foundation, the `useImperativeHandle` can be used for manual re-ordering. Supporting graceful degradation.

This implementation provides a production-ready drag-and-drop solution that can be easily integrated into existing components while maintaining full control over styling and behaviour.

**Coming soon:**

Once this work is completed and available in Carbon, a `DraggableProvider` will be developed which can be used to wrap multiple containers returned by `useDraggable` hook(s) which will facilitate the movement of items across containers.

Once this is available and we can update uses of the `DndProvider` with this new provider in Carbon, the `react-dnd` depdendency will be removed.

### Current behaviour

Currently, consumers have to use third-party draggable libraries to build out draggable functionality. Internally, we use and re-use `react-dnd` which is now out of support and results in code repetition and complexity.

### Checklist

#### QA


playwright-ct.config.ts
```diff
@@ -8,7 +8,7 @@ const playwrightDir = resolve(__dirname, "./playwright");
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
-  testDir: resolve(__dirname, "./src/components"),
+  testDir: resolve(__dirname, "./src/"),
```
DipperTheDan, line 11: **question(non-blocking):** Why has this been changed?

src/__internal__/draggable/draggable-container.tsx
```diff
@@ -0,0 +1,476 @@
...
+              if (
+                !Number.isNaN(indexOfTarget) &&
+                indexOfTarget >= 0 &&
+                destinationId !== undefined &&
+                destinationId !== null &&
+                (lastMoveRef.current.indexOfTarget !== indexOfTarget ||
+                  lastMoveRef.current.destinationId !== destinationId)
+              ) {
+                localMove(destinationId, indexOfTarget);
+                lastMoveRef.current = { indexOfTarget, destinationId };
```
DipperTheDan, line 376: **suggestion(non-blocking):** Something like this here might make this a bit easier to read:

```
const isValidIndex = Number.isInteger(indexOfTarget) && indexOfTarget >= 0
const hasDestination = destinationId != null
const isNewMove =
  lastMoveRef.current.indexOfTarget !== indexOfTarget ||
  lastMoveRef.current.destinationId !== destinationId

if (isValidIndex && hasDestination && isNewMove) {
  localMove(destinationId, indexOfTarget)
  lastMoveRef.current = { indexOfTarget, destinationId };
}
```
Obviously feel free to choose better names for the `const` declarations.

src/__internal__/draggable/draggable-container.tsx
```diff
@@ -0,0 +1,476 @@
...
+                | string
+                | number;
+
+              if (
+                !Number.isNaN(indexOfTarget) &&
+                indexOfTarget >= 0 &&
+                destinationId !== undefined &&
+                destinationId !== null &&
+                lastMoveRef.current.indexOfTarget !== indexOfTarget
+              ) {
```
DipperTheDan, line 409: **suggestion(non-blocking):** Same thing here. It might be best to make `const` declarations where you can to make these easier to read.

src/__internal__/draggable/draggable-item.test.tsx
```diff
@@ -0,0 +1,571 @@
...
+        Apple
+      </DraggableItem>,
+    );
+
+    expect(mockObjectRef.current).toBeInstanceOf(HTMLElement);
+    expect(mockObjectRef.current).not.toBeNull();
+  });
+
+  it("handles ref prop when ref is null", () => {
+    // This should not throw an error
```
DipperTheDan, line 86: **question(non-blocking):** You have the `.not.toThrow` check here. Is the comment meant to be here or was this just left over from dev work?

src/__internal__/draggable/draggable-item.test.tsx
```diff
@@ -0,0 +1,571 @@
...
+      render(
+        <DraggableItem ref={null} id="Apple" itemProps={itemDataRole}>
+          Apple
+        </DraggableItem>,
+      );
+    }).not.toThrow();
+  });
+
+  it("handles ref prop when ref is undefined", () => {
+    // This should not throw an error
```
DipperTheDan, line 97: Same here with regards to the comment being here.

src/__internal__/draggable/draggable-item.tsx
```diff
@@ -0,0 +1,189 @@
...
+      }
+    };
+
+    /**
+     * After mount, check if the first child has an ID that we can use
+     * This supports scenarios where the child already has its own ID
+     */
+    useLayoutEffect(() => {
+      // This runs after the component mounts and the children are rendered
+      if (itemRef.current && itemRef.current.children.length > 0) {
```
DipperTheDan, line 72: **suggestion(non-blocking):** You should be able to do this here:
```suggestion
      if (itemRef?.current.children.length > 0) {
```

src/__internal__/draggable/draggable-item.tsx
```diff
@@ -0,0 +1,189 @@
...
+          onDrop() {
+            setDragState(idle);
+          },
+        }),
+      );
+
+      return () => cleanup();
+    }, [uniqueId, draggableItemData, setDragState]);
+
+    return React.createElement(
```
DipperTheDan, line 172: Just an FYI here, that `createElement` is classed as a [legacy React API](https://react.dev/reference/react/legacy). React themselves state that `they are not recommended for use in newly written code`.

src/components/draggable/draggable.test.tsx
```diff
@@ -151,6 +173,36 @@ test("items are reordered when their order is manually changed", () => {
   expect(allItems[1]).toHaveTextContent("Apple");
 });
 
+test("items are reordered when their order is manually changed via imperative reo-order", () => {
```
DipperTheDan, line 176:
```suggestion
test("items are reordered when their order is manually changed via imperative re-order", () => {
```

src/__internal__/draggable/draggable-container.tsx
```diff
@@ -0,0 +1,476 @@
...
+      () => ({
+        columnId: uniqueId,
+        dragType,
+      }),
+      [uniqueId, dragType],
+    );
+
+    return (
+      <DraggableContainerContext.Provider value={contextValue}>
+        {React.createElement(
```
DipperTheDan, line 455: I think I've mentioned it elsewhere but `createElement` is part of React's legacy API's now and they state that `they are not recommended for use in newly written code`.

playwright-ct.config.ts
tomdavies73: So playwright can run on the `.pw.tsx` file added to the `useDraggable` hook, previously it would only run over components which would mean this test wouldn't be included.

src/__internal__/draggable/draggable-item.test.tsx
tomdavies73: Just looks like a leftover comment - I'll remove as the assertion is pretty self-descriptive

src/__internal__/draggable/draggable-item.test.tsx
tomdavies73: Just looks like a leftover comment - I'll remove as the assertion is pretty self-descriptive

src/__internal__/draggable/draggable-item.tsx
tomdavies73: Good spot mate, we can look at other alternatives which will let us create an element on the fly based on a prop - most likely jSX

src/__internal__/utils/helpers/guid/guid.test.ts
```diff
@@ -9,3 +9,98 @@ test("should generate a unique guid each time it is called", () => {
   const uniqueGuids = new Set(guids);
   expect(uniqueGuids.size).toEqual(5);
 });
+
+describe("isGuid", () => {
```
nineteen88, line 13: suggestion: remove redundant describe block and be more descriptive of tests/describe

We have an unnecessary double describe block going on here. It can be flattened to simplify the depth and just tidy up the describe blocks to be more descriptive instead. Revert back to continuous prose if possible.

e.g. 
```
describe("when a valid GUID string is passed", () => {
  it("should return true if the string is all zeros", () => {
    ...
  });
});
```

src/__internal__/draggable/draggable-item.tsx
tomdavies73: I'm leaning towards <div as={...}>

src/components/draggable/draggable.stories.tsx
```diff
@@ -30,34 +32,90 @@ type Story = StoryObj<typeof DraggableContainer>;
...
     </DraggableItem>
-    <DraggableItem key="4" id={4}>
+    <DraggableItem key="4" id="4">
       Some content goes here
     </DraggableItem>
   </DraggableContainer>
 );
 DefaultStory.storyName = "Default";
 
+export const ManualReOrdering: Story = () => {
```
Parsium, line 51: **suggestion**: could we consider removing the imperative handle and have consumers control the item order in state?

It seems possible to achieve manual reordering without the imperative handle (see example below). This approach would be simpler for consumers by avoiding needing to use refs, what do you think about using this instead?

We could even export a `reorder` util for consumers if we want to provide extra convenience, but this might be overkill.

```tsx
const [orderedIds, setOrderedIds] = useState<string[]>([
  "first",
  "second",
  "third",
  "fourth",
]);

// Swaps the position of two elements in an array. Returns result as a new array.
const reorder = <T,>(array: T[], index1: number, index2: number) => {
  const newArray = Array.from(array);

  const [removed] = newArray.splice(index1, 1);
  newArray.splice(index2, 0, removed);

  return newArray;
};

return (
  <DraggableContainer
    getOrder={(ids) => setOrderedIds((ids || []).filter(Boolean).map(String))}
    flexDirection="row-reverse"
  >
    {orderedIds.map((id) => (
      <DraggableItem key={id} id={id}>
        <Box
          display="flex"
          width="100%"
          justifyContent="space-between"
          alignItems="baseline"
        >
          Some {id} content goes here
          <ActionPopover m={0}>
            <ActionPopoverItem
              onClick={() =>
                setOrderedIds((prev) => {
                  const idx = prev.indexOf(id);
                  return reorder(prev, idx, 0);
                })
              }
            >
              Move Top
            </ActionPopoverItem>
            <ActionPopoverItem
              onClick={() =>
                setOrderedIds((prev) => {
                  const idx = prev.indexOf(id);
                  return reorder(prev, idx, Math.max(0, idx - 1));
                })
              }
            >
              Move Up
            </ActionPopoverItem>
            <ActionPopoverItem
              onClick={() =>
                setOrderedIds((prev) => {
                  const idx = prev.indexOf(id);
                  return reorder(
                    prev,
                    idx,
                    Math.min(idx + 1, prev.length - 1),
                  );
                })
              }
            >
              Move Down
            </ActionPopoverItem>
            <ActionPopoverItem
              onClick={() =>
                setOrderedIds((prev) => {
                  const idx = prev.indexOf(id);
                  return reorder(prev, idx, prev.length - 1);
                })
              }
            >
              Move Bottom
            </ActionPopoverItem>
          </ActionPopover>
        </Box>
      </DraggableItem>
    ))}
  </DraggableContainer>
);
```

src/components/flat-table/flat-table-body-draggable/flat-table-body-draggable.test.tsx
```diff
@@ -361,14 +387,65 @@ it("calls getOrder callback when the order is changed and getOrder prop is set",
...
+
+  act(() => {
+    jest.runAllTimers();
+  });
+
+  expect(getOrder).toHaveBeenCalledWith(["1", "0", "2"], "0");
+});
+
+test("accepts a ref and allows imperative reordering of draggable table rows", () => {
+  jest.useFakeTimers();
```
Parsium, line 408: **issue**: we should flush and restore the real timers at the end of tests that use fake timers, since [`jest.useFakeTimers` fakes timers for the whole test file](https://jestjs.io/docs/jest-object#fake-timers)

```tsx
// before test
jest.useFakeTimers();

// after test
jest.runOnlyPendingTimers();
jest.useRealTimers();
```

This will help ensure any fake and unfinished timers don't leak into other tests.

src/components/draggable/draggable.test.tsx
```diff
@@ -17,8 +23,10 @@ test("dragging an item and dropping it after another item within the target cont
     </DraggableContainer>,
   );
 
-  const apple = screen.getByText("Apple");
-  const venus = screen.getByText("Venus");
+  // eslint-disable-next-line testing-library/no-node-access
```
Parsium, line 26: **polish**: to avoid ignoring the linter, could we set some test ids on the respective elements?

src/components/flat-table/flat-table.stories.tsx
```diff
@@ -2770,6 +2774,100 @@ export const WithDraggableRows: Story = () => {
 };
 WithDraggableRows.storyName = "With Draggable Rows";
 
+export const WithDraggableRowsAndManualReOrdering: Story = () => {
```
Parsium, line 2777: **suggestion**: similarly to this story's `Draggable` counterpart, we should be able to use state to achieve this, so it would be worthwhile using it over refs to simplify the approach.

src/components/flat-table/flat-table.pw.tsx
```diff
@@ -2099,6 +2102,90 @@ test.describe("Prop tests", () => {
...
 
+  test("should apply correct specific dragging background colour", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<FlatTableDraggingRows data-drag-state="is-dragging" />);
+    const draggableItem = page.locator("td").filter({
+      hasText: "UK",
+    });
+
```
Parsium, line 2113: **polish**: if you would like to more closely resemble a user in the middle of a drag operation, the [Playwright's mouse API](https://playwright.dev/docs/api/class-mouse#mouse-move) could come in handy:

```tsx
await draggableItem.hover();
await page.mouse.down();
await page.mouse.move(1, 1); // start drag operation
```

This would allow you to represent this operation without having to set `data-drag-state` manually.

src/hooks/useDraggable/use-draggable.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
+  });
+
+  return draggableElement;
+};
+GetOrderCallback.storyName = "getOrder callback";
+
+export const ManualReOrdering: Story = () => {
+  const draggableHandle = useRef<UseDraggableHandle | null>(null);
+  const [currentOrder, setCurrentOrder] = useState<number[]>([
+    1, 2, 3, 4, 5, 6,
```
Parsium, line 157: **suggestion**: similarly to this story's `Draggable` counterpart, we should be able to use state to achieve this:

```tsx
const [orderedIds, setOrderedIds] = useState<number[]>([1, 2, 3, 4, 5, 6]);

const reorder = <T,>(array: T[], index1: number, index2: number) => {
  const newArray = Array.from(array);

  const [removed] = newArray.splice(index1, 1);
  newArray.splice(index2, 0, removed);

  return newArray;
};

const placementControls = (id: number) => (
  <ActionPopover m={0}>
    <ActionPopoverItem
      onClick={() => {
        setOrderedIds((prev) => {
          const idx = prev.indexOf(id);
          return reorder(prev, idx, 0);
        });
      }}
    >
      Move Top
    </ActionPopoverItem>
    <ActionPopoverItem
      onClick={() => {
        setOrderedIds((prev) => {
          const idx = prev.indexOf(id);
          return reorder(prev, idx, Math.max(0, idx - 1));
        });
      }}
    >
      Move Up
    </ActionPopoverItem>
    <ActionPopoverItem
      onClick={() => {
        setOrderedIds((prev) => {
          const idx = prev.indexOf(id);
          return reorder(prev, idx, Math.min(idx + 1, prev.length - 1));
        });
      }}
    >
      Move Down
    </ActionPopoverItem>
    <ActionPopoverItem
      onClick={() => {
        setOrderedIds((prev) => {
          const idx = prev.indexOf(id);
          return reorder(prev, idx, prev.length - 1);
        });
      }}
    >
      Move Bottom
    </ActionPopoverItem>
  </ActionPopover>
);

const items = orderedIds.map((id) => (
  <Box
    display="flex"
    alignItems="baseline"
    justifyContent="space-between"
    padding="2px"
    margin="2px"
    width="80px"
    id={id.toString()}
    key={id}
  >
    item {id}
    {placementControls(id)}
  </Box>
));

const { draggableElement } = useDraggable({
  draggableItems: items,
  getOrder: (ids) => setOrderedIds((ids || []).filter(Boolean).map(Number)),
});

return draggableElement;
```

src/hooks/useDraggable/use-draggable.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
+  const { draggableElement } = useDraggable({
+    draggableItems: items,
+    dragType: "onDrop",
+  });
+
+  return draggableElement;
+};
+DragType.storyName = "Drag Type";
+
+export const WithIds: Story = () => {
```
Parsium, line 114: **question**: as consumers need to pass in pre-constructed JSX elements into the hook, have we considered how to accommodate those who need control over each item's UI?

For example, if a consumer needs to implement a collection of draggable items, each with a dedicated drag handle (part of the item that can be dragged in order to drag the whole item), they would need to be able to attach listeners to an item's inner UI.

Is there a mechanism that allows them to do this? How easy would it be to amend the API if we needed to introduce such a feature?

src/hooks/useDraggable/use-draggable.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+## How it Works
+
+**Simple concept**: Every draggable item lives in a container. Pass an array of items to `useDraggable`, and it returns a container with those items now draggable.
+
+**Smart wrapping**: The hook wraps content while preserving its structure—text, components, or complex elements all work seamlessly.
+
+**Zero configuration**: IDs are optional. The hook auto-generates GUIDs or uses existing element IDs.
+
+## Key Attributes
```
Parsium, line 25: **polish**: what do you think about moving the notes on key attributes into a later section on customisation?

This might help improve the flow of the docs, by starting off with the basics then diving into more advanced capabilities.

src/hooks/useDraggable/use-draggable.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+### Custom Elements
+
+Change the container and item HTML elements:
+
+<Canvas of={useDraggableStories.ContainerAndItemsNode} />
+
+### Visual Feedback
+
+Style based on drag state using the auto-generated attributes:
```
Parsium, line 99: **suggestion**: it might be worth stating that we *only* recommend using it for components that consumers control and not with Carbon components.

If we are providing data tags for consumers to use for custom styling, we should be careful. As we discourage consumers from targeting the DOM structure of Carbon components, this design choice could be seen as a contradiction in our own guidance and cause confusion.

We could even link to the "Extending Styles of Carbon components" docs as an explanation why.

src/hooks/useDraggable/use-draggable.stories.tsx
Parsium, line 1: **issue**: looking through `useDraggable` and its use of the new internal `DraggableContainer` and `DraggableItem` components, I'm wary about us requiring consumers to pass items as pre-constructed JSX, then storing them in state.

Interlinking the JSX and data in this way seems more complex than necessary, and it could impact the hook's:

1. **Flexibility** - more difficult for consumers to control item UI, such as attaching refs or listeners, or conditionally rendering items. It could also make it difficult to attach new behaviour to each item in the future.
2. **Rendering performance** - `DraggableContainer` must store the JSX items in state, which requires triggering re-renders to register them in state. 
3. **Maintainability** - manipulating children and storing them in state may lead to fragile code since the data flow is harder to understand, and consumers/Carbon devs might not expect that children are modified. We've experienced similar fragility issues with `Select` that also manipulates child JSX elements.

___

Since consumers will likely be fetching raw data for each item:

```tsx
const [items, setItems] = useState([
  { id: "0", label: "Apple" },
  { id: "1", label: "Banana" },
  { id: "2", label: "Cherry" },
]);
```

We could address the above by amending `useDraggable` so:
1. Its used in a controlled manner (use the item state from the consumer, rather than re-create it ourselves).
2. It returns ref callbacks for registering containers and items.

Here's a rough example of what this might look like:

```tsx
const [items, setItems] = useState([
  { id: "0", label: "Apple" },
  { id: "1", label: "Banana" },
  { id: "2", label: "Cherry" },
]);

const { 
  setContainerRef,  // ref callback for registering the container element with Pragmatic
  setItemProps      // function that accepts an item, then returns ref callback for registering that item's element with Pragmatic
} = useDraggable({
  items,
  itemKey: "id"
  onOrderChange: (reorderedItems) => setItems(reorderedItems)
});

return (
  <table>
    <thead>
      <tr>Fruit</tr>
    </thead>
    <tbody ref={setContainerRef}>
      {items.map(item => (
        <tr key={item.id} ref={setItemProps(item.id)}>
          <td>{item.label}</td>
        </tr>
      ))}
    </tbody>
  </table>
);
```

This would allow us to maintain separation of concerns between UI and data by relating the item order to the data itself instead of the UI, while still providing control over the element types of the container and items.

I appreciate this suggestion is high-level and skips some technical details, I have done some experimentation on the side, so I'm happy to share this and discuss more with you and the other reviewers.

Issue #6671, opened 2024-04-11
[Textbox] - onClick handler is still being passed when the component is disabled


sianford:
### Description

When the component is disabled, NVDA screen reader registers that the input field is clickable with a mouse/trackpad. 

### Reproduction

https://carbon.sage.com/?path=/story/textbox--disabled

### Steps to reproduce

1. Open the Textbox Disabled story in Windows.
2. Open NVDA with the Speech Viewer switched on.
3. Click on the content of the disabled Textbox field.
4. Inspect the Speech Viewer window.

Expected result:
There is no response when the field is clicked on. This is not accessible at all using the keyboard (which is expected behaviour).

Actual result:
The Speech Viewer states the field is "clickable".

### JIRA ticket numbers (Sage only)

SKPU-11421

### Suggested solution

Remove the onClick handler for disabled inputs.

### Carbon version

132.1.0

### Relevant browsers

Firefox, Chrome, Microsoft Edge

### Relevant OSs

Windows

### Additional context

VoiceOver used with Safari on Mac also states "To enter text in this feild, type." indicating it also detects the field is accessible. 

### Confidentiality

- [X] I confirm there is no confidential or commercially sensitive information included.

edleeks87:
We are still passing a [handler to the underlying input](https://github.com/Sage/carbon/blob/9f9198475abd930428cdd448c911241a8b3feabe/src/__internal__/input/input.component.tsx#L254), we should add a ternary like we have in [Textbox](https://github.com/Sage/carbon/blob/9f9198475abd930428cdd448c911241a8b3feabe/src/components/textbox/textbox.component.tsx#L290) to solve this

edleeks87:
FE-6430

PR #7357, opened 2025-06-02
fix(input): nvda announces click when input disabled

mihai-albu-sage:
fixes: #6671

### Proposed behaviour

The onClick handler is no longer passed down to the input if the input is disabled or readOnly.

### Current behaviour

The onClick handler is passed to the input even if the input is disabled or readOnly.

### Checklist

#### QA


src/__internal__/input/input.test.tsx
```diff
@@ -253,3 +253,22 @@ test("when both `aria-describedby` and `ariaDescribedBy` are provided, `aria-des
   const input = screen.getByRole("textbox");
   expect(input).toHaveAttribute("aria-describedby", "description");
 });
+
+test.each(["disabled", "readOnly"])(
+  "does not call onClick handler when input is %s and is clicked",
+  async (propName) => {
```
edleeks87: suggestion: this test currently doesn't need to be `async` if there's no `await` for any actions or outcomes etc. I do think it might be worth keeping it async and matching how the inverse test is done using `await user.click(input)`

src/__internal__/input/input.test.tsx
```diff
@@ -253,3 +253,22 @@ test("when both `aria-describedby` and `ariaDescribedBy` are provided, `aria-des
...
+    const onClickMock = jest.fn();
+    render(
+      <Input
+        onFocus={onClickMock}
```
tomdavies73: **suggestion** (non-blocking): I feel like we're not getting much value by making this a `test.each` testing both `disabled` and `readOnly` here. We could just split these into two separate tests, will reduce complexity and enhance readability.

But if you disagree, I'm happy to approve if you just address @edleeks87's suggested change

src/__internal__/input/input.test.tsx
mihai-albu-sage: The tests have been kept async as suggested.

src/__internal__/input/input.test.tsx
mihai-albu-sage: The test has been split into two separate tests for better readability.

paulrobinson88 (approved):

PR #7040, opened 2024-10-25
feat(date, date-range): provides a mechanism that allows consumers to override the format of dates

damienrobson-sage:
A new property has been added to the `Locale` interface for dates to allow customers to override the displayed date formats in both the Date and DateRange inputs

fix #6930 

### Proposed behaviour

Provide a mechanism to allow default date formats to be overridden

### Current behaviour

Currently, users are forced to adhere to the date formats decreed by their chosen locale, with the only option to change being to use a different locale.

### Checklist

- [X] Commits follow our style guide
- [X] Related issues linked in commit messages if required
- [ ] Screenshots are included in the PR if useful
- [x] All themes are supported if required
- [X] Unit tests added or updated if required
- [ ] Playwright automation tests added or updated if required
- [X] Storybook added or updated if required
- [ ] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [ ] Typescript `d.ts` file added or updated if required
- [X] Related docs have been updated if required

#### QA

- [ ] Tested in provided StackBlitz sandbox/Storybook
- [ ] Add new Playwright test coverage if required
- [ ] Carbon implementation matches Design System/designs
- [ ] UI Tests GitHub check reviewed if required

### Testing instructions

Render one of the following component:

**DateInput**

```javascript
<I18nProvider
        locale={{
          locale: () => "de-DE",
          date: {
            dateFnsLocale: () => de,
            ariaLabels: {
              previousMonthButton: () => "Vorheriger Monat",
              nextMonthButton: () => "Nächster Monat",
            },
            dateFormatOverride: "yyyy-M-d",
          },
        }}
      >
        <DateInput
          label="Date `DE` locale"
          value={state}
          onChange={handleChange}
        />
      </I18nProvider>
```

**DateRange**

```javascript
      <I18nProvider
        locale={{
          locale: () => "de-DE",
          date: {
            dateFnsLocale: () => de,
            ariaLabels: {
              previousMonthButton: () => "Vorheriger Monat",
              nextMonthButton: () => "Nächster Monat",
            },
            dateFormatOverride: "y-m-ddd",
          },
        }}
      >
        <DateRange
          startLabel="Start"
          endLabel="End"
          value={state}
          onChange={handleChange}
        />
      </I18nProvider>
```
Change the value of `dateFormatOverride` and observe that the changes are correctly reflected in the inputs

PR #7359, opened 2025-06-02
fix(date, date-range): ensure date format override returns correct formatted value

### LINKED ISSUE #7345: Custom date format is changing date swapping day and month
### Description

Using `DateInput` and `DateRange` with the `dateFormatOverride` prop to customize format I see date is changed. When selecting 12th May 2025 it then changes it to 5th December 2025.

I expect the date not to be changed.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-xe8krcwx?file=src%2FApp.tsx

### Steps to reproduce

_No response_

_No response_

### Suggested solution

_No response_

### Carbon version

153.0.0

### React version

v17

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #7345

### Proposed behaviour

- Both `dateFormatOverride` prop and translation key can be used, with the prop taking precedence.
- Format override is respected regardles of default locale format. 

https://github.com/user-attachments/assets/a19a265c-11db-49ae-982a-c46c866510e1

### Current behaviour

- `dateFormatOverride` prop has no effect.
- When using the `dateFormatOverride` translation key, it fails to format values correctly and parse correct raw value for the `onChange` event, specially when the day and month are flipped.

Using de-DE locale:

https://github.com/user-attachments/assets/990ef2b9-d929-4408-97bb-4f9c396f10cf

### Checklist

#### QA

### Testing instructions

- Can be tested in `i18n Story` test story in storybook and changing the `dateFormatOverride`, see actions for returned formatted and raw values. 
- Separator formatting should also be supported with format override: e.g. `03-06-2019` should update to `03/06/2019` if date format is `dd/MM/yyyy`.


src/components/date/__internal__/date-formats/index.ts
nuria1110, line 1: Looks like James was correct [here](https://github.com/Sage/carbon/pull/7040#discussion_r1824306575) and we should be generating the list of formats to match from the override string and not from the locale format, which is what was causing the mismatch when day and month values were swapped.

src/components/date/date-test.stories.tsx
```diff
@@ -240,12 +266,18 @@ MultipleDates.storyName =
...
 
-export const I18NStory = (args: I18nArgs) => {
+const locales = {
+  "en-US": enUS,
+  "en-GB": enGB,
+  "zh-CN": zhCN,
+  "de-DE": de,
+};
+
+export const I18NStory = ({ locale, ...args }: DateInputI80NProps) => {
```
Parsium, line 280: **praise**: nice idea using Storybook Controls to change the locale for testing :+1:

src/components/date/__internal__/date-formats/date-formats.test.ts
```diff
@@ -511,33 +511,59 @@ test("should default to en-GB locale if no locale code string passed to `getForm
...
-
-      expect(
-        expectedFormats.every((formatStr) => formats.includes(formatStr)) &&
-          formats.length === expectedFormats.length,
-      ).toEqual(true);
-
-      expect(format).toEqual(dateFormatOverride);
-    },
-  );
+  test("should return the expected formats when `dateFormatOverride` is passed as NA locale", () => {
```
edleeks87: suggestion: `test` -> `it` when they're in `describe` blocks

src/components/date/date.pw.tsx
```diff
@@ -868,6 +868,41 @@ test.describe("Functionality tests", () => {
       await page.hover("body"); // hover on body to close the tooltip
     });
   });
+
+  test("should format the date input value according to `dateFormatOverride` prop when using the date-picker", async ({
```
edleeks87: comment (non-blocking): I'm not sure we need these to be PW tests as there's nothing browser specific we're relying on etc, not the end of the world having them though

nuria1110: Can't remember why I added them in the first place, removed them

dariusbercea-sage (approved):

PR #7360, opened 2025-06-03
fix(note): render title semantically with h3 when passed as text

DobroTora:
### Proposed behaviour
title prop in Note component renders using Typography with `variant="h3"`

### Current behaviour

Styling for H3 for a title prop is being overwritten in a Note component within `with title` story.

### Checklist

#### QA

### DIFF
--- a/src/components/note/note.component.tsx
+++ b/src/components/note/note.component.tsx
@@ -96,7 +96,7 @@ export const Note = ({
                   fontSize="16px"
                   lineHeight="21px"
                   paddingBottom="16px"
-                  variant="h2"
+                  variant="h3"
                 >
                   {title}
                 </Typography>
--- a/playwright/components/note/index.ts
+++ b/playwright/components/note/index.ts
@@ -11,7 +11,7 @@ const noteComponent = (page: Page) => {
   return page.locator(NOTE_COMPONENT);
 };
 const noteHeader = (page: Page) => {
-  return page.locator(NOTE_COMPONENT).locator("h2");
+  return page.locator(NOTE_COMPONENT).locator("h3");
 };
 const noteContent = (page: Page) => {
   return page.locator(DATA_CONTENTS);
--- a/src/components/note/note.test.tsx
+++ b/src/components/note/note.test.tsx
@@ -31,7 +31,7 @@ test("renders a Typography component with h2 `variant` and `title` as its child 
     <Note createdDate="23 May 2020, 12:08 PM" noteContent="" title="Title" />,
   );
 
-  const titleElement = screen.getByRole("heading", { level: 2 });
+  const titleElement = screen.getByRole("heading", { level: 3 });
 
   expect(titleElement).toHaveTextContent("Title");
   expect(titleElement).toHaveAttribute("data-role", "note-title");


damienrobson-sage (dismissed):

src/components/note/note.style.ts
```diff
@@ -53,13 +53,11 @@ const StyledInlineControl = styled.div`
...
   ${VARIANT_TYPES.map(
     (variant) => `
-      ${variant}{
-        font-weight: 700;
-        font-size: 16px;
-        line-height: 21px;
-        padding-bottom: 16px;
-      }
-    `,
+${variant}{
```
nineteen88: comment: should set the styles to default to those of h3 now

If we are to still support this behaviour then this should be defaulted to the h3 styling (previously it was h2 styling). It might be worth speaking to the team to ensure that we want to keep this behaviour as it's not completely clear from the ticket.

We also need to ensure the DS is updated as this no longer matches what is supposed to be the source of truth. So, it could be worth speaking to @harpalsingh to make sure that gets updated if this is to be correct. I'd be reluctant to approve this until it does match the source of truth.

nineteen88: In fact, upon reading the ticket again it seems it was mentioned that the override styling for the Node that is passed in should be completely removed.

`Should remove the Note specific H3 styling, if user passes text, render a h3 typography components, otherwise render the Node`

Could we please remove that styling block in that case and prompt @harpalsingh to update the Design System if Note should use H3 so Designers and Developers are both using the same styles?

damienrobson-sage (dismissed):

src/components/note/note.style.ts
```diff
@@ -1,6 +1,6 @@
 import styled, { css } from "styled-components";
 import { margin } from "styled-system";
-import applyBaseTheme from "../../style/themes/apply-base-theme";
```
nineteen88: comment: not sure this change is required

What is in Master should be correct for handling this

src/components/note/note.component.tsx
```diff
@@ -92,11 +92,8 @@ export const Note = ({
               (typeof title === "string" ? (
                 <Typography
                   data-role="note-title"
-                  fontWeight="700"
```
nineteen88: comment: the default styling is incorrect as per the ticket

See the attached for what styling is being output now

Screenshot 2025-06-19 at 15 51 19

src/components/note/note.style.ts
```diff
@@ -51,6 +51,15 @@ const StyledInlineControl = styled.div`
 `;
 
 const StyledTitleWrapper = styled.div`
+  h3 {
```
nineteen88: comment: I'm unsure as to why we've moved styling to be a h3 specific styling only

Previously the styling was passed into the Typography component and it was agreed upon that we would only need to change the variant of the Typography component. What is purpose of this change?

src/components/note/note.component.tsx
```diff
@@ -88,21 +88,17 @@ export const Note = ({
       <StyledNote width={width} {...rest} {...tagComponent("note", rest)}>
         <StyledNoteMain>
           <StyledNoteContent>
-            {title &&
-              (typeof title === "string" ? (
-                <Typography
```
nineteen88: comment: Although you've removed the changes in the style file as mentioned, they aren't being applied to the Typography element anymore

nineteen88:
LGTM

If we can just flatten the commits (should only need one commit) with a descriptive commit message, we're good to go!

damienrobson-sage (approved):

PR #7367, opened 2025-06-06
feat(responsive-vertical-menu): allows custom aria-label for menu launcher

damienrobson-sage:
### Proposed behaviour

- `ResponsiveVerticalMenu` now allows an additional prop to set the value of `aria-label`;
- The `ResponsiveVerticalMenuDivider` is now exported correctly

### Current behaviour

- The launcher button for the `ResponsiveVerticalMenu` is announced via the name of the icon used, with no additional context;
- The `ResponsiveVerticalMenuDivider` isn't exported for use

### Checklist

#### QA

### Testing instructions

Use the `Without Global Header` test story to verify that `Main menu` is announced when focus is gained and a screen reader is active

### DIFF
--- a/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
+++ b/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
@@ -16,8 +16,8 @@ import {
 } from "./responsive-vertical-menu.style";
 
 import Box from "../../box";
-
 import Modal from "../../modal";
+
 import useIsAboveBreakpoint from "../../../hooks/__internal__/useIsAboveBreakpoint";
 import useMediaQuery from "../../../hooks/useMediaQuery";
 
@@ -26,6 +26,7 @@ import tagComponent, {
 } from "../../../__internal__/utils/helpers/tags";
 import { DepthProvider } from "./__internal__/depth.context";
 import { MenuFocusProvider } from "./__internal__/focus.context";
+import useLocale from "../../../hooks/__internal__/useLocale";
 
 export interface ResponsiveVerticalMenuProps extends TagProps {
   /** The content of the menu */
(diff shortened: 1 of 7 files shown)


mihai-albu-sage (dismissed):

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
```diff
@@ -41,6 +43,7 @@ export interface ResponsiveVerticalMenuProps extends TagProps {
 const BaseMenu = ({
   children,
   height,
+  launcherButtonAriaLabel = "Main menu",
```
edleeks87: question (non-blocking): is `Main menu` an appropriate default here? Not sure who can decide on that though tbh just wondering if it should be more descriptive of what it is/ what it's doing etc.

edleeks87: suggestion (non-blocking): I wonder if it would be more appropriate not to surface a prop here as this is not being applied to the root element, we have an aria-label for the close button as well and it's highly unlikely there'll be more than one instance of the component on a given page. My thinking is we can use the translations to handle this and provide localisation at the same time for both element's aria-label attribute

edleeks87: comment: Either way I think it's worth adding tests for this so we can catch any regressions at an early stage etc

edleeks87 (approved):
I think it's worth doing the same for [this](https://github.com/Sage/carbon/blob/f7d2db23ebd2dfd3d98f40992781d58c8bb31c3a/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx#L191) before we send it off to the translation team

PR #7368, opened 2025-06-06
feat(storybook): upgrade to v10

nineteen88:
### Proposed behaviour

- update storybook and all dependencies to use v10
- change the way we hide stories to the documented method as per Storybook docs

### Current behaviour

- update is currently on v8

### Checklist

#### QA


.storybook/version-picker/index.tsx
```diff
@@ -4,15 +4,15 @@ import {
   WithTooltip,
   TooltipLinkList,
   TooltipLinkListLink,
-} from "@storybook/components";
+} from "storybook/internal/components";
```
edleeks87, line 6: question: the migration docs indicate anything moved to `internal` might be deprecated in v10, is this the right package to use here and if we have to do we have a ticket/plan for migrating off it asap?

.storybook/manager.ts
```diff
@@ -1,8 +1,8 @@
-import { addons, types } from "@storybook/manager-api";
+import { addons, types } from "storybook/manager-api";
 import sageTheme from "./sage-docs-theme";
 import { ADDON_ID, TOOL_ID } from "./version-picker/constants";
 import { VersionPicker } from "./version-picker";
-import { API_PreparedIndexEntry, API_StatusObject } from "@storybook/types";
+import { API_PreparedIndexEntry } from "storybook/internal/types";
```
edleeks87: Question: is this still needed? If it is, is it the appropriate type to use given that it's now `internal`?

.storybook/version-picker/index.tsx
nineteen88: This is within our addon (version-picker) and the migration docs also mention: 

"Addon authors may continue to use the internal packages, there is currently not yet any replacement."

I might create a ticket to keep an eye on it essentially, but not sure there's much that can be done right now other that rewrite all these components and make them internal for our Storybook only?

edleeks87 (dismissed):

.storybook/version-picker/index.tsx
Parsium: **suggestion**: what do you think about adding a comment briefing explaining this and/or linking to the ticket once its raised?

Might be worth being clear Storybook only recommend addon authors use this API and it shouldn't be used elsewhere.

Parsium (dismissed):

playwright-ct.config.ts
```diff
@@ -2,13 +2,13 @@ import { defineConfig, devices } from "@playwright/experimental-ct-react";
 
 import { resolve } from "path";
 
-const playwrightDir = resolve(__dirname, "./playwright");
+const playwrightDir = resolve(import.meta.dirname, "./playwright");
```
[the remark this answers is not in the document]
edleeks87: agree this is the better approach and will be cross-platform safe doing the following

```
import { fileURLToPath } from 'url';
import { dirname } from 'path';

const __dirname = dirname(fileURLToPath(import.meta.url));

File: playwright-ct.config.ts [hunk 4]
```diff
@@ -2,13 +2,13 @@ import { defineConfig, devices } from "@playwright/experimental-ct-react";
...
 
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
-  testDir: resolve(__dirname, "./src/components"),
+  testDir: resolve(import.meta.dirname, "./src/components"),
```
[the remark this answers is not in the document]
edleeks87: agree again with the suggested approach, whilst it should be fine to assume that Node 21 and up is used we can mitigate fairly easily by following the suggestion

.storybook/manager-head.html
```diff
@@ -5,3 +5,84 @@
...
+
+  [role="toolbar"] button[aria-label*="Locale"],
+  [role="toolbar"] button[aria-label*="Colour Mode"],
+  [role="toolbar"] button[aria-label*="Theme"],
+  [role="toolbar"] button[aria-label*="locale"],
+  [role="toolbar"] button[aria-label*="mode"],
+  [role="toolbar"] button[aria-label*="theme"],
+  [role="toolbar"] button[aria-label*="Locale"] *,
+  [role="toolbar"] button[aria-label*="Colour Mode"] *,
+  [role="toolbar"] button[aria-label*="Theme"] *,
```
edleeks87, line 59: question (non-blocking): would `*` not cover the below where we're targeting `svg` children?

src/components/tooltip/tooltip.component.tsx
```diff
@@ -68,9 +68,9 @@ export interface TooltipProps extends TagProps {
   /** Override font color of the Tooltip, provide any color from palette or any valid css color value. */
   fontColor?: string;
   /**
-   * Overrides the default flip behaviour of the Tooltip,
+   * Overrides the default flip behaviour of the Tooltip, 123
```
edleeks87: question: why not 456? 😉

tsconfig.json
```diff
@@ -14,7 +14,7 @@
       "dom.iterable"
     ],
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
edleeks87, line 17: question (non-blocking): what's the reason for updating the module resolution? I'm fine with the change (probably should have been made a while ago) just wondering if something forced it now

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "carbon-react",
   "version": "158.41.0",
   "description": "A library of reusable React components for easily building user interfaces.",
+  "type": "module",
```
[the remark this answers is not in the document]
tomdavies73: Could `babel.config.mjs` be the go to for this?

src/components/tooltip/tooltip.component.tsx
```diff
@@ -68,9 +68,9 @@ export interface TooltipProps extends TagProps {
...
    * must be an array containing some or all of ["top", "bottom", "left", "right"]
-   * (see https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements)
+   * (see <a href="https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements" target="_blank">Popper.js documentation</a>)
```
tomdavies73, line 73: **praise**: This works nicely in the props table

.storybook/main.ts
```diff
@@ -109,19 +110,29 @@ const config: StorybookConfig = {
...
+    managerHead: (head: string | undefined) => `
+      ${head || ""}
       <meta name="robots" content="noindex">
   `,
   }),
 
   typescript: {
     check: false,
     reactDocgen: "react-docgen-typescript",
+    reactDocgenTypescriptOptions: {
```
tomdavies73, line 131: **question**: when running storybook locally we're met with quite a few Vite warnings for docgen skipping 

```
Vite warning: Skipping docgen for
│  "/Users/thomas.davies/development-work/carbon/.storybook/utils/deprecation-warning.component.tsx"
│  because it is not included in the active TypeScript project.
│  Plugin: vite:react-docgen-typescript
│  File:
```

Could we do some sort of exclusion in here to avoid this?

src/components/tabs/tab/tab.stories.tsx
```diff
@@ -1,9 +1,9 @@
-import { Meta, StoryObj } from "@storybook/react";
+import { Meta, StoryObj } from "@storybook/react-vite";
 import { Tab } from "./tab.component";
 
 /**
  * This file is used primarily as a means to generate the props table.
- * It contains the tag: ["hideInSidebar"] so that it is not included in the sidebar.
+ * It contains the tag: ["!hideInSidebar"] so that it is not included in the sidebar.
```
tomdavies73: **question**: should this be `"!dev"` now?

src/components/tabs/__next__/__internal__/tab-list.stories.tsx
```diff
@@ -1,9 +1,9 @@
-import { Meta, StoryObj } from "@storybook/react";
+import { Meta, StoryObj } from "@storybook/react-vite";
 import { TabList } from "../tabs.component";
 
 /**
  * This file is used primarily as a means to generate the props table.
- * It contains the tag: ["hideInSidebar"] so that it is not included in the sidebar.
+ * It contains the tag: ["!hideInSidebar"] so that it is not included in the sidebar.
```
tomdavies73: **question**: should this be `"!dev"` now?

src/components/portrait/portrait.stories.tsx
```diff
@@ -199,9 +146,53 @@ export const CustomColors: Story = () => {
           onChange={(e) => setBgColour(e.target.value)}
           value={bgColour}
         >
-          {bgColors.map(({ label, value }) => (
-            <Option text={label} value={value} />
-          ))}
+          <Option text="paleblue" value="#A3CAF0" />
```
tomdavies73, line 159: **praise**: much more readable

.storybook/preview.ts
```diff
@@ -83,7 +83,7 @@ const globalTypes = {
       showName: true,
     },
   },
-  ...globalThemeProvider,
+  ...(globalThemeProvider as any),
```
tomdavies73: **question**: is there any other type we can use here?

.storybook/withThemeProvider.jsx
```diff
@@ -22,50 +21,38 @@ const FourColumnLayout = styled.div`
...
-        chromatic: {
-          theme: null,
-          fourColumnLayout: false,
-        },
-      },
-    },
-  ) => {
-    const { theme: chromaticTheme, fourColumnLayout } = parameters.chromatic;
-    const isChromaticBuild = isChromatic();
+const withThemeProvider = (Story, context) => {
```
tomdavies73: **question**: not necessarily related to the scope of your changes, but would we see any benefit in changing this file from `jsx` -> `tsx`?

tomdavies73 (commented):
Brilliant work @nineteen88,

The only other thing I've noticed is some leftover usage of `hideInSidebar`, this is still being used in some slightly obscure stories and some legacy stories, namely tabs.

A simple find and replace in VS Code should do the trick `hideInSidebar` -> `dev` 👍

playwright-ct.config.ts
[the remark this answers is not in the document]
nineteen88: The fileURLToPath(import.meta.url) approach also fails here. Playwright loads the config as CommonJS (since we have now removed the type in package.json as per the other feedback)

package.json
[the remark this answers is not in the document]
nineteen88: babel.config.mjs doesn't need to change as it already uses the .mjs extension so Node resolves it correctly. The root cause is that "type": "module" causes Node to interpret our CJS lib/*.js files as ESM, breaking require() for consumers

tsconfig.json
nineteen88: Storybook 10 uses package exports subpaths (e.g. storybook/internal/components) which "node" resolution can't resolve. "bundler" was required to make TypeScript happy with the upgrade

.storybook/withThemeProvider.jsx
nineteen88: Definitely scope creep but I did it to tidy it up and keep it consistent

.storybook/version-picker/index.tsx
```diff
@@ -76,14 +83,30 @@ const VersionPicker = () => {
...
+            <ActionList>
+              {items.map((item) => (
+                <ActionList.Item key={item.id}>
+                  <ActionList.Button
+                    ariaLabel={false}
+                    onClick={() => {
+                      window.location.href = item.href;
+                      item.onClick();
+                    }}
+                  >
```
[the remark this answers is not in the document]
nineteen88: ariaLabel is a Storybook-internal component prop where false explicitly signals "this button is labelled by its text content". It's not a DOM attribute, so standard ARIA rules don't apply here and will be required in storybook 11

tomdavies73 (commented):
Nothing further to add from me @nineteen88, there does seem to still be some leftover legacy usage of `hideInSidebar`. Apart from that it looks good though! 

Image

DipperTheDan (commented):
I noticed a few things. Mainly the two screenshots showing some files that shouldn't be there. 
Image

Image

<br>
I've also noticed that the interaction stories aren't working too. Screenshot of how they look.

Image

nineteen88:
> I noticed a few things. Mainly the two screenshots showing some files that shouldn't be there. Image
> 

I believe these are resolved with the next push of changes.

> Image
> I've also noticed that the interaction stories aren't working too. Screenshot of how they look. Image

Judging by this screenshot it looks like the Interaction tests are disabled on the toolbar at the top, and would explain why they're not running? They seem to work fine for me

tomdavies73 (dismissed):

DipperTheDan (dismissed):

paulrobinson88 (approved):

edleeks87 (approved):

tomdavies73 (approved):

PR #7373, opened 2025-06-10
fix(menu, multi-action-button, split-button): remove focus first list item on menu-popup open

### LINKED ISSUE #7264: Focus pulled to first menu item when activating menu with keyboard
### Description

When using keyboard to open the menu, focus is pulled to the first menu item. This fails WCAG 2.4.3 Focus Order. 

It's expected that focus stays on the menu itself once opened, until tab is pressed to go to the first menu item. 

### Reproduction

https://carbon.sage.com/?path=/docs/menu--docs

### Steps to reproduce

Use keyboard to open the menu

_No response_

### Suggested solution

_No response_

### Carbon version

152.2.0

### React version

v18

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #7264

### Proposed behaviour

`Menu`, `MultiActionButton` and `SplitButton`: 
- Maintain focus on toggle button/parent item on submenu open. 
- Do not open submenu with ArrowUp.
- Close submenu with Enter/Space on toggle button/parent item.

`MultiActionButton` and `SplitButton`:
- Closes submenu with Enter/Space or click on toggle button.
- Remove `aria-haspopup` and `add aria-controls` to toggle button.
- Focus border is not covered by hover style on menu buttons.

image

### Current behaviour

`Menu`, `MultiActionButton` and `SplitButton`: 
- Focus first menu item on submenu open. 
- Open submenu with ArrowUp.

`MultiActionButton` and `SplitButton`:
- Has `aria-haspopup` on toggle button.
- Focus border is covered by hover style on menu buttons.

image

### Checklist

#### QA

### Testing instructions

All components should:
- Open the submenu with Enter/Space/ArrowDown/Click.
- Close the submenu with Enter/Space/Click.
- Navigate into the menu with Tab/ArrowDown

Exception - in `Menu` when parent item has `href` and a submenu, menu should first open, Space should close and Enter should follow the link (expected default behaviour). See "Submenu Options" story in Menu. 

Screen readers should be able to navigate into the submenu list.

### DIFF
--- a/src/components/menu/__internal__/submenu/submenu.component.tsx
+++ b/src/components/menu/__internal__/submenu/submenu.component.tsx
@@ -116,7 +116,6 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
     const [applyFocusRadius, setApplyFocusRadius] = useState<boolean>(false);
     const [applyFocusRadiusToLastItem, setApplyFocusRadiusToLastItem] =
       useState<boolean>(false);
-    const focusFirstMenuItemOnOpen = useRef(false);
 
     const numberOfChildren = submenuItemIds.length;
 
@@ -282,28 +281,24 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
           | React.KeyboardEvent<HTMLAnchorElement>
           | React.KeyboardEvent<HTMLButtonElement>,
       ) => {
-        if (!submenuOpen) {
-          if (
-            Events.isEnterKey(event) ||
-            Events.isSpaceKey(event) ||
-            Events.isDownKey(event) ||
-            Events.isUpKey(event)
-          ) {
-            event.preventDefault();
-            openSubmenu();
-            focusFirstMenuItemOnOpen.current = !href;
-          }
+        const isToggleKey =
+          Events.isEnterKey(event) || Events.isSpaceKey(event);
+
+        if ((isToggleKey || Events.isDownKey(event)) && !submenuOpen) {
+          event.preventDefault();
+          openSubmenu();
         }
 
         if (submenuOpen) {
           const index = findCurrentIndex(submenuFocusId);
           let nextIndex = index;
 
-          if (href && !submenuFocusId) {
-            if (Events.isDownKey(event) || Events.isUpKey(event)) {
+          if (!submenuFocusId) {
+            // toggle close when Space/Enter is pressed on the parent item
+            // when parent is a link, allow default behaviour on Enter once submenu is open
+            if ((isToggleKey && !href) || (Events.isSpaceKey(event) && href)) {
               event.preventDefault();
-              setSubmenuFocusId(submenuItemIds[0]);
-              return;
+              closeSubmenu();
             }
           }
 
@@ -407,18 +402,6 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
         }
       }
     }, [children, submenuOpen, submenuRef]);
-
-    useEffect(() => {
-      if (
-        focusFirstMenuItemOnOpen.current &&
-        submenuOpen &&
-        !submenuFocusId &&
-        submenuItemIds.length
-      ) {
-        focusFirstMenuItemOnOpen.current = false;
-        setSubmenuFocusId(submenuItemIds[0]);
-      }
-    }, [submenuOpen, submenuFocusId, submenuItemIds]);
 
     const handleClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
       openSubmenu();
--- a/src/components/menu/menu-item/menu-item.test.tsx
+++ b/src/components/menu/menu-item/menu-item.test.tsx
@@ -720,7 +720,7 @@ describe("when MenuItem has a submenu", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("focuses succeeding submenu item when 'ArrowDown' key is pressed", async () => {
+  it("opens the submenu when the 'ArrowDown' key is pressed, then moves focus to the next submenu items on subsequent presses", async () => {
     const user = userEvent.setup();
     render(
       <Menu>
@@ -732,12 +732,14 @@ describe("when MenuItem has a submenu", () => {
       </Menu>,
     );
     const submenuParentItem = screen.getByRole("button", { name: "Item One" });
-    act(() => {
-      submenuParentItem.focus();
-    });
+    await user.tab();
     await user.keyboard("{arrowdown}");
     const submenuItems = screen.getAllByRole("link");
 
+    expect(submenuParentItem).toHaveFocus();
+    expect(submenuItems[0]).toBeVisible();
+
+    await user.keyboard("{arrowdown}");
     expect(submenuItems[0]).toHaveFocus();
     await user.keyboard("{arrowdown}");
     expect(submenuItems[1]).toHaveFocus();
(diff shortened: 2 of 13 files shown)


src/components/split-button/split-button.component.tsx
```diff
@@ -147,6 +148,8 @@ export const SplitButton = forwardRef<SplitButtonHandle, SplitButtonProps>(
     };
 
     const handleToggleClick = () => {
+      // ensure button is focused when clicked (Safari)
+      toggleButtonRef.current?.focus();
```
edleeks87:
```suggestion
      toggleButtonRef.current?.focus({ preventScroll: true });
```

src/components/menu/menu-item/menu-item.test.tsx
```diff
@@ -582,12 +582,14 @@ describe("when MenuItem has a submenu", () => {
...
     const submenuParentItem = screen.getByRole("button", { name: "Item One" });
-    act(() => {
-      submenuParentItem.focus();
-    });
+    await user.tab();
     await user.keyboard("{arrowdown}");
     const submenuItems = screen.getAllByRole("link");
 
+    expect(submenuParentItem).toHaveFocus();
+    expect(submenuItems[0]).toBeVisible();
```
Parsium, line 740: **question**: do we want the focus to remain on the parent menu item after the down arrow key is pressed?

Requiring the user to press the key twice seems a bit counter-intuitive, so I just want to make sure I'm not misunderstanding.

I noticed [this example in the WAI-ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/) (which shares similarities to Carbon's `Menu`) handles the behaviour slightly differently, but it does seem more intuitive to me.

src/components/menu/menu-item/menu-item.test.tsx
```diff
@@ -609,13 +611,13 @@ describe("when MenuItem has a submenu", () => {
...
-    act(() => {
-      submenuParentItem.focus();
-    });
+    await user.tab();
     await user.keyboard("{arrowup}");
     const submenuItems = screen.getAllByRole("link");
 
-    expect(submenuItems[0]).toHaveFocus();
+    expect(submenuParentItem).toHaveFocus();
+    expect(submenuItems[0]).toBeVisible();
```
Parsium, line 769: **comment**: similar case here to the question regarding the down arrow key behaviour.

src/components/split-button/split-button.test.tsx
```diff
@@ -498,13 +481,15 @@ test("should render the child buttons when a 'ArrowDown' keydown event detected
     </SplitButton>,
   );
 
-  const toggle = await screen.findByRole("button", { name: "Show more" });
+  const toggle = screen.getByRole("button", { name: "Show more" });
```
Parsium, line 484: **question**: do we want to continue supporting arrow key navigation, given users won't be able to access the popup from the toggle button via these keys?

While users might expect the down arrow key to open the popup, they could get confused when they can't then access the popup buttons with the same key as well.

As `SplitButton` isn't defined as an ARIA menu, I wonder if it might be less confusing if we removed the arrow key behaviour altogether, so users can only navigate via tabbing.

src/components/multi-action-button/multi-action-button.test.tsx
Parsium, line 1: **comment**: similar to `SplitButton`, I wonder if supporting any arrow key navigation could be confusing if we are no longer moving focus when the popup becomes visible, as the popup would no longer be accessible via the arrow keys when the toggle button is focused.

nuria1110: To respond to all of your comments, as we still support arrow keys across several other components (Menu, ActionPopover, etc.) and they should all ideally work the same in terms of keyboard navigation, my thinking was to leave the support of arrow keys for now until we have come to a decision and can consolidate this behaviour.

For now, I will fix the issue where you cannot navigate MultiActionButton/SplitButton from the toggle to the popup using arrow down as I agree that could be confusing and to match its behaviour to Menu.

src/components/menu/menu-item/menu-item.test.tsx
nuria1110: Responded below

src/components/menu/menu-item/menu-item.test.tsx
nuria1110: Responded below

src/components/split-button/split-button.test.tsx
nuria1110: Responded below

src/hooks/__internal__/useChildButtons/useChildButtons.tsx
```diff
@@ -51,10 +44,14 @@ const useChildButtons = (
...
       Events.isUpKey(ev);
 
-    if (isToggleKey) {
+    if (isToggleKey && !showAdditionalButtons) {
       ev.preventDefault();
       showButtons();
     }
+
+    if (Events.isDownKey(ev) && showAdditionalButtons) {
+      getButtonChildren()?.[0]?.focus();
```
Parsium: **suggestion**: I noticed the page scrolls when pressing the down arrow key to enter the popup buttons.

Since we are defining custom behaviour on the key press, what do you think about preventing the default browser action?: 

```suggestion
      ev.preventDefault();
      getButtonChildren()?.[0]?.focus();
```

nuria1110: I addressed the scrolling issue elsewhere but forgot to do it here 😅, will get it sorted

edleeks87: Good spot @Parsium, you might need to do the same for up arrow key as well fyi @nuria1110, worth clarifying with UX at least. Confusing part for me is that we allow the popover to open on up key still etc

edleeks87 (approved):

src/hooks/__internal__/useChildButtons/useChildButtons.tsx
nuria1110: Spoke with accessibility and UX and agreed to remove open with arrow up

Parsium (approved):

harpalsingh:
Tested updated keyboard behaviour and focus remains on parent now. Passing a11y review.

src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -282,28 +281,24 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
...
         if (submenuOpen) {
           const index = findCurrentIndex(submenuFocusId);
           let nextIndex = index;
 
-          if (href && !submenuFocusId) {
-            if (Events.isDownKey(event) || Events.isUpKey(event)) {
+          if (!submenuFocusId) {
+            // toggle close when Space/Enter is pressed on the parent item
+            // when parent is a link, allow default behaviour on Enter once submenu is open
+            if ((isToggleKey && !href) || (Events.isSpaceKey(event) && href)) {
```
Parsium, line 299: **note**: hopefully the latter condition shouldn't be reached in practice, since if a parent menu item is a link, this should trigger a page redirect, rather than an action on the same page. So the parent item should be a button to ensure correct HTML semantics.

Parsium (approved):

src/components/menu/__internal__/submenu/submenu.component.tsx
nuria1110: Yes, in theory this is not recommended but unfortunately we do support this so I wanted to make sure it still behaved as "expected".

paulrobinson88 (approved):

PR #7387, opened 2025-06-18
fix(adaptive-sidebar): ensure all elements are indexed behind the component when rendered as a modal

tomdavies73:
### Proposed behaviour

Adds a global style variable which ensures all popover based components and the select input have a lower z-index than the adaptive sidebar

### Current behaviour

Currently, when the adaptive sidebar hits the set breakpoint, any popover based components and the select input have a higher z-index, which causes them to still be visible, even when the adaptive sidebar covers them

Screenshot 2025-06-18 at 17 08 00

### Checklist

#### QA

### Testing instructions

In the `WithDropdown` test story, a chromatic test has been added to catch the fixed regression.

Also, a select has also been added inside the adaptive sidebar to verify that the z-index of the fixed components is not adversely affected when rendered inside the adaptive sidebar

### DIFF
--- a/src/__internal__/popover/popover.test.tsx
+++ b/src/__internal__/popover/popover.test.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import Popover, { PopoverProps } from "./popover.component";
-import { baseTheme } from "../../style/themes";
 import Dialog from "../../components/dialog";
 
 const PopoverWithButton = (
@@ -144,15 +143,22 @@ test("renders popup within a transparent backdrop to prevent scrolling outside t
   );
 
   const backdrop = screen.getByTestId("popup-backdrop");
+
   expect(backdrop).toHaveStyle({
     background: "transparent",
-    zIndex: baseTheme.zIndex.popover,
     position: "fixed",
     top: 0,
     right: 0,
     bottom: 0,
     left: 0,
   });
+
+  const backdropIndex = getComputedStyle(backdrop).getPropertyValue("z-index");
+
+  // non-default value
+  expect(backdropIndex).toContain("--adaptiveSidebarModalBackdrop");
+  // default value
+  expect(backdropIndex).toContain("6000");
 });
 
 test("does not render a backdrop when disableBackgroundUI is false", () => {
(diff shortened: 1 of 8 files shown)


src/__internal__/popover/popover.test.tsx
```diff
@@ -144,15 +143,18 @@ test("renders popup within a transparent backdrop to prevent scrolling outside t
...
-    zIndex: baseTheme.zIndex.popover,
     position: "fixed",
     top: 0,
     right: 0,
     bottom: 0,
     left: 0,
   });
+
+  const backdropIndex = getComputedStyle(backdrop).getPropertyValue("z-index");
+  expect(backdropIndex).toContain("6000");
```
damienrobson-sage: **suggestion (non-blocking)**: is it worth adding a test for the non-default value here too?

src/components/flat-table/flat-table.test.tsx
```diff
@@ -1553,14 +1552,16 @@ test("when an ActionPopover is opened inside the FlatTable, it will have the bac
...
     position: "fixed",
     top: 0,
     right: 0,
     bottom: 0,
     left: 0,
   });
 
+  const backdropIndex = getComputedStyle(backdrop).getPropertyValue("z-index");
+  expect(backdropIndex).toContain("6000");
+
```
damienrobson-sage: **suggestion**: same here

src/components/select/select.style.ts
```diff
@@ -44,7 +44,8 @@ const StyledSelect = styled.div.attrs(applyBaseTheme)<StyledSelectProps>`
 
       ${isOpen &&
       css`
-        z-index: ${theme.zIndex.aboveAll};
+        /* If behind an adaptive sidebar, use the backdrop z-index value, if not default to original z-index value */
+        z-index: var(--adaptiveSidebarBackdrop, ${theme.zIndex.aboveAll});
```
damienrobson-sage: **question**: given there are test scenarios for the other components, do we not need additional ones in `Select` to cover this change?

src/style/global-style.ts
```diff
@@ -29,6 +29,11 @@ const GlobalStyle = createGlobalStyle`
   [data-element=${TAB_GUARD_TOP}], [data-element=${TAB_GUARD_BOTTOM}] {
     position: fixed;
   }
+
+  body:has(.adaptive-sidebar) {
```
edleeks87: might be worth naming the class a bit more intuitively as we only set it when it's in modal view, similar comment for the token name but less important than class imo.

I also wonder if we can get access to the `theme` here and set it to the preset value from there?


```suggestion
  body:has(.adaptive-sidebar-modal-view) {
    --adaptiveSidebarModalBackdrop: ${({ theme = baseTheme }) => theme.zIndex.overlay};
```

PR #7389, opened 2025-06-19
fix(responsive-vertial-menu): menu closes when shrinking the size

mihai-albu-sage:
### Proposed behaviour

If the menu loses focus during a resize event, the `handleBlur` handler is prevented from closing the menu. This should only be prevented during a resize because the menu should close when losing focus during keyboard navigation.

To achieve this, a new resize event listener has been added. We track whether a resize is in progress by setting a flag (`isResizingRef`) to true in the resize event handler, and resetting it to false using a debounced `setTimeout` after `100ms` of no resize activity. This allows us to detect when the user is actively resizing the window.

In the `handleBlur` logic, we check this flag. If a blur occurs while `isResizingRef` is `true`, we skip closing the menu. This prevents unwanted closures caused by temporary focus shifts during layout reflows or DOM changes triggered by resizing.

This logic ensures the menu remains open during intentional layout changes (like a resize), but still closes appropriately during user-driven navigation events like pressing Tab or clicking outside the menu.

https://github.com/user-attachments/assets/7c4f5902-4407-4258-8cd2-25112dd07b1a

https://github.com/user-attachments/assets/b948fe4b-2094-4933-a2bd-ef54434054cc

### Current behaviour

During a window resize, conditional rendering causes  the menu to temporarily unmount and remount. As a result, the `document.activeElement` resets to the `<body>`. This triggers the `handleBlur` event handler, which interprets the loss of focus as the user navigating away from the menu and closes it.

This behavior is unintentional: the menu closes not because the user actively moved focus, but because the layout shift during the resize disrupted focus handling.

https://github.com/user-attachments/assets/4e0ca713-de30-4ca6-bb71-640247a4cfa1

https://github.com/user-attachments/assets/582e3efc-d9d7-445a-b192-8f11274e28c0

### Checklist

#### QA

### Testing instructions

Steps:
1. Go to the Storybook page for the component.
2. Open the menu.
3. Resize the browser window to a width greater than 700px — this should render the menu in its default (non-modal) mode.
4. Select an option from the menu.
5. Now, resize the window down to less than 700px — this should switch the menu to responsive mode (rendered inside a modal).

Expected Result:
The menu should remain open and visible after the window is resized into responsive mode.

### DIFF
--- a/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
+++ b/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
@@ -130,6 +130,32 @@ const BaseMenu = ({
     };
   }, [active, measureDimensions, menu, responsiveMode]);
 
+  const isResizingRef = useRef(false);
+  const resizeTimeoutRef = useRef<number | null>(null);
+
+  useEffect(() => {
+    const handleResize = () => {
+      isResizingRef.current = true;
+
+      if (resizeTimeoutRef.current !== null) {
+        clearTimeout(resizeTimeoutRef.current);
+      }
+
+      resizeTimeoutRef.current = window.setTimeout(() => {
+        isResizingRef.current = false;
+      }, 100);
+    };
+
+    window.addEventListener("resize", handleResize);
+
+    return () => {
+      window.removeEventListener("resize", handleResize);
+      if (resizeTimeoutRef.current !== null) {
+        clearTimeout(resizeTimeoutRef.current);
+      }
+    };
+  }, []);
+
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


damienrobson-sage (approved):

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
```diff
@@ -130,6 +130,42 @@ const BaseMenu = ({
...
+  /**
+ * 
+ *         if (
+          containerRef.current &&
+          !containerRef.current.contains(document.activeElement) &&
+          !isResizingRef.current
+        ) {
+          setActive(false);
+ * 
+ */
```
damienrobson-sage: **nitpick**: remove the commented code if it isn't needed 😄

PR #7391, opened 2025-06-19
fix(responsive-vertical-menu): fix auto-close issue in Chrome

damienrobson-sage:
### Proposed behaviour

Ensure that the blur function handles the situation correctly when the `focusout` handler is called in Chrome (Firefox and Safari behaviour is unaffected). This is down to how restrictive Chrome is when handling DOM updates.

### Current behaviour

When programmatically closing the responsive menu as a result of child menu updates (e.g. via API calls), Chrome closes the entire menu as opposed to just updating in place

### Checklist

- [x] Commits follow our style guide
- [ ] Related issues linked in commit messages if required
- [ ] Screenshots are included in the PR if useful
- [ ] All themes are supported if required
- [x] Unit tests added or updated if required
- [x] Playwright automation tests added or updated if required
- [ ] Storybook added or updated if required
- [ ] Translations added or updated (including creating or amending translation keys table in storybook) if required
- [ ] Typescript `d.ts` file added or updated if required
- [ ] Related docs have been updated if required

#### QA

- [ ] Tested in provided StackBlitz sandbox/Storybook
- [ ] Add new Playwright test coverage if required
- [ ] Carbon implementation matches Design System/designs
- [ ] UI Tests GitHub check reviewed if required

### Testing instructions

Across all browsers, and using the `Responsive Vertical Menu > Test > Delayed Menu Items` story, open all menus and await the re-load. Ensure that the secondary-level menu closes when this happens but the main menu stays open.

PR #7392, opened 2025-06-19
fix(responsive-vertical-menu): ensure menu can be closed via the open button  in safari

tomdavies73:
adds some additional code to ensure safari specific focusout false triggers are mitigated which previously prevented the menu from being closed via the open button when the menu is already open in safari

### Proposed behaviour

Adds an early return to the `handleBlur` function which is called during the `"focusOut"` event, which is wrongly triggered in Safari due to our programmatic focus on `Button`. 

This wrongly triggers a close event which caused a race condition with other updates occurring within the component and caused the menu to close and then open again in a brief flash.

### Current behaviour

Currently, a wrongly triggered close event, which causes a race condition with other updates occurring within the component and causes the menu to close and then open again in a brief flash.

### Checklist

#### QA

### Testing instructions

As this issue is Safari only, there is no specific tests we can add for this. I recommend manual testing and re-create the issue steps via the ticket on Safari to verify the issue is resolved. I would also test general navigation etc. and that behaviour in other browsers work as expected.


damienrobson-sage (approved):
Looks good, and doesn't seem to interfere with the Chrome fix in #7391

mihai-albu-sage (dismissed):

paulrobinson88 (approved):

Issue #7380, opened 2025-06-16
Text Editor character limit/character count, new line issue


allan-sage:
### Description

I am using more than one new line to space out the paragraphs in my text editor
When createFromHTML or copy&paste in paragraphs spaced out by multiple new lines
Then I start typing into the text editor, the character limit goes up

Example to copy and paste (may have to type out first then copy and paste as github is not putting in newlines):
a paragraph
\n
a paragraph
\n
a paragraph
\n
a paragraph
\n
\n
a paragraph
\n
\n
a paragraph 

### Reproduction

https://carbon.sage.com/?path=/docs/text-editor--docs

### Steps to reproduce

![Image](https://github.com/user-attachments/assets/1adaca11-4dca-46e2-81bc-c9e71ed5ef41)
![Image](https://github.com/user-attachments/assets/c03d135f-c291-45ce-9253-55c45ef3ab14)

### Carbon version

154.4.1

### React version

v18

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Confidentiality

- [x] I confirm there is no confidential or commercially sensitive information included.

PR #7398, opened 2025-06-25
fix(text-editor): character count behaving inconsistently when pasting

mihai-albu-sage:
fixes: #7380

### Proposed behaviour
To address the issue where Lexical does not consistently preserve empty paragraphs during serialization—especially after content mutations like typing a character—we now track empty lines by explicitly counting paragraph nodes instead of relying on serialized text alone.

https://github.com/user-attachments/assets/c59f443f-5feb-4aaf-b960-729670bca052

### Current behaviour

Currnetly, when users pasted content that included empty lines (e.g., multiple line breaks), Lexical would initially preserve those empty paragraphs in the editor's internal state. However, as soon as a new character was typed (which triggers a content mutation and a tree diff), Lexical would treat those empty paragraphs as insignificant and remove them from the serialized output. As a result, the empty lines would disappear in the serialized data, leading to an inaccurate character count — even though the lines were still visually present in the editor before typing.

https://github.com/user-attachments/assets/92c1ada1-7b7b-4f1f-b5d4-a7b222c15120

### Checklist

#### QA

### Testing instructions

1. Go to the Text Editor default storybook page
2. Write some text combined with multiple line breaks and keep track of the remaining characters.
3. Select the text and copy it
4. Delete the text and paste it in the text editor
5. Type an additional character

Expected: the remaining characters count should go down after typing
Actual: the remaining characters count goes up.

### DIFF
--- a/src/components/text-editor/__internal__/plugins/CharacterCounter/character-counter.component.tsx
+++ b/src/components/text-editor/__internal__/plugins/CharacterCounter/character-counter.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { $getRoot } from "lexical";
 import {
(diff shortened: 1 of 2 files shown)


src/components/text-editor/text-editor.pw.tsx
```diff
@@ -587,6 +587,52 @@ test.describe("Prop tests", () => {
     });
   });
 
+  test("counts characters correclty after pasting new line characters and interacting with it", async ({
```
Parsium, line 585: **suggestion(non-blocking)**: what do you think about using Playwright's recommended [`getBy*` locator functions](https://playwright.dev/docs/locators) instead of locating by path?

This could ensure the tests are not tied to a specific DOM structure in case we change it in the future:

```tsx
// instead of page.locator("div[role='textbox']"):
const textbox = page.getByRole("textbox"); 

// instead of page.locator("div[data-role='pw-rte-character-limit']")
// Carbon has `page.getByTestId()` configured to locate elements with the matching 'data-role' attribute: 
const displayedLimitAfterPastingText = page.getByTestId("pw-rte-character-limit").textContent();
```

src/components/text-editor/__internal__/plugins/CharacterCounter/character-counter.component.tsx
```diff
@@ -20,21 +20,40 @@ const CharacterCounterPlugin = ({
...
   useEffect(() => {
-    // The character counter plugin listens for updates to the editor state
-    // independently to ensure updates do not conflict/interrupt other state
-    // changes
-    editor.registerUpdateListener(({ editorState }) => {
-      editorState.read(() => {
-        const newContent = $getRoot().getTextContent();
+    const updateCharCount = () => {
+      editor.getEditorState().read(() => {
+        const root = $getRoot();
```
Parsium: **question**: I noticed in `TextEditor`'s change handler, we get the text content size in a similar way to what we had previously for the `CharacterCounter` plugin ([link to source](https://github.com/Sage/carbon/blob/9bd9361/src/components/text-editor/text-editor.component.tsx#L211-L213)).

Assuming the same reproduction steps you've described in the PR description are followed, could this potentially lead to a scenario where the character limit warning is displayed, even if the rendered character count is one or more characters above the limit?

src/components/text-editor/text-editor.pw.tsx
mihai-albu-sage: Updated the test to use the recommended locator functions.

src/components/text-editor/__internal__/plugins/CharacterCounter/character-counter.component.tsx
mihai-albu-sage: Although the change handler reads the content from the Text Editor similarly to the previous character counter implementation, the character limit warning displays correctly. I was unable to reproduce any issues with the warning message, even when following the mentioned reproduction steps.

Parsium: Ah brill, thanks for confirming 👍🏼

edleeks87 (approved):

Parsium (approved):

edleeks87 (approved):

paulrobinson88 (approved):

PR #7399, opened 2025-06-25
fix(text-editor): ensure onChange is called correctly

### LINKED ISSUE #7384: TextEditor: Unexpected behabiour when used as a controlled component
### Description

There are a few behaviors observed in the TextEditor component that may affect its usability when used as a controlled input. 
These are shared below for consideration:

#### 1. `onChange` fires on every click
The `onChange` event is triggered even when clicking inside the editor without modifying its content. This makes it hard to distinguish between actual changes and user interaction.

#### 2. `onChange` fires when `value` prop changes
When the parent updates the `value` prop (e.g., controlled input pattern), it causes `onChange` to trigger, even though the change came externally. This behavior creates an infinite loop if not handled carefully.

#### 3. Focus is lost on value update
If a user is typing into the editor and the `value` prop updates (e.g., due to a debounce or state sync), the editor loses focus and the cursor disappears. This makes it nearly impossible to type continuously while controlling the editor's content.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-mttmh3qh

### Steps to reproduce

1. Click inside the editor without typing → `onChange` fires.
2. Type something → the `value` updates → focus is lost.
3. Programmatically update the `value` prop → `onChange` fires again.

_No response_

### Suggested solution

_No response_

### Carbon version

154.8.0

### React version

v17

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

Windows

### Additional context

- I initially tried debouncing the `onChange` to mitigate the re-rendering and focus loss, but the issue persists — it only delays the loss of focus.
- Here's a short video (~2 min) walking through each issue: 
https://github.com/user-attachments/assets/bb9b9acf-9ade-44f1-8b80-24fd4f1f6a97

### Confidentiality

tomdavies73:
fix #7384

### Proposed behaviour

Ensures onChange is only called when the editor is typed into, the bold/italic buttons are used to format text, and the unordered/ordered list buttons are used. Also fixes an issue where focus is lost when the value is changed when the component is used in a controlled manner.

### Current behaviour

Currently, onChange is called on render, on click and when the value prop is changed programmatically. Focus is also lost when the value is changed when the component is used in a controlled manner

### Checklist

#### QA

### Testing instructions

Some test cases have been added to address this behaviour, but some manual testing will be required


src/components/text-editor/text-editor.component.tsx
```diff
@@ -212,7 +215,25 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
...
         );
 
-        if (onChange) {
+        const currentStateJSON = JSON.stringify(newState.toJSON());
+        const lastStateJSON = lastStateRef.current;
+
+        if (currentStateJSON === lastStateJSON) {
+          return; // No actual change, prevents unnecessary onChange calls
+        }
+
```
edleeks87, line 224: suggestion: if the above suggestion about usePrevious is a non-starter you could update the programmatic ref here rather than inside the if and after it etc

```
   lastStateRef.current = currentStateJSON;

   if (programmaticUpdateRef.current) {
         programmaticUpdateRef.current = false;
         return;
   }

File: src/components/text-editor/text-editor.component.tsx [hunk 2]
```diff
@@ -212,7 +215,25 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
           $getRoot().getTextContent(),
         );
 
-        if (onChange) {
+        const currentStateJSON = JSON.stringify(newState.toJSON());
+        const lastStateJSON = lastStateRef.current;
+
+        if (currentStateJSON === lastStateJSON) {
+          return; // No actual change, prevents unnecessary onChange calls
```
edleeks87, line 222: suggestion (non-blocking): this comment feels a bit superfluous to me

src/components/text-editor/text-editor.component.tsx
```diff
@@ -248,12 +287,11 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
 
     // Reset the value of the editor when the cancel trigger is updated (implements reset on cancel)
     useEffect(() => {
-      const safeValue = value || createEmpty();
-
-      /* istanbul ignore else */
-      if (editorRef.current) {
+      const safeValue = value || createEmpty(); /* istanbul ignore else */
```
edleeks87, line 290:
```suggestion
      const safeValue = value || createEmpty();
      /* istanbul ignore else */
```

src/components/text-editor/text-editor.component.tsx
```diff
@@ -212,7 +215,25 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
...
+          // but we still want to update the lastStateRef, and reset the flag
+          lastStateRef.current = currentStateJSON;
+          programmaticUpdateRef.current = false;
+          return;
+        }
+        lastStateRef.current = currentStateJSON;
+
+        // only call onChange once the lastStateJson has been initialised
+        // this is to prevent the onChange callback being called on the first render
+        if (lastStateJSON !== null && onChange) {
```
edleeks87, line 236: comment: I realise you are only updating this but we're already using nullish coalescence in the if statement so I think we can tweak this to be the following if the above suggestion about usePrevious is a non-starter

```
       if (lastStateJSON !== null) {
          const formattedValues = editorRef.current
            ? SerializeLexical(editorRef.current)
            : {};
          onChange?.(currentTextContent, formattedValues);
        }

File: src/components/text-editor/text-editor.component.tsx [hunk 5]
```diff
@@ -233,6 +254,24 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
       [characterLimit, locale.textEditor, onChange],
     );
 
+    useEffect(() => {
+      if (value && editorRef.current) {
+        const currentEditorState = editorRef.current.getEditorState();
+        const currentStateJSON = JSON.stringify(currentEditorState.toJSON());
+
+        // if the current state does not match the value prop, we need to update the editor state
```
edleeks87, line 262: suggestion (non-blocking): I think some of these comments are overkill, imo a single comment above the if statement is enough

```suggestion
        // handles for controlled updates to the value whilst preventing calling onChange unnecessarily 
```

src/components/text-editor/text-editor.component.tsx
```diff
@@ -204,6 +204,9 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
       };
     }, [namespace, readOnly, value]);
 
+    const lastStateRef = useRef<string | null>(null);
```
edleeks87, line 207: suggestion: I wonder if we can utilise [usePrevious](https://github.com/Sage/carbon/blob/master/src/hooks/__internal__/usePrevious/index.ts) here, if it works you won't need to re-assign the current each time it updates etc. It might not work with the different serialisation options and the checks below (checking if the last value was null etc, maybe a `isFirstRender` flag works instead there) but might be worth a try for the sake of consistency at the least. If it doesn't I'm happy with the current approach but would remove `Ref` from the name as imo `lastState` is intuitive enough etc

```
const previousValue = usePrevious(value || createEmpty()); // might not work as we don't know the format of value etc
const isFirstRender = useRef(true);

...
const handleChange = useCallback(
... 
const previousValueJSON = JSON.stringify(previousValue);

if (currentStateJSON === previousValueJSON || isFirstRender.current) {
  isFirstRender.current = false;
  return;
 }

File: src/components/text-editor/text-editor.component.tsx [hunk 7]
```diff
@@ -204,6 +204,9 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
...
+    const programmaticUpdateRef = useRef<boolean>(false);
```
edleeks87, line 208: suggestion (non-blocking):
```suggestion
    const hasControlledValueUpdate = useRef<boolean>(false);
```

src/components/text-editor/text-editor.test.tsx
```diff
@@ -1,24 +1,13 @@
-import { render, screen, within } from "@testing-library/react";
+import { render, screen, within, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
-import React, { act, createRef } from "react";
+import React, { act, createRef, useState } from "react";
```
edleeks87, line 4: comment: I think we should be importing `act` from `"@testing-library/react"`

src/components/text-editor/utils.ts
```diff
@@ -1,16 +1,45 @@
 import { DeserializeHTML } from "./__internal__/helpers";
 
+interface LexicalNode {
+  type: string;
+  direction?: string | null;
+  children?: LexicalNode[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any; // For other properties we don't need to specify
```
edleeks87, line 8:
```suggestion
  [key: string]: unknown;
```

if we use unknown we can remove the linter comment

Parsium: Agreed, I think we should be defaulting to `unknown` over the `any` type for most cases.

[Values with an `any` type are not type-checked by TypeScript](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any), so it's possible we could overlook typing issues before they become larger bugs.

PR #7413, opened 2025-07-03
chore: use typescript parser and upgrade to eslint v9

tomdavies73:
### Proposed behaviour

This PR upgrades ESLint to v9, switches to the TypeScript parser, and refactors several components/tests for consistency and cleaner types.

Our default overrides are still used and remain untouched, but due to air-bnb's lint rules not being compatible with ESLint v9 they have been replaced with the default recommended rules where applicable. 

This change in rules has highlighted a few instances where missed rules caused some linting issues to be missed in the codebase, some eslint ignores also remain for rules which are no longer applied, these have been left so they can be reviewed at a later date. 

### Current behaviour

Eslint is currently below v9, and uses the majority of air-bnb's lint rules and babel parser is used instead of ts parser.

### Checklist

#### QA


eslint.config.mjs
```diff
@@ -0,0 +1,425 @@
...
+import js from "@eslint/js";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const cleanGlobals = (globalObj) => {
+  const cleaned = {};
+  for (const [key, value] of Object.entries(globalObj)) {
+    const cleanKey = key.trim();
+    if (cleanKey) {
```
damienrobson-sage: **question**: what's this check for? If `trim` is attempted on a null/undefined value it'll throw an error so you'll never hit the `if` clause unless you've got something valid anyway;

tomdavies73: Good point mate, I'll see if there's a better way to do this or if we can just trim the global directly when needed.

edleeks87: I don't think it can ever be undefined as the Object.entries guarantees they will be strings so the following is enough:

```
const cleanGlobals = (globalObj) => Object.entries(globalObj).reduce((cleaned, [key, value]) => {
    cleaned[key.trim()] = value;
    return cleaned;
  }, {});

File: eslint.config.mjs [hunk 2]
```diff
@@ -0,0 +1,425 @@
...
+import typescriptEslint from "@typescript-eslint/eslint-plugin";
+import jest from "eslint-plugin-jest";
+import testingLibrary from "eslint-plugin-testing-library";
+import jestDom from "eslint-plugin-jest-dom";
+import playwright from "eslint-plugin-playwright";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import js from "@eslint/js";
+
+const __filename = fileURLToPath(import.meta.url);
```
edleeks87: We might have the same cross-platform issue we have with the rollup config. @paulrobinson88 or @stephenogorman would one of you mind testing this on windows please? I don't mind leaving it as is for now as I can address both at the same time just be good to know etc

tomdavies73:
> ## Pull Request Overview
> 
> This PR upgrades ESLint to v9, switches to the TypeScript parser, and refactors several components/tests for consistency and cleaner types.
> 
>     * Replace custom prop interfaces with direct `MarginProps` where applicable and update re-exports
> 
>     * Simplify catch blocks and no-op callbacks in Playwright tests
> 
>     * Revamp lint configuration to `eslint.config.mjs` and bump relevant dependencies
> 
> 
> ### Reviewed Changes
> 
> Copilot reviewed 15 out of 16 changed files in this pull request and generated no comments.
> Show a summary per file
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> 	
> Comments suppressed due to low confidence (3)

Yeah looks like these should all be dev deps, missed that during the initial run through

package.json
```diff
@@ -189,6 +187,7 @@
     "vite-plugin-static-copy": "^2.2.0"
   },
   "dependencies": {
```
edleeks87: Comment: as these are only used in the eslint config they can be devDependencies

eslint.config.mjs
```diff
@@ -0,0 +1,425 @@
...
+    const cleanKey = key.trim();
+    if (cleanKey) {
+      cleaned[cleanKey] = value;
+    }
+  }
+  return cleaned;
+};
+
+export default defineConfig([
+  globalIgnores(["**/__spec__.js", "**/__definition__.js"]),
```
edleeks87: comment: this can be removed we no longer have files matching that naming convention

edleeks87 (dismissed):

damienrobson-sage (approved):

edleeks87 (dismissed):

PR #7424, opened 2025-07-08
FE-7401 Date picker tab stops

damienrobson-sage:
### Proposed behaviour

Reduce the number of tab stops to one by preventing the date picker popover from opening unless the calendar icon is clicked with the mouse; keyboard users benefit from being able to type the date without needing to navigate back to the text box or to their desired date.

### Current behaviour

There are currently two tab stops when using a keyboard to navigate through date pickers, which can delay and interupt user flow.

### Checklist

#### QA

### Testing instructions

Open any date-driven story and ensure that: 
- focusing the text input via click or tab doesn't display the calendar;
- clicking the calendar icon displays the calendar.


src/components/date/date.component.tsx
Parsium, line 1: **question**: I noticed that the requirements mention pressing the down arrow key in the input box should open the picker popup.

Is there a particular reason why we might want to omit this feature?

src/components/date/date.pw.tsx
```diff
@@ -309,8 +324,8 @@ test.describe("Functionality tests", () => {
   }) => {
     await mount(<WithSiblingButton value="01/05/2022" />);
 
-    const dateInput = page.getByRole("textbox");
-    await dateInput.click();
+    const calendarIcon = getDataElementByValue(page, "calendar");
```
Parsium: **suggestion(non-blocking)**: if you'd like to use Playwright's in-built locators, we've configured Playwright so its `page.getByTestId` locator is set up to find elements with a matching `data-role` attribute:

```tsx
const calendarIcon = page.getByTestId("icon");
```

src/components/date/date.component.tsx
```diff
@@ -315,7 +315,6 @@ export const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
       isBlurBlocked.current = false;
 
       if (!open && !alreadyFocused.current) {
```
edleeks87: question: do we need this if/else at all anymore? The picker shouldn't open on focus unless I'm misunderstanding

src/components/date/date.component.tsx
edleeks87: FYI @Parsium, @damienrobson-sage and myself chatted with Hannah and it was decided to leave that requirement for now as we would need to convert the popover/calendar into a modal to make it WCAG compliant

src/components/date/date.component.tsx
damienrobson-sage: Good spot, will remove it now.

src/components/date/date.pw.tsx
damienrobson-sage: Updated - had to leave a couple lying around due to tests having multiple icons (popover close buttons, etc.)

edleeks87 (approved):

dariusbercea-sage (approved):

PR #7425, opened 2025-07-09
Update rollup

edleeks87:
### Proposed behaviour

Removes use of `fileURLToPath` as it is platform specific. Using `process.cwd` allows the scripts to
run on all platforms. Refactors external array to be a function to prevent entry files from being
treated as external.

Updates `rollup` and related plugins as follows:

- rollup 4.44.2
- @rollup/plugin-node-resolve 15.2.3
- @rollup/plugin-commonjs 25.0.7
- @rollup/plugin-json 6.1.0
- @rollup/plugin-terser 0.4.4

### Current behaviour

Rollup config relies on `fileURLtoPath` which prevents build running on windows platform
`external` files are defined as an array
Rollup and plugins are not on latest versions

### Checklist

#### QA

### Testing instructions

Ideally QA with windows machine will run build script and I can release a beta to allow brush testing of distributed files


rollup.config.mjs
```diff
@@ -40,44 +35,77 @@ export default {
...
-        /**
-         * This expands the relative paths to absolute paths
-         * e.g. src/components/foo becomes /project/src/components/foo.js
-         * */
-        fileURLToPath(new URL(file, import.meta.url)),
-      ]),
+      // .filter(file => {
+      //   // Additional safety check to exclude SVGs
+      //   return !file.endsWith('.svg');
+      // })
```
Parsium: **comment**: could we remove this unused code?

```suggestion
```

package.json
```diff
@@ -71,10 +71,10 @@
     "@dotenvx/dotenvx": "^1.25.1",
     "@playwright/experimental-ct-react": "^1.52.0",
     "@playwright/test": "^1.52.0",
-    "@rollup/plugin-commonjs": "^28.0.2",
+    "@rollup/plugin-commonjs": "^25.0.8",
```
Parsium: **question**: I noticed we're downgrading this plugin, is there a particular reason for this?

package.json
```diff
@@ -71,10 +71,10 @@
...
     "@rollup/plugin-json": "^6.1.0",
-    "@rollup/plugin-node-resolve": "^16.0.0",
-    "@rollup/plugin-terser": "^0.3.0",
+    "@rollup/plugin-node-resolve": "^15.3.1",
```
Parsium: **question**: similarly is there a particular reason why we need to downgrade this plugin?

package.json
edleeks87: I don't think they need to be let me confirm

package.json
edleeks87: I don't think they need to be let me confirm

package.json
edleeks87: Confirmed they can be reverted thanks

rollup.config.mjs
```diff
@@ -40,44 +35,73 @@ export default {
...
       preferBuiltins: true,
       browser: true,
       moduleDirectories: ["node_modules"],
-      resolveOnly: [
-        /* Exclude @swc/helpers from custom resolution */
-        /^(?!@swc\/helpers)/,
-        /* Exclude styled-components from custom resolution */
-        /^(?!styled-components)/,
-      ],
+      // resolveOnly is deprecated in favor of exportConditions
```
nineteen88: suggestion(non-blocking): not sure this comment really adds anything within the context of what a developer can see so could be removed

rollup.config.mjs
```diff
@@ -40,44 +35,73 @@ export default {
...
+    if (id.startsWith("./") || id.startsWith("../")) {
+      return false;
+    }
+    
+    // Externalise node_modules
+    if (id.includes("node_modules")) {
+      return true;
+    }
+    
+    // Externalize specific packages
```
nineteen88: nitpick(non-blocking): we have Anglo and American spellings next to each other - should we look for consistency?

rollup.config.mjs
edleeks87: yep sorry was more for my own sanity checks when I was working through it

nineteen88 (approved):

nuria1110 (approved):

PR #7446, opened 2025-07-21
chore(typescript): upgrade dependency to latest version (5.8.3)

edleeks87:
### Proposed behaviour

Bumps `typescript` to 5.8.3
Updates:
 - @typescript-eslint/parser
 - @typescript-eslint/eslint-plugin
 - @types/node @types/react
 - @types/react-dom
 - ts-node
 - ts-jest
 
### Current behaviour

Typescript is on 5.3 and other dependencies are not on latest

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
     "storybook": "~8.6.12",
     "storybook-addon-pseudo-states": "~4.0.2",
     "styled-components": "^5.3.11",
-    "typescript": "~5.3.0",
+    "typescript": "^5.8.3",
     "url-loader": "^4.1.1",
     "uuid": "^8.3.2",
     "vite": "^6.3.4",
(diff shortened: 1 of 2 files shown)


tomdavies73 (approved):

package.json
Parsium, line 1: **suggestion(non-blocking)**: I noticed we have some peer dependency warnings due to `eslint-plugin-testing-library` expecting an older version of `eslint`.

I appreciate that this upgrade does not cause the warnings, but for general housekeeping, would you be happy to bump `eslint-plugin-testing-library` to the latest version as well?

edleeks87: Good spot. I'll do it in a new PR 👍 as I need this to start on the monorepo stuff

PR #7452, opened 2025-07-22
test(split-button): add play functions

DobroTora:
### Proposed behaviour

capture popover open for all variations (eg primary/destructive etc) and sizes
capture focus state for child button
capture disabled state for a child button
capture it with Icons in child buttons
capture popover positions and alignments

### Checklist

#### QA


src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -0,0 +1,387 @@
...
+  ),
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+    const canvas = within(canvasElement);
+    const toggleButtons = canvas.getAllByRole("button", { name: /show more/i });
+
+    for (const toggle of toggleButtons) {
+      await userEvent.click(toggle);
+      await userInteractionPause(500);
+      await userEvent.keyboard("{Escape}");
```
edleeks87: question: doesn't pressing `Escape` here close the popover? We want to capture that in chromatic

edleeks87: @DobroTora all but one of the tests still calls escape and so we're taking snapshots of the component closed

src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -248,36 +289,175 @@ export const SizeVariations: Story = {
...
     await userEvent.tab();
-    expect(document.activeElement).toHaveAttribute("data-element", "toggle-button");
+    expect(document.activeElement).toHaveAttribute(
+      "data-element",
+      "toggle-button",
+    );
+
+    await userEvent.keyboard("{Escape}");
+    await userInteractionPause(300);
+    expect(toggleButton).toHaveAttribute("aria-expanded", "false");
```
edleeks87: comment: I don't think we need to test this type of thing. We should be using play functions to capture the state of components after user interaction so that chromatic can snapshot the SplitButton expanded and we can capture the baseline for any future regressions

src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -0,0 +1,384 @@
...
+  },
+  decorators: [
+    (StoryToRender) => (
+      <DefaultDecorator>
+        <StoryToRender />
+      </DefaultDecorator>
+    ),
+  ],
+};
+SizeVariations.storyName = "Sizes";
```
damienrobson-sage: **suggestion**: break this out into individual tests, one for each size, otherwise we're only capturing a snapshot of the large split button being opened, not all three. This will probably apply to each story that has a loop/map in it's render function.

src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -0,0 +1,384 @@
...
+    focusTarget.focus();
+    await userInteractionPause(600);
+
+    await userEvent.keyboard("{Escape}");
+    await userInteractionPause(300);
+
+    const toggleButtonElement = canvasElement.querySelector(
+      '[data-element="toggle-button"]',
+    );
+    expect(document.activeElement).toBe(toggleButtonElement);
```
damienrobson-sage: **suggestion**: rather than use `document.activeElement`, you could do this instead:
```javascript
await expect(toggleButtonElement).toBeFocused()
```

DobroTora: @damienrobson-sage , so toBeFocused() is a Playwright specific matcher, not a Jest-DOM matcher specific to Storybook. I have used toHaveFocus() method instead.

damienrobson-sage: No worries - I always get the selectors the wrong way around 😆

DobroTora: @Damienrobson-sage so toBeFocused() is a Playwright specific matcher, not a Jest-DOM matcher specific to Storybook. I have used toHaveFocus() method instead.

damienrobson-sage (approved):

src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -0,0 +1,408 @@
...
+  },
+};
+
+const Wrapper = (props: { children: React.ReactNode }) => (
+  <Box display="flex" flexDirection="column" gap="32px">
+    {props.children}
+  </Box>
+);
+
+export const PrimaryButtonType: Story = {
```
edleeks87: comment (non-blocking): some of these stories can be combined to reduce the number of snapshots we're using. For example we can combine some of the variant/size stories with the focus state stuff. I'm happy for this to be merged as is and we can do the consolidation in a second pass after we review all of our play function tests

edleeks87 (approved):

edleeks87:
Nice work @DobroTora FYI I've merged master in and moved it out of draft so the chromatic workflows run for QA to review

src/components/split-button/split-button.interactions.stories.tsx
```diff
@@ -0,0 +1,319 @@
...
+    if (!allowInteractions()) return;
+    const canvas = within(canvasElement);
+
+    const primaryToggle = canvas.getAllByRole("button", {
+      name: /show more/i,
+    })[0];
+    await userEvent.click(primaryToggle);
+    await userInteractionPause(300);
+    expect(primaryToggle).toHaveAttribute("aria-expanded", "true");
+    await userEvent.click(primaryToggle);
```
edleeks87: Comment: by clicking the toggles again we are closing the popover so we won't capture the open states in chromatic. Whilst it isn't replicating how a user will interact fully I think there's value in opening all the popovers and snapshotting then, this would need to be done for all the combined stories.

edleeks87 (dismissed):

damienrobson-sage (dismissed):

stephenogorman (approved):

PR #7458, opened 2025-07-23
Remove uncontrolled input support

damienrobson-sage:
### Proposed behaviour

Remove support for uncontrolled usage from the following components:

- [ ] Textbox
- [ ] Decimal
- [ ] Number
- [ ] SimpleSelect
- [ ] Filterable Select
- [ ] MultiSelect
- [ ] GroupedCharacter
- [ ] NumeralDate
- [ ] Search
- [ ] Checkbox
- [ ] RadioButton
- [ ] Switch
- [ ] SimpleColorPicker
- [ ] AdvancedColorPicker
- [ ] ButtonToggle
- [ ] Password

### Current behaviour

Our input-based components support uncontrolled usage

### Checklist

#### QA

### DIFF
--- a/src/components/decimal/decimal.component.tsx
+++ b/src/components/decimal/decimal.component.tsx
@@ -1,10 +1,4 @@
-import React, {
-  useState,
-  useEffect,
-  useRef,
-  useCallback,
-  useContext,
-} from "react";
+import React, { useState, useEffect, useCallback, useContext } from "react";
 import invariant from "invariant";
 
 import Textbox, { CommonTextboxProps } from "../textbox";
@@ -30,14 +24,12 @@ export interface DecimalProps
   align?: "left" | "right";
   /** Allow an empty value instead of defaulting to 0.00 */
   allowEmptyValue?: boolean;
-  /** The default value of the input if it's meant to be used as an uncontrolled component */
-  defaultValue?: string;
   /** The input id */
   id?: string;
   /** The width of the input as a percentage */
   inputWidth?: number;
-  /** Handler for change event if input is meant to be used as a controlled component */
-  onChange?: (ev: CustomEvent) => void;
+  /** Handler for change event */
+  onChange: (ev: CustomEvent) => void;
   /** Handler for blur event */
   onBlur?: (ev: CustomEvent) => void;
   /** The input name */
@@ -62,19 +54,16 @@ export interface DecimalProps
     | 15;
   /** If true, the component will be read-only */
   readOnly?: boolean;
-  /** The value of the input if it's used as a controlled component */
-  value?: string;
+  /** The value of the input */
+  value: string;
   /** The locale string - default en */
   locale?: string;
 }
-
-let deprecateUncontrolledWarnTriggered = false;
 
 export const Decimal = React.forwardRef(
   (
     {
       align = "right",
-      defaultValue,
       precision = 2,
       inputWidth,
       readOnly,
@@ -173,6 +162,8 @@ export const Decimal = React.forwardRef(
     const prevPrecisionValue = usePrevious(precision);
 
     useEffect(() => {
+      // Skipped coverage test - see notes in test file
+      /* istanbul ignore if */
       if (prevPrecisionValue && prevPrecisionValue !== precision) {
         Logger.error(
           "Decimal `precision` prop has changed value. Changing the Decimal `precision` prop has no effect.",
@@ -230,7 +221,7 @@ export const Decimal = React.forwardRef(
       [getSeparator, removeDelimiters],
     );
 
-    const decimalValue = getSafeValueProp(defaultValue || value || emptyValue);
+    const decimalValue = getSafeValueProp(value || emptyValue);
     const [stateValue, setStateValue] = useState(
       isNaN(toStandardDecimal(decimalValue))
         ? decimalValue
@@ -253,6 +244,7 @@ export const Decimal = React.forwardRef(
     const handleOnChange = (ev: React.ChangeEvent<HTMLInputElement>) => {
       const { value: val } = ev.target;
       setStateValue(val);
+      /* istanbul ignore else */
       if (onChange) onChange(createEvent(val));
     };
 
@@ -277,35 +269,18 @@ export const Decimal = React.forwardRef(
     };
 
     const isControlled = value !== undefined;
-    const prevControlledRef = useRef<boolean | null>();
-
-    useEffect(() => {
-      const message =
-        "Input elements should not switch from uncontrolled to controlled (or vice versa). " +
-        "Decide between using a controlled or uncontrolled input element for the lifetime of the component";
-
-      invariant(prevControlledRef.current !== isControlled, message);
-
-      prevControlledRef.current = isControlled;
-
-      return () => {
-        prevControlledRef.current = null;
-      };
-    }, [isControlled]);
 
     const prevValue = usePrevious(value);
 
     useEffect(() => {
       const standardDecimalValue = toStandardDecimal(stateValue);
 
-      if (isControlled) {
-        const valueProp = getSafeValueProp(value);
-        if (standardDecimalValue !== valueProp) {
-          if (valueProp === "" && prevValue === "") {
-            setStateValue(formatValue(emptyValue));
-          } else {
-            setStateValue(formatValue(valueProp));
-          }
+      const valueProp = getSafeValueProp(value);
+      if (standardDecimalValue !== valueProp) {
+        if (valueProp === "" && prevValue === "") {
+          setStateValue(formatValue(emptyValue));
+        } else {
+          setStateValue(formatValue(valueProp));
         }
       }
     }, [
@@ -318,13 +293,6 @@ export const Decimal = React.forwardRef(
       toStandardDecimal,
       value,
     ]);
-
-    if (!deprecateUncontrolledWarnTriggered && !isControlled) {
-      deprecateUncontrolledWarnTriggered = true;
-      Logger.deprecate(
-        "Uncontrolled behaviour in `Decimal` is deprecated and support will soon be removed. Please make sure all your inputs are controlled.",
-      );
-    }
 
     return (
       <>
--- a/src/components/decimal/decimal.pw.tsx
+++ b/src/components/decimal/decimal.pw.tsx
@@ -104,7 +104,7 @@ test.describe("check props for Decimal component", () => {
       mount,
       page,
     }) => {
-      await mount(<Decimal precision={precision} />);
+      await mount(<DefaultStory precision={precision} />);
 
       const commonDataElementInputPreviewElement =
         commonDataElementInputPreview(page);
@@ -139,7 +139,7 @@ test.describe("check props for Decimal component", () => {
       mount,
       page,
     }) => {
-      await mount(<Decimal locale={locale} precision={3} />);
+      await mount(<DefaultStory locale={locale} precision={3} />);
 
       const commonDataElementInputPreviewElement =
         commonDataElementInputPreview(page);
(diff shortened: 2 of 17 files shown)


src/components/button-toggle/button-toggle.stories.tsx
```diff
@@ -363,49 +490,77 @@ export const DisabledButton: Story = () => {
...
+  function onChangeHandler(
+    event: React.MouseEvent<HTMLButtonElement>,
+    selectedValue?: string,
+  ) {
+    setValue(selectedValue as string);
+  }
+
+  return (
+    <Box
+      width={inChromatic ? "175px" : "375px"}
```
Parsium: **polish**: what do you think about making this the default width? That way we'd ensure visitors don't notice any Chromatic logic in the example code:

```suggestion
      width="175px"
```

src/components/decimal/decimal.component.tsx
```diff
@@ -253,6 +244,7 @@ export const Decimal = React.forwardRef(
     const handleOnChange = (ev: React.ChangeEvent<HTMLInputElement>) => {
       const { value: val } = ev.target;
       setStateValue(val);
+      /* istanbul ignore else */
       if (onChange) onChange(createEvent(val));
```
mihai-albu-sage: nitpick: since `onChange` is a required prop in the TypeScript definition, the `if (onChange)` check and the `istanbul ignore` might be unnecessary. We could probably just call `onChange` directly.

src/components/decimal/decimal.component.tsx
```diff
@@ -173,6 +162,8 @@ export const Decimal = React.forwardRef(
     const prevPrecisionValue = usePrevious(precision);
 
     useEffect(() => {
+      // Skipped coverage test - see notes in test file
+      /* istanbul ignore if */
```
mihai-albu-sage: nitpick: It looks like coverage works correctly here. I enabled the skipped test and removed the `istanbul ignore`, and everything still passes with 100% coverage.

src/components/number/number-test.stories.tsx
```diff
@@ -104,30 +144,40 @@ Validation.parameters = {
...
         error="Error Message"
         inputHint="Hint text"
         mb={2}
+        onChange={(e) => setState(e.target.value)}
       />
-      <Number label="Number" value="123456" warning="Warning Message" mb={2} />
       <Number
-        validationMessagePositionTop={false}
         label="Number"
         value="123456"
```
mihai-albu-sage: suggestion: the value `"123456"` should be replaced with `{state}`

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -113,7 +107,7 @@ export interface NumeralDateProps
   /** Blur event handler */
   onBlur?: (ev: NumeralDateEvent) => void;
   /** Change event handler */
-  onChange?: (ev: NumeralDateEvent) => void;
+  onChange: (ev: NumeralDateEvent) => void;
```
mihai-albu-sage, line 110: blocking: `value` and `onChange` need to be added to the test stories from `numeral-date-test.stories.tsx`

src/components/radio-button/radio-button-test.stories.tsx
```diff
@@ -30,151 +30,172 @@ export default {
...
+  chromatic: { disableSnapshot: false },
+  themeProvider: { chromatic: { theme: "sage" } },
+};
+
+export const WithValidationsOnRadioGroup = ({ ...args }) => {
+  const [value, setValue] = useState("radio1");
+
+  return (
+    <>
+      <RadioButtonGroup
```
mihai-albu-sage, line 125: suggestion: I noticed some odd behavior in this story when selecting/hovering the radio buttons from these groups due to the fact they are sharing the same name, and the radio buttons also share the same id and state values.

src/components/decimal/decimal.component.tsx
```diff
@@ -277,35 +269,18 @@ export const Decimal = React.forwardRef(
     };
 
     const isControlled = value !== undefined;
```
Parsium: **suggestion**: looks this code isn't needed anymore?

```suggestion
```

src/components/decimal/decimal.pw.tsx
Parsium, line 1: **suggestion(non-blocking)**: what do you think about removing [these event tests](https://github.com/Sage/carbon/blob/e9ef6ef311bdac82203f853ea917b937a05f4b44/src/components/decimal/decimal.pw.tsx#L406-L451)?

Following the changes, I'm wondering how much value they are providing and if they would be better as Jest tests.

src/components/decimal/decimal.test.tsx
```diff
@@ -1468,3 +341,19 @@ test("component should render without invariant firing in strict mode", () => {
 
   consoleErrorSpy.mockRestore();
 });
+
+// Skipped coverage test - Logger.error not being called in this test despite the precision prop changing
+// and the useEffect being hit (line 164 of decimal.component.tsx).
+test.skip("logs an error when precision changes", () => {
```
Parsium: **polish**: I noticed this test is being called after unskipping it and removing the relevant ignore coverage directives in `Decimal`.

If this is the case, would we be able to unskip this and remove the comments?

src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -296,21 +287,7 @@ export const NumeralDate = forwardRef<NumeralDateHandle, NumeralDateProps>(
...
-        isControlled.current === (value !== undefined),
-        modeSwitchedMessage,
-      );
-    }, [value]);
-
-    const [dateValue, setDateValue] = useState<NumeralDateValue>({
-      ...(initialValue ||
-        Object.fromEntries(dateFormat.map((datePart) => [datePart, ""]))),
-    });
+    const [dateValue, setDateValue] = useState<NumeralDateValue>(value);
```
Parsium: **suggestion**: what do you think about removing this state?

We could use `value` directly if it should always be provided.

src/components/switch/switch.component.tsx
```diff
@@ -19,14 +18,12 @@ import HintText from "../../__internal__/hint-text";
...
-  extends CommonCheckableInputProps,
+  extends Omit<CommonCheckableInputProps, "defaultChecked">,
     MarginProps,
     TagProps {
   /** Breakpoint for adaptive label (inline labels change to top aligned). Enables the adaptive behaviour when set */
   adaptiveLabelBreakpoint?: number;
-  /** Set the default value of the Switch if component is meant to be used as uncontrolled */
-  defaultChecked?: boolean;
-  /** When true label is inline */
+  /** [Legacy] When true label is inline */
```
Parsium: **question**: It appears the Fusion DS specs still support inline labels for `Switch`.

Would it be worth us re-visiting if this needs deprecation at a later date to minimise the number of changes?

src/components/simple-color-picker/simple-color-picker.test.tsx
```diff
@@ -508,3 +547,26 @@ test.each(["error", "warning", "info"])(
...
+      <SimpleColor value="transparent" />
+      <SimpleColor value="#00A376" />
+    </SimpleColorPicker>,
+  );
+
+  // eslint-disable-next-line jest-dom/prefer-to-have-value
+  expect(screen.getByRole("radio", { checked: true })).toHaveAttribute(
+    "value",
+    "transparent",
+  );
```
Parsium: **suggestion(non-blocking)**: what do you think about adding an `aria-label` prop onto `SimpleColor` and using that to workaround the lint warning?:

```suggestion
      <SimpleColor value="transparent" aria-label="transparent" />
      <SimpleColor value="#00A376" />
    </SimpleColorPicker>,
  );

  expect(screen.getByRole("radio", { name: "transparent" })).toBeChecked();
```

It turns out we document using `aria-label` on `SimpleColor`, but its not been added to its prop interface: [Link to docs](https://carbon.sage.com/?path=/docs/simple-color-picker--docs#default)

src/components/select/multi-select/multi-select.pw.tsx
```diff
@@ -1818,7 +1830,12 @@ test.describe("Accessibility tests for MultiSelect component", () => {
...
-      await mount(<MultiSelectDefaultValueComponent defaultValue={[value]} />);
+      await mount(
+        <MultiSelectDefaultValueComponent
+          onChange={() => {}}
+          value={[value]}
+        />,
+      );
       await checkAccessibility(page);
     });
   });
```
Parsium: **suggestion**: would we be able to remove this test, given this prop no longer exists?

src/components/select/simple-select/simple-select.test.tsx
Parsium, line 1: **question**: I noticed several parameterised tests are now structured differently:

```ts
const values = [/**/]
values.forEach(() => {
  test("my test", () => {
    // ...
  })
});

```

Is there a benefit for this?

src/components/select/multi-select/multi-select.test.tsx
Parsium, line 1: **question**: I noticed several parameterised tests are now structured differently:

```ts
const values = [/**/]
values.forEach(() => {
  test("my test", () => {
    // ...
  })
});

```

Is there a benefit for this?

src/components/search/search.component.tsx
```diff
@@ -128,35 +120,19 @@ export const Search = React.forwardRef<SearchHandle, SearchProps>(
...
-    const isSearchValueEmpty = !isControlled
-      ? searchValue.length === 0
-      : value.length === 0;
+    const isSearchValueEmpty = value.length === 0;
 
     let buttonProps = {};
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+      /* istanbul ignore else */
       if (onChange) {
```
mihai-albu-sage: suggestion: `onChange` is now required, we could remove the test and the `istanbul ignore`

src/components/select/simple-select/simple-select.stories.tsx
```diff
@@ -78,6 +89,36 @@ export const Required: Story = () => {
...
+export const IsOptional = () => {
+  const [value, setValue] = useState("");
+
+  return (
+    <Box height={250}>
+      <Select
+        name="optional"
+        id="optional"
+        label="Foreground Color"
+        isOptional
```
mihai-albu-sage: suggestion: the `isOptional` prop does not exist. Same issue in the `multi-select`, `filetrable-select` and `textbox` story files. Only place where I can see it added is in the `textarea` component. Perhaps it should also be added to the `textbox` component.

damienrobson-sage: James recently did some work to remove the `isOptional` prop entirely after I deprecated it a few months ago. I'll tidy up the stories 👍

src/components/select/simple-select/simple-select.test.tsx
damienrobson-sage: The reason was around cleaning up the files and reducing redundancy. A lot of these tests used the same base logic with the only difference being execution variables so it seemed like the sensible thing to do was to simplify them.

src/components/select/multi-select/multi-select.test.tsx
damienrobson-sage: As above

src/components/checkbox/checkbox-group/checkbox-group.stories.tsx
Parsium, line 1: **suggestion**: could we move the declaration of state in some of the stories?

Placing the state in the scope of the top-level component would prevent new state from being created on each re-render.

```tsx
render: function Render(args) {
  const [checkedValues, setCheckedValues] = useState({
    "Apple": false,
    "Banana": false,
    "Cherry": false,
    "Date": false,
  });

  return (
    <CheckboxGroup {...args}>
      {Object.entries(checkedValues).map([label, isChecked] => (
        <Checkbox
          key={label}
          name={label}
          label={label}
          checked={isChecked}
          onChange={(ev) => setCheckedValues(ev.target.checked)}
        />
      ))}
    </CheckboxGroup>
  )
}
```

mihai-albu-sage (approved):

PR #7460, opened 2025-07-24
TextEditor - ensure `onChange` fires only when text content is modified or styled (FE-7384)

### LINKED ISSUE #7384: TextEditor: Unexpected behabiour when used as a controlled component

Parsium:
resolves #7384

### Proposed behaviour

For `TextEditor`:

- Ensure `onChange` prop is called only when the editor's text content is modified or styled.
- Prevent changes to `value` prop from continually overwriting Lexical's internal state.
- Introduce a new `initialValue` prop, which is an alias of the `value` prop.
- Add documentation to clarify that the editor is an uncontrolled component, with an example showing how to update its content externally.

### Current behaviour

- `onChange` is called whenever the editor is clicked.
- `onChange` is called whenever the `value` prop changes.
- Updating the `value` prop overwrites Lexical's internal state, causing two main problems: expensive re-renders on each update, and loss of editor state like cursor position and focus when `value` changes again.

### Checklist

#### QA


src/components/text-editor/text-editor.component.tsx
```diff
@@ -211,25 +209,22 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
       };
     }, [namespace, readOnly]);
 
-    // OnChangePlugin is tested separately
-    /* istanbul ignore next */
     const handleChange = useCallback(
-      (newState: EditorState) => {
-        const currentTextContent = newState.read(() => {
+      (editorState: EditorState, editor: LexicalEditor) => {
```
edleeks87, line 225: question: as `LexicalEditor` provides a `getEditorState` callback could we just have a parameter of `editor` here and remove `editorState`?

src/components/text-editor/__internal__/plugins/Toolbar/toolbar.test.tsx
```diff
@@ -181,20 +182,27 @@ describe("Events", () => {
...
+  it("calls 'onChange' when the bold button is clicked while text is selected", async () => {
     const user = userEvent.setup();
     const mockOnChange = jest.fn();
 
-    render(<TextEditor labelText="foo" onChange={mockOnChange} />);
+    render(
+      <TextEditor
+        labelText="foo"
+        onChange={mockOnChange}
+        initialValue={createFromHTML("<p>Hello world!</p>")}
```
damienrobson-sage, line 193: **suggestion (non-blocking)**: as this is the same value for every test, you could lift it up into a global variable so it's only fired once, which might save some coin.

src/components/text-editor/text-editor.component.tsx
Parsium: It seems the official `onChange` plugin expects an `onChange` handler with the above signature, so I'm not sure we can simplify this further.

dariusbercea-sage (approved):

src/components/text-editor/__internal__/plugins/Toolbar/toolbar.test.tsx
Parsium: I'll leave this for now if you're happy, since it's only a few tests. If we add more later, we can certainly move it to global scope

PR #7463, opened 2025-07-25
feat(version-picker): remove beta releases from version-picker

nineteen88:
### Proposed behaviour

- remove beta releases from appearing in the version picker list
- additionally, update typescript config to include .storybook folder for linting and resolve lint issues
- include a dev dependency for the semver types ("@types/semver")

Screenshot 2025-07-25 at 14 23 35

### Current behaviour

- The Version Picker list includes beta releases which we don't really want visitors to be able to see

Screenshot 2025-07-25 at 14 24 20

### Checklist


.storybook/version-picker/index.tsx
```diff
@@ -12,11 +12,16 @@ import fetchData from "./fetch-data";
 
 const getDisplayedItems = (
   versions: Record<string, string>,
-  onClick: TooltipLinkListLink["onClick"]
+  onClick: TooltipLinkListLink["onClick"],
 ) => {
-  let formattedVersions: TooltipLinkListLink[] = [];
+  const formattedVersions: TooltipLinkListLink[] = [];
 
   for (const [key, value] of Object.entries(versions)) {
```
edleeks87: suggestion (non-blocking): we could utilise `reduce` here but what you have is fine if you prefer a more verbose approach

```
const formattedVersions = Object.entries(versions).reduce<TooltipLinkListLink[]>((acc, [key, value]) => {
    if (!key.match(/-beta\.\d+$/)) {
      acc.push({
        id: key,
        title: key,
        onClick,
        active: false,
        href: value,
        target: "_blank",
      });
    }
    return acc;
  }, []).sort((a, b) => compareBuild(b.id, a.id));

edleeks87 (dismissed):

damienrobson-sage (approved):
LGTM 👍

PR #7468, opened 2025-07-30
feat(link): add subtle variant & underline prop

tomdavies73:
### Proposed behaviour

Adds the subtle variant to `Link`, also adds the `underline` prop, which when set to `"false"` removes `text-decoration: underline` from the component.

Screenshot 2025-07-30 at 12 53 15

Screenshot 2025-07-30 at 12 53 02

Screenshot 2025-07-30 at 12 53 22

A test story has also been added to create a baseline for the use of both props when used simultaneously.

### Current behaviour

N/A

### Checklist

#### QA

### DIFF
--- a/src/components/link/link.component.tsx
+++ b/src/components/link/link.component.tsx
@@ -73,6 +73,7 @@ export const Link = React.forwardRef<
       iconAlign = "left",
       isSkipLink,
       disabled = false,
+      underline = "always",
       ariaLabel,
       rel,
       tooltipMessage,
@@ -187,6 +188,7 @@ export const Link = React.forwardRef<
       <StyledLink
         isSkipLink={isSkipLink}
         disabled={isDisabled}
+        underline={underline}
         iconAlign={iconAlign}
         className={className}
         hasContent={Boolean(children)}
--- a/src/components/link/link.stories.tsx
+++ b/src/components/link/link.stories.tsx
@@ -240,6 +240,30 @@ export const OnADarkBackground: Story = () => {
         </React.Fragment>
       ))}
       <br />
+      <Link
+        href="https://carbon.sage.com"
+        target="_blank"
+        rel="noreferrer noopener"
+        isDarkBackground
+        variant="subtle"
+      >
+        This is a link
+      </Link>
+      {(["left", "right"] as const).map((align) => (
+        <React.Fragment key={`${align}-subtle-variant`}>
+          <br />
+          <Link
+            icon="settings"
+            isDarkBackground
+            variant="subtle"
+            iconAlign={align}
+            href="#foo"
+          >
+            This is a link
+          </Link>
+        </React.Fragment>
+      ))}
+      <br />
       <Link isDarkBackground disabled>
         This is a link
       </Link>
--- a/src/components/link/link.stories.tsx
+++ b/src/components/link/link.stories.tsx
@@ -33,6 +33,25 @@ export const WithDisabled = () => {
 };
 WithDisabled.storyName = "With Disabled";
 
+export const WithUnderlineOnlyOnHover = () => {
+  return (
+    <Link underline="hover" href="#foo">
+      This is an anchor link with an underline applied on hover
+    </Link>
+  );
+};
+WithUnderlineOnlyOnHover.storyName = "With Underline Only On Hover";
+WithUnderlineOnlyOnHover.parameters = { chromatic: { disableSnapshot: true } };
+
+export const WithNoUnderline = () => {
+  return (
+    <Link underline="never" href="#foo">
+      This is an anchor link with no underline
+    </Link>
+  );
+};
+WithNoUnderline.storyName = "Without Underline";
+
 export const WithIcon: Story = () => {
   return (
     <Link icon="settings" href="#foo">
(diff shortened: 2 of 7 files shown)


mihai-albu-sage (approved):

src/components/link/link.component.tsx
```diff
@@ -73,6 +73,7 @@ export const Link = React.forwardRef<
       iconAlign = "left",
       isSkipLink,
       disabled = false,
+      underline = true,
```
nineteen88: suggestion(non-blocking): this feels a little odd semantically as it's being used to remove something

It would probably feeler a bit better if it felt like you were opting out of the default behaviour, so calling it something like `noUnderline` which defaults to false might be a bit consuming experience.

Alternatively, other component libraries handle this slightly differently and use underline as a string to switch behaviour rather than just a boolean.

i.e. `underline = [none, hover, always]`

This second suggestion is a bit of scope creep but I don't think it's impossible that consumers might ask for the ability to have hover underline. Which actually raises the point, we haven't established whether or not it should have hover underline for the use case of it in the Global Nav

src/components/link/link.stories.tsx
```diff
@@ -33,6 +33,13 @@ export const WithDisabled = () => {
 };
 WithDisabled.storyName = "With Disabled";
 
+export const WithNoUnderline = () => {
+  return (
+    <Link underline={false}>This is an anchor link with no underline</Link>
```
nineteen88: comment: the example does not contain a href

As it doesn't contain a href, it's just appearing as plain text, you don't even get a pointer cursor when you hover over it nor any of the focus state

src/components/link/link.test.tsx
```diff
@@ -241,6 +241,60 @@ test("renders with custom data tags", () => {
   expect(screen.getByTestId("foo")).toHaveAttribute("data-element", "bar");
 });
 
+describe("underline prop", () => {
+  test("always shows underline text-decoration", () => {
```
nineteen88: suggestion: if the test is within a `describe` block it should be an `it` and be a test case based on what you are describing

```suggestion
  describe("when using the underline prop", () =>{
    it("should display an underline for all states when the prop is `always`", () => {
        ...
```

src/components/link/link.test.tsx
```diff
@@ -241,6 +241,60 @@ test("renders with custom data tags", () => {
...
+        Test Content
+      </Link>,
+    );
+
+    const linkElement = screen.getByRole("link");
+    await user.hover(linkElement);
+
+    expect(linkElement).toHaveStyle("text-decoration: none");
+  });
+});
```
nineteen88, line 288: comment: we are missing a test case for when the underline prop is set to `hover`

src/components/link/link.style.ts
```diff
@@ -180,7 +185,17 @@ const StyledLink = styled.span.attrs(applyBaseTheme)<
...
       > a,
       > button {
-        text-decoration: ${hasContent ? "underline" : "none"};
+        text-decoration: ${hasContent && underline === "always"
+          ? "underline"
+          : "none"};
+
+        :hover {
+          text-decoration: ${
+            /* istanbul ignore next - having issues testing CSS :hover pseudo-states in jsdom */
```
mihai-albu-sage, line 195: **suggestion (non-blocking):**  Since we’re testing a hover style, we can use `toHaveStyleRule` as it can check for pseudo‑classes like `:hover`. In this case, the `:hover` rule is defined on the styled `<span>` wrapper (which wraps the link) and targets the `<a>` inside, so the assertion should be made on the wrapper element rather than the link itself. It could look something like this:
```
it("applies text decoration underline on hover", async () => {
    render(
      <Link href="foo.com" underline="hover">
        Test Content
      </Link>,
    );

    const linkElement = screen.getByRole("link");
    
    // eslint-disable-next-line testing-library/no-node-access
    const linkSpanWrapper = linkElement.parentElement;
    
    expect(linkSpanWrapper).toHaveStyleRule('text-decoration', 'underline', {modifier: '> a:hover'})
  });
```

However, this method currently relies on direct node access, it might be worth exploring a cleaner way to structure the component or test so that we can query the wrapper without traversing the DOM manually.

src/components/link/link.test.tsx
tomdavies73: I've added this as a playwright test as there was no good way to test against the hover styling being conditionally applied in a unit test

src/components/link/link.style.ts
tomdavies73: Thanks for this Mihai, 

I've tested this, and it does work and checks that the rule is applied, but there is no way to test if this is applied specifically on hover, it just asserts that the rule exists. 

So I think I'm just going to stick with an Istanbul ignore in the style file, and keep the replacement test in playwright which checks the styling being applied against a hover 👍

mihai-albu-sage (dismissed):

mihai-albu-sage (dismissed):

nineteen88 (approved):

mihai-albu-sage (approved):

DobroTora (approved):
LGTM

DobroTora (approved):

PR #7471, opened 2025-07-31
[Adaptive Sidebar] Implement show/hide functionality

damienrobson-sage:
### Proposed behaviour

Add a property to indicate a third "state" to the `AdaptiveSidebar` that allows it to be hidden. This would visually remove it from the screen but not destroy the underlying elements, attached scripts, etc.

### Current behaviour

Currently, `AdaptiveSidebar` can only be open or closed, which is not fit for purpose when it is used to store and display e.g. agent chat or complex forms.

### Checklist

#### QA

### Additional context

Accessibility and DS have approved this approach and have no concerns as to the suggested implementation.

### Testing instructions

A story has been added to the `AdaptiveSidebar` documentation to demonstrate the new `hidden` property: `Hidden`

### DIFF
--- a/src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
+++ b/src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
@@ -30,6 +30,8 @@ export interface AdaptiveSidebarProps
   children?: React.ReactNode;
   /** The height of the sidebar, relative to the wrapping component */
   height?: string;
+  /** Whether the sidebar is hidden from view. In this state, the adaptive sidebar will continue to receive updates, etc. but will not be visible to users */
+  hidden?: boolean;
   /** Whether the sidebar is open or closed */
   open: boolean;
   /** Whether to render the sidebar as a modal component instead of as an inline sidebar */
@@ -44,6 +46,7 @@ export const AdaptiveSidebar = ({
   borderColor = "none",
   children,
   height = "100%",
+  hidden = false,
   open,
   renderAsModal = false,
   width = "320px",
@@ -67,18 +70,33 @@ export const AdaptiveSidebar = ({
     }
   }, [open]);
 
+  // Remove inert attribute from elements when sidebar is hidden
+  /* istanbul ignore next */
+  useEffect(() => {
+    if (!hidden || !open) return;
+
+    const inertElements = document.querySelectorAll("[inert]");
+
+    inertElements.forEach((element) => {
+      element.removeAttribute("inert");
+    });
+  }, [hidden, open]);
+
   if (renderAsModal || !largeScreen) {
     return (
       <StyledSidebar
-        className="adaptive-sidebar-modal-view"
         backgroundColor={backgroundColor}
+        className="adaptive-sidebar-modal-view"
+        data-role={"adaptive-sidebar-modal-view"}
+        enableBackgroundUI={open && hidden}
+        hidden={hidden}
         open={open}
         p={0}
         ref={adaptiveSidebarRef}
       >
         <Box
-          height="100%"
           data-role="adaptive-sidebar-content-wrapper"
+          height="100%"
           {...colours}
         >
           {children}
@@ -94,6 +112,7 @@ export const AdaptiveSidebar = ({
       data-element="adaptive-sidebar"
       data-role="adaptive-sidebar"
       height={height}
+      hidden={hidden}
       ref={adaptiveSidebarRef}
       role="region"
       tabIndex={-1}
--- a/src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
+++ b/src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
@@ -1,5 +1,5 @@
 import { Meta, StoryObj } from "@storybook/react";
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { allModes } from "../../../.storybook/modes";
 import isChromatic from "../../../.storybook/isChromatic";
 import generateStyledSystemProps from "../../../.storybook/utils/styled-system-props";
@@ -13,6 +13,7 @@ import Typography from "../typography";
 import Button from "../button";
 import Hr from "../hr";
 import { Menu, MenuItem } from "../menu";
+import SplitButton from "../split-button";
 
 const styledSystemProps = generateStyledSystemProps({
   margin: true,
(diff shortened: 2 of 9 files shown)


src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -30,6 +30,8 @@ export interface AdaptiveSidebarProps
   children?: React.ReactNode;
   /** The height of the sidebar, relative to the wrapping component */
   height?: string;
+  /** Whether the sidebar is hidden */
```
nineteen88: suggestion(non-blocking): maybe we could add some more context on exactly what it's doing

The thinking behind this suggestion is that it's technically hidden without enabling this prop, but with this prop it's only visually hidden and is not removed from the DOM.

We've added extra information on this within the Story, but it would be  better if we can self document where possible too. Reason being that it's not impossible that we will lose this story in the non-too-distant future when we alter our story approach.

src/components/adaptive-sidebar/adaptive-sidebar.test.tsx
```diff
@@ -214,3 +293,91 @@ test("should not render with a left border if the `borderColor` prop is set to `
...
+  await user.click(openButton);
+
+  expect(screen.getByTestId("adaptive-sidebar")).toHaveStyleRule(
+    "display",
+    "block",
+  );
+  expect(screen.getByText("My content")).toBeVisible();
+});
+
+test("hides the sidebar when the `hidden` prop is set to `true` and below the scren breakpoint", async () => {
```
nuria1110, line 337: **suggestion:** To avoid having to dom query, I believe `getByRole("dialog")` should be able to find the sidebar when visible. When its hidden, we could pass data-role to the Sidebar we render within AdaptiveSidebar to still be able to access the element, as it will not find it as a dialog due to `aria-modal="false"` when hidden, I've tried this locally and it looks like it works:

```
test("hides the sidebar when the `hidden` prop is set to `true` and below the scren breakpoint", async () => {
  const user = userEvent.setup();
  mockUseIsAboveBreakpoint.mockReturnValue(false);

  render(<MockWithHiddenSupport />);

  expect(screen.queryByText("My content")).not.toBeInTheDocument();

  const openButton = screen.getByTestId("adaptive-sidebar-control-button");
  await user.click(openButton);

  let sidebar = screen.getByRole("dialog");

  expect(sidebar).toHaveStyleRule("display", "block");

  const withinSidebar = within(sidebar);
  expect(withinSidebar.getByText("My content")).toBeInTheDocument();

  expect(openButton).toHaveTextContent("Close");
  const hideButton = screen.getByText("Hide");

  await user.click(hideButton);

  expect(openButton).toHaveTextContent("Show");

  sidebar = screen.getByTestId("adaptive-sidebar-modal-view");

  expect(sidebar).toHaveStyleRule("display", "none");

  expect(screen.getByText("My content")).toBeInTheDocument();
  expect(screen.getByText("My content")).not.toBeVisible();

  await user.click(openButton);

  sidebar = screen.getByRole("dialog");

  expect(sidebar).toHaveStyleRule("display", "block");
  expect(screen.getByText("My content")).toBeVisible();
});
```

nuria1110 (dismissed):

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -67,18 +70,33 @@ export const AdaptiveSidebar = ({
     }
   }, [open]);
 
+  // Remove inert attribute from elements when sidebar is hidden
+  /* istanbul ignore next */
+  useEffect(() => {
+    if (!hidden || !open) return;
```
edleeks87, line 76:
```suggestion
    if  (!(renderAsModal && hidden && open)) return;
```

thinking is do we need to run this at all if it's not a modal etc, although we may need to handle if it adapts so if omitting the modal check helps that it's fine to etc

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -67,18 +70,33 @@ export const AdaptiveSidebar = ({
     }
   }, [open]);
 
+  // Remove inert attribute from elements when sidebar is hidden
```
edleeks87, line 73: comment: this feels like it would be better handled in the manager that is adding this originally. We also aren't re-adding `inert when the modal is no longer `hidden`. Furthermore, we probably need to also update `aria-hidden` as that remains true when the `inert` attribute is removed.

image

image

image

src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
```diff
@@ -531,3 +532,98 @@ export const WithCustomBorderColor: Story = () => {
...
+              }}
+            >
+              Close
+            </Button>
+          </SplitButton>
+        </Box>
+        <Hr my={0} mx={0} />
+        <Box display="flex" flexDirection="column" p={1}>
+          <Typography>
+            This counter will update every second when the sidebar is open or
```
edleeks87, line 621: suggestion (non-blocking): I'm not sure what value this adds, would just updating the text to say `shown`/`hidden` work the same as it would still require inspecting the element in the DOM etc?

src/components/adaptive-sidebar/adaptive-sidebar.style.ts
```diff
@@ -11,20 +11,21 @@ import { getColors } from "./__internal__/utils";
...
   AdaptiveSidebarProps,
-  "backgroundColor" | "borderColor" | "height" | "width"
+  "backgroundColor" | "borderColor" | "height" | "hidden" | "width"
 > &
   MarginProps &
   PaddingProps & {
     tabIndex: number;
   };
 
 const StyledAdaptiveSidebar = styled(Box)<StyledAdaptiveSidebarProps>`
```
edleeks87, line 21: suggestion: we might not need to update the style file here if we just update `Box` to support `hidden`. It already supports `aria-hidden` so to me doesn't feel like a ridiculous idea to support both etc

src/components/adaptive-sidebar/adaptive-sidebar.stories.tsx
damienrobson-sage: The counter was designed to show that the activity carries on "behind the scenes" when the sidebar is hidden, which is the whole reason for adding this feature.

src/components/adaptive-sidebar/adaptive-sidebar.style.ts
damienrobson-sage: Not a bad shout, I'll add a `hidden` prop to `Box` and update.

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
damienrobson-sage: Spoken about this internally and these concerns will be addressed in a separate ticket (FE-7438)

nineteen88 (dismissed):

nuria1110 (dismissed):

paulrobinson88 (approved):

nuria1110 (approved):

nineteen88 (approved):

PR #7472, opened 2025-07-31
feat(badge): align with beta DS

nuria1110:
### Proposed behaviour

Align `Badge` component with the new beta DS:
- Deprecate `onClick`, `aria-label` and `color` props. 
- Make `children` optional to render `Badge` as a standalone component.
- Allow `counter` prop to accept number or string values - numbers greater than `1000` will display as `999+`, strings longer than 4 characters will be trimmed (e.g. `12345` will display `1234`). 
- Add `size` prop which accepts "small", "medium" and "large".
- Add `variant` prop which accepts "typical" and "subtle" variants. 
- Add `inverse` prop to apply inverse styles to badge. 

image

image

image

image

image

### Current behaviour

`Badge` component is not in line with the new beta DS.

image

### Checklist

#### QA

### Testing instructions

All stories should support controls for testing purposes - "Default" test story has controls to set `counterAsString` and `counterAsNumber`.

- Counter should not be displayed in small size. 
- If number or string equivalent to 0 is passed to `counter`, `Badge` will not render.
- Deprecated onClick behaviour should still be supported with new variants and inverse styles.

### DIFF
--- a/src/components/badge/badge.component.tsx
+++ b/src/components/badge/badge.component.tsx
@@ -1,56 +1,135 @@
 import React, { useState, useRef } from "react";
-import {
-  StyledBadgeWrapper,
-  StyledCrossIcon,
-  StyledCounter,
-  StyledBadge,
-} from "./badge.style";
-import { TagProps } from "../../__internal__/utils/helpers/tags";
+import { MarginProps } from "styled-system";
+import { filterStyledSystemMarginProps } from "../../style/utils";
+import { StyledBadgeWrapper, StyledCounter, StyledBadge } from "./badge.style";
+import Icon from "../icon";
+import tagComponent, {
+  TagProps,
+} from "../../__internal__/utils/helpers/tags/tags";
 import guid from "../../__internal__/utils/helpers/guid";
+import Logger from "../../__internal__/utils/logger";
 
-export interface BadgeProps extends TagProps {
-  /** Prop to specify an aria-label for the component */
+let deprecateOnClickTriggered = false;
+let deprecateAriaLabelTriggered = false;
+let deprecateColorTriggered = false;
+
+export interface BadgeProps extends TagProps, MarginProps {
+  /** @deprecated Prop to specify an aria-label for the component */
   "aria-label"?: string;
-  /** The badge will be added to this element */
-  children: React.ReactNode;
+  /** The badge will be positioned relative to this element */
+  children?: React.ReactNode;
   /** The number rendered in the badge component */
   counter?: string | number;
-  /** Prop to specify the color of the component */
+  /** @deprecated Prop to specify the color of the component */
   color?: string;
-  /** Callback fired when badge is clicked */
+  /** @deprecated Callback fired when badge is clicked */
   onClick?: (ev: React.MouseEvent<HTMLElement>) => void;
   /** Unique identifier for the component. */
   id?: string;
+  /** Size of the badge */
+  size?: "small" | "medium" | "large";
+  /** Badge variant */
+  variant?: "typical" | "subtle";
+  /** Set the style of the Badge to inverse */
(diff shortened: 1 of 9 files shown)


src/components/badge/badge.component.tsx
```diff
@@ -1,56 +1,124 @@
...
   const { current: uniqueId } = useRef(id || guid());
 
-  const renderCorrectBadge = () => {
+  const shouldDisplayBadge = counter != 0;
+  const isButton = onClick && size !== "small";
+  const isButtonInteraction = isButton && (isFocused || isHovered);
+
+  let counterToDisplay = counter;
+  if (
+    (typeof counter === "string" && counter.length > 3) ||
```
damienrobson-sage: **suggestion**: this will show "999+" if I pass the string, "0000". Might be better to parse `counter` as a number every time and then check the actual value:

```javascript
if (Number(counter) > 999) {
    counterToDisplay = "999+";
};
```

Doing this also protects against passing "abcd" as a string, as `Number()` will return `undefined` if it can't parse.

src/components/badge/badge.component.tsx
```diff
@@ -1,56 +1,124 @@
...
+    );
+    deprecateColorTriggered = true;
+  }
+
   const [isFocused, setIsFocused] = useState(false);
   const [isHovered, setIsHovered] = useState(false);
   const { current: uniqueId } = useRef(id || guid());
 
-  const renderCorrectBadge = () => {
+  const shouldDisplayBadge = counter != 0;
```
damienrobson-sage: **comment**: what would happen if I passed a string version of "0"? Would the badge show with 0 as the value, or am I missing something?

src/components/badge/badge.component.tsx
nuria1110: The way I've currently gone about it is that because `counter` accepts either a number or string, if consumers prefer a max of '99+' instead of '999+' due to e.g. limited space, they would have to set the limit and then add the '+' sign themselves and pass that as a string. Which is why I've decided to avoid parsing the value to an integer directly since this wouldn't really work if we want to allow them to provide the '+' sign.  

It just depends on how much control we leave up to consumers and how we want to limit the size of the Badge, as with current designs, they should not be displaying more than 4 characters, including the '+' sign. One thing we could do is add a mechanism for them to pass a prop e.g. `maxCounter`, where we would then internally add the '+' sign if the counter exceeded that max value, and avoid consumers having to set the limit themselves. 

I think this may be worth discussing with the team since I'm aware the way I've done this currently is probably not ideal/confusing.

src/components/badge/badge.component.tsx
nuria1110: This is loose inequality so both integer `0` and string "0" should not render the badge

damienrobson-sage: So it is - I saw `!==` but now you've pointed it out I can see that it's actually `!=` 😄

src/components/badge/badge.component.tsx
damienrobson-sage: As per Slack discussion, nothing more to do here beyond discuss with the wider consumer base etc. as to how it should be used

src/components/badge/badge.style.ts
```diff
@@ -1,107 +1,163 @@
...
+    default:
+      return css`
+        min-width: 24px;
+        height: 24px;
+        padding: 0px 2px 1px;
+      `;
+  }
+};
+
+// TODO: replace with design tokens
```
nineteen88, line 32: praise: Appreciate adding a TODO comment for this

nineteen88 (approved):
praise: great work getting this sorted

nineteen88 (approved):

PR #7476, opened 2025-08-06
[Dialog, DialogFullScreen] Merge components

damienrobson-sage:
### Proposed behaviour

Merge `Dialog` and `DialogFullScreen` into a single component. Full-screen dialogs can still be acheived through use of the `fullscreen` prop.

### Current behaviour

There is a lot of duplicated code between Dialog and DialogFullscreen, yet the only real difference is whether or not the dialog takes up the full screen or not. Both components are therefore viable candidates to be merged together into a single Dialog component, with optional flag for enabling fullscreen mode.

### Checklist

#### QA

### Additional context

`DialogFullScreen` still exists as a usable component to keep this from being a breaking change, however all it does is invoke the `Dialog` component with the `fullscreen` prop hard-coded to true.


src/components/dialog/dialog-test.stories.tsx
```diff
@@ -494,3 +496,264 @@ export const WithButton = {
     );
   },
 };
+
+export const WithStickyForm: StoryType = {
```
Parsium: **praise**: nice job re-rewriting these stories in latest CSF 3 format :+1: This should help us avoid tech debt when CSF 2 is eventually retired.

src/components/dialog/components.test-pw.tsx
```diff
@@ -682,10 +693,1039 @@ export const UsingHandle = () => {
...
+        </Box>
+      </CarbonProvider>
+    </>
+  );
+};
+
+export const FullScreenWithHeaderChildren = () => {
+  const [isOpen, setIsOpen] = React.useState(true);
+  const HeaderChildren = (
+    <Box margin="$min-width: 568px 0 26px">
```
Parsium, line 819: **question(non-blocking)**: May I clarify what this margin value achieves? This looked like a mistake at first glance.

Appreciate this was an example pre-existing in `DialogFullscreen`'s tests, so happy to leave for now.

src/components/dialog/components.test-pw.tsx
Parsium, line 1: **suggestion(non-blocking)**: To reduce the number of test examples we have, could we reuse the non-fullscreen examples where there's identical markup between them and the new fullscreen examples? For example, `FullScreenComponentFocusableSelectors` seems to be similar to `DialogComponentFocusableSelectors`.

Appreciate it's not the focus of these changes, and our Playwright examples more broadly could do with a bit of a tidy-up, so happy to if you want to defer this.

src/components/dialog/dialog-test.stories.tsx
```diff
@@ -494,3 +496,264 @@ export const WithButton = {
...
+  },
+  args: {
+    open: true,
+    title: "Example Dialog",
+    subtitle: "I have a sticky form!",
+    showCloseIcon: true,
+    onCancel: () => {},
+  },
+  decorators: [
+    (Story: StoryObj) => (
```
Parsium: **suggestion(non-blocking)**: I noticed some type errors were being flagged for the new stories where decorator parameter have been manually typed. What do you think about removing these and let typescript infer the type?


```suggestion
    (Story) => (
```

src/components/dialog/dialog.mdx
Parsium, line 1: **suggestion**: there seems to be some duplicated information between the existing `Dialog` docs and the new Fullscreen section. For example, the details on restoring focus on close is identical to its non-fullscreen version.

Could we remove some of the new information that's already been sufficiently covered previously?

src/components/dialog/dialog.pw.tsx
```diff
@@ -20,876 +21,1683 @@ import {
...
+          }}
+        />,
+      );
 
-  if (!boundingBox) throw new Error("Dialog not found or visible.");
-  if (!viewportSize) throw new Error("Unable to retrieve viewport size.");
+      const closeIcon = getDataElementByValue(page, "close").first();
+      await closeIcon.click();
+      expect(called).toBe(true);
+    });
```
Parsium: **suggestion**: what do you think about removing this test and ensure this is covered by a Jest test?

Previously we've had to remove a lot of tests from Playwright that followed this callback assertion pattern, as it was common for them to flake since the `onCancel` callback will be called asynchronously.

Here's some more info explaining why this occurs: [Link to Playwright docs](https://playwright.dev/docs/test-components#test-stories)

src/components/dialog/dialog.style.ts
```diff
@@ -25,145 +27,275 @@ const dialogSizes = {
...
 
-const StyledDialogTitle = styled.div<StyledDialogTitleProps>`
-  background-color: var(--colorsUtilityYang100);
-  padding: 23px 32px 0;
-  border-bottom: 1px solid #ccd6db;
-  border-top-right-radius: var(--borderRadius200);
-  border-top-left-radius: var(--borderRadius200);
-  ${({ showCloseIcon }) => showCloseIcon && "padding-right: 85px"};
+          ${highlightVariant &&
+          highlightVariant === "ai" &&
```
Parsium, line 236: **question**: do we want to allow consumers to use a highlight variant with the fullscreen dialog? Given its not currently possible to use one.

src/components/sidebar/sidebar.stories.tsx
```diff
@@ -506,14 +506,14 @@ export const TopModalOverride: Story = () => {
...
       </Button>
-      <DialogFullScreen
+      <Confirm
         open={isOpenDialogFullScreen && isOpenAll}
         onCancel={() => setIsOpenDialogFullScreen(false)}
-        title="Dialog fullscreen"
+        title="Confirm"
         onConfirm={() => {}}
       >
         <Textbox label="Fullscreen textbox" />
```
Parsium: **suggestion(non-blocking)**: what do you think of this just for consistency?

```suggestion
        <Textbox label="Confirm textbox" />
```

src/components/dialog/dialog.test.tsx
```diff
@@ -21,380 +26,869 @@ afterEach(() => {
...
-      <Dialog open title="Outer dialog">
+    const content = screen.getByTestId("dialog-content");
+    expect(content).toHaveStyle({
+      padding: "0px 0px 0px 0px",
+    });
+  });
+
+  /** Remove this when after Pages is re-written */
+  // TODO: in for coverage only. Prop appears not to be used - to be removed (FE-6774)
+  test.skip("applies the appropriate styles when the `pagesStyling` prop is set", () => {
```
Parsium: **comment**: I assume the commented code might have been a mistake :)

src/components/dialog/dialog.test.tsx
Parsium, line 1: **suggestion**: I noticed some of the new fullscreen dialog tests are asserting previously tested behaviour. For example, the test [asserting the dialog has the `aria-modal` attribute](https://github.com/Sage/carbon/blob/15014aee464678fb35dc1ae699f4f8f2b03eb4b8/src/components/dialog/dialog.test.tsx#L468-L476) seems identical to [this pre-existing test](https://github.com/Sage/carbon/blob/15014aee464678fb35dc1ae699f4f8f2b03eb4b8/src/components/dialog/dialog.test.tsx#L96-L103).

To simplify the test suite, would it be worth us removing some of the new tests if they aren't asserting new behaviour unique to the fullscreen dialog?

src/components/dialog/components.test-pw.tsx
damienrobson-sage: I'm not sure what's going on here, tbh. Span up the component in the test Storybook entries and this is what's rendered...:

Screenshot 2025-08-18 at 10 33 28

I'll leave it for now and maybe dig into it once every other tweak has been done.

src/components/dialog/dialog-test.stories.tsx
damienrobson-sage: Nice one, yeah my linter originally went balistic that it didn't have the types explicitly set, but I've removed them and it's happy to work it out itself, so all good 👍

src/components/dialog/dialog.style.ts
damienrobson-sage: Nice spot, removed.

src/components/dialog/dialog.test.tsx
damienrobson-sage: Yep, was left over whilst I was messing around trying to get it to pass. I've uncommented but left the test skipped for now as I didn't want to burn time trying to fix a test that would soon be removed.

src/components/dialog/dialog.pw.tsx
damienrobson-sage: Removed this as it's already covered by [this test](https://github.com/Sage/carbon/pull/7476/files#diff-f646980e207cd9a900c040057d92e3a03e28680bc0e01dfc63e7b3d7f2aa66eeR829)

src/components/dialog/dialog.test.tsx
damienrobson-sage: Removed the vast majority of the tests as they were identical or near-identical in their processes, as suggested; the remaining ones are there mainly for coverage purposes

src/components/dialog/dialog.component.tsx
```diff
@@ -190,88 +228,122 @@ export const Dialog = forwardRef<DialogHandle, DialogProps>(
...
-                  : "var(--colorsUtilityYang100)"
-              }
             >
-              {dialogTitle}
+              {title || headerChildren ? dialogTitle() : null}
               {closeIcon}
               <StyledDialogContent
                 {...contentPadding}
                 data-role="dialog-content"
+                disableContentPadding={disableContentPadding}
```
Parsium, line 337: **suggestion**: what do you think about deprecating `disableContentPadding` now that `Dialog` support both it and the `contentPadding` prop? This could help us cut down on any superfluous props in the future.

src/components/dialog/dialog.style.ts
```diff
@@ -25,145 +27,275 @@ const dialogSizes = {
...
 
-  ${({ size }) =>
-    size === "maximise"
+  ${({ disableContentPadding, fullscreen, hasHeader }) =>
+    fullscreen
       ? css`
-          height: calc(100% - var(--spacing400));
-          width: calc(100% - var(--spacing400));
+          flex: 1;
+          ${computePadding}
```
Parsium: **suggestion**: I noticed the fullscreen variant doesn't make use of `Dialog`'s `contentPadding` prop, which users might not expect.

What do you think about updating the fullscreen variant's styling to accommodate `contentPadding`?

src/components/dialog/dialog.style.ts
Parsium: I noticed the fullscreen variant doesn't support the `Dialog`'s `highlightVariant` prop, which users might not expect. Do we want both the fullscreen and non-fullscreen variants to support it?

src/components/dialog/dialog.stories.tsx
```diff
@@ -588,3 +596,1334 @@ export const HighlightVariant: Story = {
...
+          <Hr />
+          <Box p={5}>{showCorrectContent2()}</Box>
+        </Drawer>
+      </Dialog>
+    </Box>
+  );
+};
+WithComplexExample.storyName = "Fullscreen: With Complex Example";
+
+export const WithDisableContentPadding: Story = () => {
```
Parsium: **suggestion**: if you agree with deprecating `disableContentPadding`, what do you think about removing this story?

src/components/dialog/dialog.stories.tsx
```diff
@@ -588,3 +596,1334 @@ export const HighlightVariant: Story = {
...
+        </Form>
+      </Dialog>
+    </>
+  );
+};
+FullScreenWithHideableHeaderChildren.storyName =
+  "FullScreen: With Hideable Header Children";
+FullScreenWithHideableHeaderChildren.parameters = {
+  chromatic: { disableSnapshot: true },
+};
```
Parsium: **suggestion**: I assume these stories might be artifacts from a older commit, given they seem identical to a few of the newly added stories. Would we be able to remove them?

src/components/dialog/dialog-test.stories.tsx
Parsium, line 1: **suggestion**: I noticed a couple of stories that seem to be identical to ones we have currently.

What do you think of the following to reduce duplicated coverage?:

- Removing `FullscreenWithTwoDifferentNodes` in favour of `WithTwoDifferentNodes`
- Removing `WithStickyForm` in favour of `WithWrappedStickyForm`
- Removing `DialogWithLongHeaderContent` in favour of `WithLongTitle`

src/components/dialog/dialog.mdx
```diff
@@ -155,6 +154,38 @@ Setting the `highlightVariant` prop will add a highlight above the Dialog.
...
 
+## Fullscreen
+
+## Examples
+
+### Default
+
+The call-to-action element should always be focused when the full-screen `Dialog` is closed. However, in some instances it may not receive focus
+due to specific browser design choices. The below example shows how to programmatically focus the call-to-action element when the full-screen `Dialog` is closed
+to ensure behaviour is consistent across all browsers.
```
Parsium: **suggestion**: given this is identical to the non-fullscreen variant, what do you think about removing this?

```suggestion
```

src/components/dialog/dialog.style.ts
damienrobson-sage: Given there's nothing in either DS regarding the use of variants in full-screen mode, and only the new DS makes any reference to the AI variant, I'd wager that it's unlikely that we'll need to support `highlightVariant` when `fullscreen`. I'll run it by someone from the DS tomorrow and feed back here, but my initial thought is that we go with this for now and update the documentation to indicate that variants work in modal mode only.

Happy to discuss further if you think there's a valid case for supporting it now, though 😄

damienrobson-sage: Spoke to Harpal, no requirement for fullscreen modals to have the highlight so I've updated the documentation to indicate this to customers.

src/components/dialog/dialog.component.tsx
```diff
@@ -67,15 +59,50 @@ export interface DialogProps extends ModalProps, TagProps {
...
+  "data-component"?: string;
+  /* Disables auto focus functionality on child elements */
+  disableAutoFocus?: boolean;
+  /** @deprecated Determines if the Dialog can be closed */
+  disableClose?: boolean;
+  /**
+   * [Legacy] Flag to remove padding from content.
+   * @deprecated Use `contentPadding` instead.
+   */
+  disableContentPadding?: boolean;
```
Parsium, line 86: **comment**: it seems `disableContentPadding` isn't being used in the component currently. Would we be able to have it override the content's padding when the prop is provided?

src/components/dialog/dialog.style.ts
```diff
@@ -25,145 +27,255 @@ const dialogSizes = {
...
+    `}
 `;
 
-const StyledDialog = styled.div<StyledDialogProps & ContentPaddingInterface>`
-  box-shadow: var(--boxShadow300);
-  display: flex;
-  flex-direction: column;
-  position: relative;
-  border-radius: var(--borderRadius200);
+const StyledDialogContent = styled.div.attrs(applyBaseTheme)<
```
Parsium, line 123: **issue**: I noticed a small visual regression with the content's padding in fullscreen mode, where the default responsive padding isn't being applied when both `contentPadding` and `disableContentPadding` are undefined.

These were the original padding values being applied at particular breakpoints:

```css
padding: 0 16px;

@media screen and (min-width: 600px) {
  padding: 0 24px;
}
@media screen and (min-width: 960px) {
  padding: 0 32px;
}
@media screen and (min-width: 1260px) {
  padding: 0 40px;
}
```

src/components/dialog/dialog.stories.tsx
```diff
@@ -588,3 +596,699 @@ export const HighlightVariant: Story = {
...
+    </>
+  );
+};
+WithHideableHeaderChildren.storyName =
+  "Fullscreen: With Hideable Header Children";
+WithHideableHeaderChildren.parameters = {
+  chromatic: { disableSnapshot: true },
+};
+
+export const FullScreenDefault: Story = () => {
```
Parsium, line 590: **comment**: it appears we are only using this new fullscreen story in the docs page. Would we be able to remove the others just to tidy up the file?

src/components/dialog/dialog.style.ts
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Parsium: **suggestion**: could we use `PaddingProps` type to avoid using `any`? This would be better for type safety.

```suggestion
```

Parsium (approved):

mihai-albu-sage (dismissed):

Parsium (dismissed):

stephenogorman (approved):

mihai-albu-sage (approved):

PR #7479, opened 2025-08-08
feat(popover-container): various design updates

tomdavies73:
### Proposed behaviour

Adds the `type` prop with a `"inverse"` option to `Hr`

Screenshot 2025-08-08 at 15 58 31

Adds the center positon to `PopoverContainer`

Screenshot 2025-08-08 at 15 58 05

Adds the `borderRadius` prop to `PopoverContainer`

Screenshot 2025-08-12 at 10 11 48

Adds the `offset` prop to `PopoverContainer`

Screenshot 2025-08-12 at 10 12 03

Adds the `data-popover-container-button` to `renderOpenComponent`'s to prop interface

Screenshot 2025-08-12 at 10 12 30

Adds multiple border radius value support to `Box`

Adds conditional rendering for the `title` prop on `PopoverContainer`

adds a test story for the primary open button responsive example

Screenshot 2025-08-12 at 10 12 57

Screenshot 2025-08-12 at 10 13 07

Screenshot 2025-08-12 at 10 13 25

### Current behaviour

No `type` prop on `Hr`
No ability to center the `PopoverContainer`
No ability to set a border radius on `PopoverContainer`
No ability to add an offset to `PopoverContainer`
An empty title container is rendered even when `title` is not passed
Only one border radius value can be used on `Box`
`data-popover-container-button` is not currently present on the renderOpenComponent's prop interface

### Checklist

#### QA

### Testing instructions

Each prop added can be tested in isolation on each relevant components' story book docs. The full implementation can be seen on the new test story that has been added.

### DIFF
--- a/src/components/box/box.component.tsx
+++ b/src/components/box/box.component.tsx
@@ -24,10 +24,13 @@ export type BoxSizing = "content-box" | "border-box";
 
 type DesignTokensType = keyof typeof DesignTokens;
 type BoxShadowsType = Extract<DesignTokensType, `boxShadow${string}`>;
-export type BorderRadiusType = Extract<
-  DesignTokensType,
-  `borderRadius${string}`
->;
+type BorderRadiusToken = Extract<DesignTokensType, `borderRadius${string}`>;
+
+export type BorderRadiusType =
+  | BorderRadiusToken
+  | `${BorderRadiusToken} ${BorderRadiusToken}`
+  | `${BorderRadiusToken} ${BorderRadiusToken} ${BorderRadiusToken}`
+  | `${BorderRadiusToken} ${BorderRadiusToken} ${BorderRadiusToken} ${BorderRadiusToken}`;
 
 export interface BoxProps
   extends FlexboxProps,
@@ -152,7 +155,7 @@ export const Box = React.forwardRef<HTMLDivElement, BoxProps>(
         bg={bg}
         backgroundColor={backgroundColor}
         boxShadow={boxShadow}
-        borderRadius={borderRadius}
+        $borderRadius={borderRadius}
         aria-hidden={ariaHidden}
         hidden={hidden}
         {...tagComponent(dataComponent, rest)}
--- a/src/components/box/box.style.ts
+++ b/src/components/box/box.style.ts
@@ -24,13 +24,14 @@ const calculatePosition = (props: Omit<PositionProps, "zIndex">) => {
 };
 
 const StyledBox = styled.div.attrs(applyBaseTheme)<
-  BoxProps & {
+  Omit<BoxProps, "borderRadius"> & {
     cssProps?: {
       color?: string;
       opacity?: string;
       height?: string;
       width?: string;
     };
+    $borderRadius?: BoxProps["borderRadius"];
   }
 >`
   ${space}
@@ -39,9 +40,14 @@ const StyledBox = styled.div.attrs(applyBaseTheme)<
   ${grid}
   ${calculatePosition}
 
-  ${({ borderRadius = "borderRadius000" }) => css`
-    border-radius: var(--${borderRadius});
-  `}
+${({ $borderRadius = "borderRadius000" }) => {
+    const radiusValues = $borderRadius.split(" ");
+    return css`
+      border-radius: ${radiusValues
+        .map((radius) => `var(--${radius})`)
+        .join(" ")};
+    `;
+  }}
 
   ${({ cssProps, bg, backgroundColor, ...rest }) =>
     styledColor({ color: cssProps?.color, bg, backgroundColor, ...rest })}
--- a/src/components/hr/hr.component.tsx
+++ b/src/components/hr/hr.component.tsx
@@ -13,6 +13,8 @@ export interface HrProps extends MarginProps, TagProps {
   adaptiveMxBreakpoint?: number;
   /** Set the height of the component. Accepts one of "small", "medium", or "large" */
   height?: "small" | "medium" | "large";
+  /** Set the color variant of the horizontal rule. Use "typical" for standard styling or "inverse" for use in darker backgrounds */
+  type?: "typical" | "inverse";
 }
 
 export const Hr = ({
@@ -20,6 +22,7 @@ export const Hr = ({
   ml,
   mr,
   "aria-hidden": ariaHidden,
+  type = "typical",
   height = "small",
   ...rest
 }: HrProps): JSX.Element => {
@@ -35,6 +38,7 @@ export const Hr = ({
     <StyledHr
       aria-hidden={ariaHidden}
       height={height}
+      $type={type}
       ml={marginLeft}
       mr={marginRight}
       mt={rest.mt || 3}
--- a/src/components/hr/hr.style.ts
+++ b/src/components/hr/hr.style.ts
@@ -9,13 +9,18 @@ const heightMap = {
 };
 
 const StyledHr = styled.hr.attrs(applyBaseTheme)<
-  MarginProps & { height: "small" | "medium" | "large" }
+  MarginProps & { height: "small" | "medium" | "large" } & {
+    $type?: "typical" | "inverse";
+  }
 >`
   ${margin}
   width: 100%;
   border: 0;
   height: ${({ height }) => heightMap[height]}px;
-  background: var(--colorsUtilityMajor100);
+  background-color: ${({ $type }) =>
+    $type === "typical"
+      ? "var(--colorsUtilityMajor100)"
+      : "var(--colorsActionMajorYang030)"};
 `;
 
 export default StyledHr;
--- a/src/components/link/link.style.ts
+++ b/src/components/link/link.style.ts
@@ -246,6 +246,11 @@ const StyledLink = styled.span.attrs(applyBaseTheme)<
         box-shadow: 0 var(--spacing050) 0 0 var(--colorsUtilityYin090);
         border-bottom-left-radius: var(--borderRadius025);
         border-bottom-right-radius: var(--borderRadius025);
+
+        &:has([data-popover-container-button="true"]) {
+          border-bottom-left-radius: var(--borderRadius000);
+          border-bottom-right-radius: var(--borderRadius000);
+        }
       `}
 
       > button, ${StyledButton}:not(.search-button) {
--- a/src/components/popover-container/popover-container-test.stories.tsx
+++ b/src/components/popover-container/popover-container-test.stories.tsx
@@ -238,6 +238,67 @@ export const InsideMenuWithOpenButton = () => {
   );
 };
 
+export const InsideMenuWithPrimaryOpenButton = () => {
+  const [open, setOpen] = useState(false);
+  return (
+    <Menu menuType="black">
+      <MenuItem href="#">Menu Item One</MenuItem>
+      <MenuItem onClick={() => {}} submenu="Menu Item Two">
+        <MenuItem href="#">Submenu Item One</MenuItem>
+        <MenuItem href="#">Submenu Item Two</MenuItem>
+      </MenuItem>
+      <MenuItem submenu="Search">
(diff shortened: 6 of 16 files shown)


src/components/hr/hr.component.tsx
```diff
@@ -13,13 +13,16 @@ export interface HrProps extends MarginProps, TagProps {
   adaptiveMxBreakpoint?: number;
   /** Set the height of the component. Accepts one of "small", "medium", or "large" */
   height?: "small" | "medium" | "large";
+  /** Set the color variant of the horizontal rule. Use "typical" for standard styling or "inverse" for use in darker backgrounds */
+  variant?: "typical" | "inverse";
```
edleeks87: Comment: I think we should go with `type` here, that way it continues to be easily updated to add new options but if we then need to have variations within the types we can add `variant` etc

src/components/hr/hr.component.tsx
```diff
@@ -35,6 +38,7 @@ export const Hr = ({
     <StyledHr
       aria-hidden={ariaHidden}
       height={height}
+      variant={variant}
```
edleeks87: suggestion (non-blocking): whilst it's not an issue with v5 of SC and the rest of the props aren't like this I think it might be worth using the `$` prefix. In v6 anything not prefixed is automatically passed down to the element so if we every migrate we'd start hitting errors around invalid attributes etc

src/components/hr/hr.style.ts
```diff
@@ -9,13 +9,18 @@ const heightMap = {
 };
 
 const StyledHr = styled.hr.attrs(applyBaseTheme)<
-  MarginProps & { height: "small" | "medium" | "large" }
+  MarginProps & { height: "small" | "medium" | "large" } & {
+    variant?: "typical" | "inverse";
```
edleeks87: `$type?: "typical" | "inverse";` if you decide to go with the above

src/components/popover-container/popover-container.component.tsx
```diff
@@ -396,6 +409,7 @@ export const PopoverContainer = forwardRef<
         aria-label={containerAriaLabel}
         aria-describedby={ariaDescribedBy}
         p="16px 24px"
+        borderRadius={borderRadius}
```
edleeks87: suggestion (non-blocking): borderRadius -> $borderRadius

src/components/popover-container/popover-container.stories.tsx
```diff
@@ -144,6 +144,26 @@ export const BorderRadius: Story = () => {
 };
 BorderRadius.storyName = "Border Radius";
 
+export const Offset: Story = () => {
+  const [open, setOpen] = useState(defaultOpenState);
+  const onOpen = () => setOpen(true);
+  const onClose = () => setOpen(false);
+  return (
+    <div style={{ height: 100 }}>
```
edleeks87: suggestion: might be worth using Box here

src/components/box/box.style.ts
```diff
@@ -39,9 +39,14 @@ const StyledBox = styled.div.attrs(applyBaseTheme)<
   ${grid}
   ${calculatePosition}
 
-  ${({ borderRadius = "borderRadius000" }) => css`
-    border-radius: var(--${borderRadius});
-  `}
+${({ borderRadius = "borderRadius000" }) => {
+    const radiusValues = borderRadius.split(" ");
```
edleeks87: suggestion (non-blocking): as we've done this a couple of times we could move it to a shared util

src/components/box/box.component.tsx
```diff
@@ -24,10 +24,13 @@ export type BoxSizing = "content-box" | "border-box";
 
 type DesignTokensType = keyof typeof DesignTokens;
 type BoxShadowsType = Extract<DesignTokensType, `boxShadow${string}`>;
-export type BorderRadiusType = Extract<
-  DesignTokensType,
-  `borderRadius${string}`
->;
+type BorderRadiusToken = Extract<DesignTokensType, `borderRadius${string}`>;
+
+export type BorderRadiusType =
```
edleeks87, line 29: suggestion: this type could be shared between this and PopoverContainer to ensure consistency

src/components/popover-container/popover-container.component.tsx
```diff
@@ -376,6 +377,7 @@ export const PopoverContainer = forwardRef<
       ref: openButtonRef,
       "aria-label": openButtonAriaLabel || title,
       id: isOpen ? undefined : popoverContainerId,
+      className: "persist-primary-styling",
```
edleeks87: comment (non-blocking): as we use `primary` as a `buttonType`/`variant` this className might suggest something unintended. Maybe something like `persist-original-styling`

src/components/popover-container/popover-container.component.tsx
```diff
@@ -126,9 +139,13 @@ export interface PopoverContainerProps extends PaddingProps, TagProps {
...
   children?: React.ReactNode;
   /** Sets rendering position of dialog */
-  position?: "left" | "right";
+  position?: Position;
+  /** The popover offset from the reference element */
+  offset?: number;
   /** Sets the popover container dialog header name */
   title?: string;
+  /** Sets the border radius of the popover container */
+  borderRadius?: BorderRadiusType;
```
Parsium: **suggestion(non-blocking)**: to address @edleeks87's point on de-duplicating the type definitions, what do you think about indexing `BoxProps` like so?:

```tsx
import type { BoxProps } from "../box";

export interface PopoverContainerProps extends PaddingProps, TagProps {
  borderRadius?: BoxProps["borderRadius"];
}
```

src/components/popover-container/popover-container.test.tsx
```diff
@@ -154,6 +154,24 @@ test("popup has correct accessible description when ariaDescribedBy prop is pass
...
+
+test("popover title container renders when the title prop is passed", () => {
+  render(
+    <PopoverContainer title="Custom Title" open>
+      Ta da!
+    </PopoverContainer>,
+  );
+
+  expect(screen.getByTestId("popover-container-title")).toBeVisible();
+});
```
Parsium: **question(non-blocking)**: since we already test whether the title is rendered when the `title` prop is passed, do we need these two tests?

src/components/popover-container/popover-container-test.stories.tsx
```diff
@@ -238,6 +241,207 @@ export const InsideMenuWithOpenButton = () => {
...
+          onOpen={() => setOpen(true)}
+          onClose={() => setOpen(false)}
+          open={open}
+          renderOpenComponent={({ ref, onClick, className }) => (
+            <Box data-role="gblnav-notificationui-bell">
+              <Button
+                aria-label="Notifications"
+                ref={ref}
+                onClick={onClick}
+                className={className}
```
Parsium: **issue**: currently we mark `Button`'s `className` prop as private, so this could lead to confusion if we start suggesting consumers use it. However, I don't think we should make the prop public, since it would provide a back entrance for consumers to freely customise `Button`'s styling.

I appreciate given how `MenuItem` overrides the `Button`'s background colour, this complicates how we can fix this. Happy to have a chat with you and @edleeks87 about ways we can work around this.

src/components/popover-container/popover-container.component.tsx
```diff
@@ -165,19 +182,22 @@ export type PopoverContainerHandle = {
...
       offset(
         shouldCoverButton
           ? ({ rects }) => ({ mainAxis: -rects.reference.height })
-          : 6,
+          : popoverOffset,
       ),
       flip({
         fallbackStrategy: "initialPlacement",
       }),
     ],
```
Parsium: **suggestion**: I noticed at small screen widths, the popover doesn't shift position in order to keep it fully in view, which could be a problem whenever the open button isn't fully centred on the page. Here's an example at 320px:

Screenshot 2025-08-13 at 09 55 52

To fix this, could we utilise [the `shift` middleware](https://floating-ui.com/docs/shift) from `@floating-ui`?:

```suggestion
      }),
      shift()
    ],
```

tomdavies73: Great suggestion mate, for the meantime I will add the shift middleware whilst using the `center` position

src/components/popover-container/popover-container-test.stories.tsx
tomdavies73: I agree `className` is probably not the best approach here, I will pass and use a data attribute for styling instead

src/components/popover-container/popover-container.component.tsx
tomdavies73: Yeah, went for an attribute `data-popover-container-button`

src/components/box/box.style.ts
tomdavies73: As it's only used twice I've decided not to move it to a shared util, but if used again in the codebase I believe it should be

src/components/hr/hr.component.tsx
tomdavies73: Yeah agreed, I've used the `$` prefix approach now across the PR when passing props down to styled components

src/components/hr/hr.component.tsx
tomdavies73: Yeah I agree, pivoted from `variant` to `type` for this scenario

src/components/popover-container/popover-container-test.stories.tsx
```diff
@@ -238,6 +241,215 @@ export const InsideMenuWithOpenButton = () => {
...
+    <GlobalHeader logo={Logo()}>
+      <Menu menuType="black" flex="1">
+        <MenuItem flex="1" submenu="Product Switcher">
+          <MenuItem href="#">Product A</MenuItem>
+        </MenuItem>
+        <MenuItem href="#" flex="0 0 auto">
+          <PopoverContainer
+            containerAriaLabel="Create"
+            closeButtonAriaLabel="Close Create Popover"
+            position={isSmall ? "center" : !isMid ? "center" : "right"}
```
Parsium: **suggestion**: The popup seems to have a tendency to jump in its positioning when adjusting the screen width. I assume its a combination of `Popover` re-rendering to calculate the position, plus our use of `useMediaQuery` here which is also triggering re-renders.


https://github.com/user-attachments/assets/c58a757c-d33d-4bc1-9738-2711f861600a



What do you think about simplifying the positioning like so? This should make any position changes a bit smoother visually:

```suggestion
            position="center"
```

src/components/link/link.style.ts
```diff
@@ -244,8 +244,11 @@ const StyledLink = styled.span.attrs(applyBaseTheme)<
         }
         max-width: fit-content;
         box-shadow: 0 var(--spacing050) 0 0 var(--colorsUtilityYin090);
-        border-bottom-left-radius: var(--borderRadius025);
-        border-bottom-right-radius: var(--borderRadius025);
+
+        &:not(:has(.persist-primary-styling)) {
+          border-bottom-left-radius: var(--borderRadius025);
+          border-bottom-right-radius: var(--borderRadius025);
+        }
```
Parsium: **suggestion**: It seems the border radii set above on the underlying anchor/button is sufficient. What do you think about removing this?:


```suggestion
```

src/components/popover-container/popover-container-test.stories.tsx
edleeks87: I understand the thinking behind it we might need to do some sort of throttling so that it doesn't try to instantly move as James points out it creates some jumping around.

On the ternary side of things something like `position={isSmall || !isMid ? "center" : "right"}`

Parsium (dismissed):

edleeks87 (dismissed):

edleeks87 (approved):

harpalsingh:
Just adding aware the ability to control scrollbar color is something to add later and defaults will render.

paulrobinson88 (approved):

edleeks87 (approved):

PR #7492, opened 2025-08-18
fix(responsive-vertical-menu): ensure secondary menu is a child of primary

nuria1110:
### Proposed behaviour

- Moves secondary menu element to be a child of the primary menu instead of a sibling to ensure correct screen reader behaviour. 

https://github.com/user-attachments/assets/16964dbf-a351-4551-890d-3b684526f109

- Children of last tertiary item in secondary menu are accessible via Tab.

https://github.com/user-attachments/assets/119d70fb-7a27-40c4-bff3-f24f88652dea

### Current behaviour

- Secondary menu element is a sibling of the primary menu, meaning screen readers will navigate through all of the items within the primary menu before moving onto the secondary.

https://github.com/user-attachments/assets/6bc6434d-91a9-4210-8335-fb0264610cf2

- Children of last tertiary item in secondary menu are not accessible via Tab.

https://github.com/user-attachments/assets/6ab6924e-6859-46b9-a062-45dcaf28f3fe

### Checklist

#### QA

### Additional context

Additional ticket (FE-7448) has been raised to fix other keyboard nav issues regarding level 4 menu items, playwright tests have been written for these but skipped.

### Testing instructions

See [demo](https://stackblitz.com/edit/parsium-carbon-starter-avlyhfm7?file=src%2FApp.tsx) for Tabbing issue, see `WithDifferentDepthAsLastItem` test story for fix.

### DIFF
--- a/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-item/responsive-vertical-menu-item.component.tsx
+++ b/src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-item/responsive-vertical-menu-item.component.tsx
@@ -10,7 +10,7 @@ import React, {
   useState,
 } from "react";
 import { MarginProps, PaddingProps } from "styled-system";
-
+import { StyledResponsiveMenu } from "../responsive-vertical-menu.style";
 import {
   StyledIcon,
   StyledMenuItemContent,
@@ -131,6 +131,10 @@ const BaseItem = forwardRef<HTMLElement, BaseItemProps>(
       setActive,
       setActiveMenuItem,
       reducedMotion,
+      left,
+      top,
+      width,
+      height,
     } = useResponsiveVerticalMenu();
 
     const {
@@ -145,6 +149,7 @@ const BaseItem = forwardRef<HTMLElement, BaseItemProps>(
     const isActive = activeMenuItem?.id === id;
     const hasChildren = React.Children.count(children) > 0;
     const parentId = useContext(ParentItemContext);
+    const subMenuRef = useRef<HTMLUListElement>(null);
     const responsiveDisplay = useMemo(
       () => responsiveMode && depth === 0,
       [depth, responsiveMode],
@@ -176,16 +181,17 @@ const BaseItem = forwardRef<HTMLElement, BaseItemProps>(
 
           // Primary-level menu items
           if (depth === 0 && hasChildren && activeMenuItem?.id === id) {
-            e.preventDefault();
-            // If the item is active and has children, move focus to the first/last child, based on whether the Shift key is pressed or not
-            moveFocus(!e.shiftKey ? "firstChild" : "lastChild");
+            // If item is active and has children, move focus to last child if Shift is pressed
+            if (e.shiftKey) {
+              e.preventDefault();
+              moveFocus("lastChild");
+            }
           }
 
           // Secondary-level menu items
           if (depth === 1) {
             const currentIndex = siblings.findIndex((el) => el.id === id);
             const isLastSibling = currentIndex === siblings.length - 1;
-            const isFirstSibling = currentIndex === 0;
 
             // If the item is the last sibling and has no children (or it does but isn't expanded), move focus to the parent when forward-tabbing
             /* istanbul ignore else */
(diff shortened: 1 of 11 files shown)


src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.context.tsx
```diff
@@ -65,6 +73,12 @@ export const ResponsiveVerticalMenuProvider = ({
   const [responsiveMode, setResponsiveMode] = useState<boolean>(false);
   const [reducedMotion, setReducedMotion] = useState<boolean>(false);
 
+  const [left, setLeft] = useState("auto");
+  const [top, setTop] = useState("auto");
+
+  const [width, setWidth] = useState<string | undefined>(undefined);
```
edleeks87: Comment: I don't think we need state here, we can just pass the props down via the context as it never gets updated internally etc

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
```diff
@@ -103,6 +105,11 @@ const BaseMenu = ({
     }
   }, [active, menu, responsiveMode, activeMenuItem, button]);
 
+  useEffect(() => {
```
edleeks87: comment: this can be removed and just pass the `height` and `width` directly as it's updated via props

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-item/responsive-vertical-menu-item.component.tsx
```diff
@@ -195,21 +201,17 @@ const BaseItem = forwardRef<HTMLElement, BaseItemProps>(
...
-            }
           }
 
           // Tertiary-level menu items
           if (depth === 2) {
             const currentIndex = siblings.findIndex((el) => el.id === id);
             const isLastSibling = currentIndex === siblings.length - 1;
 
+            // If the item is the last sibling and has no children (or it does but isn't expanded),
+            // and parent is the last sibling, move focus to the grandparent when forward-tabbing
```
Parsium: **suggestion(non-blocking)**: what do you think of the following description tweak?:

```suggestion
            // and parent is the last sibling in the level above, move focus to the grandparent when forward-tabbing
```

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.context.tsx
```diff
@@ -50,11 +57,27 @@ export const useResponsiveVerticalMenu = () => {
...
+  if (!deprecatedWarning && !isInternal) {
+    deprecatedWarning = true;
+    Logger.deprecate(
+      "`ResponsiveVerticalMenuProvider` is deprecated and no longer needed for `ResponsiveVerticalMenu`. You can use `ResponsiveVerticalMenu` directly without wrapping it in a provider.",
+    );
+  }
+
   const [active, setActive] = useState<boolean>(false);
   const [activeMenuItem, setActiveMenuItem] =
     useState<ResponsiveVerticalMenuButtonItem | null>(null);
```
Parsium: **comment**: this will be a small breaking change, since the internal context will override the external context for the underlying components. This means the components won't be able to access the external provider's props.

To fix this, we could check in the provider if an external context exists, then use its props if it does. What do you think?:

```tsx
export const ResponsiveVerticalMenuProvider = ({
  children,
  width,
  height,
}: ResponsiveVerticalMenuProviderProps) => {
  const externalContext = useContext(ResponsiveVerticalMenuContext);
  const hasExternalContext = externalContext !== null;

  if (!deprecatedWarning && hasExternalContext) {
    deprecatedWarning = true;
    Logger.deprecate(
      "`ResponsiveVerticalMenuProvider` is deprecated and no longer needed for `ResponsiveVerticalMenu`. You can use `ResponsiveVerticalMenu` directly without wrapping it in a provider.",
    );
  }

  // other logic...

  return (
    <ResponsiveVerticalMenuContext.Provider
      value={{
        // context props...
        width: width || externalContext.width,
        height: height || externalContext.height
      }}
      >
  );
```

nuria1110: As per slack conversation, since consumers should be setting the width/height props directly into the component, these new props in the provider are for internal use only. I have marked them as private and once we move the context into an internal folder this can be removed.

paulrobinson88 (approved):

PR #7522, opened 2025-09-09
test(dialog): add play functions

DobroTora:
### Proposed behaviour
Adding play functions to the Dialog component.

### Current behaviour
There are no play functions in this component currently. 

### Checklist

#### QA

### DIFF
--- a/src/components/dialog/dialog.interactions.stories.tsx
+++ b/src/components/dialog/dialog.interactions.stories.tsx
@@ -1,0 +1,68 @@
+import React, { useState } from "react";
+import { StoryObj } from "@storybook/react";
+import { userEvent, within, expect, waitFor } from "@storybook/test";
+import Dialog from ".";
+import Button from "../button";
+import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
+
+type Story = StoryObj<typeof Dialog>;
+
+export default {
+  title: "Dialog/Interactions",
+  component: Dialog,
+  parameters: {
+    layout: "padded",
+    chromatic: { disableSnapshot: true },
+  },
+  decorators: [],
+};
+
+export const FocusManagement: Story = {
+  render: function FocusManagementRender() {
+    const [isOpen, setIsOpen] = useState(false);
+    return (
+      <>
+        <Button onClick={() => setIsOpen(true)} data-role="open-dialog-btn">
+          Open Dialog
+        </Button>
+        <Dialog
+          open={isOpen}
+          title="Title"
+          subtitle="Subtitle"
+          onCancel={() => setIsOpen(false)}
+          showCloseIcon
+        >
+          <p>Content</p>
+        </Dialog>
+      </>
+    );
+  },
+
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+    const canvas = within(canvasElement);
+    const portal = within(canvasElement.ownerDocument.body);
+
+    const openButton = canvas.getByRole("button", { name: /open dialog/i });
+
+    openButton.focus();
+    await waitFor(() => expect(openButton).toHaveFocus());
+
+    await userEvent.click(openButton);
+
+    const closeButton = await portal.findByRole("button", { name: /close/i });
+    await waitFor(() => expect(closeButton).toBeVisible());
+
+    await userEvent.tab();
+    await new Promise((resolve) => setTimeout(resolve, 200));
+
+    closeButton.focus();
+  },
+};
+
+FocusManagement.storyName = "Focus Management";
+FocusManagement.parameters = {
+  themeProvider: {
+    chromatic: { theme: "sage" },
+  },
+};


src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+type Story = StoryObj<typeof Dialog>;
+
+export default {
+  title: "Dialog/Interactions",
+  component: Dialog,
+  parameters: { layout: "padded" },
+  decorators: [],
+};
+
+export const OpenAndCloseDialog: Story = {
```
edleeks87: comment: as this component's open state can be controlled by a prop we don't need this test

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    await userEvent.tab();
+    await userInteractionPause(50);
+
+    const closeIcon = within(dialog).getByRole("button", { name: /close/i });
+    await userEvent.click(closeIcon);
+    await userInteractionPause(300);
+  },
+};
+
+export const RestoreFocus: Story = {
```
edleeks87: comment: I'm not sure we need this one we have unit and playwright tests that assert this behaviour

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const cancelBtn = within(dialog).getByRole("button", { name: /^cancel$/i });
+    await userEvent.click(cancelBtn);
+    await userInteractionPause(300);
+  },
+};
+
+export const FocusManagement: Story = {
+  render: function FocusManagementRender() {
+    const [isOpen, setIsOpen] = useState(false);
+    const buttonRef = useRef<HTMLButtonElement>(null);
```
edleeks87: comment: this ref doesn't seem to do anything

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    );
+  },
+
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+
+    const canvas = within(canvasElement);
+    const portal = within(canvasElement.ownerDocument.body);
+
+    const candidates = canvas.getAllByRole("button", { name: /open dialog/i });
```
edleeks87:
```suggestion
    const openButton = canvas.getByRole("button", { name: /open dialog/i });
```

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const openButton =
+      candidates.find(
+        (el) => el.getAttribute("data-role") === "open-dialog-btn",
+      ) ?? candidates[0];
+
+    await userEvent.click(openButton);
+    await userInteractionPause(300);
```
edleeks87: comment: I believe @Parsium discovered that we should not be using interaction pauses here. The preferred approach is to use actual assertions like so:

```
export const Foo = {
  play: async ({ canvas, userEvent }) => {
    await userEvent.click(canvas.getByRole('button', { name: 'Open dialog' }));
 
    const dialog = canvas.getByRole('dialog');

    // ❌ avoid fixed delay
    // await userInteractionPause(250); 

    // 👇 Explicitly waits for dialog to appear. Ensures dialog is visible before snapshot is taken. 
    await expect(dialog).toBeVisible();
  },
};

File: src/components/dialog/dialog.interactions.stories.tsx [hunk 6]
```diff
@@ -0,0 +1,232 @@
...
+
+    const dialog = portal.getByRole("dialog");
+
+    await userEvent.tab();
+    await userInteractionPause(50);
```
edleeks87: comment: we should expect the relevant element is focused here

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const firstInput = within(dialog).getByLabelText(/first name/i);
+    await userEvent.type(firstInput, "Jane Doe", { delay: 30 });
+    await userInteractionPause(100);
+
+    await userEvent.tab();
+    await userInteractionPause(50);
+    await userEvent.tab();
+    await userInteractionPause(50);
+
+    const closeIcon = within(dialog).getByRole("button", { name: /close/i });
```
edleeks87: comment: if we click the close icon the snapshot will just be the open button. There's limited value in interactions with Dialog as we can control the open state but capturing the focus state is worth it imo

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+  render: function OpenAndCloseDialogRender() {
+    const [open, setOpen] = useState(false);
+    const triggerRef = useRef<HTMLButtonElement>(null);
+
+    return (
+      <>
+        <Button
+          data-role="dialog-trigger"
+          onClick={() => setOpen(true)}
+          ref={triggerRef}
```
damienrobson-sage: **comment**: this `ref` can be removed as it provides no purpose

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    if (!allowInteractions()) return;
+
+    const canvas = within(canvasElement);
+    const portal = within(canvasElement.ownerDocument.body);
+
+    const candidates = canvas.getAllByRole("button", { name: /open dialog/i });
+    const openButton =
+      candidates.find(
+        (el) => el.getAttribute("data-role") === "dialog-trigger",
+      ) ?? candidates[0];
```
damienrobson-sage: **comment**: rather than looking for all button elements with a specific name and then looping over them, just look up by name:

```js
canvas.getByRole("button");
```

As you've only got one visible button at the start of the story's interaction, you don't need to search for it; just get it.

src/components/dialog/dialog.interactions.stories.tsx
DobroTora: I am using this method across my play fucntion stories to capture behavior we discussed in a past to keep some UI elements open for a accurate chromatic snapshots, so I agree with @Damo to use userInteractionPause, but with shorter waiting time.

src/components/dialog/dialog.interactions.stories.tsx
DobroTora: It does restore focus to the button after dialog is dismissed.

src/components/dialog/dialog.interactions.stories.tsx
edleeks87: That's not what Damo means, we shouldn't be using these they rely on node timeouts we are testing in browser

src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```diff
@@ -0,0 +1,80 @@
+// src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```
edleeks87: Comment: this comment is a bit odd 🤔.

src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+const RenderFromDefault: StoryFn = () =>
+  (DefaultStory as unknown as () => JSX.Element)();
+
+export const NavFocusManagement: Story = {
+  render: RenderFromDefault,
+  play: async () => {
+    if (!allowInteractions()) return;
+
+    await userEvent.tab();
+    await userInteractionPause(50);
```
edleeks87: comment: we should avoid using the interactive pause, it uses a node timeout whilst we're testing in the browser. We should be using actual await assertions here instead (eg toHaveFocus)

src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const ScrollAndFocusInput: Story = {
+  render: RenderFromDefault,
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+
+    const canvas = within(canvasElement);
+    const links = canvas.getAllByRole("link");
+
+    const targetLink = links[1] ?? links[0];
```
edleeks87: comment: using nullish coalescence here is a bit odd, we know how many links there are. Will it ever hit the second part of the conditional?

src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    await userEvent.click(targetLink);
+    await userInteractionPause(300);
```
edleeks87: comment: we should assert a given element is visible here

src/components/anchor-navigation/anchor-navigation-interactions.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    await userEvent.tab();
+    await userInteractionPause(100);
```
edleeks87: comment: asserting an element is focused is better here

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+    await userEvent.tab();
+    await userInteractionPause(50);
+    await userEvent.tab();
+    await userInteractionPause(50);
+
+    const closeIcon = within(dialog).getByRole("button", { name: /close/i });
+    await userEvent.click(closeIcon);
+    await userInteractionPause(150);
+
+    expect(openButton).toHaveFocus();
```
edleeks87: comment: I'm confused about the relevance of this change, we shouldn't be removing await here as we need to account for any latency etc

edleeks87 (changes requested):
I'm confused why some AnchorNav changes are in this PR but I've assumed it was an accident and that it was meant for a separate one so have added comments to these files to help etc

DobroTora:
> > I agree with Damo when it comes to using pauses when we want to simulate user's behavior and I think it came out also in one of the previous reviews.
> 
> Apologies Debra, looking back I think I've badly-worded my original comment (or ended it without making any actual point). What I meant was that, in 90% of cases, a simple `await expect` callback will suffice. In these tests, we don't need to simulate users waiting for anything in particular. Maybe if we were simulating an API call or something of that ilk, but here we're just waiting for animations to complete and the DOM to stabilise, so we can just use the de-facto methods.

Ok, I think I misunderstood the idea as well then, as I remember we discuss it before and agree on using this method over toBeVisible. I am following the comments now and took a note of the pattern preference for a future.

src/components/dialog/dialog.interactions.stories.tsx
damo: Please don't "agree with @damo". I have nothing to do with this project and can do without the spam. Thank you.

src/components/dialog/dialog.interactions.stories.tsx
edleeks87: I think closing this will mean the dialog isn't snapshotted, it would probably be better to tab to the close icon then `await expect(within(dialog).getByRole("button", { name: /close/i })).toBeVisible()`

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+    await userEvent.type(firstInput, "Jane Doe", { delay: 30 });
+
+    const closeButton = within(dialog).getByRole("button", { name: /close/i });
+    await waitFor(() => expect(closeButton).toBeVisible());
+
+    await waitFor(async () => {
+      await userEvent.tab();
+      expect(closeButton).toHaveFocus();
+    });
+  },
```
damienrobson-sage, line 60: **comment**: this can be simplified - all we want here is the close button to be focused and snapshotted. The rest of the form is just decoration.

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+type Story = StoryObj<typeof Dialog>;
+
+export default {
+  title: "Dialog/Interactions",
+  component: Dialog,
+  parameters: { layout: "padded" },
+  decorators: [],
+} as const;
+
+export const FocusManagement: Story = {
```
edleeks87, line 20: comment: this will fail as there's more than one input etc in the rendered story but you are asserting that pressing tab once after the first input will focus the close icon when it will actually focus the next input.

IMO I think you should remove all but one input and then assert that tabbing to the close icon means it is focused, this will show the focus trap working and allow us to capture the focus state on the close icon. Something like below:


```
export const FocusManagement: Story = {
  render: function FocusManagementRender() {
    const [isOpen, setIsOpen] = useState(false);

    return (
      <>
        <Button onClick={() => setIsOpen(true)} data-role="open-dialog-btn">
          Open Dialog
        </Button>

        <Dialog
          open={isOpen}
          title="Title"
          subtitle="Subtitle"
          onCancel={() => setIsOpen(false)}
          showCloseIcon
        >
          <Form
            stickyFooter
            leftSideButtons={
              <Button onClick={() => setIsOpen(false)}>Cancel</Button>
            }
            saveButton={
              <Button buttonType="primary" type="submit">
                Save
              </Button>
            }
          >
            <Textbox label="Name" value="" onChange={() => {}} />
          </Form>
        </Dialog>
      </>
    );
  },

    if (!allowInteractions()) return;

    const portal = within(canvasElement.ownerDocument.body);

    const openButton = canvas.getByRole("button", { name: /open dialog/i });
    await userEvent.click(openButton);

    const dialog = await portal.findByRole("dialog");
    const closeButton = within(dialog).getByRole("button", { name: /close/i });
    const firstInput = within(dialog).getByLabelText(/name/i);

    await expect(dialog).toBeVisible();

    await userEvent.tab();
    await userEvent.tab();

    await expect(firstInput).toHaveFocus();

    await userEvent.tab();
    await userEvent.tab();

    await expect(closeButton).toHaveFocus();
  },
};

File: src/components/dialog/dialog.interactions.stories.tsx [hunk 18]
```diff
@@ -0,0 +1,97 @@
...
+import Form from "../form";
+
+type Story = StoryObj<typeof Dialog>;
+
+export default {
+  title: "Dialog/Interactions",
+  component: Dialog,
+  parameters: { layout: "padded" },
+  decorators: [],
+} as const;
```
edleeks87:
```suggestion
};
```

we don't need to coerce here

src/components/dialog/dialog.interactions.stories.tsx
edleeks87: we don't need to have `waitFor` around the `expect`, `await` is sufficient

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </>
+    );
+  },
+
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+
+    const portal = within(canvasElement.ownerDocument.body);
+
+    const openButton = canvasElement.querySelector(
```
edleeks87: question: why was this refactored from the getByRole?

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    await userEvent.click(openButton);
+
+    const dialog = await portal.findByRole("dialog");
+
+    const closeButton = within(dialog).getByRole("button", { name: /close/i });
+    const firstInput = within(dialog).getByLabelText(/name/i);
+
+    await userEvent.tab();
+    expect(closeButton).toHaveFocus();
```
edleeks87: comment: the expects need to be `await`ed

edleeks87: `await expect(closeButton).toHaveFocus();`

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    await userEvent.tab();
+    expect(firstInput).toHaveFocus();
```
edleeks87: `await expect(firstInput).toHaveFocus();`

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    await userEvent.tab();
+    await userEvent.tab();
+
+    await userEvent.tab();
+    expect(closeButton).toHaveFocus();
```
edleeks87: `await expect(closeButton).toHaveFocus();`

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    const portal = within(canvasElement.ownerDocument.body);
+
+    const openButton = canvasElement.querySelector(
+      '[data-role="open-dialog-btn"]',
+    ) as HTMLElement;
+
+    if (!openButton) {
+      throw new Error("Open dialog button not found");
+    }
```
damienrobson-sage: **comment**: you don't need this - you're controlling the button so it will always be there (if it isn't then something else has gone wrong with Storybook, which would have much wider implications that breaking this play function)

src/components/dialog/dialog.interactions.stories.tsx
```diff
@@ -61,21 +61,14 @@ export const FocusManagement: Story = {
...
-    (dialog as HTMLElement).focus();
-
+    const dialog = await portal.findByRole("dialog");
     const closeButton = within(dialog).getByRole("button", { name: /close/i });
     const firstInput = within(dialog).getByLabelText(/name/i);
 
-    await expect(dialog).toHaveFocus();
-
-    await userEvent.tab();
+    (closeButton as HTMLElement).focus();
```
edleeks87: We shouldn't be programatically focusing imo, we want to test the focus trap so should be letting focus start on the dialog wrapper, I think what we had previously was better if we're doing interaction tests etc. 

I think if we just remove the code below it'll work fine, when the dialog opens, focus will be placed on the content wrapper. One Tab will take it to the firstInput, then the next three tabs will focus the form footer buttons then the close button which we will then assert is focused

```
    (closeButton as HTMLElement).focus();
    await expect(closeButton).toHaveFocus();

```

damienrobson-sage (dismissed):

edleeks87 (dismissed):

PR #7524, opened 2025-09-10
feat(text-editor): audit changes

damienrobson-sage:
Adds support for underlined text with command button, form-based hyperlink creation and input sizing. Button grouping introduced and placeholder typography dropdown added.

### Proposed behaviour

- Users can underline text
- Users can change the font style
- Users can now add links via form-based inputs
- Buttons are grouped by function, with dividers when appropriate
- Users can mention other users by providing a list of search options and then typing the `@` character into the editor followed by letters within said option
- Structure of the `TextEditor` component folder has been refreshed to make it easier to maintain and extend.

### Current behaviour

- Users cannot underline text
- Users cannot change the font style
- Users can only add links by pasting over existing text, or typing/pasting into empty space
- Buttons are not grouped by function
- Users cannot mention other users
- File structure is disorganised and random

### Checklist

- [x]  Playwright automation tests added or updated if required

#### QA

### Additional context

A new `Demo` storybook entry has been added to allow customers the opportunity to visually check what props do before the implement the component. This is also useful for testing combinations of props together.

### Testing instructions

Use the demo story mentioned above to test out the new functionalities.


src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+interface DropdownOption {
+  id: string;
+  label: string;
+  onClick: () => void;
+  ariaLabel?: string;
+}
+
+interface ToolbarDropdownProps {
+  ariaLabel?: string;
```
Parsium: **issue**: it seems overriding the translation key `textEditor.typography.selectAria` does not change this component's accessible name.

It appears it's due to our mixed use of kebab and pascal case for the prop name. Could we use the kebab version to fix this?: 

```suggestion
  "aria-label"?: string;
```

src/locales/de-de.ts
```diff
@@ -171,6 +171,25 @@ const deDE: Partial<Locale> = {
     saveButtonAria: () => "Speichern",
     toolbarAriaLabel: () => "Formatierung",
     unorderedListAria: () => "Unsortierte Liste",
+    hyperlink: {
+      buttonAria: () => "Hyperlink",
```
Parsium, line 176: **comment**: we don't necessarily need to include these translations here, since `i18nProvider` should fallback to `en-GB` if the keys don't exist.

We'll need to get these keys translated soon though.

src/components/note/note.component.tsx
```diff
@@ -12,7 +12,7 @@ import {
 } from "./note.style";
 import StatusIcon from "./__internal__/status-icon";
 import { ActionPopover } from "../action-popover";
-import ReadOnlyEditor from "../text-editor/__internal__";
+import ReadOnlyEditor from "../text-editor/__internal__/__ui__/ReadOnlyEditor/read-only-rte.component";
```
Parsium, line 15: **suggestion(non-blocking)**: would it be worth moving this export into a shared location that both `Note` and `TextEditor` can access?

src/components/text-editor/text-editor.stories.tsx
```diff
@@ -41,10 +46,81 @@ const meta: Meta<typeof TextEditor> = {
 export default meta;
 type Story = StoryObj<typeof TextEditor>;
 
-export const Default: Story = () => {
-  return <TextEditor namespace="storybook-default" labelText="Text Editor" />;
+export const Demo: Story = {
```
Parsium, line 49: **praise**: nice idea to simplify our use of storybook here :+1:

src/components/text-editor/text-editor.component.tsx
```diff
@@ -155,6 +105,18 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
         "`value` is deprecated in TextEditor and support will soon be removed. Please use `initialValue` instead.",
       );
     }
+    if (!deprecateOnCancelWarnTriggered && onCancel) {
```
Parsium, line 109: **comment**: how easy would it be to split the deprecation of these props into separate commits?

It could help with visibility by ensuring its part of our release notes.

src/components/text-editor/text-editor-test.stories.tsx
```diff
@@ -1,16 +1,17 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Parsium: **suggestion**: could we let TypeScript infer the relevant types based on the component's props? This would allow us to avoid disabling the linter.

src/components/text-editor/text-editor.stories.tsx
```diff
@@ -41,10 +46,81 @@ const meta: Meta<typeof TextEditor> = {
...
+      "typography",
+      "bold",
+      "italic",
+      "underline",
+      "unordered-list",
+      "ordered-list",
+      "link",
+    ],
+    // Deprecated
+    value: undefined,
```
Parsium: **suggestion(non-blocking)**: what do you think about using the non-deprecated version for simplicity?:

```suggestion
    initialValue: undefined,
```

src/components/text-editor/text-editor.pw.tsx
```diff
@@ -183,23 +184,24 @@ test.describe("Prop tests", () => {
     }) => {
       await mount(<TextEditorWithHeader />);
 
-      await page.keyboard.press("Tab");
       const button1 = page.getByRole("button", { name: "foo" });
+      const button2 = page.getByRole("button", { name: "bar" });
+      const button3 = page.getByRole("button", { name: "baz" });
+      const editor = page.getByRole("textbox");
+      const typographyDropdown = page.locator(
+        "button[data-role='pw-rte-typography-dropdown']",
```
Parsium: **suggestion(non-blocking)**: could we use the `getByTestId` locator for searching for elements by data-role tag?

src/components/text-editor/__internal__/__utils__/interfaces.types.ts
```diff
@@ -0,0 +1,201 @@
+import { MarginProps } from "styled-system";
```
Parsium, line 1: **suggestion(non-blocking)**: might be worth making the extension of this file `.type.ts`. Given it only contains type interfaces, this would avoid the build generating an empty JavaScript file.

src/components/text-editor/__internal__/__plugins__/LinkMonitor/link-monitor.plugin.ts
Parsium, line 1: **issue**: I noticed we are referencing nodes' double underscore methods, which [are private according to Lexical](https://lexical.dev/docs/concepts/nodes#node-properties):

>By convention, we prefix properties with __ (double underscore) so that it makes it clear that these properties are private and their access should be avoided directly.
> ...
>If you are adding a property that you expect to be modifiable or accessible, then you should always create a set of get*() and set*() methods on your node for this property.

To access the URL when a link is destroyed, what do you think about using its official getters like below? This could avoid problems if we later update Lexical and the internals of `AutoLinkNode` have changed.

```ts
// line 45
prevEditorState.read(() => {
  const node = $getNodeByKey(nodeKey);
  
  if (!$isAutoLinkNode(node)) {
    return;
  }

  onLinkAdded?.(node.getURL(), mutation);
});
```

src/components/text-editor/__tests__/utils/TestEditor.tsx
```diff
@@ -0,0 +1,136 @@
+/* istanbul ignore file */
+import React from "react";
+import { LexicalComposer } from "@lexical/react/LexicalComposer";
+import { RichTextPlugin } from "@lexical/react/LexicalRichTextPlugin";
+import { ContentEditable } from "@lexical/react/LexicalContentEditable";
+import LexicalErrorBoundary from "@lexical/react/LexicalErrorBoundary";
```
Parsium: **suggestion**: it appears the default export has been deprecated in favour of the named export.

Could we use the named one to avoid issues later?:

```suggestion
import { LexicalErrorBoundary } from "@lexical/react/LexicalErrorBoundary";
```

src/components/text-editor/__tests__/text-editor.test.tsx
```diff
@@ -464,7 +415,7 @@ test("serialisation of editor", async () => {
   render(
     <TextEditor
       labelText="Text Editor"
-      onSave={(values) => mockSave(values)}
+      onSave={(values: any) => mockSave(values)}
```
Parsium: **question**: do we need this any type here? If we can, I think we should let TypeScript infer this to avoid disabling the linter.

Parsium: **question**: could we let TypeScript infer the relevant types? This is avoid the need to use `any` and  disable the linter.

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // JSON import, used to deserialize the node from JSON
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static importJSON(serializedNode: any): MentionNode {
+    return new MentionNode(
+      serializedNode.__mention,
+      serializedNode.text,
+      serializedNode.key,
+    );
+  }
```
Parsium: **suggestion**: to avoid disabling the linter, what do you think about using a custom type for the serialised node?:

```suggestion
  exportJSON(): SerializedMentionNode {
    return {
      ...super.exportJSON(),
      type: "mention",
      version: 1,
      mention: this.__mention,
    };
  }

  static importJSON(serializedNode: SerializedMentionNode): MentionNode {
    return new MentionNode(serializedNode.mention, serializedNode.text); // omitted key based on https://lexical.dev/docs/concepts/key-management#when-to-use-__key
  }
```

where `SerializedMentionNode` is defined as:

```ts
interface SerializedMentionNode extends SerializedTextNode {
  mention: string;
}

File: src/components/text-editor/__internal__/__nodes__/styled-span.node.ts [hunk 13]
```diff
@@ -0,0 +1,221 @@
...
+      version: 1,
+      fontWeight: this.__fontWeight,
+      fontSize: this.__fontSize,
+      lineHeight: this.__lineHeight,
+    };
+  }
+
+  // JSON import, used to deserialize the node from JSON
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static importJSON(serializedNode: any): StyledSpanNode {
```
Parsium: **suggestion**: similar to `MentionNode`, could we use a type for the serialised node to improve type safety?

src/components/text-editor/text-editor.mdx
```diff
@@ -304,6 +210,29 @@ const App = () => (
 
 <Canvas of={TextEditorStories.WithCustomPlugins} />
 
+### Mentions
+
+The Text Editor supports mentions, and offers a built-in `Mentions` component to facilitate this. The `Mentions` component provides an interface for displaying a list of mentionable items and handling user interactions.
```
Parsium, line 221: **question**: if we are making mentions support a first-class feature, should we offer control via `TextEditor` props instead?

I appreciate plugin support was added as a temporary fix when projects urgently requested mentions functionality after our DraftJS to Lexical migration. Is it worth reserving plugins for truly custom functionality to simplify the interface?

src/components/text-editor/text-editor-test.stories.tsx
```diff
@@ -202,7 +196,7 @@ export const OnChangeFormattedValues: Story = () => {
       <TextEditor
         namespace="storybook-onchange-formatted-values"
         labelText="Text Editor"
-        onChange={(_, { htmlString, json }) => {
+        onChange={(_: any, { htmlString, json }: any) => {
```
Parsium: **question**: do we need the `any` types here? It seems TypeScript might be able to infer them, which could give us better type safety.

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,153 @@
...
+      options={options}
+      menuRenderFn={(
+        anchorElementRef,
+        { selectedIndex, selectOptionAndCleanUp, setHighlightedIndex },
+      ) =>
+        anchorElementRef.current && results.length
+          ? ReactDOM.createPortal(
+              <div className="typeahead-popover mentions-menu">
+                <ul data-role="mention-list">
+                  {options.map((option, i: number) => (
```
Parsium, line 147: **issue**: I noticed when the mention list appears, the currently highlighted option isn't currently being announced by a screen reader when navigating via the arrow keys or typing (see video below).

Given users may need to insert valid mentions, for example to successfully reference names registered in an application, assistive technology users may have difficulty achieving this.

https://github.com/user-attachments/assets/39ad1484-7b87-4240-a2b5-8ea09f9ff080

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,153 @@
...
+      return;
+    }
+
+    mentionsCache.set(queryString, null);
+    lookupService(queryString, (newResults) => {
+      mentionsCache.set(queryString, newResults);
+      setResults(newResults);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [queryString]);
```
Parsium: **question**: is there a reason we need to omit the missing dependencies here?

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,153 @@
...
+    [editor],
+  );
+
+  // Trigger function to check for mention matches in the text.
+  const checkForMentionMatch = useCallback((text: string) => {
+    return getPossibleQueryMatch(text);
+  }, []);
+
+  useEffect(() => {
+    const cachedResults = mentionsCache.get(queryString);
```
Parsium: **question**: do we need to cache the results here? From what I tell, a re-render shouldn't occur if `results` is unchanged, but maybe you've seen performance implications I might be missing?

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,121 @@
...
+  cursor: pointer;
+  display: inline-block;
+  font-size: 0.875rem;
+  font-weight: 400;
+  line-height: 1.25rem;
+  min-height: 21px;
+  padding: 0px 8px;
+`;
+
+export class MentionNode extends TextNode {
```
Parsium, line 29: **comment**: maybe its worth tightening what's encapsulated by this class, and clarifying what isn't meant for direct use.

Lexical has some constraints around node use and custom nodes (e.g. [not referencing private properties](https://lexical.dev/docs/concepts/nodes#node-properties) like `__mention`, [don't call `clone` directly](https://lexical.dev/docs/concepts/node-cloning#when-not-to-use-clone)), and making that explicit could help avoid accidental misuse.


```ts
export class MentionNode extends TextNode {
  #mention: string;

  getMention(): string {
    const self = this.getLatest();
    return self.#mention;
  }

  // Do not use directly. To be used by Lexical only. Use Lexical's `$copyNode` for duplicating nodes.
  static clone(node: MentionNode): MentionNode {
    return new MentionNode(node.#mention, node.__text, node.__key);
  }
}
```

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.style.ts
```diff
@@ -0,0 +1,88 @@
...
+  font-size: var(--fontSizes200);
+
+  height: ${({ size }) => getSizeForComponent(size)};
+
+  ${({ menuOpen }) =>
+    menuOpen &&
+    css`
+      /* border: 1px solid var(--colorsActionMinor500);
+      box-shadow: var(--boxShadow100);
+      background-color: var(--colorsUtilityYang100); */
```
Parsium: **comment**: there's some commented out code here :)

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+  useEffect(() => {
+    if (isOpen && focusedIndex >= 0 && menuRef.current) {
+      const items =
+        menuRef.current.querySelectorAll<HTMLElement>('[role="menuitem"]');
+      items[focusedIndex]?.focus();
+    }
+  }, [isOpen, focusedIndex]);
+
+  // Handle keyboard navigation
+  const handleKeyDown = (event: React.KeyboardEvent) => {
```
Parsium, line 86: **issue**: I noticed the keyboard behaviour for the typography option listbox appears to differ from the expected pattern for an ARIA menu or combobox. Currently, the listbox doesn't move focus between options on <kbd>Home</kbd>, <kbd>End</kbd> or character key presses, as we are using the toolbar's focus index state.

This might be unexpected for users relying on the keyboard or assistive tech. Could we adjust this to align with the expected behaviour?

https://github.com/user-attachments/assets/3f4d1d62-a2ea-4880-b178-a420374eeebc

.hintrc
edleeks87, line 1: comment: should this file be included

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,121 @@
...
+    return {
+      ...super.exportJSON(),
+      type: "mention",
+      version: 1,
+    };
+  }
+
+  // JSON import, used to deserialize the node from JSON
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static importJSON(serializedNode: any): MentionNode {
```
edleeks87:
```suggestion
    static importJSON(serializedNode: SerializedTextNode & { __mention: string; text: string; key: NodeKey }): MentionNode {

```

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,121 @@
...
+  // Segmented mode is used for mentions to allow the text to be broken into segments
+  // When a node is "directionless", it inherit the text direction from its parent
+  mentionNode.setMode("segmented").toggleDirectionless();
+
+  // Apply the node replacement to the mentionNode
+  return $applyNodeReplacement(mentionNode);
+}
+
+export function $isMentionNode(
+  node: LexicalNode | null | undefined,
```
edleeks87:
```suggestion
  node?: LexicalNode | null,
```
(non-blocking)

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function convertMentionElement(
+  domNode: HTMLElement,
+): DOMConversionOutput | null {
+  // Get the text content and mention name from the DOM node
+  const { textContent } = domNode;
+  const mentionName = domNode.getAttribute("data-lexical-mention-name");
+
+  // If the text content is null, return null
+  if (!!textContent && textContent.length > 0) {
```
edleeks87:
```suggestion
  if (textContent?.length) {
```

(non-blocking I just can't help myself)

src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
```diff
@@ -0,0 +1,221 @@
...
+  static importDOM(): DOMConversionMap | null {
+    return {
+      span: (domNode: HTMLElement) => ({
+        conversion: () => {
+          const fontWeight = domNode.style.fontWeight || "400";
+          const fontSize = domNode.style.fontSize || "14px";
+          const lineHeight = domNode.style.lineHeight || "21px";
+          return {
+            node: new StyledSpanNode(
+              domNode.textContent || "",
```
edleeks87, line 169:
```
          const { textContent: content, style } = domNode;
          const { fontWeight = "400", fontSize = "14px", lineHeight = "21px" } = style;

          return {
            node: new StyledSpanNode(
              textContent ?? "",
              fontWeight,
              fontSize,
              lineHeight,
            ),
          };
```

src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
edleeks87:
```suggestion
  static importJSON(serializedNode: SerializedTextNode & { fontWeight: string, fontSize: string, lineHeight: string }): StyledSpanNode {
```

src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
```diff
@@ -0,0 +1,221 @@
...
+    prevNode: StyledSpanNode,
+    dom: HTMLElement,
+    config: EditorConfig,
+  ): boolean {
+    const updated = super.updateDOM(prevNode, dom, config);
+
+    let styleChanged = false;
+    if (this.__fontWeight !== prevNode.__fontWeight) {
+      dom.style.fontWeight = this.__fontWeight;
+      styleChanged = true;
```
edleeks87: comment (non-blocking): I think we can just return true here and remove the styleChanged variable 

```
updateDOM(
    prevNode: StyledSpanNode,
    dom: HTMLElement,
    config: EditorConfig,
  ): boolean {
    const updated = super.updateDOM(prevNode, dom, config);

    if (this.__fontWeight !== prevNode.__fontWeight) {
      dom.style.fontWeight = this.__fontWeight;
      return true;
    }
    if (this.__fontSize !== prevNode.__fontSize) {
      dom.style.fontSize = this.__fontSize;
      return true;
    }
    if (this.__lineHeight !== prevNode.__lineHeight) {
      dom.style.lineHeight = this.__lineHeight;
      return true;
    }

    // Return true if the node was updated, false otherwise
    // This is important for the editor to know if it needs to re-render the node
    // If the text content or styles have changed, we return true
    // If the text content is the same but styles have changed, we also return true
    // If neither has changed, we return false
    // This helps optimize rendering performance and ensures that the editor
    // only re-renders nodes when necessary
    return updated;
  }

File: src/components/text-editor/__internal__/__ui__/Mentions/helpers.ts [hunk 27]
```diff
@@ -0,0 +1,45 @@
...
+  AtSignMentionsRegex,
+  AtSignMentionsRegexAliasRegex,
+} from "./constants";
+
+export function checkForAtSignMentions(
+  text: string,
+  minMatchLength: number,
+): MenuTextMatch | null {
+  // Check the text against the regex for mentions.
+  let match = AtSignMentionsRegex.exec(text);
```
edleeks87:
```suggestion
  const match = AtSignMentionsRegex.exec(text) || AtSignMentionsRegexAliasRegex.exec(text);
```

non-blocking

src/components/text-editor/__internal__/__ui__/Mentions/mentions-typeahead-menu-item.component.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onMouseEnter,
+  option,
+}: {
+  index: number;
+  isSelected: boolean;
+  onClick: () => void;
+  onMouseEnter: () => void;
+  option: MentionTypeaheadOption;
+}) => {
+  let className = "item";
```
edleeks87:
```suggestion
  const className = "item" + (isSelected ? " selected" : "");
```
non-blocking

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,153 @@
...
+        }
+        mentionNode.select();
+        closeMenu();
+      });
+    },
+    [editor],
+  );
+
+  // Trigger function to check for mention matches in the text.
+  const checkForMentionMatch = useCallback((text: string) => {
```
edleeks87: comment: there's not much value in using `useCallback` if the dep array is empty

```
const checkForMentionMatch = (text: string) => {
    return getPossibleQueryMatch(text);
  };

File: src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx [hunk 30]
```diff
@@ -0,0 +1,153 @@
...
+      setResults(cachedResults);
+      return;
+    }
+
+    mentionsCache.set(queryString, null);
+    lookupService(queryString, (newResults) => {
+      mentionsCache.set(queryString, newResults);
+      setResults(newResults);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
edleeks87: question: is this needed?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography.component.tsx
```diff
@@ -0,0 +1,204 @@
...
+        }
+      }
+    });
+  };
+
+  return (
+    <Box minWidth={"150px"} marginRight="spacing05">
+      <ToolbarDropdown
+        size={size}
+        aria-label={locale.textEditor.typography.selectAria()}
```
edleeks87: comment (non-blocking): I wonder if we need to pass this as a prop or just consume the context in the dropdown

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography.component.tsx
```diff
@@ -0,0 +1,204 @@
...
+        namespace={namespace}
+        onChange={(value) => handleChange(value)}
+        value={selectedOption}
+        isFirstButton={isFirstButton}
+        options={(
+          [
```
edleeks87: comment: if you assign this to array to an variable rather than inline we can remove the need to cast the array

```
const options: TypographyKey[] = [
  "title",
  "subtitle",
  "sectionHeader",
  "sectionSubheader",
  "paragraph",
];

File: src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography.component.tsx [hunk 33]
```diff
@@ -0,0 +1,204 @@
...
+            "title",
+            "subtitle",
+            "sectionHeader",
+            "sectionSubheader",
+            "paragraph",
+          ] as TypographyKey[]
+        ).map((key) => {
+          const ariaLabel = locale.textEditor.typography[key]();
```
edleeks87: comment: similar to above we could move the context into the dropdown and use the key to index there

src/components/text-editor/__internal__/__ui__/Toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,293 @@
...
+        });
+      }),
+    );
+  }, [updateToolbar, editor]);
+
+  // Register a keydown listener to enable keyboard navigation
+  const handleToolbarKeyDown = (event: React.KeyboardEvent) => {
+    if (toolbarControls.length === 0 || !toolbarRef.current) return;
+
+    const currentButtons = Array.from(
```
edleeks87:
```suggestion
    const currentButtons: HTMLButtonElement[] = Array.from(
```

src/components/text-editor/__internal__/__ui__/Toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,293 @@
...
+        event.preventDefault();
+        nextIndex = currentButtons.length - 1;
+        break;
+      default:
+        return;
+    }
+
+    // Update tabIndex and focus
+    currentButtons.forEach((button, index) => {
+      (button as HTMLButtonElement).tabIndex = index === nextIndex ? 0 : -1;
```
edleeks87:
```suggestion
      button.tabIndex = index === nextIndex ? 0 : -1;
```

src/components/text-editor/__internal__/__ui__/Toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,293 @@
...
+    });
+
+    (currentButtons[nextIndex] as HTMLButtonElement)?.focus();
```
edleeks87:
```suggestion
    currentButtons[nextIndex]?.focus();
```

src/components/text-editor/text-editor.component.tsx
```diff
@@ -326,48 +288,62 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
...
-                        namespace={namespace}
-                        previews={previews}
-                        rows={rows}
-                        readOnly={readOnly}
-                        required={required}
-                        error={!!error}
-                        warning={!!warning || !!characterLimitWarning}
-                        validationMessagePositionTop={
-                          validationMessagePositionTop
+                {!readOnly ? (
```
edleeks87: comment (non-blocking): I think it's better to check the truthiness first then render the non readonly if it's false. I don't have a reason for it other than that's how I've always been made to do it 😆

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/__tests__/dropdown.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+          onChange={onChange}
+          namespace="test"
+        />
+      </TestEditor>,
+    );
+    expect(screen.getByRole("button")).toBeInTheDocument();
+    expect(screen.getByRole("button")).toHaveTextContent("Option 1");
+  });
+
+  it("opens menu on button click", () => {
```
Parsium: **suggestion**: since `Toolbar` controls the dropdown list's visibility, would it make more sense for the open/close tests to be part of `Toolbar`'s tests? Right now, these look like we're testing implementation details rather than user behaviour (e.g. verifying dropdown is visible after clicking the button).

src/components/text-editor/__internal__/__nodes__/mention.node.test.tsx
Parsium, line 1: **suggestion**: what do you think about porting over most (or all) of `TextEditor`'s tests to Playwright?

Since we're re-creating a simplified version of `TextEditor` and frequently mock functionality to get around JSDOM's limitations, this feels like a good justification for it. Testing in a real browser should give us greater confidence that we're capturing actual behaviour.

We'd need to manually verify our code coverage is sufficient, but overall this seems like a stronger long-term approach.

Parsium:
Nice work @damienrobson-sage, appreciate this is a large undertaking 👍🏼 

I've left some comments on a couple of issues spotted regarding accessibility and the component's API, along with some general questions and smaller suggestions. Let me know what you think :)

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: Will take a look, might be able to refactor

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
damienrobson-sage: Took Ed's suggestion to just consume within the dropdown rather than pass in via props 👍

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: Thanks, looking into this as a priority 👍

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: As per my reply to [Ed's comment](https://github.com/Sage/carbon/pull/7524#discussion_r2386804882), I'll see if refactoring this away.

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: The idea of caching was to cover cases where users might be constantly searching a large list of possible mentions. It might not be needed but we've had little to no steer from teams in terms of just how big the mention lists could be. If you're OK with me doing so, I'd like to leave it in for the time being until we have a better understanding of how teams are implementing this.

In PDS, for example, I saw companies with 100s of named employees. If each of these was mentionable, it _could_ lead to issues. I'll see if I can add some more relevant test stories that we can use to determine performance impact.

src/components/text-editor/__internal__/__nodes__/mention.node.test.tsx
damienrobson-sage: As discussed last week, this is a good shout. What do you think about doing it separately, though, so as to not delay further? We can also do some investigation into the PlayWright + Jest coverage setup as I don't recall any other components being purely PW-tested?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/__tests__/dropdown.test.tsx
damienrobson-sage: It's a good shout but based on your comment about moving most/all tests to PW, would it be better holding off on this refactor until that's done/not done?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
damienrobson-sage: As mentioned in my other comment, I'll look into this as a priority 😄

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: Looked into this, expanding the deps so as to remove the lint rule causes an infinite render to occur.

src/components/text-editor/text-editor.component.tsx
damienrobson-sage: Should be quite straightforward to do. Will do it last so that they aren't accidentally swallowed by other commits, etc.

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
damienrobson-sage: Hannah's had a play around with this and it's a Safari-only issue. Every other screen reader behaves as expected.

src/components/text-editor/text-editor.mdx
damienrobson-sage: My thinking here was to avoid making it first-class as not everyone will want/need it; I'd rather make it opt-in initially as I don't expect this feature to be widely used (which is also why I'd like to keep it as a plugin as opposed to core functionality)

src/components/note/note.component.tsx
damienrobson-sage: Given there's no widespread usage of the read-only editor I'll leave it where it is for now, but it's a good shout if we ever consider additional house-keeping.

src/components/text-editor/text-editor.component.tsx
damienrobson-sage: These have been included in #7558 so I'll mark this as resolved.

.hintrc
```diff
@@ -0,0 +1,30 @@
+{
```
Parsium: **suggestion**: would we be able to remove this new added file from the git changes, given we've configured git to ignore it?

src/components/text-editor/text-editor.component.tsx
```diff
@@ -272,6 +235,7 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
       <StyledTextEditorWrapper
         data-role={`${namespace}-editor-wrapper`}
         onBlur={(ev) => {
+          /* istanbul ignore next */
           if (!ev.currentTarget.contains(ev.relatedTarget)) {
```
Parsium, line 251: **suggestion(non-blocking)**: given we have these blur and focus handlers here, we could use them to set the `isFocused` state and remove the effect on lines 141-160. Not essential though.

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,260 @@
...
+    if (isOpen && focusedIndex >= 0 && menuRef.current) {
+      const items =
+        menuRef.current.querySelectorAll<HTMLElement>('[role="option"]');
+      items[focusedIndex]?.focus();
+    }
+  }, [isOpen, focusedIndex]);
+
+  // Handle keyboard navigation
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    switch (event.key) {
```
Parsium, line 87: **issue**: I noticed the listbox doesn't have [typeahead support](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#keyboard_interactions), which is recommended to help users quickly jump to an option by typing its name. Would we be able to introduce this?

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,125 @@
+/**
+ * IMPORTANT: The methods and functions included in this file are to be consumed internally by
+ * `Lexical` only; they therefore should not be used directly. If you need similar functionality,
+ * please refer to the Lexical documentation and utilize the appropriate public APIs.
+ */
```
Parsium: **comment**: cheers for adding this disclaimer in @damienrobson-sage.

I think we could make it a little bit clearer which `MentionNode` methods (and helpers) are safe to use, mainly so contributors unfamiliar with Lexical can understand. For example, we'll likely need `$isMentionNode` and `MentionNode.exportJSON()` when working with nodes, but we should avoid using `MentionNode.clone()` and any private properties to prevent potential problems with Lexical.

Maybe we could instead add a jsdoc comment to `MentionNode`'s `clone` method and private properties to make this explicit?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  useEffect(() => {
+    if (isOpen && focusedIndex >= 0 && menuRef.current) {
+      const items =
+        menuRef.current.querySelectorAll<HTMLElement>('[role="option"]');
+      items[focusedIndex]?.focus();
+    }
+  }, [isOpen, focusedIndex]);
+
+  // Handle keyboard navigation
```
Parsium, line 85: **issue**: it seems the focus behaviour of the toolbar may be conflicting with the dropdown listbox. For example, pressing the left or right arrow keys moves focus in the toolbar. Should these keys be ignored while the listbox is visible?

https://github.com/user-attachments/assets/3183b878-7af7-4f3c-a724-f296fbeaad5c

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,260 @@
...
+        onKeyDown={handleKeyDown}
+        onClick={handleButtonClick}
+        className="toolbar-button"
+        tabIndex={isFirstButton ? 0 : -1}
+        data-role={`${namespace}-typography-dropdown`}
+        id={`${namespace}-typography-dropdown`}
+        onMouseDown={
+          /* istanbul ignore next */ (e: React.MouseEvent<HTMLButtonElement>) =>
+            e.preventDefault()
+        }
```
Parsium: **question**: just wondering what's this used for?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/hyperlink.component.tsx
```diff
@@ -0,0 +1,164 @@
...
+          }
+          iconType="link_on"
+          buttonType={"tertiary"}
+          data-role={`${namespace}-hyperlink-button`}
+          id={`${namespace}-hyperlink-button`}
+          tabIndex={isFirstButton ? 0 : -1}
+          className="toolbar-button"
+          ref={ref}
+        />
+        <Dialog
```
Parsium: **issue**: it seems on close the focus isn't being restored to the hyperlink button. Debugging this, we are correctly refocusing the hyperlink button programmatically when the dialog is closed, but it seems there's a timing issue as focus is lost to the document body.

src/components/text-editor/__internal__/__ui__/Toolbar/toolbar.component.tsx
```diff
@@ -0,0 +1,293 @@
...
+    );
+  }, [updateToolbar, editor]);
+
+  // Register a keydown listener to enable keyboard navigation
+  const handleToolbarKeyDown = (event: React.KeyboardEvent) => {
+    if (toolbarControls.length === 0 || !toolbarRef.current) return;
+
+    const currentButtons: HTMLButtonElement[] = Array.from(
+      toolbarRef.current?.querySelectorAll(
+        'button[role="button"].toolbar-button, button.toolbar-button',
```
Parsium: **suggestion(non-blocking)**: could we simplify this selector like so?:


```suggestion
        'button.toolbar-button',
```

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  // Apply the node replacement to the mentionNode
+  return $applyNodeReplacement(mentionNode);
+}
+
+export function $isMentionNode(node?: LexicalNode | null): node is MentionNode {
+  return node instanceof MentionNode;
+}
+
+export function convertMentionElement(
```
Parsium: **nitpick**: might be worth prefixing this as per [Lexical's conventions](https://lexical.dev/docs/concepts/nodes#extending-elementnode-with-config):

```suggestion
export function $convertMentionElement(
```

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,212 @@
...
+import MentionTypeaheadOption from "./mention-typeahead-option.class";
+
+import {
+  $createMentionNode,
+  $isMentionNode,
+} from "../../__nodes__/mention.node";
+import Icon from "../../../../icon";
+import useLocale from "../../../../../hooks/__internal__/useLocale";
+
+import "./style.css";
```
Parsium: **issue**: it seems this stylesheet isn't being included in final Carbon build, causing the following:

Screenshot 2025-10-08 at 15 56 54

Could we get around this by making our mentions menu a styled component?

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
```diff
@@ -0,0 +1,212 @@
...
+      setResults(cachedResults);
+      return;
+    }
+
+    mentionsCache.set(queryString, null);
+    lookupService(queryString, (newResults) => {
+      mentionsCache.set(queryString, newResults);
+      setResults(newResults);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Parsium: **comment**: I see what you mean about optimising performance if `searchOptions` is large, that makes sense. I'd still be wary about omitting effect dependencies though. If doing so results in an infinite loop, it could be hinting at a broader problem.

To optimise repeated lookups, maybe we could memorise a value derived from `queryString` instead? Something like this:

```tsx
  const [queryString, setQueryString] = useState<string | null>(null);
  const results = useMemo(
    () =>
      queryString
        ? searchOptions.filter((mention) =>
            mention.name.toLowerCase().startsWith(queryString.toLowerCase()),
          )
        : [],
    [queryString, searchOptions],
```

src/components/text-editor/__internal__/__ui__/Toolbar/__tests__/toolbar.test.tsx
```diff
@@ -0,0 +1,893 @@
...
+        />
+      </LexicalComposer>,
+    );
+    const buttonGroups = screen.getAllByTestId(/btg-test/i);
+    expect(buttonGroups.length).toEqual(4);
+
+    // Because the separator is hidden via aria-hidden we need to use the container
+    // to find it instead of screen.getByRole
+    // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+    const dividers = container.querySelectorAll("[role='separator']");
```
Parsium: **suggestion(non-blocking)**: what do you think of the following? This way we can avoid disabling the linter:


```suggestion
    const dividers = screen.queryAllByRole("separator", { hidden: true });
```

src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx
edleeks87: I think it might be due to the lack of memoisation on the `lookUpService`, worth exploring if wrapping in a `useCallback` resolves this intentionally omitting the deps can lead to weird behaviour. I also wondered why is the `mentionsCache` being set to null then updated, wouldn't an overwrite be enough? If it's still causing infinite loops you might want to check if the `mentionsCache` has already been set with the same results and bail out if it has but I'm not sure if we want to prevent it being updated if it's the same etc.


```
const lookupService = useCallback((
    string: string,
    callback: (results: Array<Mention>) => void,
  ): void => {
    setTimeout(() => {
      const results = searchOptions.filter((mention) =>
        mention.name.toLowerCase().includes(string.toLowerCase()),
      );
      callback(results);
    }, 500);
  }, [searchOptions]);

File: src/components/text-editor/__internal__/__ui__/Mentions/style.css
Comment (edleeks87) [inline src/components/text-editor/__internal__/__ui__/Mentions/style.css:1]: comment: I've noticed this file isn't being included in the bundle when we run the build script. We'll need to update the rollup config. You can probably just copy the file across like we are doing but also we could minify if you want. Happy to help with this if you need. I've also noticed that `__tests__` files are being included in the bundle so we'll either need to update the naming convention to avoid dev code being bundled or tweak the config to exclude them, again happy to help on that if you need.

Image

File: src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx (hunk 44 above)
Comment (damienrobson-sage) [reply src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx]: Checked and apparently nothing... I'll remove it 😅

File: src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx (hunk 49 above)
Comment (damienrobson-sage) [reply src/components/text-editor/__internal__/__ui__/Mentions/mentions.component.tsx]: I've reworked to pretty much remove the caching and everything works just as well without half of the code 👍

File: src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography.component.tsx [hunk 51]
```diff
@@ -0,0 +1,234 @@
...
+        let newNode: TextNode | StyledSpanNode;
+
+        /* istanbul ignore if */
+        if (key === "paragraph") {
+          // Use a regular TextNode for paragraphs. Doing so ensures that
+          // the mentions plugin can still function correctly
+          newNode = new TextNode("");
+        } else {
+          // Use StyledSpanNode for other typography styles.
+          newNode = StyledSpanNode.createFromKey(key, "");
```
Parsium: **suggestion(non-blocking)**: what do you think about changing the naming of key to option/style?

This could help alleviate potential confusion with Lexical keys, given there is no similarity between the two.

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+  },
+}));
+
+describe("TypographySelector", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("renders with default paragraph option", () => {
+    const ref = React.createRef<HTMLDivElement>();
```
Parsium: **suggestion(non-blocking)**: might be worth avoiding `createRef` here, given [its part of the legacy class component API.](https://react.dev/reference/react/createRef)

```suggestion
    const ref = { current: null }
```

tsconfig-build.json
```diff
@@ -15,6 +17,7 @@
     "**/*.test.*",
     "**/*.stories.*",
     "**/*.pw.ts*",
-    "**/*.test-pw.ts*"
+    "**/*.test-pw.ts*",
+    "**/TestEditor.tsx"
```
Parsium: **suggestion**: what do you think about adding a `.test` extension to `TestEditor` instead of excluding it in the repo configs? This would allow us to keep our repo config files generalised.

src/components/text-editor/__internal__/__ui__/ContentEditor/content-editor.component.tsx
```diff
@@ -68,6 +54,26 @@ const ContentEditor = forwardRef<HTMLDivElement, ContentEditorProps>(
...
+              const targetId = event.relatedTarget?.id;
+              /* istanbul ignore else */
+              if (targetId) {
+                const targetElement = document.querySelector(
+                  `[id="${targetId}"]`,
+                );
+                /* istanbul ignore else */
+                if (targetElement) (targetElement as HTMLElement).focus();
+              }
+            }
```
Parsium, line 75: **question**: may I clarify what use case this covers?

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.component.tsx
```diff
@@ -0,0 +1,256 @@
...
+    } else {
+      setFocusedIndex?.(-1);
+    }
+  };
+
+  const handleOptionClick = (option: DropdownOption) => {
+    option.onClick();
+    setIsOpen?.(false);
+    setFocusedIndex?.(-1);
+    // buttonRef.current?.focus();
```
Parsium: **comment**: looks like we've got some commented code here :)

src/components/text-editor/__internal__/__nodes__/mention.node.ts
```diff
@@ -0,0 +1,163 @@
+/**
```
Parsium: **suggestion**: might be worth only adding an internal disclaimer for `MentionNode`'s `clone` method and private properties and removing the others.

This could help avoid potential confusion that the other public methods and helpers cannot be used, given there are use cases for them.

src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
```diff
@@ -0,0 +1,315 @@
+/**
```
Parsium: **suggestion**: similarly to `MentionNode`, it might be worth only adding an internal disclaimer to `StyledSpanNode`'s `clone` method and private properties to avoid potential confusion.

src/components/text-editor/__internal__/__ui__/ContentEditor/content-editor.component.tsx
damienrobson-sage: If you are in the `contenteditable` and shift-tab, the focus gets stuck between the div and the toolbar, however checking the target destination it is trying to get back to the toolbar. I've had to do this elsewhere fo similar reasons.

src/components/text-editor/__internal__/__nodes__/mention.node.ts
damienrobson-sage: Updated both of these to make sure that only publicly-consumable methods are documented "normally". Removed the top comment and marked anything private as `INTERNAL`

src/components/text-editor/__internal__/__ui__/Toolbar/toolbar.style.ts
```diff
@@ -5,30 +5,39 @@ import Button, { ButtonProps } from "../../../../button";
 interface FormattingButtonProps extends ButtonProps {
   tabIndex?: number;
   isActive?: boolean;
+  onMouseDown?: React.MouseEventHandler<HTMLButtonElement>;
 }
 
-const StyledToolbar = styled.div<{ hasHeader?: boolean }>`
+export const getPaddingForSize = (size: "small" | "medium" | "large") => {
```
edleeks87: comment (non-blocking): if you change this to have a default value you won't have to cast it below

```

export const getPaddingForSize = (size?: "small" | "medium" | "large") => {
  switch (size) {
    case "small":
      return "8px";
    case "large":
      return "16px";
    default:
      return "12px";
  }
};

File: src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.style.ts [hunk 59]
```diff
@@ -0,0 +1,90 @@
...
+type ButtonProps = { size: "small" | "medium" | "large"; menuOpen?: boolean };
+type ListProps = { size: "small" | "medium" | "large" };
+
+const getSizeForComponent = (size: "small" | "medium" | "large") => {
+  switch (size) {
+    case "small":
+      return "32px";
+    case "large":
+      return "48px";
+    case "medium":
```
edleeks87: comment (non-blocking): you don't need `case "medium"` here as the default will catch it

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.style.ts
```diff
@@ -0,0 +1,90 @@
...
+  cursor: pointer;
+  font-size: var(--fontSizes100);
+  font-weight: var(--fontWeights500);
+
+  height: ${({ size }) => getSizeForComponent(size)};
+
+  ${({ menuOpen }) =>
+    menuOpen &&
+    css`
+      ${addFocusStyling(false)}
```
edleeks87: comment(non-blocking): you don't need to pass false as it's the default in the util

src/components/text-editor/__internal__/__ui__/Toolbar/buttons/typography-dropdown/dropdown.style.ts
```diff
@@ -0,0 +1,90 @@
...
+export const StyledMenuItem = styled.li<{
+  isFocused: boolean;
+}>`
+  background: ${(props) => (props.isFocused ? "#f0f0f0" : "white")};
+  cursor: pointer;
+  user-select: none;
+  padding: 8px 16px;
+  color: var(--colorsUtilityYin055);
+
+  &:hover {
```
edleeks87: suggestion (non-blocking): we could avoid `!important` and potential inconsistencies it causes by using `&` to bump specificity instead

```
&& {
  &:hover {
      background: #f0f0f0;
      color: var(--colorsUtilityYin100);
    }
}

**Review by edleeks87 — DISMISSED** [review]

File: jest.config.ts [hunk 62]
```diff
@@ -44,7 +44,7 @@ const baseProjectConfig: Config = {
 
 const clientConfig: Config = {
   displayName: "Client",
-  testMatch: ["**/!(*.server).+(spec|test).[jt]s?(x)"],
+  testMatch: ["**/!(*.server|TestEditor).+(spec|test).[jt]s?(x)"],
```
Parsium: **suggestion(non-blocking)**: is this config change required, given `TestEditor`'s file has a `*.test.tsx` extension?

Parsium (dismissed):

src/components/text-editor/text-editor.component.tsx
```diff
@@ -176,6 +139,7 @@ export const TextEditor = forwardRef<TextEditorHandle, TextEditorProps>(
       [],
     );
 
+    // We need to handle focus and blur this way to prevent
```
Parsium: **nitpick**:

```suggestion
```

edleeks87 (dismissed):

Parsium (dismissed):

designerlisa:
Hi @damienrobson-sage, 
Most of it looks really great! Aside from Niko's comments, we’ve made a few fixes—please double-check that these are now aligned:

1. Text size updates to reflect hierarchy (this also affects the NOTE component):
Title text – `global.font.static.heading.M` (24px)
Subtitle text – `global.font.static.heading.S` (21px)
Section Heading text – `global.font.static.subheading.L` (18px)
Section Subheading text – `global.font.static.subheading.M` (16px)
[Link to Figma Design](https://www.figma.com/design/Cie8BIDWKj4tEXCVGsen5A/-BETA--DS-Components?node-id=113943-1633576&t=i6JKhA5ilgGvWoaw-4)

2. Spacing between hint text / container / text count:
S = 4px
M = 8px
L = 12pxx 
[Link to Figma Design](https://www.figma.com/design/xmJTHKX6KaUIrjUSZ77RHr?node-id=10-12838#1502174744)

Let me know if all are aligned!

PR #7525, opened 2025-09-11
fix(responsive-vertical-menu): fixes a bug when navigating with keyboard and VoiceOver

damienrobson-sage:
When navigating the menu in Safari with VoiceOver enabled, there was a potential race condition which could cause the menu to close instead of moving to the next available target

### Proposed behaviour

Update the `handleBlur` event handler to use `composedPath` and determine whether events raised come from outside of RVM that way before deciding to close the menu.

### Current behaviour

When navigating the `ResponsiveVerticalMenu` in Safari via keyboard and with VoiceOver running, it is possible to hit a condition whereby the menu would close automatically when a menu item is focused.

### Checklist

#### QA

### Testing instructions

- Open the `Default` story in the `Testing` section of the Responsive Vertical Menu storybook in Safari
- Open the menu with the keyboard and navigate around with the VO key (Control+Option) and the left/right arrows
- Ensure that the menu remains open throughout


src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu.component.tsx
```diff
@@ -166,29 +167,26 @@ const BaseMenu = ({
...
-              setActive(false);
-            }
-          }, 10);
+      timeout = setTimeout(() => {
+        if (!containerRef.current?.contains(document.activeElement)) {
+          setActive(false);
         }
+      }, 0);
+
+      /* istanbul ignore if */
```
edleeks87: comment: I'm not sure we need this but if we do it might be better to move it to the top of the function so it short circuits early

tomdavies73 (approved):

paulrobinson88 (approved):

PR #7526, opened 2025-09-11
feat(pill): add translation support to pill

### LINKED ISSUE #7514: Remove pill button in MultiSelect is not accessible
### Description

The remove pill button in a MultiSelect just reads "remove pill" and it isn't clear which option it refers to which fails WCAG 1.3.1 Info and Relationships

The HTML structure or the button's accessible name should make it clear which pill it removes

### Reproduction

https://carbon.sage.com/?path=/docs/select-multiselect--docs

### Steps to reproduce

Select multiple options in a MultiSelect

With a screen reader, if you go through buttons it'll just read "remove pill" several times

SBS-143005

### Suggested solution

This is a button with just an icon so aria could be acceptable

Allow users to provide aria properties for each button (through Option ?)

Or add an aria-label `${removePillTranslation} ${pillName}` 

Or add an aria-label `${removePillTranslation}` and aria-labelledby `${selfId} ${pillNameId}` 

Maybe there's a better solution with pure HTML

### Carbon version

156.3.1

### React version

v17

### Design tokens version

_No response_

### Relevant browsers

Chrome, Safari

### Relevant OSs

MacOS

### Additional context

_No response_

### Confidentiality

damienrobson-sage:
fixes #7514

### Proposed behaviour

Adds a translation entry to the locale files to allow customised aria-label messages. This is now applied by default; unless the user provides a value via `ariaLabelOfRemoveButton`, this translation entry will be used (defaults to `Remove XYZ pill` where "XYZ" is the value as displayed in the Pill)

### Current behaviour

The remove pill button fails WCAG 1.3.1 due to it not announcing the value of the pill it removes, which can be confusing to screen reader users when there are multiple pills on-screen

### Checklist

#### QA

### Testing instructions

A new story has been added into Pill's testing stories to demonstrate Pills with divergent aria labels, which can be used for testing

### DIFF
--- a/src/components/pill/pill-test.stories.tsx
+++ b/src/components/pill/pill-test.stories.tsx
@@ -1,7 +1,8 @@
-import React from "react";
+import React, { useState } from "react";
 import { action } from "@storybook/addon-actions";
-import Pill from "./pill.component";
+import Pill, { PillProps } from "./pill.component";
 import Box from "../box";
+import { MultiSelect, Option } from "../select";
 
 export default {
   title: "Pill/Test",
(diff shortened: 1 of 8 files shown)


src/components/pill/pill.component.tsx
```diff
@@ -89,7 +92,12 @@ export const Pill = ({
         <IconButton
           onClick={onDelete}
           data-element="close"
-          aria-label={ariaLabelOfRemoveButton}
+          // Use custom ariaLabelOfRemoveButton if provided, else fallback to locale string
+          aria-label={
+            ariaLabelOfRemoveButton && ariaLabelOfRemoveButton !== "remove pill"
```
tomdavies73: **suggestion**: I think this may be a tad too much complexity, if we just go with `ariaLabelOfRemoveButton || locale.pill.remove(children)`

src/components/pill/pill.mdx
```diff
@@ -90,3 +92,20 @@ Color variants for dark theme, to apply these set `isDarkBackground` to `true`.
 ### Pill
 
 <ArgTypes of={PillStories} />
+
+## Translation keys
```
tomdavies73, line 96: **suggestion**: I think we should also add this content to the translation keys table 

https://carbon.sage.com/?path=/docs/documentation-i18n--docs#where-to-find-translation-keys here too

src/components/pill/pill-test.stories.tsx
```diff
@@ -89,3 +90,58 @@ StatusDarkBackground.story = {
     colorVariant: "neutral",
   },
 };
+
+export const WithCustomAriaLabels = ({ ...args }: PillProps) => {
+  const noop = () => action("delete");
+
+  const [selectedPills, setSelectedPills] = useState(["1", "4"] as string[]);
```
Parsium: **nitpick**: do we need the type assertion here? It seems like TypeScript should be able to infer the type

```suggestion
  const [selectedPills, setSelectedPills] = useState(["1", "4"]);
```

PR #7527, opened 2025-09-11
test(date-range): change date to avoid incorrect snapshot change

DobroTora:
### Proposed behaviour
Story relies on `handleOnKeyDown`, so onBlur is not fired, when switching between two fields.

### Current behaviour
DateRange calls `onBlur` even when the user is just switching between the start and end inputs

### Checklist

#### QA

### DIFF
--- a/src/components/date-range/date-range-interactions.stories.tsx
+++ b/src/components/date-range/date-range-interactions.stories.tsx
@@ -304,6 +304,7 @@ export const TypingSyncDateRange: Story = {
   ),
   play: async ({ canvasElement }) => {
     if (!allowInteractions()) return;
+
     const canvas = within(canvasElement);
 
     const [startIcon] = canvas.getAllByTestId("icon");
@@ -312,11 +313,11 @@ export const TypingSyncDateRange: Story = {
 
     const startInput = canvas.getByRole("textbox", { name: /start date/i });
     await userEvent.clear(startInput);
-    await userEvent.keyboard("10/09/2025");
+    await userEvent.type(startInput, "05/05/2022");
     await userInteractionPause(300);
 
     const caption = canvas.getByRole("status");
-    expect(caption).toHaveTextContent(/September 2025/i);
+    expect(caption).toHaveTextContent(/May 2022/i);
   },
 };
 


src/components/date-range/date-range-interactions.stories.tsx
```diff
@@ -304,19 +304,21 @@ export const TypingSyncDateRange: Story = {
...
+    const portal = within(document.body);
 
     const [startIcon] = canvas.getAllByTestId("icon");
     await userEvent.click(startIcon);
     await userInteractionPause(250);
 
     const startInput = canvas.getByRole("textbox", { name: /start date/i });
     await userEvent.clear(startInput);
-    await userEvent.keyboard("10/09/2025");
+    await userEvent.type(startInput, "05/05/2022");
```
tomdavies73, line 316: **suggestion**" for the sake of consistency I think we should still use `.keyboard()` here as `.keyboard` is used for all the other tests

src/components/date-range/date-range-interactions.stories.tsx
```diff
@@ -304,19 +304,21 @@ export const TypingSyncDateRange: Story = {
   ),
   play: async ({ canvasElement }) => {
     if (!allowInteractions()) return;
+
     const canvas = within(canvasElement);
+    const portal = within(document.body);
```
Parsium: **comment**: it seems the issue we're observing with unexpected snapshot changes is caused by the start date happening to coincide with the current month as of writing, September 2025, since we highlight the current date.

Because of this, I don't think querying the document body is necessary here, as changing the start date month should be sufficient.

Can we just keep the modified start date?

src/components/date-range/date-range-interactions.stories.tsx
Parsium: Personally I think either `.type` or `.keyboard` is fine here, as it should achieve the same result

tomdavies73: Agreed @Parsium, I'd just argue for consistency’s sake it would be better to stick with `.keyboard`, especially as that's the approach elsewhere and the change from `keyboard` -> `type` isn't needed for this fix

DobroTora: Hello both, I would stick to keyboard as it targets the actual issue here I think.

src/components/date-range/date-range-interactions.stories.tsx
DobroTora: sure we can 👍

dariusbercea-sage (approved):

Issue #7518, opened 2025-09-08
Impossible to Set Custom Input Field id prop in NumeralDate Component for Accessibility


hagad:
### Description

In the context of accessibility, we need to be able to provide a custom id prop on the NumeralDate component.

Currently, there is no way to control the IDs of the three internal input fields. This prevents components like from rendering anchor links that correctly reference these input fields.

We need to be able to assign e ids to each input field so users can quickly navigate to the first invalid field when clicking on an error link.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-jdqypja1?file=src%2FApp.tsx

### Suggested solution

Allow passing a custom id

to the NumeralDate component

### Carbon version

156.3.2

### React version

v17

### Relevant browsers

Chrome

### Relevant OSs

MacOS

### Confidentiality

- [x] I confirm there is no confidential or commercially sensitive information included.

Parsium:
To provide context for those in triage, this pattern is commonly used in forms with multiple inputs, where an error summary message is displayed at the top of the form, with links to the individual invalid fields.

We should either surface the ability to control all three ids of each date field, or just the day field.

PR #7529, opened 2025-09-11
fix(numeral-date): add ability to set ID values against individual inputs

mihai-albu-sage:
fixes: #7518

### Proposed behaviour

Optional prop `inputIds` has been added which allows the user to provide custom IDs for each of the field inputs by specifying a value for the `day`, `month` and `year`. In the absence of these IDs, the component uses the random generated IDs, in order to keep backwards compatibility.

### Current behaviour

Currently there is no way for the user to specify custom ID values for the field inputs.

### Checklist

#### QA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [157.3.1](https://github.com/Sage/carbon/compare/v157.3.0...v157.3.1) (2025-09-23)
+
+### Bug Fixes
+
+* **link:** link audit changes ([7c6d54d](https://github.com/Sage/carbon/commit/7c6d54db52d601723752e0e140b387cff3b09ad5))
+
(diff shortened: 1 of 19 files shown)


src/components/numeral-date/numeral-date.component.tsx
```diff
@@ -60,6 +60,12 @@ export interface NumeralDateEvent {
   };
 }
 
+export interface InputIdsValue {
```
tomdavies73: **suggestion** (non-blocking): I think we could go with a different name which has more semantic meaning like `DateInputIds` etc.

tomdavies73 (dismissed):

Parsium (dismissed):

src/components/numeral-date/numeral-date.component.tsx
mihai-albu-sage: Good point on the naming, I’ve gone ahead and renamed the interface for clearer semantics.

Parsium (approved):

PR #7533, opened 2025-09-12
refactor: updates internal fieldset to render validation and hintText

nuria1110:
### Proposed behaviour

Internal Fieldset components now renders `HintText` and `ValidationMessage`, these have been removed from individual grouped components.
Components involved:
- `RadioButtonGroup`
- `CheckboxGroup`
-  `NumeralDate`
- `Time`

### Current behaviour

Currently, our grouped components import and render `HintText` and `ValidationMessage` individually, creating a lot of duplication and possible inconsistencies.  

### Checklist

#### QA

### Testing instructions

- There should be no major changes to any component and behaviour should be as expected. 
- Minor spacing changes in legend/hint text to be expected due to inconsistent styles applies across components. 
- Small accessibility improvement to `Time` as incorrect accessible description was being applied to the group. (see unit test updates).


src/components/time/time.component.tsx
```diff
@@ -315,136 +297,103 @@ const Time = React.forwardRef<TimeHandle, TimeProps>(
 
     return (
       <Fieldset
+        applyNewValidation
+        // because internalId is a combination of both input ids, we need to generate a new one as that was setting the incorrect accessible desc.
+        id={guid()}
```
Parsium: **suggestion**: could we update `internalId` to a different value? Perhaps to a new guid?:

```tsx
// line 140
const internalId = useRef(guid());
```

This would allow us to get round the accessible description issue, while ensuring a guid isn't re-generated on subsequent re-renders.

Parsium (commented):
Looks much neater! 👍🏼

src/components/time/time.component.tsx
nuria1110: I was weary of doing this since the component does not support passing a custom `id` prop to the group. Since it is a combination of the input ids, which consumers have the ability to pass, they are still able to set/access this value and I was thinking that removing this may create issues for consumers.

Parsium: Ah I see what you mean about the fieldset's id being a combination of the input ids.

We do have an issue with the fieldset id currently, as [it shouldn't contain whitespace](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id#syntax). What do you think about hyphenating the fieldset's id for the time being? That should allow the accessible description to be computed correctly, while fixing the invalid syntax?

In the future, I think we should remove setting a fieldset id as a breaking change (since the fieldset id is returned in `onChange`), as I don't think consumers should need to reference the fieldset element itself in practice, but rather the individual input fields.

damienrobson-sage (approved):

PR #7540, opened 2025-09-16
Fe 7470 breadcrumbs focus safari

DobroTora:
### Checklist

#### QA

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs.pw.tsx
+++ b/src/components/breadcrumbs/breadcrumbs.pw.tsx
@@ -166,41 +166,42 @@ test.describe("should render Breadcrumbs component", async () => {
     );
   });
 
-  test("should have correct color when Crumb is current and isDarkBackground is true", async ({
+  test("should have non-interactive cursor when Crumb is current and isDarkBackground is true", async ({
     mount,
     page,
   }) => {
     await mount(<OnDarkBackground />);
+    await page.waitForSelector('[data-component="breadcrumbs"]');
 
-    const currentCrumb = crumbAtIndex(page, 3).locator("a");
-    await expect(currentCrumb).toHaveCSS("color", "rgb(255, 255, 255)");
-    await currentCrumb.hover();
-    await expect(currentCrumb).toHaveCSS("color", "rgb(255, 255, 255)");
-    await expect(currentCrumb).toHaveCSS("cursor", "text");
+    const currentCrumb = crumbAtIndex(page, 3).locator(
+      'span[data-component="crumb"]',
+    );
+
+    await expect(currentCrumb).toBeVisible();
+    await expect(currentCrumb).toHaveAttribute("aria-current", "page");
+    await expect(currentCrumb).toHaveCSS("cursor", "auto");
   });
 
-  [
-    {
-      isCurrent: true,
-      expectedAttribute: "aria-current",
-      expectedValue: "page",
-    },
-    {
-      isCurrent: false,
-      expectedAttribute: "href",
-      expectedValue: "#",
-    },
-  ].forEach(({ isCurrent, expectedAttribute, expectedValue }) => {
-    test(`should check Crumb with isCurrent prop is ${isCurrent}`, async ({
-      mount,
-      page,
-    }) => {
-      await mount(<DefaultCrumb isCurrent={isCurrent} />);
-      await expect(crumbAtIndex(page, 0).locator("a")).toHaveAttribute(
-        expectedAttribute,
-        expectedValue,
-      );
-    });
+  test("should check Crumb with isCurrent prop is true", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<DefaultCrumb isCurrent={true} />);
+    await expect(crumbAtIndex(page, 0).locator("span")).toHaveAttribute(
+      "aria-current",
+      "page",
+    );
+  });
+
+  test("should check Crumb with isCurrent prop is false", async ({
+    mount,
+    page,
+  }) => {
+    await mount(<DefaultCrumb isCurrent={false} />);
+    await expect(crumbAtIndex(page, 0).locator("a")).toHaveAttribute(
+      "href",
+      "#",
+    );
   });
 
   test("should check Crumb with href prop set to default val", async ({
@@ -222,13 +223,14 @@ test("when Crumb's isCurrent prop is true, Crumb divider should not exist", asyn
   await mount(<DefaultCrumb isCurrent />);
 
   const crumbElement = crumbAtIndex(page, 0);
-  await expect(crumbElement.locator("a")).toHaveAttribute(
+
+  await expect(crumbElement.locator("span")).toHaveAttribute(
     "aria-current",
     "page",
   );
-  await expect(crumbElement.locator("span").nth(1)).toHaveCSS(
-    "color",
-    "rgba(0, 0, 0, 0.9)",
+
+  await expect(crumbElement.locator('[data-role="crumb-divider"]')).toHaveCount(
+    0,
   );
 });
 
--- a/src/components/breadcrumbs/crumb/crumb.component.tsx
+++ b/src/components/breadcrumbs/crumb/crumb.component.tsx
@@ -24,26 +24,52 @@ export interface CrumbProps
       | "disabled"
     >,
     TagProps {
-  /** This sets the Crumb to current, does not render Link */
   isCurrent?: boolean;
+  onClick?: (
+    ev: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>,
+  ) => void;
 }
 
-const Crumb = React.forwardRef<HTMLAnchorElement, CrumbProps>(
-  ({ href, isCurrent, children, onClick, ...rest }: CrumbProps, ref) => {
+const Crumb = React.forwardRef<HTMLLIElement, CrumbProps>(
+  ({ href, isCurrent, children, onClick, ...rest }, ref) => {
     const { isDarkBackground } = useBreadcrumbsContext();
 
+    const isSafari = React.useMemo(() => {
+      if (typeof navigator === "undefined") return false;
+      return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    }, []);
+
+    const handleClick = React.useCallback(
+      (
+        event: React.MouseEvent<
+          HTMLAnchorElement | HTMLButtonElement,
+          MouseEvent
+        >,
+      ) => {
+        if (!isCurrent && isSafari) {
+          (event.currentTarget as HTMLElement).focus({ preventScroll: true });
+        }
+        onClick?.(event);
+      },
+      [isCurrent, isSafari, onClick],
+    );
+
     return (
-      <li>
+      <li ref={ref}>
         <StyledCrumb
-          ref={ref}
           isCurrent={isCurrent}
           aria-current={isCurrent ? "page" : undefined}
           isDarkBackground={isDarkBackground}
+          data-testid={isCurrent ? "current-crumb" : "link-anchor"}
           {...rest}
           {...tagComponent("crumb", rest)}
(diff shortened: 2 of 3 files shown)


src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,26 +24,52 @@ export interface CrumbProps
...
+          (event.currentTarget as HTMLElement).focus({ preventScroll: true });
+        }
+        onClick?.(event);
+      },
+      [isCurrent, isSafari, onClick],
+    );
+
     return (
-      <li>
+      <li ref={ref}>
```
edleeks87: question: why is the `ref` being updated here?

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,26 +24,52 @@ export interface CrumbProps
...
+  ) => void;
 }
 
-const Crumb = React.forwardRef<HTMLAnchorElement, CrumbProps>(
-  ({ href, isCurrent, children, onClick, ...rest }: CrumbProps, ref) => {
+const Crumb = React.forwardRef<HTMLLIElement, CrumbProps>(
+  ({ href, isCurrent, children, onClick, ...rest }, ref) => {
     const { isDarkBackground } = useBreadcrumbsContext();
 
+    const isSafari = React.useMemo(() => {
```
edleeks87: comment: We should avoid browser sniffing, having this work on all browsers isn't going to break anything etc

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,26 +24,52 @@ export interface CrumbProps
       | "disabled"
     >,
     TagProps {
-  /** This sets the Crumb to current, does not render Link */
   isCurrent?: boolean;
+  onClick?: (
```
edleeks87: comment: the interface already extends `LinkProps` which has `onClick` defined so you don't need this it can just be destructured like you have below

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,26 +24,52 @@ export interface CrumbProps
...
+
+    const handleClick = React.useCallback(
+      (
+        event: React.MouseEvent<
+          HTMLAnchorElement | HTMLButtonElement,
+          MouseEvent
+        >,
+      ) => {
+        if (!isCurrent && isSafari) {
+          (event.currentTarget as HTMLElement).focus({ preventScroll: true });
```
edleeks87: suggestion: what you have is fine but if you use the ref (as long as it's passed to the anchor like before) you want have to cast the type to `HTMLElement`

```suggestion
          ref.current?.focus({ preventScroll: true });
```

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,26 +24,52 @@ export interface CrumbProps
...
+
     return (
-      <li>
+      <li ref={ref}>
         <StyledCrumb
-          ref={ref}
           isCurrent={isCurrent}
           aria-current={isCurrent ? "page" : undefined}
           isDarkBackground={isDarkBackground}
+          data-testid={isCurrent ? "current-crumb" : "link-anchor"}
```
edleeks87: comment: we already override the locator to use `data-role` in our unit tests so we don't need to use `data-testid`

src/components/breadcrumbs/crumb/crumb.test.tsx
```diff
@@ -27,27 +27,80 @@ test("passes href to the anchor element when isCurrent is false", () => {
...
+    </Breadcrumbs>,
+  );
 
-  expect(anchor).not.toHaveAttribute("href", "foo");
+  expect(screen.getByTestId("baz")).toHaveAttribute("data-element", "bar");
 });
 
-test("calls onClick callback when the crumb link is clicked", async () => {
+test("calls focus({ preventScroll: true }) on Safari when not current", async () => {
+  const originalUA = window.navigator.userAgent;
```
edleeks87: comment: we can just remove this and assert it's called no matter the browser

src/components/breadcrumbs/crumb/crumb.test.tsx
```diff
@@ -57,14 +110,31 @@ test("calls onClick callback when the crumb link is clicked", async () => {
...
+
+  focusSpy.mockRestore();
+  Object.defineProperty(window.navigator, "userAgent", {
+    value: originalUA,
+    configurable: true,
+  });
 });
 
-test("does not call onClick callback when isCurrent is true", async () => {
+test("does not render a link or call onClick on Safari when current", async () => {
```
edleeks87: comment: this test isn't needed if we remove the browser sniffing

src/components/breadcrumbs/crumb/crumb.test.tsx
```diff
@@ -73,20 +143,57 @@ test("does not call onClick callback when isCurrent is true", async () => {
...
+        Current Page
+      </Crumb>
+    </Breadcrumbs>,
+  );
+
+  const el = screen.getByText("Current Page");
+  expect(el).toHaveAttribute("aria-current", "page");
+});
+
+test("isSafari returns false when navigator is undefined (no preventScroll focus call)", () => {
```
edleeks87: Comment: this can also be removed

edleeks87 (commented):
Hey @DobroTora this is pretty close to what we need, I think we can just remove browser sniffing and then some small other tweaks and I'm happy with it 😄

src/components/breadcrumbs/breadcrumbs.pw.tsx
```diff
@@ -166,41 +166,42 @@ test.describe("should render Breadcrumbs component", async () => {
     );
   });
 
-  test("should have correct color when Crumb is current and isDarkBackground is true", async ({
```
edleeks87, line 169: question: have we got this captured anywhere in chromatic atm? If not I think we should keep this until we do

src/components/breadcrumbs/crumb/crumb.component.tsx
DobroTora: It shouldn't be here, but on a` StyledCrumb`

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,43 +24,54 @@ export interface CrumbProps
...
+  /** This sets the Crumb to current, does not render Link */
   isCurrent?: boolean;
-  onClick?: (
-    ev: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>,
-  ) => void;
 }
 
-const Crumb = React.forwardRef<HTMLLIElement, CrumbProps>(
-  ({ href, isCurrent, children, onClick, ...rest }, ref) => {
+function mergeRefs<T>(...refs: Array<React.Ref<T> | undefined>) {
```
edleeks87: comment: we don't need this we can do something like

```
    const [internalRef, setInternalRef] = useState<HTMLAnchorElement | null>(null);
    const setRefs = React.useCallback(
      (reference: HTMLAnchorElement) => {
        setInternalRef(reference);
        if (!ref) return;
        if (typeof ref === "object") ref.current = reference;
          ref(reference);
        }
      },
      [ref],
    );

    const handleClick = (event: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) => {
      internalRef?.focus();
      onClick?.(event as React.MouseEvent<HTMLAnchorElement>);
    };

...

        <StyledCrumb
          ref={setRefs}
          isCurrent={isCurrent}
          aria-current={isCurrent ? "page" : undefined}
          isDarkBackground={isDarkBackground}
          {...rest}
          {...tagComponent("crumb", rest)}
          {...(!isCurrent && {
            href,
            onClick: handleClick,
          })}
        >
          {children}
        </StyledCrumb>

File: src/components/breadcrumbs/crumb/crumb.component.tsx [hunk 11]
```diff
@@ -24,43 +24,54 @@ export interface CrumbProps
...
-      return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    const focusCrumb = React.useCallback(() => {
+      try {
+        crumbRef.current?.focus?.({ preventScroll: true } as FocusOptions);
+      } catch {
+        crumbRef.current?.focus?.();
+      }
     }, []);
 
+    type LinkOnClick = NonNullable<LinkProps["onClick"]>;
```
edleeks87: comment: we don't need these types, they also shouldn't be initialised in a function body

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -24,43 +24,54 @@ export interface CrumbProps
...
+
+const Crumb = React.forwardRef<HTMLElement, CrumbProps>(
+  ({ href, isCurrent, children, onClick, ...rest }, forwardedRef) => {
     const { isDarkBackground } = useBreadcrumbsContext();
+    const crumbRef = React.useRef<HTMLElement | null>(null);
 
-    const isSafari = React.useMemo(() => {
-      if (typeof navigator === "undefined") return false;
-      return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    const focusCrumb = React.useCallback(() => {
```
edleeks87: comment: we don't need this see suggestion above

edleeks87: This can be moved into the onClick, creating a useCallback function with no dependency array is just wasting memory as it will never update

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -28,47 +28,32 @@ export interface CrumbProps
...
-    });
-  };
-}
-
 const Crumb = React.forwardRef<HTMLElement, CrumbProps>(
   ({ href, isCurrent, children, onClick, ...rest }, forwardedRef) => {
     const { isDarkBackground } = useBreadcrumbsContext();
     const crumbRef = React.useRef<HTMLElement | null>(null);
 
+    React.useImperativeHandle(forwardedRef, () => crumbRef.current, []);
```
edleeks87: Comment: this isn't the way to use imperative handle (I'm not convinced the TS types will allow it either way), I've already recommended an approach that will work here

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -28,47 +28,32 @@ export interface CrumbProps
...
-      if (typeof ref === "function") {
-        ref(node);
-      } else {
-        (ref as React.MutableRefObject<T | null>).current = node;
-      }
-    });
-  };
-}
-
 const Crumb = React.forwardRef<HTMLElement, CrumbProps>(
```
edleeks87: Question: why is this now `HTMLElement` it needs to be `HTMLAnchorElement`

edleeks87 (changes requested):
We can't assert it's not null with `!` and it looks like some comments haven't been addressed

src/components/breadcrumbs/crumb/crumb.component.tsx
```diff
@@ -33,7 +33,7 @@ const Crumb = React.forwardRef<HTMLElement, CrumbProps>(
     const { isDarkBackground } = useBreadcrumbsContext();
     const crumbRef = React.useRef<HTMLElement | null>(null);
 
-    React.useImperativeHandle(forwardedRef, () => crumbRef.current, []);
+    React.useImperativeHandle(forwardedRef, () => crumbRef.current!, []);
```
edleeks87: Comment: We shouldn't be doing this, we can't use `!` to assert it's not null as it might be given the type above. I think the better approach is to avoid an imperative handle we would then need to surface the `focus` and `blur` functions as part of it that would be a breaking change as we restricting what the implementing teams are able to do with the ref they pass

PR #7541, opened 2025-09-16
    feat: ensure popover based components cannot remain open when an adaptive sidebar renders as a modal

tomdavies73:
### Proposed behaviour

Due to a custom blur event which has been added to `AdaptiveSidebar` and listened to in:

1. `ActionPopover`
2. `FilterableSelect`
3. `MultiActionButton`
4. `MultiSelect`
5. `SplitButton` 
6. `SimpleSelect`

these components can no longer stay open when an `AdaptiveSidebar` is conditionally rendered as a modal

The `restoreFocusOnClose` prop has also been added to the `AdaptiveSidebar` interface to provide consumers with control over where focus is returned to then the `AdaptiveSidebar` renders from modal to a div.

### Current behaviour

The adaptive sidebar can be conditionally rendered as a modal and popover based components (listed above) remain open

### Checklist

#### QA

### Testing instructions

The `With Dropdown` test story can be used to play around with a select being open and an adaptive sidebar rendering between `div` -> `modal` based on screen re-size


src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -82,14 +86,35 @@ export const AdaptiveSidebar = ({
...
+   * all elements which could be open are closed. */
+  useEffect(() => {
+    if (renderModal) {
+      setTimeout(() => {
+        /* Due to component design elsewhere, directly blurring the previously
+         * active element, and handling said blur can cause issues.
+         * By creating and handling a custom blur event instead we can better control
+         * behaviour and avoid unintended issues. */
+        document.dispatchEvent(
+          new CustomEvent("adaptiveSidebarBlur", {
```
edleeks87: Comment: I'm not sure calling it blur is right here as it's a focus event etc. I think calling it something like `adaptiveSidebarModalFocusIn` (rolls of the tongue I know) would be more intuitive to what it happening etc

src/components/multi-action-button/multi-action-button.component.tsx
```diff
@@ -101,6 +102,21 @@ export const MultiActionButton = forwardRef<
       }
     };
 
+    useEffect(() => {
```
edleeks87: comment: as this is duplicated a fair few time I think it's worth making an internal custom hook that takes a callback argument that calls the set state etc

eg:

```
const useAdaptiveSidebarModalFocus = (closeFn: Dispatch<SetStateAction<boolean>>) => {
    useEffect(() => {
 const handleBlur = () => closeFn();

      document.addEventListener(
        "adaptiveSidebarBlur",
        handleBlur,
      );
      return () => {
        document.removeEventListener(
          "adaptiveSidebarBlur",
          handleBlur,
        );
      };
}, [closeFn]);
}

File: src/components/select/filterable-select/filterable-select.component.tsx [hunk 3]
```diff
@@ -564,6 +564,21 @@ export const FilterableSelect = React.forwardRef<
       }
     }
 
+    useEffect(() => {
```
edleeks87: comment: if you use a custom hook this would be:

```
useAdaptiveSidebarModalFocus(() => setOpen(false));

File: src/components/adaptive-sidebar/adaptive-sidebar.component.tsx [hunk 4]
```diff
@@ -82,14 +86,35 @@ export const AdaptiveSidebar = ({
     });
   }, [hidden, open]);
 
-  if (renderAsModal || !largeScreen) {
+  /* When the adaptive sidebar is conditionally rendered as a modal
+   * we need to blur the previously active element to ensure
+   * all elements which could be open are closed. */
+  useEffect(() => {
+    if (renderModal) {
```
edleeks87, line 93: question: below we're rendering a modal if `renderModal || !largeScreen` do we not need to match that here?

damienrobson-sage (commented):
If I open the "main" dropdown in the `Adaptive Sidebar > Test > With Dropdown` and resize down to invoke the modal, it closes (which is good). If I go back to non-modal again, the dropdown reopens again (which though helpful is probably bad)

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
```diff
@@ -50,9 +52,11 @@ export const AdaptiveSidebar = ({
   open,
   renderAsModal = false,
   width = "320px",
+  restoreFocusOnClose,
```
edleeks87: comment: this defaults to true in Dialog fyi, I think it should be false in this component but worth UX  input on it

tomdavies73: I believe this should be defaulted to `false` due to the nature of the component and my prev chat with Hannah. Will default this to `false`

src/components/adaptive-sidebar/adaptive-sidebar.component.tsx
tomdavies73: I've added a const `renderModal` with has a value of `renderAsModal || !largeScreen` to avoid repetition

edleeks87: ah sorry I missed that

src/components/adaptive-sidebar/components.test-pw.tsx
```diff
@@ -30,6 +30,9 @@ export const DefaultAdaptiveSidebar = ({ ...props }) => {
           </Typography>
         </Box>
         <AdaptiveSidebar open={adaptiveSidebarOpen} {...props}>
+          {adaptiveSidebarOpen && (
```
edleeks87: question (non-blocking): why are we conditionally rendering this?

edleeks87 (dismissed):

edleeks87 (approved):

PR #7543, opened 2025-09-17
build(lint): enable recommended rules for playwright

Parsium:
### Proposed behaviour

- Enable the recommended rules from the Playwright ESLint plugin and fix immediate errors.
- Set a maximum threshold on the number of permitted linting warnings.
- Re-enable the linting rule for forbidding null assertion operators in TypeScript.

### Current behaviour

- Our Playwright tests contain several anti-patterns that don't adhere to recommendations from the Playwright community. Enabling the recommended linting rules results in many new errors and warnings.

### Checklist

#### QA

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -346,6 +346,7 @@ export default defineConfig([
     },
   },
   {
+    ...playwright.configs["flat/recommended"],
     files: ["**/*.pw.ts*", "**/*.test-pw.ts*", "playwright/**/*.ts*"],
     plugins: {
       playwright: fixupPluginRules(playwright),
@@ -362,6 +363,24 @@ export default defineConfig([
       },
     },
     rules: {
+      ...playwright.configs["flat/recommended"].rules,
+      "playwright/expect-expect": [
+        "error",
+        {
+          assertFunctionNames: [
+            "checkAccessibility",
+            "checkCSSOutline",
+            "checkGoldenOutline",
+            "containsClass",
+            "assertCssValueIsApproximately",
+            "verifyRequiredAsterisk",
+            "verifyRequiredAsteriskForLabel",
+            "verifyRequiredAsteriskForLegend",
+            "checkElementBorderColours",
+            "checkNewFocusStyling",
+          ],
+        },
+      ],
       "playwright/no-commented-out-tests": "error",
       "playwright/no-focused-test": "error",
       "playwright/no-skipped-test": "warn",
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -271,6 +271,7 @@ export default defineConfig([
           allow: ["arrowFunctions"],
         },
       ],
+      "@typescript-eslint/no-non-null-assertion": "error",
       "@typescript-eslint/no-unused-vars": [
         "error",
         {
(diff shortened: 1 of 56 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "test-update": "jest --config=./jest.config.ts --updateSnapshot",
     "test-storybook": "test-storybook --url http://127.0.0.1:9001",
     "format": "prettier --write './{src,playwright}/**/*.{js,jsx,ts,tsx}'",
-    "lint": "eslint ./src ./playwright",
+    "lint": "eslint ./src ./playwright --max-warnings=636 --report-unused-disable-directives",
```
damienrobson-sage, line 16: **question**: why 636...?

eslint.config.mjs
```diff
@@ -271,6 +271,7 @@ export default defineConfig([
           allow: ["arrowFunctions"],
         },
       ],
+      "@typescript-eslint/no-non-null-assertion": "error",
```
tomdavies73, line 274: **question**: was there any particular reason why this was added as part of this change? I agree it should be there, I'm just thinking if it's not in scope we could move it to a different commit

damienrobson-sage (approved):

package.json
Parsium: After these changes, we have 636 lint warnings in total. The threshold will ensure no further warnings are introduced.

damienrobson-sage: Good enough for me, just checking it wasn't a magic number situation 😄

eslint.config.mjs
```diff
@@ -361,6 +363,24 @@ export default defineConfig([
       },
     },
     rules: {
+      ...playwright.configs["flat/recommended"].rules,
+      "playwright/expect-expect": [
+        "error",
+        {
+          assertFunctionNames: [
```
tomdavies73, line 370: **question**: Have we missed any functions here, just looking around I can see `assertCssValueIsApproximately` `isInViewport`, `waitForElementFocus` etc. I might be misunderstanding the critera for what is needed in here tbf

tomdavies73 (approved):

eslint.config.mjs
Parsium: I've updated our linting script so that any old disabled directive comments are automatically removed if they no longer apply. 

After running the lint script, several `no-non-null-assertion` disable directives were removed, which I didn't feel comfortable with, given that, in my opinion, this rule is important for enforcing type-safe code, and we'd be losing visibility on areas where it has been disabled.

eslint.config.mjs
Parsium: This rule flags any tests that don't contain assertions. Here's some examples of [tests with and without them](https://github.com/mskelton/eslint-plugin-playwright/blob/main/docs/rules/expect-expect.md). `isInViewport` and `waitForElementFocus` aren't actually asserting any behaviour here, so we don't need them specified as exemptions for this rule.

tomdavies73: Nice one @Parsium, just needed that extra bit of context

eslint.config.mjs
tomdavies73: Makes sense, happy with it staying in as part of this commit 👍

PR #7547, opened 2025-09-23
test(button): amend existing test to prevent focus being inherited

paulrobinson88:
### Proposed behaviour
Correct an issue where focus and hover states from the 'Focus and Hover States' Play Function are inherited in the next story which is opened.

https://github.com/user-attachments/assets/d21d2a98-8b97-48f9-af5e-e37ece942e12

### Current behaviour
Once the hover and focus is applied for the Play Function the next story opened has the same focus and hover state applied.

https://github.com/user-attachments/assets/4c10cca3-adfb-408e-8190-36a5f58d8446

### Checklist

#### QA

### Testing instructions
Check that the focus and hover state does not apply to the next story opened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [157.3.2](https://github.com/Sage/carbon/compare/v157.3.1...v157.3.2) (2025-09-24)
+
+### Bug Fixes
+
+* **numeral-date:** add ability to set ID values against individual inputs ([b9751a5](https://github.com/Sage/carbon/commit/b9751a55b4596da58f521950397020ed153a7a54)), closes [#7518](https://github.com/Sage/carbon/issues/7518)
+
+## [157.3.1](https://github.com/Sage/carbon/compare/v157.3.0...v157.3.1) (2025-09-23)
+
+### Bug Fixes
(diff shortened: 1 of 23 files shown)


src/components/button/button-interactions.stories.tsx
```diff
@@ -168,8 +201,8 @@ export const FocusAndHoverStates: Story = {
 FocusAndHoverStates.storyName = "Focus and Hover States";
 FocusAndHoverStates.parameters = {
   pseudo: {
-    hover: "button",
-    focus: "button",
+    hover: '[data-testid="target"]',
+    focus: '[data-testid="target"]',
```
Parsium: **suggestion(non-blocking)**: would we be able to use `Button`'s `data-role` prop instead? 

I'm a bit puzzled why TypeScript hasn't flagged this, as `data-testid` isn't a prop the component supports. Might be something separate for us to look into :)

edleeks87: data- props are implicitly supported fyi @Parsium. Do agree lets use data-role though for consistency

dariusbercea-sage (approved):

PR #7555, opened 2025-09-26
fix(link): enable focus on Safari

DobroTora:
### Proposed behaviour

Enable focus on Safari for Link component

### Current behaviour
Focus is disabled on Safari 

### Checklist

#### QA

### DIFF
--- a/src/components/link/link.component.tsx
+++ b/src/components/link/link.component.tsx
@@ -1,4 +1,11 @@
-import React, { useContext, useEffect, useMemo, useState } from "react";
+import React, {
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+  useRef,
+  useCallback,
+} from "react";
 
 import Icon, { IconType } from "../icon";
 import MenuContext from "../menu/__internal__/menu.context";
@@ -113,6 +120,7 @@ export const Link = React.forwardRef<
     const { inMenu } = useContext(MenuContext);
     const { batchSelectionDisabled } = useContext(BatchSelectionContext);
     const isDisabled = disabled || batchSelectionDisabled;
+    const linkRef = useRef<HTMLAnchorElement | HTMLButtonElement | null>(null);
 
     if (!deprecatedDisabledWarning && disabled) {
       deprecatedDisabledWarning = true;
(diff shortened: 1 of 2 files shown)


src/components/link/link.component.tsx
```diff
@@ -152,17 +160,38 @@ export const Link = React.forwardRef<
...
+          ).current = reference;
         if (typeof ref === "function") {
-          ref(reference);
+          ref(reference as HTMLAnchorElement & HTMLButtonElement);
         }
       },
       [ref],
     );
 
+    const focusLink = useCallback(() => {
```
edleeks87: comment: we don't need a `focusLink` callback. By using the hook with an empty array we are only memoising it at the point the component mounts and no more so it wastes memory. We also don't need the types, if we did they shouldn't be defined inside the component but at the top of the file. Something like the below will be enough here


```
const handleClick = useCallback(
      (event:
      | React.MouseEvent<HTMLAnchorElement>
      | React.MouseEvent<HTMLButtonElement>) => {
        linkRef.current?.focus?.({ preventScroll: true });
        onClick?.(event);
      },
      [onClick],
    );

File: src/components/link/link.component.tsx [hunk 2]
```diff
@@ -152,17 +160,38 @@ export const Link = React.forwardRef<
     const effectiveInverse = inverse ?? isDarkBackground;
     const effectiveVariant = variantAlias[variant] ?? variant;
 
-    const setRefs = React.useCallback(
-      (reference: HTMLAnchorElement) => {
+    const setRefs = useCallback(
```
edleeks87, line 163: comment: we don't need the type coercion here

```
 const setRefs = useCallback(
      (reference: HTMLAnchorElement | HTMLButtonElement | null) => {
        linkRef.current = reference;
        if (!ref) return;
        if (typeof ref === "object") ref.current = reference;
        if (typeof ref === "function") ref(reference);
      },
      [ref],
    );

Comment (edleeks87) [reply src/components/link/link.component.tsx:163]: We don't know that the ref is an object, so it's risky to assign it to current without checking

File: src/components/link/link.component.tsx [hunk 3]
```diff
@@ -164,32 +164,26 @@ export const Link = React.forwardRef<
...
-            ref as React.MutableRefObject<
-              HTMLAnchorElement | HTMLButtonElement | null
-            >
-          ).current = reference;
+
         if (typeof ref === "function") {
-          ref(reference as HTMLAnchorElement & HTMLButtonElement);
+          ref(reference);
+        } else {
+          ref.current = reference;
```
edleeks87, line 173: see comment about checking if it's an object

edleeks87 (approved):

damienrobson-sage (approved):

stephenogorman (approved):

PR #7557, opened 2025-09-30
feat(textbox): align component with fusion DS

tomdavies73:
### Proposed behaviour

Aligns Textbox with new fusion designs.
The following props will no longer work:
```
  adaptiveLabelBreakpoint,
  info,
  helpAriaLabel,
  labelAlign,
  labelWidth,
  labelSpacing,
  labelId,
  reverse,
  tooltipPosition,
  tooltipId,
  validationOnLabel
```

### Current behaviour

Textbox is not aligned with fusion spec

### Checklist

#### QA


src/__internal__/validation-message/__next__/validation-message.style.ts
```diff
@@ -0,0 +1,38 @@
...
+      flex-wrap: wrap;
+      margin: 0px;
+
+      margin: 0px;
+
+      color: ${isError
+        ? "var(--input-validation-label-error)"
+        : "var(--input-validation-label-warn)"};
+
+      font-family: var(--fontFamiliesDefault);
```
edleeks87: question: why aren't we using the new tokens for the font?

edleeks87: comment: The font tokens contain everything already as far as I know so you could just set `font` unless I'm missing something, you might need to pick the relevant one based on size and error vs warning though (worth checking tokens with DS/UX though)

```
   --global-font-static-comp-medium-xs: 500 13px/1.5 Sage UI;
  --global-font-static-comp-medium-s: 500 14px/1.5 Sage UI;
  --global-font-static-comp-medium-m: 500 14px/1.5 Sage UI;
  --global-font-static-comp-medium-l: 500 16px/1.5 Sage UI;
  --global-font-static-comp-regular-xs: 400 13px/1.5 Sage UI;
  --global-font-static-comp-regular-s: 400 14px/1.5 Sage UI;
  --global-font-static-comp-regular-m: 400 14px/1.5 Sage UI;
  --global-font-static-comp-regular-l: 400 16px/1.5 Sage UI;
```

```
font-family: var(--fontFamiliesDefault);
font-size: ${isLarge
        ? "var(--global-font-static-body-regular-l)"
        : "var(--global-font-static-body-regular-m)"};

font-style: normal;
font-weight: ${isError ? "500" : "400"};
line-height: 150%;
```
can be

```
${isLarge && css`
  font: ${isError ? "var(--global-font-static-comp-medium-l)" : ""var(--global-font-static-comp-regular-l)"};
`}
${!isLarge && css`
     font: ${isError ? "var(--global-font-static-comp-medium-m)" : ""var(--global-font-static-comp-regular-m)"};
`}

File: src/components/textbox/__next__/__internal__/error-border/error-border.style.ts [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+    position: absolute;
+    left: -10px;
+    top: 0px;
+    bottom: 0px;
+    z-index: 6;
+    background-color: ${warning
+      ? "var(--input-validation-bar-warn)"
+      : "var(--input-validation-border-error)"};
+  `}
+  transform: scaleX(1);
```
edleeks87: question: is this transform doing anything here as it's scaling to 100% with no animation?

src/components/textbox/__next__/__internal__/hint-text/hint-text.style.ts
```diff
@@ -0,0 +1,26 @@
...
+  disabled?: boolean;
+}
+
+const StyledHintText = styled.span<StyledHintTextProps>`
+  color: ${({ disabled }) =>
+    disabled
+      ? "var(--input-labelset-label-disabled)"
+      : "var(--input-labelset-label-alt)"};
+
+  font-family: var(--fontFamiliesDefault);
```
edleeks87: Comment: same as above wrt to using `font` and the token etc

src/components/textbox/__next__/__internal__/input/input.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ...props
+    },
+    ref,
+  ) => {
+    const localRef = useRef<HTMLInputElement>(null);
+    const inputRef = ref || localRef;
+    const stateProps = { disabled, readOnly };
+
+    useEffect(() => {
+      if (autoFocus && inputRef && "current" in inputRef) {
```
edleeks87: suggestion (non-blocking): what you have is fine (don't need to check inputRef as if current is in it it's truthy) but we've done the following elsewhere in the codebase so I think for consistency sake it's worthwhile adopting it

```
 const setRefs = (node: HTMLInputElement | null) => {
      internalRef.current = node;
        ref(node);
      } else if (ref) {
        (ref as React.MutableRefObject<HTMLInputElement | null>).current = node;
      }
    };
  
    const stateProps = { disabled, readOnly };

    useEffect(() => {
      if (autoFocus) {
        internalRef.current?.focus();
      }
    }, [autoFocus]);

Comment (edleeks87) [reply src/components/textbox/__next__/__internal__/input/input.component.tsx]: suggestion (non-blocking): we could actually make this a custom (internal) hook so it's DRY and reuse the pattern where needed

```
// useCombinedRefs

import {MutableRefObject, RefCallback} from "react";

export default function combineRefs<T = unknown>(
  ...refs: (MutableRefObject<T | null> | RefCallback<T> | null | undefined)[]
): RefCallback<T> {
  return (node: T | null) => {
    refs.forEach((ref) => {
      if (!ref) return;

        ref(node);
      } else {
        ref.current = node;
      }
    });
  };
};
```

```
// input.component
 
...

const combinedRef = useCombinedRefs(internalRef, ref);

...

useEffect(() => {
  if (autoFocus) {
      internalRef.current?.focus();
   }
}, [autoFocus]);

...

<input type="text" ref={combinedRef} {...stateProps} {...props} />

src/__internal__/error-border/error-border.style.ts
edleeks87, line 1: comment (non-blocking): if this is the same for all components I wonder if it's more appropriate in the shared `_internal_` directory (still as a `_next_`)

src/components/textbox/__next__/__internal__/input/input.style.ts
```diff
@@ -0,0 +1,123 @@
...
+    z-index: 2;
+  }
+
+  [data-role="input-text-container"] {
+    display: flex;
+
+    padding: ${({ size = "medium" }) => {
+      switch (size) {
+        case "small":
+          return "0 var(--global-space-comp-s)";
```
edleeks87:
```suggestion
          return "var(--global-space-none) var(--global-space-comp-s)";
```

worth confirming with DS first though

src/components/textbox/__next__/__internal__/input/input.style.ts
```diff
@@ -0,0 +1,123 @@
...
+        case "small":
+          return "var(--global-space-comp-xs)";
+        case "large":
+          return "var(--global-space-comp-m)";
+        default:
+          return "var(--global-space-comp-s)";
+      }
+    }};
+
+    flex: 1 0 0;
```
edleeks87: suggestion (non-blocking): purely for readability it might be better to have the flex styling with the `display` flex above, ie move padding below this just so the related parts are grouped etc

src/components/textbox/__next__/__internal__/label/label.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+  disabled,
+  readOnly,
+}: LabelProps) => {
+  return (
+    <StyledLabel
+      id={id}
+      htmlFor={htmlFor}
+      isLarge={isLarge}
+      isRequired={isRequired}
+      disabled={disabled}
```
edleeks87:
```suggestion
      $disabled={disabled}
```

without it's setting the disabled attribute on the label. Might be worth adding the`$` to all of them bar `id` and `htmlFor` as they're the only ones valid on `label` elements

src/components/textbox/__next__/__internal__/label/label.style.ts
```diff
@@ -0,0 +1,55 @@
+import styled, { css } from "styled-components";
+
+interface StyledLabelProps {
```
edleeks87: suggestion:

```
interface StyledLabelProps {
  $isLarge?: boolean;
  $isRequired?: boolean;
  $disabled?: boolean;
  $readOnly?: boolean;
}

File: src/components/textbox/__next__/__internal__/label/label.style.ts [hunk 9]
```diff
@@ -0,0 +1,55 @@
...
+    if (disabled) {
+      return "var(--input-labelset-label-disabled)";
+    }
+    if (readOnly) {
+      return "var(--input-labelset-label-readOnly)";
+    }
+    return "var(--input-labelset-label-default)";
+  }};
+
+  font-family: var(--fontFamiliesDefault);
```
edleeks87: same comment wrt to font token above

src/components/textbox/__next__/__internal__/hint-text/hint-text.style.ts
```diff
@@ -0,0 +1,26 @@
+import styled from "styled-components";
+
+interface StyledHintTextProps {
```
edleeks87: suggestion:

```
interface StyledHintTextProps {
  $isLarge?: boolean;
  $disabled?: boolean;
}

File: src/__internal__/validation-message/__next__/validation-message.style.ts [hunk 11]
```diff
@@ -0,0 +1,38 @@
+import styled, { css } from "styled-components";
+
+interface StyledValidationMessageProps {
```
edleeks87, line 3: suggestion (non-blocking): these won;'t be added to the DOM element so up to you

```
interface StyledValidationMessageProps {
  $isError?: boolean;
  $isDarkBackground?: boolean;
  $isLarge?: boolean;
}

File: src/components/textbox/__next__/__internal__/input/input.style.ts [hunk 12]
```diff
@@ -0,0 +1,123 @@
+import styled from "styled-components";
+import addFocusStyling from "../../../../../style/utils/add-focus-styling";
+
+interface InputContainerProps {
```
edleeks87: suggestion: disabled needs doing rest are up to you

```
interface InputContainerProps {
  $error?: boolean;
  $inputWidth?: number;
  disabled?: boolean;
  readOnly?: boolean;
  $size?: "small" | "medium" | "large";
}

File: src/components/textbox/__next__/__internal__/input/input.component.tsx [hunk 13]
```diff
@@ -0,0 +1,85 @@
...
+  /** Specify a callback triggered on focus */
+  onFocus?: (ev: React.FocusEvent<HTMLInputElement>) => void;
+  /** Placeholder string to be displayed in input */
+  placeholder?: string;
+  /** If true, the component will be read-only */
+  readOnly?: boolean;
+  /** Flag to configure component as mandatory */
+  required?: boolean;
+  /** The value of the input */
+  value: string | readonly string[] | number | undefined;
```
edleeks87: question: if value is controlled do we need `undefined`?

src/components/textbox/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+  ) => {
+    const validationState = { error, warning };
+    const stateProps = { disabled, readOnly };
+
+    const [uniqueId, uniqueName] = useUniqueId(id, name);
+
+    const { labelId, validationId, ariaDescribedBy } = useInputAccessibility({
+      id: uniqueId,
+      validationRedesignOptIn: true,
+      error: validationState.error,
```
edleeks87: suggestion: you can just spread the object here
```
      ...validationState,
```

src/components/textbox/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+    const validationMessage = (
+      <ValidationMessage
+        id={validationId}
+        isLarge={size === "large"}
+        {...validationState}
+      />
+    );
+
+    const errorBorder = (validationState.error || validationState.warning) && (
```
edleeks87:
```suggestion
    const errorBorder = (error || warning) && (
```

src/components/textbox/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <ErrorBorder
+        data-role="error-border"
+        warning={Boolean(!validationState.error && validationState.warning)}
```
edleeks87:
```suggestion
        warning={Boolean(!error && warning)}
```

src/components/textbox/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+        </LabelSet>
+        <InputSet
+          data-role="input-set"
+          inputWidth={resolvedInputWidth}
+          size={size}
+        >
+          <Input
+            id={uniqueId}
+            name={uniqueName}
+            aria-invalid={Boolean(validationState.error)}
```
edleeks87:
```suggestion
            aria-invalid={Boolean(error)}
```

src/components/textbox/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+            aria-describedby={ariaDescribedByString}
+            value={value}
+            placeholder={!(disabled || readOnly) ? placeholder : undefined}
+            required={required}
+            ref={ref}
+            error={Boolean(validationState.error)}
```
edleeks87:
```suggestion
            error={Boolean(error)}
```

src/components/textbox/__internal__/__next__/text-input.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+### Label Inline
+
+<Canvas of={TextInputStories.LabelInline} />
+
+### Sizes
+
+<Canvas of={TextInputStories.Sizes} />
+
+### Input Width
```
edleeks87, line 76: question: @ljemmo @harpalsingh is the implementation of the `labelInline` with `inputWidth` correct here? If no `inputWidth` is set there space on the lhs, in order to get it to align to left we have to set the `inputWidth` to 80 or (use `containerWidth`) as the default label width is computed as 20% etc

Image

Image

src/components/textbox/__internal__/__next__/text-input.mdx
```diff
@@ -0,0 +1,97 @@
...
+- [Quick Start](#quick-start)
+- [Examples](#examples)
+- [Validation states](#validation-states)
+- [Props](#props)
+- [Translation keys](#translation-keys)
+
+## Quick Start
+
+```javascript
+import TextInput from "carbon-react/lib/components/textbox/__next__";
```
edleeks87, line 30: comment: as the future component is being renamed to `TextInput` I wonder if we should create and alias for this. Think it needs discussing with the rest of the team or at least lets see what further reviewers think

src/components/textbox/__internal__/__next__/text-input.stories.tsx
```diff
@@ -0,0 +1,345 @@
...
+      value={state}
+      onChange={setValue}
+    />
+  );
+};
+
+LabelInline.storyName = "Label Inline";
+
+export const InputWidth = () => {
+  const [state, setState] = useState("");
```
edleeks87, line 145: suggestion (non-blocking): might be worth creating state and handlers for each input as I suspect someone may flag that they don't work independently

src/components/textbox/textbox.style.ts
```diff
@@ -2,16 +2,15 @@ import styled, { css } from "styled-components";
...
-    width: 2px;
     background-color: ${warning
       ? "var(--colorsSemanticCaution500)"
-      : "var(--colorsSemanticNegative500)"};
-    left: -12px;
-    bottom: 0px;
-    top: 0px;
+      : "var(--input-validation-border-error, --colorsSemanticNegative500)"};
   `}
+  transform: scaleX(1);
```
edleeks87: same question as above

src/components/textbox/textbox.style.ts
```diff
@@ -2,16 +2,15 @@ import styled, { css } from "styled-components";
 
 const ErrorBorder = styled.span`
   ${({ warning }: { warning: boolean }) => css`
+    width: 2px;
```
edleeks87: comment: if we do get tokens to use lets leave it here as we don't have a legacy fallback

src/__internal__/error-border/error-border.style.ts
```diff
@@ -0,0 +1,18 @@
+import styled, { css } from "styled-components";
+
+const ErrorBorder = styled.span`
+  ${({ warning }: { warning: boolean }) => css`
+    width: 2px;
```
edleeks87, line 5: comment: there are a couple of spacing tokens that have 2px for value, might be worht checking with DS/UX if any apply here

src/__internal__/error-border/error-border.style.ts
```diff
@@ -0,0 +1,18 @@
...
+    position: absolute;
+    left: -10px;
+    top: 0px;
```
edleeks87, line 8: same comment about checking if any tokens are usable here

src/components/textbox/__next__/text-input.style.ts
```diff
@@ -0,0 +1,82 @@
+import styled from "styled-components";
+import { margin } from "styled-system";
+import applyBaseTheme from "../../../style/themes/apply-base-theme";
+
+interface StyledTextInputProps {
```
edleeks87: interface StyledTextInputProps {
  $labelInline?: boolean;
  $containerWidth?: number;
  $size?: "small" | "medium" | "large";
}

only `size` is needed really rest if up to you

src/components/textbox/__next__/text-input.style.ts
```diff
@@ -0,0 +1,82 @@
...
+  containerWidth?: number;
+  size?: "small" | "medium" | "large";
+}
+
+interface LabelSetProps {
+  labelInline?: boolean;
+  labelSetWidth?: number;
+}
+
+interface InputSetProps {
```
edleeks87: interface InputSetProps {
  $inputWidth?: number;
  $size?: "small" | "medium" | "large";
}

docs/validations.stories.tsx
```diff
@@ -134,15 +135,15 @@ export const ValidationRedesign: StoryObj = () => {
         <RequiredFieldsIndicator mb={2}>
           Fill in all fields marked with
         </RequiredFieldsIndicator>
-        <Textbox
+        <TextInput
```
edleeks87: comment: we need to consider what we need to still capture for the old textbox visual regression wise, again worth discussing with the rest of the team in slack

edleeks87 (commented):
@ljemmo @harpalsingh do we have any designs for inline labels with an inputHint? the alignment looks a little off on this but I can't compare it to anything


Image

src/__internal__/validation-message/__next__/validation-message.style.ts
ljemmo: @edleeks87 i don't believe there's any nice way to show hint text inline. When switching it on within figma it looks identical to Tom's implementation.  TBF inline is generally discouraged. Where is see it used, it never has hint text enabled. e.g for quick filtering on top of tables. I think we have 2 options:
1. Not allow hint text on inline and restrict from a code POV
2. just leave as is knowing that its never going to be used and doesn't matter how it therefore appears.

docs/validations.stories.tsx
damienrobson-sage: Agreed - maybe use one `TextInput` and one `Textbox` to cover off both?

src/__internal__/validation-message/__next__/validation-message.style.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      font-family: var(--fontFamiliesDefault);
+
+      font-size: ${isLarge
+        ? "var(--global-font-static-body-regular-l)"
+        : "var(--global-font-static-body-regular-m)"};
+
+      font-style: normal;
+      font-weight: ${isError ? "500" : "400"};
+      line-height: 150%;
```
damienrobson-sage: **question**: would these need to be tokens, too? Or are we leaving those aside for now?

src/components/textbox/__next__/__internal__/hint-text/hint-text.component.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import StyledHintText from "./hint-text.style";
+
+export interface HintTextProps {
+  /** Children elements */
```
damienrobson-sage: **nitpick**: `Child elements` or `Children` 😆

src/components/textbox/__next__/__internal__/input/input.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+  /** If true, the input will display error styling */
+  error?: boolean;
+  /** The width of the input as a percentage (e.g., 50 for 50%) */
+  inputWidth?: number;
```
damienrobson-sage: **question**: just to confirm, definitely only percentage values to be supported?

src/components/textbox/__next__/__internal__/label/label.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+      readOnly={readOnly}
+    >
+      {children}
+    </StyledLabel>
+  );
+};
+
+export default React.memo(Label);
```
damienrobson-sage: **question**: what's the reason for the manual memoisation?

src/__internal__/validation-message/__next__/validation-message.style.ts
ljemmo: > **question**: would these need to be tokens, too? Or are we leaving those aside for now?

@damienrobson-sage we don't currently distribute core tokens like line-height or font-weight. instead Font tokens can be applied which contain those core values.

tomdavies73: Yep, this is due to a little misunderstanding on my part, some hard coded font-familys, weights, colours, sizes etc. have been used instead of the font shorthand with the corresponding font token. I'll make sure this is addressed where i've used them 👍

src/__internal__/validation-message/__next__/validation-message.style.ts
tomdavies73: I think personally we should keep inline input hints, if we're providing the capacity to make labels inline via `labelInline`, it makes sense to have everything available in both label states. We've made accommodations for error states when the label is inline, it makes sense in my opinion to make those accommodations for the hint text too. What do we think @edleeks87 @ljemmo ?

src/components/textbox/__next__/__internal__/error-border/error-border.style.ts
tomdavies73: So I added this as when the window is being repositioned the absolutely positioned error border can sometimes shrink when the label is inline. We didn't observe this previously as the error border was always in a fixed position, however when inline it is in the centre of the screen and needs to reposition. Adding the transform stopped this shrinking from occurring, and it appeared to be the only solution I could find that worked.

src/components/textbox/__next__/__internal__/input/input.component.tsx
tomdavies73: The DS seem to point toward using percentages for widths on the component, I feel like it keeps things consistent and adaptive. Also gives us more granular control, opposed to supporting any string values etc.

src/components/textbox/__next__/__internal__/label/label.component.tsx
tomdavies73: The old internal label component was memorized, so I followed suit here, after thinking about it, I don't think it's really needed, so I will remove 👍

src/components/textbox/__next__/__internal__/input/input.component.tsx
damienrobson-sage: Perfect, just double-checking given some components support "any valid CSS values" 👍

src/__internal__/validation-message/__next__/validation-message.style.ts
edleeks87: I'm inclined to not include support for the hint inline, mainly as it looks terrible and if it's there it might encourage people to use it and we would therefore create a scenario where we need to improve/fix it etc. My POV is if it shouldn't be used in the wild it shouldn't be included etc

src/components/textbox/__next__/__internal__/error-border/error-border.style.ts
edleeks87: Might need you to demo this as I'm not entirely sure what you mean wrt to it needing to reposition, it should absolutely position itself to the next relative parent so it sounds like something is wrong there

edleeks87: I've been able to repro what you mean, it looks like sub-pixel rendering as the browser tries to recalculate it's positioning. I still see it with your transform but notice it's fixed by using `transform: translateZ(0)` instead

src/__internal__/error-border/error-border.style.ts
edleeks87: decided against moving it for now

docs/validations.stories.tsx
edleeks87, line 1: Need to remove the TextInput from these stories, will do after reviews

docs/validations.mdx
```diff
@@ -64,6 +64,9 @@ Component props that are not supported if the opt-in flag is set to true are lab
 The legacy validation pattern is still available for use, but we recommend using the new validation pattern for a more consistent and user-friendly experience.
 This pattern uses tooltips to provide feedback to users about the state of their input.
 
+The `Textbox` component is used below for documentation purposes, but it is recommend to use the new `TextInput` component instead which only uses the new validation pattern
```
nineteen88: comment: this comment appears to be outdated compared to what the stories are showing below

.storybook/main.ts
```diff
@@ -19,6 +19,8 @@ const getStories = () =>
       ignore: [
         `${projectRoot}/src/**/*-@(interaction|test).stories.@(js|jsx|ts|tsx)`,
         `${projectRoot}/src/__internals__/__stories__/*.*`,
+        // Ignore stories in __internal__ directories, except for those in __stories__ directories
+        `${projectRoot}/src/components/**/__internal__/**/*.{mdx,stories.@(js|jsx|ts|tsx)}`,
```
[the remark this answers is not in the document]
edleeks87: it's intentionally ignored for the production build of storybook, the PR description needs updating

src/components/icon/icon.style.ts
```diff
@@ -85,6 +85,42 @@ function adjustIconBgSize(fontSize?: FontSize, bgSize?: BgSize) {
...
+  }
+
+  .mentions-list-item && {
+    color: currentColor;
+  }
+
+  .mentions-list-item:hover &&,
+  .mentions-list-item.selected && {
+    color: currentColor;
+  }
```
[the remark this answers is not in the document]
edleeks87: changes not related to this PR

src/__internal__/utils/helpers/combine-refs/index.ts
```diff
@@ -0,0 +1,17 @@
...
+export default <T = unknown>(
+  ...refs: (MutableRefObject<T | null> | RefCallback<T> | null | undefined)[]
+): RefCallback<T> => {
+  return (node: T | null) => {
+    refs.forEach((ref) => {
+      if (!ref) return;
+
+      if (typeof ref === "function") {
+        ref(node);
+      } else {
```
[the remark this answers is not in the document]
edleeks87: If I use `Ref` I'd have to coerce it to `MutableRefObject` to assign to `current`

src/__internal__/legacy-input/input-presentation.style.ts
```diff
@@ -125,18 +124,16 @@ const InputPresentationStyle = styled.div<
...
-        : `var(--colorsUtilityReadOnly600)`};
+      border-color: transparent;
     `}
 
-  ${({ align, prefix }) =>
+  ${({ align }) =>
     align === "right" &&
-    `flex-direction: ${prefix ? "row" : "row-reverse"};
+    `flex-direction: row-reverse;
   `}
```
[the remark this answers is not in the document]
edleeks87: Legacy component not really worth updating

DipperTheDan (dismissed):

src/components/textbox/__internal__/__next__/text-input.component.tsx
```diff
@@ -0,0 +1,305 @@
...
+      type = "text",
+      validationMessagePositionTop,
+      value,
+      warning,
+      ...rest
+    }: TextInputProps,
+    ref: React.ForwardedRef<HTMLInputElement>,
+  ) => {
+    const localRef = useRef<HTMLInputElement>(null);
+    // change to imperative handleFocus method when moving this from __next__
```
nineteen88: comment(minor): is it worth adding `todo:` to this comment for traceability?

nineteen88 (dismissed):

docs/validations.mdx
edleeks87: Good spot, I'll update

src/components/textbox/__internal__/__next__/text-input.component.tsx
edleeks87: yep I'll expand this a bit, we might need to consider something custom here in the long run, we want to only expose the handle type but some of our internal usage relies on being able to get a ref to the underlying HTMLElement. Either way a conversation for another day

DipperTheDan (dismissed):

nineteen88 (dismissed):

DipperTheDan (dismissed):

DipperTheDan (dismissed):

paulrobinson88 (dismissed):

nineteen88 (dismissed):

DipperTheDan (dismissed):

paulrobinson88 (approved):

DipperTheDan (approved):

nineteen88 (approved):

PR #7560, opened 2025-10-01
fix(flat-table): ensure focus is only applied when using keyboard navigation - FE-7454

### LINKED ISSUE #7467: Enable prop to disable focus on FlatTable
### Description

To meet accessibility point raised by the Sage Design team together with Sage Accessibility team, we would like to have an option to disable focus on FlatTable. This is needed as when non interactive elements are in FlatTable, we don't want focus to be enable on whole FlatTable.  

### Suggested solution

Add prop that can control FlatTable focus (enabled by default, disabled as on option).

### Demo URL

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Confidentiality

DipperTheDan:
fixes #7467

### Proposed behaviour

`Flat-Table` should only be focusable when using keyboard navigation.

### Current behaviour

`Flat-Table` can be focused when using keyboard navigation or clicking with a mouse.

### Checklist

#### QA

### Additional context

I have moved one of the unit tests that asserts the styles applied via focusing of the flat table container in to a Playwright test. I have also included an interaction story so that we can capture this via chromatic too. 

### Testing instructions

- Focus styling should only be applied to Flat-Table via keyboard navigation.
- No other functional or styling regressions associated with Flat-Table.
- New Chromatic snapshot should capture what is intended.

### DIFF
--- a/src/components/flat-table/flat-table.interaction.stories.tsx
+++ b/src/components/flat-table/flat-table.interaction.stories.tsx
@@ -1,0 +1,61 @@
+import React from "react";
+import { StoryObj, composeStory } from "@storybook/react";
+import { userEvent, within, expect, waitFor } from "@storybook/test";
+import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
+import DefaultDecorator from "../../../.storybook/utils/default-decorator";
+import { FlatTable } from "./flat-table.component";
+import meta, { DefaultStory, Paginated } from "./flat-table.stories";
+
+type Story = StoryObj<typeof FlatTable>;
+
+export default {
+  title: "Flat Table/Interactions",
+  component: FlatTable,
+  parameters: {
+    themeProvider: { chromatic: { theme: "sage" } },
+  },
+};
+
+export const FlatTableFocused: Story = {
+  ...DefaultStory,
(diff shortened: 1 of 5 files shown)


damienrobson-sage (approved):

src/components/flat-table/flat-table.interaction.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  ...DefaultStory,
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+
+    const canvas = within(canvasElement);
+    const focusableTableContainer = canvas.getByTestId("flat-table-container");
+
+    await userEvent.tab();
+
+    await waitFor(() => expect(focusableTableContainer).toHaveFocus());
```
edleeks87: question (non-blocking): do we need to use `waitFor` here, will `await expect` not suffice?

DipperTheDan: Yeah, I think you're correct. A usual `await expect...` should work. I'll get this amended.

damienrobson-sage (approved):

dariusbercea-sage (approved):

PR #7561, opened 2025-10-01
test(button-toggle): add play functions

sianford:
### Proposed behaviour

Add Play functions to the `Button Toggle` component to capture the following:
- click states
- hover and focus states
- keyboard interactions
- button deselection/untoggling
- navigation through a wrapped button group

### Current behaviour

There are currently no Play functions for the `Button Toggle` component

### Checklist

#### QA

### Testing instructions

Ensure the interaction tests pass and are captured as expected in Chromatic.
Five stories added:
`Button Toggle > Interactions >`
- `On Click`
- `Focus and Hover States`
- `Kyboard Interaction`
- `Deselection`
- `Wrapped Buttons`

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs-interactions.stories.tsx
+++ b/src/components/breadcrumbs/breadcrumbs-interactions.stories.tsx
@@ -1,0 +1,150 @@
+import React from "react";
+import { StoryFn, StoryObj } from "@storybook/react";
+import { userEvent, within, expect } from "@storybook/test";
+
+import Breadcrumbs from "./breadcrumbs.component";
+import Box from "../box";
+
+import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
+import userInteractionPause from "../../../.storybook/utils/user-interaction-pause";
+import { Crumb } from "./crumb";
+
+type Story = StoryObj<typeof Breadcrumbs>;
+
+export default {
+  title: "Breadcrumbs/Interactions",
+  component: Breadcrumbs,
+  parameters: {
+    themeProvider: { chromatic: { theme: "sage" } },
+  },
+  decorators: [(StoryToRender: StoryFn) => <StoryToRender />],
+};
+
+export const DefaultBreadcrumbs: Story = {
+  render: () => (
+    <Breadcrumbs>
+      <Crumb href="#">Breadcrumb 1</Crumb>
+      <Crumb href="#">Breadcrumb 2</Crumb>
+      <Crumb href="#" isCurrent>
+        Current Page
+      </Crumb>
+    </Breadcrumbs>
+  ),
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) return;
+    const canvas = within(canvasElement);
+
+    const nav = canvas.getByRole("navigation", { name: /breadcrumb/i });
+    expect(nav).toBeInTheDocument();
+
+    const links = canvas.getAllByRole("link");
+    expect(links.length).toBe(2);
+
+    await userEvent.tab();
+    await userInteractionPause(100);
+    expect(links[0]).toHaveFocus();
+
+    await userEvent.tab();
+    await userInteractionPause(100);
+    expect(links[1]).toHaveFocus();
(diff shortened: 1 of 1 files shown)


src/components/button-toggle/button-toggle-interactions.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+            onChange={onChangeHandler}
+          >
+            <ButtonToggle value="far">Foo</ButtonToggle>
+            <ButtonToggle value="bor">Bar</ButtonToggle>
+            <ButtonToggle value="boz">Baz</ButtonToggle>
+          </ButtonToggleGroup>
+        </Box>
+      </>
+    );
+  },
```
damienrobson-sage, line 69: **comment**: can we refactor this into a reusable component to remove the need for the disabled linter, prevent unnecessary redefinitions on re-render, etc.? Hapy to talk you through it - it's not a biggie but will make the code a bit lighter and a bit easier to read.

```ts
const InteractionComponent = (props) => {
    const [value, setValue] = useState("");
    const onChangeHandler = (
      event: React.MouseEvent<HTMLButtonElement>,
      selectedValue?: string,
    ) => {
      setValue(selectedValue as string);
    }

    return (
      <>
        <Box margin={4} width="250px" display="flex" flexWrap="nowrap">
          <ButtonToggleGroup
            id="button-toggle-group-id"
            value={value}
            onChange={onChangeHandler}
          >
            <ButtonToggle value="foo">Foo</ButtonToggle>
            <ButtonToggle value="bar">Bar</ButtonToggle>
            <ButtonToggle value="baz">Baz</ButtonToggle>
          </ButtonToggleGroup>
        </Box>
      </>
    );
}
```

and then render just becomes this for every story:

```ts
render: ()=> <InteractionComponent />
```

You can also set it up to allow different child components, etc. as needed.

src/components/button-toggle/button-toggle-interactions.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+
+export const OnClick: Story = {
+  render: () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const [value, setValue] = useState("");
+    function onChangeHandler(
+      event: React.MouseEvent<HTMLButtonElement>,
+      selectedValue?: string,
+    ) {
+      setValue(selectedValue as string);
```
Parsium: **suggestion(non-blocking)**: what do you think about clearing the value when `selectedValue` is undefined? (e.g. when a button is deselected)

This avoids the type assertion and ensures type safety by defaulting to an empty string. Not essential though.

```suggestion
      selectedValue?: string,
    ) {
      setValue(selectedValue ?? "");
```

src/components/button-toggle/button-toggle-interactions.stories.tsx
sianford: After we looked at this, we concluded that for the more complex examples for this component, with various disabled buttons and whole button toggles, it would be difficult to use this method. I've made a note of this for future components though.

src/components/button-toggle/button-toggle-interactions.stories.tsx
sianford: Thanks. I've amended these.

damienrobson-sage (approved):

src/components/button-toggle/button-toggle-interactions.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+type Story = StoryObj<typeof ButtonToggle>;
+
+export default {
+  title: "Button Toggle/Interactions",
+  parameters: { themeProvider: { chromatic: { theme: "sage" } } },
+};
+
+export const OnClick: Story = {
+  render: () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
Parsium, line 20: **suggestion(non-blocking)**: using a named function for the rendered component should allow us to avoid disabling the linter:

```suggestion
  render: function OnClick() {
```

PR #7569, opened 2025-10-07
test(test-utils): adds a helper to replace Logger deprecation checks

damienrobson-sage:
### Proposed behaviour

Create a new helper util for testing: `assertDeprecationWarning`. This can be used in place of e.g.:

Example:

```ts
test("should log a deprecation warning for onClick prop", () => {
  assertDeprecationWarning({
    component: <Badge onClick={() => {}} />,
    deprecationMessage:
      "The `onClick` prop in `Badge` is deprecated and will soon be removed.",
  });
});
```

The helper also takes a third option, `method`, which can be used to test `Logger.warn` and `Logger.error`. `Logger.deprecate` is used as the default option if this is not set.

### Current behaviour

There are a number of test scenarios that use variations on a similar approach to test that the deprecation warnings from `Logger`, but the implementation is inconsistent.

Example:

```ts
test("should log a deprecation warning for onClick prop", () => {
  const loggerSpy = jest.spyOn(Logger, "deprecate");
  render(<Badge onClick={() => {}} />);

  expect(loggerSpy).toHaveBeenCalledWith(
    "The `onClick` prop in `Badge` is deprecated and will soon be removed.",
  );
  expect(loggerSpy).toHaveBeenCalledTimes(1);

  loggerSpy.mockRestore();
  loggerSpy.mockClear();
});
```

### Checklist

#### QA

### Testing instructions

Run the `Badge` or `TextArea` component unit tests. All tests should pass using the new structure.

**IMPORTANT**: due to the ongoing audit work, the changed unit tests in Badge and TextArea have been done for demonstration purposes only. The work to update the unit tests to use the new helper should be done when the relevant components are updated.

### DIFF
--- a/src/__spec_helper__/__internal__/test-utils.ts
+++ b/src/__spec_helper__/__internal__/test-utils.ts
@@ -9,6 +9,7 @@ import {
 
 import { space } from "../../style/themes/base/base-theme.config";
 import { mockMatchMedia } from "../mock-match-media";
+import Logger from "../../__internal__/utils/logger";
 
 const isUpper = (char: string) => char.toUpperCase() === char;
 const humpToDash = (acc: string, char: string) =>
@@ -344,7 +345,56 @@ const testStyledSystemFlexBox = (
   );
 };
 
+// Helper to assert that a deprecation warning is logged only once
+interface LoggerComponentMessageAssertionProps {
+  component: React.ReactNode;
+  message: string;
+  method?: "deprecate" | "warn" | "error";
+}
+
+/**
+ * assertLoggerComponentMessage
+ * @param component: The React component to render that triggers the deprecation warning.
+ * @param message: The expected deprecation warning message.
+ * @param method: The Logger method to spy on, either "deprecate", "warn" or "error". Defaults to "deprecate".
+ *
+ * This function renders the provided component and checks that the specified deprecation warning
+ * is logged exactly once during the initial render. It then re-renders the component and verifies
+ * that the warning is not logged again, ensuring that deprecation warnings are not repeated.
+ */
+const assertLoggerComponentMessage = ({
+  component,
+  message,
+  method = "deprecate",
+}: LoggerComponentMessageAssertionProps) => {
+  const loggerSpy = jest.spyOn(Logger, method);
+
+  const { rerender } = render(component);
+
+  // We do it this way to compensate for other logs that might be present
+  // e.g. in Password, ButtonMinor is used to render the toggle
+  // button. Using this helper in a test for Password would
+  // also capture the deprecation warning from ButtonMinor
+  // and cause issues.
+  const loggedMessages = loggerSpy.mock.calls.map((call) => call[0]);
+  expect(loggedMessages).toContain(message);
+
+  const messageInstances = loggedMessages.filter((msg) => msg === message);
+  expect(messageInstances).toHaveLength(1);
+
+  loggerSpy.mockReset();
+
+  rerender(component);
+  const loggedMessagesAfterRerender = loggerSpy.mock.calls.map(
+    (call) => call[0],
+  );
+  expect(loggedMessagesAfterRerender).not.toContain(message);
+
+  loggerSpy.mockRestore();
+};
+
 export {
+  assertLoggerComponentMessage,
   assertStyleMatch,
   toCSSCase,
   makeArrayKeys,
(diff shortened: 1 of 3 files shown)


src/__spec_helper__/__internal__/test-utils.ts
```diff
@@ -344,7 +345,58 @@ const testStyledSystemFlexBox = (
...
+  const loggerSpy = jest.spyOn(Logger, method);
+
+  const { rerender } = render(component);
+
+  // We do it this way to compensate for other logs that might be present
+  // e.g. in Password, ButtonMinor is used to render the toggle
+  // button. Using this helper in a test for Password would
+  // also capture the deprecation warning from ButtonMinor
+  // and cause issues.
+  const loggedMessages = loggerSpy.mock.calls.map((call) => call[0]);
```
tomdavies73, line 379: **praise**: this makes sense, I had to do something similar in `Textbox` as we have multiple logs there - this ensures we can still assert each call is only logged once without relying on `toHaveBeenCalledTimes` which will fail 👍

src/__spec_helper__/__internal__/test-utils.ts
```diff
@@ -344,7 +345,58 @@ const testStyledSystemFlexBox = (
...
+// Helper to assert that a deprecation warning is logged only once
+interface AssertDeprecationWarning {
+  component: React.ReactNode;
+  deprecationMessage: string;
+  method?: "deprecate" | "warn" | "error";
+}
+
+/**
+ * assertDeprecationWarning
+ * @param param0 - component: The React component to render that triggers the deprecation warning.
```
tomdavies73: **suggestion**: `param0` has been used for each param which appears to be some kind of placeholder. 

If you do the following: 

```
/**
 * assertDeprecationWarning
 * @param component The React component to render that triggers the deprecation warning.
 * @param deprecationMessage The expected deprecation warning message.
 * @param method The Logger method to spy on, either "deprecate", "warn" or "error". Defaults to "deprecate".
 *
 * This function renders the provided component and checks that the specified deprecation warning
 * is logged exactly once during the initial render. It then re-renders the component and verifies
 * that the warning is not logged again, ensuring that deprecation warnings are not repeated.
 */
```

This will result in a more intuitive description of the functions parameters 👍 

Screenshot 2025-10-07 at 09 49 51

src/__spec_helper__/__internal__/test-utils.ts
```diff
@@ -344,7 +345,58 @@ const testStyledSystemFlexBox = (
...
+ * @param method: The Logger method to spy on, either "deprecate", "warn" or "error". Defaults to "deprecate".
+ *
+ * This function renders the provided component and checks that the specified deprecation warning
+ * is logged exactly once during the initial render. It then re-renders the component and verifies
+ * that the warning is not logged again, ensuring that deprecation warnings are not repeated.
+ */
+const assertDeprecationWarning = ({
+  component,
+  deprecationMessage,
+  method = "deprecate",
```
Parsium, line 368: **question**: do we need this granularity, or could we simplify things by ensuring we only use `Logger.deprecate` for deprecation warnings?

src/components/textarea/textarea.test.tsx
```diff
@@ -758,25 +747,22 @@ test("renders with the expected custom border radius styling as an array", () =>
...
-      ]}
-    />,
-  );
-
-  expect(loggerSpy).toHaveBeenCalledWith(
-    "The `borderRadius` prop in `Textarea` component only supports up to 4 values.",
-  );
-
-  loggerSpy.mockRestore();
+  assertDeprecationWarning({
```
Parsium: **question**: just wondering, would this be considered as a warning rather than a deprecation?

damienrobson-sage: It only covered deprecation warnings to begin with, and then it evolved. Not sure if that answers your question, though... 👍

damienrobson-sage: I picked this example specifically because it used `Logger.warn` instead of of `Logger.deprecate` and thus was a handy example to demo the method switching capability

src/__spec_helper__/__internal__/test-utils.ts
tomdavies73: I don't have any issue with the method having multiple options if it's needed, @damienrobson-sage could we rename this util to be more implementation agnostic? That could be a happy medium.

damienrobson-sage: I'll rename it - the fact that different components use different approaches may (for now at least) be something to leave alone and address as the components are aligned with the DS.

tomdavies73 (approved):

Parsium (approved):

PR #7576, opened 2025-10-08
 fix(responsive-vertical-menu): ensure margin & padding props are applied correctly

tomdavies73:
### Proposed behaviour

Moves the application of margin props on `ResponsiveVerticalMenuItem` and `ResponsiveVerticalMenuDivider` to the `li` or `div` element, opposed to the `a` or `hr` anchor element(s). This means the margin props added now are applied against siblings, which can facilitate various designs and patterns. 

One of which being one `ResponsiveVerticalMenuDivider` and `ResponsiveVerticalMenuItem` being placed at the bottom of the `ResponsiveVerticalMenu` with the use of `marginTop="auto"`

To facilitate this pattern, a mandatory 40px margin bottom to the last item was removed as this was code which was no longer needed and caused issues with specific patterns.

Also, the application of padding props has been fixed with a new min-height introduced to each menu item with some additional internal padding to prevent overlaps with focus styling.

Screenshot 2025-10-08 at 17 17 32

### Current behaviour

Currently, margin based props are applied on the `a` element, which means the props can not be applied effectively. 

### Checklist

#### QA

### Testing instructions

A new interactions story has been added to show `ResponsiveVerticalMenuDivider` and `ResponsiveVerticalMenuItem` being placed at the bottom of the `ResponsiveVerticalMenu` with the use of `marginTop="auto"`. This will also add a chromatic snapshot to catch future regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [157.6.1](https://github.com/Sage/carbon/compare/v157.6.0...v157.6.1) (2025-10-16)
+
+### Bug Fixes
+
+* ensure aria-hidden & inert is handled correctly with open modals in different carbon instances ([0d0ccc8](https://github.com/Sage/carbon/commit/0d0ccc83107169f3e63bd621d47aab2ac7cbe930))
+
+## [157.6.0](https://github.com/Sage/carbon/compare/v157.5.3...v157.6.0) (2025-10-10)
+
+### Features
+
+* **tabs:** add warnings for upcoming deprecations ([5c797a6](https://github.com/Sage/carbon/commit/5c797a6a09677b83247c05aa50429946bcb5de78))
+
 ## [157.5.3](https://github.com/Sage/carbon/compare/v157.5.2...v157.5.3) (2025-10-08)
 
 ### Bug Fixes
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "carbon-react",
-  "version": "157.5.3",
+  "version": "157.6.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "carbon-react",
-      "version": "157.5.3",
+      "version": "157.6.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@floating-ui/dom": "~1.2.9",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "carbon-react",
-  "version": "157.5.3",
+  "version": "157.6.1",
   "description": "A library of reusable React components for easily building user interfaces.",
   "files": [
     "lib",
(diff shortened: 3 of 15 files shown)


src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-item/responsive-vertical-menu-item.style.ts
```diff
@@ -137,7 +136,7 @@ export const StyledNestedMenu = styled.ul<{
 
 export const StyledResponsiveMenuListItem = styled.li`
   list-style: none;
-  margin: 0;
+  ${margin}
   padding: 0;
   width: 100%;
   max-width: 340px;
```
Parsium: **issue**: I noticed it's possible for an item to overflow its menu with certain margins:

Screenshot 2025-10-13 at 13 07 23

Screenshot 2025-10-13 at 13 28 47


One way we could fix this would be to remove the width constraints on items and let the parent menu handle their width:

```diff
export const StyledResponsiveMenuListItem = styled.li`
  list-style: none;
  ${margin}
  padding: 0;
- width: 100%;
- max-width: 340px;

export const StyledResponsiveMenu = styled.ul<StyledResponsiveMenuProps>`
- align-items: center;
+ align-items: stretch;
`

export const StyledResponsiveVerticalMenuDivider = styled.div<StyledResponsiveVerticalMenuDividerProps>`
  ${margin}
- width: ${({ depth, responsive }) =>
-    depth > 0 && responsive ? "88%" : "100%"};
+ ${({ depth, responsive }) => depth > 0 && responsive && `max-width: 88%`};
`;
```

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-interaction.stories.tsx
```diff
@@ -156,6 +182,31 @@ export const ToggleOpenOnClick: Story = {
...
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) {
+      return;
+    }
+
+    const canvas = within(canvasElement);
+    const menuToggle = canvas.getByRole("button");
+
+    await userEvent.click(menuToggle);
+    await userInteractionPause(1000);
```
DipperTheDan, line 196: **suggestion(non-blocking):** Appreciate you're following the convention of the example above, and this fix is urgently needed so we don't need to sort it now but I think it's best to use `await expect(vertical-menu-locator).someAssertionHere` to reduce the potential flake in the Chromatic snapshot.

tomdavies73: Good point mate, as I've just followed the approach used elsewhere in the file I will probably just leave this to be addressed when the rest of the file is re-aligned with best practice 👍

dariusbercea-sage (approved):

PR #7581, opened 2025-10-15
fix: ensure aria-hidden & inert is handled correctly with open modals in different carbon instances

tomdavies73:
### Proposed behaviour

Adds an extra step of validation in `useModalAria` with the addition of a `data-modal-hidden` attribute to ensure multiple instances open at the same time only take ownership of managing elements it encounters. Stops separate instances of the hook restoring elements or marking elements (adding attributes) which they should not have any control over, and places the hook into a queue, with the next hook only managing elements after the previous one has finished.

### Current behaviour

Currently, we use a strong method of marking all elements which are not in the root toward the modal as `aria-hidden` and `inert`, we then clean up by removing all the attributes we previously added. This is fine in standard use cases, but causes issues when we have two open stacked modals running in two separate carbon instances, as they can incorrectly restore or add attributes, which causes a bug in production where everything remains `aria-hidden` and `inert` even when all modals are closed.

### Checklist

#### QA

### Testing instructions

I have added the `"Separate CarbonInstances With Dialog"` Adaptive Sidebar test story to replicate an edge case where two instances of Carbon are used at the same time. 

**Steps to reproduce are:**

Open the adaptive sidebar on a large screen size, open the dialog, then re-size so the adaptive sidebar becomes a standard sidebar, then close the sidebar & close the dialog after. You can still interact with all other elements and even repeat the steps again if you wish. 

However, before the fix using this reproduction and repeating the same steps, you can see that this is not possible. The rest of the page remains `aria-hidden` and `inert`: 

https://stackblitz.com/edit/parsium-carbon-starter-a3u4txxq?file=src%2FApp.tsx%3AL22-L22,src%2Fmain.tsx,index.html

### DIFF
--- a/src/components/link/link-interaction.stories.tsx
+++ b/src/components/link/link-interaction.stories.tsx
@@ -1,0 +1,338 @@
+import React from "react";
+import { StoryObj } from "@storybook/react";
+import { within, expect, userEvent } from "@storybook/test";
+
+import Link from ".";
+import Box from "../box";
+import { Menu, MenuItem } from "../menu";
+
+import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
+import DefaultDecorator from "../../../.storybook/utils/default-decorator";
+
+type Story = StoryObj<typeof Link>;
+
+export default {
+  title: "Link/Interactions",
+  parameters: {
+    info: { disable: true },
+    themeProvider: { chromatic: { theme: "sage" } },
+    chromatic: {
+      disableSnapshot: false,
+    },
+  },
+};
+
+export const Focus: Story = {
+  render: () => (
+    <Box width="max-content" display="flex" flexDirection="column" gap="32px">
+      <Link href="https://carbon.sage.com" target="_blank" data-role="target">
+        Typical link
+      </Link>
+      <Link disabled>Disabled link</Link>
+      <Link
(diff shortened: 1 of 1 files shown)


src/components/adaptive-sidebar/adaptive-sidebar-test.stories.tsx
```diff
@@ -469,6 +472,73 @@ ExampleImplementation.parameters = {
...
+            <Instance2 />
+          </CarbonProvider>
+        </>,
+      );
+    }
+
+    return () => {
+      if (container2) {
+      }
+    };
```
damienrobson-sage: **question**: why is this useEffect's clean-up empty?

Parsium: **issue**: this effect initialises the React root but doesn't destroy it afterwards. Adding a cleanup would ensure we don't leave root in memory when the component unmounts:

```suggestion
    let root = Root | null = null;

    const container = document.getElementById("mfe-2-root");
    if (container && !container.hasChildNodes()) {
      root = createRoot(container);
      root.render(
        <CarbonProvider>
          <Instance2 />
        </CarbonProvider>
      );
    }

    return () => {
      root?.unmount();
      root = null;
    };
```

PR #7591, opened 2025-10-22
test(batch selection): add play functions

divyajindel:
### Proposed behaviour

Add Play Functions to batch selection component for testing the below.
- icon click
- icon hover

### Current behaviour

There are currently no Play Functions for the batch Selection component

### Checklist

#### QA

### Testing instructions

Ensure tests are executed in the expected way.

### DIFF
--- a/src/components/batch-selection/batch-selection-interactions.stories.tsx
+++ b/src/components/batch-selection/batch-selection-interactions.stories.tsx
@@ -1,0 +1,112 @@
+import React from "react";
+import { StoryFn, StoryObj } from "@storybook/react";
+import { userEvent, within } from "@storybook/test";
+import BatchSelection from ".";
+import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
+import IconButton from "../icon-button";
+import Icon from "../icon";
+import Button from "../button";
+import DefaultDecorator from "../../../.storybook/utils/default-decorator";
+
+type Story = StoryObj<typeof BatchSelection>;
+
+export default {
+  title: "Batch Selection/Interactions",
+  component: BatchSelection,
+  parameters: {
+    themeProvider: { chromatic: { theme: "sage" } },
+  },
+  decorators: [(StoryToRender: StoryFn) => <StoryToRender />],
+};
+
+export const ClickBatchSelection: Story = {
+  render: () => (
+    <BatchSelection selectedCount={0}>
+      <Button size="small" mx={1} buttonType="secondary">
+        Select All 38 items
+      </Button>
+      <IconButton onClick={() => {}}>
+        <Icon type="csv" />
+      </IconButton>
+      <IconButton onClick={() => {}}>
+        <Icon type="bin" />
+      </IconButton>
+      <IconButton onClick={() => {}}>
+        <Icon type="pdf" />
+      </IconButton>
+    </BatchSelection>
+  ),
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


src/components/batch-selection/batch-selection-interactions.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <IconButton onClick={() => {}}>
+        <Icon type="pdf" />
+      </IconButton>
+    </BatchSelection>
+  ),
+  play: async ({ canvasElement }) => {
+    if (!allowInteractions()) {
+      return;
+    }
+    await userInteractionPause(500);
```
paulrobinson88: **Question:** We are trying to avoid using the `userInteractionPause` due to issues flagged with Chromatic inconsistencies but also, is this required in this location? As there doesn't appear to be a `userEvent` before or after it?

src/components/batch-selection/batch-selection-interactions.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <DefaultDecorator>
+        <StoryToRender />
+      </DefaultDecorator>
+    ),
+  ],
+};
+hoverBatchSelection.storyName = "Hover";
+hoverBatchSelection.parameters = {
+  pseudo: {
+    hover: '[data-component="icon-button"]',
```
paulrobinson88: question (non-blocking): It doesn't appear that the `hover` is applied to the first `button` just the `icon-buttons` is this intentional?

src/components/batch-selection/batch-selection-interactions.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (!allowInteractions()) {
+      return;
+    }
+    await userInteractionPause(500);
+
+    const canvas = within(canvasElement);
+    const itemButton = canvas.getAllByRole("button");
+
+    await userInteractionPause(500);
+    await userEvent.click(itemButton[0]);
```
paulrobinson88: **Suggestion:** To avoid using `userInteractionPause` you may be able to use something like the below. 
```suggestion
    await userEvent.click(itemButton[0], {delay:500});
```
But only if there is a pause required, for example to wait for an animation to complete in the `accordion` component

src/components/batch-selection/batch-selection-interactions.stories.tsx
divyajindel: I've used  **userInteractionPause**  to check things more clearly but i'll avoid using it now as it's flagging unexpected snapshot changes.

src/components/batch-selection/batch-selection-interactions.stories.tsx
divyajindel: I've added hover on icon buttons but forgot to add for button so will add it now.

src/components/batch-selection/batch-selection-interactions.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+      <DefaultDecorator>
+        <StoryToRender />
+      </DefaultDecorator>
+    ),
+  ],
+};
+hoverBatchSelection.storyName = "Hover";
+hoverBatchSelection.parameters = {
+  pseudo: {
+    hover: '[data-component="button"], [data-component="icon-button"]',
```
paulrobinson88: After doing a quick QA on this, it appears that the hover styling is applied to the next story selected in Storybook. We may need to be more specific about the pseudo hover on this story so this doesn't happen. 
I'll send on an example of avoiding this in the `button` component

PR #7599, opened 2025-10-29
feat: styled system replacement poc work - FE-5531

DipperTheDan:
WIP POC. I am aware that we have a coverage issue, but this is due to the changes I have had to make in the menu-item styling. This is a proof of concept so my aim was to show that my proposal could work with a complex component.

### Proposed behaviour

POC for replacing `styled-system`

### Current behaviour

`styled-system` exists

### Checklist

#### QA

### Additional context

I have refrained from running the Chromatic workflow due to the not wanting to burn through Chromatic snapshots.


docs/styled-system-replacement-poc.mdx
```diff
@@ -0,0 +1,141 @@
...
+- Expand the utility functions to cover more `styled-system` features as needed.
+- Thoroughly test the new implementation to ensure parity with `styled-system`.
+- Measure performance and bundle size improvements.
+- Improve folder structure and organization of the utilities (open to suggestions or do we even move it to a different repo altogether?).
+- Receive feedback from the team and make improvements to the POC if required.
+
+## Potential Names
+- `carbon-style` - might be a bit vague.
+- `styled` - sounds similar to `styled-system` and `styled-components` and might also imply a connection to those libraries.
+- `Styley McStyleFace` - just to see who actually read this far.
```
damienrobson-sage, line 141: **praise**: no further discussion needed IMO. This is the way.

docs/styled-system-replacement-poc.mdx
```diff
@@ -0,0 +1,141 @@
...
+If the consumer passes a value of `ml="XS"`, it will be interpreted as `ml={3}` internally and the appropriate spacing will be applied based on the multiplier; `3 * 8 = 24px` in this case.
+The end result will be a style of `margin-left: 24px;` applied to the component. <br />
+Seeing as this is a POC, support for negative values has been included as well, so `-XS` would be interpreted as `-3`. The rationale behind this is that negative CSS values do exist. 
+As a team, we can have a discussion on this and whether we want to support negative sizing letters or not, or we support it but do not advertise or advocate the use.
+
+### Example: Spacing
+
+Component file: 
+```tsx
+import { paddingProps } from "../utils/spacing-types";
```
damienrobson-sage, line 58: **nitpick**: `PaddingProps` to match casing in `spacing-types`

src/components/dips-box/dips-box.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+        Margin Left and Padding Top String
+      </DipsBox>
+
+      <DipsBox ml="XS" pt="L">
+        Margin Left and Padding Top String
+      </DipsBox>
+    </>
+  );
+};
+Spacing.storyName = "Spacing";
```
damienrobson-sage, line 39: **suggestion**: demonstrate that the types can mix-and-match e.g.:
```javascript
<DipsBox m="3px" pt="L" pb={4}>
```

DipperTheDan: Yeah that's a good idea. I'll get that added in the next round of changes.

src/components/dips-box/utils/spacing-types.ts
```diff
@@ -0,0 +1,147 @@
...
+  boxSizing?: string;
+  /** Short hand `visibility` property */
+  visibility?: string;
+  /** Short hand `aspect-ratio` property */
+  aspectRatio?: string;
+}
+
+export interface MaxWidthProps {
+  maxWidth?: string;
+}
```
Parsium: **question(minor)**: do we need this interface if `maxWidth` is part of `LayoutProps`?

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
+import { css } from "styled-components";
+
+/** If we want to add TS support, we can use this. For now, leaving out to avoid complexity in JS life.
+ * Size range from XS to XL
+ * export type SpacingSize = "none" | "3XS" | "2XS" | "XS" | "S" | "M" | "L" | "XL"| "2XL" | "3XL" | "4XL";
+ */
+
+// Mapping of size to multiplier value for spacing calculations
+export const spacingSizeMap = {
```
Parsium: **thought**: could we defer these new size aliases to a second pass?

Since we didn't support them before, I think focusing on replacing the existing functionality from `styled-system` first would help minimise the number of changes and make testing for regressions easier.

src/components/menu/menu-item/menu-item.component.tsx
```diff
@@ -345,7 +351,15 @@ export const MenuItem = forwardRef<
...
 
-    const paddingProps = filterStyledSystemPaddingProps(rest);
+    const paddingPropsRaw = filterStyledSystemPaddingProps(rest);
+    // Remove any null values from paddingProps to ensure type compatibility
+    const paddingProps = Object.fromEntries(
+      Object.entries(paddingPropsRaw).filter(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        ([_, paddingValue]) => paddingValue !== null,
+      ),
+    );
```
Parsium, line 379: **comment**: if we can update `spacing.js` to be written in TypeScript, we might be able to remove these null checks, assuming TypeScript would be able to infer the values correctly.

src/__spec_helper__/__internal__/test-utils.ts
```diff
@@ -251,6 +270,44 @@ const testStyledSystemPadding = (
...
+  assertOpts?: jest.Options,
+) => {
+  describe.each(paddingProps)(
+    'when a custom spacing is specified using the "%s" styled props',
+    (styledProp, propName) => {
+      it(`should set ${propName} styling correctly`, () => {
+        const props = { [styledProp]: 3 };
+        render(component({ ...props }));
+
+        assertStyleMatch({ [propName]: "24px" }, elementQuery(), assertOpts);
```
Parsium: **suggestion**: I noticed the default values of the new test utils differ from their old counterparts.

Keeping the old default values could help limit the amount of component code changes we'd need to make, as if we've replaced `styled-system`'s functionality, we'd only need to swap the references to the old util with the new one:

```suggestion
        const props = { [styledProp]: 2 };
        render(component({ ...props }));

        assertStyleMatch({ [propName]: "var(--spacing200)" }, elementQuery(), assertOpts);
```

src/components/menu/menu-item/menu-item.style.ts
Parsium, line 1: **comment**: would it be possible to revert these `Menu` changes?

I noticed if we update the `testStyledPadding` util to use the same default padding values as the old `testStyledSystemPadding` util, our tests still pass without these component changes. Keeping things scoped to the new `styled-system` replacement might make it easier for us to spot any behavioural differences.

Happy to discuss if I'm overlooking something :)

src/components/dips-box/utils/spacing-types.ts
DipperTheDan: Probably not, I did this because we have a `MaxWidthProps` that is imported from `styled-system` that is used in `MenuItem` and `Submenu`. I did this to take make sure my solution would work as a drop in replacement for `styled-system`.

src/components/dips-box/utils/spacing.js
DipperTheDan: Yeah, we could. I mainly did this to prove to us and DS that this solution would support the sizing alias when we choose to implement them.

src/components/menu/menu-item/menu-item.component.tsx
DipperTheDan: I did consider using TS initially but I figured for a POC just using plain JS would be better. When we come to actually implement it, we could do it in TS. I don't think it would be too difficult to do.

src/components/menu/menu-item/menu-item.style.ts
DipperTheDan: Do you mean revert everything excluding the changes:

```
padding-right: ${(props) =>
              parsePadding(props as Partial<PaddingProps>).padding};
```
And these:

```
right: ${(props) =>
            parsePadding(props as Partial<PaddingProps>).iconSpacing};
```

And also keep the new imports?

docs/styled-system-replacement-poc.mdx
```diff
@@ -0,0 +1,141 @@
...
+`styled-system` was a library for handling styling as props, but it comes with some drawbacks:
+- Increased bundle size due to its large footprint (especially with Box).
+- Difficulty in tree-shaking unused styles.
+- No longer supported or maintained (Website seems to have been hijacked by a casino/gambling scam).
+
+By creating custom style functions using just plain old JavaScript, we can tailor the implementation to our specific needs and optimise for performance.
+
+## Implementation
+
+The implementation involves creating utility functions that mimic the behavior of `styled-system` for common use cases, such as spacing, layout, and flexbox.
```
edleeks87:
```suggestion
The implementation involves creating utility functions that mimic the behaviour of `styled-system` for common use cases, such as spacing, layout, and flexbox.
```

src/components/dips-box/utils/spacing.js
edleeks87: Yeah for me it would be a *nice* to have rather than *need* at this stage. For me the first pass must have a replacement for `styled-stystem` props that maintains the `string` and `number` (where applicable) interface

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+
+  return String(value);
+}
+
+/* Flexbox and Layout support */
+export function getFlexboxStyles(props) {
+  const style = {};
+  for (const key in flexboxKeys) {
+    const value = props[key];
+    if (value == null) continue;
```
edleeks87:
```suggestion
    if (!value) continue;
```

you could just check for any falsy value as it will only be null or `""`

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+  }
+  return style;
+}
+
+/** Apply flexbox styles
+ * This can be implemented by doing:
+ * `${(props) => flexboxCss(props)}`
+ * in a styled component.
+ */
+// Number has been added to the logic here in case we choose to support numbers. For now flexGrow and flexShrink are string values
```
edleeks87: comment: For me strings would be enough here, the only one I'd support numbers on is the spacing values and even on that I'm a bit on the fence as it's a bit of an opaque interface that requires documenting that the number is multiplied by a factor of 8 (mainly keeping them to lessen the impact of the breaking change)

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+// but they are supposed to be numbers in CSS.
+export function flexboxCss(props) {
```
edleeks87: suggestion: unless we need the `get...` functions to be separate you could combine them so that the returned object is stringified rather than separating the concerns unnecessarily

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+/** format a single value for output */
+function formatValue(cssKey, value) {
+  if (typeof value === "number") {
+    // If it's suffix-less return as raw number
+    if (PropertyWithoutSuffix.has(cssKey)) return String(value);
+
+    // 0 is special: "0" is valid everywhere
+    if (value === 0) return "0";
+
+    // treat numbers in [0, 1] as percentages for percent-capable props
```
edleeks87: comment: imo this part of the interface was always terrible. Happy to go with the flow so if general consensus is to keep it to limit the impact of the breaking change then I can live with it. For me it's not very intuitive how to pass a float to achieve a percentage, for me a string suffices etc

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+  "right",
+  "top",
+  "bottom",
+  "flex-basis",
+]);
+
+/** format a single value for output */
+function formatValue(cssKey, value) {
+  if (typeof value === "number") {
+    // If it's suffix-less return as raw number
```
edleeks87, line 132: question: this has a suffix though?

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+    const token = sizingLetters[1];
+    const size = sizeMap[token] ?? 0;
+    return applyScale(sign * size, scale);
+  }
+
+  // If value is something like '1rem', 'auto', etc., pass through
+  return value;
+}
+
+export function getSpacingStyles(props, scale = 8, sizeMap = spacingSizeMap) {
```
edleeks87: comment: same about combining this with the css one below

src/components/dips-box/utils/spacing.js
```diff
@@ -0,0 +1,299 @@
...
+/** If we want to add TS support, we can use this. For now, leaving out to avoid complexity in JS life.
+ * type Scale =
+ *   | number
+ *   | string
+ *   | ((n: number) => number | string)
+ *   | Array<number | string>;
+ */
+
+/** Apply the chosen scale to a numeric step/multiplier */
+function applyScale(n, scale) {
```
edleeks87: comment: I'd be pretty keen not to support this anymore in the interest of simplifying the implementation considerably. As far as I know this doesn't work in carbon anyway so I don't think removing it will have much impact. If I'm wrong about that then I still think it's worth discussing if there's any value persisting it

src/components/dips-box/utils/spacing.js
DipperTheDan: What I'll do for this then is make a copy of the branch in it's current form (with the sizing map) and keep it to one side for when/if we want to implement this. I've proven we can support it with the way the POC currently is, with no great deal of change required to the core spacing functions.

src/components/dips-box/utils/spacing.js
DipperTheDan: Strings alone would be easier to support, so I'm all for this. I did this number stuff as we have unit tests that test numbers for `flexGrow` and a few others. I can amend the data we pass in when I update the utils like @Parsium has requested earlier on in the PR.

src/components/dips-box/utils/spacing.js
DipperTheDan: I can mash them together no problem. For me it was easier to have them separated for debugging purposes, that was the main reason I did it this way.

src/components/dips-box/utils/spacing.js
DipperTheDan: Yeah, the idea behind this was to minimise the breaking change impact. It was also to demonstrate that if needs be, we can support with existing float stuff too if enough consumers shout about it. I'm happy to remove support for the percentage stuff and we can just document how to do percentages in the docs for the new approach.

src/components/dips-box/utils/spacing.js
DipperTheDan: The main idea for this was that DS have different spacing values based on `component` or `layout` styling. The idea being that `layout` might be a multiplier of 8 and `component` might be of 2. They were also keen to support passing token values as multipliers too. I'm more than happy to kick this can down the road for now though (it's been demonstrated here that we can add this functionality in to the core functions of this _thing_) and we just support what we currently support and implement this as a feature/improvement later on.

edleeks87 (dismissed):
Approving the proposed changes demonstrated in this POC

src/components/dips-box/utils/spacing-types.ts
```diff
@@ -0,0 +1,113 @@
...
+  /** Short hand `flex` property */
+  flex?: string;
+  /** Short hand `align-self` property */
+  alignSelf?: string;
+  /** Short hand `order` property */
+  order?: string;
+}
+
+export interface LayoutProps {
+  /** Short hand `width` property */
```
nineteen88, line 85: comment(minor, non-blocking): these props aren't short hand are they?

nineteen88 (dismissed):
Tiny minor comment overall. Looks really good to me. Good work!

nicktitchmarsh (dismissed):

PR #7601, opened 2025-10-30
test(play-functions): remove userinteractionpause

paulrobinson88:
### Proposed behaviour

- Removes `userInteractionPause` from Play Functions due to issues with Chromatic snapshot consistency.
- Adds `expect await` assertions to interaction stories so we can have more accurate Pass/Fail results on these tests.
- Disables Chromatic snapshots for some stories where the state is already captured.

This applies to the components below - 
- Accordion
- Action Popover
- Adaptive Sidebar
- Advanced Color Picker
- Breadcrumbs
- Checkbox
- Date Range
- Decimal
- Menu
- Multi-Action button
- Multi select
- Simple select
- Split button
- Vertical Menu
- Responsive Vertical Menu

### Current behaviour

- `userInteractionPause` is used throughout a number of component interaction stories.
- Some tests pass regardless of what the state of the story is due to no assertions.
- Snapshots are taken of a state which is already captured in other stories.

### Checklist

#### QA

### Testing instructions
- Ensure interaction tests pass as expected
- Check Chromatic snapshots are taken as expected


src/components/decimal/decimal-interactions.stories.tsx
```diff
@@ -51,14 +50,15 @@ export const DefaultExample: Story = {
     });
 
     await userEvent.click(decimalInput);
-    await userInteractionPause(80);
     await userEvent.keyboard("357");
-    await userInteractionPause(120);
     await userEvent.keyboard("{tab}");
 
-    expect((decimalInput as HTMLInputElement).value).toBe("357.00");
+    await expect((decimalInput as HTMLInputElement).value).toBe("357.00");
```
Parsium: **suggestion(non-blocking)**: it seems the [`@storybook/test` package supports a `toHaveValue` assertion](https://storybook.js.org/docs/writing-tests/interaction-testing#asserting-with-expect), would we be able to use it? This would allow us to avoid using a type assertion:


```suggestion
    await expect(decimalInput.value).toHaveValue("357.00");
```

src/components/decimal/decimal-interactions.stories.tsx
```diff
@@ -144,6 +147,6 @@ export const FocusedExample: Story = {
 
     await userEvent.click(decimalInput);
 
-    expect(document.activeElement).toBe(decimalInput);
+    await expect(document.activeElement).toBe(decimalInput);
```
Parsium: **suggestion(non-blocking)**: what do you think about using the `toHaveFocus` assertion? This would make the error message more specific if the test fails:

```suggestion
    await expect(decimalInput).toHaveFocus();
```

src/components/date-range/date-range-interactions.stories.tsx
Parsium, line 1: **suggestion(non-blocking)**: what do you think about using the `toBeInvalid` and `toHaveValue` assertions over `toBe`?

If there a specific assertion exists from `@storybook/test`, it might be worth using it over `toBe` to help make the errors from failing tests more specific, which would be useful for debugging.

src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -87,6 +86,11 @@ export const ClickToOpen: Story = {
     const accordionToggle = canvas.getAllByRole("button");
 
     await userEvent.click(accordionToggle[0]);
+    await userEvent.click(accordionToggle[0]);
+    await userEvent.click(accordionToggle[0]);
+    await expect(
+      await within(document.body).findByText("Content1"),
+    ).toBeVisible();
```
Parsium: **suggestion**: this should help avoid the flaky behaviour observed with `toBeVisible()`. Given the disclosure content is animated, so there is a chance the text won't be visible when the assertion is executed.  

```suggestion

    await expect(
      await within(document.body).findByText("Content1"),
    ).toBeInTheDocument();
```

src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -9,7 +9,6 @@ import Textbox, { TextboxProps } from "../textbox";
 
 import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
 import DefaultDecorator from "../../../.storybook/utils/default-decorator";
-import userInteractionPause from "../../../.storybook/utils/user-interaction-pause";
```
Parsium, line 12: **suggestion**: would we be able to remove the `userInteractionPause` util? This would help prevent accidentially re-introducing it in test code.

paulrobinson88: I wasn't seeing this when I pushed the commit and not seeing in the new one, so hopefully that's removed now.

src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -310,15 +261,16 @@ export const NestedComponentInteractions: Story = {
...
 
-    await userEvent.click(accordionToggle[0]);
-
-    await userInteractionPause(500);
+    await userEvent.click(accordionToggle[0], { delay: 100 });
 
     const textboxes = canvas.queryAllByRole("textbox");
     const textbox = textboxes[0];
     await userEvent.type(textbox, "Text input in Accordion", { delay: 100 });
     await userEvent.click(accordionToggle[0], { delay: 1000 });
```
sianford: Can you reduce this delay to 100 from 1000, and also on the next line too? It seems to take a noticeably long time when it runs.

paulrobinson88: Certain tests needed longer delays for the animation to prevent flakiness but seems that one is fine with the lower delay so I've amended 2 lines to `delay:100`

src/components/accordion/accordion-interaction.stories.tsx
Parsium: Ah apologies @paulrobinson88 - I actually meant removing the util entirely from the repo:

```diff
-.storybook/utils/user-interaction-pause.ts
```

This would help prevent us accidentally using it again elsewhere.

src/components/action-popover/action-popover-interaction.stories.tsx
```diff
@@ -202,5 +201,57 @@ export const CloseOnEscPress: Story = {
     ),
   ],
 };
-
+CloseOnEscPress.parameters = { chromatic: { disableSnapshot: true } };
 CloseOnEscPress.storyName = "Close On ESC Press";
+
+export const SubmenuClick: Story = {
```
sianford: I think the name of this story is misleading. There is no `Click` action to open the submenu. It is opened on hover.

src/components/accordion/accordion-interaction.stories.tsx
Parsium: @paulrobinson88 @sianford this might be something for us to look into later. If we start to frequently need delayed actions in play functions, and its drastically impacting the runtime of Chromatic checks on PRs, we can prioritise this.

Ultimately, to avoid delays we'd either need to:

1. Explicitly asserting a condition we know is true if and only if the element has finished animating (e.g. element has specific styling once stable)
2. Using fake timers for the interaction stories, so we essentially fast forward the animations

src/components/breadcrumbs/breadcrumbs-interactions.stories.tsx
```diff
@@ -97,18 +92,19 @@ export const LongTrailBreadcrumbs: Story = {
...
-    expect(links[0]).toHaveFocus();
+    await expect(links[0]).toHaveFocus();
 
     await userEvent.tab();
-    await userInteractionPause(120);
-    expect(links[1]).toHaveFocus();
+    await expect(links[1]).toHaveFocus();
   },
 };
+LongTrailBreadcrumbs.parameters = {
```
sianford: Do we need this story at all? I don't think it is providing anything that the default or dark background one does. You could may be add another breadcrumb to the dark background one and take a snapshot there which would be of value.

src/components/decimal/decimal-interactions.stories.tsx
Parsium: **suggestion**:

```suggestion
    await expect(decimalInput).toHaveValue("357.00");
```

src/components/decimal/decimal-interactions.stories.tsx
```diff
@@ -77,14 +77,15 @@ export const frFRLocaleExample: Story = {
     });
 
     await userEvent.click(decimalInput);
-    await userInteractionPause(80);
     await userEvent.keyboard("442");
-    await userInteractionPause(120);
     await userEvent.keyboard("{tab}");
 
-    expect((decimalInput as HTMLInputElement).value).toBe("442,00");
+    await expect((decimalInput as HTMLInputElement).value).toBe("442,00");
```
Parsium: **suggestion**:

```suggestion
    await expect(decimalInput).toHaveValue("442,00");
```

src/components/decimal/decimal-interactions.stories.tsx
```diff
@@ -103,14 +104,15 @@ export const WithPrecisionExample: Story = {
     });
 
     await userEvent.click(decimalInput);
-    await userInteractionPause(80);
     await userEvent.keyboard("90210");
-    await userInteractionPause(120);
     await userEvent.keyboard("{tab}");
 
-    expect((decimalInput as HTMLInputElement).value).toBe("90,210.00000");
+    await expect((decimalInput as HTMLInputElement).value).toBe("90,210.00000");
```
Parsium: **suggestion**:

```suggestion
    await expect(decimalInput).toHaveValue("90,210.00000");
```

src/components/decimal/decimal-interactions.stories.tsx
```diff
@@ -123,14 +125,15 @@ export const DecimalInputClearedExample: Story = {
     });
 
     await userEvent.click(decimalInput);
-    await userInteractionPause(80);
     await userEvent.keyboard("{selectall}{backspace}");
-    await userInteractionPause(120);
     await userEvent.keyboard("{tab}");
 
-    expect((decimalInput as HTMLInputElement).value).toBe("0.00");
+    await expect((decimalInput as HTMLInputElement).value).toBe("0.00");
```
Parsium: **suggestion**:

```suggestion
    await expect(decimalInput).toHaveValue("0.00");
```

src/components/accordion/accordion-interaction.stories.tsx
sianford: @Parsium would there ever be a use case when we'd want to mimic a user pause? I can't think of one, at least not for this type of test, so I think it could be removed.

src/components/select/multi-select/multi-select-interaction.stories.tsx
sianford: I know this story was already in before you made your changes, but I'm a bit confused by the story name of "HoverItem' when the hover action is never tested. Can it be changed to may be "Keyboard Interaction" or something like that?

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-interaction.stories.tsx
```diff
@@ -253,11 +250,17 @@ export const NavigateToSecondaryResponsive: Story = {
     }
 
     await userEvent.tab(); // focus toggle
-    await userEvent.keyboard("{Enter}"); // open menu
-    await userInteractionPause(1000);
-
-    await userEvent.tab(); // close button
-    await userEvent.tab(); // secondary item
+    await userEvent.keyboard("{Enter}", { delay: 500 }); // open menu
```
sianford, line 257: Can the delays possiby be replaced with assertions?

src/components/vertical-menu/vertical-menu-interaction.stories.tsx
```diff
@@ -57,8 +56,10 @@ export const ClickToOpenSubmenu: Story = {
 
     const canvas = within(canvasElement);
     const submenuToggle = canvas.getByRole("button", { name: /item 3/i });
-
-    await userEvent.click(submenuToggle);
+    await userEvent.click(submenuToggle, { delay: 200 });
```
sianford: Is this delay required with there being an assertion on the next line that `ChildItem 1` is visible?

src/components/breadcrumbs/breadcrumbs-interactions.stories.tsx
paulrobinson88: I'd agree, that's removed.

src/components/select/multi-select/multi-select-interaction.stories.tsx
paulrobinson88: The `hover` is applied at the end in the parameters on two items in the dropdown but I'd agree it should really match what the interactions are, I've changed this.

src/components/vertical-menu/responsive-vertical-menu/responsive-vertical-menu-interaction.stories.tsx
paulrobinson88: I've added in an extra assertion for when the menu initially opens but can't avoid having the delay to account for the animation, otherwise the interaction fails to find the button element.

src/components/vertical-menu/vertical-menu-interaction.stories.tsx
paulrobinson88: I've tested removing this and it seems to still work fine.

sianford (approved):

src/components/accordion/accordion-interaction.stories.tsx
Parsium: @sianford yeah, ideally we should be using explicit assertions for this 👍🏼

sianford (approved):

Parsium (approved):

tomdavies73 (approved):

PR #7602, opened 2025-10-30
fix(flat-table): prevent sorting indicators from impacting text wrapping

### LINKED ISSUE #7586: [FlatTable] Sort header placeholder impacts wrapping when not shown
### Description

When using a sortable FlatTable with long header texts, the wrapping behaves oddly as screen size is reduced.

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-vjwnakan?file=src%2FApp.tsx

### Steps to reproduce

1. Open SB demo
2. Resize render pane until wrapping occurs

_No response_

### Suggested solution

The `sort-placeholder` element is causing extra spacing to be added to the header which impacts wrapping. If no sort icon is shown, we should consider whether it's needed at all (it's currently fixed to `inline-block`, but if it **must** remain in the DOM then maybe switching between that and `none` would be better).

### Carbon version

157.4.0

### React version

v18

### Design tokens version

4.35.0

### Relevant browsers

Chrome, Firefox

### Relevant OSs

MacOS

### Assistive technologies

_No response_

### Additional context

_No response_

### Confidentiality

damienrobson-sage:
fixes #7586 

### Proposed behaviour

Use the correct semantic element for the sorting buttons, and ensure that the table headers respect the spacing appropriately

### Current behaviour

When using a sortable FlatTable with long header texts, the wrapping behaves oddly as screen size is reduced. The current implementation of `Sort` in `FlatTable` is also built using a clickable `div` as opposed to a button.

### Checklist

#### QA

### Testing instructions

A test story has been added under `FlatTable`: `Extended Column Sorting` which can be used to test the changes.

### DIFF
--- a/playwright/components/flat-table/index.ts
+++ b/playwright/components/flat-table/index.ts
@@ -40,7 +40,7 @@ export const flatTableBodyRows = (page: Page) =>
 export const flatTableClickableRow = (page: Page, index: number) =>
   page.locator(FLAT_TABLE_COMPONENT).locator("tbody tr").nth(index);
 export const flatTableSortable = (page: Page) =>
-  page.locator(FLAT_TABLE_COMPONENT).locator("thead tr th div [role=button]");
+  page.locator(FLAT_TABLE_COMPONENT).locator("thead tr th button");
 export const flatTableCell = (page: Page, index: number) =>
   page.locator(FLAT_TABLE_CELL).nth(index);
 export const flatTableCheckboxCell = (page: Page, index: number) =>
@@ -73,8 +73,8 @@ export const flatTableCurrentPageInput = (page: Page) =>
 
 export const flatTableRowHeader = (page: Page) =>
   page.locator(FLAT_TABLE_STICKY_ROW);
-export const flatTableHeaderCellsIcon = (page: Page) =>
-  flatTableHeader(page).locator("th > div > div").locator("span");
+export const flatTableHeaderCellsButtonIcon = (page: Page) =>
+  flatTableHeader(page).locator("[data-element='sort-icon']");
 export const flatTableExpandableIcon = (page: Page, index: number) =>
   flatTableCell(page, index).locator(FLAT_TABLE_ICON);
 
--- a/src/components/flat-table/flat-table-test.stories.tsx
+++ b/src/components/flat-table/flat-table-test.stories.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useMemo, useState } from "react";
 import { action } from "@storybook/addon-actions";
 import {
   FlatTable,
@@ -13,6 +13,7 @@ import {
   FlatTableHeaderProps,
   FlatTableRowHeaderProps,
   FlatTableCellProps,
+  Sort,
 } from ".";
 import Button from "../../../src/components/button";
 import Box from "../../../src/components/box";
@@ -39,20 +40,6 @@ import PopoverContainer from "../popover-container";
 
 export default {
   title: "Flat Table/Test",
-  includeStories: [
-    "FlatTableStory",
-    "ExpandableWithLink",
-    "ExpandableWithActionPopover",
-    "SortableStory",
-    "SubRowsAsAComponentStory",
-    "FlatTableSizeFocus",
-    "FlatTableInsideDrawer",
-    "FlatRowHeaderWithNoPaddingAndButtons",
-    "FlatTableThemesWithAlternateHeaderBackground",
-    "FlatTableThemesWithStickyHead",
-    "FlatTableWithStickyHeadAndFooter",
-    "WithLongRowHeader",
-  ],
   parameters: {
     info: { disable: true },
     chromatic: {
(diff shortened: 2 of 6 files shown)


playwright/components/flat-table/index.ts
```diff
@@ -74,7 +74,11 @@ export const flatTableCurrentPageInput = (page: Page) =>
 export const flatTableRowHeader = (page: Page) =>
   page.locator(FLAT_TABLE_STICKY_ROW);
 export const flatTableHeaderCellsIcon = (page: Page) =>
-  flatTableHeader(page).locator("th > div > div").locator("span");
+  flatTableHeader(page).locator("th > div > div > span[data-component='icon']");
+export const flatTableHeaderCellsButtonIcon = (page: Page) =>
+  flatTableHeader(page).locator(
+    "th > div > button > span[data-component='icon']",
+  );
```
nuria1110: **suggestion:** Could we locate these elements by targeting them using `data-element` or `data-role` instead? This can easily break if the component structure changes and will also help with clarity on what component is being located.

src/components/flat-table/flat-table-test.stories.tsx
```diff
@@ -1239,3 +1226,146 @@ export const WithLongRowHeader = () => {
...
+export const ExtendedColumnSorting = (args: FlatTableProps) => {
+  const initialHeaders = [
+    { name: "client", label: "Client", isActive: true },
+    { name: "region", label: "Region", isActive: false },
+    { name: "total", label: "Total (£)", isActive: false },
+    { name: "orders", label: "Orders", isActive: false },
+    { name: "lastPurchase", label: "Last Purchase", isActive: false },
+  ];
+
+  const data: TableRowData[] = useMemo(() => {
```
Parsium, line 1242: **question(non-blocking)**: does this need to be memoised? I'm guessing it's because the component renders it, which is needed for it to appear in the code snippet. Typically though, I'd expect consumers wouldn't define initial data like this. 

```
  const data: TableRowData[] = useMemo(() => {
```

src/components/flat-table/sort/sort.style.ts
```diff
@@ -1,43 +1,64 @@
...
 
-const StyledSortIcon = styled(Icon)<StylesSortIconProps>`
-  padding-left: var(--spacing075);
-  color: ${({ iconColor }) => `var(${iconColor})`};
+    span[data-component="icon"] {
+      color: ${colorTheme === "dark"
+        ? "var(--colorsActionMinorYang100)"
+        : "var(--colorActionMinor500)"};
+      margin-bottom: 1px;
+    }
```
Parsium, line 68: **suggestion**: what do you think of defining these styles within `Icon`? This help ensure `Icon` has full control over its own styling.

`Icon`'s `className` prop could help us achieve this:

```
// icon.style.ts
cosnt StyledIcon = styled.span`
  .flat-table--sort-icon {
      color: ${colorTheme === "dark"
        ? "var(--colorsActionMinorYang100)"
        : "var(--colorActionMinor500)"};
      margin-bottom: 1px;
  }
`

// sort.component.tsx
const icon = () =>
    sortType ? (
    <Icon className="flat-table--sort-icon" type={sortType === "ascending" ? "sort_up" : "sort_down"} />
  ) : (
    <span data-role="sort-placeholder" />
  );
```

src/components/flat-table/flat-table-test.stories.tsx
damienrobson-sage: Correct, it doesn't need to be memoised, this is all just because of how the story is being set up, etc. Customers wouldn't need to rely on this approach.

src/components/flat-table/sort/sort.style.ts
damienrobson-sage: I get where you're coming from but I'd rather keep the change focused within `Sort` so that I don't need to start making changes to Icon that would see extra props being added to it and whatnot; `Icon` has no concept of `colorTheme`, for example, so I'd need to add that in, and then add tests for coverage, etc. and the whole ticket's scope inflates beyond what it needs to be.

Parsium: No worries - happy to defer this for now, given it's a small change.

In the near future, it'd be good for us to move away from overriding styles of public components, in favour of a keeping these components decoupled. This would reduce the complexity of our styling and make public components like `FlatTable` less prone to regressions if we change their current dependents.

playwright/components/flat-table/index.ts
```diff
@@ -74,7 +74,9 @@ export const flatTableCurrentPageInput = (page: Page) =>
 export const flatTableRowHeader = (page: Page) =>
   page.locator(FLAT_TABLE_STICKY_ROW);
 export const flatTableHeaderCellsIcon = (page: Page) =>
-  flatTableHeader(page).locator("th > div > div").locator("span");
+  flatTableHeader(page).locator("th > div > div > span[data-component='icon']");
```
Parsium: **suggestion**: could we use a test attribute here, similar to what @nuria1110 suggested earlier? This would help make our tests more robust by being less dependent on a complex page structure.

damienrobson-sage: As discussed, I've removed this selector and accompanying test given it was only checking that a passed-in, user-specified icon rendered correctly

nuria1110 (dismissed):

PR #7606, opened 2025-10-31
 chore: add new commit script using inquirer 

tomdavies73:
### Proposed behaviour

Adds a new commit script which runs on `npm run commit`. Follows a similar structure to commitizen which was used previously but was removed due to a lack of support.

### Current behaviour

Currently, there is no commit script, all commits are done manually.

### Checklist

#### QA


scripts/commit/commit.mjs
```diff
@@ -0,0 +1,218 @@
...
+  { value: 'docs', name: 'docs: Documentation only changes' },
+  { value: 'style', name: 'style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)' },
+  { value: 'refactor', name: 'refactor: A code change that neither fixes a bug nor adds a feature' },
+  { value: 'perf', name: 'perf: A code change that improves performance' },
+  { value: 'test', name: 'test: Adding missing tests or correcting existing tests' },
+  { value: 'build', name: 'build: Changes that affect the build system or external dependencies' },
+  { value: 'ci', name: 'ci: Changes to CI configuration files and scripts' },
+  { value: 'chore', name: 'chore: Other changes that don\'t modify src or test files' },
+  { value: 'revert', name: 'revert: Reverts a previous commit' },
+];
```
damienrobson-sage, line 17: **question (non-blocking)**: probably one for wider discussion, but now would be a good time to clean these up. Get rid of any we're not using, or add extra ones to cover off additional cases, etc. Not essential, though; these ones work perfectly fine as-is 😄

tomdavies73: Good shout mate, I just made this consistent with what we had used previously, we could definitely remove ones we don't use but the impact of keeping it is probably quite negligible. 

A later review on extra commits we'd need could definitely be useful, but I've yet to find an instance where the above list of commits hasn't covered what we need. But we can always just add more in the future now as we have control 👍

damienrobson-sage: Absolutely, just something that sprang to mind whilst I was reading the list. Nothing needs changing, s'all good

nineteen88: FYI they are derived from the conventional commits spec:

https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional

So, I personally believe we should keep it aligned to that.

nineteen88 (approved):

damienrobson-sage (approved):

PR #7607, opened 2025-11-01
feat(tabs): add support for programmatic tab focus to new Tabs

damienrobson-sage:
### Proposed behaviour

Adds an `imperativeHandle` to the `TabList` component that allows users to programmatically select specific `Tabs` via `id`

### Current behaviour

Users do not have the ability to programmatically focus the newer `Tabs` elements.

### Checklist

#### QA

### Additional context

The handle lives within the `TabList` component to ensure that the programmatic focus handles on a more global context, rather than living on individual tabs.

### Testing instructions

A new Storybook entry, `Programmatic Focus` has been added to the `Tabs` documentation


src/components/tabs/__next__/tabs.types.ts
```diff
@@ -7,6 +7,10 @@ export interface TabContextProps {
   tabId?: string;
 }
 
+export interface TabListHandle {
```
tomdavies73: **suggestion**:

Instead of an `interface`, I think `type` may be more suitable for defining a function. 

Also, it might be worth using some sort of comment to outline the function too, similar to what already exists in the legacy version of the component:

```
export type TabsHandle = {
  /**
   * Programmatically focus on a specific tab.
   * @param tabId - The ID of the tab to focus. Must match the `tabId` prop of the target `Tab` component.
   */
  focusTab: (tabId: string) => void;
} | null;
```

src/components/tabs/__next__/tabs.types.ts
```diff
@@ -7,6 +7,14 @@ export interface TabContextProps {
   tabId?: string;
 }
 
+export type TabListHandle = {
```
tomdavies73: **suggestion**: all mdx docs etc. are still pointing toward an import of `TabsHandle` and that is the name currently used in the legacy component.

We can update the mdx to match this or change this to `TabsHandle`. Sorry missed this on my first review pass

DipperTheDan (approved):

PR #7610, opened 2025-11-04
fix(date-picker): ensure calendar picker is not hidden by dialog overflows

### LINKED ISSUE #7593: DateInput overlapped by Dialog container
### Description

Place a DateInput control inside a Dialog since version 155.0.0. The Dialog cover the DateInput when calendar is opened

### Reproduction

https://stackblitz.com/edit/parsium-carbon-starter-aki8qpzl?file=package.json,src%2Fmain.tsx

### Steps to reproduce

_No response_

_No response_

### Suggested solution

_No response_

### Carbon version

155.0.0

### React version

v18

### Design tokens version

_No response_

### Relevant browsers

Chrome

### Relevant OSs

Windows

### Assistive technologies

_No response_

### Additional context

_No response_

### Confidentiality

damienrobson-sage:
fixes #7593 

### Proposed behaviour

Adjust the `useFloating` hook to use `fixed` positioning and set a `z-index` of `1001` to render it above the sticky footer

### Current behaviour

If a `Date` input is rendered within a `Dialog` that uses a sticky footer, the calendar appears above the modal but behind the footer when opened.

### Checklist

#### QA

### Testing instructions

Use the DialogOverflow test story under DateInput in Storybook to check the correct appearance.

### DIFF
--- a/src/components/badge/badge-test.stories.tsx
+++ b/src/components/badge/badge-test.stories.tsx
@@ -8,7 +8,7 @@ import SplitButton from "../split-button";
 import { Menu, MenuItem } from "../menu";
 import Icon from "../icon";
 import Portrait from "../portrait";
-import { Tabs, Tab } from "../tabs/";
+import { Tabs, TabList, Tab, TabPanel } from "../tabs/__next__";
 import Typography from "../typography";
 
 const meta: Meta<typeof Badge> = {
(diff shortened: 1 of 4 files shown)


src/hooks/__internal__/useFloating/useFloating.ts
```diff
@@ -63,6 +63,12 @@ const useFloating = ({
             left: `${x}px`,
             top: `${y}px`,
             position: strategy,
+            /**
+             * The z-index of the popover needs to be adjusted to ensure that
+             * the floating element appears over the top of other elements on
+             * the same "layer" correctly (i.e. a date picker in a dialog)
+             * */
+            "z-index": 1001,
```
tomdavies73, line 71: **thought**: If there was a better way to do this I'd prefer it, but yeah, I don't think there is really. As we've used fixed z-index's elsewhere, this should be fine. But it would be nice if we could pivot away from z-index reliance in general in the future 👍

src/components/date/date-test.stories.tsx
```diff
@@ -363,3 +366,44 @@ AutoFocus.parameters = {
   themeProvider: { chromatic: { fourColumnLayout: true, theme: "sage" } },
   chromatic: { viewports: [1800], disableSnapshot: false },
 };
+
+export const DialogOverflow = () => {
```
tomdavies73: **suggestion:**: I have no problem with this being a test story, and i can see that it currently does not capture a snapshot, which is fine. 

However, I think there is some value in capturing the picker being open at the same time as the `Dialog`. We could do this via an interaction story etc. and it will give a solid baseline against the fix. Therefore, if anything is changed in the future (which can be an issue with fixed "magic" number z-index's) we will know about it 👍

src/hooks/__internal__/useFloating/useFloating.ts
Parsium, line 71: **suggestion**: might be worth adding this new value into the [base theme](https://github.com/Sage/carbon/blob/FE-7515-date-picker-dialog-overflow/src/style/themes/base/base-theme.config.ts#L54-L64). Just so we have it in a centralised location alongside the other z-indices we manage:

damienrobson-sage: I get where you're coming from but as this isn't a style file and the hook doesn't already have access to the theme I'm reticent to add support in for getting/reading the theme, and to add this as a hard-coded value when it's single use (and we want to move away from it at some point if possible).

It's also worth considering that we have a lot of other seemingly random z-index values floating around (if you'll pardon the pun).

Parsium (dismissed):

tomdavies73 (approved):

PR #7624, opened 2025-11-17
feat(button): align with the new DS styling and functionality

damienrobson-sage:
### Proposed behaviour

- Introduce the "next" version of Button that consumers can start to import and use in place of the existing `Button` implementation
- Deprecate `darkBackground`, `gradient-grey` and `gradient-white` ButtonType from current button implementation
- Deprecate `isWhite` prop from current button implementation
- Deprecate `destructive` prop from current button implementation
- Deprecate `subtext` prop from current button implementation
- Deprecate `buttonType` prop from current button implementation

### Current behaviour

The `Button` component is not aligned with the latest DS requirements.

### Checklist

#### QA

### Additional context

- The existing Button component's documentation has been moved to the `Deprecated` section of Storybook and points users towards the newer documentation
- The newer documentation references the older documentation for consumers that still require it.

### Testing instructions

Use the new Storybook documentation to check the new functionality


src/components/button/__next__/button.style.ts
nuria1110, line 1: **suggestion (non-blocking):** Since this component in particular has a large number of prop variations, it could be a bit cleaner to separate out the style maps into a separate `config` file so that the style file is a bit more readable.

src/components/button/__next__/button.style.ts
```diff
@@ -0,0 +1,645 @@
...
+    borderRadius: "24px",
+    fontSize: "16px",
+    fontWeight: "var(--fontWeights500)",
+    height: "48px",
+    iconOnlyWidth: "48px",
+    lineHeight: "150%",
+    paddingVertical: "6px",
+    paddingHorizontal: "16px",
+  },
+};
```
nuria1110: **comment:** It looks like some of the properties are consistent throughout all of the sizes and could be just passed directly to the styled component, e.g. `lineHeight`, `fontWeight`, etc. 

Think we could also be using new design-tokens for most of these values here if possible.

src/components/button/__next__/button.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+Icons.storyName = "Icons";
+
+export const Inverse: Story = (args: ButtonProps) => {
+  return (
+    <Box backgroundColor="#333" p={2}>
+      <Button {...args} inverse />
+    </Box>
+  );
+};
+Inverse.storyName = "Inverse";
```
nuria1110: **suggestion:** Think it would be good to capture the different `variantTypes` for inverse, at least just as a test story so that we can snapshot with chromatic.

src/components/button/button.component.tsx
edleeks87, line 1: Question: instead of the the props being marked isn't this whole component not being deprecated?

tsconfig-build.json
```diff
@@ -4,7 +4,11 @@
     "declaration": true,
     "noEmit": false,
     "rootDir": "src/",
-    "emitDeclarationOnly": true
+    "emitDeclarationOnly": true,
+    "types": [
```
edleeks87: comment: this file already extends the `tsconfig.json` so we shouldn't need to define the `types` again. Also I'm not sure why we'd need to define them for the build as test files and jest isn't part of the distributed files

src/components/button/__next__/button.style.ts
```diff
@@ -0,0 +1,318 @@
...
+  justify-content: center;
+  outline-offset: 0;
+  text-decoration: none;
+  vertical-align: middle;
+
+  ${({ allowMotion, disabled, inverse, size, variant, variantType }) => {
+    if (inverse) {
+      return getCSSForInverseStyle({
+        disabled,
+        size: size as NonNullable<Size>,
```
edleeks87: comment: these don't need to be coerced when the type is already set to NonNullable above

```
return getCSSForInverseStyle({
        disabled,
        size,
        variantType,
...

File: src/components/button/__next__/button.style.ts [hunk 5]
```diff
@@ -0,0 +1,318 @@
...
+          disabled,
+          size: size as NonNullable<Size>,
+        })}
+      `;
+    }
+
+    /**
+     * XS should fall back to the "default secondary" colour scheme
+     * */
+    /* istanbul ignore if */
```
edleeks87: question: do we need to ignore here, is there no way to create some tests to hit these?

src/components/button/__next__/button.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+  /** The HTML type that this button should use. */
+  type?: "button" | "reset" | "submit";
+  /** The target attribute to pass to the button when acting as a link. Ignored if `href` is not set. */
+  target?: string;
+  /** The rel attribute to pass to the button when acting as a link. Ignored if `href` is not set. */
+  rel?: string;
+  /** The variant of the button. */
+  variant?: "default" | "destructive" | "ai";
+  /** The variant type of the button. */
+  variantType?: "primary" | "secondary" | "tertiary" | "subtle";
```
edleeks87: suggestion (non-blocking): you could avoid the need to make these props NonNullable if you defined the types separately and exported them and used them with the `?` like you have etc.

```
export type Size = "xs" | "small" | "medium" | "large";
export type Variant = "default" | "destructive" | "ai";
export type VariantType = "primary" | "secondary" | "tertiary" | "subtle";
...
size?: Size;
...
variant?: Variant;
  /** The variant type of the button. */
variantType?: VariantType;

File: src/components/button/__next__/button.component.tsx [hunk 7]
```diff
@@ -0,0 +1,262 @@
...
+  /** The text that the button displays. */
+  children?: ReactNode;
+  /** Flag to indicate that the button is disabled. */
+  disabled?: boolean;
+  /** Flag to indicate that the button can be full-width. */
+  fullWidth?: boolean;
+  /** The click destination of the button. Using this will transform the button into an anchor. */
+  href?: string;
+  /** Defines the position of the chosen icon relative to the children. */
+  iconPosition?: "left" | "right";
```
edleeks87: question: with the composable `children` do we need to have an `iconPosition` prop anymore? If we do should `children` just be a string?

src/components/button/__next__/button.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+        },
+      }),
+      [focusButton],
+    );
+
+    const allowMotion = useMediaQuery(
+      "screen and (prefers-reduced-motion: no-preference)",
+    );
+
+    const hasValidChildren = children !== undefined && children !== false;
```
edleeks87: comment (non-blocking): checking if they've passed false seems like overkill to me

src/components/button/__next__/button.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+      event:
+        | React.MouseEvent<HTMLAnchorElement>
+        | React.MouseEvent<HTMLButtonElement>,
+    ) => {
+      buttonRef.current?.focus({ preventScroll: true });
+
+      onClick?.(event);
+    };
+
+    const showIcon = () => {
```
edleeks87: suggestion: we're checking the same conditions a few times here and going through each if every time, it could be refactored a bit like so to be a bit more efficient

```
const getIconColor = () => {
  if (inverse) {
    if (variantType === "primary") {
      return "#000" 
    }
    if (disabled) {
      return "rgba(255, 255, 255, 0.42)";
    }
     return "#FFF";
  }

  if (variant === "ai") {
    return disabled
      ? "var(--button-ai-label-disabled, #0000006B)"
      : "var(--button-ai-label-active, #000)";
  }

  if (variantType === "primary") {
    return "var(--button-typical-primary-label-default, #FFF)";
  }

  if (variant === "destructive") {
    return disabled
      ? "var(--button-destructive-secondary-label-disabled, rgba(0, 0, 0, 0.42))"
      : "var(--button-destructive-secondary-label-default, #DB004E)";
  }

 return disabled
    ? "var(--button-typical-secondary-label-disabled, #0000006B)"
    : "var(--button-typical--secondary-label-active, #000)";
};

const showIcon = () => {
  if (size === "xs") return null;

  return <Icon type={iconType as IconType} color={getIconColor()} />;
};

File: src/components/button/__next__/button.component.tsx [hunk 10]
```diff
@@ -0,0 +1,262 @@
...
+
+      let iconColour =
+        variantType === "primary"
+          ? "var(--button-typical-primary-label-default, #FFF)"
+          : "var(--button-typical-secondary-label-default, #000000E6)";
+
+      if (variantType !== "primary") {
+        iconColour = disabled
+          ? "var(--button-typical-secondary-label-disabled, #0000006B)"
+          : "var(--button-typical--secondary-label-active, #000)";
```
edleeks87:
```suggestion
          : "var(--button-typical-secondary-label-active, #000)";
```

src/components/button/__next__/button.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+        <StyledLoadingContainer>
+          <LoaderSpinner
+            showSpinnerLabel={false}
+            size="extra-small"
+            variant={
+              variantType === "primary" && variant !== "ai"
+                ? "inverse"
+                : "action"
+            }
+          />{" "}
```
edleeks87: question (non-blocking): would it be better to have the spacing applied with margin on the loader?

src/components/button/__next__/button.mdx
```diff
@@ -0,0 +1,169 @@
...
+### HTML Button Type
+
+Standard HTML buttons can be one of three types: `button`, `submit` or
+`reset`. This functionality is available via the `type` prop, which
+allows `Button` instances to be used as standard buttons in elements
+such as forms.
+
+<Canvas of={ButtonStories.HTMLButtonType} />
+
+### As Links
```
edleeks87: Question: apologies I'm likely out of the loop here but we've confirmed this is still needed? Asking as we here constantly that it's bad practice etc

src/components/button/__next__/button.style.ts
edleeks87: It might be better to do my above suggestion around removing NonNullable usage in this dedicated file and then import where needed

src/components/button/__next__/button.test.tsx
```diff
@@ -0,0 +1,401 @@
...
+  "destructive-primary-large",
+  "destructive-secondary-small",
+  "destructive-secondary-medium",
+  "destructive-secondary-large",
+  "ai-primary-small",
+  "ai-primary-medium",
+  "ai-primary-large",
+];
+
+sizeAndVariantCombinations.forEach((combo) => {
```
edleeks87: suggestion (non-blocking): this is quite complex, I appreciate there's a lot of different variations that need testing but I think it might be better split out a bit more. If we add anything further to any of the combinations this becomes quite difficult to update vs splitting it out is a bit more maintainable

src/components/button/__next__/button.test.tsx
```diff
@@ -0,0 +1,401 @@
...
+  );
+
+  const button = screen.getByRole("link", { name: "Test Button" });
+  button.focus();
+  await user.keyboard("{ArrowRight}");
+
+  expect(clickMock).not.toHaveBeenCalled();
+});
+
+test("calling exposed focusTab method with correct tabId focuses chosen tab", async () => {
```
edleeks87: comment: this description doesn't really match the test

tsconfig-build.json
damienrobson-sage: TBH I'm struggling to work out what this change was for, or why. I'll roll it back and see if it breaks anything.

src/components/button/__next__/button.component.tsx
damienrobson-sage: The container is flex and has a gap, I think this is just my linter having fun with whitespace

src/components/button/__next__/button.mdx
damienrobson-sage: As discussed, awaiting feedback from UX on this

edleeks87: Discussed with @harpalsingh on slack and agreed we can omit this for now and add as a new feature if it proves necessary in the future

src/components/button/__next__/button-interactions.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+export default {
+  title: "Button/Interactions",
+  parameters: {
+    themeProvider: { chromatic: { theme: "sage" } },
+  },
+};
+
+type Story = StoryObj<typeof Button>;
+
+export const FocusAndHoverStates: Story = {
```
nuria1110: **suggestion (non-blocking):** I don't really see much value in this interaction test - there are no specific focus styles apart from the focus border.
I do think would be beneficial to capture the hover states & the active states but that could be done through the pseudo selectors to capture the different variants all in one go:

```
export const HoverAndActive: Story = {
  render: () => (
    <>
      <Box>
        <Button data-role="hover" variantType="primary">Primary</Button>
        {...etc}
      </Box>
      <Box>
        <Button data-role="active" variantType="primary">Primary</Button>
        {...etc}
      </Box>
    </>
  ),
};
HoverAndActive.parameters = {
  pseudo: {
    hover: "[data-role='hover']",
    active: "[data-role='active']",
  },
};
```

src/components/button/__next__/button.config.ts
```diff
@@ -0,0 +1,318 @@
...
+    paddingVertical: "6px",
+    paddingHorizontal: "16px",
+  },
+  large: {
+    borderRadius: "var(--global-radius-action-2-xl)",
+    font: "var(--global-font-static-comp-medium-l)",
+    height: "var(--global-size-l)",
+    iconOnlyWidth: "var(--global-size-l)",
+    paddingVertical: "6px",
+    paddingHorizontal: "16px",
```
nuria1110: **question:** Are there no design-tokens available for these? Same with the other paddings.

src/components/button/__next__/button.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+      </Box>
+      <Box display="flex" flexDirection="column" alignItems="flex-start">
+        <h1>Destructive</h1>
+        <h2>Primary</h2>
+        <>
+          <Button {...args} variant="destructive" variantType="primary" />
+        </>
+        <h2>Secondary</h2>
+        <>
+          <Button {...args} />
```
nuria1110: **nit-pick:** Missing the secondary & destructive props here

src/components/button/__next__/button.test.tsx
```diff
@@ -0,0 +1,894 @@
...
+});
+
+test("renders correctly when inverse", () => {
+  render(<Button inverse>Test Button</Button>);
+  const button = screen.getByRole("button");
+
+  expect(button).toBeInTheDocument();
+  expect(button).toHaveStyleRule(
+    "background-color: var(--button-typical-primary-inverse-bg-default)",
+  );
```
nuria1110, line 43: **nit-pick:** Formatting for `toHaveStyleRule`, there are a few like across some of the tests
```suggestion
  expect(button).toHaveStyleRule(
    "background-color", "var(--button-typical-primary-inverse-bg-default)",
  );
```

src/components/button/__next__/button.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+<Canvas of={ButtonStories.Default} />
+
+### Button Content
+
+Button content is composable; in other words, it's up to you to decide what you
+want the button to show, and how. In the following example, you can see how to
+implement [Icons](../?path=/story/icon--docs) into buttons. If you have multiple
+child elements, wrap them in a `Fragment` to ensure that the correct styling is
+maintained.
```
nuria1110, line 50: **question:** Not too sure why would we need a fragment if we are passing children?

src/components/button/__next__/button.test.tsx
```diff
@@ -0,0 +1,894 @@
...
+  const button = screen.getByRole("button");
+
+  expect(button).toBeInTheDocument();
+  expect(button).toHaveStyleRule(
+    "border: 1px solid var(--button-typical-tertiary-inverse-border-default)",
+  );
+  expect(button).toHaveStyle("border-radius: var(--global-radius-action-xl)");
+  expect(button).toHaveStyle("height: var(--global-size-m)");
+  expect(button).toHaveStyleRule("padding", "6px 16px");
```
nuria1110: **question:** I remember we previously had some issues with the old design-tokens giving false positives when using `toHaveStyle`, is this still the case for the new tokes?

src/components/button/__next__/button.mdx
edleeks87: If it's composed with a group of child elements react requires a wrapper, if they pass a span/div wrapper the flex gap styles are applied to that instead of the elements

Eg:
```
<>
  <Icon iconType="home" />
  Home
</>

File: src/components/button/__next__/button.test.tsx (hunk 20 above)
Comment (edleeks87) [reply src/components/button/__next__/button.test.tsx]: no idea tbh but I'll update to use styled-comps helper instead

File: src/components/button/__next__/button.config.ts (hunk 16 above)
Comment (edleeks87) [reply src/components/button/__next__/button.config.ts]: there are for some but not for 6px, I'll update what I can

File: src/components/button/__next__/button-interactions.stories.tsx (hunk 15 above)
Comment (edleeks87) [reply src/components/button/__next__/button-interactions.stories.tsx]: There's a focus box-shadow applied so I'm inclined to leave so it's captured. I'd move it to playwright but those tests are a bit in flux atm

Comment (edleeks87) [reply src/components/button/__next__/button-interactions.stories.tsx]: I've moved them into the test stories and deleted this file

File: src/components/button/__next__/button.component.tsx [hunk 21]
```diff
@@ -0,0 +1,285 @@
...
+            loaderType="ring"
+            size={size !== "large" ? "extra-small" : "small"}
+            inverse={useWhiteRing}
+            showLabel={false}
+            loaderLabel={locale.loaderSpinner.loading()}
+          />
+          {allowMotion && locale.loaderSpinner.loading()}
+        </>
+      );
+    };
```
Parsium: **question(non-blocking)**: do we want to offer a dedicated `loading` prop for `Button` as an alternative option to achieve the loading button pattern?

Currently, it should be possible to achieve this via composition:

```tsx
<Button variantType="primary">
  <Loader loaderType="ring" variant="inline" size="extra-small" showLabel />
</Button>
```

src/components/button/__next__/button.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+Similarly, variants can be one of up to four variant types: `primary`,
+`secondary`, `tertiary` and `subtle`. Supported variant-type combinations are
+as follows:
+
+| Variant       | Primary | Secondary | Tertiary | Subtle |
+| ------------- | ------- | --------- | -------- | ------ |
+| `default`     | ✅      | ✅        | ✅       | ✅      |
+| `destructive` | ✅      | ✅        | ❌       | ❌      |
+| `gradient`    | ❌      | ✅        | ❌       | ❌      |
```
Parsium, line 75: **praise**: the breakdown of all the supported visual states is really clear here 👍

src/components/button/__next__/button.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+      <Button size="xs">XS</Button>
+      <Button size="small">Small</Button>
+      <Button>Medium</Button>
+      <Button size="large">Large</Button>
+    </Box>
+  );
+};
+Sizes.storyName = "Sizes";
+
+export const Disabled: Story = () => {
```
Parsium, line 128: **suggestion(non-blocking)**: It might help to set the cursor to `not-allowed` on hover, so mouse users don't assume the element is interactive if its labelled as such.

src/components/button/__next__/button.mdx
```diff
@@ -0,0 +1,162 @@
...
+`reset`. This functionality is available via the `type` prop, which
+allows `Button` instances to be used as standard buttons in elements
+such as forms.
+
+<Canvas of={ButtonStories.HTMLButtonType} />
+
+### Legacy support
+
+To help with migration, this component can also accept the following legacy props:
+- the `buttonType` prop, which maps to the `variant` and `variantType` props as follows:
```
Parsium, line 148: **nitpick**: it might help to include a link to the old component docs to provide context:

```suggestion
### Support of legacy `Button` props 

If you are migrating from [Carbon's legacy `Button` component](?path=/docs/deprecated-button--docs), this component also accepts the following legacy props:
```

src/components/button/button.component.tsx
```diff
@@ -37,19 +43,25 @@ export interface ButtonProps extends SpaceProps, TagProps {
   "aria-labelledby"?: string;
   /** Identifies the element(s) offering additional information about the button the user might require. */
   "aria-describedby"?: string;
-  /** Color variants for new business themes: "primary" | "secondary" | "tertiary" | "darkBackground" */
+  /**
```
Parsium, line 46: **suggestion**: for the deprecated props, it could be useful to provide context behind the deprecation by including reason:


```ts
// in src/components/button/__next__/button.component.tsx:

import { type ButtonProps as LegacyButtonProps } from "../button.component";

interface ButtonProps extends Omit<
    LegacyButtonProps,
    "size" | "type" | "iconTooltipMessage" | "iconTooltipPosition" | "buttonType"
  >,
  SpaceProps,
  TagProps {
    /**
     * @deprecated Please use `variantType` prop instead.
    */
    buttonType: LegacyButtonProps["buttonType"];
     /**
     * @deprecated Please use `variant="destructive"` instead.
     * */
    destructive: LegacyButtonProps["destructive"];
  }
```

tsconfig.json
```diff
@@ -19,7 +23,11 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es6"
+    "target": "es6",
+    "types": [
+      "jest",
+      "@testing-library/jest-dom"
+    ]
```
Parsium: **issue**: after this change, our `npm run build-types` script fails with several errors. If this isn't strict required, could we revert?

src/components/button/__next__/button.style.ts
```diff
@@ -0,0 +1,297 @@
+import styled, { css } from "styled-components";
```
Parsium, line 1: **nitpick**: it might worth replacing the "AI" instances with "gradient" for consistency

src/components/button/__next__/button.test.tsx
```diff
@@ -0,0 +1,930 @@
...
+    const abbreviation = size === "xs" ? size : size.charAt(0);
+
+    expect(button).toHaveStyleRule(
+      "font",
+      `var(--global-font-static-comp-medium-${abbreviation})`,
+    );
+  },
+);
+
+describe("variant: default", () => {
```
Parsium: **suggestion**: what do you think about these render tests on lines 146-627 parameterised tests instead? 

This could help improve general readability:

```ts
describe.each<{ variant: Variant; variantType: VariantType; size: Size }>([
  { variant: "default", variantType: "primary", size: "small" },
  { variant: "default", variantType: "primary", size: "medium" },
  { variant: "default", variantType: "primary", size: "large" },
  { variant: "default", variantType: "secondary", size: "small" },
  { variant: "default", variantType: "secondary", size: "medium" },
  { variant: "default", variantType: "secondary", size: "large" },
  { variant: "default", variantType: "tertiary", size: "small" },
  { variant: "default", variantType: "tertiary", size: "medium" },
  { variant: "default", variantType: "tertiary", size: "large" },
  { variant: "default", variantType: "subtle", size: "small" },
  { variant: "default", variantType: "subtle", size: "medium" },
  { variant: "default", variantType: "subtle", size: "large" },
  { variant: "destructive", variantType: "primary", size: "small" },
  { variant: "destructive", variantType: "primary", size: "medium" },
  { variant: "destructive", variantType: "primary", size: "large" },
  { variant: "destructive", variantType: "secondary", size: "small" },
  { variant: "destructive", variantType: "secondary", size: "medium" },
  { variant: "destructive", variantType: "secondary", size: "large" },
  { variant: "gradient", variantType: "secondary", size: "small" },
  { variant: "gradient", variantType: "secondary", size: "medium" },
  { variant: "gradient", variantType: "secondary", size: "large" },
])(
  "when variant is $variant, variantType is $variantType and size is $size",
  ({ variant, variantType, size }) => {
    it("renders correctly", () => {
      // assertions...
    });

    it("renders correctly when disabled", () => {
      // assertions...
    });

    it("renders correctly when loading", () => {
      // assertions...
    });

    it("renders correctly with an icon", () => {
      // assertions...
    });

    it("renders correctly with disabled and with an icon", () => {
      // assertions...
    });
  },
);
```

src/components/button/__next__/button.component.tsx
edleeks87: Yep I wasn't sure if this was added for a reason that I was missing context, but on reflection I think the composition pattern is cleaner, I'll double check with @damienrobson-sage and remove if there was no steer to implement it this way

src/components/button/__next__/button.mdx
edleeks87: I will take all the credit for @damienrobson-sage's work

src/components/button/__next__/button.stories.tsx
edleeks87: I believe the steer from accessibility was to have the default browser cursor instead of `not-allowed`

tsconfig.json
edleeks87: I think it can be removed iirc when talking to Damo

src/components/button/__next__/button.test.tsx
edleeks87: I'll have a think about this, my personal pov is that complex parameterised tests with a lot of options can be less readable

Parsium (dismissed):

nuria1110 (dismissed):

src/components/button/__next__/button.config.ts
```diff
@@ -0,0 +1,318 @@
...
+    paddingVertical: "var(--global-space-comp-2-xs)",
+    paddingHorizontal: "var(--global-space-comp-m)",
+  },
+  medium: {
+    borderRadius: "var(--global-radius-action-xl)",
+    font: "var(--global-font-static-comp-medium-m)",
+    height: "var(--global-size-m)",
+    iconOnlyWidth: "var(--global-size-m)",
+    paddingVertical: "6px",
+    paddingHorizontal: "var(--global-space-layout-2-xs)",
```
ljemmo: @edleeks87 the spacing token consumed here says layout. These are reserved for layout stacks such as groups of tiles or buttons. Instead the token consumed should be --global-space-comp-L for 16px left and right padding.

src/components/button/__next__/button.config.ts
```diff
@@ -0,0 +1,318 @@
...
+  },
+  large: {
+    borderRadius: "var(--global-radius-action-2-xl)",
+    font: "var(--global-font-static-comp-medium-l)",
+    height: "var(--global-size-l)",
+    iconOnlyWidth: "var(--global-size-l)",
+    paddingVertical: "6px",
+    paddingHorizontal: "var(--global-space-layout-2-xs)",
```
ljemmo: @edleeks87 Large size should consume --global-space-comp-L also

ljemmo (dismissed):
Hey @edleeks87  - left a couple of comments inline where layout spacing tokens were used inline instead of comp spacing tokens.

Additionally, i found these other elements to consider:

1. When a button is composed with an icon, it appears we are consuming old tokens instead of new:
Image

2. Icon only button example is not a circle, it should have the same width as height. [Figma link here.](https://www.figma.com/design/Cie8BIDWKj4tEXCVGsen5A/Fusion-DS-Components?node-id=79171-391833&t=l4MSZX30DY2yZo1l-4)
Image

3. Inverse only has 1 example (primary button). For completeness it might be useful to show all the other inverse options in Storybook for devs to understand what they would look like:
Image

4. I would expect the `cursor: not-allowed;` treatment to be applied to disabled button but currently there's no cursor change.

DipperTheDan (approved):

nuria1110 (approved):

harpalsingh:
an review was requested after some updates, looks good to me 👍 , already has passed review so thats fine.

PR #7626, opened 2025-11-18
feat(text-editor): upgrade Lexical packages

damienrobson-sage:
Upgrades the Lexical packages to the latest compatible version to ensure we are using as up-to-date an ecosystem as possible

### Proposed behaviour

Upgrade to v35.0.0 of Lexical. Note that this is not the latest available version but taking that version would require changes to Babel and Jest's underlying configs, which risks impacting other components and processes. It should be revisited once the work to move away from Babel is completed

### Current behaviour

The Lexical package has not been updated since the initial implementation.

### Checklist

#### QA

### Testing instructions

Use the existing Storybook entries to ensure that no functionality is broken/different (all unit and PW tests are passing so it's unlikely to be the case)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "carbon-react",
-  "version": "158.11.0",
+  "version": "158.12.0",
   "description": "A library of reusable React components for easily building user interfaces.",
   "files": [
     "lib",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## [158.12.0](https://github.com/Sage/carbon/compare/v158.11.0...v158.12.0) (2025-11-26)
+
+### Features
+
+* **tabs:** add squared borders at active indicator edges ([ba18714](https://github.com/Sage/carbon/commit/ba18714b1d050205f2c36ec3bc0584a041b3d47b))
+
+### Bug Fixes
+
+* **tabs:** fix aria-label issue ([40e421e](https://github.com/Sage/carbon/commit/40e421ee3eacda1b9c015588f9005f53ecb5f4ad))
+
 ## [158.11.0](https://github.com/Sage/carbon/compare/v158.10.1...v158.11.0) (2025-11-25)
 
 ### Features
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "carbon-react",
-  "version": "158.11.0",
+  "version": "158.12.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "carbon-react",
-      "version": "158.11.0",
+      "version": "158.12.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@atlaskit/pragmatic-drag-and-drop": "^1.7.7",
--- a/src/components/tabs/__next__/tabs.component.tsx
+++ b/src/components/tabs/__next__/tabs.component.tsx
@@ -6,8 +6,6 @@ import React, {
   useRef,
   useState,
 } from "react";
-
-import Typography from "../../typography";
 import { TabsProvider, useTabs } from "./tabs.context";
 import {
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -208,7 +208,7 @@
     "date-fns": "^4.1.0",
     "immutable": "~3.8.2",
     "invariant": "^2.2.4",
-    "lexical": "^0.21.0",
+    "lexical": "^0.35.0",
```
Parsium: >Note that this is not the latest available version but taking that version would require changes to Babel and Jest's underlying configs

**suggestion**: would you be open to upgrading to the latest?

It seems the reason behind the failing jest tests is an updated transitive dependency now being an ESM only package. We could resolve this by updating `jest.config.ts` like so:

```ts
// jest.config.ts

const esmOnlyPackages = ["react-error-boundary"];

const baseProjectConfig: Config = {
   // other settings...

  transformIgnorePatterns: [
    `<rootDir>/node_modules/(?!(${esmOnlyPackages.join("|")}))`,
  ]
};
```

src/components/anchor-navigation/anchor-navigation.test.tsx
```diff
@@ -241,7 +241,8 @@ test("cleans up event listeners after unmounting", () => {
   unmount();
 
   expect(
-    addEventListenerSpy.mock.calls.filter((call) => call[0] === "scroll"),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Parsium: **suggestion**: could we type the spy here to avoid disabling the linter?

```ts
// line 239
const addEventListenerSpy: jest.SpyInstance = jest.spyOn(window, "removeEventListener");
```

```suggestion
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```

src/components/text-editor/__internal__/__ui__/ReadOnlyEditor/read-only-rte.component.tsx
```diff
@@ -80,6 +80,7 @@ const ReadOnlyEditor = ({
         <RichTextPlugin
           contentEditable={
             <ContentEditable
+              aria-label="note content"
```
Parsium: **suggestion**: would it worth introducing a prop for controlling this, so a consumer can set a different name in `Note` and `TextEditor`? This would help mitigate translation issues as well.

src/components/text-editor/__internal__/__ui__/ReadOnlyEditor/read-only-rte.component.tsx
```diff
@@ -80,6 +81,7 @@ const ReadOnlyEditor = ({
         <RichTextPlugin
           contentEditable={
             <ContentEditable
+              aria-label={ariaLabel || "note content"}
```
Parsium: **suggestion**: would it be worth removing the fallback value here to account for localisation?

As `TextEditor` doesn't render a `ReadOnlyEditor` with an `ariaLabel` prop, we wouldn't be able to control this accessible name if a different locale is used.

Parsium (dismissed):

package.json
```diff
@@ -196,10 +196,10 @@
     "@atlaskit/pragmatic-drag-and-drop-hitbox": "^1.1.0",
     "@floating-ui/dom": "~1.2.9",
     "@floating-ui/react-dom": "~1.3.0",
-    "@lexical/headless": "^0.21.0",
-    "@lexical/link": "^0.21.0",
-    "@lexical/react": "^0.21.0",
-    "@lexical/selection": "^0.21.0",
+    "@lexical/headless": "^0.38.2",
```
edleeks87: Comment: as lexical are not following semver for the pre alpha v1 releases, and can/will include breaking changes in any release before then, I think we should pin these versions. Even though `npm i` will not bump 0 versioned it will if `update` is run or someone does a ful install after deleting the lock file etc

PR #7641, opened 2025-11-24
Tab adjustments

damienrobson-sage:
### Proposed behaviour

- Update the focus border to have squared-off edges at the bottom when horizontal/right when vertical instead of rounded ones
- Remove the visually-hidden typography and `aria-labelledby`, as `aria-label` will take precedence regardless.

### Current behaviour

- The focus border that appears on `Tab` instances does not match the designs in the DS
- The use of `aria-labelledby` in the `TabList` component causes accessibility issues if more than one `Tabs` instance is used in the same page.

### Checklist

#### QA

### Testing instructions

Any existing Storybook story can be used to demonstrate the focus border adjustments. The `Focus and Hover` test story can be used to verify the accessibility fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [158.11.0](https://github.com/Sage/carbon/compare/v158.10.1...v158.11.0) (2025-11-25)
+
+### Features
+
+* **tabs:** re-introduce handler for monitoring tab changes ([50cb9c7](https://github.com/Sage/carbon/commit/50cb9c7fa61b4355cd31cc59fee558d49d5fa92e)), closes [#7622](https://github.com/Sage/carbon/issues/7622)
+
 ## [158.10.1](https://github.com/Sage/carbon/compare/v158.10.0...v158.10.1) (2025-11-24)
 
 ### Bug Fixes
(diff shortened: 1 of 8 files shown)


src/components/tabs/__next__/tabs.style.ts
```diff
@@ -244,7 +244,14 @@ export const StyledTab = styled.button<StyledTabProps>`
     ${addFocusStyling()}
     z-index: 1;
 
-    border-radius: 8px;
+    ${({ orientation }) => css`
+      border-top-left-radius: 8px;
```
edleeks87, line 248: question: should these be tokens? @harpalsingh also will we need fallbacks for spacing or is just colour you're concerned about in Tabs?

damienrobson-sage: @edleeks87 there are other non-token spacings throughout the file - would you like those updating too?

harpalsingh: For now I was just doing the color tokens since those were actually difference fallbacks, the sizing dont need fallbacks because the size should be the same as we have, but if its new tokens then use the old token or just hard coded value.

edleeks87: Ok based on that @damienrobson-sage let's leave these without tokens for now

paulrobinson88 (approved):

PR #7642, opened 2025-11-25
fix(submenu): ensure correct focus styling is applied to menu-item

DipperTheDan:
### Proposed behaviour

Square corner styling should be applied to the last menu-item in a segment if another segment succeeds it and there are no other menu-items outside of the segment blocks.

Screenshot 2025-11-25 at 14 32 01

Screenshot 2025-11-25 at 14 32 11

### Current behaviour

Rounded corner styling is being applied to the last menu-item in a segment if another segment succeeds it and there are no other menu-items outside of the segment blocks.

Screenshot 2025-11-25 at 14 33 18

Screenshot 2025-11-25 at 14 33 25

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

- No other visual regression with focusing styling with menu-items, segment blocks, scrollable blocks or submenus
- These can be tested using the story `MenuWithTwoSegments` as this has various examples in all the menu-items that should cover most cases.

### DIFF
--- a/src/components/menu/__internal__/submenu/submenu.component.tsx
+++ b/src/components/menu/__internal__/submenu/submenu.component.tsx
@@ -180,40 +180,63 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
 
     useEffect(() => {
       const handleBorderRadiusStyling = () => {
+        /* istanbul ignore next */
+        if (!submenuRef) return;
+
+        const allMenuItems = Array.from(
+          submenuRef.querySelectorAll("[data-component='menu-item']") ||
+            /* istanbul ignore next */ [],
+        );
+
+        const lastMenuItem =
+          allMenuItems.length > 0
+            ? allMenuItems[allMenuItems.length - 1]
+            : null;
+
+        lastMenuItemElement.current = lastMenuItem ?? undefined;
+
+        // Clear any previous markers in this submenu in case menu items have changed since last opening
+        submenuRef
+          .querySelectorAll("[data-last-visible-menu-item]")
+          .forEach((el) => el.removeAttribute("data-last-visible-menu-item"));
+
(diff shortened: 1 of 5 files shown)


src/components/menu/__internal__/submenu/submenu.component.tsx
```diff
@@ -180,40 +180,63 @@ const Submenu = React.forwardRef<HTMLAnchorElement, SubmenuProps>(
 
     useEffect(() => {
       const handleBorderRadiusStyling = () => {
-        // Finds all ul elements that are not submenus
-        const ulElements = Array.from(
-          submenuRef?.querySelectorAll("ul:not([data-component='submenu'])") ||
+        /* istanbul ignore next */
```
Parsium, line 183: **comment**: I realise we need to have a thorough review of `Menu`'s style files, given it has grown complexity, so if a CSS-only approach would be too difficult, I'm happy to take this effect approach for now. 

Having said that, I do think we need to simplify `Menu`'s styling files sooner rather than later, as it will become harder to maintain if we need to introduce new changes.

DipperTheDan: A CSS-only approach was tried originally to solve this, and it wasn't possible to catch every scenario where rounded corners was and wasn't needed. The JS approach was the least worst reliable way to get it done, which is mental really 🤣  

I 100% agree with regards to Menu's styling stuff too. It was a beast before we had to support rounded corners and now having to factor in all these permutations of Menu with rounded corners, it's getting insane what we're having to do.

tomdavies73: I agree with @Parsium's sentiment that a pure CSS solution is probably the best option here but agree with you both that this is probably the best solution in the interim. We can review menu styling as a whole and hopefully simplify our wider approach in the future 👍

PR #7645, opened 2025-11-27
feat(typography): update component with fusion DS

tomdavies73:
### Proposed behaviour

1. Adds a new internal next version of the `Typography` component, this will eventually become the "main" version in the future and has been build to be fully agnostic
2. Adapts the legacy version of the `Typography` component to support legacy props, functionality and styling whilst also supporting next props, functionality and styling

Deprecation warnings have been added for the usage of legacy  `List` and `ListItem`, individual deprecation warnings have been added to each prop description, pointing towards a suitable alternative where possible. 

Internally the legacy component now renders the next `Typography` but with a styled component approach, therefore highly used existing styling overrides can still be applied by consumers (`lineHeight`, `fontWeight`, `fontSize`), and these override the rendered next `Typography.`

This prevents consumers from encountering any `"Breaking changes"` e.g. actual code changes which will break their project, but also prevents major visual changes which are arguably as breaking for a component such as `Typography` which is heavily used by our consumers.

### Current behaviour

N/A

### Checklist

#### QA


src/components/typography/__next__/config.ts
edleeks87, line 1: Question: Have you confirmed this won't be included in the build?

src/components/typography/__internal__/__next__/typography.component.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return "h2";
+    }
+    if (variant === "section-subheading") {
+      return "h3";
+    }
+    return variant;
+  }
+
+  return (
+    <StyledTypography
```
edleeks87, line 119: suggestion: this props will be transitive we should add `$` to any not intended for the underlying DOM element (ie variant, fluid, inverse, size, tint, weight)

edleeks87 (commented):
pausing this review as we need to explore other options as well

mihai-albu-sage (dismissed):

src/components/typography/typography.component.tsx
```diff
@@ -101,84 +142,108 @@ export interface TypographyProps extends SpaceProps, TagProps {
...
+  b: "b",
+  ul: "ul",
+  ol: "ol",
+};
+
+const getAs = (variant?: VariantTypes): React.ElementType =>
+  (variant && VARIANT_TO_ELEMENT[variant]) ?? "p";
+
+/**
+ * @deprecated This component is deprecated and will be removed in a future release.
```
edleeks87: comment: we shouldn't be marking the whole component as deprecated

src/components/typography/typography.style.ts
```diff
@@ -1,183 +1,20 @@
...
-    const size = fontSize || getSize(variant);
-    const weight = fontWeight || getWeight(variant);
-    const textTransform = transform || getTransform(variant);
-    const textDecoration = decoration || getDecoration(variant);
-    const lineHeight = lHeight || getLineHeight(variant);
-    const defaultMargin = variant === "p" ? "0 0 16px" : "0";
+import { TypographyProps } from "./typography.component";
+import Typography from "./__internal__/__next__";
 
+const StyledTypography = styled(Typography).attrs(
```
edleeks87, line 8:
```suggestion
const StyledTypography = styled(Typography).attrs(({className, rest}) => { return applyBaseTheme(rest); }
```

edleeks87: classNames are duplicated as the theme function re-applies all the props which styled-component concatenates

src/components/typography/__internal__/__next__/typography.style.ts
```diff
@@ -0,0 +1,106 @@
...
+  h1: "heading-l",
+  h2: "heading-m",
+  h3: "heading-s",
+  h4: "subheading-l",
+  h5: "subheading-m",
+};
+
+const StyledTypography = styled.span.attrs(applyBaseTheme)<TypographyProps>`
+  ${({
+    display,
```
edleeks87: comment: anything that isn't a valid DOM attribute should have the `$` prefix to prevent the props being transitive

src/components/typography/__internal__/__next__/typography.style.ts
```diff
@@ -0,0 +1,106 @@
+import styled, { css } from "styled-components";
+import { space } from "styled-system";
+import applyBaseTheme from "../../../../style/themes/apply-base-theme";
+import { TypographyProps } from "./typography.component";
+import visuallyHidden from "../../../../style/utils/visually-hidden";
+
+const headingVariants = {
```
edleeks87: comment (non-blocking): if you type this you shouldn't need to coerce the type to index

src/components/typography/__internal__/__next__/typography.style.ts
```diff
@@ -0,0 +1,106 @@
...
+    const isSuperOrSub = variant === "sup" || variant === "sub";
+    const headingFontVar = `--global-font-${fontType}-${isHeading ? headingVariants[variant as keyof typeof headingVariants] : ""}`;
+    const bodyMediumFontVar = `--global-font-${fontType}-body-medium-${sizeValue}`;
+    const bodyFontVar = `--global-font-${fontType}-body-${weight}-${sizeValue}`;
+
+    return css`
+      /* Margin is unset as browsers can add default margin based on the underlying HTML element,
+      this can cause a disconnect between the styling added via underlying element (as prop) opposed to the 
+      styling added via the variant prop. Consumers can pass their own margin instead if needed. */
+      ${!screenReaderOnly && "margin: unset;"}
```
edleeks87: praise: good catch

src/components/typography/__internal__/__next__/typography.style.ts
```diff
@@ -0,0 +1,106 @@
...
+      ${isStrongOrBold &&
+      css`
+        font: var(${bodyMediumFontVar});
+        color: ${baseColorToken};
+      `}
+
+      ${variant === "sup" &&
+      css`
+        font: var(${bodyFontVar});
+        font-size: 0.75em !important;
```
edleeks87: suggestion: I'd prefer to explore alternatives to `!important` such as bumping specificity with `&& {}`

src/components/typography/typography.stories.tsx
```diff
@@ -15,180 +52,479 @@ const meta: Meta<typeof Typography> = {
...
 };
 
 export default meta;
 type Story = StoryObj<typeof Typography>;
 
 export const VariantsStory: Story = () => (
-  <>
-    <Typography variant="h1-large">Heading Level 1 Large</Typography>
+  <Box display="flex" flexDirection="column" gap={2}>
+    <Typography variant="p">P (Default)</Typography>
```
edleeks87:
```suggestion
    <Typography variant="p">Paragraph (Default)</Typography>
```

src/components/typography/__internal__/__next__/typography.test.tsx
```diff
@@ -0,0 +1,797 @@
...
+
+describe("h1", () => {
+  it("should render as h1 with correct font/color tokens", () => {
+    render(<Typography variant="h1">Heading text</Typography>);
+
+    const heading = screen.getByRole("heading", { level: 1 });
+
+    expect(heading).toHaveStyle("font: var(--global-font-static-heading-l)");
+    expect(heading).toHaveStyle(
+      "color: var(--container-standard-standard-txt-default)",
```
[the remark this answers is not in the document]
tomdavies73: This is correct, I changed the token round but forgot to update the respective test.

src/components/typography/__internal__/__next__/typography.style.ts
tomdavies73: Agreed I'll use a `$` prefix and create a local interface in the style file

src/components/typography/__internal__/__next__/typography.style.ts
tomdavies73: Doesn't actually appear to be much difference, looks safe to remove 👍

src/components/typography/typography.component.tsx
```diff
@@ -101,84 +142,108 @@ export interface TypographyProps extends SpaceProps, TagProps {
...
+  "section-heading": "h2",
+  "section-subheading": "h3",
+  p: "p",
+  span: "p",
+  small: "p",
+  big: "p",
+  sup: "sup",
+  sub: "sub",
+  strong: "strong",
+  em: "strong",
```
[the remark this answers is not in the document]
tomdavies73: I think this is a fair point, I think we should preserve some of the casts to match the passed in variants. They may look different, but under the hood they will still be the same element which means the browser will treat them the same as previously.

edleeks87 (dismissed):

mihai-albu-sage (dismissed):

ljemmo (dismissed):

mihai-albu-sage (dismissed):

skills/carbon-react/components/box.md
```diff
@@ -148,7 +148,7 @@ description: Carbon Box component props and usage examples.
           position="sticky"
           top="0"
         >
-          <Typography color="white">This box has position sticky</Typography>
+          <Typography inverse>This box has position sticky</Typography>
```
edleeks87: comment: I wonder if it's worth mapping `color="white"` to `inverse` to prevent text being completely lost against a black background when this is merged

src/components/typography/typography.style.ts
```diff
@@ -1,18 +1,24 @@
...
 import { TypographyProps } from "./typography.component";
 import Typography from "./__internal__/__next__";
+import visuallyHidden from "../../style/utils/visually-hidden";
 
 const StyledTypography = styled(Typography).attrs(({ className, ...rest }) => {
   return applyBaseTheme(rest);
 })<TypographyProps>`
-  ${({ fontSize, fontWeight, lineHeight }) => {
+  ${({ fontSize, fontWeight, lineHeight, variant, screenReaderOnly }) => {
+    const defaultMargin = variant === "p" ? "0 0 16px" : "0";
```
edleeks87: comment: we should use the relevant tokens here

edleeks87 (commented):
Looking over the chromatic, are the [fluid](https://www.chromatic.com/test?appId=5ecf782fe724630022d27d7d&id=69dd192477506126b02489cd) snapshots adding anything? I also notice we're capturing both `sage` and `none` themes when I don't think we need to. It might also be possible to do some consolidation of some test stories to capture what we want in chromatic with less snapshots (non-blocking).

src/components/typography/typography.style.ts
tomdavies73: I'm not too sure we need the new tokens here? I'm not sure we'd get much value from using a token, especially with it being a temporary override which will eventually be removed

src/components/typography/typography.style.ts
```diff
@@ -1,183 +1,26 @@
...
-    textTransform: transform,
-    truncate,
-    screenReaderOnly,
-  }) => {
-    const size = fontSize || getSize(variant);
-    const weight = fontWeight || getWeight(variant);
-    const textTransform = transform || getTransform(variant);
-    const textDecoration = decoration || getDecoration(variant);
-    const lineHeight = lHeight || getLineHeight(variant);
+const StyledTypography = styled(Typography).attrs(({ className, ...rest }) => {
```
edleeks87: suggestion (non-blocking): it's probably safe to shorthand this again, I have a fix to the base theme in my other PR so this will be redundant soon anyway. It'll duplicate custom class names in the interim but it's trivial imo.

```
const StyledTypography = styled(Typography).attrs(applyBaseTheme)<TypographyProps>`

src/components/typography/typography.style.ts
tomdavies73: After a bit more thought using a token won't hurt tbf, i'll just use a preset `global-space...` token

edleeks87 (approved):

mihai-albu-sage (approved):

paulrobinson88 (approved):

PR #7654, opened 2025-12-03
chore: update commit script to work cross platform

tomdavies73:
### Proposed behaviour

Ensures that the commit script can be used cross browser with a temporary file approach for commit message

### Current behaviour

Shell escaping has been used heavily for the creation of the commit message, this works fine on some operating systems but is handled differently on Windows OS. This means some users on Windows machines cannot use the script properly to the desired effect.

### Checklist

#### QA


scripts/commit/commit.mjs
```diff
@@ -197,20 +201,33 @@ async function commit() {
...
   } catch (error) {
     console.error('\n❌ Commit failed:', error.message);
     process.exit(1);
+  } finally {
+    // Clean up temp file at the end
+    if (tempFile && fs.existsSync(tempFile)) {
+      try {
+        fs.unlinkSync(tempFile);
+      } catch (e) {
+        // Ignore cleanup errors
```
nineteen88: suggestion: avoid having empty catches if possible

It's probably worth just adding a console warning that there was an error. It's highly unlikely it could ever hit here, but we should log something out at the very least anyway, rather than it just swallowing up any errors.

```typescript
console.warn('Failed to clean up temp file:', e);
```

DipperTheDan (approved):

PR #7669, opened 2025-12-09
fix(adaptive-sidebar): modal aria is correct when hidden

tomdavies73:
### Proposed behaviour

Ensures the `useModalAria` hook recognises the hidden functionality in `AdaptiveSidebar` as a valid reason to run the clean-up function. This ensures all relevant attributes previously added are now removed. This aligns hiding the component with the component being closed/unmounted, as there is no perceivable differences between all of these cases for an end user.

### Current behaviour

Currently, there is an issue where modal aria attributes persist when the `AdaptiveSidebar` is hidden, meaning when it is unhidden the component cannot be interacted with correctly. 

### Checklist

#### QA

### Testing instructions

1. Navigate to the `Hidden` story in `AdaptiveSidbar`
2. Expand the browser window to as large a size as possible
3. Open the `AdaptiveSidebar`
4. Hide the `AdaptiveSidebar` via the `SplitButton`
5. Shrink the browser window to as small a size as possible
6. Try and open the `AdaptiveSidebar` in this smaller viewport

Repeating these steps on [master](https://carbon.sage.com/) vs a local reproduction of this PR will show the broken behaviour (master) and the fixed behaviour (local reproduction).

### DIFF
--- a/src/components/sidebar/sidebar.component.tsx
+++ b/src/components/sidebar/sidebar.component.tsx
@@ -96,6 +96,8 @@ export interface SidebarProps
    * @internal
    * Sets className for component. INTERNAL USE ONLY. */
   className?: string;
+  /** @private @ignore Whether the `Sidebar` is hidden from view when rendered in an `AdaptiveSidebar`. */
+  hidden?: boolean;
 }
 
 export const Sidebar = React.forwardRef<HTMLDivElement, SidebarProps>(
@@ -129,6 +131,7 @@ export const Sidebar = React.forwardRef<HTMLDivElement, SidebarProps>(
       topModalOverride,
       restoreFocusOnClose = true,
       className,
+      hidden,
       ...rest
     }: SidebarProps,
     ref,
@@ -149,7 +152,7 @@ export const Sidebar = React.forwardRef<HTMLDivElement, SidebarProps>(
       [ref],
     );
 
-    const isTopModal = useModalAria(sidebarRef);
+    const isTopModal = useModalAria(sidebarRef, hidden);
 
     const closeIcon = () => {
       if (!onCancel) return null;
--- a/src/hooks/__internal__/useModalAria/useModalAria.test.tsx
+++ b/src/hooks/__internal__/useModalAria/useModalAria.test.tsx
@@ -33,9 +33,10 @@ const ModalComponent = ({
   closeButtonText,
   dialogLabel = "dialog",
   children,
-}: ModalComponentProps) => {
+  hidden = false,
+}: ModalComponentProps & { hidden?: boolean }) => {
   const modalRef = useRef(null);
-  const isTopModal = useModalAria(modalRef);
+  const isTopModal = useModalAria(modalRef, hidden);
   const [isOpen, setOpen] = useState(false);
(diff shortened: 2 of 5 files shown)


src/hooks/__internal__/useModalAria/useModalAria.test.tsx
```diff
@@ -332,6 +357,39 @@ it("restores any previously-overridden aria-hidden and inert properties when mod
   );
 });
 
+it("restores any previously-overridden aria-hidden and inert properties when modal is hidden", async () => {
+  const user = userEvent.setup();
+  render(
+    <>
+      <div data-role="old-aria-hidden" aria-hidden="false" />
+      <HiddenModal />
+      {/* @ts-expect-error inert property not recognised by React. Support to be added in React 19 https://github.com/facebook/react/pull/24730 */}
```
edleeks87: suggestion: you can add the following to `global.d.ts` to resolve the type error

```
declare module "react" {
  interface HTMLAttributes {
    inert?: "" | "true" | string;
  }
}

File: src/components/sidebar/sidebar.component.tsx [hunk 2]
```diff
@@ -96,6 +96,8 @@ export interface SidebarProps
    * @internal
    * Sets className for component. INTERNAL USE ONLY. */
   className?: string;
+  /** @private @ignore */
```
edleeks87: suggestion (non-blocking): whilst this is a private prop it's probably worth adding a description above it for sake of maintenance etc

src/hooks/__internal__/useModalAria/useModalAria.tsx
```diff
@@ -75,7 +76,10 @@ export default function useModalAria(
           element.removeAttribute("data-modal-hidden");
         },
       );
-  }, [topModal, isTopModal]);
+    /* hidden is included as a dependency so that when it becomes true, cleanup runs
```
edleeks87: comment (non-blocking): this comment is a little redundant imo as the prop is used in the effect in needs to be included in the dependency array already etc

src/hooks/__internal__/useModalAria/useModalAria.tsx
```diff
@@ -55,7 +56,7 @@ export default function useModalAria(
       }
     };
 
-    if (isTopModal) {
+    if (isTopModal && !hidden) {
```
edleeks87: suggestion (non-blocking): I wonder if it might be better to check `hidden` prop first and short-circuit the rest of this effect if it's true etc

```
  if (hidden || rootElement.dataset.notInert === "true") {
        // stop recursing, and do nothing, if the container is hidden or has the "data-not-inert" flag
        return;
      }

src/hooks/__internal__/useModalAria/useModalAria.tsx
tomdavies73: Yeah, I just thought it was worth adding some context as to why it's in the deps array, as it seems a little out of place at first glance. But the other changes to the file will give enough context I think now, so I'll remove this comment

nuria1110 (approved):

paulrobinson88 (approved):

PR #7676, opened 2025-12-12
test(filterable-select, multi-select): remove flaky playwright tests

Parsium:
### Proposed behaviour

Remove selection-related Playwright tests, ensure these scenarios are covered by Jest instead.

### Current behaviour

Several selection-related Playwright tests for `FilterableSelect` and `MultiSelect` are known to commonly to flake when ran.

### Checklist

#### QA


src/components/select/filterable-select/filterable-select.test.tsx
```diff
@@ -1467,136 +1467,170 @@ describe("when the user types in the input", () => {
   });
 });
 
-describe("when the user selects an option", () => {
```
nineteen88, line 1470: suggestion(non-blocking): keep the describe block

We now have it test statements outside of a describe block as a result of removing this. They should really be contained by a describe block or changed to regular `test()` instead but it seems there's a fairly clear block of tests that should be grouped together here so a describe should probably remain. The describe block itself could be changed to be a better capture statement of the tests it's describing though. 

`describe("when a user interacts with an Option", () => {`

nineteen88 (dismissed):

DipperTheDan (dismissed):

DipperTheDan (approved):

PR #7688, opened 2025-12-23
fix: resolve strict mode incompatibilities with portals and popups (FE-7197)

Parsium:
### Proposed behaviour

For Carbon components involving React Portals, resolve inconsistencies in rendering behaviour:

- `Portal` - when unmounting, properly clean up DOM node created as the container for the portal content
- `Popover` - always render portal content in document body or nearest Carbon modal, to avoid the need to create and cleanup a DOM node.
- `CarbonProvider` - ensure global used in updating topmost modal is cleaned up properly.

### Current behaviour

Since #7148 was released, several Carbon components that use React Portals show unexpected behaviour when React Strict Mode is enabled, such as modals and popups not appearing at all. This occurs because these components have rendering inconsistencies that are surfaced by the [stricter Strict Mode checks introduced in React v18](https://react.dev/blog/2022/03/29/react-v18#new-strict-mode-behaviors). 

Before upgrading Carbon to support newer React versions, we should ensure all components behave correctly under Strict Mode to prepare for future React changes. 

### Checklist

#### QA

### Testing instructions

1. Run Storybook with React Strict Mode enabled: `npm run start:strict-mode`
2. Check the following components and compare behaviour with their production docs counterparts:
- `AdvancedColorPicker`
- `Alert`
- `Confirm`
- `Dialog`
- `Sidebar`
- `MenuFullscreen`
- `ResponsiveVerticalMenu`
- `Toast`
- `Tooltip`
- `VerticalMenuFullscreen`

### DIFF
--- a/src/hooks/__internal__/useIsBrowser/index.ts
+++ b/src/hooks/__internal__/useIsBrowser/index.ts
@@ -1,0 +1,20 @@
+import { useEffect, useState } from "react";
+import { getWindow, getDocument } from "../../../__internal__/dom/globals";
+
+/**
+ * Hook that detects whether your React component is being rendered in a client-side or server-side environment.
+ *
+ * Useful for handling logic that should *only* be executed in the browser, such as DOM manipulations or accessing browser APIs.
+ */
+function useIsBrowser() {
+  const canUseDOM = Boolean(getWindow() && getDocument());
+  const [isBrowser, setIsBrowser] = useState(canUseDOM);
+
+  useEffect(() => {
+    setIsBrowser(true);
+  }, []);
+
+  return { isBrowser };
+}
+
+export default useIsBrowser;
--- a/src/__internal__/popover/popover.component.tsx
+++ b/src/__internal__/popover/popover.component.tsx
@@ -1,17 +1,12 @@
-import React, {
-  MutableRefObject,
-  useContext,
-  useEffect,
-  useRef,
-  RefObject,
-} from "react";
-import ReactDOM from "react-dom";
+import React, { MutableRefObject, useContext, useRef, RefObject } from "react";
+import { createPortal } from "react-dom";
 import { flip, Placement, Middleware } from "@floating-ui/dom";
 
 import useFloating from "../../hooks/__internal__/useFloating";
 import { StyledBackdrop, StyledPopoverContent } from "./popover.style";
 import CarbonScopedTokensProvider from "../../style/design-tokens/carbon-scoped-tokens-provider/carbon-scoped-tokens-provider.component";
 import ModalContext, { ModalContextProps } from "../modal/modal.context";
+import useIsBrowser from "../../hooks/__internal__/useIsBrowser";
 
 export interface PopoverProps {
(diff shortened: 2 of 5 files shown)


src/components/carbon-provider/__internal__/top-modal-provider.component.tsx
```diff
@@ -1,24 +1,16 @@
-import React, { useState, useEffect, useRef, ReactNode } from "react";
+import React, { useState, useEffect, ReactNode } from "react";
 import TopModalContext from "./top-modal.context";
 
 export const TopModalProvider = ({ children }: { children: ReactNode }) => {
   const [topModal, setTopModal] = useState<HTMLElement | null>(null);
 
-  // can't add the setter to the global list inside useEffect because that doesn't run until
```
DipperTheDan, line 7: **question(non-blocking):** Are we all good to remove this ref and condition for first render? The comment states that we can't add the setter to the global list inside a `useEffect` and we're now doing that.

Parsium: This should be okay now, since `Modal` no longer attempts to access this global on initial render.

`Portal` now creates its attachment node on a separate render to when the portal content is displayed. Because of this change, `Modal` has been re-structured (see below) so its internal hooks, like `useModalManager`, are only called once the modal content is rendered.

```tsx
// src/__internal__/modal/modal.component.tsx#L179

const Modal = (props: ModalProps) => (
  <Portal>
   {/* 👇 `Portal` only renders this once its attachment node is present */}
    <ModalRoot {...props} />
  </Portal>
);
```

src/components/portal/__internal__/safe-portal.tsx
```diff
@@ -0,0 +1,81 @@
...
+const SafePortal = ({
+  children,
+  id,
+  className,
+  attributes,
+}: SafePortalProps) => {
+  const [portalNode, setPortalNode] = useState<HTMLElement | null>(null);
+  const [didCreateNode, setDidCreateNode] = useState(false);
+
+  useEffect(() => {
```
edleeks87, line 26: suggestion (non-blocking):

```
useEffect(() => {
  // Find or create portal node
  let node = document.getElementById(id);
  const shouldCreateNode = !node;

  if (shouldCreateNode) {
    node = document.createElement("div");
    node.setAttribute("id", id);
    document.body.appendChild(node);
  }

  setDidCreateNode(shouldCreateNode);
  setPortalNode(node);

  // Cleanup portal node if component created it
  return () => {
    if (shouldCreateNode) {
      node?.remove();
    }
  };
}, [id]);
```

src/hooks/__internal__/useIsBrowser/index.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+function canUseDOM() {
```
edleeks87: suggestion (non-blocking): we could use the `getDocument` and `getWindow` utils here instead

PR #7694, opened 2026-01-08
feat(locale): Add portuguese locale

nicktitchmarsh:
### Proposed behaviour

Fixes FE-7547

Adding locale file for Portuguese translations

### Checklist

#### QA

### DIFF
--- a/.storybook/manager.ts
+++ b/.storybook/manager.ts
@@ -9,18 +9,14 @@ import { InteractionToggle } from "./interaction-toggle/reduced-motion";
 import VersionPicker from "./version-picker";
 import { API_PreparedIndexEntry, API_StatusObject } from "@storybook/types";
 
-const useVersionPicker = process.env.USE_VERSION_PICKER === "true";
-
-if (useVersionPicker) {
-  addons.register(ADDON_ID, () => {
-    addons.add(TOOL_ID, {
-      type: types.TOOL,
-      title: "Version picker",
-      match: ({ viewMode }) => !!(viewMode && viewMode.match(/^(story|docs)$/)),
-      render: VersionPicker,
-    });
+addons.register(ADDON_ID, () => {
+  addons.add(TOOL_ID, {
+    type: types.TOOL,
+    title: "Version picker",
+    match: ({ viewMode }) => !!(viewMode && viewMode.match(/^(story|docs)$/)),
+    render: VersionPicker,
   });
-}
+});
 
 addons.register(INTERACTION_TOGGLE_ADDON_ID, () => {
   addons.add(INTERACTION_TOGGLE_TOOL_ID, {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [158.18.0](https://github.com/Sage/carbon/compare/v158.17.0...v158.18.0) (2026-01-15)
+
+### Features
+
+* **version-picker:** remove env variable check ([1954632](https://github.com/Sage/carbon/commit/1954632b9be9088de4fa33e3d461eb803e24ffc2))
+
 ## [158.17.0](https://github.com/Sage/carbon/compare/v158.16.1...v158.17.0) (2026-01-15)
 
 ### Features
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "carbon-react",
-  "version": "158.17.0",
+  "version": "158.18.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "carbon-react",
-      "version": "158.17.0",
+      "version": "158.18.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@atlaskit/pragmatic-drag-and-drop": "^1.7.7",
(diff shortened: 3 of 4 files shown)


src/locales/pt-pt.ts
```diff
@@ -0,0 +1,288 @@
...
+    ai: () => "Informação gerada por IA",
+    callout: () => "Alerta",
+  },
+  numeralDate: {
+    validation: {
+      day: (month, daysInMonth) => {
+        if (month && daysInMonth) {
+          return `O dia de ${month} eve ser um número entre 1 e ${daysInMonth}.`;
+        }
+        return "Day should be a number within a 1-31 range.";
```
tomdavies73: **issue**: appears to be a hardcoded English value here

.storybook/preview.ts
```diff
@@ -61,6 +61,7 @@ const globalTypes = {
         { value: "es-ES", right: "🇪🇸", title: "Español" },
         { value: "fr-FR", right: "🇫🇷", title: "Français" },
         { value: "fr-CA", right: "🇨🇦", title: "Français - Canada" },
+        { value: "pt-PT", right: "🇵🇹", title: "Português - Portugal" },
```
edleeks87, line 64: nitpick: we don't need to hyphenate here, it's only done for the one above to distinguish that one of the French locales is Canadian.

src/locales/pt-pt.ts
```diff
@@ -0,0 +1,288 @@
...
+    warning: () => "Aviso",
+    neutral: () => "Informação",
+    ai: () => "Informação gerada por IA",
+    callout: () => "Alerta",
+  },
+  numeralDate: {
+    validation: {
+      day: (month, daysInMonth) => {
+        if (month && daysInMonth) {
+          return `O dia de ${month} eve ser um número entre 1 e ${daysInMonth}.`;
```
tomdavies73: **issue**: `O dia de ${month} eve` should be `O dia de ${month} deve`. A `d` appears to have been missed

src/locales/pt-pt.ts
```diff
@@ -0,0 +1,288 @@
...
+      listAriaLabel: () => "Lista de pessoas que podem ser mencionadas",
+    },
+  },
+  search: {
+    searchButtonText: () => "Pesquisar",
+  },
+  select: {
+    actionButtonText: () => "Adicionar novo item",
+    placeholder: () => "Selecione...",
+    noResultsForTerm: (term) => `Não há resultados para "${term}"`,
```
tomdavies73: **issue** (maybe?? / non-blocking): the `"` has been left around the concatenation here. It can make logical sense as we're returning a term, but this has not been done in other locale files so it may be worth removing for consistency sake

src/locales/pt-pt.ts
```diff
@@ -0,0 +1,288 @@
...
+    undo: () => "Anular",
+  },
+  textEditor: {
+    boldAria: () => "Negrito",
+    cancelButton: () => "Cancelar",
+    cancelButtonAria: () => "Cancelar",
+    characterCounter: (count: number | string) => {
+      const formattedCount =
+        typeof count === "number" ? count.toString() : count;
+      return count == "1"
```
tomdavies73: **suggestion** (non-blocking): I don't personally mind the use of loose equality here, however we tend to use strict equality in the rest of the codebase. 

A strict equality check against `formattedCount` may be better, but keeping it as it is won't stop be from approving 👍

src/locales/pt-pt.ts
```diff
@@ -0,0 +1,288 @@
+import { pt as ptPTDateLocale } from "date-fns/locale/pt";
+
+import Locale from "./locale";
+
+const isSingular = (count: string | number): boolean =>
+  (typeof count === "string" ? parseInt(count) : count) === 1;
+
+const ptPT: Locale = {
```
tomdavies73: **suggestion**: we typically type non `en-GB` locales as `Partial<Locale>`, this means we can add locale types in the future without `pt-PT` support whilst we await translations and avoid a type error

src/locales/pt-pt.ts
edleeks87:
```suggestion
      return isSingular(count)
```

edleeks87: You don't really need to have a formatted count as it's being interpolated into the string anyway etc

```
return isSingular(count)
? `Resta ${count} caráter`
        : `Restam ${count} carateres`;

src/locales/pt-pt.ts
nicktitchmarsh: Good spot

.storybook/preview.ts
nicktitchmarsh: You can have pt-BR for Brazilian Portuguese

tomdavies73 (approved):

.storybook/preview.ts
chrisbarber86: So either `Français` needs to be `Français - French` to match that then.

or we only do it when it's not the originating country (e.g. fr-FR is `Français` but fr-CA is `Français - Canada`) - so `pt-BR` would be `Português - Brazil` but `pt-PT` would just be `Português`

PR #7695, opened 2026-01-08
feat(accordion): add support to search content within collapsed panels

### LINKED ISSUE #7672: Accordion: allow finding of content in collapsed panels
### Description

It is currently not possible for users to search for content via browser find-in-page functionality within collapsed accordion content. This is problematic for keyboard-only and screen reader users as the only way to find content currently is to expand all accordions on-screen first and then perform a find.

### Suggested solution

Apply the `hidden="until-found"` attribute to the content container to allow searching within collapsed accordions. Note this is not currently supported by Safari but is [supported in all other browsers](https://caniuse.com/?search=hidden%3Duntil-found) (though it might be on the way to Safari too)

[Example implementation](https://github.com/user-attachments/files/24100680/demo.html)

### Demo URL

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Confidentiality

nuria1110:
fix #7672

### Proposed behaviour

Adds `hidden="until-found"` to the `Accordion'`s content container to support search for content within collapsed panels via browser find-in-page functionality (Ctrl+F or Cmmd+F) .

### Current behaviour

`Accordion`'s content cannot be found within a page via browser find-in-page functionality (Ctrl+F or Cmmd+F) . 

### Checklist

#### QA

### Additional context

Support for `hidden="until-found"` is quite new:
- Chrome 102 added support in 2022
- Firefox 139 added support in May 2025
- Safari 26.2 added support in December 2025 (latest)

React have yet to add support for `hidden="until-found"` and `onBeforeMatch` event listeners. 
See https://github.com/facebook/react/issues/24740

### Testing instructions

Make sure to test in versions of browsers that have introduced support, specifically Safari only supports this in the latest version of macOS Tahoe 26.2. 

In all browsers when searching for content within a page, if there are any matches within the Accordion's collapsed panel, the panel should expand. 

Safari and Chrome seem to struggle to then highlight matching content however this should then appear when amending the string being searched for. This is deemed as default browser behaviour by accessibility and should hopefully get addressed in future versions since FireFox has no issue with this. 

There is a known bug in Safari and Firefox where they do not scroll to the matching text. See:
https://bugs.webkit.org/show_bug.cgi?id=304174
https://bugzilla.mozilla.org/show_bug.cgi?id=2006040 (fixed in 148)


src/components/accordion/accordion.component.tsx
```diff
@@ -133,6 +134,37 @@ export const Accordion = React.forwardRef<
...
+    }, [isExpanded]);
+
+    useEffect(() => {
+      const container = accordionContentContainer.current;
+      // istanbul ignore next
+      if (!container) return;
+
+      const handleBeforeMatch = (e: Event) => {
+        toggleAccordion(
+          e as unknown as
```
tomdavies73, line 157: **question** (non-blocking): is there any way we can avoid casting as unknown here?

nuria1110: I'm not too sure tbh, the type of event being returned by `beforematch` is a native event and not a react one so this was the only way I could think of doing this. Happy to take suggestions though if you can think of a nicer way of doing this

PR #7711, opened 2026-01-15
feat(vertical-menu-item): add onClick & customIcon

tomdavies73:
### Proposed behaviour

Adds a public facing `onClick` handler which provides consumers with the means to trigger specific behaviour on click. 

Also adds a `customIcon` prop which will render in the same place and take precedence over the `iconType` prop. This means consumers can now pass any valid react node in place of the icon, providing more flexibility and the ability to add custom Icons which are not natively available in Carbon via a SVG etc. 

Screenshot 2026-01-15 at 17 12 31

Screenshot 2026-01-15 at 17 12 40

### Current behaviour

Currently, there is only an internal `onClick` handler which is used to trigger the open/closed state of the `VerticalMenu`, there is no external handler consumers can use to trigger specific behaviour on click.

Also, consumers can only set a specific `Icon` on each `VerticalMenuItem` via the `iconType` prop.

### Checklist

#### QA

### Testing instructions

The new story added to `VerticalMenuItem`, `ItemWithOnClickHandler` will show how an `onClick` is triggered on an item which is an anchor or button element. 

The new test story added to `VerticalMenuItem` will show a new SVG icon being used in both test stories, this shows how the `customIcon` prop now works.

### DIFF
--- a/.storybook/manager.ts
+++ b/.storybook/manager.ts
@@ -9,18 +9,14 @@ import { InteractionToggle } from "./interaction-toggle/reduced-motion";
 import VersionPicker from "./version-picker";
 import { API_PreparedIndexEntry, API_StatusObject } from "@storybook/types";
 
-const useVersionPicker = process.env.USE_VERSION_PICKER === "true";
-
-if (useVersionPicker) {
-  addons.register(ADDON_ID, () => {
-    addons.add(TOOL_ID, {
-      type: types.TOOL,
-      title: "Version picker",
-      match: ({ viewMode }) => !!(viewMode && viewMode.match(/^(story|docs)$/)),
-      render: VersionPicker,
-    });
+addons.register(ADDON_ID, () => {
+  addons.add(TOOL_ID, {
+    type: types.TOOL,
+    title: "Version picker",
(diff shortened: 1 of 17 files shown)


src/components/vertical-menu/vertical-menu-item/vertical-menu-item.test.tsx
```diff
@@ -107,6 +124,27 @@ describe("VerticalMenuItem", () => {
       IconMock.mockClear();
     });
 
+    it("should render a custom icon over an Icon with a set icon type when both `customIcon` & `iconType` are passed", () => {
```
DipperTheDan: **suggestion(non-blocking):** I prefer the test description below where you described `should render anchor with precedence when both `href` and children are passed`, So something along those lines I think is more fitting for this too.

```suggestion
    it("should render a custom icon with precedence when both `customIcon` & `iconType` are passed", () => {
```

tomdavies73: Agreed, good shout

DipperTheDan (approved):

PR #7729, opened 2026-01-22
refactor(loader): declare keyframes instead of having them inline - FE-7578

DipperTheDan:
### Proposed behaviour

Refactor Loader keyframes to be declared as `const` due to a known issue with styled-components where having them inline can cause visual bugs.

### Current behaviour

Keyframes are declared inline using the `@keyframes` syntax.

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

There should be no visual regressions with Loader.

### DIFF
--- a/src/components/loader/__next__/loader.test.tsx
+++ b/src/components/loader/__next__/loader.test.tsx
@@ -219,11 +219,20 @@ test("renders correctly when `loaderType` is `ring` and `inverse` prop is set", 
 
 test("renders correctly when `loaderType` is `ring` and `trackedAnimation` prop is set", () => {
   render(<Loader loaderLabel="Loading" loaderType="ring" isTracked />);
-  expect(screen.getByRole("presentation")).toHaveStyleRule(
-    "animation-name",
-    "trackedAnimation",
+  const element = screen.getByRole("presentation");
+  const styles = window.getComputedStyle(element);
+
+  expect(styles.animationName).not.toBe("none");
+
+  expect(element).toHaveStyleRule(
+    "animation-duration",
+    expect.stringMatching(/\d+(\.\d+)?s/),
     { modifier: "circle[data-role='inner-arc']" },
   );
+
+  expect(element).toHaveStyleRule("stroke-dashoffset", "95px", {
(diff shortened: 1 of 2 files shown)


src/components/loader/__next__/loader.test.tsx
```diff
@@ -221,7 +221,7 @@ test("renders correctly when `loaderType` is `ring` and `trackedAnimation` prop
   render(<Loader loaderLabel="Loading" loaderType="ring" isTracked />);
   expect(screen.getByRole("presentation")).toHaveStyleRule(
     "animation-name",
-    "trackedAnimation",
+    expect.stringMatching(/^[a-zA-Z]{6}$/), // matches a 6-letter hash of the tracked animation
```
DipperTheDan: I am not too happy with this. It's needed for coverage. The test's goal is to verify that the tracked animation is applied, not to check the exact string value of the generated hash. Styled-components intentionally generates unique names to avoid conflicts, so testing for the hash pattern was a last resort. I am more than happy to do something else if a reviewer has a better way to approach it.

edleeks87: Suggestion (non-blocking): Rather than checking the dynamic name of the animation we could test that the animation and properties are correct

```
const element = screen.getByRole("presentation");
const styles = window.getComputedStyle(element);

// animation exists
expect(styles.animationName).not.toBe("none");

// test animation property
expect(element).toHaveStyleRule(
  "animation",
  expect.stringMatching(/\d+(\.\d+)?s/), // matches duration like "0.3s" or "1s"
);

// properties animation affects
expect(element).toHaveStyleRule("left", "0%");
expect(element).toHaveStyleRule("width", "15px");

// SVG animations
expect(element).toHaveStyleRule("stroke-dashoffset", "20");

edleeks87: What you have already is fine though if you prefer to leave it

DipperTheDan: Yeah, doing it your way actually makes more sense than my current test of just checking the animation hash exists. I'll give it a shot and see if it gets coverage.

PR #7730, opened 2026-01-22
test(accordion): refine accordion tests - (FE-7567)

sianford:
### Proposed behaviour

Our unit testing and regression testing frameworks need to be optimised. Use GitHub Copilot in VS Code to analyse the `*.test.tsx`, `*.pw.tsx`, and `*.stories.tsx` files to rapidly identify duplicate tests across the platforms, remove unnecessary code, flaky tests and speed up the carbon repo build time.

### Current behaviour

The Playwright test suite takes ~20 minutes to run. Some tests are flaky, requiring repeated reruns which adds additional time to the build. Historically, requirements around test coverage have resulted in tests being duplicated in the regression test suite that are already in unit test, resulting in a longer build time than is necessary. Accessibility tests exist for most components that would not detect an accessibility violation.

### Checklist

#### QA

### Testing instructions

Aim to move individual keystroke and prop test cases into Jest where possible if not covered already, move styling tests, presentation type (icon positions, button types etc) tests, and user interactions into Chromatic snapshot tests, leave convoluted test scenarios and a11y tests in Playwright. Assess coverage and remove unnecessary tests that add no value, particularly among the accessibility test cases. More details are contained in the associated tickets in Jira, including suggested successful prompts and guidelines. After implementing the changes:

1. Check Jest coverage is still at 100% after changes have been made (run `npm run test`).
2. Check the Playwright suite runs without error.
3. Ensure Chromatic differences are expected. The changes have not caused any unexpected changes in other components.

### DIFF
--- a/src/components/accordion/accordion-interaction.stories.tsx
+++ b/src/components/accordion/accordion-interaction.stories.tsx
@@ -1,11 +1,12 @@
 import { action } from "@storybook/addon-actions";
 import { StoryObj } from "@storybook/react";
 import { userEvent, within, expect } from "@storybook/test";
-import React from "react";
+import React, { useState } from "react";
 
 import { Accordion, AccordionGroup } from ".";
 import Box from "../box";
 import Textbox, { TextboxProps } from "../textbox";
+import Button from "../button";
 
 import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
 import DefaultDecorator from "../../../.storybook/utils/default-decorator";
@@ -107,58 +108,14 @@ ClickToOpen.parameters = {
   chromatic: { disableSnapshot: false },
 };
 
-export const FocusManagement: Story = {
-  render: (args) => (
-    <Accordion
-      onChange={action("expansionToggled")}
-      {...{
-        ...args,
-        "data-role": "accordion",
-        customPadding: 0,
-        title: "Title",
-        subTitle: "Sub Title",
-        width: "100%",
-      }}
-    >
-      <Box mt={2}>Content1</Box>
-      <Box>Content2</Box>
-      <Box>Content3</Box>
-    </Accordion>
-  ),
-  play: async () => {
-    if (!allowInteractions()) {
-      return;
-    }
-    await userEvent.tab();
-
-    await userEvent.keyboard("{Enter}", { delay: 100 });
-    await expect(
-      await within(document.body).findByText("Content1"),
-    ).toBeVisible();
-    await userEvent.keyboard("{Enter}");
-  },
-  decorators: [
-    (StoryToRender) => (
-      <DefaultDecorator>
-        <StoryToRender />
-      </DefaultDecorator>
-    ),
-  ],
-};
-
-FocusManagement.storyName = "Focus Management";
-FocusManagement.parameters = {
-  themeProvider: { chromatic: { theme: "sage" } },
-  chromatic: { disableSnapshot: false },
-};
-
-export const MixedAccordionStates: Story = {
+export const MixedAccordionStatesAndIcons: Story = {
   render: (args) => (
     <AccordionGroup>
       <Accordion
         {...args}
         onChange={action("expansionToggled")}
         title="First Accordion"
+        iconType="chevron_down"
       >
         <Box p={2}>
           <Textbox
@@ -172,6 +129,7 @@ export const MixedAccordionStates: Story = {
         {...args}
         onChange={action("expansionToggled")}
         title="Second Accordion"
+        iconType="chevron_down_thick"
       >
         <Box p={2}>
           <Textbox
@@ -185,6 +143,7 @@ export const MixedAccordionStates: Story = {
         {...args}
         onChange={action("expansionToggled")}
         title="Third Accordion"
+        iconType="dropdown"
       >
         <Box p={2}>
           <Box mt={2}>Content</Box>
@@ -203,9 +162,12 @@ export const MixedAccordionStates: Story = {
     const accordionToggles = canvas.getAllByRole("button");
 
     await userEvent.click(accordionToggles[2]);
-    await userEvent.click(accordionToggles[0], { delay: 200 });
-    await userEvent.click(accordionToggles[2], { delay: 200 });
-    await userEvent.click(accordionToggles[1], { delay: 200 });
+    await userEvent.click(accordionToggles[0]);
+    await expect(accordionToggles[0]).toHaveAttribute("aria-expanded", "true");
+    await userEvent.click(accordionToggles[2]);
+    await expect(accordionToggles[2]).toHaveAttribute("aria-expanded", "false");
+    await userEvent.click(accordionToggles[1]);
+    await expect(accordionToggles[1]).toHaveAttribute("aria-expanded", "true");
     await expect(accordionToggles[1]).toHaveFocus();
   },
   decorators: [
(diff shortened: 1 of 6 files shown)


src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -152,13 +152,14 @@ FocusManagement.parameters = {
   chromatic: { disableSnapshot: false },
 };
 
-export const MixedAccordionStates: Story = {
+export const MixedAccordionStatesAndIcons: Story = {
```
edleeks87, line 111: Comment: I think it's worth doing something similar for the icon position stuff, there's not much value in the current PW tests for it, they're very much tied to the implementation and subject to failing if/when we change the styling etc, realistically a chromatic snapshot of both is all we need

src/components/accordion/accordion.pw.tsx
```diff
@@ -160,49 +136,57 @@ test.describe("should render Accordion component", () => {
     },
   );
 
-  (["left", "right"] as const).forEach((iconAlign) => {
-    test(`should set Accordion iconAlign to ${iconAlign}`, async ({
-      mount,
+  // Separated out conditions to two tests to avoid having conditionals in tests
+  test("should set Accordion iconAlign to right", async ({ mount, page }) => {
```
edleeks87: See comment above about chromatic snapshotting this stuff instead

src/components/accordion/accordion.pw.tsx
```diff
@@ -331,64 +285,6 @@ test.describe("should render Accordion component", () => {
...
-    await mount(<AccordionComponent title="Closed" openTitle="Open" />);
-
-    await expect(accordionTitleContainer(page)).toContainText("Closed");
-
-    await accordionIcon(page).nth(0).click();
-
-    await expect(accordionTitleContainer(page)).toContainText("Open");
-  });
-
   test("should verify accordion subtitle does not render when variant is subtle", async ({
```
edleeks87: comment: this can just be a unit test 

if one doesn't exist something like this will work

```
test("should not render `subtitle` when `variant` is 'subtle'", () => {
   render(
      <Accordion
         variant="subtle"
         title="title"
         subtitle="subtitle"
      >
        FOO
      </Accordion>
   );

  expect(screen.queryByText("subtitle").not.toBeInTheDocument();
});

File: src/components/accordion/accordion.pw.tsx [hunk 4]
```diff
@@ -421,72 +317,8 @@ test.describe("should render Accordion component", () => {
...
-      "box-shadow",
-      "rgb(255, 188, 25) 0px 0px 0px 3px, rgba(0, 0, 0, 0.9) 0px 0px 0px 6px",
-    );
-    await expect(accordionTitleContainer(page).nth(0)).toBeVisible();
-  });
-});
-
 test.describe("should change content height when children change", () => {
-  test("should have proper height", async ({ mount, page }) => {
+  test("should have correct height", async ({ mount, page }) => {
```
edleeks87: Comment: this could potentially be captured in chromatic with an interaction test

src/components/accordion/accordion.pw.tsx
```diff
@@ -555,52 +378,6 @@ test.describe("Accessibility tests for Accordion", () => {
...
-      mount,
-      page,
-    }) => {
-      await mount(<AccordionDefault m={margin} p={padding} />);
-
-      await checkAccessibility(page);
-    });
-  });
-
   [0, 1, 2, 3, 4, 5, 6].forEach((padding) => {
```
edleeks87: Comment: does updating the spacing have any impact on accessibilty?

src/components/accordion/accordion.pw.tsx
```diff
@@ -663,15 +431,6 @@ test.describe("Accessibility tests for Accordion", () => {
...
-  test("should pass accessibility tests for Accordion when title is a string", async ({
-    mount,
-    page,
-  }) => {
-    await mount(<AccordionDefault title="title" />);
-
-    await checkAccessibility(page);
-  });
-
   test("should pass accessibility tests for Accordion when title is a React node", async ({
```
edleeks87: comment: this is a little bit implementation specific as it's down to the implementing dev what they pass etc, they can pass something with a bg/colour that fails contrast etc

src/components/accordion/accordion.pw.tsx
sianford: Removed.

src/components/accordion/accordion.pw.tsx
sianford: Moved to Chromatic.

src/components/accordion/accordion.pw.tsx
sianford: Moved to unit test.

src/components/accordion/accordion.pw.tsx
sianford: Moved to Chromatic.

src/components/accordion/accordion-interaction.stories.tsx
sianford: Moved to Chromatic.

src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -290,111 +252,49 @@ NestedComponentInteractions.parameters = {
...
-        </Box>
-      </Accordion>
-    </AccordionGroup>
-  ),
-  play: async ({ canvasElement }) => {
-    if (!allowInteractions()) {
-      return;
+const DynamicContentHeightComponent = () => {
+  const [contentCount, setContentCount] = useState(3);
+  const modifyContentCount = (modifier: number) => {
```
edleeks87, line 257: suggestion (non-blocking):

```
const modifyContentCount = (modifier: number) => {
    if (!contentCount && modifier === -1) {
      return;
    }
    
    setContentCount(contentCount + modifier);
  };

File: src/components/accordion/accordion.pw.tsx [hunk 8]
```diff
@@ -378,28 +234,11 @@ test.describe("Accessibility tests for Accordion", () => {
...
-          }}
-        />,
-      );
-
-      await checkAccessibility(page);
-    });
-  });
-
-  test("should pass accessibility tests for Accordion with Box and different paddings", async ({
+  test("should pass accessibility tests for Accordion with Box and custom paddings", async ({
```
edleeks87: Question: what accessibility issue are we checking for when using the spacing interface? Either way this is a contrived scenario that we are confirming has no accessibility issues but more than likely won't reflect actual implementations etc

sianford: I figured it would highlight if any items would be obscured by objects above. Since removed.

src/components/accordion/accordion.pw.tsx
sianford: Removed.

src/components/accordion/accordion-interaction.stories.tsx
```diff
@@ -287,111 +252,49 @@ NestedComponentInteractions.parameters = {
...
-    ).toBeVisible();
-    await userEvent.keyboard(" ");
+    const button = canvas.getAllByRole("button");
+    const accordionToggle = canvas.getAllByRole("button")[2];
+
+    await userEvent.click(accordionToggle);
+    await userEvent.click(button[0]);
+    await userEvent.click(button[0]);
+    await userEvent.click(button[1]);
+    await userEvent.click(button[1]);
```
paulrobinson88, line 297: **comment** (non blocking): There doesn't appear to be an assertion such as `await expect().toHaveAttribute("aria-expanded", "true");` on this test like the others meaning it would pass regardless of the outcome.
I know we are really only using the interaction tests for snapshots mainly so it may not be required in this instance.

paulrobinson88 (commented):
Hi Sian, just one comment from me which may be fine to go ahead as is. 
Thanks

src/components/accordion/accordion-interaction.stories.tsx
sianford: Great spot. Yes, I removed the previous text search assertions and delays and replaced with more involved addition and removal of content. However, I hadn't added an assertion to verify the content is now correct. While it is mainly there to capture the screenshot for Chromatic, it still needed to assert the content is correct. Updated.

edleeks87 (approved):

PR #7732, opened 2026-01-23
docs: update usage docs

tomdavies73:
### Proposed behaviour

Updates the usage docs to remove any mentions of prop types and offer direct guidance regarding component imports.

### Current behaviour

N/A

### Checklist

#### QA

### DIFF
--- a/docs/usage.mdx
+++ b/docs/usage.mdx
@@ -7,13 +7,14 @@ import { Meta } from "@storybook/blocks";
 ## Contents
 
 - [Quick start](#quick-start)
+- [Importing components](#importing-components)
 - [Globals](#globals)
   - [Global styles](#global-styles)
   - [Theming](#theming)
   - [Localisation](#localisation)
   - [Validation](#validation)
-  - [A Note On PropTypes](#a-note-on-proptypes)
-
+  - [PropTypes Support Removed](#proptypes-support-removed)
+  
 ## Quick start
 
 A basic project `index.(jsx|tsx)` file would resemble this example:
@@ -44,6 +45,19 @@ root.render(<App />);
 You can also find this in this live sandbox:
 
 [![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/fork/github/Parsium/carbon-starter)
+
+
+## Importing components
+
+All components should be imported from their respective component index paths. For example:
+
+```jsx
+import { StepFlow, StepFlowTitle, StepFlowHandle, Steps } from "carbon-react/lib/components/step-flow";
+import Textbox from "carbon-react/lib/components/textbox";
+```
+**Important:** Only import components from their index paths as shown above. Any files such as component, context or style files are implementation details and should not be imported directly. 
+These files may change or be removed at any time without notice, so importing from them directly could break your application unexpectedly.
+Always refer to the documented import paths for each component.
 
 ## Globals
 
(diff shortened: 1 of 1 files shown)


docs/usage.mdx
```diff
@@ -7,12 +7,12 @@ import { Meta } from "@storybook/blocks";
 ## Contents
 
 - [Quick start](#quick-start)
+- [Importing components](#importing-components)
 - [Globals](#globals)
   - [Global styles](#global-styles)
   - [Theming](#theming)
   - [Localisation](#localisation)
   - [Validation](#validation)
-  - [A Note On PropTypes](#a-note-on-proptypes)
```
edleeks87, line 15: I think in the long run we can remove this but for now I think updating it is better, something saying we no longer support `prop-types` but if you require type checking in a non-typescript project you can utilise a tsconfig and `checkJS: true` etc

docs/usage.mdx
```diff
@@ -45,6 +45,24 @@ You can also find this in this live sandbox:
...
+
+All components should be imported from their respective component index paths. For example:
+
+```jsx
+import { StepFlow, StepFlowTitle, StepFlowHandle, Steps } from "carbon-react/lib/components/step-flow";
+import Textbox from "carbon-react/lib/components/textbox";
+```
+
+Some components use named exports while others use default exports depending on their design and use cases. 
+Named exports are typically used when a component has related sub-components or utilities that are commonly imported together (for example, `StepFlow` with `StepFlowTitle`, `StepFlowHandle` and `Steps`).
```
edleeks87: comment (non-blocking): to me this more info than is needed wrt the types of exports etc

docs/usage.mdx
```diff
@@ -45,6 +45,24 @@ You can also find this in this live sandbox:
...
+```jsx
+import { StepFlow, StepFlowTitle, StepFlowHandle, Steps } from "carbon-react/lib/components/step-flow";
+import Textbox from "carbon-react/lib/components/textbox";
+```
+
+Some components use named exports while others use default exports depending on their design and use cases. 
+Named exports are typically used when a component has related sub-components or utilities that are commonly imported together (for example, `StepFlow` with `StepFlowTitle`, `StepFlowHandle` and `Steps`). 
+Default exports are used for standalone components that are typically used in isolation. The import paths documented in our component documentation reflect the correct export type for each component.
+
+**Important:** Only import components from their index paths as shown above. Internal files such as component, context or style files are implementation details and should not be imported directly.
```
edleeks87: comment: I think we should avoid `internal` as it might confuse given we have files in the actual internal folders etc. Saying `Any files such as...` works better for me as it's conveying anything imported from non entry files is against our advice etc

docs/usage.mdx
```diff
@@ -45,6 +45,24 @@ You can also find this in this live sandbox:
...
+import { StepFlow, StepFlowTitle, StepFlowHandle, Steps } from "carbon-react/lib/components/step-flow";
+import Textbox from "carbon-react/lib/components/textbox";
+```
+
+Some components use named exports while others use default exports depending on their design and use cases. 
+Named exports are typically used when a component has related sub-components or utilities that are commonly imported together (for example, `StepFlow` with `StepFlowTitle`, `StepFlowHandle` and `Steps`). 
+Default exports are used for standalone components that are typically used in isolation. The import paths documented in our component documentation reflect the correct export type for each component.
+
+**Important:** Only import components from their index paths as shown above. Internal files such as component, context or style files are implementation details and should not be imported directly. 
+These internal files may change or be removed at any time without notice, so importing from them directly could break your application unexpectedly.
```
edleeks87: praise: I like this as long as we remove the internal part

docs/usage.mdx
tomdavies73: I thought this was worth including as we have had inquiries in Carbon before about named/default exports not working when they should be the other way round etc.. It could be worth having especially for those who are slightly unfamiliar with import syntax, but happy to remove if others agree

docs/usage.mdx
```diff
@@ -86,11 +105,19 @@ Carbon provides built-in validation states for input components. For more inform
 
 To opt into the new validation pattern, set the `validationRedesignOptIn` flag to true in the [CarbonProvider](../?path=/docs/carbon-provider--docs).
 
-### A Note On PropTypes
+### PropTypes Support Removed
+
+As of the React 19 upgrade, Carbon components no longer support `prop-types`. This change encourages the use of TypeScript for type safety and better development experience.
```
edleeks87:
```suggestion
As of the React 18 upgrade, Carbon components no longer support `prop-types`. This change encourages the use of TypeScript for type safety and better development experience.
```

docs/usage.mdx
edleeks87: The issue for me is the above doesn't fully capture what leads to a given export pattern and in the long run we will likely consolidate them to be named

PR #7740, opened 2026-01-29
fix(tabs): ensure __next__ version updates focus index when user tabs off component

edleeks87:
### Proposed behaviour

Ensures that when a user tabs away from the `__next__` version `Tabs` that the index controlling the focus order is updated so that when the `Tabs` are refocused and the user uses arrow keys to navigate the expected `Tab` is focused.

### Current behaviour

Focus order is incorrect if a user tabs away from the component then refocuses and uses arrow key, the first tab is focussed each time etc

### Checklist

#### QA

### DIFF
--- a/src/components/tabs/__next__/tabs.test.tsx
+++ b/src/components/tabs/__next__/tabs.test.tsx
@@ -565,3 +565,47 @@ test("invokes the onTabChange handler the active tab changes", async () => {
   await user.click(tab1);
   expect(mockTabChangeFn).toHaveBeenCalledWith("tab-1");
 });
+
+test("applies the `data-element` and `data-role` props to expected elements", () => {
+  render(
+    <Tabs data-element="tabs-foo" data-role="tabs-bar">
+      <TabList
+        ariaLabel="Sample Tabs"
+        data-element="tablist-foo"
+        data-role="tablist-bar"
+      >
+        <Tab
+          data-element="tab-foo"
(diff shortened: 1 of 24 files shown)


src/components/tabs/__next__/tabs.test.tsx
```diff
@@ -205,6 +205,31 @@ test("responds to horizontal keyboard interactions", async () => {
   expect(screen.getByRole("tabpanel")).toHaveTextContent("Content 1");
 });
 
+test("updates the focus index when after a user presses 'tab' key", async () => {
```
DipperTheDan: **nitpick(non-blocking):** This makes more sense to me

```suggestion
test("updates the focus index when a user presses 'tab' key", async () => {
```

src/components/tabs/__next__/tabs.test.tsx
```diff
@@ -260,6 +285,31 @@ test("responds to vertical keyboard interactions", async () => {
   expect(screen.getByRole("tabpanel")).toHaveTextContent("Content 1");
 });
 
+test("updates the focus index when after a user presses 'tab' key and `orientation` is 'vertical'", async () => {
```
DipperTheDan: same thing here.

```suggestion
test("updates the focus index when a user presses 'tab' key and `orientation` is 'vertical'", async () => {
```

dariusbercea-sage (approved):

PR #7743, opened 2026-01-29
build: remove babel, configure swc to be main test transpiler

tomdavies73:
### Proposed behaviour

Removes all `babel` related dependencies and instead uses `swc/jest` as the main test transpiler. In order to transition to `swc/jest` some additional changes have also been made. 

`babel/jest`'s less aggressive tree-shaking meant it would often leave unused exports in place during transpilation. `swc/jest` takes a stricter approach - it analyses exports more carefully and expects us to be more explicit about what we're exporting and using.

All exports which are unused (not imported anywhere else internally) were highlighted and have had to be removed. 

There were multiple instances of consts which were only used locally in one file being exported, or instances of some costs being exported and never being used locally or elsewhere.

In some files we have also "doubled up" our exports, exporting twice as a named export and a default export. Where possible the named export has been removed, and only the default export remains.

Some additional steps have been made to meet coverage requirements with some exports which are needed but not directly tested, as `instanbul ignore` is now ineffectual against uncovered const statements

The removal of unused exports has been marked as a breaking change as technically some consumers may have imported these directly from legacy projects etc. However, the actual likelihood that this will break consumers' applications is slim to none. 

`config.ts` and `index.ts` files are now exempt from the coverage report, as `config.ts` consts are typically used in stories and therefore are treated as unused. `index.ts` files exports are consumer facing entry points, so they have also been exempt as some exports are never imported and used elsewhere. 

### Current behaviour

Currently, there are multiple `babel` dependencies which are unused and `babel/jest` is used for test transpilation.

### Checklist

#### QA


.storybook/version-picker/index.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from "@storybook/components";
 import compareBuild from "semver/functions/compare-build";
 
-import { TOOL_ID } from "./constants";
+import { TOOL_ID } from "./constants/constants";
```
edleeks87, line 10: suggestion: you can just make `constants.ts` an `index.ts`

src/__internal__/focus-trap/focus-trap.component.tsx
edleeks87, line 287: comment: not part of your changes but we should add the following to the top of the file imo so we don't get the linter warnings anymore

```
/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
/* eslint-disable jsx-a11y/no-static-element-interactions */

File: jest.config.ts [hunk 2]
```diff
@@ -25,6 +25,8 @@ const baseProjectConfig: Config = {
     "<rootDir>/src/locales",
     "<rootDir>/lib",
     "<rootDir>/esm",
+    ".*index\\.ts$",
+    ".*[.-]config\\.ts$",
```
edleeks87, line 29: comment: we're still including files that aren't needed in the coverage report

Image

edleeks87: I think it's worth generating a report from a test run in master and comparing it to one from this PR

src/components/action-popover/action-popover-item/action-popover-item.component.tsx
```diff
@@ -81,7 +81,7 @@ function calculateSubmenuPosition(
   return windowWidth >= right + offsetWidth ? "right" : "left";
 }
 
-export const ActionPopoverItem = ({
+const ActionPopoverItem = ({
```
edleeks87, line 84: comment: I think this change has broken the prop table for the sub-components

Image

tomdavies73:
Closing due to changes being unviable, for more information see the supporting doc written and attached to FE-7424

PR #7750, opened 2026-02-04
feat(tabs): add de-De, fr-CA, fr-FR & pt-PT status translations

tomdavies73:
### Proposed behaviour

Adds:

1. `de-DE`
2. `fr-CA`
3. `fr-FR`
4. `pt-PR`

Translations for info, error and warning states. `es-ES` will be added a later date, consumers are advised to use a local override if needed.

### Current behaviour

Currently, only an `en-GB` translation is provided for info, error and warning states

### Checklist

#### QA

### Testing instructions

The `With Error And Warning` stories should now have a correct accessible name which contains the tab title in every locale apart from spanish (es-ES).

1. Navigate to the  `With Error And Warning` story 
2. Turn on voiceOver 
3. Tab (pardon the pun) onto a Tab 
4. Assert that the accessible name is correct for the respective locale (swap locales using the locale dropdown at the top of story book)

Note: you can see the original translations in the attached JSON files on the parent ticket on JIRA (FE-7580)


src/style/utils/extract-text.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (!node) {
+    return "";
+  }
+
+  if (typeof node === "string" || typeof node === "number") {
+    return String(node);
+  }
+
+  if (Array.isArray(node)) {
+    return node.map(extractTextFromNode).join("");
```
DipperTheDan: **suggestion(non-blocking):** If we have an array of React nodes like `[<span>First</span>, <span>Second</span>]`, joining with an empty string produces "FirstSecond", but joining with a space produces "First Second", which is more likely what we want for accessible labels and generally reads better. I might just be thinking of an edge-case scenario though. 

```suggestion
    return node.map(extractTextFromNode).join(" ");
```

src/__internal__/utils/helpers/extract-text/index.ts
DipperTheDan, line 1: And just so we don't forget. This and the test file will need moving to a more suitable location than `style/utils`.

divyajindel (approved):

PR #7759, opened 2026-02-06
test(badge): refine badge tests - (FE-6443)

divyajindel:
### Proposed behaviour
Eliminated all Playwright tests in badge.pw.tsx except accessibility tests, and added missing test coverage to the Jest test file.

### Current behaviour
A number of tests are present in multiple test suites, and certain flaky tests may cause failures or substantially increase execution time.

### Checklist

#### QA

### Additional context
This should significantly reduce the amount of time taken for Playwright test to run in the Badge component without losing any coverage in Jest.

### Testing instructions


src/components/badge/badge.pw.tsx
```diff
@@ -1,117 +1,13 @@
...
-    test(`should check Badge counter is not visible when passing ${countInput}`, async ({
-      mount,
-      page,
-    }) => {
-      await mount(<BadgeComponent counter={countInput} />);
-      await expect(badgeCounter(page)).toBeHidden();
-    });
-  });
-
-  test("should render with correct styles when onClick is passed and Badge is hovered", async ({
```
paulrobinson88, line 60: **comment**: This test appears to check that the correct badge and cross styles are applied when using an `onClick` and the badge is hovered. The Jest test does cover that the cross is rendered but the styles aren't checked. We may want to either add this into Jest, or potentially create an interaction test which could check via a snapshot. 
However, as the onClick is deprecated it may not be required, I'll revert to whatever the other reviewer thinks.

paulrobinson88 (commented):
Hi Divya, good work on this, just one comment from me.
Thanks.

src/components/badge/badge.pw.tsx
edleeks87: I think for deprecated props we should test functionality but styling is less of an issue imo

src/components/badge/badge.test.tsx
```diff
@@ -77,6 +83,12 @@ test("should not render badge when counter is `0` as a number", () => {
   expect(screen.queryByTestId("badge")).not.toBeInTheDocument();
 });
 
+test("should not render badge when counter is a negative number", () => {
```
sianford: Not blocking, but we don't need to state the component name in this line for any tests. When the tests run, it is clear that it is for Badge, so just `"should not render when counter is a negative number"` is fine.

src/components/badge/badge.test.tsx
```diff
@@ -103,6 +115,13 @@ test("should render with provided id", () => {
...
 });
 
+test("should render with provided id when used as a button", () => {
+  render(<Badge counter={9} id="badge-button-id" onClick={() => {}} />);
+
+  const badgeButton = screen.getByRole("button");
+  expect(badgeButton).toHaveAttribute("id", "badge-button-id");
+});
+
 test("should not render as a button if onClick is not set", () => {
```
sianford: `onClick` has been deprecated. This test can be removed.

src/components/badge/badge.test.tsx
```diff
@@ -103,6 +115,13 @@ test("should render with provided id", () => {
   expect(badge).toHaveAttribute("id", "custom-id");
 });
 
+test("should render with provided id when used as a button", () => {
+  render(<Badge counter={9} id="badge-button-id" onClick={() => {}} />);
```
sianford: `onClick` has been deprecated. That prop can be removed.

src/components/badge/badge.pw.tsx
sianford: @edleeks87 so should the deprecated prop tests remain in Jest then?

src/components/badge/badge.test.tsx
sianford: The `id` prop is not dependent on having an `onClick`. I think a diacritics type test for `id` is better placed in the Playwright file.

src/components/badge/badge.test.tsx
```diff
@@ -77,6 +83,12 @@ test("should not render badge when counter is `0` as a number", () => {
   expect(screen.queryByTestId("badge")).not.toBeInTheDocument();
 });
 
+test("should not render badge when counter is a negative number", () => {
+  render(<Badge data-role="badge" counter={-1} />);
+
+  expect(screen.queryByTestId("badge")).not.toBeInTheDocument();
+});
+
 test("should not render badge when counter is a decimal number", () => {
```
sianford: As you're in this file already making changes, it would be good to tidy up this ine by removing "badge" from here, as per the comment above to keep the naming uniform.

src/components/badge/badge.test.tsx
divyajindel: Yes , That's right @sianford but after discussing with Ed and Nuria, we still need to keep functionalities tests in jest.

src/components/badge/badge.test.tsx
```diff
@@ -6,31 +6,31 @@ import { assertLoggerComponentMessage } from "../../__spec_helper__/__internal__
 
 jest.mock("../../__internal__/utils/logger");
 
-test("should log a deprecation warning for onClick prop", () => {
+test("should log deprecation warning for onClick prop", () => {
```
sianford: Why did you remove the word "a" from this?

src/components/badge/badge.test.tsx
```diff
@@ -6,31 +6,31 @@ import { assertLoggerComponentMessage } from "../../__spec_helper__/__internal__
...
   assertLoggerComponentMessage({
     component: <Badge onClick={() => {}} />,
     message:
       "The `onClick` prop in `Badge` is deprecated and will soon be removed.",
   });
 });
 
-test("should log a deprecation warning for aria-label prop", () => {
+test("should log deprecation warning for aria-label prop", () => {
```
sianford: Why did you remove the word "a" from this?

src/components/badge/badge.test.tsx
```diff
@@ -6,31 +6,31 @@ import { assertLoggerComponentMessage } from "../../__spec_helper__/__internal__
...
   assertLoggerComponentMessage({
     component: <Badge aria-label="Test label" />,
     message:
       "The `aria-label` prop in `Badge` is deprecated and will soon be removed.",
   });
 });
 
-test("should log a deprecation warning for color prop", () => {
+test("should log deprecation warning for color prop", () => {
```
sianford: Why did you remove the word "a" from this?

src/components/badge/badge.test.tsx
```diff
@@ -53,43 +53,55 @@ test("should render counter as a string", () => {
   expect(screen.getByText("99+")).toBeVisible();
 });
 
-test("should render `999+` when counter is a number higher than 999", () => {
+test("should render counter when it is exactly 999", () => {
```
sianford: This would read better as `should render '999' when counter is 999` as it states the expected result.

src/components/badge/badge.test.tsx
```diff
@@ -187,21 +197,20 @@ test("should render cross icon when onClick is set and badge is hovered", async
   expect(badgeCounter).toBeVisible();
 });
 
-test("should have the relevant aria-label when aria-label is specified", () => {
+test("should have accessible name when aria-label is specified", () => {
```
sianford: This change was unnecessary.

src/components/badge/badge.test.tsx
```diff
@@ -279,8 +284,7 @@ test("should render with correct style when variant is subtle and inverse prop i
   );
 });
 
-// coverage
-test("should render with correct style size is small and badge has children", () => {
+test("should apply small size styling when positioned on children", () => {
```
sianford: This no longer reads correctly. It would read better as `should apply small size styling with children`.

src/components/badge/badge.test.tsx
```diff
@@ -297,8 +301,7 @@ test("should render with correct style size is small and badge has children", ()
   expect(badge).toHaveStyleRule("height", "var(--global-size-4-xs,8px)");
 });
 
-// coverage
-test("should render with correct style size is large and badge has children", () => {
+test("should apply large size styling when positioned on children", () => {
```
sianford: As above, this no longer reads correctly. It would read better as `should apply large size styling with children`.

paulrobinson88 (approved):

PR #7761, opened 2026-02-09
test(action-popover): refine testing suite

paulrobinson88:
### Proposed behaviour
Removes a number of Playwright tests which are duplicated elsewhere in the testing suite and adds more detail to one of the interaction tests in Chromatic for the `Action Popover` component

### Current behaviour
There are a number of tests which exist in more than one testing suite along with flaky tests which can cause the test to fail or take a significant amount of time.

### Checklist

#### QA

### Additional context
This should significantly reduce the amount of time taken for Playwright test to run in the Action Popover component without losing any coverage in Jest.

### Testing instructions


src/components/action-popover/action-popover.pw.tsx
mihai-albu-sage, line 344: question (non-blocking): `boundingBox()` method returns either an object or null. Could we use `expect(iconPosition).not.toBeNull()` instead of `if (!iconPosition) throw new Error("Icon isn't visible");` to avoid any CI / Lint warnings?

src/components/action-popover/action-popover.pw.tsx
```diff
@@ -22,17 +18,9 @@ import {
 } from "../../../src/components/action-popover";
 import {
   ActionPopoverCustom,
-  ActionPopoverMenuWithProps,
-  ActionPopoverPropsComponent,
   ActionPopoverWithIconsAndNoSubmenus,
```
tomdavies73, line 21: **suggetion**: I know we've removed a fair bulk of test components here which is super positive, but I think a manual review of the remaining test components could be beneficial, especially: 

1. `ActionPopoverWithIconsAndNoSubmenus`
2. `ActionPopoverWithProps`
3. `ActionPopoverWithSubmenusAndIcons`
4. `ActionPopoverWithDownloadButton`

Might be worth seeing where these are used, if the tests are useful etc. It may also be worth seeing if these different variations are captured in a chromatic snapshot, if they aren't, this could be worth exploring too 👍

src/components/action-popover/action-popover.pw.tsx
paulrobinson88: Thanks for pointing this one out Mihai, I've attempted to workaround this but it appears to cause issues with the `expect(iconPosition.x).toBeGreaterThan(textPosition.x);` and the linter has an issue when we try to ignore it potentially being null with `if (iconPosition && textPosition) {
  expect(iconPosition.x).toBeLessThan(textPosition.x);
}`
Seems we're doing similar in other areas of the codebase like in the `Dialog` component so may leave this as is for now.

src/components/action-popover/action-popover.pw.tsx
sianford: The only comments I have are along the lines that Tom mentioned above. As a lot of tests have been removed, can you be sure that the styling is covered in either other components or does not warrant a snapshot by Chromatic to check against. Otherwise, good job! The more flaky tests removed from Playwright, the better 👍

tomdavies73 (approved):

sianford (approved):

PR #7763, opened 2026-02-09
feat(tabs): map legacy interface to next version of component

edleeks87:
Maps the legacy interface to the next version of `Tabs` and `Tab`. Marks relevant props as deprecated and adds Logger warning to inform consumers of future removal etc.

### Proposed behaviour

Updates legacy Tabs to map old interface to `next` version of the component, the next versions are now rendered when importing the legacy versions
Marks relevant props on legacy interface as deprecated

### Current behaviour

Legacy Tabs does not render next versions of components

### Checklist

#### QA


DipperTheDan (approved):

src/components/dialog/dialog.pw.tsx
```diff
@@ -27,7 +27,6 @@ import {
   FullScreenWithAutoFocusSelect,
   FullScreenComponentFocusableSelectors,
   FullScreenWithBox,
-  FullScreenWithComplexExample,
```
nuria1110, line 30: **nit-pick (non-blocking):** If this story isn't needed anymore its probably worth also deleting it from the `components.test-pw.tsx` file.

src/components/tabs/__next__/tabs.component.tsx
```diff
@@ -402,17 +427,17 @@ export const TabList = forwardRef<TabsHandle, TabListProps>(
       return leftVisible ? (
         <StyledScrollButton
           data-role="tab-navigation-button-left"
-          id="tab-navigation-button-left"
+          id={`tab-navigation-button-left-${idLeftNav.current}`}
           onClick={() => onClickHandler("left")}
-          size={size}
+          $size={size}
           tabIndex={-1}
           title="Scroll Tabs Left"
```
nuria1110: **comment (non-blocking):** Aware this isn't relevant to your changes but we will probably want a translation key here instead.

src/components/tabs/__next__/tabs.component.tsx
```diff
@@ -423,32 +448,34 @@ export const TabList = forwardRef<TabsHandle, TabListProps>(
       return rightVisible ? (
         <StyledScrollButton
           data-role="tab-navigation-button-right"
-          id="tab-navigation-button-right"
+          id={`tab-navigation-button-right-${idRightNav.current}`}
           onClick={() => onClickHandler("right")}
-          size={size}
+          $size={size}
           tabIndex={-1}
           title="Scroll Tabs Right"
```
nuria1110: Same here, needs translations.

nuria1110 (dismissed):

src/components/tabs/__next__/tabs.component.tsx
edleeks87: Yeah fair point, I'll raise a ticket for the non-English ones and add something for en-GB now

DipperTheDan (approved):

src/components/dialog/dialog.pw.tsx
edleeks87: removed thanks

PR #7785, opened 2026-02-20
test(checkbox): refine testing suite

dariusbercea-sage:
### Proposed behaviour

Removes a number of Playwright tests that duplicate either Jest or Chromatic ones for the `Checkbox ` Component

### Current behaviour

### Checklist

#### QA

### Additional context

### Testing instructions

### DIFF
--- a/src/components/checkbox/checkbox.pw.tsx
+++ b/src/components/checkbox/checkbox.pw.tsx
@@ -2,754 +2,23 @@ import React from "react";
 import { CommonCheckableInputProps } from "../../__internal__/checkable-input/checkable-input.component";
 import { test, expect } from "../../../playwright/helpers/base-test";
 import Box from "../box";
-import { Checkbox, CheckboxProps, CheckboxGroupProps } from ".";
+import { CheckboxProps, CheckboxGroupProps } from ".";
 import {
   CheckboxComponent,
   CheckboxGroupComponent,
-  CheckboxGroupComponentNewValidation,
   Sizes,
   Reversed,
   WithCustomLabelWidth,
 } from "./components.test-pw";
 import {
-  checkboxComponent,
-  checkboxRole,
-  checkboxGroup,
-  checkboxLabel,
-  checkboxgroupLegend,
-  checkboxInlineFieldHelp,
   checkboxIcon,
   checkboxGroupIcon,
-  checkboxSvg,
-  checkboxHelpIcon,
-  checkboxGroupFlex,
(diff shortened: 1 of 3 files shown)


src/components/checkbox/checkbox.pw.tsx
```diff
@@ -1,756 +1,25 @@
...
-          .nth(position);
-        await expect(checkboxElement).toHaveText("Checkbox Label");
-      }
-    });
-  });
-
-  (
-    ["bottom", "left", "right", "top"] as CheckboxProps["tooltipPosition"][]
-  ).forEach((position) => {
-    test(`should render CheckboxComponent component with tooltip positioned to the ${position}`, async ({
```
paulrobinson88, line 333: **Question**(non-blocking): Possibly one for a reviewing dev, we don't appear to have this covered in any `checkbox` tests, however, it is marked as 'legacy' in the docs rather than deprecated. If it is deprecated it may not be required.

src/components/checkbox/checkbox.pw.tsx
```diff
@@ -1,756 +1,25 @@
...
-
-  (
-    [
-      "top",
-      "bottom",
-      "left",
-      "right",
-    ] as CheckboxGroupProps["tooltipPosition"][]
-  ).forEach((position) => {
-    test(`should render CheckboxGroupComponent component with tooltip positioned to the ${position}`, async ({
```
paulrobinson88, line 646: Same as above

src/components/checkbox/checkbox.pw.tsx
dariusbercea-sage: as far as I've understood from the copilot analysis and checked in the Playwright accessibility tests, there is a test that checks the accessibility of the tooltip in all of the positions so they should be tested there.
https://github.com/Sage/carbon/blob/e2159e306b45065d88b86be140f0efbcda75d0e4/src/components/checkbox/checkbox.pw.tsx#L136

DipperTheDan: Yeah the idea is for tooltips to be removed in favour of the input hint text type pattern we have. This discussion has been going on since about 2021, I think and we haven't had the go-ahead to fully remove the tooltip stuff yet. As long as we have a test that covers it somewhere just in case any teams/products are still using it. They shouldn't be, but that hasn't stopped anyone from doing it.

DipperTheDan (approved):

mihai-albu-sage (approved):

PR #7787, opened 2026-02-23
test(message): refine testing suite

divyajindel:
### Proposed behaviour
This PR aims to eliminate redundant Playwright tests in the Message component, keeping only accessibility tests in Playwright.

### Current behaviour
A number of tests are present in multiple test suites, and certain flaky tests may cause failures or substantially increase execution time.

### Checklist

#### QA

### Additional context
This should significantly reduce the amount of time taken for Playwright test to run in the Message component without losing any coverage in Jest.

### Testing instructions

### DIFF
--- a/src/components/message/message-test.stories.tsx
+++ b/src/components/message/message-test.stories.tsx
@@ -49,7 +49,7 @@ export const WithNoTitle: Story = {
   },
   parameters: {
     themeProvider: { chromatic: { theme: "sage" } },
-    chromatic: { disableSnapshot: false },
+    chromatic: { disableSnapshot: true },
   },
 };
 
@@ -73,7 +73,7 @@ export const WithNoCloseIcon: Story = {
   },
   parameters: {
     themeProvider: { chromatic: { theme: "sage" } },
-    chromatic: { disableSnapshot: false },
+    chromatic: { disableSnapshot: true },
   },
 };
 
@@ -91,7 +91,7 @@ export const WithLongTextWrapping: Story = {
   },
   parameters: {
     themeProvider: { chromatic: { theme: "sage" } },
-    chromatic: { disableSnapshot: false },
+    chromatic: { disableSnapshot: true },
   },
 };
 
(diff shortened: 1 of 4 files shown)


mihai-albu-sage (dismissed):

src/components/message/message-interaction.stories.tsx
nuria1110, line 1: **comment:** I don't think interaction tests are necessary for this component as there aren't any interactive elements apart from the Button. If we want to capture the focus styles for the Button we just could add a pseudo selector to one of the test-stories. Happy to see what the other reviewers think.

paulrobinson88: yeah i'd agree, we should be able to apply a pseudo focus to the close button on one of the current test stories which might just be a simple one line change to avoid having to add a new file.

src/components/message/message.pw.tsx
```diff
@@ -1,133 +1,14 @@
 import React from "react";
-import { test, expect } from "../../../playwright/helpers/base-test";
-import {
-  MessageComponent,
-  MessageComponentWithRef,
```
paulrobinson88, line 5: **comment:** As this import `MessageComponentWithRef` has been removed, it would be worth checking if this pattern in the `components.test-pw` file is used elsewhere in the repo, if not, it could be removed.

paulrobinson88 (commented):
Hi Divya, just a couple of comments from me, one thing you may want to look at also is the Chromatic snapshots which are currently being taken, I reckon we can significantly reduce these by disabling some which aren't required and are captured in other stories, or, consolidating together the majority into one test story and capturing a snapshot. 

Currently all of the external stories, apart from `Show Close Icon` and `Transparent` are being snapshotted. Along with `With No Title`, `With No Close Icon` and `With Long Text Wrapping` in the internal test stories.

src/components/message/message-test.stories.tsx
```diff
@@ -50,6 +50,9 @@ export const WithNoTitle: Story = {
   parameters: {
     themeProvider: { chromatic: { theme: "sage" } },
     chromatic: { disableSnapshot: false },
+    pseudo: {
+      focus: 'button[data-element="close"]',
```
paulrobinson88: **comment:** we might need to be a bit more specific to this story as when the pseudo is applied to this story, opening the `With Long Text Wrapping` story, the same is applied. It might be a case of applying a `data-role` or `data-element` to the component in this one story to target.

paulrobinson88 (commented):
**comment**: We do still have some snapshots being taken in message.stories.tsx. Only Show Close Icon and Transparent aren't being captured. I reckon we can disable some of those to reduce down the overall number

paulrobinson88 (approved):

nuria1110 (approved):

mihai-albu-sage (approved):

PR #7792, opened 2026-02-24
test(definition-list): refine testing suite

dariusbercea-sage:
### Proposed behaviour

Removes a number of Playwright tests that duplicate either Jest or Chromatic ones for the `Definition list` component

### Current behaviour

### Checklist

#### QA

### Additional context

### Testing instructions


tomdavies73 (commented):
Nice work @dariusbercea-sage, I think we may be able to strip out more playwright tests though. We still seem to have quite a lot of tests which are testing visuals which should be handled in chromatic. 

In my opinion, the `definition-list.pw.tsx` file could just be accessibility tests, even one single test for a default configuration may be enough in my opinion as all these components are just wrappers for the `dl`, `dt` and `dd` elements with some added styling and functionality. All we would need to be concerned about from an accessibility standpoint is the correct semantic HTML structure and colour contrast. 

Also, I think chromatic may need a review, I've done a quick check of the build there, and we're snapshooting quite a lot of stories in both the typical stories file and the test stories file.

In my opinion we should be snapshotting our default configuration, the `w` prop, the `asSingleColumn` prop, the `dtTextAlign` prop and the `ddTextAlign` prop. Everything else like the usage of `Box` for background colours or other elements are children are not implementation details and are added by consumers so no need to snapshot on our end. 

If you have any questions, feel free to reach out 👍

paulrobinson88 (approved):

src/components/definition-list/definition-list-test.stories.tsx
```diff
@@ -129,7 +129,7 @@ export const UsingBoxToOverrideBackgroundColor = () => (
 );
 
 UsingBoxToOverrideBackgroundColor.parameters = {
-  chromatic: { disableSnapshot: false },
+  chromatic: { disableSnapshot: true },
```
tomdavies73: **suggestion**: we are already disabling the snapshot at the top of the file, instead of changing the value here to be true we can just remove it entirely. 

Same comment applies for the other changes in this file

tomdavies73 (approved):

PR #7793, opened 2026-02-24
test(button-toggle): refine testing suite

paulrobinson88:
### Proposed behaviour
- Removed Playwright testing where tests are either covered in Jest or Chromatic tests.
- Added tests to Jest to cover ones removed from Playwright.
- Disabled Chromatic snapshots where component state is already captured.
- Removed unused code in `components.test-pw` file. 

### Current behaviour
There are a number of Playwright tests which are no longer needed which are causing a loss of efficiency in the testing suite of the repo.

### Checklist

#### QA

### Testing instructions
Ensure Playwright tests pass and Chromatic snapshots are capturing the expected state.


src/components/button-toggle/button-toggle-interaction.stories.tsx
nuria1110, line 1: **comment:** I think that the interactions tests could also be simplified overall, general keyboard navigation and callback testing can be done in unit tests and imo we should only be capturing focus/hover styles here, mainly to avoid over-use of chromatic where its not really necessary. 

We could also avoid disabling the linter in every test if we create re-usable a controlled component, happy to help if you need a hand with this.

```
const ControlledButtonToggleGroup = () => {
  const [value, setValue] = useState("");
  function onChangeHandler(
    event: React.MouseEvent<HTMLButtonElement>,
    selectedValue?: string,
  ) {
    setValue(selectedValue ?? "");
  }

  return (
      <Box margin={4} width="250px" display="flex" flexWrap="nowrap">
        <ButtonToggleGroup
          id="button-toggle-group-id"
          value={value}
          onChange={onChangeHandler}
        >
          <ButtonToggle value="foo">Foo</ButtonToggle>
          <ButtonToggle value="bar">Bar</ButtonToggle>
          <ButtonToggle value="baz">Baz</ButtonToggle>
        </ButtonToggleGroup>
      </Box>
  );
}

export const FocusAndHoverStates: Story = {
  render: ControlledButtonToggleGroup,
    if (!allowInteractions()) {
      return;
    }

    // focus and hover
  },
  decorators: [
    (StoryToRender) => (
      <DefaultDecorator>
        <StoryToRender />
      </DefaultDecorator>
    ),
  ],
};
```

PR #7807, opened 2026-02-27
feat(dialog): align component with fusion DS - FE-7151

DipperTheDan:
### Proposed behaviour

Screenshot 2026-03-12 at 13 09 59

Screenshot 2026-03-12 at 13 09 40

Screenshot 2026-03-12 at 13 09 23

Screenshot 2026-03-12 at 13 09 08

Screenshot 2026-03-12 at 13 14 16

### Current behaviour

Dialog is currently out of alignment with the Fusion DS. 

### Checklist

#### QA

### Additional context

Due to `Dialog` also being used in `AdvancedColorPicker`, `Alert`, `Confirm` and `Pages`, some styling updates will be seen in these components. I have confirmed that this is expected with UX and I have been assured that we can accept the changes in styling that will inevitably happen due to these audit changes.


src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,436 @@
...
+    `}
+`;
+
+const StyledDialogTitle = styled.div<StyledDialogTitleProps>`
+  background: var(--container-standard-bg-default, #fff);
+  padding: var(--global-space-comp-XL, 24px);
+  gap: var(--global-space-comp-L, 16px);
+  border-bottom: 1px solid #a3adb5;
+  border-top-right-radius: var(--borderRadius200);
+  border-top-left-radius: var(--borderRadius200);
```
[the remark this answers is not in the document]
edleeks87: remove the fallback values though

src/components/dialog/dialog.component.tsx
```diff
@@ -1,374 +1,152 @@
...
+      return "medium";
+  }
+};
+
+const mapHighlightVariantToGradientKeyLine = (
+  highlightVariant?: string,
+  gradientKeyLine?: boolean,
+): boolean => {
+  if (gradientKeyLine !== undefined) return gradientKeyLine;
+  return highlightVariant !== undefined && highlightVariant !== "default";
```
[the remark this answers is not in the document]
edleeks87: lets call it `gradient` instead of `ai`

src/components/dialog/__internal__/__next__/dialog.component.tsx
```diff
@@ -0,0 +1,408 @@
...
+    const footerRef = useRef<HTMLDivElement>(null);
+    const { current: titleId } = useRef(createGuid());
+    const { current: subtitleId } = useRef(createGuid());
+
+    const isTopModal = useModalAria(containerRef);
+
+    // Detect small screen for accessibility requirements
+    const isSmallScreen = useMediaQuery("(max-width: 599px)");
+    const shouldDisableSticky = disableStickyOnSmallScreen && isSmallScreen;
+
```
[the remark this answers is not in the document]
edleeks87: Would also be worth exploring a pure CSS approach vs a JS one, it's more efficient and I'm not a huge fan of mixing both approaches

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,436 @@
...
+      `;
+    }
+
+    return css`
+      ${applyDefaultPadding()}
+    `;
+  };
+
+const DialogPositioner = styled.div.attrs(
+  applyBaseTheme,
```
edleeks87, line 102: comment: fine to leave for now but we should have a wider conversation with the rest of the team about how we handle zindexing as ideally we drop the use of the legacy themes in the next versions

src/components/dialog/__internal__/__next__/dialog.component.tsx
```diff
@@ -0,0 +1,408 @@
...
+          additionalWrapperRefs={focusableContainers}
+          autoFocus={!disableAutoFocus}
+          bespokeTrap={bespokeFocusTrap}
+          focusableSelectors={focusableSelectors}
+          focusFirstElement={focusFirstElement}
+          isOpen={open}
+          wrapperRef={containerRef}
+        >
+          <DialogPositioner
+            $size={size}
```
edleeks87, line 348: praise: nice one for handling the transitive props

src/components/alert/alert.pw.tsx
```diff
@@ -16,12 +16,12 @@ import { getDataElementByValue } from "../../../playwright/components";
 const specialCharacters = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
 const viewportHeights = [250, 500, 650];
 const viewportWidths = [
```
edleeks87: comment: if these are already being captured in chromatic we don't need this

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+type StyledDialogFooterProps = SharedTransientProps & {
+  $sticky?: boolean;
+};
+
+type StyledDialogPositionerProps = SharedTransientProps & {
+  $fullscreen?: boolean;
+};
+
+const applyDefaultPadding = () => css`
+  padding: 0 var(--spacing200);
```
edleeks87: comment: we should update these to use the new tokens if possible

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+  @media screen and (min-width: 960px) {
+    padding: 0 var(--spacing400);
+  }
+  @media screen and (min-width: 1260px) {
+    padding: 0 var(--spacing500);
+  }
+`;
+
+const hasPaddingProps = (props: PaddingProps) => {
+  return (
```
edleeks87: Question: do we need to check this? Wouldn't it just override if we did the following


```
const applyContentPadding =
  (disableContentPadding = false) =>
  (props: PaddingProps) => {
    if (disableContentPadding) {
      return css`
        padding: 0;
      `;
    }

    return css`
      ${applyDefaultPadding()}

      ${padding}
    `;
  };

File: src/components/dialog/__internal__/__next__/dialog.style.ts [hunk 9]
```diff
@@ -0,0 +1,502 @@
...
+  box-sizing: border-box;
+  display: block;
+  overflow-y: auto;
+  width: 100%;
+
+  ${({ disableContentPadding, $size, hasHeader, hasFooter, ...props }) =>
+    $size === "fullscreen"
+      ? css`
+          flex: 1;
+          ${applyContentPadding(Boolean(disableContentPadding))(props)}
```
edleeks87: suggestion: we shouldn't need to coerce to Boolean, there's a default value on the function but we dont' really need that as undefined is still falsy etc

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+            ${StyledForm}.sticky {
+              ${StyledFormContent} {
+                ${applyContentPadding(Boolean(disableContentPadding))(props)}
+              }
+              ${StyledFormFooter} {
+                border-top: 1px solid #a3adb5;
+                box-shadow:
+                  0 -2px 4px 0 rgba(0, 0, 0, 0.1),
+                  0 -10px 60px 0 rgba(0, 0, 0, 0.1);
+                background: var(--container-standard-bg-default, #fff);
```
edleeks87:
```suggestion
                background: var(--container-standard-bg-default);
```

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+  overflow-y: auto;
+  width: 100%;
+
+  ${({ disableContentPadding, $size, hasHeader, hasFooter, ...props }) =>
+    $size === "fullscreen"
+      ? css`
+          flex: 1;
+          ${applyContentPadding(Boolean(disableContentPadding))(props)}
+
+          &:has(${StyledForm}.sticky) {
```
edleeks87, line 144: comment (non-blocking): doesn't have to be now as this is unlikely to break as the usage is limited to form and here but it might be worth exploring if we can style the sticky footer if it's a child of StyledDialog

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+          ${!hasFooter &&
+          css`
+            padding-bottom: var(--spacing300);
+          `}
+        `
+      : css`
+          flex-grow: 1;
+          padding: ${disableContentPadding
+            ? "0px"
+            : "var(--spacing300) var(--spacing400)"};
```
edleeks87: comment: legacy tokens

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+        }
+      }
+    `}
+`;
+
+const StyledDialogFooter = styled.div<StyledDialogFooterProps>`
+  box-sizing: border-box;
+  width: 100%;
+  height: var(--global-size-3XL, 72px);
+  padding-top: var(--global-space-comp-L, 16px);
```
edleeks87:
```suggestion
  padding-top: var(--global-space-comp-l);
```

edleeks87: lower case needed here and fallbacks should be removed

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,502 @@
...
+      @media screen and (max-width: ${smallScreenBreakpoint}) {
+        position: static;
+        border-radius: 0;
+      }
+    `}
+`;
+
+const StyledDialogTitle = styled.div<StyledDialogTitleProps>`
+  background: var(--container-standard-bg-default, #fff);
+  padding: var(--global-space-comp-XL, 24px);
```
edleeks87: comment: same as above lowercase and no fallbacks etc

src/components/dialog/dialog.pw.tsx
```diff
@@ -57,12 +57,12 @@ test.describe("Dialog component", () => {
     });
 
     [
-      { size: SIZE.EXTRASMALL, width: "300px" },
-      { size: SIZE.SMALL, width: "380px" },
-      { size: SIZE.MEDIUMSMALL, width: "540px" },
-      { size: SIZE.MEDIUM, width: "750px" },
-      { size: SIZE.MEDIUMLARGE, width: "850px" },
-      { size: SIZE.LARGE, width: "960px" },
+      { size: SIZE.EXTRASMALL, width: "540px" },
```
edleeks87: comment: similar to above this can be removed if there's chromatic

src/components/dialog/dialog.stories.tsx
```diff
@@ -21,7 +21,7 @@ import type { DialogHandle, DialogProps } from ".";
 import Dialog from ".";
 
 const meta: Meta<typeof Dialog> = {
-  title: "Dialog",
+  title: "Deprecated/Dialog",
```
edleeks87:
```suggestion
  title: "Dialog",
```

edleeks87: We should show old and new in local/dev docs but only the public facing one in production etc. To the end consumer there shouldn't be a change when we remove legacy etc

edleeks87 (commented):
Looks like there's a lot of tokens that are uppercase when they should be lower etc, there's also fallbacks which is preventing the undefined tokens from being caught

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -172,33 +149,33 @@ const StyledDialogContent = styled.div.attrs(applyBaseTheme)<
...
-            padding-bottom: var(--spacing300);
+            padding-bottom: var(--global-space-layout-xs);
           `}
         `
       : css`
           flex-grow: 1;
           padding: ${disableContentPadding
             ? "0px"
-            : "var(--spacing300) var(--spacing400)"};
+            : "var(--global-space-layout-xs) var(--global-space-layout-§s)"};
```
edleeks87: typo

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -379,21 +351,21 @@ const StyledDialog = styled.div<StyledDialogProps & ContentPaddingInterface>`
...
+          border-radius: var(--global-radius-container-xl);
           background: ${backgroundColor};
           max-height: 90vh;
           width: 100%;
 
           ${$size === "small" &&
           css`
             min-width: 288px;
-            max-width: var(--container-size-dialog-maxW-S, 540px);
+            max-width: 540px;
```
edleeks87: Can we check with UX/DS to confirm there's no appropriate tokens to use here?

DipperTheDan: Checked with DS. They did have some originally but they were stripped out for hard coded values. I've let the DS know that we'd prefer token values incase these values need updating in future, it'll save us a manual job.

mihai-albu-sage (dismissed):

src/components/dialog/__internal__/__next__/dialog.style.ts
```diff
@@ -0,0 +1,470 @@
...
+                #00d639 0%,
+                #00d6de 40%,
+                #9d60ff 90%
+              );
+              border-radius: var(--global-radius-container-xl)
+                var(--global-radius-container-xl) 0 0;
+            }
+          `}
+
+          @media screen and (max-width: ${dialogSizes[$size]}) {
```
edleeks87, line 408: comment: I'm not sure this will work as the media-query will be evaluated before the value is set as they depend on the cascade

mihai-albu-sage (dismissed):

src/components/dialog/__internal__/__next__/dialog.config.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/** Minimum width for all dialog sizes */
+export const DIALOG_MIN_WIDTH = "288px";
+
+/**
+ * Size values matching Carbon's defined max-width values.
+ * All dialogs have a min-width of 288px.
+ */
+export const DIALOG_SIZE_CONFIG = {
```
edleeks87, line 36: comment: as these are used in the media queries I think we'll need to leave them as pixel values (or add fallbacks here) as they won't have been set when the queries are evaluated. And keep them as tokens where they're setting the component's values

jest.config.ts
```diff
@@ -18,6 +18,8 @@ const baseProjectConfig: Config = {
   moduleDirectories: ["src", "node_modules"],
   moduleNameMapper: {
     "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js",
+    "^@sage/design-tokens-fusion/js/es6/(.*)$":
+      "@sage/design-tokens-fusion/js/common/$1",
```
DipperTheDan, line 22: I have had to add these here to avoid an error of `SyntaxError: Unexpected token 'export'` when running the test suite because Jest runs in a Node CJS environment and ignores node_modules by default, so the export const syntax in the ES6 token files is never transpiled.

edleeks87 (dismissed):

mihai-albu-sage (dismissed):

mihai-albu-sage (dismissed):

dariusbercea-sage (dismissed):

dariusbercea-sage (approved):

edleeks87 (approved):

mihai-albu-sage (approved):

PR #7809, opened 2026-03-02
test(help): refine testing suite

dariusbercea-sage:
### Proposed behaviour

Removes a number of Playwright tests that duplicate either Jest or Chromatic ones for the `Help` component

### Current behaviour

### Checklist

#### QA

### Additional context

### Testing instructions

### DIFF
--- a/src/components/help/help.pw.tsx
+++ b/src/components/help/help.pw.tsx
@@ -5,14 +5,10 @@ import HelpComponentTest from "./component.test-pw";
 import Box from "../../../src/components/box";
 import {
   getDataElementByValue,
-  icon,
   tooltipPreview,
 } from "../../../playwright/components/index";
 import helpComponent from "../../../playwright/components/help";
-import {
-  checkAccessibility,
-  containsClass,
-} from "../../../playwright/support/helper";
+import { checkAccessibility } from "../../../playwright/support/helper";
 import { COLOR, CHARACTERS } from "../../../playwright/support/constants";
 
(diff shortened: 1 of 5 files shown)


src/components/help/help.pw.tsx
```diff
@@ -25,133 +21,7 @@ const colors = [
   ["brown", COLOR.BROWN, COLOR.WHITE],
 ];
 
-test("should have the expected styling when focused", async ({
```
paulrobinson88, line 28: **comment:** It doesn't appear that we are currently capturing the focus style in our Chromatic snapshots, it would be worth adding a pseudo focus within one of the already captured stories so we have this covered.

src/components/help/help.stories.tsx
```diff
@@ -68,6 +67,9 @@ export const WithTooltipColorOverrides: Story = () => (
   </Box>
 );
 WithTooltipColorOverrides.storyName = "With Tooltip Color Overrides";
+WithTooltipColorOverrides.parameters = {
```
paulrobinson88, line 70: **comment:** It looks like the snapshot will be capturing the help icons without the tooltip showing with the custom colour which would already be covered, may need to either use an interaction to focus or follow the same pattern as the `WithTooltipPosition `story above.

src/components/help/help.pw.tsx
dariusbercea-sage: added it to the icons stories, I can maybe add a new one strictly for focus if needed

paulrobinson88 (approved):

PR #7810, opened 2026-03-02
fix(text-editor): clicking placeholder does not focus the editor

mihai-albu-sage:
### Proposed behaviour

Added `pointer-events: none` to the placeholder so that clicks pass through to the editor beneath it, ensuring the editor receives focus when the placeholder is clicked.

### Current behaviour

Clicking the placeholder does not focus the editor, as the placeholder intercepts the click event instead of passing it through to the editor beneath it.

### Checklist

#### QA

### Testing instructions

1. Go to the TextEditor storybook page
2. Click on the placeholder.


src/components/text-editor/text-editor.test.tsx
```diff
@@ -394,6 +394,17 @@ test("placeholder prop renders correctly when placeholder prop is provided", ()
...
 
+test("placeholder does not block editor focus on click", () => {
+  render(
+    <TextEditor labelText="Example" placeholder="This is a nice placeholder" />,
+  );
+
+  const placeholder = screen.getByText("This is a nice placeholder");
+  expect(placeholder).toBeVisible();
+
+  expect(placeholder).toHaveStyle("pointer-events: none");
```
tomdavies73: **suggestion**: Instead of a styling assertion here, I would prefer a test to assert that the editor is focused correctly with a `toBeFocused` assertion

tomdavies73 (commented):
@mihai-albu-sage nice work, just tested locally there and it appears a simple CSS fix has worked. My only other comment would be on the commit, technically the editor is a content editable DIV, not a textarea. Also, the title seems to be a bit long, perhaps something like `fix(text-editor): clicking placeholder does not focus the editor` would be more succinct and accurate 👍.

src/components/text-editor/text-editor.test.tsx
mihai-albu-sage: Changed the commit message and refactored the test to check that the underlying elements receives the focus, when the placeholder is clicked.

src/components/text-editor/text-editor.test.tsx
```diff
@@ -394,6 +396,23 @@ test("placeholder prop renders correctly when placeholder prop is provided", ()
...
+  render(
+    <TextEditor labelText="Example" placeholder="This is a nice placeholder" />,
+  );
+
+  const editor = screen.getByTestId("carbon-rte-editor");
+  editor.setAttribute("tabindex", "0");
+
+  const placeholder = screen.getByText("This is a nice placeholder");
+
+  await userEvent.click(placeholder, {
```
tomdavies73: **suggestion** (non-blocking): I can see because of the attribute we've added that the typical click event can't be fired as we'd like. 

I think there may actually be some benefit in writing this as a `pw.tsx` test instead, as it should be as simple as using the `{force: true}` parameter on a `click()` event. 

This would work if you'd like to give it a try:

```
    test("placeholder does not block editor focus on click", async ({ mount, page }) => {
      await mount(
        <TextEditorDefaultComponent 
          labelText="Example" 
          placeholder="This is a nice placeholder" 
        />,
      );

      const editor = page.getByRole("textbox");
      const placeholder = page.locator("div[data-role='pw-rte-placeholder']");

      await placeholder.click({force: true});

      await expect(editor).toBeFocused();
    });
```

mihai-albu-sage: I agree, and I removed the unit test in favor of the Playwright test. Even though Playwright test might not have seemed necessary in the beginning, writing the test this way makes it clearer and easier to maintain.

tomdavies73 (approved):

PR #7833, opened 2026-03-12
fix(text-editor): text formatting doesn't apply to typography options different than paragraph

mihai-albu-sage:
### Proposed behaviour

This PR fixes an issue where inline formatting (bold/italic/underline) was not applied immediately for typography variants using the custom `StyledSpanNode`. While Lexical correctly updated the editor state, the DOM was not reconciling because `StyledSpanNode.updateDOM()` incorrectly called `super.updateDOM(this, ...)` instead of passing the previous node.

This prevented Lexical from detecting format changes and recreating the DOM element (e.g., span to em or span to strong). The fix passes the correct `prevNode` to `super.updateDOM`, allowing Lexical to properly detect formatting changes and update the DOM.

https://github.com/user-attachments/assets/402d38ed-c275-4a5b-ae7a-9dcbf071d4be

### Current behaviour

The formatting only becomes visible after another action triggers a structural DOM update (for example pressing the unordered list button), which forces Lexical to re-render the node and the formatting then appears.

https://github.com/user-attachments/assets/a44a319b-7e9f-4d55-bda3-34032003a4f9

### Checklist

#### QA


src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
```diff
@@ -211,7 +211,7 @@ export class StyledSpanNode extends TextNode {
     dom: HTMLElement,
     config: EditorConfig,
   ): boolean {
-    let updated = super.updateDOM(this, dom, config);
+    let updated = super.updateDOM(prevNode as this, dom, config);
```
DipperTheDan, line 214: suggestion(non-blocking): You may not need to cast the `prevNode` as `this`. 

The lexical text node type has `updateDOM(prevNode: this, dom: HTMLElement, config: EditorConfig): boolean;`

```suggestion
    let updated = super.updateDOM(prevNode, dom, config);
```

mihai-albu-sage: In our override the function signature expects `prevNode` to be a `StyledSpanNode`, since that’s the node type we’re working with and it extends `TextNode` with additional typography properties.

The base `TextNode.updateDOM` method uses a polymorphic this type (`prevNode: this`), which means it expects the previous node to be the same subtype as the current instance.

TypeScript can’t infer that relationship automatically when we pass `prevNode` to `super.updateDOM`, so the `prevNode as this` cast helps satisfy the type system.

DipperTheDan: Ah that's fine then. Thanks for clarifying this 😄 Nothing is ever straightforward with the TextEditor.

DipperTheDan (approved):

PR #7840, opened 2026-03-17
test(form): refine testing suite

paulrobinson88:
### Proposed behaviour
- Removes tests which are either unnecessary or already covered elsewhere in the testing suite.

### Current behaviour
- There are a large number of unnecessary tests in the Playwright file which are covered elsewhere in the testing suite.

### Checklist

#### QA

### Testing instructions
Check the Playwright tests pass.


mihai-albu-sage (approved):

src/components/form/form-test.stories.tsx
sianford: Can this and the CustomMarginNonTextInputs be in one story, so like the All Inputs story, but with custom margins, which will then mean it will only require one Chromatic snapshot?

src/components/form/form.pw.tsx
```diff
@@ -376,25 +95,7 @@ test.describe("Accessibility tests for Form component", () => {
...
-    mount,
-    page,
-  }) => {
-    await mount(<WithLabelsInline />);
-
-    await checkAccessibility(page);
-  });
-
-  test(`should pass tests for WithWarningsSummary example`, async ({
+  test(`should pass accessibility tests for WithWarningsSummary example`, async ({
```
sianford: I think this one can be removed, as we have the with errors only example and the errors and warnings one. I can't think of any other a11y issue that would be detected in the warnings only example.

src/components/form/form.stories.tsx
sianford, line 90: Can we disable the snapshot in the main stories for this one, `DefaultWithStickyFooter` and `StickyFooterVariant` and add a single story in the Test folder with all three in so Chromatic can take just one snapshot?

sianford, line 245: Can we disable the snapshot in the main stories for this one, `WithWarningsSummary` and `WithErrorsAndWarningsSummary` and add a single story in the Test folder with all three in so Chromatic can take just one snapshot?

sianford: As above, can we add a single story in the Test folder with all three errors/warnings examples in so Chromatic can take just one snapshot?

sianford: As above, can we add a single story in the Test folder with all three errors/warnings examples in so Chromatic can take just one snapshot?

src/components/form/form.stories.tsx
```diff
@@ -172,9 +171,8 @@ export const WithFullWidthButtons: Story = (args: FormProps) => (
 );
 WithFullWidthButtons.storyName = "With Full Width Buttons";
 WithFullWidthButtons.parameters = {
-  themeProvider: { chromatic: { theme: "sage" } },
+  chromatic: { disableSnapshot: true },
 };
-
 export const FieldSpacing: Story = (args: FormProps) => {
```
sianford, line 174: Can Chromatic be disabled for this one and the `OverrideFieldSpacing` examples and a Test story for both be created containing both so Chromatic can take a single snapshot?

src/components/form/form.stories.tsx
sianford, line 209: Can Chromatic be disabled for this one and the `FieldSpacing` examples and a Test story for both be created containing both so Chromatic can take a single snapshot?

paulrobinson88: Good point, updated

paulrobinson88: I've added the this to a test story, it's fine capturing the static footer and the first sticky footer but will have to see what the outcome of the snapshot is for the third.

src/components/form/form.stories.tsx
paulrobinson88: In one story now

src/components/form/form.stories.tsx
sianford: It might need a scroll action adding to scroll to the bottom of the viewport.

PR #7842, opened 2026-03-17
feat(accordion): align to Fusion DS

nuria1110:
### Proposed behaviour

Aligns `Accordion` with Fusion DS and updates styles with fusion-tokens.

- Changes icon type depending on `variant`/`size` selected. 
- Sets "medium" `size` as default. 
- `HeaderSpacing` storybook example shows how to achieve 'No padding' variation. 

"Standard" `variant`:
- Supports `size` "small" and "medium"
image

"Simple" `variant`:
- Renames "subtle" `variant` to "simple".
- Renders "tertiary"  __next__ `Button`.
- "Simple" `variant` supports `size` "small", "medium" and "large":
image

Deprecated props:
- "full" option in `borders`
- "subtle" option in `variant`
- `disableContentPadding`
- `iconType`
- `iconAlign`
- `error`
- `warning`
- `info`

### Current behaviour

`Accordion` is not aligned with Fusion DS. 

### Checklist

#### QA


src/components/accordion/accordion.component.tsx
```diff
@@ -185,96 +230,129 @@ export const Accordion = React.forwardRef<
...
+        data-element="accordion-icon"
+        data-role="accordion-icon"
+        type={iconType || getIconType()}
+        $isExpanded={isExpanded}
+        $allowMotion={allowMotion}
+      />
+    );
+
+    // temporary wrapper for deprecated validation
+    const renderTitleWithValidation = () => (
```
tomdavies73: **suggestion** (non-blocking): honestly we could drop this completely and just let the validation related props have no effect. Depends on how nice we want to be with this, could be worth testing the water to see what the actual consumer impact would be. 

But if you want to leave this in, It's completely fine and won't be a blocker on my end.

src/components/accordion/accordion.component.tsx
```diff
@@ -185,96 +230,129 @@ export const Accordion = React.forwardRef<
...
-          {...headerSpacing}
-        >
-          <StyledAccordionHeadingsContainer
-            data-element="accordion-headings-container"
-            hasValidationIcon={showValidationIcon}
+        {actualVariant === "simple" ? (
+          <Button
+            data-role="accordion-simple-button"
+            id={headerId}
+            aria-expanded={isExpanded}
```
tomdavies73, line 284: **suggestion** (non-blocking): in this ternary here when we're rendering the "simple button" or the title container, bothy have a lot of shared props. We could potentially put all shared props between the two in an object like `titleProps` etc, then spread them onto both instances.

This is just personal preference though, if you prefer to do it like this feel free to leave it.

src/components/accordion/accordion.component.tsx
```diff
@@ -185,96 +230,129 @@ export const Accordion = React.forwardRef<
...
+            aria-controls={contentId}
+            onClick={toggleAccordion}
+            onKeyDown={handleKeyDown}
+            tabIndex={0}
+            ref={ref}
+            role="button"
+            $size={standardSize}
+            $iconAlign={iconAlign}
+            {...headerSpacing}
```
tomdavies73, line 307: **question**: would consumers expect this header spacing to also be applied to the "simple button"?

src/components/accordion/accordion.component.tsx
```diff
@@ -185,96 +230,129 @@ export const Accordion = React.forwardRef<
...
-          isExpanded={isExpanded}
-          maxHeight={contentHeight}
+          $isExpanded={isExpanded}
+          $height={contentHeight}
+          $allowMotion={allowMotion}
           data-role="accordion-content-container"
         >
+          {actualVariant === "simple" && <StyledAccordionLine />}
           <StyledAccordionContent
             role="region"
```
tomdavies73, line 332: **question**: do we need a `role` of `"region"` here? I understand you didn't add this and its a legacy implementation, but an aria-live region seems unnecessary as the other `aria-` attributes we're using should be enough.

src/components/accordion/accordion.component.tsx
```diff
@@ -185,96 +230,129 @@ export const Accordion = React.forwardRef<
...
+              )}
+            </StyledAccordionTitleWrapper>
+          </StyledAccordionTitleContainer>
+        )}
         <StyledAccordionContentContainer
           ref={accordionContentContainer}
-          isExpanded={isExpanded}
-          maxHeight={contentHeight}
+          $isExpanded={isExpanded}
+          $height={contentHeight}
```
tomdavies73, line 326: **suggestion**: I see we're passing `height` down here, which we get from a `useResizeObserver` and a side effect. We then use this height in the styled component file like so 

`${$height}px;`

I think we can avoid doing this by simply making the height of the accordion when expanded fit its content. 

`height: fit-content;`

If this suggestion does work, we can remove this expensive computations which is re-triggered on every re-size event or when the component is expanded.

src/components/accordion/accordion.style.ts
```diff
@@ -4,247 +4,212 @@ import { space, margin } from "styled-system";
...
 `;
 
-export interface StyledAccordionContainerProps {
-  /** Toggles left and right borders, set to none when variant is subtle */
-  borders?: "default" | "full" | "none";
-  /** Sets accordion width */
-  width?: string;
-  /** Sets accordion variant */
-  variant?: "standard" | "subtle";
+interface StyledAccordionContainerProps {
```
tomdavies73: **suggestion** a similar sort of interface is declared a lot here with various different combinations of the transient $ prefix props you've used. You could benefit from using one shared interface, then just picking from that wherever you need below.

src/components/accordion/accordion.style.ts
```diff
@@ -4,247 +4,212 @@ import { space, margin } from "styled-system";
...
+  $borders?: "default" | "full" | "none";
```
tomdavies73, line 23: **suggestion** (non-blocking): even though these are transient props you could link up their types to be that of the OG interface, therefore if more types are ever added in the actual prop these don't have to be updated manually.

src/components/accordion/accordion.pw.tsx
```diff
@@ -1,69 +1,40 @@
...
-  AccordionWithDefinitionList,
-  AccordionWithSplit,
+  AccordionWithSplitButton,
 } from "./components.test-pw";
 
-import { additionalButton as splitAdditionalButtons } from "../../../playwright/components/split-button";
-
 const testData = [CHARACTERS.DIACRITICS, CHARACTERS.SPECIALCHARACTERS];
 
-test.describe("should render Accordion component", () => {
```
tomdavies73, line 29: **suggestion** (non-blocking): after a quick glance through this file we could purge even more tests which are not strictly interaction tests (not including the accessibility tests) e.g. asserting the title or subtitle props work could be removed as these are covered in unit tests and chromatic snapshots

playwright/components/accordion/index.ts
```diff
@@ -1,43 +1,9 @@
...
-
-export {
-  accordionTitleContainer,
-  accordionDefaultTitle,
-  accordionTitleContainerByPosition,
-  accordionIcon,
-  accordionContent,
-  accordion,
-};
+export default accordionDefaultTitle;
```
tomdavies73: **suggestion** we no longer use this in accordions pw tests it seems, just seems to be left in action-popover and split-button. You could just replace those instances with a direct data-element check and delete this entire folder.

src/components/accordion/accordion.component.tsx
nuria1110: I don't mind, thought I would be nice for this one but hopefully no one is using these anymore. I'll probably just remove to clean it up a bit more.

src/components/accordion/accordion.component.tsx
nuria1110: I personally avoid putting props into objects as I've seen in some legacy components they can get quite messy, I'll probably leave it like this, I also think it helps in terms of readability.

src/components/accordion/accordion.component.tsx
nuria1110: Depends I guess, I was mainly repurposing this prop to apply the different header spacing variations in the DS so I didn't think it was really necessary to pass it to both variants.  Since the `title` prop also accepts a node if someone really wanted to change the default button spacing they could just pass their own wrapper with spacing.

src/components/accordion/accordion.component.tsx
nuria1110: I don't think "region" roles come with an aria-live, but I agree that its probably not necessary. I'll double check with the accessibility team.

src/components/accordion/accordion.component.tsx
nuria1110: I'll have another look but I tested quite a few ways to set the height and they all seemed to break the animation in some way -  I'll try to see if I can simplify the events anyways.

mihai-albu-sage (dismissed):

mihai-albu-sage (dismissed):

tomdavies73 (dismissed):

src/components/accordion/accordion.component.tsx
tomdavies73: Yeah mb, got confused

designerlisa:
Checked with @nuria1110 regarding these findings. Those options are deprecated and will be removed in the future breaking change as per the general approach. @ljemmo are we aligned to proceed as-is? 

Once confirmed, I'll move this to QA passed.

ljemmo:
@designerlisa yes thats the expectation. We want to give consuming teams a chance to understand that elements will be removed in the near future and have chance to migrate to alternative properties.

paulrobinson88 (approved):

tomdavies73 (approved):

mihai-albu-sage (approved):

nineteen88 (approved):

PR #7858, opened 2026-03-25
fix(text-editor): span from initialValue not being styled

mihai-albu-sage:
### Proposed behaviour

To correctly apply text styling (`italic`, `bold`, `underline`) across all the typography variants, two issues needed to be addressed:

1. DOM reconciliation after styling changes

We needed to ensure the DOM updates correctly after applying styles to selected text. This was resolved in a previous PR: https://github.com/Sage/carbon/pull/7833

2. Bold styling not reflected visually

When an `initialValue` is provided to the `TextEditor`, the `importDom()` method from `StyledSpanNode` is used. This method returns `StyledSpanNode` elements with default `font-weight` values applied via inline styles.

When the Bold button is clicked in the toolbar, the editor applies a `textBold` class to the selected `StyledSpanNode`. However, because inline styles have higher specificity than CSS classes, the `font-weight` change is not visually reflected.

This issue is specific to bold styling because `font-weight` is applied by default. Italic and underline stylings are not affected, as there are no default inline values for `font-style` (italic) or `text-decoration` (underline) that need to be overridden.

Possible solutions:
- Refactor `StyledSpanNode` to use classes instead of inline styles
Instead of applying `font-weight`, `font-size`, and `line-height` via inline styles when creating `StyledSpanNode`, we could use CSS classes. This would allow future overrides without specificity conflicts. However, this approach would require a significant refactor.

- Remove inline `font-weight` before applying bold
Strip the inline `font-weight` from the selected `StyledSpanNode` before dispatching the bold command. This is more complex, as we would need to store the original value and restore it when bold is toggled off.

- Remove the default value of `400` for `font-weight` inside `importNode()`
Eliminating the default inline value would resolve the conflict, but it could introduce unintended styling changes elsewhere.

- Increase CSS specificity of the bold rule
By adding `!important` to `font-weight: bold;` in the `.textBold` class, we can override the inline style and ensure the bold styling is visible across all typography variants.

Final decision

Although using `!important` is generally discouraged, it is the most practical solution in this case. It allows us to override the default inline `font-weight` consistently across all the typography variants and revert to the original value when bold is toggled off, without requiring major refactoring or added complexity.

### Current behaviour

When the TextEditor is used with `initialValue`, the styling (bold, italic, underline) is not applied across all the typography variants.

### Checklist

#### QA

### Testing instructions

Open the storybook
Open the storybook page : `text-editor--setting-initial-values` or `text-editor-interactions--applies-styling-correctly`

### DIFF
--- a/src/components/text-editor/text-editor.style.ts
+++ b/src/components/text-editor/text-editor.style.ts
@@ -27,7 +27,7 @@ export const StyledWrapper = styled.div`
   position: relative;
 
   .textBold {
-    font-weight: bold;
+    font-weight: bold !important;
   }
 
   .textItalic {
--- a/src/components/text-editor/text-editor-interaction.stories.tsx
+++ b/src/components/text-editor/text-editor-interaction.stories.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { StoryFn, StoryObj } from "@storybook/react";
 import { userEvent, within, expect, waitFor } from "@storybook/test";
 
-import TextEditor, { Mention, MentionsPlugin } from ".";
+import TextEditor, { createFromHTML, Mention, MentionsPlugin } from ".";
 import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
 import DefaultDecorator from "../../../.storybook/utils/default-decorator";
 import CarbonProvider from "../carbon-provider";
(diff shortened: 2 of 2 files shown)


src/components/text-editor/text-editor.style.ts
```diff
@@ -27,7 +27,7 @@ export const StyledWrapper = styled.div`
   position: relative;
 
   .textBold {
-    font-weight: bold;
+    font-weight: bold !important;
```
tomdavies73, line 30: **note**: I was slightly confused at first why we only needed a fix for font-weight here, but yeah after reading your PR comments and using a more recent carbon version on the OG issue reproduction only a fix for bold is needed now 👍 

Also, I couldn't find another nice way to do this with specifity etc. so I'm happy to go with this approach

tomdavies73 (dismissed):
Good stuff @mihai-albu-sage, 

I

DipperTheDan (dismissed):

dariusbercea-sage (dismissed):

dariusbercea-sage (approved):

tomdavies73 (approved):

DipperTheDan (approved):

PR #7870, opened 2026-03-31
fix(a11y): resolve contrast failures for label, hint, legend and button-toggle-group

nineteen88:
### Proposed behaviour

- add aria-disabled to label, hint and legend when set to disabled
- update label, hint, legend and button-toggle-group to use the new tokens for the disabled states

### Current behaviour

- contrast errors are being flagged by axe when button-toggle-group is disabled but this is a more general issue with label and hint when it's set to disabled

### Checklist

#### QA

### DIFF
--- a/src/__internal__/fieldset/fieldset.component.tsx
+++ b/src/__internal__/fieldset/fieldset.component.tsx
@@ -206,6 +206,7 @@ const Fieldset = ({
             align={legendAlignment}
             isRequired={isRequired}
             isDisabled={isDisabled}
+            aria-disabled={isDisabled || undefined}
             data-element="legend"
             data-role="legend"
             isLarge={size === "large"}
@@ -218,6 +219,7 @@ const Fieldset = ({
           <HintText
             id={inputHintId}
             isDisabled={isDisabled}
+            aria-disabled={isDisabled || undefined}
             align={legendAlignment}
             isLarge={size === "large"}
           >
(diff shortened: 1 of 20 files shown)


DipperTheDan (dismissed):

src/__internal__/fieldset/fieldset.component.tsx
```diff
@@ -255,6 +256,7 @@ const Fieldset = ({
                 align={legendAlignment}
                 rightPadding={legendSpacing}
                 {...legendMargin}
+                aria-disabled={isDisabled || undefined}
```
tomdavies73, line 260: **question**: do we need to do this? `isDisabled` is not given a value so it initialises as `undefined`.

We could just pass `isDisabled` like so:

`aria-disabled={isDisabled}` 

Unless I'm missing some context here.

tomdavies73 (commented):
Good stuff @nineteen88!

Just one question from me, also is there any specific reasons why our playwright suite didn't identify these accessibility issues previously?

It could be worth exploring our accessibility tests and seeing if we could add some more to create a baseline against this behaviour 👍

src/__internal__/fieldset/fieldset.component.tsx
nineteen88: It is just for safeguarding purposes. We don't want it to render the aria if for some reason isDisabled was set to false. By adding the OR it ensures we don't include the aria unless explicitly isDisabled is true

DipperTheDan (dismissed):

tomdavies73 (dismissed):

tomdavies73 (approved):

DipperTheDan (approved):

divyajindel (approved):

PR #7888, opened 2026-04-09
feat(switch): align with ds audit - FE-7248

DipperTheDan:
### Proposed behaviour

Screenshot 2026-04-09 at 14 43 24

Screenshot 2026-04-09 at 14 43 05

Screenshot 2026-04-09 at 14 42 56

Screenshot 2026-04-09 at 14 42 50

Screenshot 2026-04-09 at 14 42 44

Screenshot 2026-04-09 at 14 42 39

Screenshot 2026-04-09 at 14 44 53

### Current behaviour

`Switch` is currently not aligned with the current state of the Fusion DS.

### Checklist

#### QA

### Additional context

N/A


src/components/switch/__internal__/__next__/switch.style.tsx
```diff
@@ -0,0 +1,286 @@
...
+    $disabled,
+    $inputHint,
+    $labelInline,
+    $labelSpacing,
+    $labelWidth,
+    $required,
+  }) => css`
+    display: flex;
+    color: ${$disabled
+      ? "var(--input-labelset-label-disabled, rgba(0, 0, 0, 0.42))"
```
nuria1110: **suggestion:** I don't think we are meant to be using fallbacks along with the fusion-tokens anymore as it was pointed out some were inaccurate on figma so we can just remove them. This also masks whether the token is being applied correctly or it it is just depending on the fallback.

src/components/switch/__internal__/__next__/switch.style.tsx
```diff
@@ -0,0 +1,286 @@
...
+      : "var(--input-labelset-label-default, rgba(0, 0, 0, 0.95))"};
+    font-family: var(--core-fontFamilies-component, "Sage UI");
+    font-size: var(--core-fontSize-static-large-step0, 14px);
+    font-style: normal;
+    font-weight: 500;
+    line-height: 150%;
```
nuria1110: **suggestion:** Instead of setting the font styles individually we can just use a font token (not very clear in figma, it is commented above the styles). The medium and large sizes also seem to have different font sizes so they will need different tokens. 

```
   // medium
    font: var(--global-font-static-comp-medium-m)
    
    // large
    font: var(--global-font-static-comp-medium-l)
```

src/components/switch/__internal__/__next__/switch.style.tsx
```diff
@@ -0,0 +1,286 @@
...
+    cursor: pointer;
+    ${$required &&
+    css`
+      ::after {
+        content: "*";
+        color: var(--input-labelset-label-required);
+        font-weight: var(--fontWeights500);
```
nuria1110: **nit-pick:** I don't think we should be using old design-tokens here, if the DS haven't specified a fusion-token I think we could just set it using a value:


```suggestion
        font-weight: 500;
```

src/components/switch/__internal__/__next__/switch.component.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /** Text shown beside the spinner during loading. Defaults to "Processing..." */
+  processingLabel?: string;
+  /** When true, the processing label is rendered below the switch rather than to its right */
+  processingLabelBelowSwitch?: boolean;
+  /** Size of the switch track */
+  size?: "small" | "large";
+  /** When true, the label is displayed inline (beside the switch) rather than above it */
+  labelInline?: boolean;
+  /** Spacing between the label and switch when labelInline is true (multiplier of base spacing unit) */
+  labelSpacing?: 1 | 2;
```
nuria1110: **question:** Is this prop still needed? I don't think the new DS supports different label spacing so not sure if this should also be deprecated.

src/components/switch/__internal__/__next__/switch.component.tsx
```diff
@@ -0,0 +1,222 @@
...
+        {inputHint && (
+          <HintText
+            data-element="input-hint"
+            id={inputHintId.current}
+            isComponentInline={labelInline}
+            marginTop="2px"
+          >
+            {inputHint}
+          </HintText>
+        )}
```
nuria1110: **question:** I see that the hint text is not going to be rendered if the label is inline - is this correct? Seems to be rendered if turned on in figma:

Image

src/components/switch/switch.component.tsx
```diff
@@ -60,298 +58,110 @@ const SwitchComponent = React.forwardRef(
...
       size = "small",
       name,
       adaptiveLabelBreakpoint,
-      tooltipPosition,
+      "data-element": dataElement,
+      "data-role": dataRole,
+      // Deprecated props — destructured to prevent them from being spread
+      reverse,
+      validationOnLabel: _validationOnLabel,
+      labelHelp,
```
nuria1110, line 139: **suggestion:** I think we could map the `labelHelp` and maybe the `fieldHelp` props to `inputHint` so that consumers don't completely lose all functionality when updating: 


```
      inputHint={inputHint || labelHelp || fieldHelp}
```

src/components/switch/switch.component.tsx
```diff
@@ -35,7 +24,7 @@ export interface SwitchProps
   tooltipPosition?: "top" | "bottom" | "left" | "right";
   /** [Legacy] Aria label for rendered help component */
   helpAriaLabel?: string;
-  /** Whether this component resides on a dark background */
+  /** @deprecated Whether this component resides on a dark background */
   isDarkBackground?: boolean;
```
nuria1110: **nit-pick (non-blocking):** We could add a deprecation message that would show in the IDE here:

```suggestion
  /** 
  * Whether this component resides on a dark background 
  * @deprecated This prop is no longer supported. 
  */
  isDarkBackground?: boolean;
```

nuria1110 (commented):
If we aren't doing any mapping we could avoid having an internal next and just export the new component directly, since we are just rendering the next within the "old". This would also avoid having two sets of story and test files, the new files could just replace the old version and we would just need to add the deprecated props to the new interface.

src/components/switch/__internal__/__next__/switch.component.tsx
DipperTheDan: Yeah, this should be deprecated but still work for now. Good spot.

src/components/switch/switch.component.tsx
DipperTheDan: I like the idea but I'll consult UX/DS on this, just incase they are dead set against it.

DipperTheDan:
> If we aren't doing any mapping we could avoid having an internal next and just export the new component directly, since we are just rendering the next within the "old". This would also avoid having two sets of story and test files, the new files could just replace the old version and we would just need to add the deprecated props to the new interface.

We'll see what the second reviewer says about this @nuria1110. I'm inclined to agree but I'd hate to do all that work and then be told to undo it👌🏻

src/components/switch/switch.component.tsx
edleeks87: labelHelp is fine for me, fieldHelp we can either maintain it separately (it's just some text below the input etc) but feel free to chat to UX. I have maintained it as part of the Textbox alignment but easy enough to remove if we get a consensus

src/components/switch/__internal__/__next__/switch.component.tsx
edleeks87: For me it would be fine to fall off but check with UX

src/components/switch/__internal__/__next__/switch.component.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    const internalId = useRef(guid());
+    const inputId = id || internalId.current;
+
+    const labelId = useRef(`${guid()}-label`);
+    const inputHintId = useRef(`${guid()}-hint`);
+
+    const largeScreen = useIsAboveBreakpoint(adaptiveLabelBreakpoint);
+    /* istanbul ignore next */
+    const shouldLabelBeInline = adaptiveLabelBreakpoint
```
edleeks87: comment: I believe we're okay to remove this stuff, I haven't supported it in Textbox

src/components/switch/switch.style.tsx
```diff
@@ -0,0 +1,306 @@
...
+        font: ${$size === "large"
+          ? "var(--global-font-static-comp-medium-l)"
+          : "var(--global-font-static-comp-medium-m)"};
+        margin-left: var(--global-space-comp-xs);
+      }
+    `}
+    /* required to give space above the Switch input when focus styling is applied. */
+    ${!$inputHint &&
+    css`
+      margin-bottom: var(--global-space-comp-s);
```
edleeks87, line 74: suggestion (non-blocking): you could use `gap` here as it's all flex layout

src/components/switch/__internal__/__next__/switch.style.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+interface StyledNextSwitchRowProps {
+  $size?: "small" | "large";
+}
+
+const StyledNextSwitchRow = styled.div<StyledNextSwitchRowProps>`
+  ${({ $size }) => css`
+    display: flex;
+    align-items: center;
+    gap: 8px var(--global-space-comp-s);
```
edleeks87: question: aren't these both 8px?

```suggestion
    gap: var(--global-space-comp-s);
```

src/components/switch/switch.style.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+const StyledNextSwitchRow = styled.div<StyledNextSwitchRowProps>`
+  ${({ $size }) => css`
+    display: flex;
+    align-items: center;
+    gap: 8px var(--global-space-comp-s);
+    /* This is required to give space above the Switch input when focus styling is applied. */
+    ${$size === "large" &&
+    css`
+      margin-top: var(--global-space-comp-xs);
```
edleeks87, line 103: suggestion (non-blocking) : as the switch wrapper is a flex box you could use `gap` to achieve the spacing

src/components/switch/__internal__/__next__/switch.style.tsx
```diff
@@ -0,0 +1,306 @@
...
+    height: ${$size === "large"
+      ? "var(--global-size-m)"
+      : "var(--global-size-xs)"};
+    border-radius: var(--global-radius-container-circle, 999px);
+    border: var(--global-borderwidth-s) solid var(--input-switch-border-default);
+    box-sizing: border-box;
+    transition:
+      background-color 0.2s ease,
+      border-color 0.2s ease;
+    @media (prefers-reduced-motion: reduce) {
```
edleeks87: comment: I think it's better to be consistent here and use the custom hook and a prop to conditionally set the animation or to override it

src/components/switch/switch-interaction.stories.tsx
```diff
@@ -4,15 +4,15 @@ import { userEvent, within, expect } from "@storybook/test";
...
+import Switch from "./switch.component";
 
 import { allowInteractions } from "../../../.storybook/interaction-toggle/reduced-motion";
 import DefaultDecorator from "../../../.storybook/utils/default-decorator";
 
 type Story = StoryObj<typeof Switch>;
 
 export default {
-  title: "Switch/Interactions",
+  title: "Deprecated/Switch/Interactions",
```
edleeks87: comment: We shouldn't be marking this for deprecation, we now have a switch in two sections. There should just be one entry for Switch, as far as consumers know nothing is happening to switch etc

src/components/switch/switch.component.tsx
```diff
@@ -47,6 +44,16 @@ export interface SwitchProps
   checked: boolean;
 }
 
+let switchLegacyWarned = false;
```
edleeks87: comment: With so much being deprecated, I think we can just rely on annotation and IDE to indicate this as some teams are suppressing the logs anyway

edleeks87 (commented):
Agree with @nuria1110 it feels like this could just be one file vs a next one as there's no mapping really

src/components/switch/switch.component.tsx
DipperTheDan: Yeah, I spoke with DS and they were ok for us to map labelHelp and fieldHelp to the inputHint prop.

src/components/switch/__internal__/__next__/switch.style.tsx
DipperTheDan: Yeah, my bad. It's meant to be just the token value. I hard-coded it 8px originally and then added the token value after.

src/components/switch/switch.style.tsx
DipperTheDan: Tried `gap` and it just made the distance between the label text and the required `*` 8px more.

src/components/switch/switch.style.tsx
DipperTheDan: Using `gap` here just changed the distance between the Switch and the `Off`/`On` text. It didn't give the space required between the hint text and the Switch.

edleeks87 (dismissed):

nuria1110 (dismissed):

dariusbercea-sage (approved):

nuria1110 (approved):

edleeks87 (approved):

PR #7898, opened 2026-04-15
feat(tabs): updated to use correct new tokens

harpalsingh:
### Proposed behaviour

Update the Tabs to use the correct Fusion tokens.

### Current behaviour

Mostly using hard coded hex values.

### Visual comparison

**Before** 
Screenshot 2026-04-15 at 16 13 32

**After**
Please note the background is updated to match the app background token.

Screenshot 2026-04-15 at 16 13 15

### Checklist

#### QA

### Testing instructions

View the tabs story and should see no difference from current live build (other than story background update)


src/components/tabs/__next__/tabs.style.ts
```diff
@@ -125,18 +125,14 @@ interface StyledTabProps
...
   border-top-left-radius: 8px;
   border-top-right-radius: 8px;
-
   white-space: nowrap;
   height: fit-content;
-
-  color: var(--colorsYin090);
-
+  color: var(--tab-label-default);
   font-weight: var(--fontWeights500);
```
edleeks87, line 133: comment: we should use the font tokens here and remove font-weight/line-height etc

src/components/tabs/__next__/tabs.style.ts
```diff
@@ -125,18 +125,14 @@ interface StyledTabProps
 export const StyledTab = styled.button<StyledTabProps>`
   background-color: transparent;
   border: none;
-  border-bottom: 2px solid #8b8b8bff;
-
+  border-bottom: 2px solid var(--tab-border-default);
```
edleeks87, line 128: comment: if there are any relevant tokens for radius we should use them

src/components/tabs/__next__/tabs.style.ts
```diff
@@ -88,10 +88,10 @@ export const StyledScrollButton = styled.button<{
   border-radius: 0px;
```
edleeks87: comment: if there are any relevant tokens for radius we should use them

src/components/tabs/__next__/tabs.style.ts
```diff
@@ -104,9 +104,9 @@ export const StyledScrollButtonPlaceholder = styled.div<{
   height: ${({ $size }) => ($size === "medium" ? "40px" : "48px")};
```
edleeks87, line 104: comment: if there are any appropriate tokens we should use them for the height/width

harpalsingh: I would avoid height ones for now until we are certain the ones added are viable and consistent expected usage, so ive left those out for now.

src/components/tabs/__next__/tabs.style.ts
harpalsingh: The font size is currently applied through size mapping option, so its not a token value, so I think we need to resolve that first before we can adjust to using a font token that passes size/line height and weight.

DipperTheDan (approved):

edleeks87 (approved):

divyajindel (approved):

PR #7903, opened 2026-04-16
test(password): refine testing suite

divyajindel:
### Proposed behaviour
- Removes duplicate and unnecessary tests from the `Password` Playwright file.
- Consolidates Chromatic snapshots to be more efficient.

### Current behaviour
- There are a number of duplicated tests in the Playwright file which are covered in other tools
- There are extra snapshots being taken which aren't currently required.

### Checklist

#### QA

### Testing instructions
Ensure Playwright tests pass and Chromatic snapshots are as expected.


src/components/password/password.test.tsx
```diff
@@ -127,6 +127,39 @@ test("should render with provided data- attributes", () => {
   expect(screen.getByTestId("bar")).toHaveAttribute("data-element", "baz");
 });
 
+test("should render with special characters in the label", () => {
```
tomdavies73, line 130: **suggestion**: I don't believe there is much value in testing this, we're just using the value of a text input under the hood. This isn't required for coverage either.

src/components/password/password.test.tsx
```diff
@@ -127,6 +127,39 @@ test("should render with provided data- attributes", () => {
...
+  const label = "Pàßwørd!";
+  render(<Password label={label} value="" onChange={() => {}} />);
+
+  expect(screen.getByLabelText(label)).toBeVisible();
+});
+
+test("should render with special characters in fieldHelp", () => {
```
tomdavies73, line 137: **suggestion**: I don't think there's much value in this either

src/components/password/password.pw.tsx
```diff
@@ -14,451 +9,7 @@ import {
...
-    mount,
-    page,
-  }) => {
-    await mount(<PasswordComponent onChange={() => {}} />);
-
-    const passwordParent = password(page).locator("..");
-    await expect(passwordParent).toHaveCSS("max-width", "100%");
-  });
-});
+import { checkAccessibility } from "../../../playwright/support/helper";
```
tomdavies73, line 12: **suggestion**: I think we can remove a few more accessibility tests, there is no value in testing when margin props are applied. 

Also, we should probably just do one accessibility test for this: 

`"should pass accessibility tests when boolean validations are passed"`

and 

`"should pass accessibility tests when string validations are passed"`

I think we just need the string validations test here as that will include the same styling as the boolean one plus the validation string. So you can remove the boolean test

tomdavies73 (commented):
Nice work @divyajindel, I think we could review our snapshot usage too. `Password` is essentially just `Textbox` which renders a minor button as a child with some added state logic for the obscurity of the visible value. 

We already sufficiently snapshot this specific component behaviour in our interaction stories, we can disable all other snapshots as these overlap the snapshots we already do in `Textbox`.

You can add this to our `password.stories.itsx` file 

`chromatic: { disableSnapshot: true } `

inside the `parameters` object on line 17 👍

PR #7909, opened 2026-04-17
fix(text-editor): styling does not appear in content output from TextEditor

mihai-albu-sage:
### Proposed behaviour

This PR includes the following fixes:

- `importJSON` now restores base `TextNode` fields
Previously, the method only set the custom fields (`text`, `fontWeight`, `fontSize`, `lineHeight`), leaving inherited fields like `format` at their defaults. A node saved with `"format": 8` (underline) would reload with `format: 0`. The method now calls `updateFromJSON` to restore all base `TextNode` fields from the serialized data.

- `exportDOM` now emits format tags
Previously, the method returned a plain `<span>` with inline styles, discarding any active inline formats. It now reads `this.getFormat()` and wraps the span in the appropriate format tags (`<strong>`, `<em>`, `<u>`) based on the node's format bits. Theme classes are not needed here because Lexical applies those automatically via `createDOM` during rendering. The tags alone are sufficient for the import side to reconstruct the correct format bits on deserialization.

- missing styling has been added to the `note.style.ts` to match the classes used by the `TextEditor`.

### Current behaviour

Inline formats (bold, italic, underline) applied to text with custom typography (e.g., `Subtitle`) are lost when the editor state is saved and reloaded. Specifically:

- JSON round-trip: A node with underline (`format: 8`) serializes correctly to JSON, but on reload the format resets to 0 because `importJSON` only restores custom fields and ignores inherited `TextNode` fields.
- HTML export: Exporting the editor state to HTML produces a plain `<span style="...">` with no format tags (`<u>`, `<strong>`, etc.), so any inline formatting is silently dropped from the output.

- Text using the default `paragraph` typography is unaffected, as it uses plain `TextNode` which handles format serialization correctly. The original reported issue regarding underline decoration not being visible on the `paragraph` was due to the missing `.textUnderline` css class. The rest of the issues occur only on `StyledSpanNode`, where custom overrides bypass the base class format handling.

### Checklist

#### QA


src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
```diff
@@ -148,11 +151,31 @@ export class StyledSpanNode extends TextNode {
   }
 
   exportDOM(): DOMExportOutput {
-    const element = document.createElement("span");
-    element.style.fontWeight = this.__fontWeight;
-    element.style.fontSize = this.__fontSize;
-    element.style.lineHeight = this.__lineHeight;
-    element.textContent = this.getTextContent();
+    const span = document.createElement("span");
```
tomdavies73: **suggestion**: we could define this as: 

`let element = document.createElement("span")`

And then re-use this `let`further down below. This prevents us defining a `const` then a `let` just below

src/components/text-editor/text-editor-test.stories.tsx
```diff
@@ -439,6 +440,39 @@ LinkToEditor.parameters = {
   chromatic: { disableSnapshot: true },
 };
 
+export const DisplayHTMLContent: Story = () => {
+  const [, setBodyJsonString] = useState("");
+  const [bodyHTML, setBodyHTML] = useState("");
+  const [isVisible, setIsVisible] = useState(false);
+
+  const onChange = (_value: string, formattedValues: EditorFormattedValues) => {
+    setBodyJsonString(JSON.stringify(formattedValues.json));
```
tomdavies73: **question**: Is there any specific reason why we're updating the state here? If we're not using said state elsewhere I don't feel like there's any practical reason for doing the state update. Unless I'm missing something?

DipperTheDan (commented):
Nothing more to add from me on this. Happy to approve once the comments left by @tomdavies73 have been addressed.

src/components/text-editor/__internal__/__nodes__/styled-span.node.ts
mihai-albu-sage: Removed unnecessary const declaration.

src/components/text-editor/text-editor-test.stories.tsx
mihai-albu-sage: Removed unnecessary state update.

DipperTheDan (approved):

tomdavies73 (approved):

dariusbercea-sage (approved):

PR #7910, opened 2026-04-17
feat(icon): align component with fusion DS - FE-7171

DipperTheDan:
### Proposed behaviour

Align component with current Fusion DS

### Current behaviour

Component is currently not aligned with Fusion DS

### Checklist

#### QA

### Additional context

N/A

### Testing instructions

N/A


src/components/icon/icon.component.tsx
```diff
@@ -73,48 +135,75 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
...
       role,
       ...rest
     }: IconProps,
     ref,
   ): JSX.Element => {
-    const flipBehaviourCheck =
-      Array.isArray(tooltipFlipOverrides) &&
-      tooltipFlipOverrides.every((override) =>
-        ICON_TOOLTIP_POSITIONS.includes(override),
+    if (fontSize && !deprecateFontSizeTriggered) {
```
nuria1110: **comment (non-blocking):** We have quite a large number of deprecated props here, we could just rely on the IDE and docs to indicate deprecations, it would avoid having to add tests for the console warnings too.

src/components/icon/icon.component.tsx
```diff
@@ -30,22 +51,38 @@ export interface IconProps
   id?: string;
   /** The ARIA role to be applied to the Icon */
   role?: string;
-  /** [Legacy] The message to be displayed within the tooltip */
+  /**
+   * @deprecated Tooltip support has been removed from `Icon`. Use a dedicated `Tooltip`
+   * component wrapping the `Icon` instead. This prop no longer has any effect.
+   */
```
nuria1110, line 42: **nit-pick:** I think we could just say we discourage the use of Tooltips in general, since the Tooltip component has been deprecated we would not want to encourage consumers to use it either.

src/components/icon/icon.component.tsx
```diff
@@ -18,8 +19,28 @@ export type LegacyIconTypes =
...
+let deprecateColorTriggered = false;
+let deprecateDisabledTriggered = false;
+let deprecateTooltipPropsTriggered = false;
+
 export interface IconProps
-  extends Omit<StyledIconProps, "type">,
+  extends Omit<
+      StyledIconProps,
+      "type" | "bg" | "bgShape" | "bgSize" | "color" | "disabled" | "fontSize"
+    >,
```
nuria1110: **suggestion:** I think we could use `Pick<>` here instead of `Omit<>` to avoid props being accidentally added to the component's public interface.

Aware this is probably out of scope for this but just thought it was worth noting that instead of extending `StyledIconProps`, I think we should keep these interfaces separate and only define props intended for the public interface in `IconProps`, and `StyledIconProps` only contain props intended to be passed to the styled component internally.

nuria1110 (commented):
I think the `icon-config` file may also need to be updated as it seems to be using some had-coded pixel values and old design-tokens. I was also under the assumption that Icons were minimum 20px now, but I could be wrong, I can't seem to find any size information on the DS.

src/components/icon/icon.component.tsx
DipperTheDan: Yeah the advice was if they really need a workaround for tooltips in the meantime, but I'm happy to follow your wording for discouragement.

src/components/action-popover/action-popover.stories.tsx
DipperTheDan, line 1: I've added the margin-bottom to these examples as due to the slight increase in icon size, it looked like a column of 9 dots. To me, this makes it now look like three separate components and gives the focus border room to breathe.

nuria1110 (dismissed):

src/components/icon/icon.component.tsx
```diff
@@ -136,69 +161,41 @@ const Icon = React.forwardRef<HTMLSpanElement, IconProps>(
...
-    const isFocusable =
-      focusableFromContext !== undefined ? focusableFromContext : focusable;
-    const hasTooltip =
-      !disabled && !disabledFromContext && !!tooltipMessage && isFocusable;
-
-    const computedTabIndex = useMemo(() => {
-      if (isInTab) {
-        return undefined;
+    const resolvedSize = useMemo((): "small" | "medium" | "large" => {
+      if (fontSize) {
```
edleeks87, line 165:
```
 if (fontSize === "extra-large") {
    Logger.warn(
      `[Icon] The \`fontSize\` value "extra-large" is no longer supported and has been mapped to "large".`,
    );
    return "large";
  }

return fontSize ?? size ?? "small";
```

you could even set a default value on size and remove `small` here if needed

src/__internal__/validations/validation-icon.component.tsx
```diff
@@ -120,32 +126,37 @@ export const ValidationIcon = ({
...
-      <Icon
-        aria-describedby={validationTooltipId.current}
-        key={`${validationType}-icon`}
-        type={validationType}
-        tabIndex={tabIndex}
-        tooltipId={validationTooltipId.current}
-        tooltipMessage={validationMessage}
-        tooltipPosition={tooltipPosition}
-        tooltipVisible={
+      <Tooltip
```
edleeks87, line 129: comment: Don't mind this approach as Tooltips falling off the legacy validation makes having the legacy validation pointless and we decided to remove piecemeal etc

src/components/button/button.component.tsx
```diff
@@ -205,12 +206,17 @@ function renderChildren({
           focusable={false}
           target={tooltipTarget}
         >
-          <Icon
-            type={iconType}
-            {...iconProps}
-            tooltipMessage={iconTooltipMessage}
-            tooltipPosition={iconTooltipPosition}
-          />
+          {iconTooltipMessage ? (
```
edleeks87: comment: I'm less keen on this but I think it's probably fine given it's the legacy button, but let's check with @harpalsingh as we might be able to just drop it as I doubt it's as critical as it is for the legacy validation etc

src/components/help/help.pw.tsx
edleeks87, line 1: question: are the accessibility tests no longer valid?

src/components/icon/icon.test.tsx
```diff
@@ -80,15 +81,10 @@ test("renders with a specific role, via the `role` prop", () => {
   expect(icon).toBeVisible();
 });
 
-test("renders a tooltip, populated with a custom value that is passed to the 'tooltipMessage' prop", async () => {
+test("does not render the `tooltipMessage` prop value", () => {
```
edleeks87: question: is there any coverage we'd lose without these tooltip test cases? Feels like they're `false === false` type tests etc so if we can I'd just remove

src/components/icon/icon.style.ts
```diff
@@ -172,29 +187,20 @@ const StyledIcon = styled.span.attrs(applyBaseTheme)<
         display: block;
       }
 
+      // We can remove this fully once we stop supporting tooltips
```
edleeks87: comment: not sure we'll ever hit this as `hasTooltip` is never passed. Maybe we set this if `tabIndex` has been set?

src/components/help/help.pw.tsx
DipperTheDan: Actually they probably are valid. I went a bit wild here and thought we could get rid of a good chunk of Playwright tests. I'll get the accessibility stuff added back in where valid.

src/components/icon/icon.test.tsx
DipperTheDan: I'll see if we can, any that we have to keep I'll comment with a coverage required comment.

src/components/button/button.component.tsx
```diff
@@ -65,10 +63,6 @@ export interface ButtonProps extends SpaceProps, TagProps {
   href?: string;
   /** Defines an Icon position related to the children: "before" | "after" */
   iconPosition?: ButtonIconPosition;
-  /** [Legacy] Provides a tooltip message when the icon is hovered. */
-  iconTooltipMessage?: string;
-  /** [Legacy] Provides positioning when the tooltip is displayed. */
-  iconTooltipPosition?: TooltipPositions;
```
nuria1110, line 71: **comment:** I think we would still need to deprecate these props imo as removing them from the interface would be a breaking change, even if we are removing support.

nuria1110 (dismissed):

edleeks87 (dismissed):

dariusbercea-sage (dismissed):

dariusbercea-sage (approved):

nuria1110 (approved):

edleeks87 (approved):

